PR #20, opened 2023-07-07
Linux fixes

skmp:
Nearly worked on first try!

Also added a check if run with no args.

### DIFF
--- a/src/Core/Memory.cpp
+++ b/src/Core/Memory.cpp
@@ -7,12 +7,24 @@
 	#include <sys/mman.h>
 #endif
 
+#if !defined(_WIN64)
+enum PosixPageProtection
+{
+	PAGE_NOACCESS = 0,
+    PAGE_READONLY          = PROT_READ,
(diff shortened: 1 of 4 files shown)


georgemoralis:
if it is from skmp it should be fine :D

src/Core/Memory.cpp
```diff
@@ -7,12 +7,24 @@
 	#include <sys/mman.h>
 #endif
 
+#if !defined(_WIN64)
+enum PosixPageProtection
+{
+	PAGE_NOACCESS = 0,
+    PAGE_READONLY          = PROT_READ,
```
VelocityRa, line 14: de sou emathe h mama sou na min anakateveis spaces kai tabs

wheremyfoodat: clang-format re malakes

georgemoralis: σιωπη ρε βλαμενα θα μου διωξετε τον dev

PR #21, opened 2023-07-07
Add clang-format

wheremyfoodat:


georgemoralis:
hmm okay and how we apply it to the existing src?

wheremyfoodat:
> hmm okay and how we apply it to the existing src?

`clang-format -i src/**/*.cpp src/**/*.h` will format every single cpp and h file
You can also run it file-by-file if it's easier

PR #22, opened 2023-07-07
ci: Add linux.yml

skmp:
- Builds on windows.yml
- Installs a few needed packages for gh action env
- Does parallel build (using `nproc`)
- Stores main elf and SDL3 as artifacts


wheremyfoodat:
> (will likely need some more packages, incl. libxext-dev
True, but it's probably safe to assume most Linux installations have the Xlib packages

PR #46, opened 2023-10-02
stubbed some HLE functions for openorbis hello world demo (not working)

georgemoralis:
dummy functions to be able to run openorbis , seems to get stuck in execution though

### DIFF
--- a/src/Core/PS4/HLE/LibKernel.cpp
+++ b/src/Core/PS4/HLE/LibKernel.cpp
@@ -26,6 +26,12 @@ namespace HLE::Libs::LibKernel {
 
     static PS4_SYSV_ABI void stack_chk_fail() { BREAKPOINT();
     }
+
+    void sched_yield() { BREAKPOINT();
+    }
+    void sysconf() { BREAKPOINT();
+    }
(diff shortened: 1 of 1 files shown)


skmp:
in Linker.cpp

```c++

static void run_main_entry(u64 addr, EntryParams* params, exit_func_t exit_func) {
    //reinterpret_cast<entry_func_t>(addr)(params, exit_func); // can't be used, stack has to have a specific layout

	asm volatile ("movq %%rsp, %%rbp\n"
        "andq $-16, %%rsp\n"
        "subq $8, %%rsp\n"	// videoout_basic expects the stack to be misaligned

        "pushq 8(%1)\n"		// copy EntryParams to top of stack like the kernel does
        "pushq 0(%1)\n"		// OpenOrbis expects to find it there

		"movq %1, %%rdi\n"	// also pass params and exit func
        "movq %2, %%rsi\n"	// as before

        "jmp *%0\n"			// can't use call here, as that would mangle the prepared stack.
							// there's no coming back
        :
        : "r"(addr), "r"(params), "r"(exit_func)
        : "rax", "rsi", "rdi", "rsp", "rbp"
		);
}
```

PR #48, opened 2023-10-02
stubbed function for hello world OpenOrbis (needs pthread_mutex_lock …

georgemoralis:
Will implemented the rest of needed stuff here

PR #49, opened 2023-10-04
Stubs: Auto add aerolib, add __error to libc, libkernel

georgemoralis:


skmp:
(@georgemoralis i'll fold this to #50, the rest are in #48)

PR #50, opened 2023-10-06
Add auto stubs, update aerolib, faster init

skmp:
### Overview
- Adds a python script to generate the tables from function list
- Auto resolves the stubs

### Details
- Reworked to use the small aerolib, as the full version is too big to include in such a way
- Uses a sorted array instead of std::map

@georgemoralis thoughts?


skmp:
Hmm, looks like clang needs ~ 2GB to compile the file, which we don't have in the runner.

georgemoralis:
i think we should reduce the NID to standard libs only

skmp:
yeah, will re-work with the original list

skmp:
Re worked with the original list ~

georgemoralis:
ci got stucked :D

skmp:
This clearly needs a different approach, CI can't handle such big source files

skmp:
This now uses a slot-based system, and will resolve up to 128 stubs to named stubs. Hopefully solves compile time issues?

skmp:
Compile time seems to be inline with other test runs now. @georgemoralis are there any multi threading concerns around the linker code right now? The stubs code needs a mutex to be multi threading safe.

georgemoralis:
linker class has a mutex  (linker.h) you can do a Lib::LockMutexGuard lock(m_mutex); to make sure

PR #54, opened 2023-10-14
Stubs: Log nids for unknown ones

skmp:
## Overview
Adds support to log stubs with only nid info (up to the stub limit)

Ex:
Screenshot 2023-10-14 190907


georgemoralis:
Ahh very nice

Issue #23, opened 2023-07-11
Initial plan for cpu soft-virtualization


skmp:
### Overview

The idea is to soft-virtualize, by reserving a few registers for the host system, and leaving everything else as is with static register allocation. RIP-relocations also need to be done.

- Decode x86 instructions (core, rex, mmx/sse, and vex encodings)
- Use a register as context and some scratch registers
- Can use similar  branch lookup structures as hex-emu does
- RIP-relative accesses will have to be relocated
- fs and gs segments need to be virtualized

#### Implementing using zydis
- r14 as temp
- r15 as context
- free more temp regs as needed
- RIP REL as double deref with rip moves

#### Kind of patches
- RIP_REL
- +r
- MOD_r
- MOD_m
- SIB_b
- SIB_i

x2  for rex versions
+ vex encodings

#### Notes

Windows TLS doesn't give direct access to TLS slots in the same way linux TLS does, so we'll definitely need a context register.

<details>
<summary> Original writeup </summary>
#### Vasika
- Elafria eikoniki mixani pou doulevei se leitourgia xristi gia x86/64 arxitetoniki ipodoxis
- Vasismeno stous metapinakes apo to hex
- Gennitiki arxitekoniki me metaprogramatismo
- Eikonefsi ton kataxoriton me statiki antistixisi gia tous perisoterous, mono oti spanizei se dinamiki katanomi
- Eikonefsi diefthiseon ektelesis meson paromion domon antistixisis me to hex

#### Epipleon
- Eikonefsi entolon SSE4a pou isos spanizoun

#### Ektimiseis
- ~ 5-15K grammes gia ton pirina + metapinakes
- me xrisi mono metafraseon gramikis ektelesis, ~ 10% fira sta almata rois
- me kapia epipleon poliplokotita kai xrisi diagramaton rois elenxou/ektelesis, < 2% fira?, alla isos kai travlisma kata tin eikonopiisi tis filoksenoumenis arxitektonikis

#### Xronika
- ~ 1w gia PoC
- ~ + 2w gia aploki, gramiki ektelesi
- ~ + 4w gia DRE me polapla simia eisodou 

### Pithano sxedio ergasion
- [ ] Aploiki ilopiisi me ton fortoti ksotikon tou hex + metapinakes + ola apla
- [ ] Anaparagontopiisi meta gia artiotita kai aisthitiki kodika
- [ ] Sindesi me ton fortoti ksotikon gia to PS4

### Sxolia
[...]

PR #56, opened 2023-10-15
virt: Initial implementation using zydis and xbyak

skmp:
### Overview
This is a cut-down version of #23, meant to provide very basic, simple virtualization at the cost of perfomance.

### Goals
- Simple implementation
- Virtualize all registers
- No register allocation
- Support most (all?) ps4 instructions/regs, though just passthrough
- Basic, per thread code cache
- No SMC support
- No instruction rewriting for most operations

### Todo
- [x] Get basic hack-and-slash demo working
- [ ] Decide on a testing stategy
- [ ] Rework to less horrific code
- [ ] Support multiple threads, TCB overflows, and other corner cases gracefully

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -41,3 +41,8 @@
 	path = third-party/xxHash
 	url = https://github.com/Cyan4973/xxHash.git
 	branch = dev
+[submodule "third-party/xbyak"]
+	path = third-party/xbyak
+	url = https://github.com/herumi/xbyak.git
+	shallow = true
+	branch = v6.73
--- a/src/core/PS4/HLE/LibKernel.cpp
+++ b/src/core/PS4/HLE/LibKernel.cpp
@@ -3,6 +3,7 @@
 #include <Util/log.h>
 #include <debug.h>
 #include <windows.h>
+#include <io.h>
 
 #include "Emulator/Util/singleton.h"
 #include "../Loader/Elf.h"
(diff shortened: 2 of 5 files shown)


skmp:
Gets to the first call now!

georgemoralis:
;0

skmp:
video out test will now run, glitchy.

Screenshot 2023-10-23 012140

Also tetris, and openbor gets quite far (crashes at some vsnprintf with null format string)

src/Core/PS4/Linker.cpp
```diff
@@ -593,6 +597,21 @@ void Linker::Relocate(Module* m)
 	}
 }
 
+void GenerateTrampoline(u64 hle_handler, u64 context_base);
+
+__declspec(align(32)) struct Context {
```
raphaelthegreat: Can this use alignas(32) as more portable alternative?

src/Core/PS4/Linker.cpp
```diff
@@ -650,37 +669,561 @@ static PS4_SYSV_ABI void ProgramExitFunc() {
     printf("exit function called\n");
 }
 
+auto gen = new Xbyak::CodeGenerator(64 * 1024 * 1024);
```
raphaelthegreat: std::make_unique

src/Core/PS4/Linker.cpp
```diff
@@ -650,37 +669,561 @@ static PS4_SYSV_ABI void ProgramExitFunc() {
...
+
+
+constexpr auto rip_offs = offsetof(Context, rip);
+constexpr auto ymm_offs = offsetof(Context, ymm[0]);
+constexpr auto ymm_size = 32;
+constexpr auto rflags_offs = offsetof(Context, rflags);
+constexpr auto host_rsp_offs = offsetof(Context, host_rsp);
+constexpr auto trampoline_ret_offs = offsetof(Context, trampoline_ret);
+
+std::unordered_map<u64, u64 (*)()> translated_entries;
```
raphaelthegreat: Better mark this as static if it's not going to be used in other cpp files

raphaelthegreat (commented):
From an architecture perspective, I feel generation could be simplified by making a class that inherits from xbyak::CodeGenerator. Also you can better control the generator lifetime instead of it being in global scope.

georgemoralis:
Vnspritnf use the same algorithm as printf. So if it crashes That will crash as well

georgemoralis:
Printf is special implementation since sysv abi can't parse va-list for arguments

skmp:
@GPUCode this is more or less a hack-and-slash for now, more to explore what needs to be done and less to be merged, for now. The code is also quite horrific as is, it will be rewritten before merging.

A couple of notes so far
- zydis + xbyak isn't a great combination
- x86 decoding, as always, is a huge gotcha within a gotcha
- segment overrides need to be handled
- I have no plan on how to handle multi threaded contexts for now

skmp:
video out sample works now, as well as openbor, and also release builds.

src/Core/PS4/Linker.cpp
```diff
@@ -650,37 +669,556 @@ static PS4_SYSV_ABI void ProgramExitFunc() {
...
 
-        "pushq 8(%1)\n"		// copy EntryParams to top of stack like the kernel does
-        "pushq 0(%1)\n"		// OpenOrbis expects to find it there
+        rsp = rsp & ~16;
+        rsp = rsp - 8;
 
-		"movq %1, %%rdi\n"	// also pass params and exit func
-        "movq %2, %%rsi\n"	// as before
+        rsp = rsp - 8;
+        *(void**)rsp = params->argv;
```
red-prig: In this case, it is not a link to the argv array that should be passed, but the array of argv links itself directly. So the size will be variable.

raphaelthegreat:
> @GPUCode this is more or less a hack-and-slash for now, more to explore what needs to be done and less to be merged, for now. The code is also quite horrific as is, it will be rewritten before merging.
> 
> A couple of notes so far
> 
>     * zydis + xbyak isn't a great combination
> 
>     * x86 decoding, as always, is a huge gotcha within a gotcha
> 
>     * segment overrides need to be handled
> 
>     * I have no plan on how to handle multi threaded contexts for now

Noted, then better leave these notes for future refactor work :p

src/Core/PS4/Linker.cpp
skmp: Yup, can confirm that the array should be inline there from the assembly
```
TranslateCode: 00000009000B0D18, ctx: 7FF613C59720
00000009000B0D18  xor rbp, rbp
00000009000B0D1B  mov rdi, rsp
00000009000B0D1E  lea rsi, [0x000000090010A628
00000009000B0D25  and rsp, 0xFFFFFFFFFFFFFFF0
00000009000B0D29  call 0x00000009000B0D30
TranslateCode: 00000009000B0D30, ctx: 7FF613C59720
00000009000B0D30  mov esi, [rdi]
00000009000B0D32  lea rdx, [rdi+0x08]
```
Funny how that doesn't match with the openorbis source code, must be looking at the wrong place? Hmm.

skmp: with `long* p` Open Orbis uses ```	char **argv = (void *)(p+1);```, seems I missread that last time. Thanks @red-prig :)

skmp:
(Rebased over main, squashed a bit)

To cleanup and merge to main this needs
- [ ] Reorganized out of linker
- [ ] Handling of code buffer overflows
- [ ] Separation between Trampoline Generation and linker
- [ ] Separation of global state and per thread state
- [ ] Cleanup of the codegen code, so it is bearable
- [ ] Support for multiple threads

georgemoralis:
quite old and deprecated will check again sometime

PR #57, opened 2023-10-17
initial file system work

georgemoralis:


src/Core/PS4/HLE/LibC.cpp
```diff
@@ -107,6 +107,9 @@ void PS4_SYSV_ABI qsort(void* ptr, size_t count,size_t size, int(PS4_SYSV_ABI* c
     std::qsort(ptr, count, size, qsort_compair);
 }
 
+void* PS4_SYSV_ABI fopen(const char* filename, const char* mode) { 
+    return std::fopen(filename, mode);
```
georgemoralis, line 111: this needs to connected to fs , since it will return /app0/ prefix , also seems it can also return relative paths

georgemoralis:
redone in a new pr

PR #58, opened 2023-10-19
filesystem work

georgemoralis:
Some work on file system . Focusing on Openbor but might do others as well


georgemoralis:
ps4nes reads from path = /mnt/usb0/roms

games normally reads from /app0/

PR #61, opened 2023-10-22
core: Switch threading primitives to STL types

raphaelthegreat:


src/Core/PS4/HLE/Graphics/Objects/video_out_ctx.cpp
```diff
@@ -83,52 +84,53 @@ bool FlipQueue::submitFlip(VideoConfigInternal* cfg, s32 index, s64 flip_arg) {
...
+        std::unique_lock lock{m_mutex};
+        m_submit_cond.wait_for(lock, std::chrono::microseconds(micros),
+                               [&] { return !m_requests.empty(); });
+        if (m_requests.empty()) {
+            return nullptr;
         }
+        return &m_requests.at(0);  // Process first request
+    }();
+    
+    if (!request) {
```
georgemoralis, line 103: something is not looking correctly here . I think it will always return false

PR #62, opened 2023-10-26
code: Rework elf loader and general cleanup

raphaelthegreat:


src/Emulator/Util/singleton.h
georgemoralis, line 1: i thought that was for removal , but i guess that is just the first step :D

georgemoralis:
oh god let's merge and see what happends :D

PR #64, opened 2023-10-27
log exceptions and flush log file

georgemoralis:
this will log exception and flush log file so that we can have the exception error logged and flushing all log file before exception

### DIFF
--- a/src/Util/log.cpp
+++ b/src/Util/log.cpp
@@ -33,7 +33,7 @@ static LONG WINAPI exception_handler(PEXCEPTION_POINTERS pExp) noexcept {
                 default: break;
             }
             break;
-        case EXCEPTION_ARRAY_BOUNDS_EXCEEDED: LOG_CRITICAL_IF(log_file_exceptions,"Exception EXCEPTION_ARRAY_BOUNDS_EXCEEDED ({}). ", log_hex(ec)); break;
+        case EXCEPTION_ARRAY_BOUNDS_EXCEEDED: LOG_CRITICAL_IF(log_file_exceptions,"Exception EXCEPTION_ARRAY_BOUNDS_EXCEEDED ({:#x}). ", ec); break;
         case EXCEPTION_DATATYPE_MISALIGNMENT: LOG_CRITICAL_IF(log_file_exceptions,"Exception EXCEPTION_DATATYPE_MISALIGNMENT ({}). ", log_hex(ec)); break;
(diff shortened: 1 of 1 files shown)


src/Util/log.cpp
```diff
@@ -7,8 +7,53 @@
...
+        case EXCEPTION_ACCESS_VIOLATION:
+            LOG_CRITICAL_IF(log_file_exceptions,"Exception EXCEPTION_ACCESS_VIOLATION ({}). ", log_hex(ec));
+            switch (pExp->ExceptionRecord->ExceptionInformation[0]) {
+                case 0: LOG_CRITICAL_IF(log_file_exceptions,"Read violation at address {}.", log_hex(pExp->ExceptionRecord->ExceptionInformation[1])); break;
+                case 1: LOG_CRITICAL_IF(log_file_exceptions,"Write violation at address {}.", log_hex(pExp->ExceptionRecord->ExceptionInformation[1])); break;
+                case 8:LOG_CRITICAL_IF(log_file_exceptions,"DEP violation at address {}.", log_hex(pExp->ExceptionRecord->ExceptionInformation[1])); break;
+                default: break;
+            }
+            break;
+        case EXCEPTION_ARRAY_BOUNDS_EXCEEDED: LOG_CRITICAL_IF(log_file_exceptions,"Exception EXCEPTION_ARRAY_BOUNDS_EXCEEDED ({}). ", log_hex(ec)); break;
```
raphaelthegreat:
```suggestion
        case EXCEPTION_ARRAY_BOUNDS_EXCEEDED: LOG_CRITICAL_IF(log_file_exceptions,"Exception EXCEPTION_ARRAY_BOUNDS_EXCEEDED ({:#x}). ", ec); break;
```
fmt can format hex numbers without dedicated functions

src/Util/log.cpp
```diff
@@ -7,8 +7,53 @@
...
 namespace logging {
 std::vector<spdlog::sink_ptr> sinks;
+constexpr bool log_file_exceptions = true;  // disable it to disable logging
+
+void flush() { spdlog::details::registry::instance().flush_all(); }
+
+#ifdef _WIN64
+
+static LONG WINAPI exception_handler(PEXCEPTION_POINTERS pExp) noexcept {
+    const unsigned ec = pExp->ExceptionRecord->ExceptionCode;
```
raphaelthegreat:
```suggestion
    const u32 ec = pExp->ExceptionRecord->ExceptionCode;
```
Best to use the common types everywhere

src/Util/log.cpp
```diff
@@ -20,9 +65,28 @@ int init(bool use_stdout) {
...
+        }
+        flush();
+        if (old_terminate) old_terminate();
+    });
+
     return 0;  // all ok
 }
 
+
+
```
raphaelthegreat:
```suggestion
```

Spaces are not needed here

PR #80, opened 2024-02-16
common: Rework timekeeping with native RDTSC and port to linux

raphaelthegreat:
This PR fixes several build errors on linux and implements some missing behavior that was windows only beforehand. Notably time management has been completely reworked and should be much more accurate both for Windows and Linux.

The previous timer has been replaced with a native clock that reads RDSTC directly, adapted from yuzu. The clock will manually measure the RDTSC frequency, which is used to implement sceKernelGetProcessTimeCounterFrequency. sceKernelGetProcessTime is implemented using clock_gettime from libkernel disassembly because we cannot measure process time very accurately ourselves, it can be implemented on windows with GetProcessTimes. The other functions simply query RDTSC directly or a variant of it so they are simple to implement now.

FPS counting has been stubbed for now, until its re-implemented using std::chrono as the RDTSC clock is a little overkill for it. I've also added a hack to work around wonky synchronization in the vulkan backend that gets exposed under linux, this makes the framebuffer demo work correctly. The whole thing needs a rewrite though


src/core/hle/libraries/libc/libc_math.cpp
```diff
@@ -4,35 +4,35 @@
 namespace Core::Libraries::LibC {
 
 float PS4_SYSV_ABI ps4_atan2f(float y, float x) {
-    return std::atan2f(y, x);
```
wheremyfoodat, line 7: Any particular reason for dropping the std namespace on these

raphaelthegreat: I'm not sure why but the compiler generated errors that these functions do not belong in namespace std, so I removed it for now. From searching online some say it's a libstdc++ bug but that's a bit different to believe.

src/core/hle/libraries/libkernel/libkernel.cpp
```diff
@@ -86,6 +89,14 @@ int PS4_SYSV_ABI sceKernelMmap(void* addr, u64 len, int prot, int flags, int fd,
     }
     *res = ret;
     return 0;
+#else
+    void* result = mmap(addr, len, prot, flags, fd, offset);
+    if (result != MAP_FAILED) {
+        *res = result;
+        return 0;
+    }
+    std::abort();
```
wheremyfoodat, line 98: This will probably compile on all major compilers probably still, but it's UB to have a path that doesn't return a value even if you are guaranteed to abort with `std::abort()`

src/core/linker.cpp
```diff
@@ -658,12 +658,12 @@ void Linker::Resolve(const std::string& name, int Symtype, Module* m, Loader::Sy
 		}
 		else
 		{
-            __debugbreak();//den tha prepei na ftasoume edo
```
wheremyfoodat, line 661: Best to replace these with `std::abort` so that they trigger a breakpoint in the debugger instead of silent failure

src/core/virtual_memory.cpp
```diff
@@ -79,7 +79,8 @@ bool memory_protect(u64 address, u64 size, MemoryMode mode, MemoryMode* old_mode
     }
     return true;
 #else
-#error Unimplement memory_protect function
+    int ret = mprotect(reinterpret_cast<void*>(address), size, convertMemoryMode(mode));
+    return true;
```
wheremyfoodat, line 83: return ret == 0;

wheremyfoodat:
It might also be best to bump from GPLv2 to GPLv3 for better Yuzu-et-al compatibility, not sure why current is v2 (cc @georgemoralis)

raphaelthegreat:
There is also a license problem regarding vulkan at the moment, shadps4 uses the C API headers which are licensed under Apache 2.0 which is incompatible with GPL. This can be solved by switching to the hpp headers that Citra/Panda use which are dual licensed with Apache and MIT, the latter being compatible with GPL

georgemoralis:
doesn't seem to work on windows though :D

PR #84, opened 2024-02-24
Linux CI

qurious-pixel:
create Linux CI workflow


georgemoralis:
As I understand this is the qt not the sdl one.

qurious-pixel:
This is the SDL3 build, but left the parts for Qt when it becomes available. 

Produces the binary and sdl3 artifact, and an appimage with all the libs packaged.

georgemoralis:
i prefer to have different ci for sdl and different for qt , since it will produce different artifacts
so linux.yml and linux-qt.yml probably @GPUCode  what you think?

raphaelthegreat:
Having different CIs for each frontend will become a problem the more platforms that are supported (even now needing 4 targets for just windows and linux sounds a bit much). The Qt and SDL should have different build targets and the CI for each OS should build both of them for each platform to ensure that everything works

georgemoralis:
i still believe it is better to have different ci for the different frontends for the following reasons.
1) Firstly the 2 frontends isn't currently synced , so for emulation side mostly the sdl one is considered atm
2)Secondly it will be more easy to understand which frontend failed.
3)Compilation time will be different so at least for sdl which is faster we can know the results faster than qt.
4)what else do you want to say to convince y :D

qurious-pixel:
No convincing needed. I will take out the Qt parts and rebase for REUSE.

georgemoralis:
i was actually to convince @GPUCode but hmm *somehow* he agreed :D

PR #86, opened 2024-02-26
common: Rewrite logging based on cut down citra logger

raphaelthegreat:
Reworks the previous system and introduces a new logger based on citra. From my view this has multiple benefits:
* Spdlog is no longer needed, thus we get one less dependency which is always a good thing
* All logging is performed asynchronously by default, so the performance should be impacted minimally
* The logger is made with the specific needs of an emulator and has been battle tested for many years. The code is also quite small and easy to understand. Thus it's a good base for a new project, as reliable logging can help during development.
* The previous logger used simple constexpr booleans to turn off logging for various subsystems. This logger supports relatively advanced filters, which allow controlling what gets logged not only globally but per subsystem too. An option for the filter has been added to the config file

I've also imported some additional handy utilities, such as the assertion macros. These are useful for reducing code bloat when checking conditions that should abort the program if they fail, are generally more reliable than normal C asserts and express intent better.

Finally added a small filesystem library based on the excellent filesystem layer yuzu has. It's quite a small subset for now, as the emulator is still quite young and doesn't need much, but these will be very useful for general config and frontend work, filesystem emulation as well (can that be LLEed?). It's a thin wrapper over std::filesystem that does some simple error checking which can be better viewed in the logs and eliminates the exceptions normally thrown by the filesystem API by using error_codes. Other than that, the only other change is that the config file alongside any data shad needs in the future is now stored in a dedicated user folder next to the application, so it can be easily moved in the future.


raphaelthegreat:
Here is an example log file produced by the new logger. The logging format is different from citra, I'm open to suggestions about how the final format should look like
[shad_log.txt](https://github.com/shadps4-emu/shadPS4/files/14398738/shad_log.txt)

georgemoralis:
maybe logging only the filename instead of the whole path? i predict that log file will be huge

src/core/libraries/libscegnmdriver.cpp
```diff
@@ -2,1264 +2,1264 @@
...
-#include "common/log.h"
+#include "common/logging/log.h"
 #include "error_codes.h"
 #include "libscegnmdriver.h"
 
 namespace Libraries::GnmDriver {
 
 int PS4_SYSV_ABI sceGnmAddEqEvent() {
-    PRINT_UNIMPLEMENTED_FUNCTION_NAME();
+    LOG_WARNING(Lib_GnmDriver, "(STUBBED) called");
```
georgemoralis: hmm why changed? the previous was printing the function name , now everyone gets the same output

raphaelthegreat: It will print the name of the function as well for each one

georgemoralis: also why warning? most of the times that call might be fatal  for emulation :D

raphaelthegreat: Can make it into an error yeah. If you have suggestion on what unimplemented function format should look, open to suggestions too

raphaelthegreat:
Also if anyone has comments about the filesystem, would like to hear. The purpose of the wrapper is mostly to avoid exceptions but if it's preferable to always use std::filesystem directly, I can remove it

src/Util/config.cpp
```diff
@@ -13,6 +13,7 @@ bool isNeo = false;
 u32 screenWidth = 1280;
 u32 screenHeight = 720;
 u32 logLevel = 0; // TRACE = 0 , DEBUG = 1 , INFO = 2 , WARN = 3 , ERROR = 4 , CRITICAL = 5, OFF = 6
```
georgemoralis: should be removed

src/core/hle/libraries/libkernel/file_system.cpp
```diff
@@ -11,45 +11,41 @@
...
 
 int PS4_SYSV_ABI posix_open(const char* path, int flags, /* SceKernelMode*/ u16 mode) {
-    LOG_INFO_IF(log_file_fs, "posix open redirect to sceKernelOpen\n");
+    LOG_INFO(Kernel_Fs, "posix open redirect to sceKernelOpen\n");
     int result = sceKernelOpen(path, flags, mode);
-    if (result < 0) {
-        BREAKPOINT(); // posix calls different only for their return values
-    }
+    // Posix calls different only for their return values
+    ASSERT(result >= 0);
```
georgemoralis, line 38: opposite?

PR #105, opened 2024-03-30
URL migration for Build

Xphalnos:
I forwarded the URL for Build ShadPS4 from georgemoralis to shadps4-emu. As the old link was archived, this will allow the document to be modified. I also updated the versions for Windows.


georgemoralis:
nice thank y!

PR #111, opened 2024-04-09
core: Rewrite videoout library and bringup new vulkan backend

raphaelthegreat:
On master the video/graphics is relatively hard to understand as it's split into multiple folders and directories without much cohesion on what every does. In addition the texture cache is extremely basic and works based on hashing: it will track changes to memory regions by computing its hash. This is fine for simple demos but when real games are put to the test, hashing large blocks of memory every draw call isn't going to be fun. The vulkan code was also a bit fragile and broken under wayland, needing a hack to make it synchronize properly.

So this PR does 3 things, it reworks video_out to be more accurate based on my reverse engineering, fully reworks the vulkan backend side of things to have better abstractions that will make the 3d engine implementation easier and fully reworks the texture caching system to be based on fault tracking.

The first part is mostly self explanatory, the implementation has been split into a separate class for easier state management and some additional error codes have been added, but the result isn't all the different from before. Presentation now occurs in the game thread instead of the window thread, which makes things a bit easier. In the future there should be a separate gpu thread to handle all the extra work but that isn't needed here.

The new vulkan backend is based on the Citra one and uses vulkan-hpp instead of raw vulkan as it's a little bit less verbose and solves the previous license problem, as the C headers are licensed under Apache which is incompatible with GPLv2. Vulkan-Hpp on the other hand is licensed with MIT as well which is ok. Like Citra, initialization is handled by the Instance class where all extensions are also loaded, the Scheduler has been ported as well as it will prove useful for parallel shader building in the future and makes validation layer performance a bit less miserable.

The main change here however is the texture cache. When a new image is stored in the cache, the region it owns is marked as protected using an mprotect call. This means that any reads or writes from the guest will go through the texture cache's exception handler, which will allow it to decide on the appropriate action (either invalidation or flush). When the image is requested again, it is validated with an upload and reprotected. In general that's a relatively clean way to handle readbacks or related accesses emulating a UMA system entails. In the future this can be tuned to be better suited for the PS4s memory model, but it's a good base


CMakeLists.txt
```diff
@@ -374,14 +397,18 @@ add_custom_command(TARGET shadps4 POST_BUILD
...
+#add_custom_command(TARGET shadps4 POST_BUILD
+#   COMMAND ${CMAKE_COMMAND} -E copy_if_different
+#     $<TARGET_FILE:SDL3-shared>
+#     $<TARGET_FILE_DIR:shadps4>)
 endif()
 
 if (WIN32)
- add_custom_command(TARGET shadps4 POST_BUILD
-   COMMAND ${CMAKE_COMMAND} -E copy_if_different
+    add_custom_command(TARGET shadps4 POST_BUILD
```
SachinVin: Is there still a need to copy the shared object when you linked it as static?

src/common/config.cpp
```diff
@@ -24,11 +24,11 @@ bool isNeoMode() {
     return isNeo;
 }
 
-u32 getScreenWidth() {
+u32 GetScreenWidth() {
```
SachinVin: Nit, the rest of the project uses camel case, it would be preferable to be consistent

src/core/libraries/kernel/memory_management.h
```diff
@@ -0,0 +1,54 @@
...
+    GpuReadWRite = 0x30,
+};
+
+struct VirtualQueryInfo {
+    uintptr_t start;
+    uintptr_t end;
+    uintptr_t offset;
+    int protection;
+    int memory_type;
+    u32 is_flexible_memory : 1;
```
SachinVin: Can we steal `bitfield.h` from dolphin

src/core/libraries/kernel/memory_management.h
```diff
@@ -0,0 +1,54 @@
...
+    u32 is_direct_memory : 1;
+    u32 is_stack : 1;
+    u32 is_pooled_memory : 1;
+    u32 is_committed : 1;
+    char name[32];
```
SachinVin: Nit std::array

src/core/libraries/kernel/memory_management.h
```diff
@@ -0,0 +1,54 @@
...
+    MapNoCoalesce = 0x400000,
+};
+
+enum class MemoryProtection : u32 {
+    CpuRead = 0x01,
+    CpuReadWrite = 0x02,
+    CpuWrite = 0x02,
+    GpuRead = 0x10,
+    GpuWrite = 0x20,
+    GpuReadWRite = 0x30,
```
SachinVin:
```suggestion
    GpuReadWrite = 0x30,
```

src/core/libraries/kernel/memory_management.h
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+#include "common/types.h"
+
+namespace Core::Kernel {
+
+enum class MemoryTypes : u32 {
+    WriteBackOnion = 0,
+    WriteCombOnion = 3,
+    WriteBackGarlic = 10,
```
SachinVin: Q) Is this not hex?

src/sdl_window.cpp
```diff
@@ -0,0 +1,135 @@
...
+    default:
+        break;
+    }
+}
+
+void WindowSDL::OnResize() {
+    SDL_GetWindowSizeInPixels(window, &width, &height);
+}
+
+void WindowSDL::OnKeyPress(SDL_Event* event) {
```
SachinVin:
```suggestion
void WindowSDL::OnKeyPress(const SDL_Event* event) {
```

src/sdl_window.cpp
```diff
@@ -0,0 +1,135 @@
...
+        button = ScePadButton::SQUARE;
+        break;
+    case SDLK_RETURN:
+        button = ScePadButton::OPTIONS;
+        break;
+    default:
+        break;
+    }
+    if (button != 0) {
+        controller->checKButton(0, button, event->type == SDL_EVENT_KEY_DOWN);
```
SachinVin:
```suggestion
        controller->checkButton(0, button, event->type == SDL_EVENT_KEY_DOWN);
```

src/video_core/texture_cache/types.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include "common/types.h"
+#include "video_core/texture_cache/slot_vector.h"
+
+namespace VideoCore {
+
+using ImageId = SlotId;
+using ImageViewId = SlotId;
+
+struct Offset2D {
```
SachinVin, line 14: Nit, maybe it would be better to use something generic like Vec2 and Vec3 like in Citra

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -0,0 +1,232 @@
...
+        throw std::runtime_error(fmt::format(
+            "Vulkan {}.{} is required, but only {}.{} is supported by instance!",
+            VK_VERSION_MAJOR(TargetVulkanApiVersion), VK_VERSION_MINOR(TargetVulkanApiVersion),
+            VK_VERSION_MAJOR(available_version), VK_VERSION_MINOR(available_version)));
+    }
+
+    const auto extensions = GetInstanceExtensions(window_type, enable_validation);
+
+    const vk::ApplicationInfo application_info = {
+        .pApplicationName = "Citra",
```
psucien: Oops

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+vk::UniqueDebugUtilsMessengerEXT CreateDebugCallback(vk::Instance instance) {
+    const vk::DebugUtilsMessengerCreateInfoEXT msg_ci = {
+        .messageSeverity = vk::DebugUtilsMessageSeverityFlagBitsEXT::eInfo |
+                           vk::DebugUtilsMessageSeverityFlagBitsEXT::eError |
+                           vk::DebugUtilsMessageSeverityFlagBitsEXT::eWarning |
+                           vk::DebugUtilsMessageSeverityFlagBitsEXT::eVerbose,
+        .messageType = vk::DebugUtilsMessageTypeFlagBitsEXT::eGeneral |
+                       vk::DebugUtilsMessageTypeFlagBitsEXT::eValidation |
+                       vk::DebugUtilsMessageTypeFlagBitsEXT::eDeviceAddressBinding |
```
psucien: This guy requires `VK_EXT_device_address_binding_report`

src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -0,0 +1,271 @@
...
+            .timelineSemaphore = true,
+        },
+        vk::PhysicalDeviceExtendedDynamicStateFeaturesEXT{
+            .extendedDynamicState = true,
+        },
+        vk::PhysicalDeviceExtendedDynamicState2FeaturesEXT{
+            .extendedDynamicState2 = true,
+            .extendedDynamicState2LogicOp = true,
+        },
+        vk::PhysicalDeviceExtendedDynamicState3FeaturesEXT{},
```
psucien: This needs `VK_EXT_extended_dynamic_state3`

src/video_core/texture_cache/image.cpp
```diff
@@ -0,0 +1,151 @@
...
+        cmdbuf.pipelineBarrier(vk::PipelineStageFlagBits::eTopOfPipe,
+                               vk::PipelineStageFlagBits::eTopOfPipe,
+                               vk::DependencyFlagBits::eByRegion, {}, {}, init_barrier);
+    });
+
+    const bool is_32bpp = info.pixel_format == vk::Format::eB8G8R8A8Srgb ||
+                          info.pixel_format == vk::Format::eA8B8G8R8SrgbPack32;
+    ASSERT(info.is_tiled && is_32bpp);
+
+    if (Config::isNeoMode()) {
```
psucien: I would prefer to hide size calculation into `ImageInfo` itself

src/video_core/renderer_vulkan/vk_platform.cpp
raphaelthegreat: Nobody saw anything xD

src/video_core/texture_cache/image.cpp
raphaelthegreat: I've done this locally actually, lemme backport it here

src/video_core/renderer_vulkan/vk_platform.cpp
georgemoralis: indeed ;D

PR #113, opened 2024-04-13
src: Reorganize and cleanup libraries

raphaelthegreat:
The library implementations are mess on master, some of them are in hle/libraries, others are in libraries, parts of the kernel in hle/kernel (this is my fault oops). This PRs just puts everything in libraries folder and organizes them in large categories. When each implementation gets big enough it can be split into its separate directory, but for now this makes things easier


CMakeLists.txt
```diff
@@ -236,96 +217,120 @@ set(COMMON src/common/logging/backend.cpp
...
+    src/qt_gui/main_window_themes.cpp
+    src/qt_gui/main_window_themes.h
+    src/qt_gui/main.cpp
+    ${RESOURCE_FILES}
+    )
+endif()
+
 if(ENABLE_QT_GUI)
 qt_add_executable(shadps4
+    ${AUDIO_CORE}
```
SachinVin, line 315: Audio core requires SDL, which I think is not compiled when qt is enabled

georgemoralis: we have fixed that earlier :D

SachinVin: gotcha

PR #114, opened 2024-04-20
Flexible memory and others

georgemoralis:
Initial work on flexible memory , seems many games need that when running at with lle libc 

Now "we are doomed" seems to allocate heaps for libc , we gonna check other games to see what we have


georgemoralis:
code redone in other prs

PR #118, opened 2024-04-30
core: Rewrite thread local storage implementation

raphaelthegreat:
It's not uncommon for ps4 guest applications to launch and use many threads, which also necessitates handling thread local storage properly. In x86 thread local accesses are performed by loading the pointer in the fs segment register. ~~This is a problem as Windows doesn't allow you to change the value of this register to what the guest expects~~. Not quite true, see first reply

On master this is handled with a simple exception handler that will patch the value of the destination register with a thread_local buffer. This works fine but will be a problem later on. Obviously the performance impact is pretty large for any access. In addition, the new texture cache that does fault tracking also needs a custom exception handler, so they end up conflicting. Also, guest apps can use negative offsets when accessing the buffer, so the current implementation would trigger UB in these cases.

This PR attempts to fix all of the above, by using assembly trampolines instead of the exception handler. For storing the TLS image pointer, a new TLS slot is allocated from the parent process and the logic from wine's TlsGetValue is used to retrieve the value. This means we also don't have to rely on undefined/unused spaces in TEB structure to store our data. Each mov instruction from FS segment is patched with a jump to a trampoline that loads the actual pointer.

While at it, also fixed a problem with fault tracking that caused crashing in pngdec demo. The tracking was being performed in the texture cache page size, when it should be on 4KB boundary like the host/guest. Also bumped the cache page size to vastly reduce the amount of page table accesses.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -61,3 +61,6 @@
 [submodule "externals/boost"]
 	path = externals/boost
 	url = https://github.com/raphaelthegreat/ext-boost
+[submodule "externals/xbyak"]
+	path = externals/xbyak
+	url = https://github.com/herumi/xbyak
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -341,46 +341,46 @@ set(QT_GUI
     )
 endif()
 
-if(ENABLE_QT_GUI)
-qt_add_executable(shadps4
-    ${AUDIO_CORE}
-    ${INPUT}
-    ${QT_GUI}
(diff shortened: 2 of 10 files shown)


red-prig:
I’ll make a small clarification that you can change the fs_base value with a special processor instruction, but this turned out to be pointless since any Windows context switch resets this value to its initial state

raphaelthegreat:
Thanks, edited the post to include the clarification

PR #121, opened 2024-05-01
Readme: improvements

Xphalnos:
I improved the readme so that it is more understandable.  I also removed the Build Status because it's not useful to put it in the main readme. I updated GPUCode's nickname to raphaelthegreat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,13 @@ Check building instructions in [Windows build](https://github.com/shadps4-emu/sh
 
 Check building instructions in [Linux build](https://github.com/shadps4-emu/shadPS4/blob/main/documents/linux_building.md)
 
+## Build status
+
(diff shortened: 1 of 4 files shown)


georgemoralis:
doesn't look bad , but why erasing build status ?

Xphalnos:
> doesn't look bad , but why erasing build status ?

I removed the build status because I don't see the point of putting it in the main readme.  If you want me to put it back I can do that.

georgemoralis:
we want it back pls

Xphalnos:
It's okay, I put the build status back.

georgemoralis:
psucien https://github.com/psucien , should be add to the team (after raph)

Xphalnos:
psucien added !

PR #128, opened 2024-05-08
gnm_driver: Gnm eventq and GPU flips

psucien:
This PR implements Gnm EOP event handling and support for GPU submitted flips.
The next functions were added

GnmDriver: 
* `sceGnmAddEqEvent`
* `sceGnmDeleteEqEvent`
* `sceGnmInsertWaitFlipDone`
* `sceGnmSetEmbeddedVsShader`
* `sceGnmSubmitAndFlipCommandBuffers`

VideoOut:
* `sceVideoOutGetBufferLabelAddress`
* `sceVideoOutSubmitEopFlip`

To maintain proper sync between GPU and presenter, a new entity `Platform::IrqController` was introduced.

### DIFF
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -7,6 +7,7 @@
 #include "core/libraries/gnmdriver/gnmdriver.h"
 #include "core/libraries/libs.h"
 #include "core/libraries/videoout/video_out.h"
+#include "core/platform.h"
 #include "video_core/amdgpu/liverpool.h"
 #include "video_core/amdgpu/pm4_cmds.h"
 #include "video_core/renderer_vulkan/renderer_vulkan.h"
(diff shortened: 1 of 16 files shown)


src/core/platform.h
```diff
@@ -0,0 +1,78 @@
...
+struct IrqController {
+    void RegisterOnce(IrqHandler handler) {
+        std::unique_lock lock{m_lock};
+        one_time_subscribers.emplace(handler);
+    }
+
+    void Register(IrqHandler handler) {
+        ASSERT_MSG(!persistent_handler.has_value(),
+                   "Too many persistent handlers"); // Add a slot map if so
+        {
```
raphaelthegreat: The scopes here and in Signal are superfluous

src/core/libraries/kernel/event_queue.cpp
```diff
@@ -20,6 +20,14 @@ int EqueueInternal::addEvent(const EqueueEvent& event) {
     return 0;
 }
 
+int EqueueInternal::removeEvent(u64 id) {
+    const auto& event_q = std::find_if(m_events.cbegin(), m_events.cend(),
+                                       [id](auto& ev) { return ev.event.ident == id; });
```
raphaelthegreat:
```suggestion
    const auto& event_q = std::ranges::find_if(m_events, 
                                       [id](auto& ev) { return ev.event.ident == id; });
```

src/core/libraries/gnmdriver/gnmdriver.cpp
```diff
@@ -1186,44 +1275,142 @@ int PS4_SYSV_ABI sceGnmSqttWaitForEvent() {
...
+    }
+
+    uintptr_t label_addr{};
+    VideoOut::sceVideoOutGetBufferLabelAddress(vo_handle, &label_addr);
+
+    // Write event to lock the VO surface
+    auto* write_lock = reinterpret_cast<PM4CmdWriteData*>(cmdbuf);
+    write_lock->header = PM4Type3Header{PM4ItOpcode::WriteData, 3};
+    write_lock->raw = 0x500u;
+    *reinterpret_cast<uintptr_t*>(&write_lock->dst_addr_lo) =
```
raphaelthegreat: You can probably add a template to Address() member function so this is written like the eop packets for example

PR #127, opened 2024-05-08
Update Workflows

Xphalnos:
Qt 6.6.1 ⇢ Qt 6.7.0
fsfe/reuse-action@v1 ⇢ fsfe/reuse-action@v3
actions/checkout@v3 ⇢ actions/checkout@v4
actions/upload-artifact@v2 ⇢ actions/upload-artifact@v4

PR #129, opened 2024-05-08
update Clang-format.sh

Xphalnos:


Xphalnos:
oops this is an error, This was planned for #127

PR #130, opened 2024-05-09
Kernel stuff & others

georgemoralis:
1. Added winpthreads , from mingw64
2. Removed old vulkan headers
3. Update submodules


Xphalnos:
A brand new version of Vulkan-Headers has just been released. Look: https://github.com/KhronosGroup/Vulkan-Headers/commit/5677bafb820e476441e9e1f745371b72133407d3

georgemoralis:
Thanks but we don't need to do update to submodules every minute :) . I will do it the next time we update something

PR #133, opened 2024-05-14
core: Implement new memory manager

raphaelthegreat:
The previous memory tracker was kinda dodgy and didn't perform much necessary tracking needed to implement some later memory functions. This PR rewrites it to be more accurate to the console.

The biggest problem to solve is the allocation granularity difference between windows and posix. Windows enforces 64KB granularity on VirtualAlloc calls while PS4 can use 16KB granularity. To get around this we use memory placeholders, a relatively new feature in Windows kernel, that allows us to reserve a virtual region and map inside it with 4KB granularity.

Using placeholders has many other benefits. We can reserve the user area upfront which will prevent any emulator allocations from polluting it. ASLR is disabled for this reason, as with it upper regions of user area are polluted from existing mappings. Later games appear to [use](https://github.com/red-prig/fpps4-game-compatibility/issues/1617) dmem aliasing which cannot be emulated with plain VirtualAlloc. To effectively support aliasing, a backing file mapping is created and mapped that represents physical memory the guest application can map. Thus when a physical address is provided this backing file is mapped to the placeholder virtual area.

Tracking is performed using a simple std::map that tracks user and system managed areas. In the future it should be expanded to include the code areas as well but those reside in system reserved area so that's for later. Each virtual memory area stores a bunch of properties about it so VirtualQuery can be implemented relatively easily later. Some special cases with FIXED mapping and sceKernelMunmap have also been implemented


src/video_core/amdgpu/pm4_cmds.h
```diff
@@ -319,7 +319,7 @@ struct PM4CmdEventWriteEop {
         switch (int_sel.Value()) {
         case InterruptSelect::None: {
             // No interrupt
-            break;
+            // break;
```
psucien: You can put it back now

PR #135, opened 2024-05-15
Show title splash while the game is loading

psucien:
This PR adds cover art support to our fancy music player


third-party/stb_image.h
georgemoralis, line 1: that should go to externals/ , third-party is deprecated and will be removed soon

georgemoralis: also needs to be excluded from .reuse/dep5

src/core/libraries/videoout/driver.cpp
```diff
@@ -175,8 +176,11 @@ void VideoOutDriver::Flip(std::chrono::microseconds timeout) {
         requests.pop();
     }
 
-    // Present the frame.
-    renderer->Present(req.frame);
+    if (!renderer->ShowSplash(
+            req.frame)) { // Whatever the game is rendering show splash if it is active
```
raphaelthegreat: Nit, the comment can be placed on top of the if statement so clang format doesn't mess up this badly

src/video_core/renderer_vulkan/renderer_vulkan.cpp
```diff
@@ -157,10 +160,38 @@ void RendererVulkan::RecreateFrame(Frame* frame, u32 width, u32 height) {
...
+        const auto* splash = Common::Singleton<Splash>::Instance();
+
+        VideoCore::ImageInfo info{};
+        info.pixel_format = vk::Format::eR8G8B8A8Srgb;
+        info.type = vk::ImageType::e2D;
+        info.size =
+            VideoCore::Extent3D{splash->GetImageInfo().width, splash->GetImageInfo().height, 1};
+        info.pitch = splash->GetImageInfo().width * 4;
+        info.guest_size_bytes = splash->GetImageData().size();
+        auto& image = texture_cache.FindImage(info, VAddr(splash->GetImageData().data()));
```
raphaelthegreat: Using the texture cache also means registering the image, so using it should ideally be reserved for guest images. You may want to directly upload the image data to the frame instead instead

PR #138, opened 2024-05-17
Migrating from third-party to externals

Xphalnos:
- Minor cleanup
- Rename shadps4 for shadPS4
- Migrating submodules
- Update Vulkan-Headers to latest


qurious-pixel:
Renaming the linux binary (which is case sensitive) should also be changed in the `linux-appimage-qt.sh`,  `linux-appimage-sdl.sh`, and `shadps4.desktop`, but that is non-standard formatting. 
It is better to leave it lower-cased which _is_ expected behavior.

Xphalnos:
Ready for review !

raphaelthegreat (approved):

PR #146, opened 2024-05-26
video_core: initial support for CE and ASC queues

psucien:
In emulation of multiple GPU queues, it is convenient to spawn processing threads to serve submission requests. This approach has limitations as we need to have SH* registers banked, maintain synchronization between threads on wait packets and write several emulation contexts in parallel. The last issue is probably the trickiest one, as having several contexts usually means several host submission queues, which may not be available on a user's system. Also, resources ownership and barriers tracking logic quickly becomes complex. Instead of this, a coroutine based approach is proposed. The submissions served cooperatively in a single thread by a simple round-robin scheduler. This also allows tasks to yield between queue work and CE state progression smoothly. As an additional benefit, we have all GFX and ASCs works serialized in accordance to labels in guest command lists. This makes it possible to record a single host context and submit to one queue only.

This PR also adds support for Constant Engine command lists processing.


raphaelthegreat (approved):
Omg

PR #153, opened 2024-05-30
video_core: multiple render targets binding

psucien:
Added support for:
- passes w\o bound pixel shader
- multiple color attachments
- `NOP` and `RSQ` instructions to recompiler
- color attachments enable and component masks


src/shader_recompiler/ir/opcodes.inc
```diff
@@ -142,6 +142,7 @@ OPCODE(FPRecip64,                                           F64,            F64,
 OPCODE(FPRecipSqrt32,                                       F32,            F32,                                                                            )
 OPCODE(FPRecipSqrt64,                                       F64,            F64,                                                                            )
 OPCODE(FPSqrt,                                              F32,            F32,                                                                            )
+OPCODE(FPInvSqrt,                                           F32,            F32,                                                                            )
```
raphaelthegreat: This opcode already exists (FPRecipSqrt32).

psucien: Oops, opcode blind

PR #164, opened 2024-06-05
core: Properly implement TLS

raphaelthegreat:
Turns out thread local storage implementation on main was completely wrong and a lot of games are starting to get stuck on __tls_get_addr so it's time to actually implement it correctly. This was based on libkernel reversing. The Linker class was a natural place to implement this code, however it was starting to get quite large. So before adding TLS I split off the module loading code into a separate Module class for convenience reasons.

### How does it work
This [article](https://chao-tic.github.io/blog/2018/12/25/tls) has been a pretty big help in understand what is happening, alongside glibc code. The dynamic linker allocates space that contains a TLS block for each module, the size of each block being determined by the PT_TLS [section](https://github.com/bminor/glibc/blob/92c270d32caf3f8d5a02b8e46c7ec5d9d0315158/elf/dl-load.c#L1174). The section also provided an init image for us to copy over, which is often smaller than the actual section size.

The main module block lives right next to the TCB, which is what %FS points to, so the main application can access it directly using negative offsets from FS. For other modules, there exists an additional table, called the DTV table that is indexed by the TLS index of each module and stores the pointer to the TLS block of that module.

However libkernel doesn't seem to quite match what glibc is doing. Each entry in DTV table in glibc is 16-byte [structure](https://github.com/bminor/glibc/blob/92c270d32caf3f8d5a02b8e46c7ec5d9d0315158/sysdeps/generic/dl-dtv.h#L22) and dtv[0] contains the generation counter. In libkernel each DTV entry only seems to be 8 bytes. dtv[0] also holds generation counter but dtv[1] holds the number of allocated DTVs. This means that modules start at index 2 and onwards.

In addition the orbis dynamic linker has a few different paths for TLS bloc allocations. For newer SDK versions (> 1.7) primary thread allocation is done with flexible memory mapping, while secondary threads attempt to use the libc heap function API (probably since libc has already been loaded). Older sdks use plain malloc. Here I've implemented only the newer behaviour for both types of threads.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -287,8 +287,9 @@ set(CORE src/core/aerolib/stubs.cpp
          src/core/linker.h
          src/core/memory.cpp
          src/core/memory.h
+         src/core/module.cpp
+         src/core/module.h
          src/core/platform.h
-         src/core/memory.h
          src/core/tls.cpp
          src/core/tls.h
          src/core/virtual_memory.cpp
--- a/src/common/alignment.h
+++ b/src/common/alignment.h
@@ -9,7 +9,7 @@
 namespace Common {
 
 template <typename T>
-[[nodiscard]] constexpr T alignUp(T value, std::size_t size) {
+[[nodiscard]] constexpr T AlignUp(T value, std::size_t size) {
     static_assert(std::is_unsigned_v<T>, "T must be an unsigned value.");
     auto mod{static_cast<T>(value % size)};
     value -= mod;
@@ -17,14 +17,14 @@ template <typename T>
 }
 
 template <typename T>
-[[nodiscard]] constexpr T alignDown(T value, std::size_t size) {
+[[nodiscard]] constexpr T AlignDown(T value, std::size_t size) {
     static_assert(std::is_unsigned_v<T>, "T must be an unsigned value.");
     return static_cast<T>(value - value % size);
 }
 
 template <typename T>
     requires std::is_integral_v<T>
-[[nodiscard]] constexpr bool is16KBAligned(T value) {
+[[nodiscard]] constexpr bool Is16KBAligned(T value) {
     return (value & 0x3FFF) == 0;
 }
 
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -102,14 +102,14 @@ struct AddressSpace::Impl {
 
         // Perform the map.
         void* ptr = nullptr;
-        if (phys_addr) {
+        if (phys_addr != -1) {
             ptr = MapViewOfFile3(backing_handle, process, reinterpret_cast<PVOID>(virtual_addr),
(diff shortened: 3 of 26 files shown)


raphaelthegreat:
The difference can also be confirmed by looking at __tls_get_addr implementation of libkernel. It does the following, which is a convoluted way of writing (index - 1) * 8 + 16. So first 2 sections of DTV are occupied already, and each DTV entry is 8 bytes
image

red-prig:
```
struct FS_OFFSET {  /* tcb */
    struct FS_OFFSET * tcb_self;      /* +0x00 */
    struct t_dtv     * tcb_dtv;       /* +0x08 */
    struct pthread   * tcb_thread;    /* +0x10 */
    void             * tcb_spare [2]; /* +0x18 */
    void             * tcb_spare2[3]; /* +0x28 */
};

struct t_dtv {
    qword tls_dtv_generation;
    qword tls_max_index;
    void * data[1];
};
```

raphaelthegreat:
Thanks, seem to match what my RE has. I've also noted some members of the Module structure used in libkernel, though it just seems to hold info from PT_TLS
```
Module:
0x20: image_ptr
0x28: provided_image_size
0x2c: tls_image_size
0x30: size_till_fs_base
0x34: alignment
```

red-prig:
```

struct SceKernelModuleSegmentInfo { /* SceKernelModuleSegmentInfo */
    void * addr;
    int size;
    int prot;
};

struct SceKernelModuleInfo { /* SceKernelModuleInfo */
    qword size;
    char name[256];
    struct SceKernelModuleSegmentInfo segmentInfo[4];
    dword segmentCount;
    byte fingerprint[20];
};

struct SceKernelModuleInfoEx { /* SceKernelModuleInfoEx */
    qword st_size;
    char name[256];
    int id;
    dword tls_index;
    void * tls_init_addr;
    dword tls_init_size;
    dword tls_size;
    dword tls_offset;
    dword tls_align;
    void * init_proc_addr;
    void * fini_proc_addr;
    qword reserved1;
    qword reserved2;
    void * eh_frame_hdr_addr;
    void * eh_frame_addr;
    dword eh_frame_hdr_size;
    dword eh_frame_size;
    struct SceKernelModuleSegmentInfo segments[4];
    dword segment_count;
    dword ref_count;
};

struct SceKernelModuleInfoForUnwind { /* SceKernelModuleInfoForUnwind */
    qword st_size;
    char name[256];
    void * eh_frame_hdr_addr;
    void * eh_frame_addr;
    qword eh_frame_size;
    void * seg0_addr;
    qword seg0_size;
};

```
Looks like this is what you mean

raphaelthegreat:
Oh nice didn't know the structures were known. Do you also have the layout of the proc param by any chance? Seems kernel is checking a few things in it

red-prig:
https://github.com/red-prig/fpPS4/blob/f498b0da64d22fe6ea9249f4df6c456dc7f3d8da/sys/elf64.pas#L621

I have already analyzed this multi-level structure in detail

PR #168, opened 2024-06-06
Proposal for shadPS4 0.0.4

Xphalnos:
I think it would be a good idea to upgrade to 0.0.4 because the development since the last version has been huge and we were able to launch our first game. Additionally on Github we are at 360 stars and almost 3000 downloads!

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -23,6 +23,7 @@ jobs:
     - uses: actions/checkout@v4
       with:
           submodules: recursive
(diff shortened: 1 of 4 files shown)


raphaelthegreat:
I agree that a new release should definitely be made, the previous one doesn't even have a renderer. Could you close this PR and make it into an issue instead though, so we can close it when the task is done

Xphalnos:
It's okay, I opened an issue instead

PR #173, opened 2024-06-07
Added Fullscreen mode

Xphalnos:
I added the possibility of putting shadPS4 in Fullscreen. it works but I would need help to make the link between `SDL_SetWindowFullscreen(window, SDL_TRUE);` and the config.toml file. I removed `SDL_SetWindowResizable(window, SDL_FALSE);` because it is initially set to FALSE.


raphaelthegreat:
Oh nice to have this here. You are quite close, simply replace SDL_TRUE with Config::isFullscreenMode() and you should be set

Xphalnos:
It works !

Xphalnos:
Ready for review !

src/common/config.cpp
```diff
@@ -102,6 +108,7 @@ void load(const std::filesystem::path& path) {
             auto general = generalResult.unwrap();
 
             isNeo = toml::find_or<toml::boolean>(general, "isPS4Pro", false);
+            isFullscreen = toml::find_or<toml::boolean>(general, "Fullscreen", true);
```
wheremyfoodat, line 111: This setting should ideally have the same capitalization as other settings (first letter non-capital)

src/sdl_window.cpp
```diff
@@ -17,7 +18,7 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameController* controller_
         UNREACHABLE_MSG("Failed to initialize SDL video subsystem: {}", SDL_GetError());
     }
 
-    const std::string title = "shadps4 v" + std::string(Common::VERSION);
+    const std::string title = "shadPS4 V." + std::string(Common::VERSION);
```
wheremyfoodat: "v0.0.4" probably looks a lot more natural than "V.0.0.4"

src/common/config.cpp
Xphalnos: Yes but this is made up of just one word

PR #177, opened 2024-06-09
TMNT Cowabunga Collection

georgemoralis:
This pr is done by testing TMNT game

- Added posix_pthread_create
- Reworked scekernelCond* functions , this might break something as i don't understant why posix calls asks for e.g broadcast without already call sceKernelCondInit , we do it manually now (as before) but i don't know if it is correct behaviour

TODO: More to come


georgemoralis:
later..

PR #179, opened 2024-06-09
Workflows: Use by default the latest CMake

Xphalnos:
The cmake-version is empty because this allows you to use the latest version by default.
I tested it on my fork and it works


Xphalnos:
The commit is called "Test for CMake" but it works fine.

PR #184, opened 2024-06-10
Tracy profiler

psucien:


psucien:
Okay, this can be considered as "done". During integration, I've faced with two issues though:
1. There is a known bug with Vulkan symbols loading (https://github.com/wolfpld/tracy/pull/639) and since we're using fixed release version 0.10, the fix isn't there yet. I've added it as a patch.
2. Sometimes the server stops with throwing an error that fiber exits on a thread which isn't running it. I tracked thread ids around fiber marks and they all time the same. I believe this is an issue related to on-demand mode (similar to https://github.com/wolfpld/tracy/issues/402) and will check if it will get fixed in the next update. If it will be annoying, we can disable fibers tracking feature in the profiler.

PR #186, opened 2024-06-10
Separation between Release mode and RelWithDebInfo

Xphalnos:
I added a real Release mode for people who need it and shortened the build path.
Tested and working


raphaelthegreat:
Won't changing the build paths mess up the builds for people who have already cloned and built this in Visual Studio?

Xphalnos:
No because it does not modify the code

PR #194, opened 2024-06-12
core: Many things

raphaelthegreat:
Mostly from trying to get Cuphead to show anything
* More shader opcodes
* Dwarf exception handling
* Better dynamic relocation (resolve symbols of modules that depend on dylib)
* Dynamic addition of TLS blocks
* More pthread functions
* Semaphores implementation
* Fixed sceKernelMmap for files
* Flexible memory tracking
* Increase stub limit to 1024 (Unity has way too many :/)
* Blank frame present for videoout


src/core/libraries/gnmdriver/gnmdriver.cpp
```diff
@@ -66,6 +66,9 @@ static inline u32* WriteTrailingNop(u32* cmdbuf) {
 
 s32 PS4_SYSV_ABI sceGnmAddEqEvent(SceKernelEqueue eq, u64 id, void* udata) {
     LOG_TRACE(Lib_GnmDriver, "called");
+    if (id != SceKernelEvent::Type::GfxEop) {
+        return ORBIS_OK;
```
georgemoralis, line 70: shouldn't return an error instead?

raphaelthegreat: Technically shouldn't, but we don't support them atm so just fake success for now

src/core/libraries/libc/libc.cpp
```diff
@@ -463,10 +463,10 @@ void libcSymbolsRegister(Core::Loader::SymbolsResolver* sym) {
 
     // stdio functions
     LIB_FUNCTION("xeYO4u7uyJ0", "libc", 1, "libc", 1, 1, ps4_fopen);
-    LIB_FUNCTION("hcuQgD53UxM", "libc", 1, "libc", 1, 1, ps4_printf);
+    // LIB_FUNCTION("hcuQgD53UxM", "libc", 1, "libc", 1, 1, ps4_printf);
```
georgemoralis, line 466: is there a reason for commenting these?

Issue #166, opened 2024-06-06
Linux compilation errors for gnmdriver & amdgpu


gilcel:
Hi,

FYI: Getting the following errors when trying to compile the latest shadPS4-git source code (3faeba8) on OpenSUSE Tumbleweed (gcc 13.3.0 / clang v18.1.6 / llvm 18.2.2).

```
shadps4-emu/shadPS4-git/src/video_core/amdgpu/liverpool.h:9,
                 from <edited>/shadPS4-git/src/core/libraries/gnmdriver/gnmdriver.cpp:13:
<edited>/shadPS4-git/src/video_core/amdgpu/resource.h: In member function ‘u32 AmdGpu::Buffer::GetStride() const’:
<edited>/shadPS4-git/src/video_core/amdgpu/resource.h:35:28: error: operands to ‘?:’ have different types ‘unsigned int’ and ‘const BitField<48, 14, long unsigned int>’
   35 |         return stride == 0 ? 1U : stride;
      |                ~~~~~~~~~~~~^~~~~~~~~~~~~
<edited>/shadPS4-git/src/video_core/amdgpu/resource.h:35:28: note:   and each type can be converted to the other
<edited>/shadPS4-git/src/video_core/amdgpu/liverpool.h: In member function ‘AmdGpu::NumberFormat AmdGpu::Liverpool::ColorBuffer::NumFormat() const’:
<edited>/shadPS4-git/src/video_core/amdgpu/liverpool.h:683:70: error: operands to ‘?:’ have different types ‘AmdGpu::NumberFormat’ and ‘const BitField<8, 2, AmdGpu::NumberFormat>’
  683 |             return info.number_type == AmdGpu::NumberFormat::Uscaled ? AmdGpu::NumberFormat::Srgb
      |                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
  684 |                                                                      : info.number_type;
      |                                                                      ~~~~~~~~~~~~~~~~~~
<edited>/shadPS4-git/src/video_core/amdgpu/liverpool.h:683:70: note:   and each type can be converted to the other
make[2]: *** [CMakeFiles/shadps4.dir/build.make:510: CMakeFiles/shadps4.dir/src/core/libraries/gnmdriver/gnmdriver.cpp.o] Error 1
[...]

PR #195, opened 2024-06-13
core: Fixup linux support

raphaelthegreat:
Linux support has been missing a few things for a while to be functional. This should make it work better


raphaelthegreat:
Closes https://github.com/shadps4-emu/shadPS4/issues/166

PR #200, opened 2024-06-15
scePad library rewrite

georgemoralis:
Purpose of this pr is 

- [ ] Design an interface to support different input types such as keyboard and gamepad via SDL 
- [ ] Support more than 1 player (2 probably is ideal)
- [ ] Support keymapping


Xphalnos:
For the key mapping, will you put it for the Qt interface or only in the configuration file?

PR #202, opened 2024-06-15
core: Address a few more regressions

raphaelthegreat:
* Actually fix TLS on linux (third time's the charm)
* Force scheduling to SCHED_FIFO again
  > Real time policies need appropriate sched params and the ones in PS4 don't match Linux so it causes functions to error out and guest games don't like that. Revisit it later.
* Fix incorrect Unmap function in Linux


aaalloc:
Hello, I've came across this PR because there seems to be an issue about SCHED_FIFO policies. After the character creation in bloodborn, it get stuck in a unreachable state with
```
[Kernel.Pthread] <Error> thread_management.cpp:scePthreadAttrSetschedpolicy:284: policy=1 not supported by winpthreads
```

forcing SCHED_OTHER policy for ```policy``` fix the issue, ```ppolicy``` is already forced to SCHED_OTHER also.  I wanted to know if there is a specific reason of why it hasn't been made that way and that changes

https://github.com/shadps4-emu/shadPS4/pull/202/commits/7577bd4e77cf438f28a6693b34b41cbec4c784ad#diff-1179139663057cf3ceb4647b8ed2cc9030d0873ccf224290152b49cf2fec2f09R275
```cpp
    if (policy != SCHED_OTHER) {
        LOG_ERROR(Kernel_Pthread, "policy={} not supported by winpthreads", policy);
...
```
has been made like this

PR #208, opened 2024-06-23
Adding Quickstart

Xphalnos:
I fixed a problem displaying screenshots.
I updated the submodules but I think it would be good to do so for Zlib-ng and SDL as they received updates recently.
Currently I'm working on a Quickstart (how to get the latest WIP version, how to unzip PKG files, how to play...)
Ready for changes or review !


georgemoralis:
Nice ! can you also add a section for keyboard mapping? Seems a lot of users doesn't know what to press 🗡️

Xphalnos:
I didn't do the key mapping because there is this commit https://github.com/shadps4-emu/shadPS4/pull/210 which may change the key mapping.
The quick start: https://github.com/Xphalnos/shadPS4/blob/main/documents/Quickstart/Quickstart.md.
The images are not displayed but this is normal because they are not part of the shadPS4 source code.
I will subsequently make other commits to complete this one.

PR #210, opened 2024-06-24
- Initial gamepad support (needs more work..)

raziel1000:
- Initial gamepad support using SDL3 (tested quickly with an xbox one controller and only supports one connected pad for now), vibration should work too
- Added a quick combobox to switch between keyboard and gamepad in Qt version.
- to enable this add controller = 1 in config.toml under [General], or delete your config.toml and let the emu create a new one.
controller = 0 for keyboard
controller = 1 for gamepad


raziel1000:
will reopen

Shoegzer:
Just curious why this was shelved?  It actually seems to work pretty well - even produces rumble effects with a dualsense in linux - but I'm probably missing something. Nice work.

PR #211, opened 2024-06-24
fix system include paths

abouvier:
This make the include paths compatible with system installed headers.


abouvier:
https://github.com/shadps4-emu/ext-cryptoppwin/pull/1 is needed for the Windows builds.

raphaelthegreat:
Windows build seems to be failing

abouvier:
Well, the result is uglier than I thought. The system headers are installed in a subdirectory that is nonexistent in the project source tree, so it must be created by cmake in the build tree. As explained [here](https://github.com/KhronosGroup/glslang/issues/3509#issuecomment-1942529771) for example.

Xphalnos (approved):

psucien:
Thank you!

PR #216, opened 2024-06-26
cmake: prefer system libraries

abouvier:
Is there a reason why `sdl3` and `zlib-ng` were compiled as shared libraries?


Xphalnos (approved):
Approved. Tested on Windows and works.
This reduces the size of the artifacts (Before Windows-Qt was 28.1MB and now it is 27.9. For Windows before it was 5.15MB and now 5.01)

raphaelthegreat (approved):

PR #235, opened 2024-07-04
Fix ShowSplash size

DanielSvoboda:
Fix ShowSplash size

Fix ShowSplash size


Xphalnos (approved):
Approved. Works

psucien:
Thank you for the fix. Since the regression went unnoticed for quite a long time, could you also make the splash enabled by default?

Xphalnos:
It's in the config.cpp file on line 21, you must modify false by true.

PR #236, opened 2024-07-04
Recompiler: sampler patching

psucien:
In some shaders, a pattern has been observed where the anisotropy setting in S# is reset if a sampled texture lacks mips in its view. To address this issue, this PR introduces a heuristic based on pattern detection. Specifically, if expected sharp fields are accessed, the emitted sampler descriptor will have the `disable_aniso` flag set. This allows the runtime to check it later and modify parameters of the bound sampler as needed.

Also, restored BFS in search of image instruction producer.


psucien:
Allows [shadps4-game-compatibility/issues/32](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/32) to pass through splashes.

PR #247, opened 2024-07-06
Improve physical device selection in Vulkan renderer

viniciuslrangel:
Fix when two discrete gpu are present & memory heap size is now considered while sorting physical devices

Hello, it's my first contribution. Feel free to criticise


raphaelthegreat (dismissed):

psucien:
I see two problems there:
- check for heap0 only is not sufficient as GPU may have several heaps and the largest one not necessarily to be the first in the properties
- assume you have two GPUs listed in the order: 0 - integrated, 1 - discrete. And the integrated one has more heap0 memory. Sort will fail with comparator error (at least on debug builds)

viniciuslrangel:
@psucien I sorted first ensuring the API version and the largest dedicated(local) heap

PR #250, opened 2024-07-06
Misc implementations and fixes.

polybiusproxy:
This PR implements both the `sceKernelFtruncate` and `sceKernelUnlink` functions from `libkernel.sprx`.
As seen on games using PSPHD (Sony's PSP in-house emulator) - tested with CUSA05289.

PSPHD won't boot due to its start-up routine; at launch, it will try to create a temporary file, resize it - then delete it, and reserve (with `sceKernelReserveVirtualRange`) and map a ~4GB block of memory for that file.
This PR should improve PSPHD games by implementing the resize and delete methods.

Improves [(shadps4-game-compatibility#124)](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/124)


src/core/libraries/kernel/file_system.cpp
```diff
@@ -359,6 +391,29 @@ s32 PS4_SYSV_ABI sceKernelFsync(int fd) {
...
+    }
+
+    if (file->m_host_name.empty()) {
+        return SCE_KERNEL_ERROR_EACCES;
+    }
+
+    // CUSA05289 (PSPHD emulator): Tries to resize a temporary file after deleting it
+    try {
+        // TODO: Check for sufficient space?
+        std::filesystem::resize_file(file->m_host_name, length);
```
psucien: It would be better to use an overload which returns `std::error_code` instead of exception handling

polybiusproxy:
Marking as draft until I can get an accurate implementation done.

polybiusproxy:
I have now reworked the PR so the functions are accurate (**`sceKernelUnlink` lacks a Linux implementation**) and implemented `sceKernelReserveVirtualRange`. In addition, I have moved the NT API functions to a separate common header, and have removed an assert on `MemoryManager::MapFile`, as games can call `sceKernelMmap` without the address being NULL (on real hardware the address is used as a hint for allocating memory if not null).

This should now improve all PSPHD games to the point that the PSP emulation starts, and games using `sceKernelReserveVirtualRange`.

Improves (https://github.com/shadps4-emu/shadps4-game-compatibility/issues/124), (https://github.com/shadps4-emu/shadps4-game-compatibility/issues/140), and LocoRoco Remastered.
Should also improve Dreams (CUSA08010).

raphaelthegreat:
The sceKernelReserveVirtualRange implementation doesn't look correct to me. The function is supposed to reserve a virtual address range for later FIXED mapping to occupy and doesn't map memory itself (so it prevents the space being taken by mmaps with addr = 0). A better way would be to add a new VMAType::Reserved and a new function in the memory system that has similar logic to MapMemory but carves a reserved block in the virtual address range. And also account of it in the assert of FIXED mappings and AddMapping

raphaelthegreat:
Also if you need to remove the assert in MapFile you also have to adjust the line below to start searching at the specified virtual address, otherwise it will search from the start of system managed

src/common/io_file.cpp
```diff
@@ -216,6 +218,26 @@ void IOFile::Close() {
...
+    FILE_DISPOSITION_INFORMATION disposition;
+    IO_STATUS_BLOCK iosb;
+
+    const int fd = fileno(file);
+    HANDLE hfile = reinterpret_cast<HANDLE>(_get_osfhandle(fd));
+
+    disposition.DeleteFile = TRUE;
+    NtSetInformationFile(hfile, &iosb, &disposition, sizeof(disposition),
+                         FileDispositionInformation);
+#endif
```
raphaelthegreat, line 245: Add a #else here with UNREACHABLE() so when this is called on linux it won't silently fail

src/common/ntapi.h
```diff
@@ -0,0 +1,123 @@
...
+    };
+    ULONG_PTR Information;
+} IO_STATUS_BLOCK, *PIO_STATUS_BLOCK;
+
+typedef struct _FILE_DISPOSITION_INFORMATION {
+    BOOLEAN DeleteFile;
+} FILE_DISPOSITION_INFORMATION, *PFILE_DISPOSITION_INFORMATION;
+
+// http://stackoverflow.com/a/31411628/4725495
+static u32(__stdcall* NtDelayExecution)(BOOL Alertable, PLARGE_INTEGER DelayInterval) =
```
raphaelthegreat: Won't this execute for each compilation unit that includes this header? Maybe move this to a ntapi.cpp file and have an initialization function.

polybiusproxy: Shouldn't be the case due to the `#pragma once` above.

raphaelthegreat: pragma once iirc only prevents multiple inclusions in the same source file. In different cpp files it should duplicate still I think

polybiusproxy: Now the initialization has been moved to a source file.

raphaelthegreat (approved):
Thanks!

polybiusproxy:
To summarize - this PR has:
- Implemented `sceKernelFtruncate`, `sceKernelUnlink`.
- Implemented `sceKernelReserveVirtualRange`.
- Fixed a bug where file handles were always mapped as read-only on Windows.

Now all PSPHD games should boot up to the point where the PSP emulation starts.
Improves all PSPHD games mentioned earlier (PaRappa, LocoRoco, Patapon, etc.).

PR #254, opened 2024-07-06
fix linux again

DanielSvoboda:
added the shortcut extension for linux ".desktop" and fix name as suggested by qurious


qurious-pixel:
LGTM

PR #255, opened 2024-07-07
strdup for Windows

DanielSvoboda:
Adds an implementation of the strdup function for the Windows environment, where it is not standard. This implementation uses strcpy_s to ensure safe string copying. The code is guarded with #ifdef _WIN32 to ensure it compiles only on Windows platforms.

### DIFF
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include <cstdlib>
+#include <cstring>
 #include "common/assert.h"
 #include "common/logging/log.h"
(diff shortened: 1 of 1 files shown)


georgemoralis:
looking good but where it doesn't work? vstudio seems to compile it just fine

DanielSvoboda:
in https://github.com/shadps4-emu/shadPS4/blob/main/src/core/libraries/kernel/file_system.cpp#L146
For me in VS, in 'sceKernelWrite', it appears as 'the identified "strdup" is not defined.". this appears in Windows but the compilation is still carried out...

DanielSvoboda:
I think this is shown because of my compiler, strdup should work in MinGW, but I'm using MSVC, if I just switch to `_strdup` it should work in both. I'm using visual studio.

georgemoralis:
looks like an issue with intelisense rather than missing function.

DanielSvoboda:
yes, it is very likely, I will close the 'pull'

PR #256, opened 2024-07-07
Forwarding V_MAX_LEGACY_F32 to V_MAX_F32. 

CoreyHendrey:
Just forwarded the legacy op, as I can see done for other legacy ops.

Fixes Translation error in Geometry Wars 3.


psucien:
Unfortunately, current implementation of legacy instructions is a leftover, and they should be reworked. You can still use such hack locally to progress further in game, but for the upstream version we will prepare a proper implementation.

PR #258, opened 2024-07-07
sceSaveDataDirNameSearch: Removed early failure exit

CoreyHendrey:
Geometry wars will throw a save corrupt error with this early exit, however, if functions perfectly if we let this flow through.
I am not sure why the early exit was implemented originally, was it just to make sure the log didn't throw an nullptr exception?


georgemoralis:
#281 seems to have better implementation if you want to check

PR #267, opened 2024-07-08
Fix sceAudioOutOpen not handling audio param attributes

viniciuslrangel:
and returning error incorrectly when some mask is applied


georgemoralis:
looking good ;)

PR #270, opened 2024-07-09
WIP: Started implementing Ult library

CoreyHendrey:


raphaelthegreat:
For new modules would also recommend using module generator https://github.com/shadps4-emu/moduleGenerator which generates functions and ensures all NIDs are correct (signatures still need to be made manually). Ult library also uses fiber library underneath so you can probably use boost fibers here

CoreyHendrey:
Ah, I looked for the module generator and couldn't find it haha. Thanks!

I'll look into the Fiber boost library. I noticed it all uses a lot of calls into the existing kernel threading that we already have, so that should be useful.

raphaelthegreat:
Yeah. Ult can actually be run as LLE if libSceFiber is available (which only works on linux atm as LLE). So you can probably easily translate calls to that or implement scefiber and call it from here like real module does (will allow games like persona to work without lle modules). Can see which one is easiest tbh

raphaelthegreat:
Any updates on this?

georgemoralis:
are you gonna continue work on this? If not we should be close that since it is a very initial work

georgemoralis:
close due to inactivity , if needed we will reopen it

PR #279, opened 2024-07-11
use new bloodborne screenshot

shubhisroking:


shubhisroking:
ah, the pull request is not yet created for the color fix. I will reinitiate this PR once it has been created and merged

im-tem:
even after the fix pr is merged it would make sense to only update the readme after the version with a fix is actually released to general public else it'd be false advertising 😅

shubhisroking:
> even after the fix pr is merged it would make sense to only update the readme after the version with a fix is actually released to general public else it'd be false advertising 😅

yeah, guess we gotta wait then

PR #290, opened 2024-07-12
Fix pthread deprecation warnings

jas0n098:
pthread_attr_(set,get)stackaddr are marked as deprecated in glibc and musl libc doesn't implement them, causing build failures. pthread_attr_(set,get)stack was used to replace them, as recommended by pthread docs.
While I haven't observed any games break that I have, it might still be a good idea to test this before merging.

### DIFF
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -318,7 +318,8 @@ int PS4_SYSV_ABI scePthreadAttrGetstackaddr(const ScePthreadAttr* attr, void** s
         return SCE_KERNEL_ERROR_EINVAL;
     }
 
-    int result = pthread_attr_getstackaddr(&(*attr)->pth_attr, stack_addr);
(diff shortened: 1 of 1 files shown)


src/core/libraries/kernel/thread_management.cpp
```diff
@@ -318,7 +318,10 @@ int PS4_SYSV_ABI scePthreadAttrGetstackaddr(const ScePthreadAttr* attr, void** s
         return SCE_KERNEL_ERROR_EINVAL;
     }
 
-    int result = pthread_attr_getstackaddr(&(*attr)->pth_attr, stack_addr);
+    size_t stack_size = 0;
+    pthread_attr_getstacksize(&(*attr)->pth_attr, &stack_size);
```
raphaelthegreat: This getstacksize call is superfluous I think

jas0n098: You're right, getstack would set the stack size just like getstacksize would

PR #294, opened 2024-07-15
Add macOS support

squidbus:
Adds basic support for macOS. Currently requires a minimum of macOS 11 for Intel Macs, and macOS 15 for M series Macs for the AVX support added to Rosetta 2.

Summary of changes:
* Fixes for some compile issues on macOS, such as missing includes and lack of `std::chrono::time_zone`
* CPU/HLE changes:
  * Address space mapping code paths for missing `memfd_create` and inability to map to the desired fixed location
  * Code patches for missing BMI1 instructions when running under Rosetta 2. Rather than using fixed byte patterns, it uses Zydis to disassemble and search for instructions
  * Thread local storage patches for macOS. Also moved these patches to the new system created for the other code patches
  * Using `SIGBUS` for the texture cache signal handler instead of `SIGSEGV`
  * `pthread_mutex_timedlock` polyfill when unavailable
* GPU changes:
  * Reducing the requirements from Vulkan 1.3 to Vulkan 1.2 with the needed extensions that were rolled into 1.3, as MoltenVK has not reached full 1.3 support yet
  * Support for Metal-based Vulkan surfaces
  * Support for using alternative formats and swizzles when a surface format is unavailable
  * Fix for an off-by-one issue with accessing shader user data
* Misc
  * SDL face button mappings that are usable on keyboards without a keypad, e.g. Macbook keyboards
  * GitHub Actions workflows for macOS builds

Updated TLS patches are not fully tested on other platforms as I'm not currently set up to do so.

screenshot


Adamillo:
What's the minimum requirement for Intel mac users?

squidbus:
> What's the minimum requirement for Intel mac users?

I haven't tested on Intel Macs since I don't have one, but it should probably work on at least 11 and up. Let me set a definite minimum version and make sure it compiles.

Adamillo:
I see! Do make sure that the requirement for macOS on M series doesn't change though

squidbus:
Minimum is now set to macOS 11, although macOS 15 is still required for ARM devices because of AVX support.

src/core/cpu_patches.cpp
```diff
@@ -0,0 +1,381 @@
...
+               regs.size());
+
+    std::call_once(register_save_init_flag, &InitializeRegisterSaveSlots);
+
+    u32 index = 0;
+    for (const auto& reg : regs) {
+        const auto offset = reinterpret_cast<void*>(register_save_slots[index++] * sizeof(void*));
+
+        c.putSeg(Xbyak::util::gs);
+        c.mov(Xbyak::util::qword[offset], reg.cvt64());
```
raphaelthegreat: Why not push the registers to the stack instead? The GS segment could be used as guest TLS on unix platforms, avoiding the need for patches. I have done this in my linux fixes PR

src/core/cpu_patches.cpp
```diff
@@ -0,0 +1,381 @@
...
+            return Xbyak::Reg32e(index, static_cast<int>(bits));
+        }
+    }
+    UNREACHABLE_MSG("Out of scratch registers!");
+}
+
+static bool GenerateTlsAccess(const ZydisDecodedOperand* operands, Xbyak::CodeGenerator& c) {
+    const auto& dst_op = operands[0];
+    const auto& src_op = operands[1];
+    if (src_op.type != ZYDIS_OPERAND_TYPE_MEMORY || src_op.mem.segment != ZYDIS_REGISTER_FS ||
```
raphaelthegreat: Will this also account for 64-bit TLS access pattern

src/core/module.cpp
```diff
@@ -84,8 +84,9 @@ void Module::LoadModuleToMemory(u32& max_tls_index) {
 
     // Map module segments (and possible TLS trampolines)
     void** out_addr = reinterpret_cast<void**>(&base_virtual_addr);
-    memory->MapMemory(out_addr, LoadAddress, aligned_base_size + TrampolineSize,
-                      MemoryProt::CpuReadWrite, MemoryMapFlags::Fixed, VMAType::Code, name, true);
+    memory->MapMemory(out_addr, memory->VirtualOffset() + LoadAddress,
```
raphaelthegreat: Why is this needed? If the virtual base is allocated too high it will break a lot of games

src/video_core/renderer_vulkan/liverpool_to_vk.cpp
```diff
@@ -281,6 +281,63 @@ vk::BorderColor BorderColor(AmdGpu::BorderColor color) {
     }
 }
 
+const std::vector<vk::Format>& GetAllFormats() {
+    static const std::vector formats{
```
raphaelthegreat:
```suggestion
    static constexpr std::array formats{
```
Make this a constexpr array and return a std::span<const vk::Format> from this function

src/core/address_space.cpp
```diff
@@ -245,21 +245,36 @@ struct AddressSpace::Impl {
     Impl() {
         // Allocate virtual address placeholder for our address space.
         void* hint_address = reinterpret_cast<void*>(SYSTEM_MANAGED_MIN);
+#ifdef __APPLE__
+        constexpr int virtual_flags = MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE;
```
raphaelthegreat: If fixed mapping fails might be worth doing something similar to what windows does. Depends on how much of the area macos process already has reserved though. Also consider disabling ASLR to make this deterministic

raphaelthegreat:
I'm unsure about adding a macos CI so early, nobody in the team has a Macbook, so the question of maintenance is important. But nice work, lots of good changes here

src/core/cpu_patches.cpp
squidbus: Originally this pushed registers to the stack, but you run into problems when guest code writes below the stack pointer without decrementing. I could try to guess a safe distance to move the stack down to avoid clobbering that data, but the safest way I could think of to do this was to just set up thread local storage to save registers to.

src/core/cpu_patches.cpp
squidbus: It accounts for `mov (64-bit register), fs:[0]` with any size for the zero, 32 or 64 bit. We could support non-zero offset as well if needed by removing some of the checks here, but I kept it in line with the current support. Especially since the more specific we get here, the better it guards against false-read instructions from data in between code.

src/core/module.cpp
squidbus: I haven't been able to find any way to reserve the range that is currently asked for; the best I've been able to get is starting at `0x7000000000`. The reason being that dynamic libraries and Rosetta runtime components may be allocated in the range first.

There's a workaround that wine uses to reserve low address space by creating a `zerofill` section in the executable covering the desired range, but the size of the virtual memory space is way too large, and creating a `zerofill` section that large causes it to crash on load.

squidbus:
> I'm unsure about adding a macos CI so early, nobody in the team has a Macbook, so the question of maintenance is important. But nice work, lots of good changes here

I figured since you usually do releases separate from CI builds it would probably be fine, but I can disable uploads and just do verification CI if you want, or no CI if really preferred.

raphaelthegreat:
Hm alright we can keep this in for now and see later how it goes

src/core/address_space.cpp
squidbus: I'll see if I can get a lower address to work.

src/core/module.cpp
squidbus: Is there a known limit on how high the virtual base can be before games start to break? Or specific games that can be checked against.

raphaelthegreat: PS4 memory system has system managed area, system reserved and user area. There are games that expect mappings to be at most 40bits in size, so higher addresses are not allowed. For system managed the emulator currently reserves from the lowest portion of the area, however we can get away with starting from 0xfc0000000 instead as red-prig discovered in kernel that is the search start.
So I would say you can try the zerofill trick from 0xfc0000000 to end of system managed (there are constants for these in address_space.h). Then let Rosetta take whatever it wants from system reserved area (reserved area is only used for code modules atm and tessellation mapping, and games can't map there. Hopefully it doesn't reach user area) and reserve user area. Games can pass fixed addresses in user area and map them freely so we need to make sure it's reserved properly.

src/core/cpu_patches.cpp
squidbus: Also FYI the fs/gs relationship is reversed on macOS, similar to Windows. This is just accessing some pre-allocated TLS slots to store data. The equivalent of what you're doing on Linux in your other PR would be storing the guest TLS in FS (which I suppose makes things even simpler, assuming macOS lets you actually modify it).

raphaelthegreat: Would be good to check (on windows I believe red-prig found that FS is getting reset by the OS on context switches, making it impractical to use)

squidbus: Unfortunately can't find a way to write the FS base at all on macOS, so it'll probably have to retain the patch to read from a host TLS slot. I'll adapt the patch for other Unix-like OSes for your changes when they're merged in, though.

raphaelthegreat: Sure thing

src/core/module.cpp
squidbus: There seems to be a region starting somewhere near 0x100000000 and ending at 0x7000000000 that just refuses to be mapped, even if looking at the virtual memory map of the process it appears to be unused. I've tried tricks like reserving the area for stack or other memory in the executable, and while that works for regions higher up in memory, if it overlaps with that unmappable region my system actually crashes completely. It turns out that this is the issue I was having earlier, not that the zerofill was too big.

So I don't think it's actually possible to get the user area mapping any lower than 0x7000000000

src/core/cpu_patches.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    return true;
+}
+
+#endif // __APPLE__
+
+using InstructionGenerator = bool (*)(const ZydisDecodedOperand*, Xbyak::CodeGenerator&);
+static const std::unordered_map<ZydisMnemonic, InstructionGenerator> InstructionGenerators = {
+    {ZYDIS_MNEMONIC_MOV, GenerateTlsAccess},
+#ifdef __APPLE__
```
georgemoralis, line 337: i wonder if that can be more generic , for example using xbyak cpu feature detection , so it can work on other platforms that doesn't support BMI1 . Also same stuff should be done for SSE4a which is amd only and it isn't supported in rosseta 2 too (i assume)

squidbus: It's possible to use feature detection but I don't think it would be very useful. It's a unique situation for Rosetta 2 that it supports newer extensions like AVX but not all of BMI1; you probably wouldn't find such a situation anywhere else. And supporting patching out AVX and others as well would add a lot of complexity.

For SSE4a I'm not sure on the details, but I'm mainly bringing it up to parity with the other platforms here. If that's an issue running the emulator on e.g. Windows with an Intel CPU, could be something to explore down the line.

georgemoralis: SSE4a is amd exclusive (only 4 instructions) . That is an issue for all platforms that has intel cpu , and since it is so rare i am sure also rosseta doesn't have it

raphaelthegreat: With this new patch framework I feel we can add any instructions we see relatively easily yeah

src/core/module.cpp
raphaelthegreat: User area usage is less common as system managed so as long as that is reserved properly it should be a good first step compatibility wise

abouvier:
Could you add `find_package(date 3.0.1 CONFIG)` in the root `CMakeLists.txt`?

src/core/module.cpp
squidbus: Is it okay if the system managed and user space areas are discontiguous? I think I could maybe get the system area reserved going up to its end, but the user area would have to go all the way over at 0x7000000000 as mentioned.

raphaelthegreat: Should be okay as games don't map to system reserved which is in-between those. Just add an assertion if a game requests to map directly in the unmapped portion of user area. Should be quite rare but can happen

squidbus:
Changes:
* Added missing `find_package` for `date`.
* Rebased with the changes to TLS on Linux. The patching system now has support for in-place patches instead of trampolines, which is used for this case.
* Refactored the address space mapping code a bit to allow separate areas for each region, and used that on macOS to allocate system managed in the right location and the rest wherever we can get.

src/core/cpu_patches.cpp
```diff
@@ -0,0 +1,393 @@
...
+#include "cpu_patches.h"
+
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <pthread.h>
+#endif
+
+namespace Core {
+
```
raphaelthegreat, line 25: I feel doing ```using namespace Xbyak::util;``` which help reducing some noise

src/core/cpu_patches.cpp
```diff
@@ -0,0 +1,393 @@
...
+                if (patch_info.trampoline) {
+                    const auto trampoline_ptr = c.getCurr();
+
+                    patch_info.generator(operands, c);
+
+                    // Return to the following instruction at the end of the trampoline.
+                    c.jmp(code + instruction.length);
+
+                    // Replace instruction with near jump to the trampoline.
+                    patch_gen.jmp(trampoline_ptr, Xbyak::CodeGenerator::LabelType::T_NEAR);
```
raphaelthegreat, line 382: I don't see the prefix bytes to nop patch anymore. This will cause some openorbis demos to crash if you leave it out

src/core/address_space.cpp
```diff
@@ -52,18 +57,38 @@ struct AddressSpace::Impl {
...
+
+        system_managed_size = SystemManagedSize;
+        system_reserved_size = SystemReservedSize + ReductionOnFail;
+        user_size = UserSize;
         for (u32 i = 0; i < MaxReductions && !virtual_base; i++) {
-            virtual_size -= ReductionOnFail;
-            virtual_base = static_cast<u8*>(VirtualAlloc2(process, NULL, virtual_size,
-                                                          MEM_RESERVE | MEM_RESERVE_PLACEHOLDER,
-                                                          PAGE_NOACCESS, &param, 1));
+            system_reserved_size -= ReductionOnFail;
```
raphaelthegreat: Typically on windows the lower part of system_managed is reserved. And since this VirtualAlloc2 has address requirements which force it to the range we want, reducing the virtual size will most definitely cut system managed size and not reserved size. So you probably need to subtract this from system_managed_size

src/core/address_space.cpp
```diff
@@ -52,18 +57,38 @@ struct AddressSpace::Impl {
...
+            virtual_base = static_cast<u8*>(
+                VirtualAlloc2(process, NULL, system_managed_size + system_reserved_size + user_size,
+                              MEM_RESERVE | MEM_RESERVE_PLACEHOLDER, PAGE_NOACCESS, &param, 1));
         }
         ASSERT_MSG(virtual_base, "Unable to reserve virtual address space!");
 
+        system_managed_base = virtual_base;
+        system_reserved_base = virtual_base + system_managed_size;
+        user_base = system_reserved_base + system_reserved_size;
```
raphaelthegreat: Add an assert user_base is the expected for good measure

src/core/address_space.cpp
```diff
@@ -244,22 +274,72 @@ enum PosixPageProtection {
...
         // Allocate virtual address placeholder for our address space.
-        void* hint_address = reinterpret_cast<void*>(SYSTEM_MANAGED_MIN);
-        virtual_size = SystemSize + UserSize;
-        virtual_base = reinterpret_cast<u8*>(
-            mmap(hint_address, virtual_size, PROT_READ | PROT_WRITE,
+        system_managed_size = SystemManagedSize;
+        system_reserved_size = SystemReservedSize;
+        user_size = UserSize;
+#ifdef __APPLE__
+        system_managed_base = reinterpret_cast<u8*>(
```
raphaelthegreat, line 295: Make a constexpr variable with the common flags MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE to reduce visual noise

src/core/address_space.cpp
```diff
@@ -244,22 +274,72 @@ enum PosixPageProtection {
...
+        if (backing_fd < 0) {
+            LOG_CRITICAL(Kernel_Vmm, "shm_open failed: {}", strerror(errno));
+            throw std::bad_alloc{};
+        }
+        shm_unlink(shm_path.c_str());
+#else
         madvise(virtual_base, virtual_size, MADV_HUGEPAGE);
 
+        const VAddr start_addr = reinterpret_cast<VAddr>(virtual_base);
+        m_free_regions.insert({start_addr, start_addr + virtual_size});
```
raphaelthegreat: Isn't this the same as the insertion of system_managed_addr above?

src/core/cpu_patches.cpp
squidbus: The nop bytes are done further down, it's set up so that you either patch in a new instruction or a jump to a trampoline, then regardless of which happens it will fill the remainder with nops.

raphaelthegreat:
Out of curiosity what where the final mapping addresses you were able to achieve?

src/core/address_space.cpp
squidbus: Accidentally got left in, thanks

squidbus:
> Out of curiosity what where the final mapping addresses you were able to achieve?

```
[Kernel.Vmm] <Info> address_space.cpp:Impl:313: System managed virtual memory region: 0x400000 - 0x7ffffbfff
[Kernel.Vmm] <Info> address_space.cpp:Impl:316: System reserved virtual memory region: 0x7000000000 - 0x77ffffffff
[Kernel.Vmm] <Info> address_space.cpp:Impl:318: User virtual memory region: 0x7800000000 - 0xf7ffffffff
```

src/core/cpu_patches.cpp
raphaelthegreat: I see, I think zydis will handle it properly in this case as it fully parses the instructions

squidbus: Ah sorry I misread, yeah the prefix bytes should be handled by Zydis, and then any remaining data gets nop'd out.

raphaelthegreat:
Question, could you try doing the reduce trick windows does but on system reserved area? I.e how much can you map from end of system managed until you reach that "unmappable" region. Large portion of system reserved area is unused by the emulator so we need very little for it technically. Would make things more convenient to not allocate it that high

squidbus:
> Question, could you try doing the reduce trick windows does but on system reserved area? I.e how much can you map from end of system managed until you reach that "unmappable" region. Large portion of system reserved area is unused by the emulator so we need very little for it technically. Would make things more convenient to not allocate it that high

I'm still testing things out, but it seems to behaves very strangely if I got beyond the managed region at all. Mapping 0x4000 seems to go fine but is obviously way too small, then going higher than that it starts to either SIGBUS or hang before the mmap returns, up until I get to the upper range of sizes where it just fails to map.

EDIT: I think this is because the mappings for whatever is placed there are able to be overwritten, I'll see if I can tweak the reservations at all.

squidbus:
I'm able to reserve and map the first 0x700000000 of the reserved area, I assume that would be enough?

raphaelthegreat:
Yes looks enough

squidbus:
New map:
```
[Kernel.Vmm] <Info> address_space.cpp:Impl:318: System managed virtual memory region: 0x400000 - 0x7ffffbfff
[Kernel.Vmm] <Info> address_space.cpp:Impl:321: System reserved virtual memory region: 0x800000000 - 0xeffffffff
[Kernel.Vmm] <Info> address_space.cpp:Impl:323: User virtual memory region: 0x7000000000 - 0xefffffffff
```

raphaelthegreat:
Nice. Also you might want to adjust the tessellation ring buffer address for macos https://github.com/shadps4-emu/shadPS4/blob/0dda5eb6ef320e16d8e49fd1da5798d28ca6f819/src/core/libraries/gnmdriver/gnmdriver.cpp#L291C59-L291C68

georgemoralis:
can you rebase to latest master , we need to do some tests in windows and linux and we probably gonna merge after

squidbus:
> can you rebase to latest master , we need to do some tests in windows and linux and we probably gonna merge after

Sure, rebased.

georgemoralis:
window build gives this at startup :/

[Debug] <Critical> address_space.cpp:operator():82: Assertion Failed!
Unexpected user address space location: 0xfc1950000

squidbus:
> window build gives this at startup :/
> 
> [Debug] address_space.cpp:operator():82: Assertion Failed! Unexpected user address space location: 0xfc1950000

Should be fixed, and rebased again.

I've made the Windows code match the intent more, where the system managed region gets adjusted if needed and the other regions are guaranteed to be fixed in place.

georgemoralis:
sceKernelConvertUtcToLocaltime  throws a runtime exception in date::current_zone() if that works on mac maybe we should do it with an #ifdef

squidbus:
> sceKernelConvertUtcToLocaltime throws a runtime exception in date::current_zone() if that works on mac maybe we should do it with an #ifdef

Weird, since the standardized version is directly based on the library; guess the implementation must be different. I've updated it to only use the date library on macOS.

georgemoralis:
ok did some tests seems ok , can i have a small request? can you ifdef the keyboard keys as well and leave keypad for windows -linux users? we gonna add a configurator soon but i don't wanna end user start searching for keys again :D

georgemoralis:
also why not joining discord? it might be a lot easier than this :D

squidbus:
> ok did some tests seems ok , can i have a small request? can you ifdef the keyboard keys as well and leave keypad for windows -linux users? we gonna add a configurator soon but i don't wanna end user start searching for keys again :D

Sure, done.

> also why not joining discord? it might be a lot easier than this :D

Sure, I'm having some internet problems right now and pushing updates from a phone hotspot, but I can join within the next day or two.

VasylBaran:
Any updates? Looking forward to this =)

squidbus:
> Any updates? Looking forward to this =)

As far as I know it's good to go, just waiting on one of the maintainers to check it doesn't affect Linux.

raphaelthegreat:
Tested it on linux, works okay

raphaelthegreat (approved):

PR #299, opened 2024-07-16
Return EBUSY from sceKernelPollEventFlag instead of ETIMEDOUT.

squidbus:
sceKernelPollEventFlag should return EBUSY instead of ETIMEDOUT when the event flag is not set.

Progresses CUSA16429 further until shader compilation issues. Without this you can see it logging out:
```
assertion failed. file D:\dev\arukas\work\dev\arukas\Library\needle\build\Rsdx\source\Core\Algo\hhRsdxAlgoMutex.cpp line(763) function(RsdxWaitEvent) condition(res == SCE_OK || res == SCE_KERNEL_ERROR_EBUSY)
```
because the return code is wrong.

May help with CUSA24365 as well since it's from the same developer and has the same issue, but I have not tested it.


squidbus:
Realized this should probably go in `EventFlagInternal::Poll` instead, so moved it there.

georgemoralis:
ok this appears to be valid for the existing design  (eventflags need a rewrite anyway)

PR #303, opened 2024-07-18
Move presentation to separate thread/improve sync

raphaelthegreat:
* Moves presentation to dedicated thread instead of sharing the same thread as input polling. Gives us better control over presentation rate and reduces input latency. Swapchain could block for up to 16ms before preventing inputs during that time.
* More proper vblank event firing. Before vblank was triggered on every present request which could be a lot more often than 16.6ms. Now we keep track of time when it should actually get triggered.
* Proper emulation of flip rate limiting. Games can set the number of vblank intervals that need to occur before a flip request can be processed. This means that games like Sonic Mania run at normal speed now.
* Fix sync problems with cpu side flips. Phyre engine games like Flow and Journey perform flips by calling sceVideooutSubmitFlip directly instead of using gpu flips. The flip also comes from a different thread than main, so using the drawing scheduler that could be in use by the gpu thread is a bad idea. Assign different command buffer scheduler in this case to resolve problem.
* Small overhaul to scheduler sync interface. It now receives a struct where all sync info like semaphores and fences is stored which gives us more flexibility

Todo:
* Make presentation rate configurable as from now on FPS will be locked to 60
* Make use of more advanced extensions like VK_GOOGLE_display_timing and VK_KHR_present_wait for better latency control and to query the refresh rate of the monitor accurately.
* Test this as it might break things


abouvier:
You forgot to put a `FindRenderDoc.cmake` file in the `cmake` folder to make `find_package(RenderDoc MODULE)` meaningful.

PR #309, opened 2024-07-20
Update building-linux.md

zqpvr:
add git to arch and fedora

### DIFF
--- a/documents/building-linux.md
+++ b/documents/building-linux.md
@@ -14,12 +14,12 @@ sudo apt-get install build-essential libasound2-dev libpulse-dev libopenal-dev z
 
 #### Fedora
 ```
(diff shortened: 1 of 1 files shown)


ryzendew:
Both arch and fedora come with git by default even on vanilla installs they changed that long ago.

abouvier:
> Both arch and fedora come with git by default even on vanilla installs they changed that long ago.

None of Arch base packages depends on `git`.

ryzendew:
> > Both arch and fedora come with git by default even on vanilla installs they changed that long ago.
> 
> None of Arch base packages depends on `git`.

I just did both a by hand install and archinstall and git was installed with both … I don’t say words for nothing i test before speaking like people should

abouvier:
> > > Both arch and fedora come with git by default even on vanilla installs they changed that long ago.
> > 
> > 
> > None of Arch base packages depends on `git`.
> 
> I just did both a by hand install and archinstall and git was installed with both … I don’t say words for nothing i test before speaking like people should

Can you do `pacman -Qi git` and paste the line starting with `Required By`?

PR #310, opened 2024-07-21
64 bits OP, impl V_ADDC_U32 & V_MAD_U64_U32

viniciuslrangel:
~Still need to implement SSA rewrite auto cast between 64 and 32 bits~

Fixed op src being wrongly forced to float


viniciuslrangel:
@raphaelthegreat Can you review?

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -92,6 +92,26 @@ void Translator::V_ADD_I32(const GcnInst& inst) {
     // TODO: Carry
 }
 
+void Translator::V_ADDC_U32(const GcnInst& inst) {
+
+    const auto src0 = GetSrc<IR::U32>(inst.src[0]);
+    const auto src1 = GetSrc<IR::U32>(inst.src[1]);
+
+    IR::U1 scarry;
```
raphaelthegreat: Is this really a U1? It makes more sense for the shader to use lower part of VCC mask as U32 carry

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -294,6 +314,20 @@ void Translator::V_SUBREV_I32(const GcnInst& inst) {
...
+    const auto src0 = GetSrc<IR::U32>(inst.src[0]);
+    const auto src1 = GetSrc<IR::U32>(inst.src[1]);
+    const auto src2 = GetSrc64<IR::U64>(inst.src[2]);
+
+    IR::U64 result;
+    result = ir.IMul(src0, src1);
+    result = ir.IAdd(ir.UConvert(64, result), src2);
+
+    const IR::VectorReg dst_reg{inst.dst[0].code};
+    ir.SetVectorReg64(dst_reg, result);
```
raphaelthegreat: Docs say this also sets VCC but it is unclear to me if it is full 64bit register or only lower 32bits

src/shader_recompiler/ir/ir_emitter.cpp
```diff
@@ -145,34 +145,64 @@ void IREmitter::SetThreadBitScalarReg(IR::ScalarReg reg, const U1& value) {
...
 }
 
 template <>
 F32 IREmitter::GetScalarReg(IR::ScalarReg reg) {
     return BitCast<F32>(GetScalarReg<U32>(reg));
 }
 
+template <>
+U64 IREmitter::GetScalarReg(IR::ScalarReg reg) {
+    return Inst<U64>(Opcode::GetScalarRegister, reg, Imm32(64));
```
raphaelthegreat: This is more complicated than necessary I think. You can avoid modifying the IR instruction by loading 2 scalar/vector registers and combining them with PackUint2x32. So this instruction is kept clean

raphaelthegreat: Also IREmitter should not handle this at all I think. Have Translator do this in 64-bit version of GetSrc

raphaelthegreat: This also depends on how guest shader uses 64-bit arithmetic. For single instances in some instructions I feel this will be a simpler way to achieve the result in the current system. Otherwise if a larger section of the shader performs U64 arithmetic it might be worth investigating other solutions. Probably a constant propagation pass that detects this Pack -> Unpack -> Pack patterns and preserves the original U64 value

src/shader_recompiler/frontend/translate/vector_alu.cpp
red-prig: > Is this really a U1? It makes more sense for the shader to use lower part of VCC mask as U32 carry

Setting in VCC, each bit corresponds to the thread number in wavefront [0..63] that is, if you do not yet emulate all wavefront functionality, you can, for example, set the first bit

raphaelthegreat: Yeah a lot of instructions use it for per thread bit storage like compares. Some instructions can also use the low 32-bits as temporary storage though

src/shader_recompiler/ir/passes/ssa_rewrite_pass.cpp
```diff
@@ -310,7 +310,8 @@ class Pass {
     DefTable current_def;
 };
 
-void VisitInst(Pass& pass, IR::Block* block, IR::Inst& inst) {
+void VisitInst(Pass& pass, IR::Block* block, const IR::Block::iterator& iter) {
```
raphaelthegreat: These refactors are not needed now it seems

src/shader_recompiler/ir/opcodes.inc
```diff
@@ -289,6 +293,9 @@ OPCODE(ConvertF64S32,                                       F64,            U32,
 OPCODE(ConvertF64U32,                                       F64,            U32,                                                                            )
 OPCODE(ConvertF32U16,                                       F32,            U16,                                                                            )
 OPCODE(ConvertU16U32,                                       U16,            U32,                                                                            )
+OPCODE(ConvertU64U32,                                       U64,            U32,                                                                            )
```
raphaelthegreat: Are these opcodes needed anymore?

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -92,6 +93,30 @@ void Translator::V_ADD_I32(const GcnInst& inst) {
...
+    } else { // VOP2
+        scarry = ir.GetVccLo();
+    }
+
+    IR::U32 result = ir.IAdd(ir.IAdd(src0, src1), scarry);
+
+    const IR::VectorReg dst_reg{inst.dst[0].code};
+    ir.SetVectorReg(dst_reg, result);
+
+    IR::U1 less_src0 = ir.ILessThan(result, src0, false);
```
raphaelthegreat:
```suggestion
    const IR::U1 less_src0 = ir.ILessThan(result, src0, false);
```
ditto for other small cases of const correctness

raphaelthegreat (commented):
Nice work only small nits left

raphaelthegreat (approved):

PR #314, opened 2024-07-21
Replace remaining uses of `QDir::currentPath() / "user"` with `GetUserPath(UserDir)`.

squidbus:
Instead of relying on `QDir::currentPath() + "/user"` to be correct, use `GetUserPath(UserDir)` as the single source of truth on the user directory like the rest of the codebase.

I think I handled the cases where `wstring` is needed for Windows correctly, let me know if there's anything wrong there.


qurious-pixel:
Works with AppImages.

PR #317, opened 2024-07-23
Fix macOS builds

VasylBaran:
Install molten-vk using brew (as oppose to fetching it directly with url) to fix macOS build

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -30,10 +30,7 @@ jobs:
 
     - name: Install MoltenVK
       run: |
-        curl -LO https://github.com/KhronosGroup/MoltenVK/releases/download/v1.2.10/MoltenVK-macos.tar
-        tar xf MoltenVK-macos.tar
-        sudo mkdir -p /usr/local/lib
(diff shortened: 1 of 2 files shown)


.github/workflows/macos.yml
```diff
@@ -45,7 +42,7 @@ jobs:
       run: |
         mkdir upload
         mv ${{github.workspace}}/build/shadps4 upload
-        cp /usr/local/lib/libMoltenVK.dylib upload
+        cp /opt/homebrew/opt/molten-vk/lib/libMoltenVK.dylib upload
```
shinra-electric, line 45: This shouldn't be changed to `opt/homebrew`. 
x86 binaries are kept in `usr/local` and arm binaries are kept in `opt/homebrew`

VasylBaran: molten-vk didn't exist in usr/local

squidbus:
This fix is not correct, you are installing ARM MoltenVK for an x86_64 build. It seems that the issue is just that 1.2.10 was revoked, so the simpler fix would just be to change the release in the URL to v1.2.9

PR #318, opened 2024-07-23
Install x86_64 MoltenVK from Homebrew.

squidbus:
The macOS runners in GitHub Actions are ARM based now, we need to explicitly use x86_64 Homebrew to install x86_64 libraries as they do not ship universal binaries.


raphaelthegreat (approved):
Thanks

PR #322, opened 2024-07-24
Add partial unmap support

Borchev:
Adds support for unmapping subsections of previously mapped sections. P.T appears to require this functionality and will crash early on in the boot process without it.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -54,7 +54,7 @@ PAddr MemoryManager::Allocate(PAddr search_start, PAddr search_end, size_t size,
     free_addr = alignment > 0 ? Common::AlignUp(free_addr, alignment) : free_addr;
 
     // Add the allocated region to the list and commit its pages.
-    auto& area = CarveDmemArea(free_addr, size);
+    auto& area = CarveDmemArea(free_addr, size)->second;
     area.memory_type = memory_type;
     area.is_free = false;
     return free_addr;
@@ -63,9 +63,8 @@ PAddr MemoryManager::Allocate(PAddr search_start, PAddr search_end, size_t size,
 void MemoryManager::Free(PAddr phys_addr, size_t size) {
     std::scoped_lock lk{mutex};
 
-    const auto dmem_area = FindDmemArea(phys_addr);
-    ASSERT(dmem_area != dmem_map.end() && dmem_area->second.base == phys_addr &&
-           dmem_area->second.size == size);
+    auto dmem_area = CarveDmemArea(phys_addr, size);
+    ASSERT(dmem_area != dmem_map.end() && dmem_area->second.size >= size);
 
     // Release any dmem mappings that reference this physical block.
     std::vector<std::pair<VAddr, u64>> remove_list;
@@ -169,6 +168,7 @@ int MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, size_t size, M
     new_vma.prot = prot;
     new_vma.name = name;
     new_vma.type = type;
+    new_vma.is_exec = is_exec;
 
     if (type == VMAType::Direct) {
         new_vma.phys_base = phys_addr;
@@ -216,10 +216,16 @@ void MemoryManager::UnmapMemory(VAddr virtual_addr, size_t size) {
     std::scoped_lock lk{mutex};
 
     const auto it = FindVMA(virtual_addr);
-    ASSERT_MSG(it->second.Contains(virtual_addr, size),
+    const auto& vma_base = it->second;
+    ASSERT_MSG(vma_base.Contains(virtual_addr, size),
                "Existing mapping does not contain requested unmap range");
 
-    const auto type = it->second.type;
+    const auto vma_base_addr = vma_base.base;
+    const auto vma_base_size = vma_base.size;
+    const auto phys_base = vma_base.phys_base;
+    const bool is_exec = vma_base.is_exec;
+    const auto start_in_vma = virtual_addr - vma_base_addr;
+    const auto type = vma_base.type;
     const bool has_backing = type == VMAType::Direct || type == VMAType::File;
     if (type == VMAType::Direct) {
         UnmapVulkanMemory(virtual_addr, size);
@@ -239,7 +245,8 @@ void MemoryManager::UnmapMemory(VAddr virtual_addr, size_t size) {
     MergeAdjacent(vma_map, new_it);
 
     // Unmap the memory region.
-    impl.Unmap(virtual_addr, size, has_backing);
+    impl.Unmap(vma_base_addr, vma_base_size, start_in_vma, start_in_vma + size, phys_base, is_exec,
+               has_backing);
     TRACK_FREE(virtual_addr, "VMEM");
 }
 
@@ -404,7 +411,7 @@ MemoryManager::VMAHandle MemoryManager::CarveVMA(VAddr virtual_addr, size_t size
     return vma_handle;
 }
 
-DirectMemoryArea& MemoryManager::CarveDmemArea(PAddr addr, size_t size) {
+MemoryManager::DMemHandle MemoryManager::CarveDmemArea(PAddr addr, size_t size) {
     auto dmem_handle = FindDmemArea(addr);
     ASSERT_MSG(dmem_handle != dmem_map.end(), "Physical address not in dmem_map");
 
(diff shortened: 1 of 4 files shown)


src/core/memory.cpp
```diff
@@ -238,13 +251,31 @@ int MemoryManager::MapFile(void** out_addr, VAddr virtual_addr, size_t size, Mem
...
-
-    const auto type = it->second.type;
+    auto vma_handle = FindVMA(virtual_addr);
+    auto& entire_vma = vma_handle->second;
+    const VAddr start_in_vma = virtual_addr - entire_vma.base;
+    const VAddr end_in_vma = start_in_vma + size;
+    const auto type = entire_vma.type;
     const bool has_backing = type == VMAType::Direct || type == VMAType::File;
+
+    // There does not appear to be comparable support for partial unmapping on Windows.
```
raphaelthegreat: The memory system should not handle OS specific operations. Move this logic inside address space Unmap()

raphaelthegreat: You can probably use the placeholders array to check which region is currently mapped

Borchev: Alright that makes sense. I'll move the logic into Unmap().

raphaelthegreat:
Also note the current approach I believe will not work with flexible mappings (i.e private mappings). For full partial unmap support on windows you might need to move all mappings to use the backing memory file. This is fine as its pretty explicit that 448MB maximum of that file is reserved for flexible memory

raphaelthegreat:
I think it might be fine in the end to implement this in memory as pushing this logic exclusively in address space will require additional tracking structures. However it will be good to have it return a bool flag if this workaround is needed so unix based systems can simply call unmap once

Borchev:
I initially looked into moving unmap support into the AddressSpace without changing the signature of the Unmap and it ended up getting messy. As you already mentioned, doing it that way required adding an additional mapping structure. Instead, I opted to change the Unmap signature to pass the info needed for the Windows workaround. Let me know if this is an acceptable compromise. 

I can look into moving the flexible memory allocations to use the backing mapped file but I think that should be in a separate commit. I'm not completely clear on why that its necessary to do this so I probably need to think it over more.

raphaelthegreat:
I have merged my memory cleanup PR which implements the partial unmap on linux (also uses CarveVMA instead of manually splitting like you did here) so feel free to rebase on it. For flexible memory unmaps the way you implemented partial unmaps is by unmapping the entire vma first and then remapping the parts that are necessary. But if the mapping was done with VirtualAlloc2() the data those mappings might hold can be lost.

src/core/memory.cpp
```diff
@@ -64,9 +64,21 @@ PAddr MemoryManager::Allocate(PAddr search_start, PAddr search_end, size_t size,
 void MemoryManager::Free(PAddr phys_addr, size_t size) {
     std::scoped_lock lk{mutex};
 
-    const auto dmem_area = FindDmemArea(phys_addr);
-    ASSERT(dmem_area != dmem_map.end() && dmem_area->second.base == phys_addr &&
-           dmem_area->second.size == size);
+    auto dmem_area = FindDmemArea(phys_addr);
+    ASSERT(dmem_area != dmem_map.end() && dmem_area->second.size >= size);
```
raphaelthegreat, line 67: Does P.T do partial direct memory area free as well (I tested it on linux and it didn't hit this assert)? Unless it does that better to leave this unchanged

Borchev: Yes, the version I tested allocates 0-7f0000, then frees 0-300000, followed by 700000-7f0000.

raphaelthegreat: In that case I would recommend doing similar thing to VMA logic. Change the assertion to a Contains() check and use CarveDmemArea at the bottom of the function

Borchev:
> I have merged my memory cleanup PR which implements the partial unmap on linux (also uses CarveVMA instead of manually splitting like you did here) so feel free to rebase on it. For flexible memory unmaps the way you implemented partial unmaps is by unmapping the entire vma first and then remapping the parts that are necessary. But if the mapping was done with VirtualAlloc2() the data those mappings might hold can be lost.

Thanks for the explanation. I forgot that the pages would be decommitted instantly with virtualfree so I'll need to address that like you said. I'll try to get it rebased soon.

Emulator-Team-2:
I have found some games that crashes right after launch due to this error.

```
[Debug] <Critical> address_space.cpp:operator():467: Assertion Failed!
Partial unmapping of flexible allocations is not supported
```

- Gintama Rumble
- Koihime Enbu Ryo Rai Rai
- Phantom Breaker: Omnia
- UNDER NIGHT IN-BIRTH Exe:Late[st]
- DENGEKI BUNKO FIGHTING CLIMAX IGNITION

##

```
[Debug] <Critical> memory.cpp:operator():403: Assertion Failed!
Adding an allocation to already allocated region
```
- THE BIT.TRIP

##

Let me know if you need logs from these games and I'll add them here.

raphaelthegreat:
Please post logs with all of these games

Emulator-Team-2:
@raphaelthegreat 
I added the logs in my previous post.

Borchev:
Thanks for testing this and providing logs. I messed up an assertion clause and forgot to update another. This latest change should fix those.

Emulator-Team-2:
The latest changes fixes this `Assertion Failed!` error.

Borchev:
Is anything else needed here? I have a follow-up PR to add support for partial unmapping of flexible allocations on Windows that relies on this one (Although, I'd rather not check it in if no titles actually do this as it complicates the logic in AddressSpace a fair bit).

src/core/memory.cpp
```diff
@@ -63,9 +63,8 @@ PAddr MemoryManager::Allocate(PAddr search_start, PAddr search_end, size_t size,
...
     std::scoped_lock lk{mutex};
 
-    const auto dmem_area = FindDmemArea(phys_addr);
-    ASSERT(dmem_area != dmem_map.end() && dmem_area->second.base == phys_addr &&
-           dmem_area->second.size == size);
+    auto dmem_area = CarveDmemArea(phys_addr, size);
+    ASSERT(dmem_area != dmem_map.end() && dmem_area->second.size >= size);
 
     // Release any dmem mappings that reference this physical block.
     std::vector<std::pair<VAddr, u64>> remove_list;
```
raphaelthegreat, line 70: This logic needs to be adjusted for partial unmaps as currently it will unmap the entire VMA

ryzendew:
could you rebase this PR so i can test it please i found a game that mentions unmap range

Borchev:
Done. Good luck.

ryzendew:
[shad_log.txt](https://github.com/user-attachments/files/16445440/shad_log.txt)
sad bear it didn't help but maybe this log could

Borchev:
Thanks for the logs. Ill try to take a look soon.

Borchev:
Changed the approach slightly. Added a backed-allocation map to AddressSpace to keep track of things instead of passing extra parameters just for the Windows implementation.

Emulator-Team-2:
With the latest changes Gintama Rumble crashes with this error.

```
[Kernel.Vmm] <Info> memory.cpp:Free:78: Unmaping direct mapping 0x29f070000 with size 0x4000
[Kernel.Vmm] <Info> memory.cpp:Free:78: Unmaping direct mapping 0x29f070000 with size 0x4000
[Debug] <Critical> address_space.cpp:operator():187: Assertion Failed!
(FormatMessageA failed to format error)
```

Log:
[shad_log.zip](https://github.com/user-attachments/files/16451288/shad_log.zip)

Borchev:
Thanks for the logs. Unfortunately, I don't have a copy of the decrypted system libraries so I can't replicate the issue. I tested against exactly the same version of that game and didn't hit that assertion. However, I did find an issue with the way I was doing address lookups in the backed-allocation map and just pushed a fix. Hopefully, that also addresses the assertion error with Gintama Rumble.

Emulator-Team-2:
The latest changes fixes assertion error in Gintama Rumble.

raphaelthegreat:
Could we go back to providing the arguments to Unmap? I feel it was much simpler approach and avoids hidden bugs with having to maintain additional tracking structure >:

Borchev:
Sure, I'll revert it to the original implementation shortly.

Borchev:
This PR has been open for three weeks now. Please let me know if there is anything else needed here. Otherwise, I'll close it.

raphaelthegreat:
Apologies for delay lgtm

PR #323, opened 2024-07-24
address_space: Fix windows placeholder mapping

raphaelthegreat:
Placeholder interval set needs to be contiguous, otherwise windows complains

### DIFF
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -59,10 +59,10 @@ struct AddressSpace::Impl {
         static constexpr size_t MaxReductions = 10;
 
         size_t reduction = 0;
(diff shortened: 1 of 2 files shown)


red-prig:
Windows is quite strict about splitting and merging intervals and will give an error if you try to split an interval that does not need splitting and join an interval that does not need to be merged, personally I do strict checks and use three intervals:
```
0x00010000000..0x00070000000  PROC_AREA  //(original:0x400000-0x80000000)
0x00080000000..0x00100000000  DL_AREA    //(original:0x80000000-0x200000000)
0x00200000000..0x10000000000  USR_HEAP   //(original:0x00200000000-0x800000000000)
```
In the case of a patched individual process, I free up space starting at 0x400000 by moving the stack of the running process to another location.

PR #324, opened 2024-07-24
memory: Cleanups and refactors

raphaelthegreat:
Allows Rayman Legends to go in game. Shadow of Collosus remake and P.T also progress to GPU crashes on Linux (on Windows they need partial unmap support)


raphaelthegreat:
Latest commit fixes Dragon Ball Fighterz memory problems (i.e when free VMA exists that can fit the mapping but is not aligned to provided value)

PR #330, opened 2024-07-28
log improvement ThrowInvalidType

DanielSvoboda:
Helps with debugging, makes it easier to locate which function generated the error in this code

BEFORE:
```
[Debug] <Critical> ir_emitter.cpp:ThrowInvalidType:12: Unreachable code!
Invalid type 256
```

AFTER:
image

### DIFF
--- a/src/shader_recompiler/ir/ir_emitter.cpp
+++ b/src/shader_recompiler/ir/ir_emitter.cpp
@@ -8,8 +8,8 @@
 
 namespace Shader::IR {
 namespace {
-[[noreturn]] void ThrowInvalidType(Type type) {
-    UNREACHABLE_MSG("Invalid type {}", u32(type));
+[[noreturn]] void ThrowInvalidType(Type type, const std::string& functionName) {
+    UNREACHABLE_MSG("Invalid type= {}, functionName= {}", u32(type), functionName);
 }
 
 Value MakeLodClampPair(IREmitter& ir, const F32& bias_lod, const F32& lod_clamp) {
@@ -382,7 +382,7 @@ F32F64 IREmitter::FPAdd(const F32F64& a, const F32F64& b) {
     case Type::F64:
         return Inst<F64>(Opcode::FPAdd64, a, b);
     default:
-        ThrowInvalidType(a.Type());
+        ThrowInvalidType(a.Type(), "FPAdd");
     }
 }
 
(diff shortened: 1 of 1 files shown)


DanielSvoboda:
Taking advantage of the topic, would the missing type be the F8 of this function? (Invalid type 256)

```cpp
F32F64 IREmitter::FPSaturate(const F32F64& value) {
    switch (value.Type()) {
    case Type::F32:
        return Inst<F32>(Opcode::FPSaturate32, value);
    case Type::F64:
        return Inst<F64>(Opcode::FPSaturate64, value);
    default:
        ThrowInvalidType(value.Type(), "FPSaturate");
    }
}
```

raphaelthegreat:
> Taking advantage of the topic, would the missing type be the F8 of this function? (Invalid type 256)

There is no 8bit floating point type so this function is not missing anything I think

src/shader_recompiler/ir/ir_emitter.cpp
```diff
@@ -8,8 +8,8 @@
 
 namespace Shader::IR {
 namespace {
-[[noreturn]] void ThrowInvalidType(Type type) {
-    UNREACHABLE_MSG("Invalid type {}", u32(type));
+[[noreturn]] void ThrowInvalidType(Type type, const std::string& functionName) {
```
raphaelthegreat: Try using std::source_location here https://en.cppreference.com/w/cpp/utility/source_location

DanielSvoboda:
that made it better, thank you
image

DanielSvoboda:
Regarding number 256, I did a test with this:
```cpp 
const std::string typeName = Shader::IR::NameOf(type);
```
And it returned U32, I know it doesn't make sense, but it includes U32 in FPSaturate, and this error no longer appears.
```cpp
F32F64 IREmitter::FPSaturate(const F32F64& value) {
    switch (value.Type()) {
    case Type::F32:
        return Inst<F32>(Opcode::FPSaturate32, value);
    case Type::F64:
        return Inst<F64>(Opcode::FPSaturate64, value);
    case Type::U32:
        return Inst<F32>(Opcode::FPSaturate32, value);
    default:
        ThrowInvalidType(value.Type());
    }
}
```

raphaelthegreat:
Yeah this not a correct solution. Saturation is a floating point operation which clamps between 0.0 and 1.0 so doing it for integers will produce wrong results. You need to find which GCN instruction emitted this (set a breakpoint in the place the error is log and determine why its not float)

DanielSvoboda:
Yes, I'll look at this issue of 256 later with more time. This pull in question is ok, I added the name of the function and the line.

raphaelthegreat (approved):
Thanks

PR #333, opened 2024-07-28
Translate/Emit/Convert

DanielSvoboda:
translate.cpp
- [X] DS_WRITE_B64
- [X] DS_READ_B64
- [X] DS_READ2_B64
- [X] S_MIN_U32
- [X] S_MAX_U32
- [X] V_CMP_U_F32
- [X] S_CMP_GT_U32
- [X] V_MIN3_I32
- [X] V_MBCNT_LO_U32_B32
- [X] V_MBCNT_HI_U32_B32
- [ ] V_MAD_U32
- [ ] V_MBCNT_U32_B32 -check
------------------------------------
UConvert
- [X] ConvertU32U64
- [X] ConvertU64U32
- [ ] ConvertU32x2U32
- [ ] ConvertU32x2U64
------------------------------------
- [ ] EmitLoadBufferFormatF64xN
- [ ] EmitLoadBufferFormatF64
- [ ] EmitLoadBufferFormatF64x2
- [ ] EmitLoadBufferFormatF64x3
- [ ] EmitLoadBufferFormatF64x4
68 to 68 - these are really necessary? 
some games call it 70,73,78,81... what's that?
------------------------------------
- [X] GetGpuClock64 - it was not implemented, but it only has a hack with break

and some more corrections...


src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -65,4 +65,20 @@ void Translator::V_READFIRSTLANE_B32(const GcnInst& inst) {
     UNREACHABLE();
 }
 
+void Translator::DS_MAX(int bit_size, const GcnInst& inst) {
+    const IR::U32 addr{ir.GetVectorReg(IR::VectorReg(inst.src[0].code))};
+    const IR::U32 data{ir.GetVectorReg(IR::VectorReg(inst.src[1].code))};
+    const IR::U32 current_max{ir.GetVectorReg(IR::VectorReg(inst.dst[0].code))};
+    const IR::U32 max_value{ir.UMax(current_max, data)};
```
raphaelthegreat: Implementing shared atomics with normal UMin is not the same and will break things

src/shader_recompiler/frontend/translate/vector_memory.cpp
```diff
@@ -381,4 +381,63 @@ void Translator::IMAGE_ATOMIC(AtomicOp op, const GcnInst& inst) {
...
+        return {};
+    }();
+
+    // Construct the handle
+    const IR::Value handle = ir.GetScalarReg(sharp_reg);
+
+    // Perform the atomic operation
+    const IR::Value prev = [&] {
+        switch (op) {
+        case AtomicOp::Swap:
```
raphaelthegreat: Using image atomics for buffer atomic instructions is questionable.

raphaelthegreat (changes requested):
A lot of changes in this PR are not correct sadly. I would normally wait until draft is removed but there is no indication in description about these being corrected so letting you know upfront

DanielSvoboda:
to make this pull a little simpler, I removed these: BUFFER_ATOMIC_ADD, EmitImageAtomicInc32, EmitImageAtomicDec32. I will try to fix DS_MAX and DS_MIN and convert ConvertU32x2U32 and ConvertU32x2U64

raphaelthegreat:
Also remove EmitBufferAtomicIMax32/Min as they not correct. The remaining DS changes are okay though.  Feel free to also look at my dark souls branch for some additional changes to support more DS read/write operations (like 64-bit pairs)

src/shader_recompiler/backend/spirv/emit_spirv_integer.cpp
```diff
@@ -242,6 +242,14 @@ Id EmitUClamp32(EmitContext& ctx, IR::Inst* inst, Id value, Id min, Id max) {
     return result;
 }
 
+Id EmitSLessThan(EmitContext& ctx, Id lhs, Id rhs) {
```
raphaelthegreat: This is the exact same as the EmitSLessThan32 right below

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -62,7 +73,25 @@ void Translator::S_BARRIER() {
 }
 
 void Translator::V_READFIRSTLANE_B32(const GcnInst& inst) {
```
raphaelthegreat: Keep the assert in V_READFIRSTLANE_B32 as the implementation is a stub that happens to work for RDR only. Also DS_MIN/DS_MAX should be removed

src/shader_recompiler/frontend/translate/translate.cpp
```diff
@@ -138,7 +138,11 @@ IR::U32F32 Translator::GetSrc(const InstOperand& operand, bool force_flt) {
         value = ir.Imm32(-0.5f);
         break;
     case OperandField::ConstFloatNeg_1_0:
-        value = ir.Imm32(-1.0f);
+        if (force_flt) {
+            value = ir.Imm32(-1.0f);
+        } else {
+            value = ir.Imm32(-1);
```
raphaelthegreat:
```suggestion
            value = ir.Imm32(std::bit_cast<u32>(-1.f));
```

src/shader_recompiler/frontend/translate/translate.cpp
```diff
@@ -160,6 +164,8 @@ IR::U32F32 Translator::GetSrc(const InstOperand& operand, bool force_flt) {
             value = ir.GetVccHi();
         }
         break;
+    case OperandField::M0:
```
raphaelthegreat: M0 stuff should be removed they were hacks

src/shader_recompiler/frontend/translate/translate.cpp
```diff
@@ -1158,6 +1200,9 @@ void Translate(IR::Block* block, u32 block_base, std::span<const GcnInst> inst_l
         case Opcode::S_WQM_B64:
         case Opcode::V_INTERP_P1_F32:
         case Opcode::S_ENDPGM:
+        case Opcode::BUFFER_ATOMIC_ADD:
```
raphaelthegreat: These should not be ignored on main

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -595,4 +604,27 @@ void Translator::V_CMP_CLASS_F32(const GcnInst& inst) {
     }
 }
 
+void Translator::V_MAD_U32(const GcnInst& inst) {
+    const IR::U32 src0{GetSrc(inst.src[0])};
+    const IR::U32 src1{GetSrc(inst.src[1])};
+    const IR::U32 src2{GetSrc(inst.src[2])};
+    // TODO
+    const IR::U32 result = ir.IAdd(ir.IMul(src0, src1), src2);
```
raphaelthegreat: This is incomplete from my branch, to get this in main you will need to add proper bitfield extracts as the ISA documentation specifies for V_MAD_U32_U24 (also change the name to that)

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -595,4 +604,27 @@ void Translator::V_CMP_CLASS_F32(const GcnInst& inst) {
...
+    SetDst(inst.dst[0], result);
+    ir.SetVccLo(result);
+}
+
+void Translator::V_MBCNT_U32_B32(bool is_low, const GcnInst& inst) {
+    const IR::U32 src0{GetSrc(inst.src[0])};
+    const IR::U32 src1{GetSrc(inst.src[1])};
```
raphaelthegreat: This implementation was tailored for NVIDIA (warp size 32). Query the subgroup size from vulkan backend, pass it in the profile structure and assert for it here

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -595,4 +604,27 @@ void Translator::V_CMP_CLASS_F32(const GcnInst& inst) {
...
+    const IR::U32 lane_id = ir.LaneId();
+    const IR::U32 thread_mask = ir.ISub(ir.ShiftLeftLogical(ir.Imm32(1), lane_id), ir.Imm32(1));
+    const IR::U1 is_odd_warp = ir.INotEqual(ir.BitwiseAnd(ir.WarpId(), ir.Imm32(1)), ir.Imm32(0));
+    const IR::U32 mask = IR::U32{ir.Select(is_odd_warp, is_low ? ir.Imm32(0) : thread_mask,
```
raphaelthegreat:
```suggestion
    const IR::U32 mask = IR::U32{ir.Select(is_odd_warp, is_low ? ir.Imm32(~0U) : thread_mask,
```
Small mistake I had found later

raphaelthegreat (changes requested):
I feel there is a misunderstanding with how atomics work here. There are 3 types of atomics, image atomics that work on storage images (implemented already), buffer atomics (all BUFFER_ATOMIC_* instructions) that work on storage buffers and shared atomics (DS instructions). You are attempting to use image atomics for the later 2 which will crash the recompiler if you tested it

src/shader_recompiler/frontend/translate/vector_alu.cpp
DanielSvoboda: Now the V_MAD_U32 function is performing a 24-bit extraction. However, I'm a little confused about whether this function should handle 32-bit operations instead.

Additionally, the V_MAD_U32_U24 function is calling V_MAD_I32_I24(inst, false). Should we consider changing this function as well?
updated code: https://github.com/shadps4-emu/shadPS4/pull/333/files#diff-9a6c2e2027c03231e88aaaab30908baecae202661839f35c31a777fec2500c7aR606

PR #339, opened 2024-07-29
Fix SearchFree function bug

Borchev:
MemoryManager::SearchFree throws when called with a virtual_addr below the minimum mapped address. This was seen when testing with Don't Starve Together: Console Edition. The fix is to clamp the virtual_addr parameter minimum to the system managed virtual base address.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -348,7 +348,15 @@ void MemoryManager::NameVirtualRange(VAddr virtual_addr, size_t size, std::strin
     it->second.name = name;
 }
 VAddr MemoryManager::SearchFree(VAddr virtual_addr, size_t size, u32 alignment) {
+    // If the requested address is below the mapped range, start search from the lowest address
+    auto min_search_address = impl.SystemManagedVirtualBase();
+    if (virtual_addr < min_search_address) {
+        virtual_addr = min_search_address;
+    }
+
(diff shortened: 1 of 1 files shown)


raphaelthegreat:
Can you send a log of the game? I find it strange the address ended up below system manged and mind hide an more underlying bug

Borchev:
Sure, attached. This log is from a build with this change: 
[shad_log_DST.txt](https://github.com/user-attachments/files/16418896/shad_log_DST.txt)

raphaelthegreat:
A log without this change would be a bit more useful as I can see which mapping function crashed it

Borchev:
Sure. Here you go.
[shad_log_DST_Stock.txt](https://github.com/user-attachments/files/16419056/shad_log_DST_Stock.txt)

raphaelthegreat:
And it crashes inside that MapDirectMemory call? The address looks normal 0x100000000

Borchev:
Yes, here is the callstack:
image

Here is the state of vma_map:
image

The problem is that the lowest mapped address range (on my machine) is 0x141fb0000 so the requested address doesnt exist in vma_map and FindVMA will return vma_map.end() which SearchFree attempts to access and crashes

raphaelthegreat:
I see. Makes sense then

Issue #342, opened 2024-07-30
Deleted


Emulator-Team-2:

EmulationPsycho:
This happens to me here too, it used to work fine!

![Captura de Tela (268)](https://github.com/user-attachments/assets/b5dc4ea6-6f81-4675-b5ad-ee3b85327a53)
![Captura de Tela (271)](https://github.com/user-attachments/assets/cffcf721-549c-454e-9ba2-4a20b937f45e)
![Captura de Tela (272)](https://github.com/user-attachments/assets/fa23df9a-81ad-4010-87c0-66ded6ddfae4)

viniciuslrangel:
I already found the problem, fixing it

EmulationPsycho:
> I already found the problem, fixing it

And where is the build to test? 😜

viniciuslrangel:
@EmulationPsycho https://github.com/viniciuslrangel/shadPS4/commit/cc8a299c034573eaf2ef94aa64da9cde9fdc54c0 Check the actions from my repo

EmulationPsycho:
> @EmulationPsycho [viniciuslrangel@cc8a299](https://github.com/viniciuslrangel/shadPS4/commit/cc8a299c034573eaf2ef94aa64da9cde9fdc54c0) Check the actions from my repo

Yes I can confirm that the problem is fixed now!

![Captura de Tela (284)](https://github.com/user-attachments/assets/d965c633-f1a0-449a-a67b-86d914f30430)

Emulator-Team-2:
Fixed by #414.

PR #347, opened 2024-07-31
Op arg float as default

viniciuslrangel:
After removing the 'force_flt' assumption based on input modifications, some games stopped running due to an early MOV operation that restricted the parser from inferring the value as a float, so I thought we should keep all values as float by default and force to integer only when needed by I32 OPs.

However I started getting a lot of errors from invalid types and had to run, wait to crash, and fix. So I changed the GetSrc implementation to infer the correct value based on return type, not only cleaning the code but making it easier for new OP to be implemented without needing to handle type conversion from GetSrc, including 64-bit types.

### DIFF
--- a/src/shader_recompiler/frontend/translate/translate.cpp
+++ b/src/shader_recompiler/frontend/translate/translate.cpp
@@ -73,9 +73,11 @@ void Translator::EmitPrologue() {
     }
 }
 
-template <>
-IR::U32F32 Translator::GetSrc(const InstOperand& operand, bool integer) {
+IR::U32F32 Translator::GetSrcRaw(const InstOperand& operand, bool integer) {
     IR::U32F32 value{};
+
+    ASSERT(operand.type != ScalarType::Float64 && operand.type != ScalarType::Uint64 &&
+           operand.type != ScalarType::Sint64);
 
     const bool is_float = operand.type == ScalarType::Float32 && !integer;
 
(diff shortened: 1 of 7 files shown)


viniciuslrangel:
Also fixes #342

src/shader_recompiler/frontend/translate/translate.cpp
```diff
@@ -313,12 +309,69 @@ IR::U64F64 Translator::GetSrc64(const InstOperand& operand, bool force_flt) {
 }
 
 template <>
-IR::U64 Translator::GetSrc64(const InstOperand& operand, bool force_flt) {
-    return GetSrc64<IR::U64F64>(operand, force_flt);
+Translator::SrcAuto Translator::GetSrc<Translator::SrcAuto>(const InstOperand& operand) {
```
raphaelthegreat: Lots of complication here I think. Not sure why this needed but if you want to avoid moving GetSrc in header, you can make it a template and add special definition for U32/F32 here. Also make the default type in the template as IR::F32

viniciuslrangel: This SrcAuto is to make the type infer based on the return type. Having this wrapper, it can call the correct method using the cast operator to the expected type, otherwise, all GetSrc would require the template parameter as it would construct a new IR::Value from whatever type it returns, losing all the compile-time check

viniciuslrangel:
Closing. Follow up #370

PR #348, opened 2024-07-31
Reorganization of includes

Xphalnos:


raphaelthegreat:
Nice, always wanted to fix those

roamic:
IMO includes should be (top-down) more-local to more-global in order to eliminate the errors where you remove include from .cpp which is not required and suddenly nothing compiles because your local header was depending on it.

PR #352, opened 2024-07-31
Better logo for shadPS4

Xphalnos:
I modified the logo I made some time ago by making it square (the old one was rectangular) and giving it white outlines (inspired by Cemu).
I'm always ready for changes!

Capture d'écran 2024-07-31 181246


shinra-electric:
If you need an .icns file for the macOS bundle, here is one I made earlier...
[shadPS4.icns.zip](https://github.com/user-attachments/files/16445851/shadPS4.icns.zip)

This is what it looks like: 

shadps4

Xphalnos:
> If you need an .icns file for the macOS bundle, here is one I made earlier...

Okay but where should I put it?

shinra-electric:
I guess `src/images`. CMake should point to it with: 
```
set( MACOSX_BUNDLE_ICON_FILE ${CMAKE_CURRENT_SOURCE_DIR}/src/images/shadPS4.icns )
```
and then it needs to be specified in the `Info.plist` 
```
<key>CFBundleIconFile</key>
<string>shadPS4.icns</string>
```

I'm not great at using CMake to build app bundles, so something else might need to be done. Best to check with @qurious-pixel or someone else who has Mac experience.

Xphalnos:
Indeed, I think it is better to do this in another commit because I don't know anything about Mac.

squidbus:
macOS icon added here: https://github.com/shadps4-emu/shadPS4/pull/356

PR #358, opened 2024-08-02
reuse: fix license for externals

abouvier:
These externals have their own license.


raphaelthegreat (approved):
Thanks for this, nice to see this oversight corrected

PR #362, opened 2024-08-03
Adding CMAKE_INTERPROCEDURAL_OPTIMIZATION

Xphalnos:
While exploring the Cemu source code, I [**found this**](https://github.com/cemu-project/Cemu/blob/517e68fe57ac1ac112f37c321d3d40a30ea5a8d6/CMakeLists.txt#L50) which might be useful.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,10 @@ function(create_target_directory_groups target_name)
     endforeach()
 endfunction()
 
+# enable link time optimization for release builds
+set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE ON)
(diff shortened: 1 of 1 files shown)


wheremyfoodat:
This should probably be enabled behind a CMake option. Enabling LTO for RelWithDebInfo builds unconditionally will inflate compile times and make development slower.

Xphalnos:
> This should probably be enabled behind a CMake option. Enabling LTO for RelWithDebInfo builds unconditionally will inflate compile times and make development slower.

Should I remove the config for RelWithDebInfo?

Xphalnos:
I'm closing this because I have another more important commit.

PR #367, opened 2024-08-05
Updated BatchMap2 to Handle ORBIS_KERNEL_MAP_OP_TYPE_PROTECT

menaman123:
* Implemented MProtect in memory.cpp that uses Protect function in address_space.cpp
* Implemented sceKernelMprotect in order for BatchMap2 to handle  ORBIS_KERNEL_MAP_OP_TYPE_PROTECT

### DIFF
--- a/documents/Quickstart/Quickstart.md
+++ b/documents/Quickstart/Quickstart.md
@@ -37,13 +37,18 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 - Windows 10 or Ubuntu 22.04
 
-## Have the latest WIP version
+## How to run the latest Work-in-Progress builds of ShadPS4
 
-When you go to Github Release, you have the latest major versions (e.g. v0.0.3), but if you want to have the latest Work-In-Progress version, you can go to Actions on Github to download it (Please note a Github account is required to be able to download).
+1. Go to <https://github.com/shadps4-emu/shadPS4/actions> and make sure you are logged into your GitHub account (important!)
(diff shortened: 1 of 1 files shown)


georgemoralis:
do you have any reference game?

menaman123:
> do you have any reference game?

Yes I was using Crash N Sane trilogy (CUSA07402)

StevenMiller123:
If fpPS4 is anything to go by, aren't sceKernelMprotect and sceKernelMtypeprotect two different functions? They use Mprotect for operation 2, and Mtypeprotect for operation 4.

menaman123:
> If fpPS4 is anything to go by, aren't sceKernelMprotect and sceKernelMtypeprotect two different functions? They use Mprotect for operation 2, and Mtypeprotect for operation 4.

Yup i just noticed, i will make those changes. I have a question. Looking at fpPS4 i see one of the parameters is mtype. What is the equivalent in shadPS4. I know prot is the MemoryProt

StevenMiller123:
Also, as a tester, I've found that your Mprotect implementation doesn't improve any games that crash on it (after adding Mprotect to the libkernel symbols). I tried a game that uses BatchMap op 4 as well (Crash Team Racing Nitro Fueled), and that doesn't progress any further than when removing the unimplemented line in the BatchMap function. As for your question, I'm not familiar enough with the shadPS4 source to give a clear answer.

menaman123:
Okay so i added MTypeprotect to it and just changed MProtect to ORBIS_KERNEL_MAP_OP_PROTECT

menaman123:
It seems like it calls the BatchMap2 function over and over again with the op 4 then it crashes. Not sure how to move pass this if anyone has any ideas.

red-prig:
mtype -> MemoryTypes

PR #371, opened 2024-08-06
Workflows cleanup + misc fixes

Xphalnos:
I cleaned the Workflows by removing anything unnecessary.
I also deleted things in CMakeLists.txt which was copied from Citra but is not useful to us.
I removed the "Ready" because it is not necessary to see it here.
The size of the macOS Qt artifact has been reduced by 0.4MB.

With the "Ready":
Capture d'écran 2024-08-06 131449

Without the "Ready":
Capture d'écran 2024-08-06 131402

### DIFF
--- a/.github/workflows/linux-qt.yml
+++ b/.github/workflows/linux-qt.yml
@@ -10,7 +10,6 @@ on:
     branches: [ "main" ]
 
 env:
-  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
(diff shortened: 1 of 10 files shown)


Xphalnos:
With the commit "SDL3: Disabling unnecessary options" the size of the artifacts has been further reduced.
Ready for review!

Before:
Capture d'écran 2024-08-07 172258
 
After:
Capture d'écran 2024-08-07 172811

PR #374, opened 2024-08-08
qt_gui: Refreshing game list after install directory change

SamuelFontes:
The game list wasn't being refreshed automaticly after a manual directory change on the QT GUI,  now the RefreshGameTable will be called after the GameInstallDialog is executed.


SamuelFontes:
Closed because I will recreate again in a proper branch, so when I sync it won't pollute the commits

PR #370, opened 2024-08-05
Surface management rework (3/3)

psucien:
Major changes:
- Initial support for overlapping images. Currently used for mip gen scenarios, can be extended to aliasing in future
- Handy debug features: rdoc markers, NV checkpoints, GPU assisted validation
- Gnm driver functions: `sceGnmGetGpuCoreClockFrequency` and `sceGnmSetVgtControl`

And small fixes here and there

PR #377, opened 2024-08-08
translator: Use templates for type information

raphaelthegreat:
The type enum isn't correct sometimes and when emitting instruction we always know what type we need. Also use float for V_MOV_B32 which fixes upside down grade like Kingdom Hearts


georgemoralis:
intergrate into #370

PR #384, opened 2024-08-08
Avoid crash when access violation address not in page table

Borchev:
Fix issue seen with Don't Starve Together.


raphaelthegreat:
Does the game use addresses in user area above 39bits? If that is the case the proper solution is to increase page table size

Borchev:
No, the highest address used before hitting the exception is 0x9C0988000. The first access exception address that causes an issue on my machine is 0xffffffffd14a4008 which seems suspect.

raphaelthegreat:
Yeah I feel there might an underlying cause here that needs to be fixed

Borchev:
That makes sense. I should have chosen my words better in my initial comment. This change was intended to suppress the crash and try to continue rather than fix the underlying issue. I'll abandon the PR if you prefer not to do this.

raphaelthegreat:
Unless it's really necessary probably avoid adding more checks as buffer fetch is a very hot path. So less the better

PR #386, opened 2024-08-08
Kernel-Related Fixes

StevenMiller123:
I noticed some easy to fix problems while testing some of my games. Please let me know if I should remove any of these changes.

### DIFF
--- a/src/core/libraries/kernel/memory_management.h
+++ b/src/core/libraries/kernel/memory_management.h
@@ -63,7 +63,7 @@ struct OrbisVirtualQueryInfo {
 
 struct OrbisKernelBatchMapEntry {
     void* start;
-    off_t offset;
+    size_t offset;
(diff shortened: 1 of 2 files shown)


red-prig:
`off_t` is not equal in size to `size_t` ? or what?
Alignment:
```
 [KernelBatchMapEntry]
  start:64
  offset:64
  length:64
  protection:8
  mtype:8
  pad1:16
  operation:32
```

raphaelthegreat:
Seems off_t is 32bits on some platforms. Strange

StevenMiller123:
When I looked it up, they were the same size, but off_t is signed while size_t is unsigned (I think). I tested on Windows and Ubuntu 23.10, and the change didn't break anything for me.

raphaelthegreat:
Looks good to me

PR #389, opened 2024-08-09
Use create_directories in sceKernelMkdir instead of create_directory

nblockbuster:
Fixes an error thrown when launching CUSA00568 because it attempts to make `/app0/temp/stress_cores` without making `/app0/temp` first


raziel1000:
Hi, this is incorrect. sceKernelMkdir calls mkdir which creates 1 dir only and need the parent dirs to exist which is equivalent to create_directory. if there is an error with this game it must be originating from somewhere else.

georgemoralis:
hmm i suppose that temp dir should be create in fs

georgemoralis:
https://github.com/shadps4-emu/shadPS4/pull/368/commits/61a6f633fd10cb9139b204ab1125c0e0e313bdaa

this should probably fix the issue without mkdirs

nblockbuster:
I see, thank you

PR #390, opened 2024-08-09
gui: Implement settings dialog

dima-xd:
As simple as it can be, without any abstraction and etc.

Example:
image


Xphalnos:
This is incredible progress!
Only one thing would be necessary: ​​remove the Save button. I think it's better to have a button that does both.

dima-xd:
> This is incredible progress! Only one thing would be necessary: ​​remove the Save button. I think it's better to have a button that does both.

Not sure about that.  For me it's not that necessary. Let's see what others say

ryzendew:
can you add 1 last thing ? to disable logs but have it on by default

dima-xd:
> can you add 1 last thing ? to disable logs but have it on by default

I don't want to add new functionalities in this PR. Maybe me or someone else will add it after this PR

PR #391, opened 2024-08-09
scePthreadAttrSetstack implementation

StevenMiller123:
This function is used by Final Fantasy XV. If there are any other Pthread functions I should implement in this PR, please let me know.

### DIFF
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -1086,6 +1086,16 @@ int PS4_SYSV_ABI scePthreadAttrGetstack(ScePthreadAttr* attr, void** addr, size_
     return SCE_KERNEL_ERROR_EINVAL;
 }
 
+int PS4_SYSV_ABI scePthreadAttrSetstack(ScePthreadAttr* attr, void* addr, size_t size) {
+    int result = pthread_attr_setstack(&(*attr)->pth_attr, addr, size);
+    LOG_INFO(Kernel_Pthread, "scePthreadAttrSetstack: result = {}", result);
+
+    if (result == 0) {
+        return SCE_OK;
+    }
+    return SCE_KERNEL_ERROR_EINVAL;
+}
+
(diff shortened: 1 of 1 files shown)


src/core/libraries/kernel/thread_management.cpp
```diff
@@ -1086,6 +1086,16 @@ int PS4_SYSV_ABI scePthreadAttrGetstack(ScePthreadAttr* attr, void** addr, size_
     return SCE_KERNEL_ERROR_EINVAL;
 }
 
+int PS4_SYSV_ABI scePthreadAttrSetstack(ScePthreadAttr* attr, void* addr, size_t size) {
+    int result = pthread_attr_setstack(&(*attr)->pth_attr, addr, size);
```
psucien, line 1093: Before calling this function, we need to ensure that:
* `attr` is a valid pointer
* `stack_addr` is a valid pointer
* `size` is less than specified by PS4 system maximum

src/core/libraries/kernel/thread_management.cpp
```diff
@@ -1086,6 +1086,16 @@ int PS4_SYSV_ABI scePthreadAttrGetstack(ScePthreadAttr* attr, void** addr, size_
...
+    LOG_INFO(Kernel_Pthread, "scePthreadAttrSetstack: result = {}", result);
+
+    if (result == 0) {
+        return SCE_OK;
+    }
+    return SCE_KERNEL_ERROR_EINVAL;
```
psucien: we use `ORBIS_` prefix there

psucien (changes requested):

StevenMiller123:
I've made the requested changes, let me know if there is anything else I need to change.

psucien:
Looks good now. However, I would ask you to change the stack size check in a way so we are sure that requested size is *not less* than system page size. So basically make it `size < 0x4000`

StevenMiller123:
I've made the change, it seems to work fine in games.

psucien (approved):
Thank you!

PR #392, opened 2024-08-09
qt_gui: Added stopping games by escape key or stop button

ElBread3:
This PR should theoretically add stopping games by either pressing escape or the stop button in the main window. This is implemented by creating a SDL window manager that's only current use is to hold a function that simply sends a quit event in order to close the window (less invasive than adding SDL headers into QT classes, also can be expanded in the future). It is important to note that currently the emulator doesn't seem to have any de-initialization functions which would need to be implemented so that exiting isn't basically a crash.


DeveloperVaibhav1:
try an alternative method for this .

DeveloperVaibhav1:
> This PR should theoretically add stopping games by either pressing escape or the stop button in the main window. This is implemented by creating a SDL window manager that's only current use is to hold a function that simply sends a quit event in order to close the window (less invasive than adding SDL headers into QT classes, also can be expanded in the future). It is important to note that currently the emulator doesn't seem to have any de-initialization functions which would need to be implemented so that exiting isn't basically a crash.
Its simply awesome

DeveloperVaibhav1:
fix it for clang format

DeveloperVaibhav1:
how to access the source code

ElBread3:
Closing as this is now old and will need to find a better way to do this in a more integrated fashion

Issue #232, opened 2024-07-02
VK 1.3 is required


EmuAl:
I have a AMD RX 6900XT GPU and the games i try crash with the following error:

[Debug] <Critical> vk_instance.cpp:operator():80: Assertion Failed!
Vulkan 1.3 is required, but only 1.2 is supported by device

any idea if this is fixable please

thanks

raphaelthegreat:
You need to update your drivers

EmuAl:
thanks but i am on latest amd gpu driver 24.6.1 ?

raphaelthegreat:
Then your GPU should definitely support 1.3

EmuAl:
![2024-07-02](https://github.com/shadps4-emu/shadPS4/assets/47227078/8fecfef1-d039-4a04-913c-83d29c21813b)
is this enough to use the emulator ?

raphaelthegreat:
Looks fine to me

EmuAl:
so must be a bug then ?

raphaelthegreat:
Hm probably. Try installing latest Vulkan SDK

EmuAl:
will do, thanks

EmuAl:
[Debug] <Critical> vk_instance.cpp:operator():80: Assertion Failed!
Vulkan 1.3 is required, but only 1.2 is supported by device!

same error after installation of VulkanSDK-1.3.283.0
hopefully it can fixed, will keep trying new versions as and when released

l33tm4st3r:
Hi all.

 Just remove "OpenCL™, OpenGL®, and Vulkan® Compatibility Pack" package from windows installed apps.

Have a nice day.

EmuAl:
Sorry but this isn't installed on my pc

RainKikyou:
I have the same with problem
![QQ截图20240705235727](https://github.com/shadps4-emu/shadPS4/assets/20301499/88183a2b-50d7-4301-8da6-d8f7b54548ae)
![image](https://github.com/shadps4-emu/shadPS4/assets/20301499/31c2a6f5-8837-45d0-827b-cb76e8f370ba)
![image](https://github.com/shadps4-emu/shadPS4/assets/20301499/de4878e6-1afc-41d9-8471-7949774d722e)
![QQ图片20240706124712](https://github.com/shadps4-emu/shadPS4/assets/20301499/3c0b9cbd-e46c-4aea-8c28-66f0634e2b51)

My PC
[...]

PR #393, opened 2024-08-09
vk_instance.cpp: Prefer device with most recent ApiVersion

SamuelFontes:
Changed to prefer the GPU with support for the most recent VulkanApiVersion to avoid selecting the onboard GPU. If both GPUs have support for the same version it will consider the memory heap size while sorting physical devices.

This should solve the issues people are having on https://github.com/shadps4-emu/shadPS4/issues/232.
The problem is that in some cases the GPU selected was the onboard GPU. Another solution would be to change the TargetVulkanApiVersion from VK_API_VERSION_1_2 to VK_API_VERSION_1_3.
https://github.com/shadps4-emu/shadPS4/blob/ab56665d4b75e13e99e824dc531b8642d0377632/src/video_core/renderer_vulkan/vk_platform.h#L20
But since I don't know what problems this might cause this pull request should be the safest fix.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_instance.cpp
+++ b/src/video_core/renderer_vulkan/vk_instance.cpp
@@ -73,7 +73,7 @@ Instance::Instance(Frontend::WindowSDL& window, s32 physical_device_index,
             const vk::PhysicalDeviceProperties& left_prop = std::get<1>(left).properties;
             const vk::PhysicalDeviceProperties& right_prop = std::get<1>(right).properties;
             if (left_prop.apiVersion >= TargetVulkanApiVersion &&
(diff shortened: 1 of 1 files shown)


squidbus:
> Another solution would be to change the TargetVulkanApiVersion from VK_API_VERSION_1_2 to VK_API_VERSION_1_3.

It's set to 1.2 because the minimum requirement is now 1.2 with certain extensions that were made core in 1.3, not full 1.3.

SamuelFontes:
Closed due to this not being the best way to solve the issue.
I'm going to be working on providing a way for people to select the GPU on the Qt GUI.

PR #394, opened 2024-08-09
thread_management.cpp: Various Mandatory Threading Fixes | Resolve #398

### LINKED ISSUE #398: Shadow of the colossus crashing before boot screen so far.
Here's my log file after trying to run Shadow Of the colossus on 0.1.1 WIP today's release of Windows QT, just to further help improve this emulator, keep up the good work :)

[shad_log.txt](https://github.com/user-attachments/files/16573432/shad_log.txt)

lzardy:
- scePthreadMutexInit did not return default when the mutex attributes were empty, resulting in invalid accesses.
Now, we return default as expected.
- Current HeapAPI was incomplete.
Now it has a full struct with all 10 preset function replacement fields.
Additionally, we now account for the full size 10 array parameter passed to _sceKernelRtldSetApplicationHeapAPI.
- HeapAPI malloc replacement unaccounted for
Now, we at least have a fallback to `std::malloc`, such that allocation completes without early crash within user thread.

Resolves #398

### DIFF
--- a/src/core/linker.cpp
+++ b/src/core/linker.cpp
@@ -305,7 +305,7 @@ void* Linker::TlsGetAddr(u64 module_index, u64 offset) {
         // Module was just loaded by above code. Allocate TLS block for it.
         Module* module = m_modules[module_index - 1].get();
         const u32 init_image_size = module->tls.init_image_size;
-        u8* dest = reinterpret_cast<u8*>(heap_api_func(module->tls.image_size));
+        u8* dest = reinterpret_cast<u8*>(heap_api->heap_malloc(module->tls.image_size));
         const u8* src = reinterpret_cast<const u8*>(module->tls.image_virtual_addr);
         std::memcpy(dest, src, init_image_size);
         std::memset(dest + init_image_size, 0, module->tls.image_size - init_image_size);
@@ -335,8 +335,8 @@ void Linker::InitTlsForThread(bool is_primary) {
             &addr_out, tls_aligned, 3, 0, "SceKernelPrimaryTcbTls");
         ASSERT_MSG(ret == 0, "Unable to allocate TLS+TCB for the primary thread");
     } else {
-        if (heap_api_func) {
-            addr_out = heap_api_func(total_tls_size);
+        if (heap_api) {
+            addr_out = heap_api->heap_malloc(total_tls_size);
         } else {
             addr_out = std::malloc(total_tls_size);
         }
--- a/src/core/linker.cpp
+++ b/src/core/linker.cpp
@@ -336,9 +336,17 @@ void Linker::InitTlsForThread(bool is_primary) {
         ASSERT_MSG(ret == 0, "Unable to allocate TLS+TCB for the primary thread");
     } else {
         if (heap_api) {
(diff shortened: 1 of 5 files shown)


dima-xd:
FYI Shadow of the Colossus crashes on `heap_api->heap_malloc(total_tls_size)`

georgemoralis:
@lzardy  any games you have based your work on ?

lzardy:
> @lzardy any games you have based your work on ?

Yes, these changes are for both CUSA08809 (Shadow of the Colossus) and CUSA18672 (Crysis 2 Remastered). Both of these games crash for the first case, when `scePthreadMutexInit` did not return default. The change is based on the kernel decomp to match functionality, in addition to the other fix for HeapAPI, although this one does not fix any crashes. I am currently working to resolve crashes related to the `heap_malloc` calls, as some games like Shadow of the Colossus (developed for earlier firmwares) expect default behavior that is missing although this is unconfirmed.

lzardy:
Shadow of the Colossus now launches without unexpected crash and Crysis 2 Remastered now launches up network initialization, where it crashes from a separate issue. This can be marked as a resolved PR, but for future reference it is recommended that tls block malloc replacement be implemented, as this PR simply maintains stability through a standard library fallback.

StevenMiller123:
Fixes hanging in Attack on Titan 2 as well. Doesn't cause any regressions for me.

src/core/linker.h
```diff
@@ -46,7 +46,21 @@ struct EntryParams {
...
+    PS4_SYSV_ABI void* (*heap_memalign)(size_t, size_t);
+    PS4_SYSV_ABI int (*heap_posix_memalign)(void**, size_t, size_t);
+    // NOTE: Fields below may be inaccurate
+    PS4_SYSV_ABI int (*heap_reallocalign)(void);
+    PS4_SYSV_ABI void (*heap_malloc_stats)(void);
+    PS4_SYSV_ABI int (*heap_malloc_stats_fast)(void);
+    PS4_SYSV_ABI size_t (*heap_malloc_usable_size)(void*);
+};
+
+typedef HeapAPI* AppHeapAPI;
```
raphaelthegreat:
```suggestion
using AppHeapAPI = HeapAPI*;
```

src/core/linker.cpp
```diff
@@ -335,10 +335,18 @@ void Linker::InitTlsForThread(bool is_primary) {
...
         ASSERT_MSG(ret == 0, "Unable to allocate TLS+TCB for the primary thread");
     } else {
-        if (heap_api_func) {
-            addr_out = heap_api_func(total_tls_size);
-        } else {
+        if (heap_api) {
+            LOG_ERROR(Core_Linker, "TLS user malloc called, using std::malloc");
+            // TODO: ref rtld_tls_block_malloc, tls malloc replacement
+            // addr_out = heap_api->heap_malloc(total_tls_size);
             addr_out = std::malloc(total_tls_size);
```
raphaelthegreat, line 347: Can this be limited to windows only? iirc linux has no such breakage and it's probably safer to use this. malloc on linux might return high addresses

PR #399, opened 2024-08-11
qt-gui: Added GPU device selection functionality

SamuelFontes:
This is a good solution for the problem people are having on the issue: https://github.com/shadps4-emu/shadPS4/issues/232
image

### DIFF
--- a/src/qt_gui/main_window.cpp
+++ b/src/qt_gui/main_window.cpp
@@ -168,7 +168,7 @@ void MainWindow::GetPhysicalDevices() {
             auto prop = physical_device.getProperties();
             QString name = QString::fromUtf8(prop.deviceName, -1);
             if (prop.apiVersion < Vulkan::TargetVulkanApiVersion) {
-                name += " * Unsupported API Version";
+                name += " * Unsupported Vulkan Version";
             }
             m_physical_devices.push_back(name);
         }
--- a/src/qt_gui/main_window.cpp
+++ b/src/qt_gui/main_window.cpp
@@ -161,17 +161,17 @@ void MainWindow::LoadGameLists() {
 }
 
 void MainWindow::GetPhysicalDevices() {
-        Vulkan::Instance instance(false, false);
-        auto physical_devices = instance.GetPhysicalDevices();
-        for (const vk::PhysicalDevice physical_device : physical_devices) {
-            auto prop = physical_device.getProperties();
-            QString name = QString::fromUtf8(prop.deviceName, -1);
-            if (prop.apiVersion < Vulkan::TargetVulkanApiVersion) {
-                name += " * Unsupported Vulkan Version";
-            }
-            m_physical_devices.push_back(name);
-        }
-    }
+	Vulkan::Instance instance(false, false);
+	auto physical_devices = instance.GetPhysicalDevices();
+	for (const vk::PhysicalDevice physical_device : physical_devices) {
+		auto prop = physical_device.getProperties();
+		QString name = QString::fromUtf8(prop.deviceName, -1);
+		if (prop.apiVersion < Vulkan::TargetVulkanApiVersion) {
+			name += " * Unsupported Vulkan Version";
+		}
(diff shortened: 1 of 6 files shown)


SamuelFontes:
This is how it will look when GPU doesn't support the target vulkan API version.
image

lzardy:
It may be better to say "Unsupported Vulkan Version", but good either way.

SamuelFontes:
> It may be better to say "Unsupported Vulkan Version", but good either way.


I agree, here is how it will look now:
image

georgemoralis:
Actually after macos we support vulkan 1.2 with needed 1.3 extensions

SamuelFontes:
This visual warning uses the Vulkan::TargetVulkanApiVersion variable, right now this is set to vulkan 1.2.
https://github.com/shadps4-emu/shadPS4/blob/2ba3221fc94df83d79103b957fd8888296c4f305/src/video_core/renderer_vulkan/vk_platform.h#L20
The main problem on the issue https://github.com/shadps4-emu/shadPS4/issues/232 was that users had to edit the config files manually to select the correct GPU by trial and error.  
This change provides a easy to use way for them to select the correct GPU, the warning is just a visual warning to make it clear to the user that the GPU doesn't support our targeted vulkan API version.

src/qt_gui/main_window.cpp
```diff
@@ -158,6 +160,21 @@ void MainWindow::LoadGameLists() {
     }
 }
 
+void MainWindow::GetPhysicalDevices() {
+    if (m_physical_devices.size() == 0) {
```
raphaelthegreat: If main window init is called once the size check here is probably unnecessary

src/qt_gui/settings_dialog.cpp
```diff
@@ -4,13 +4,20 @@
...
+    : QDialog(parent), ui(new Ui::SettingsDialog) {
     ui->setupUi(this);
     ui->tabWidgetSettings->setUsesScrollButtons(false);
     const auto config_dir = Common::FS::GetUserPath(Common::FS::PathType::UserDir);
 
     ui->buttonBox->button(QDialogButtonBox::StandardButton::Close)->setFocus();
 
+    // Add list of available GPUs
+    ui->graphicsAdapterBox->addItem("Auto Select"); // -1, auto selection
+    for (auto device = physical_devices->begin(); device != physical_devices->end(); ++device) {
```
raphaelthegreat:
```suggestion
    for (const auto& device : physical_devices) {
```

src/qt_gui/settings_dialog.h
```diff
@@ -16,7 +16,8 @@ class SettingsDialog;
 class SettingsDialog : public QDialog {
     Q_OBJECT
 public:
-    explicit SettingsDialog(QWidget* parent = nullptr);
+    explicit SettingsDialog(QWidget* parent = nullptr,
+                            std::vector<QString>* physical_devices = nullptr);
```
raphaelthegreat:
```suggestion
                            std::span<const QString> physical_devices);
```

src/qt_gui/main_window.cpp
SamuelFontes: Forgot to remove that, thanks for pointing it out.
It was some left over of me trying to get this done some other way.

src/qt_gui/settings_dialog.h
SamuelFontes: I've tried doing that way but for some reason it doesn't compile due to an error: missing default argument on parameter 'physical_devices'
image
Not sure if this is something on my environment only

raphaelthegreat: That is because the first function argument has a default argument. Default argument are only supported at end of list. Either remove default or move std::span as first arg

SamuelFontes: Thanks that solved the issue.

raphaelthegreat (approved):
Thanks

PR #407, opened 2024-08-12
Basic gamepad support through SDL

counter185:
Ideally we'd have a whole GUI to configure gamepads for every player but that's kind of a long-term goal, so for now I made it just take the first SDL-compatible controller and pass it's buttons and axes to `WindowSDL`'s `controller`, just like the current keyboard handler does. It's not much, but I think it will make early testing less annoying, especially for me with an 80% keyboard with no numpad.

### DIFF
--- a/src/sdl_window.cpp
+++ b/src/sdl_window.cpp
@@ -289,7 +289,7 @@ void WindowSDL::onGamepadEvent(const SDL_Event* event) {
 
     u32 button = 0;
     Input::Axis axis = Input::Axis::AxisMax;
-    switch (event->type) { 
+    switch (event->type) {
     case SDL_EVENT_GAMEPAD_BUTTON_DOWN:
     case SDL_EVENT_GAMEPAD_BUTTON_UP:
         button = sdlGamepadToOrbisButton(event->gbutton.button);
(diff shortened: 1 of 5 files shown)


Shoegzer:
This works well as tested in Linux Mint 22 / kernel 6.8 with a dualsense pad.  If you haven't already, you may want to take a look at #210 which was not merged and closed for unknown reasons but includes working rumble.

src/sdl_window.cpp
```diff
@@ -276,4 +294,49 @@ void WindowSDL::onKeyPress(const SDL_Event* event) {
...
+void WindowSDL::onGamepadEvent(const SDL_Event* event) {
+    using Libraries::Pad::OrbisPadButtonDataOffset;
+
+    u32 button = 0;
+    Input::Axis axis = Input::Axis::AxisMax;
+    switch (event->type) { 
+    case SDL_EVENT_GAMEPAD_BUTTON_DOWN:
+    case SDL_EVENT_GAMEPAD_BUTTON_UP:
+        button = 
+            event->gbutton.button == SDL_GAMEPAD_BUTTON_DPAD_DOWN ? OrbisPadButtonDataOffset::ORBIS_PAD_BUTTON_DOWN
```
raphaelthegreat: I feel this can be done as a helper function with switch statement to convert the enum

counter185:
Made some changes, added lightbar functionality, vibration support based on #210 and moved the button bindings to a helper function. Not sure if `scePadSetLightBar` is supposed to return `ORBIS_PAD_ERROR_INVALID_ARG` or a different error code on null parameter, but I don't have a way of checking that

raphaelthegreat:
Thanks a lot for this

PR #414, opened 2024-08-12
video_core: Crucial buffer cache fixes + proper GPU clears

raphaelthegreat:
Cherry-pick from https://github.com/shadps4-emu/shadPS4/pull/370 mostly.
* Buffer offsets are now defined upfront. We can't define them on demand like previously as a buffer access might be inside a scope. If the same buffer is accessed again under a different scope it is invalid to reuse the Id of offset.
* Use hashes to protect images from accidental re-upload. Guest might write some CPU data to the same page as a render target for example. This also fixes some flickering in simpler 2D games due to that.
* When refreshing an image attempt to use a buffer provided from the buffer cache instead of uploading always from CPU. This particularly affects clears as guest performs them by writing to memory region of render target viewed as SSBO. Now the entire process can happen in the GPU giving more performance, and more correctness; we no longer upload whatever garbage values the CPU had in that region

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -248,6 +248,19 @@ std::pair<Buffer*, u32> BufferCache::ObtainBuffer(VAddr device_addr, u32 size, b
     return {&buffer, buffer.Offset(device_addr)};
 }
 
+std::pair<const Buffer*, u32> BufferCache::ObtainTempBuffer(VAddr gpu_addr, u32 size) {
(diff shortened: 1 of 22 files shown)


src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -8,6 +8,7 @@
 #include "video_core/amdgpu/liverpool.h"
 #include "video_core/buffer_cache/buffer_cache.h"
 #include "video_core/renderer_vulkan/liverpool_to_vk.h"
+#include "video_core/renderer_vulkan/liverpool_to_vk.h"
```
Exhigh: duplicate

PR #420, opened 2024-08-12
Fix some Vulkan validation errors on macOS.

squidbus:
* Set up portability extension requirements to allow using the standard Vulkan loader on macOS, which in turn allows loading validation layers.
* Fix some validation errors for unsupported extensions in the device structure chain.
* Fix using mailbox present mode when unsupported, which was also causing frame pacing issues on macOS.


squidbus:
Rebased

raphaelthegreat (approved):
Thanks looks good

PR #427, opened 2024-08-13
spirv: Simplify shared memory handling

raphaelthegreat:
* Remove 8 and 16bit functions. They are unused add some additional complexity. We can always add it back later.
* Remove usage of shared memory aliasing. Dark Souls uses shared memory in fragment stage and causes the game to freeze as this feature is for compute only


EmulationChannel:
[shad_log.txt](https://github.com/user-attachments/files/16605947/shad_log.txt)
RED DEAD REDEMPTION

raphaelthegreat:
@EmulationChannel Can you stop posting the same log on every pull request? We will let you know when the RDR issue is fixed in PR description

PR #428, opened 2024-08-14
shader_recompiler: BUFFER_ATOMIC & DS_* Opcodes

lzardy:
- Emitters for BufferAtomic and SharedAtomic
- Addition of BufferAtomic and SharedAtomic opcodes to IR
- Translator for BUFFER_ATOMIC Opcodes (currently only BUFFER_ATOMIC_ADD is supported)
- Translators for DS_MAX/MIN/ADD_U32 Opcodes


src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
```diff
@@ -21,6 +34,52 @@ Id ImageAtomicU32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords, Id va
...
+}
+
+Id EmitBufferAtomicInc32(EmitContext&, IR::Inst*, u32, Id, Id) {
+    // TODO: This is not yet implemented
+    throw NotImplementedException("SPIR-V Instruction");
+}
+
+Id EmitBufferAtomicDec32(EmitContext&, IR::Inst*, u32, Id, Id) {
+    // TODO: This is not yet implemented
+    throw NotImplementedException("SPIR-V Instruction");
```
roamic: use `UNREACHABLE_MSG` instead of throws so that it goes to log

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -110,6 +116,50 @@ void Translator::DS_WRITE(int bit_size, bool is_signed, bool is_pair, const GcnI
...
+    if (inst.dst[0].type != ScalarType::Undefined) {
+        SetDst(inst.dst[0], old_value);
+    }
+}
+
+void Translator::DS_ADD_U32(const GcnInst& inst) {
+    const IR::U32 addr{GetSrc(inst.src[0])};
+    const IR::U32 data{GetSrc(inst.src[1])};
+    const IR::U32 offset = ir.Imm32(u32(inst.control.ds.offset0));
+    const IR::U32 addr_offset = ir.IAdd(addr, offset);
```
raphaelthegreat, line 153: This also need a special IR instruction (can call it SharedAtomicIAdd32). Also I don't think its necessary to align down address as guest should pass dword aligned values when reading dwords

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -110,6 +116,50 @@ void Translator::DS_WRITE(int bit_size, bool is_signed, bool is_pair, const GcnI
...
+    const IR::U32 aligned_addr = ir.BitwiseAnd(addr_offset, ir.Imm32(~3));
+    const IR::U32 old_value = IR::U32(ir.LoadShared(32, false, aligned_addr));
+    const IR::U32 new_value = ir.IAdd(old_value, data);
+    ir.WriteShared(32, new_value, aligned_addr);
+    if (inst.dst[0].type != ScalarType::Undefined) {
```
raphaelthegreat: Are there cases where dst type is undefined?

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -110,6 +116,50 @@ void Translator::DS_WRITE(int bit_size, bool is_signed, bool is_pair, const GcnI
     }
 }
 
+void Translator::DS_MAX_U32(const GcnInst& inst) {
+    const IR::U32 addr{GetSrc(inst.src[0])};
+    const IR::U32 data{GetSrc(inst.src[1])};
+    const IR::U32 offset = ir.Imm32(u32(inst.control.ds.offset0));
+    const IR::U32 addr_offset = ir.IAdd(addr, offset);
+
+    const IR::U32 old_value = IR::U32(ir.LoadShared(32, false, addr_offset));
```
raphaelthegreat: This needs a special IR opcode (something like SharedAtomic(U/S)Max32) so you can call OpAtomic(U/S)Max in spirv backend. The difference between buffers and this is the pointer type, buffers have spv::StorageClass::StorageBuffer while shared atomics have spv::StorageClass::Workgroup

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -110,6 +116,50 @@ void Translator::DS_WRITE(int bit_size, bool is_signed, bool is_pair, const GcnI
...
+}
+
+void Translator::DS_MIN_U32(const GcnInst& inst) {
+    const IR::U32 addr{GetSrc(inst.src[0])};
+    const IR::U32 data{GetSrc(inst.src[1])};
+    const IR::U32 offset = ir.Imm32(u32(inst.control.ds.offset0));
+    const IR::U32 addr_offset = ir.IAdd(addr, offset);
+
+    const IR::U32 old_value = IR::U32(ir.LoadShared(32, false, addr_offset));
+    const IR::U32 new_value = ir.UMin(old_value, data);
```
raphaelthegreat: Ditto here as for max

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -110,6 +116,42 @@ void Translator::DS_WRITE(int bit_size, bool is_signed, bool is_pair, const GcnI
...
 }
 
+void Translator::DS_ADD_U32(const GcnInst& inst) {
+    const IR::U32 addr{GetSrc(inst.src[0])};
+    const IR::U32 data{GetSrc(inst.src[1])};
+    const IR::U32 offset = ir.Imm32(u32(inst.control.ds.offset0));
+    const IR::U32 addr_offset = ir.IAdd(addr, offset);
+
+    const IR::U32 value = ir.SharedAtomicIAdd(addr_offset, data);
+    ir.WriteShared(32, value, addr_offset);
```
raphaelthegreat: The WriteShared call is not needed as OpAtomicIAdd will write the result as well. The value returned is the original anyway so this is undoing the operation

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -110,6 +116,42 @@ void Translator::DS_WRITE(int bit_size, bool is_signed, bool is_pair, const GcnI
...
+}
+
+void Translator::DS_MIN_U32(const GcnInst& inst) {
+    const IR::U32 addr{GetSrc(inst.src[0])};
+    const IR::U32 data{GetSrc(inst.src[1])};
+    const IR::U32 offset = ir.Imm32(u32(inst.control.ds.offset0));
+    const IR::U32 addr_offset = ir.IAdd(addr, offset);
+
+    const IR::U32 value = ir.SharedAtomicUMax(addr_offset, data);
+    ir.WriteShared(32, value, addr_offset);
```
raphaelthegreat: Ditto

src/shader_recompiler/frontend/translate/vector_memory.cpp
```diff
@@ -430,6 +434,67 @@ void Translator::BUFFER_STORE_FORMAT(u32 num_dwords, bool is_typed, bool is_form
...
+    IR::BufferInstInfo info{};
+    info.index_enable.Assign(mtbuf.idxen);
+    info.offset_enable.Assign(mtbuf.offen);
+    info.inst_offset.Assign(mtbuf.offset);
+
+    const IR::Value handle =
+        ir.CompositeConstruct(ir.GetScalarReg(sharp), ir.GetScalarReg(sharp + 1),
+                              ir.GetScalarReg(sharp + 2), ir.GetScalarReg(sharp + 3));
+
+    const IR::Value value = ir.LoadBufferFormat(num_dwords, handle, address, info);
```
raphaelthegreat: The value being added in the buffer atomic does not come from the buffer, it is from VDATA register (src[1]). Also if GLC == 1 you need to writeback the result of the atomic to VDATA as well

src/shader_recompiler/ir/ir_emitter.h
```diff
@@ -96,6 +104,33 @@ class IREmitter {
     void StoreBufferFormat(int num_dwords, const Value& handle, const Value& address,
                            const Value& data, BufferInstInfo info);
 
+    [[nodiscard]] Value BufferAtomicIAdd(const Value& handle, const Value& a, const Value& b,
+                                         BufferInstInfo info);
+    [[nodiscard]] Value BufferAtomicSMin(const Value& handle, const Value& a, const Value& b,
```
raphaelthegreat: If BufferAtomicIMin exists with is_signed argument, having separate BufferAtomicSMin/BufferAtomicUMin function is unnecessary, so you can remove them and select which opcode to emit directly from IMin

raphaelthegreat:
Oh additionally you also need to add all the new instructions added here to https://github.com/shadps4-emu/shadPS4/blob/8c77d4dde661ff565e4a3dccc72a0b7028178e68/src/shader_recompiler/ir/microinstruction.cpp#L41 as the dead code pass might eliminate the atomic otherwise if it does not write its result

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
```diff
@@ -12,6 +12,29 @@ std::pair<Id, Id> AtomicArgs(EmitContext& ctx) {
...
+    const Id shift_id{ctx.ConstU32(2U)};
+    const Id index{ctx.OpShiftRightArithmetic(ctx.U32[1], offset, shift_id)};
+    const Id pointer{ctx.OpAccessChain(ctx.shared_u32, ctx.shared_memory_u32, index)};
+    const auto [scope, semantics]{AtomicArgs(ctx)};
+    return (ctx.*atomic_func)(ctx.U32[1], pointer, scope, semantics, value);
+}
+
+Id BufferAtomicU32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id address, Id value,
+                   Id (Sirit::Module::*atomic_func)(Id, Id, Id, Id, Id)) {
+    // Get srsrc buffer
```
raphaelthegreat: Comments here can prob be removed the code is pretty clear what it does (mostly style nit this tbh)

src/shader_recompiler/frontend/translate/vector_memory.cpp
```diff
@@ -435,6 +439,46 @@ void Translator::BUFFER_STORE_FORMAT(u32 num_dwords, bool is_typed, bool is_form
     }
 }
 
+void Translator::BUFFER_ATOMIC(u32 num_dwords, AtomicOp op, const GcnInst& inst) {
+    // Get controls for mubuf-specific instructions
+    const auto& mubuf = inst.control.mubuf;
```
raphaelthegreat, line 444: Here as well

src/shader_recompiler/frontend/translate/vector_memory.cpp
```diff
@@ -435,6 +439,46 @@ void Translator::BUFFER_STORE_FORMAT(u32 num_dwords, bool is_typed, bool is_form
     }
 }
 
+void Translator::BUFFER_ATOMIC(u32 num_dwords, AtomicOp op, const GcnInst& inst) {
```
raphaelthegreat:
```suggestion
void Translator::BUFFER_ATOMIC(AtomicOp op, const GcnInst& inst) {
```

src/shader_recompiler/frontend/translate/vector_memory.cpp
```diff
@@ -435,6 +439,46 @@ void Translator::BUFFER_STORE_FORMAT(u32 num_dwords, bool is_typed, bool is_form
...
+
+    // Construct srsrc SGPRs
+    const IR::Value handle =
+        ir.CompositeConstruct(ir.GetScalarReg(srsrc), ir.GetScalarReg(srsrc + 1),
+                              ir.GetScalarReg(srsrc + 2), ir.GetScalarReg(srsrc + 3));
+
+    // Apply atomic op
+    // derefs srsrc buffer and adds vdata value to it
+    // then returns original srsrc buffer value
+    const IR::Value original_val = ir.BufferAtomicIAdd(handle, address, vdata_val, info);
```
raphaelthegreat: I think the switch that was here is fine, since most buffer atomics are implemented on backend side anyway.

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -42,6 +42,17 @@ bool IsBufferInstruction(const IR::Inst& inst) {
...
+    case IR::Opcode::BufferAtomicIAdd32:
+    case IR::Opcode::BufferAtomicSMin32:
+    case IR::Opcode::BufferAtomicUMin32:
+    case IR::Opcode::BufferAtomicSMax32:
+    case IR::Opcode::BufferAtomicUMax32:
+    case IR::Opcode::BufferAtomicInc32:
+    case IR::Opcode::BufferAtomicDec32:
+    case IR::Opcode::BufferAtomicAnd32:
+    case IR::Opcode::BufferAtomicOr32:
+    case IR::Opcode::BufferAtomicXor32:
```
raphaelthegreat, line 34: These instructions need to be added to IsBufferStore() so the buffer handle is translated as storage buffer. Given they are used in many places I recommend you make a dedicated function IsBufferAtomic() and use it where needed

raphaelthegreat (approved):
Thanks

PR #434, opened 2024-08-14
control_flow_graph: Initial divergence handling

raphaelthegreat:
AMD hardware is inherently incapable of "diverging" (some threads in a warp to take a different code path than others). The GPU maintains a separate register named EXEC which is a bit-mask of all currently active threads. If the corresponding EXEC bit of a thread is zero, vector operations will be suppressed.

Normally this would be quite a pain to handle, however the compiler helps us quite a bit here. Since masking threads means that the GPU will still execute the instruction, but discard its result, the compiler will often add checks if EXEC = 0 (i.e all threads agree on a certain condition = condition is uniform) and use S_CBRANCH_EXECZ to skip over code blocks in more rapid succession. Our recompiler uses these branches as hints for reconstructing control flow in SPIR-V.

However that is often not enough. If the code that needs to be conditionally executed is small enough, where the branch instruction itself might be more costly, then the compiler will instead mask EXEC with condition, perform necessary operations and then restore it. This is currently not handled in the recompiler which causes a variety of bugs.

A simple example of this is the following code, a bitwise or needs to be conditionally executed so it wrapped in EXEC saving and restoring instructions.
```
/*0000000001a8*/ s_and_saveexec_b64 s[4:5], vcc
/*0000000001ac*/ v_or_b32        v3, 4, v3
/*0000000001b0*/ s_mov_b64       exec, s[4:5]
```
This can happen with more than 1 instruction as well as shown here
```
/*0000000004ec*/ s_and_saveexec_b64 vcc, vcc
/*0000000004f0*/ v_rcp_f32       v0, v17
/*0000000004f4*/ v_mad_f32       v2, v0, -v16, 1.0 clamp
/*0000000004fc*/ s_load_dwordx8  s[20:27], s[8:9], 0x0
/*000000000500*/ s_mov_b64       exec, vcc
```
Sometimes the compiler can also insert an instruction between an EXEC saving instruction and a branch. This is not strictly necessary but might be chosen by the compiler for more optimal instruction pipelining
```
/*00000000028c*/ s_and_saveexec_b64 s[2:3], vcc
/*000000000290*/ s_cbranch_execz .L700_0
...
.L700_0:
/*0000000002bc*/ s_andn2_b64     exec, s[2:3], exec
/*0000000002c0*/ v_madak_f32     v17, 2.0, v15, 0x3f800000
/*0000000002c8*/ s_cbranch_execz .L860_0
```

Instructions inside theses EXEC "scopes" need to be wrapped in conditionals to be emulated properly. This is the purpose of this PR. The handling of this is done at the CFG level, where we can easily insert new basic blocks and have later stages auto-magically convert these into valid SPIR-V scopes. Between label emissions from branches and block linking we insert a new stage which attempts to annotate these scopes with additional labels so they get dedicated basic blocks. Most cases should be handled, however not all. There are various combinations of open and close scope instructions (some like S_ANDN2_B64 can even be used for both!)

### DIFF
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -1064,16 +1064,7 @@ ScePthread PThreadPool::Create() {
         }
     }
 
-#ifdef _WIN64
     auto* ret = new PthreadInternal{};
-#else
-    // TODO: Linux specific hack
-    static u8* hint_address = reinterpret_cast<u8*>(0x7FFFFC000ULL);
-    auto* ret = reinterpret_cast<PthreadInternal*>(
-        mmap(hint_address, sizeof(PthreadInternal), PROT_READ | PROT_WRITE,
-             MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0));
-    hint_address += Common::AlignUp(sizeof(PthreadInternal), 4_KB);
-#endif
     ret->is_free = false;
     ret->is_detached = false;
     ret->is_almost_done = false;
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -1064,7 +1064,16 @@ ScePthread PThreadPool::Create() {
         }
     }
 
+#ifdef _WIN64
     auto* ret = new PthreadInternal{};
+#else
+    // TODO: Linux specific hack
+    static u8* hint_address = reinterpret_cast<u8*>(0x7FFFFC000ULL);
+    auto* ret = reinterpret_cast<PthreadInternal*>(
+        mmap(hint_address, sizeof(PthreadInternal), PROT_READ | PROT_WRITE,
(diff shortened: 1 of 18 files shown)


raphaelthegreat:
This fixed Dark Souls blurry backgrounds and improves Bloodborne Alpha internal rendering (no more green framebuffer)

raphaelthegreat:
This needs a lot of testing in case some game regress or crash

bigol83:
Red Dead Redemption crashes before main menu with this log. I am on Intel i5 12400f rtx 3060 Windows 10

[shad_log.txt](https://github.com/user-attachments/files/16623973/shad_log.txt)

gandalfthewhite19890404:
> Red Dead Redemption crashes before main menu with this log. I am on Intel i5 12400f rtx 3060 Windows 10
> 
> [shad_log.txt](https://github.com/user-attachments/files/16623973/shad_log.txt)

crashes on main build, so it is not current PR issue.

PR #436, opened 2024-08-14
ERROR codes

DanielSvoboda:
Removed error_codes present in separate files, unifying them all in one place, then these codes were removed:
> src/core/libraries/dialogs/error_codes.h
> src/core/libraries/save_data/error_codes.h
> src/core/libraries/disc_map/disc_map_codes.h

Removed some duplicates. Standardizes all as ORBIS_

Added the remaining codes present in psdevwiki.

The comment of the value in decimal was included next to all codes, to make it easier to locate:

image


georgemoralis:
After some talk with the other team members we think that is better each library has it's own error code header files and common error codes goes to root library folder.

PR #438, opened 2024-08-15
gpu: handle primitive restart index register

viniciuslrangel:
Fix most 3d models vertex in project diva future tone


raphaelthegreat (approved):
Thanks

PR #439, opened 2024-08-15
[WIP] libSceNgs2

lzardy:


georgemoralis:
seems nice are you going to continue it? If so one adjustment is to load ngs2 library like the others that has lle alternatives in emulator.cpp , so if no lle ngs2 file found it will load the HLE alternative instead.

Surely the purpose is one day to have all the modules in HLE but that will require additional work that will be done a bit later

src/core/libraries/libs.cpp
```diff
@@ -79,6 +80,7 @@ void InitHLELibs(Core::Loader::SymbolsResolver* sym) {
     Libraries::ErrorDialog::RegisterlibSceErrorDialog(sym);
     Libraries::ImeDialog::RegisterlibSceImeDialog(sym);
     Libraries::AvPlayer::RegisterlibSceAvPlayer(sym);
+    Libraries::Ngs2::RegisterlibSceNgs2(sym);
```
georgemoralis: This should be removed since you added it emulator.cpp

lzardy:
> seems nice are you going to continue it? If so one adjustment is to load ngs2 library like the others that has lle alternatives in emulator.cpp , so if no lle ngs2 file found it will load the HLE alternative instead.
> 
> Surely the purpose is one day to have all the modules in HLE but that will require additional work that will be done a bit later

Yes I will continue it. If it is necessary that development be faster, I can stop here for the review and you can do push to main, so others can contribute to each function.
Otherwise, I can continue on my own to do the system from top down.

src/core/libraries/error_codes.h
```diff
@@ -472,3 +472,56 @@ constexpr int ORBIS_AVPLAYER_ERROR_INFO_OTHER_ENCRY = 0x806A00BF;
 
 // AppContent library
 constexpr int ORBIS_APP_CONTENT_ERROR_PARAMETER = 0x80D90002;
+
+// Ngs2 library
+constexpr int ORBIS_NGS2_ERROR_INVALID_PARAMETERS = 0x804A0001;
```
georgemoralis: Error codes should go to an ngs2_error.h file in the same directory as ngs2. That is what we decided lately

georgemoralis:
Left some review fixes

georgemoralis:
Ok all good now. I guess we can merge it if you want so it will be synced with main and you can open other following pr's for this

lzardy:
i think it is good enough spot for people to work off of, am okay with merge

PR #443, opened 2024-08-15
Qt-GUI: Cleaning the option menu

Xphalnos:
I deleted the "Emulator" tab because there was only the language selection.
I moved it to "General".
Always ready for changes!

Capture d'écran 2024-08-15 150952


Xphalnos:
Could someone restart the "Windows-Qt" build?
This is due to an internal error.

image

PR #448, opened 2024-08-16
Fix for issue 331

Sailanarmo:
After studying the for loop I became aware that the loop only needed to iterate to the size of labels - 2. 

This line: 
```cpp
  const bool is_last = next_it == labels.end();
```
Would check if `next_it` was the last position in the container and return from the function. I simplified the for loop by terminating the loop as it gets to the second to last position in the container. So the logic goes: 
```cpp
[1,2,3,4,5,6]
bounds = size - 2 = 4

i = 1;
i + 1 = 2;
insert[1,2];
++i;
1 < bounds true

i = 2;
i + 1 = 3;
insert[2,3];
++i;
2 < bounds true;

i = 3;
i + 1 = 4;
insert[3,4];
++i;
3 < bounds true;

i = 4;
i + 1 = 5;
insert[4,5];
++i;

4 < 4 false
terminate loop
```
This should eliminate the need for the check if it is the last element in the container because we are only iterating to the second to last position of the container. The only concern I have is `inst_list`. I did not find anywhere that makes it safe to assume that the size of `labels` is the same size of `inst_list`. I think that should have some looking into.


Exhigh:
Was this patch even tested as a fix for the problem ?

src/shader_recompiler/frontend/control_flow_graph.cpp
```diff
@@ -94,25 +94,25 @@ void CFG::EmitBlocks() {
...
+
+        // TODO: Investigate how to make sure this does not go out of bounds. 
+        //       Is inst_list always the size of labels? 
+        const auto& end_inst = inst_list[i+1]; 
 
-        // Insert block between the labels using the last instruction
-        // as an indicator for branching type.
         Block* block = block_pool.Create();
         block->begin = start;
-        block->end = end;
```
Exhigh: why was this removed ?

src/shader_recompiler/frontend/control_flow_graph.cpp
```diff
@@ -94,25 +94,25 @@ void CFG::EmitBlocks() {
...
-        const bool is_last = next_it == labels.end();
-        if (is_last) {
-            // Last label is special.
-            return;
-        }
-        const Label end = *next_it;
-        const size_t end_index = get_index(end) - 1;
-        const auto& end_inst = inst_list[end_index];
+    // Check to make sure there are lables in the vector and that they have data.
+    ASSERT(!lables.empty() && lables.size() > 2);
```
Exhigh:
```suggestion
    ASSERT(!labels.empty() && labels.size() > 2);
```

src/shader_recompiler/frontend/control_flow_graph.cpp
```diff
@@ -94,25 +94,25 @@ void CFG::EmitBlocks() {
...
+
+    // Loop through every pair except the last two: 
+    // [0,1,2,3,4,5,6]
+    // [0,1], [1,2], [2,3], [3,4], [4,5] 
+    for (int i = 0; i < labels.size() - 2; ++i)
+    {
+        const Label start = lables[i];
```
Exhigh:
```suggestion
        const Label start = labels[i];
```

Exhigh:
Even with these suggested changes, this patch doesn't actually work.

raphaelthegreat:
This is wrong in many ways. You are indexing instruction list with the index of the label which doesn't make sense. You removed the end address of the block. You also changed begin and end indices to be label indexes when they should be instruction list indices.

src/shader_recompiler/frontend/control_flow_graph.cpp
Sailanarmo: It should not have been removed I must have missed that. Sorry.

Sailanarmo:
> This is wrong in many ways. You are indexing instruction list with the index of the label which doesn't make sense. You removed the end address of the block. You also changed begin and end indices to be label indexes when they should be instruction list indices.

image

If that is the case then why is labels the container that is being iterated over and not `inst_list`? I kept the functionality the same. However, I am now noticing the lambda function `get_index`. I'll assumed wrongly that it was grabbing the index of itself and not the index of another container.

Sailanarmo:
> Was this patch even tested as a fix for the problem ?

@Exhigh, I do not have the game set up to test it on. I saw a problem with the loop and I presented a fix. The comment that has the TODO on it is really going to be the issue. There is an assumption that `inst_list` must be the same size as `labels` or `index_to_pc`. 

image

`index_to_pc` gets resized here by `inst_list` size + 1 but the logic here:

image

Doesn't really guarantee that labels are going to always be added.

viniciuslrangel:
index_to_pc is the same size as inst_list because it's the address of each instruction, used to track branching which uses relative addresses. The `get_index` returns the index of the instruction for a given Label. The labels are only created when there is a branching instruction, so it will always be smaller than the number of instructions. You are mixing up the purpose of these two arrays

Sailanarmo:
> index_to_pc is the same size as inst_list because it's the address of each instruction, used to track branching which uses relative addresses. The `get_index` returns the index of the instruction for a given Label. The labels are only created when there is a branching instruction, so it will always be smaller than the number of instructions. You are mixing up the purpose of these two arrays

@viniciuslrangel Yeah, I saw the `getindex` function and I am putting those back in. However, that still does not answer why we are then iterating over the `labels` container and indexing out of the `inst_list`. Would it not have been better to have labels as an `std::unordered_map` that would map directly to the contents of the two other containers? That way it is always guaranteed to map to it and it would not have to worry about out of bounds indexing? 

>  index_to_pc is the same size as inst_list

Is this true? If you have a `std::vector<int> a{1};` and another `std::vector<int> b` and you call `b.resize(a.size() + 1);` then the size of `b` will be 2. If it is supposed to be the same size why is the +1 added to `.resize()`?

viniciuslrangel:
@Sailanarmo The get_index is trying to find the nearest instruction, it's not exactly a specific value. If you check the implementation, it uses `lower_bound`, not an index. The resize has the `+ 1` because at the end of `EmitLabels`, the last value of index_to_pc points to the end address (`index_to_pc[inst_list.size()] = pc;`)

Sailanarmo:
> @Sailanarmo The get_index is trying to find the nearest instruction, it's not exactly a specific value. If you check the implementation, it uses `lower_bound`, not an index. The resize has the `+ 1` because at the end of `EmitLabels`, the last value of index_to_pc points to the end address (`index_to_pc[inst_list.size()] = pc;`)

@viniciuslrangel Gotcha! Thanks for the explanation. Maybe `getindex` should be changed to `getinstruction` then? If that is truly what it is doing then I feel renaming the lambda function to `getinstruction` would be more appropriate since that is what led to my confusion in assigning the index's of the loop instead. 

I still feel changing `labels` to a map would be better since you could iterate over all the keys and return the instructions instead of looking them up with `getindex`. 

One more question I have for you: Why is there a -1 at line 106? 
```cpp
  const size_t end_index = get_index(end) - 1;
```
I'm trying to understand why that was needed.

raphaelthegreat:
> I feel renaming the lambda function to getinstruction would be more appropriate since that is what led to my confusion in assigning the index's of the loop instead.

The lambda returns an index and not instruction itself. The name might be a little vague but being too verbose here might make code less readable

> I still feel changing labels to a map would be better since you could iterate over all the keys and return the instructions instead of looking them up with getindex.

index_to_pc is meant as a bi-directional translation between an address and its corresponding index in the list of instructions. index -> address conversion can be done with a simple indexing of the array, while address -> index is with lower_bound(). So this is faster in all cases than using std::map since array gives data locality.

> One more question I have for you: Why is there a -1 at line 106?

The "end" label is the address where the next basic block starts. So to get the index of the instruction where this block ends we subtract 1

Sailanarmo:
@raphaelthegreat Thank you for your help! It's greatly appreciated. I have made new updates. Please review and leave feedback.

Exhigh:
I think your coming at this from the wrong perspective, the initial assumption that the issue arises from within the initial loop iterating beyond labels size is just wrong, the issue comes from get_index, when the labels value address isn't present in index_to_pc which ends up causing the out of bounds access. I would suggest you do local testing to figure out why the issue is happening and try to come up with a solution and test it locally rather than theory guessing what's wrong and submitting a patch. This patch from the beginning should have been a discussion rather than a pull request, because it tries to fix something that isn't the issue.

PR #453, opened 2024-08-16
gpu: check right register for primitive restart index

viniciuslrangel:
@reg-prig suggestion

Not sure if it is possible to have trash in the register (maybe if primitive restart is toggled off), but just to make sure.

Related to #438


src/video_core/renderer_vulkan/vk_graphics_pipeline.h
```diff
@@ -39,7 +39,8 @@ struct GraphicsPipelineKey {
     Liverpool::StencilRefMask stencil_ref_front;
     Liverpool::StencilRefMask stencil_ref_back;
     Liverpool::PrimitiveType prim_type;
-    u32 prim_restart_index;
+    bool enable_primitive_restart;
```
raphaelthegreat: This will cause padding issues in the structure, better make it a u32

viniciuslrangel: Oh, didn't know it was relevant here, thx

red-prig:
```
	union VGT_MULTI_PRIM_IB_RESET_EN {
	struct {
		unsigned int                        RESET_EN : 1;
		unsigned int                                 : 31;
	} bitfields, bits;
```
The register is 32 bits in size but only the first bit needs to be checked.

PR #456, opened 2024-08-16
Improve posix_sem functions

StevenMiller123:
Some games are currently crashing due to the return of the posix_sem functions. These functions return -1 when an error occurs, and 0 when there's no error. When -1 is returned, games call the __Error function in libkernel to check the error. Since the posix_sem functions do not update the g_posix_errno variable appropriately, games see an unexpected error and can crash. 

The specific error I ran into was posix_sem_timedwait returning -1 when the error was an ETIMEDOUT. The games normally wouldn't crash when this occurs, but they were because the __Error function was returning 0 instead.

I've attempted to cover all possible errors here, however I based the errors off the ones set by the winpthreads library, so there may be some missing ones compared to what the PS4 could set, or what Linux/Mac OS may set. 

This fixes the current crash in Grand Theft Auto V and Assassin's Creed Unity, and doesn't appear to cause regressions in the games I tested.

### DIFF
--- a/src/core/libraries/kernel/libkernel.cpp
+++ b/src/core/libraries/kernel/libkernel.cpp
@@ -126,7 +126,7 @@ int ErrnoToSceKernelError(int e) {
 }
 
 void SetPosixErrno(int e) {
-    //Some error numbers are different between supported OSes or the PS4
+    // Some error numbers are different between supported OSes or the PS4
     switch (e) {
     case EPERM:
         g_posix_errno = POSIX_EPERM;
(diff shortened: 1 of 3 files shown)


georgemoralis:
An SetPosixErrno function that just set errno should be easier and need less code . Converting to sce error code and back is not that optimal

StevenMiller123:
I've added the suggested SetPosixErrno function, which appropriately converts the errno value to it's posix equivalent, then sets g_posix_errno appropriately. Let me know if there are any more changes I should make.

georgemoralis:
yes all good now

PR #464, opened 2024-08-18
Fix stencil buffer not being used

roamic:


raphaelthegreat (approved):
The hardest problems often have the stupidly simple solutions

raphipod:
I'm sorry to intervene, but #464 actually fails to build on Arch Linux. 

[Here is the link to the cmake output. (Pastebin)](https://pastebin.com/nhd1Kx7U)

PR #465, opened 2024-08-18
Some bloodborne hacks

raphaelthegreat:
Mostly a few hacks to allow retail game to boot. Not mergable but we can have this around for people to test while features are getting properly implemented in main

### DIFF
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -1070,7 +1070,16 @@ ScePthread PThreadPool::Create() {
         }
     }
 
+#ifdef _WIN64
     auto* ret = new PthreadInternal{};
+#else
+    // TODO: Linux specific hack
+    static u8* hint_address = reinterpret_cast<u8*>(0x7FFFFC000ULL);
+    auto* ret = reinterpret_cast<PthreadInternal*>(
+        mmap(hint_address, sizeof(PthreadInternal), PROT_READ | PROT_WRITE,
+             MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0));
+    hint_address += Common::AlignUp(sizeof(PthreadInternal), 4_KB);
+#endif
     ret->is_free = false;
     ret->is_detached = false;
     ret->is_almost_done = false;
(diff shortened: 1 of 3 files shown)


KioryFreeman:
Amazing update on this one, FPS seems really consistent now and definitely steps into playable territory.

noxifoxi:
At this point we should call the emulator **chadPS4**.

chadps4
*Sorry, but I couldn't hold it.*

MaxMo20:
Question, are these hack temporary? I'm assuming they are but i don't like to assume anything that I'm unsure about. I didn't know there was motion blur in the first place, must be per object motion blur which I like and smoothens the objects in motion without too much blur.  Maybe when this is playable could we configure the motion blur amounts since we have the values or any of these hacks individually? Thanks!

HermitsA:
OS: Nixos unstable (24.11)
GPU: RX 6700xt
CPU: Ryzen 5 7600
RAM: 32GB DDR5 6000 Mhz

It seems that the latest appimages for bb-hacks linux-qt has had a few regressions making the game unable to get to the menu.

Last seen working:
https://github.com/shadps4-emu/shadPS4/actions/runs/10458195832

Regressed at: https://github.com/shadps4-emu/shadPS4/actions/runs/10524940281
Also seen in: https://github.com/shadps4-emu/shadPS4/actions/runs/10526470363

[shad_log.txt](https://github.com/user-attachments/files/16736577/shad_log.txt)
360984682-57478af3-d7aa-4571-8f6e-10fbb4056f1a

rafael-57:
Latest has some regressions, the code for skipping these...shaders? is gone so crash on death happens again:
image

Tested by manually putting this piece of code back and crash on death doesn't happen anymore.

rafael-57:
Hunter's Dream actually works now!! Game is unmodded!

New game -> Die to the wolf -> Hunter's Dream!

image
[shad_log.txt](https://github.com/user-attachments/files/16803017/shad_log.txt)

Before this, trying to load a save crashed 2 the emu two times in a row though, not sure why. Haven't been able to replicate for now. Will provide logs if it happens again.

rafael-57:
Lightning is still kinda broken (this is an interior scene)

image

TedMosly:
infinite health cheat doesn't work, only stamina

echolessowl:
tried the linux-qt version, was able to load character creator but immediately crashed after 'finalizing contract' with a `[Debug] <Critical> page_manager.cpp:GuestFaultSignalHandler:207: Unreachable code!` error

raphaelthegreat:
Very few hacks are needed anymore before game boots, so this is not worth it to maintain anymore. Closing as main will soon be able to boot without any hacks

PR #466, opened 2024-08-18
Alphabetical order console language

DanielSvoboda:
image


georgemoralis:
nice , although i think it would be nicer to have something like this
https://doc.qt.io/qt-6/qcompleter.html

PR #475, opened 2024-08-19
PlayGo: basic implementation, credits to red-prig

bax-cz:
Some more PlayGo implementation inspired by red-prig.

### DIFF
--- a/src/core/file_format/playgo_chunk.h
+++ b/src/core/file_format/playgo_chunk.h
@@ -3,29 +3,128 @@
 
 #pragma once
 #include <filesystem>
-#include "common/types.h"
+#include <mutex>
+#include "common/io_file.h"
+#include "core/libraries/playgo/playgo_types.h"
+
+constexpr u32 PLAYGO_MAGIC = 0x6F676C70;
+
+struct chunk_t {
+    u32 offset;
+    u32 length;
+} __attribute__((packed));
 
 struct PlaygoHeader {
     u32 magic;
 
     u16 version_major;
     u16 version_minor;
-    u16 image_count;
+    u16 image_count;    // [0;1]
+    u16 chunk_count;    // [0;1000]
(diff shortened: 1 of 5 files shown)


georgemoralis:
looking good any specific game tested? i think infamous first (or second) needed that . Worms Also needed this but was working with the previous implementation

Graubek:
It fixed  [Lib.PlayGo] <Error> playgo.cpp:scePlayGoGetLocus:53: (STUBBED)called handle = 1, chunkIds = 2, numberOfEntries = 69  loop in InFamous second son and first light

bax-cz:
Yep, as 

> It fixed [Lib.PlayGo] playgo.cpp:scePlayGoGetLocus:53: (STUBBED)called handle = 1, chunkIds = 2, numberOfEntries = 69 loop in InFamous second son and first light

Yes, this. But still won't boot :(

src/core/file_format/playgo_chunk.h
```diff
@@ -3,29 +3,128 @@
 
 #pragma once
 #include <filesystem>
-#include "common/types.h"
+#include <mutex>
```
georgemoralis, line 6: probably needs an #include <vector>  to be able to compile under linux as well

bax-cz: ok, I'm in the process of setting up a linux build environment, this might take a while..

georgemoralis: No needed just add an #include <vector> and it should work

georgemoralis: vector

bax-cz: ok, should be fixed now

george24f:
I m a bit stuck on this what can i do to fix my infamous second son

PR #477, opened 2024-08-19
Fix gcc compilation error in vk_graphics_pipeline.cpp

Random06457:
The main branch currently doesn't compile with GCC.

GCC fails to infer the type of the two parts of a ternary expression whose types are different but both contain an implicit cast operator to the same type.

See example here: https://godbolt.org/z/Y5KdM77EW

Maybe adding a GCC pipeline to the CI could be a good idea for GCC/Clang differences(?)


georgemoralis:
We are using clang so if you want to pr a gcc ci feel free to do it

PR #481, opened 2024-08-19
Set game window size based on the config (in windowed mode)

kotn3l:
Changing the width and height in the emulator config didn't affect the game window size (in non-fullscreen mode). This little PR changes that. 

I don't actually know what resolution the games render at, so this might be an issue -- please let me know below. (I guess the games have the info on what resolution to render at for both PS4 and Pro?)

A possible next step would be to make the window properly resizeable, and then just do a basic scaling on the framebuffer based on it, while maintaining aspect ratio. 

EDIT: same for fullscreen mode -- for example, I have an ultrawide monitor, and usiong fullscreen mode just stretches the framebuffer.

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -34,9 +34,6 @@ Frontend::WindowSDL* g_window = nullptr;
 
 namespace Core {
 
-static constexpr s32 WindowWidth = 1280;
-static constexpr s32 WindowHeight = 720;
-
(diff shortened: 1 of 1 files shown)


georgemoralis:
looks ok however you need to fix clang format

kotn3l:
Ok sorry I think I understand the formatting now lol, should be good now.

PR #485, opened 2024-08-20
rtc errors

lzardy:


georgemoralis:
Someone forgot to add it to cmakelist?

lzardy:
> Someone forgot to add it to cmakelist?

Ok it is good now for merge, and I added the missing Ngs2 lib files to CMake as well.

PR #497, opened 2024-08-20
shader_recompiler: fix BranchTarget sign flip for sopp.simm

xezrunner:
This used to cause a crash that would prevent Amplitude [CUSA02480] from booting beyond initialization.

A conditional true label would get an address starting with `0xffff....`, which wasn't realistic with the given shader.  
With some addresses, the multiplication by 4 causes the value to have its MSB wrongly set due to the small `s16` type - updated it to `s32`.


xezrunner:
Does this change require additional info/changes/testing?

Issue #369, opened 2024-08-05
Memory-mapped file bug on Windows


Borchev:
Memory mapping of read-only files fails on windows. This is because MapViewOfFile requires the view size to be a multiple of the system page size, but the size of the files are not necessarily multiples of the system page size. Additionally, CreateFileMapping cannot create mappings larger than the size of a file on disk when they are read-only (correct me if I'm wrong here) so padding upwards this way won't work.

One potential solution:
VirtualAlloc an area large enough for the mapped file (rounded up to the next page size multiple), ReadFile the file contents into the alloc'd area and then set the pages to readonly.

I've tested the aforementioned solution with Cuphead, it fixes the crash-on-boot and gets sit further in the boot process, getting stuck with a repeating "Forcing submitdone to avoid..." message.

PR #502, opened 2024-08-20
Workaround for readonly memory mapping of files issue

Borchev:
Potential solution/workaround for #369. Fixes Cuphead crash on boot.


StevenMiller123:
Seems to fix the current issue seen in Cooking Mama: Cookstar too. https://github.com/shadps4-emu/shadps4-game-compatibility/issues/40

PR #504, opened 2024-08-21
bump submodules

shubhisroking:


Xphalnos:
I don't think it's necessary to do this because it was already done a short time ago in [**this commit**](https://github.com/shadps4-emu/shadPS4/commit/70576035b04051f04d832c2d5ff160ba6ff17442).

shubhisroking:
> I don't think it's necessary to do this because it was already done a short time ago in [**this commit**](https://github.com/shadps4-emu/shadPS4/commit/70576035b04051f04d832c2d5ff160ba6ff17442).

latest commits in GSlang and TOML removed some unused headers, but I get your point.

PR #507, opened 2024-08-21
sceNetAccept sleep and SDLAudio::AudioOutOutput latency and cpu improvements

kh0nsu:
The sceNetAccept sleep reduces CPU usage, possibly memory usage depending on game behaviour, and log spam if the game calls it in a loop. BB at least is known to do this. This is still far from ideal but I expect this is beneficial for most games.

The SDLAudio::AudioOutOutput changes will target a particular latency rather than a fixed number of bytes which should make behaviour consistent across games. Handle is validated further. A NULL buffer is supported (I saw this in some OpenOrbis samples). The sleep is changed from 0ms to 1ms. In testing an audio-only homebrew sample, CPU usage is lowered. It's hard to measure exactly (on Windows at least) but a sleep-0 theoretically constantly yields and re-schedules the thread possibly using up a whole core. Sleep-1 is not guaranteed to be 1ms but it should be lower than the (arbitrary) 20ms latency target on all platforms.

20ms is meant to provide ~1-frame latency in 60fps games but may be too low in some cases so this may need tuning. This should be obvious as audio will become crunchy in games where it is currently smooth.

Feedback welcome. This is my first PR.

### DIFF
--- a/src/audio_core/sdl_audio.cpp
+++ b/src/audio_core/sdl_audio.cpp
@@ -84,7 +84,7 @@ int SDLAudio::AudioOutOpen(int type, u32 samples_num, u32 freq,
             port.stream =
                 SDL_OpenAudioDeviceStream(SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK, &fmt, NULL, NULL);
             SDL_ResumeAudioDevice(SDL_GetAudioStreamDevice(port.stream));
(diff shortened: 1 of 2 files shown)


georgemoralis:
can someone test games with sound issues to see if it behaves better (lagging , cropping issues)

PR #514, opened 2024-08-21
Fix a few issues with the intel anv vulkan driver from mesa

Random06457:
- intel doesn't support `VK_FORMAT_D16_UNORM_S8_UINT`, so use `VK_FORMAT_D24_UNORM_S8_UINT` instead.
- the drivers panics when a texture has `usage & VK_IMAGE_USAGE_STORAGE_BIT && aspect == VK_IMAGE_ASPECT_DEPTH_BIT` (https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/intel/vulkan/anv_image.c#L760)
- `TextureCache::FindDepthTarget` wasn't setting the `image.aspect_mask` of the depth image properly so it still had its default value `VK_IMAGE_ASPECT_COLOR_BIT` which causes this assert to trigger: https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/vulkan/runtime/vk_image.c#L279

### DIFF
--- a/src/video_core/renderer_vulkan/liverpool_to_vk.cpp
+++ b/src/video_core/renderer_vulkan/liverpool_to_vk.cpp
@@ -304,6 +304,7 @@ std::span<const vk::Format> GetAllFormats() {
         vk::Format::eBc7UnormBlock,
         vk::Format::eD16Unorm,
         vk::Format::eD16UnormS8Uint,
+        vk::Format::eD24UnormS8Uint,
         vk::Format::eD32Sfloat,
         vk::Format::eD32SfloatS8Uint,
         vk::Format::eR4G4B4A4UnormPack16,
--- a/src/video_core/renderer_vulkan/vk_instance.cpp
+++ b/src/video_core/renderer_vulkan/vk_instance.cpp
@@ -481,6 +481,8 @@ bool Instance::IsFormatSupported(const vk::Format format) const {
 vk::Format Instance::GetAlternativeFormat(const vk::Format format) const {
     if (format == vk::Format::eB5G6R5UnormPack16) {
         return vk::Format::eR5G6B5UnormPack16;
+    } else if (format == vk::Format::eD16UnormS8Uint) {
+        return vk::Format::eD24UnormS8Uint;
     }
     return format;
 }
--- a/src/video_core/texture_cache/image.cpp
+++ b/src/video_core/texture_cache/image.cpp
@@ -73,14 +73,15 @@ static vk::ImageUsageFlags ImageUsageFlags(const ImageInfo& info) {
         if (!info.IsBlockCoded() && !info.IsPacked()) {
             usage |= vk::ImageUsageFlagBits::eColorAttachment;
         }
+
+        // In cases where an image is created as a render/depth target and cleared with compute,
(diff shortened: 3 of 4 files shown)


LeDragoX:
I could progress a little further after skipping this Assert in CUSA14876:

https://github.com/shadps4-emu/shadPS4/blob/79680c50c085f45bd5a4ce01f52565d9a2c175d8/src/core/libraries/kernel/memory_management.cpp#L287

Would give this log: [CrashTR_normal_shad_log.txt](https://github.com/user-attachments/files/16699988/CrashTR_normal_shad_log.txt)

But adding this hack + the PR where should stop at `PM4 type 2` will throw this new error instead, hope this helps.
```ts
[Debug] <Critical> image_info.cpp:ConvertImageType:43: Unreachable code!
```
New log: [CrashTR_pr514_shad_log.txt](https://github.com/user-attachments/files/16698894/CrashTR_pr514_shad_log.txt)

src/video_core/amdgpu/liverpool.cpp
```diff
@@ -175,6 +175,11 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
     while (!dcb.empty()) {
         const auto* header = reinterpret_cast<const PM4Header*>(dcb.data());
         const u32 type = header->type;
+        if (type == 2) {
```
psucien: I doubt it is a real type2 packet here. To convince me in opposite, show pm4 stream dump or better gnm function which writes it. We haven't encountered any packets other than type3 so far, and I don't think your case is special. This may be the result of command list corruption, or something else bad is happening.

Random06457: I had the same doubt but the overall command list looked legit to me for the following reasons:
- Beside that one type-2 packet, every other command list that was submitted has no incoherent data
- The AMD driver in mesa does emit some type-2 packets (look for string `PKT2_NOP_PAD`)
- The type-2 packet itself is just `0x80000000`. So it doesn't look like uninitialized data or whatever

However I didn't go much further to check if this is indeed what the game code does. I could do that but I don't have enough experience with the PS4-specific driver to easily locate the functions that assemble the command list throughout all the code without spending too much time on it. If you happen to be knowledgeable about it, I'd appreciate some pointers.

You can find the command list here as well as a disassembled version:
[00000000_dcb_1_0.zip](https://github.com/user-attachments/files/16736210/00000000_dcb_1_0.zip)

psucien: To me, this cmd list looks like a part of something bigger, so I assume that type2 is actually the last dword of the previous type3 packet. Dst address of write data also looks suspicious, but it is another thing.
I would recommend excluding this change from the PR, so we will be able to merge it. This can be addressed as a separate fix, however more investigations are needed. Also, please specify what game you're testing, as it will be helpful in tracking such issues.

Random06457: Sure, I'll move the type-2 specific changes to another PR and will add more info there.

Random06457: Done.

PR #517, opened 2024-08-21
Misc fixes

roamic:
* Fixed hanging subtitles after video has finished playing by reporting 0 in `sceAvPlayerCurrentTime`
* Fixed a crash in `sceMsgDialogOpen` (used in Resident Evil 4)
* Fixed Depth/Stencil usage. Fixes most of the validation errors that mention depth/stencil buffer usage
* Set W/H of `renderArea` to `1` to suppress validation errors for draw calls that do not use any render targets.


georgemoralis:
a lot of changes

PR #522, opened 2024-08-22
Optimize initialize UserPaths reserve count known paths

GermanAizek:
I think it will be useful for faster init emulator

### DIFF
--- a/src/common/path_util.cpp
+++ b/src/common/path_util.cpp
@@ -93,6 +93,7 @@ static auto UserPaths = [] {
         paths.insert_or_assign(shad_path, new_path);
     };
 
+    paths.reserve(11);
     create_path(PathType::UserDir, user_dir);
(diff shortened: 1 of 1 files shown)


georgemoralis:
Probably faster but I think someone can easy forget to update it when adding new folder so it might be risky

GermanAizek:
> Probably faster but I think someone can easy forget to update it when adding new folder so it might be risky

@georgemoralis is there way to make unordered_map initially already initialized? For example, like consteval or constexpr, such solution is better than reserve. Alternatively, I can get count iterations through for loop and reserve it. Tell me what is the best way to do this.

raphaelthegreat:
I don't think this is needed. Calling reserve is nice but when you are adding it to these initialization paths it doesn't benefit much and just adds more lines of code

georgemoralis:
close as not needed

PR #526, opened 2024-08-22
L2/R2 input threshold adjustment

kh0nsu:
- Input threshold changed to ~~150 (pressed) 135 (released)~~ >31 <16
- Fixes L2/R2 sticking where the rest point is ~~127~~ slightly above 0
- Avoids glitches when only partially pressed or drifting
- Still compatible with 0/255 inputs from keyboard (and now controller obviously)
- Tested with xbox controller and keyboard
- TRACE logging to help with threshold tuning; shouldn't compile in on release builds

### DIFF
--- a/src/input/controller.cpp
+++ b/src/input/controller.cpp
@@ -5,6 +5,7 @@
 #include "core/libraries/kernel/time_management.h"
 #include "core/libraries/pad/pad.h"
 #include "input/controller.h"
+#include "common/logging/log.h"
 
 namespace Input {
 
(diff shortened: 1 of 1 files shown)


georgemoralis:
Looking nice and someone test if that solves the l2/r2 issues?

SuperPlasmaPanda:
Doesn't seem to work in DSR
edit: on Dualsense

efigr:
Doesn't seem to make any difference on RE4 (2005), and Dark Souls Remastered.
I've also made a custom build to test it on Bloodborne, no difference there either.

Using an Xbox Series controller, if that matters.

bigol83:
> Doesn't seem to make any difference on RE4 (2005), and Dark Souls Remastered. I've also made a custom build to test it on Bloodborne, no difference there either.
> 
> Using an Xbox Series controller, if that matters.

Same for me with DS4, doesn't fix the issue in Dark Souls remastered

kh0nsu:
Very curious what's going on if this still doesn't fix it for everyone. Try enabling the logging (maybe change them to LOG_DEBUG or higher so you don't need all the trace stuff compiled in). That should show the range of values then; if it's not going above 150 or below 135 then the thresholds need adjusting. If the messages stop appearing after the first press or if it varies between games then something else is going on.

efigr:
> Very curious what's going on if this still doesn't fix it for everyone. Try enabling the logging (maybe change them to LOG_DEBUG or higher so you don't need all the trace stuff compiled in). That should show the range of values then; if it's not going above 150 or below 135 then the thresholds need adjusting. If the messages stop appearing after the first press or if it varies between games then something else is going on.

Gonna look into it.

georgemoralis:
#531 solves l2/r2 issue . Since I don't have a gamepad I dunno if that pr is needed now

kh0nsu:
> #531 solves l2/r2 issue . Since I don't have a gamepad I dunno if that pr is needed now

Oh, great! So the axis gets 0 to 255 then? I still recommend a non-zero threshold for the digital buttons in case there are games using that (it seems the above games prefer the analog value?) to deal with axis drift.

kh0nsu:
Rebased to after #531, updated and tested the thresholds with the corrected scaling.

Hermiten:
Can someone test it ? :)

rafael-57:
> Can someone test it ? :)
@Hermiten can you pls bring your branch up to date with main? I'll test

khmin84 (approved):

khmin84 (approved):

georgemoralis:
closed due to inactivity

PR #531, opened 2024-08-22
R2/L2 fix

georgemoralis:
might fix l2/r2 issue. waiting for feedback


efigr:
Fixes the L2 / R2 sticking bug in the games I've tested:
- Resident Evil 4 (2005)
- Bloodborne
- Dark Souls Remastered

PR #533, opened 2024-08-22
core, video_core: Use reserve() before loops where possible

GermanAizek:
Hi @georgemoralis these are the more useful reserves. Im waiting code review.


psucien:
This is not a hot code anyhow, I don't see the reason

PR #538, opened 2024-08-23
shader_recompiler: handle fetch shader address offsets

viniciuslrangel:
parse vertex & instance offset sgpr from fetch shader and propagate them to vkCmdDraw

still need to handle inputs with OverStepRate

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -51,24 +51,17 @@ void Rasterizer::Draw(bool is_indexed, u32 index_offset) {
     BeginRendering();
     UpdateDynamicState(*pipeline);
 
-    u32 instance_offset = 0;
-    if (vs_info.fetch_offset_sgpr != -1) {
-        instance_offset = vs_info.user_data[vs_info.fetch_offset_sgpr];
-    }
+    const auto [vertex_offset, instance_offset] = vs_info.GetDrawOffsets();
 
     if (is_indexed) {
-        u32 vertex_offset = 0;
-        if (vs_info.fetch_index_sgpr != -1) {
-            vertex_offset = vs_info.user_data[vs_info.fetch_index_sgpr];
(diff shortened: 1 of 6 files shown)


src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -51,13 +51,24 @@ void Rasterizer::Draw(bool is_indexed, u32 index_offset) {
...
+        }
+
+        cmdbuf.drawIndexed(num_indices, regs.num_instances.NumInstances(), 0, s32(vertex_offset),
+                           instance_offset);
     } else {
         const u32 num_vertices = regs.primitive_type == AmdGpu::Liverpool::PrimitiveType::RectList
                                      ? 4
                                      : regs.num_indices;
-        cmdbuf.draw(num_vertices, regs.num_instances.NumInstances(), 0, 0);
+        cmdbuf.draw(num_vertices, regs.num_instances.NumInstances(), 0, instance_offset);
```
viniciuslrangel: Could you dear reviewer double check if this instance offset makes sense?

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -51,13 +51,24 @@ void Rasterizer::Draw(bool is_indexed, u32 index_offset) {
     BeginRendering();
     UpdateDynamicState(*pipeline);
 
+    u32 instance_offset = 0;
```
raphaelthegreat: Can you add a method function to Shader::Info that returns this and vertex offset? Less clutter in this function I think

src/video_core/renderer_vulkan/vk_rasterizer.cpp
raphaelthegreat: I think it makes sense. Vertex offset also applies to non-indexed draws so you should add it here too

raphaelthegreat (approved):

PR #551, opened 2024-08-24
video_core: re-enable culling

viniciuslrangel:
Removes some geometric artifacts

(at least on project diva)


bigol83:
This breaks Dark Souls remastered main menu. The text "Dark Souls Remastered" appears seconds after the main menu options, At first it looks almost completely black.

Cattura

PR #556, opened 2024-08-24
Handle PM4 type-2 packets

Random06457:


Random06457:
https://github.com/shadps4-emu/shadPS4/pull/514#discussion_r1729935818
@psucien 

> To me, this cmd list looks like a part of something bigger, so I assume that type2 is actually the last dword of the previous type3 packet.

How do you think this is possible? From what I see from the `sceGnmSubmitCommandBuffers` calls, multiple command lists are sent, including the one I sent and the previous dcb command list seemed to end properly.

Though I am basing this on the assumption that the CPU "emulation" is 100% correct, which I would believe to be a safe assumption unless there's a weird edge case with an instruction that doesn't behave the same.

Otherwise I guess maybe one of the HLE calls could be incorrect and cause such a difference? But again, without proper knowledge of the GL driver used by the game, it's hard to tell.

> Dst address of write data also looks suspicious, but it is another thing.

Why so? Looks like a perfectly valid CPU address for a PIE binary to me.

The game I was testing this with was Deponia (CUSA03770).

Here is the total list of the sent command lists and their disassembly up to the one containing the type-2 packet. Please don't trust the disassembler too much as it is still WIP.
[pm4.zip](https://github.com/user-attachments/files/16736649/pm4.zip)

psucien:
> How do you think this is possible?

We observed cases, when due to race condition on CPU side (bugs in sync primitives implementation or timings being different to ones on hardware) a command list was overwritten during processing by GPU thread. This caused parser crashes on unexpected packet types.

> The game I was testing this with was Deponia (CUSA03770).

A quick RE of the game showed that it indeed uses type2 packets for paddings insertion. This is a kind of evidence I wanted to see before giving a green light for the PR.

Before the merge, I would ask you to:
- make packet type check to use `switch` operator
- remove the commentary for type3 case as it doesn't make sense now
- change your commentary so it will reflect the actual meaning of the packet, rather than "don't do anything". You can refer to the fact that it is emitted by Gnm for cases of 1dw NOP paddings.

Random06457:
Done.

PR #560, opened 2024-08-24
Fixed video dimensions alignment and image cache

roamic:
AvPlayer:
The library should provide raw frames and ask user to remove the unused parts with crop offsets.
Since this library supports only AVX (H.264) video we may safely assume that frame buffer dimensions are aligned to 16. Unfortunately, I found no way to get raw frame buffers from ffmpeg, so the data is positioned in the memory manually.

Image cache:
The image should be erased only in `DeleteImage` instead of `UnregisterImage`.


StevenMiller123:
Image cache fix seems to resolve the Clicker Heroes crash reported [here](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/39).

Issue #496, opened 2024-08-20
[Aggregate issue] Missing opcodes


rafael-57:
## Format ##
-- {Game Serial} {Game Name}
{Some log lines with "Unknown opcode"}
{upload full logfile}

## Currently implemented ##
- [x] BUFFER_ATOMIC_SWAP #566 
- [x] V_LSHL_B64 opcode #608
- [x] V_BFM_B32 and V_FFBH_U32 #663 
- [x] PM4 0x16 and 0x11 #637 
- [x] PM4 0x24 and 0x25 #678 
- [x] PM4 0x42 #426 
- [x] TBUFFER_STORE_FORMAT_XYZW #802 
- [x] IMAGE_SAMPLE_CD and IMAGE_GATHER4_C_LZ #802
- [x] S_ANDN2_B32 and S_NAND_B32 #833
- [x] S_XOR_B32 #911 
- [x] V_MAX_F64 #915 
- [x] DS_READ2ST64_B32  #916 
- [x] V_MUL_F64, V_FMA_F64 and V_RCP_F64 #944 
- [x] V_CVT_PKNORM_U16_F32, V_CMPX_EQ_I32 #1013
- [x] V_CVT_PK_U8_F32 #1022 
- [x] V_CMPX_LE_I32 #1056 
- [x] V_CMP_CLASS_F32 #1153
- [x] V_CVT_F64_I32 #1219 
- [x] S_BCNT1_I32_B64, S_ORN2_SAVEEXEC_B64, V_RCP_CLAMP_F32, V_CMPX_GE_I32, V_CVT_PK_U16_U32, and PM4 0x50 DMA_DATA #1412
- [x] S_BITCMP0_B32, S_BITCMP1_B32, S_BITCMP0_B64, S_BITCMP1_B64 #1550
- [x] V_MED3_U32 #1553 
- [x] DS_OR_B32 #1590 
- [x] DS_AND_B32, DS_OR_B32, DS_XOR_B32 #1593
- [x] S_FF1_I32_B32 https://github.com/shadps4-emu/shadPS4/commit/cde84e4bac94cfaa4c63f38f2b0494b78c6955eb
- [x] S_ABS_I32 #1713
- [x] V_FREXP_EXP_I32_F64, V_FREXP_MANT_F64, V_FRACT_F64, V_FREXP_EXP_I32_F32, and V_FREXP_MANT_F32 #1766
- [x] IMAGE_STORE_MIP #1770
[...]

PR #557, opened 2024-08-24
core/kernel: Miscellaneous memory fixes

polybiusproxy:
This PR:
- Changes the direct memory size to be accurate (SCE_KERNEL_MAIN_DMEM_SIZE aka `sceKernelGetDirectMemorySize`).
- Fixes an inaccuracy where `sceKernelAvailableDirectMemorySize` also accounted for non-free direct memory areas.
- Fixed an inaccuracy where `sceKernelAvailableDirectMemorySize` did not return `ENOMEM`.

Fixes Rock Band 4 ([CUSA02084](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/144)) not booting on version 2.21.
Fixes DRAGON BALL FighterZ ([CUSA09072](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/43)) memory allocations (at @StevenMiller123).
Fixes JoJo's Bizarre Adventure: Eyes of Heaven ([CUSA04909](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/291)) not booting.

PR #566, opened 2024-08-24
shader_recompiler: BUFFER_ATOMIC_SWAP Opcode

lzardy:
Additions:
- Translation of `BUFFER_ATOMIC_SWAP` MUBUF instruction
- BufferAtomicSwap Emitter and IR Opcode
  - Replaces nonexistent BufferAtomicExchange32

Related to issue #496, ref: [BUFFER_ATOMIC_SWAP](https://github.com/shadps4-emu/shadPS4/issues/496#issuecomment-2308520893)


raphaelthegreat:
If you want to rename Exchange to Swap could you also add the 32 postfix so its consistent with other opcodes? Otherwise lgtm (is there any reason this is draft actually?)

lzardy:
> If you want to rename Exchange to Swap could you also add the 32 postfix so its consistent with other opcodes? Otherwise lgtm (is there any reason this is draft actually?)

Okay I have added 32 to end. It is draft as I was only able to display this missing Opcode with an unreal engine game, which needed [this PR](https://github.com/shadps4-emu/shadPS4/pull/557) to start.

PR #576, opened 2024-08-25
video_core: implement PM4 disassembler

Random06457:
This PR adds a simple PM4 disassembler for debugging purpose.

The disassembled PM4 display lists are saved to disk next to their binary version when `Config::dumpPM4()` is true.

Not all PM4 commands are supported yet but the ones supported by the emulator should disassemble fine.

### DIFF
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -4,6 +4,9 @@
 #include "gnm_error.h"
 #include "gnmdriver.h"
 
+#define MAGIC_ENUM_RANGE_MIN 0
+#define MAGIC_ENUM_RANGE_MAX 0x88
+#include <magic_enum.hpp>
 #include "common/assert.h"
 #include "common/config.h"
 #include "common/debug.h"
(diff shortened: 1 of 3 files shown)


psucien:
Thank you for your efforts, having a pm4 analyzer is indeed very helpful. At this moment we already have a tool for this purpose, and probably it will be good if you check it out: https://github.com/psucien/tlg-emu-tools
You can reach anyone from dev team in Discord if you are interested in its development.

src/core/libraries/gnmdriver/gnmdriver.cpp
```diff
@@ -320,6 +323,296 @@ static void WaitGpuIdle() {
...
+        const auto* header = reinterpret_cast<const PM4Header*>(cmds.data());
+        const u32 type = header->type;
+        switch (type) {
+        case 0:
+            append(fmt::format("PM4Type0: base={}, count={}\n", header->type0.base.Value(),
+                               header->type0.count.Value()));
+            for (size_t i = 0; i < header->type0.count.Value(); i++)
+                append(fmt::format("- 0x{:08X}\n", cmds.data()[1 + i]));
+            break;
+            cmds = cmds.subspan(header->type0.NumWords() + 1);
```
baggins183, line 351: break should be after this line

Random06457: Thanks

Hermiten:
What do you plan to do with this @psucien ?

psucien:
Unfortunatelly it won't go by at least three reasons:
- high chance of sync issues caused by heavy io in submission functions. These functions shouldn't be blocking and we already risking here by having binary pm4 dump;
- some games show about a megabyte binary dumps _per frame_. Imagine this being written as text and accumulated on disk for a some time;
- since we aready have ImGui integrated, we are going to move the existing parser as overlay into main emulator window.

psucien:
Done in scope of https://github.com/shadps4-emu/shadPS4/pull/1094

PR #577, opened 2024-08-25
Update README.md game showcase table

Ritax2003:


raphaelthegreat:
I don't think having the status next to the images is any useful, but it adds visual clutter. If a game is shown on the readme it can be assumed that it boots

Ritax2003:
> I don't think having the status next to the images is any useful, but it adds visual clutter. If a game is shown on the readme it can be assumed that it boots

yeah my mistake.
sorry.

PR #580, opened 2024-08-25
feat: add pl_PL localization

mavethee:
Initial polish localization for shadPS4 :3

PR #582, opened 2024-08-25
Polish translation

Centek96:
Made polish translation without specialized settings names.

### DIFF
--- a/src/qt_gui/translations/pl_PL.ts
+++ b/src/qt_gui/translations/pl_PL.ts
@@ -6,7 +6,7 @@
         <message>
             <location filename="../about_dialog.ui" line="16"/>
             <source>About shadPS4</source>
-            <translation>About shadPS4</translation>
+            <translation>O shadPS4</translation>
         </message>
         <message>
             <location filename="../about_dialog.ui" line="60"/>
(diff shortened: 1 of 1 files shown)


src/qt_gui/translations/pl_PL.ts
```diff
@@ -16,322 +16,322 @@
...
             <location filename="../about_dialog.ui" line="78"/>
             <source>shadPS4 is an experimental open-source emulator for the PlayStation 4.</source>
-            <translation>shadPS4 is an experimental open-source emulator for the PlayStation 4.</translation>
+            <translation>shadPS4 to eksperymentalny otwarto-źródłowy emulator PlayStation 4.</translation>
         </message>
         <message>
             <location filename="../about_dialog.ui" line="99"/>
             <source>This software should not be used to play games you have not legally obtained.</source>
-            <translation>This software should not be used to play games you have not legally obtained.</translation>
+            <translation>Nie powinien być używany w celu ogrywania nielegalnych kopii gier.</translation>
```
wojek07, line 24: Nit: Even thought entire sentence conveys appropriate message, I wonder whether used verb "ogrywać" is somewhat informal and as such not fits into vocabulary used in legal disclaimers?

src/qt_gui/translations/pl_PL.ts
```diff
@@ -16,322 +16,322 @@
...
             <location filename="../game_install_dialog.cpp" line="74"/>
             <source>Error</source>
-            <translation>Error</translation>
+            <translation>Błąd</translation>
         </message>
         <message>
             <location filename="../game_install_dialog.cpp" line="75"/>
             <source>The value for location to install games is not valid.</source>
-            <translation>The value for location to install games is not valid.</translation>
+            <translation>Ścieżka jest nieprawidłowa.</translation>
```
wojek07, line 78: I think the source states something about the path and that information is omitted in translation.

src/qt_gui/translations/pl_PL.ts
masermb: I'm backing it up as I had exactly the same feelings about it.

src/qt_gui/translations/pl_PL.ts
masermb: I agree that there is a difference to the original, though the english source doesn't sound right and chosen translation might be the improvement overall.  Probably original should be  "Location of games directory is invalid." or  "Path to games location folder is invalid." etc.

georgemoralis:
Already have #580 merged with this . Thanks for your efford ! If you have some instances with translation you can open a new pr

PR #594, opened 2024-08-26
Fix ja_JP.ts

nishinji:
Sorry


Xphalnos:
The error is `Encountered incorrectly encoded content.` isn't this an unsupported character?

PR #598, opened 2024-08-26
Dear ImGui Implementation

viniciuslrangel:
Press F10 to show FPS 😄 

- [x] Add core headers
- [x] Adapt SDL3 & Vulkan backend
- [x] Integrate to driver pipeline
- [x] Basic input handling (keyboard/mouse)
- [x] Gamepad capture
- [x] Layer abstraction
- [x] Basic video information layer (frame timing)


viniciuslrangel:
~Dear ImGui is not as a submodule yet because it needs patches to compile with C++23~

Forked and added as submodule [ext-imgui](https://github.com/shadps4-emu/ext-imgui)

.gitmodules
```diff
@@ -81,4 +81,9 @@
 [submodule "externals/ffmpeg-core"]
 	path = externals/ffmpeg-core
 	url = https://github.com/shadps4-emu/ext-ffmpeg-core.git
-	shallow = true
\ No newline at end of file
+	shallow = true
+[submodule "externals/dear_imgui"]
+	path = externals/dear_imgui
+	url = git@github.com:shadps4-emu/ext-imgui.git
```
sampletext32: all neighboring urls are `https`. 
```suggestion
	url = https://github.com/shadps4-emu/ext-imgui.git
```

.gitmodules
```diff
@@ -81,4 +81,9 @@
 [submodule "externals/ffmpeg-core"]
 	path = externals/ffmpeg-core
 	url = https://github.com/shadps4-emu/ext-ffmpeg-core.git
-	shallow = true
\ No newline at end of file
+	shallow = true
+[submodule "externals/dear_imgui"]
+	path = externals/dear_imgui
```
abouvier, line 89: Why not just `imgui`?

externals/CMakeLists.txt
```diff
@@ -149,6 +149,19 @@ if (APPLE AND NOT TARGET date::date-tz)
     add_subdirectory(date)
 endif()
 
+# Dear ImGui
+if (NOT TARGET Dear_ImGui)
```
abouvier: The target will never exists if there is no `find_package` ;-)

.gitmodules
viniciuslrangel: I'll forward the declaration later

externals/CMakeLists.txt
```diff
@@ -155,6 +155,19 @@ if (APPLE)
     endif()
 endif()
 
+# Dear ImGui
+block()
```
abouvier: [block()](https://cmake.org/cmake/help/latest/command/block.html) was added in CMake 3.25 but shadps4 uses `cmake_minimum_required(VERSION 3.16.3)`.

viniciuslrangel: Oh, thanks for the tip. Will fix later

gandalfthewhite19890404:
It is for QT builds only?

viniciuslrangel:
> It is for QT builds only?

No, both!

viniciuslrangel:
Removed fix commits & ensured stability in each

raphaelthegreat (approved):
Looks awesome, thanks a lot for this work!

Digote:
When I try to press here, this error occurs and the emulator closes.

image

viniciuslrangel:
> When I try to press here, this error occurs and the emulator closes.
> 
> image
> 

Can you create a issue with more details, please

PR #600, opened 2024-08-26
kernel: Reimplement POSIX semaphores using std::counting_semaphore.

squidbus:
Reimplements POSIX semaphores using `std::counting_semaphore` as a more cross-platform solution. Each function does error checking as defined by the POSIX spec.

For Windows this reduces reliance on winpthreads, and for macOS this fixes POSIX semaphores as apparently `sem_init`, `sem_destroy`, and `sem_getvalue` are not only marked as deprecated but also not implemented; only named POSIX semaphores are implemented.

One caveat is that `std::counting_semaphore` does not provide an equivalent for `sem_getvalue`. Instead we keep an atomic counter alongside the semaphore to track its value based on successful waits and posts.

Tested this on the Persona Dancing games and it resolves the issue I was having on macOS where the games would crash on launch. The invalid semaphores make the games' job threads think there is work to do as they do not properly check for errors, resulting in a jump to a null pointer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,10 +95,6 @@ check_symbol_exists(pthread_mutex_timedlock "pthread.h" HAVE_PTHREAD_MUTEX_TIMED
 if(HAVE_PTHREAD_MUTEX_TIMEDLOCK OR WIN32)
     add_compile_options(-DHAVE_PTHREAD_MUTEX_TIMEDLOCK)
 endif()
-check_symbol_exists(sem_timedwait "semaphore.h" HAVE_SEM_TIMEDWAIT)
-if(HAVE_SEM_TIMEDWAIT OR WIN32)
-    add_compile_options(-DHAVE_SEM_TIMEDWAIT)
-endif()
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
     # libc++ requires -fexperimental-library to enable std::jthread and std::stop_token support.
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include <mutex>
+#include <semaphore>
 #include <thread>
-#include <semaphore.h>
 #include "common/alignment.h"
 #include "common/assert.h"
 #include "common/error.h"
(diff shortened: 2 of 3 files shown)


StevenMiller123:
Breaks audio in Assassins Creed Unity (CUSA00663), and causes Grand Theft Auto V (CUSA00419) to hang on startup. Note that the memory error in Assassins Creed Unity's log is unrelated. These tests were performed on Windows, not sure if they behave differently on Mac OS or Linux.

[GTAV log.txt](https://github.com/user-attachments/files/16751143/GTAV.log.txt)
[ACU log.txt](https://github.com/user-attachments/files/16751144/ACU.log.txt)

mavethee:
Tested a few games, I hope it might be helpful, not much changed for the games I've own at least I can't see change in logs.

Hope it will help a bit 😅 

<details>

Beyond Two Souls_CUSA00513_v01 00

</details>

[Beyond Two Souls_CUSA00513_v01.00.txt](https://github.com/user-attachments/files/16752279/Beyond.Two.Souls_CUSA00513_v01.00.txt)

<details>

Chaos_on_Deponia_CUSA08542_v01 01

</details>

[Chaos_on_Deponia_CUSA08542_v01.01.txt](https://github.com/user-attachments/files/16752280/Chaos_on_Deponia_CUSA08542_v01.01.txt)

<details>

Deponia_CUSA03770_v1 00

</details>

[Deponia_CUSA03770_v1.00.txt](https://github.com/user-attachments/files/16752281/Deponia_CUSA03770_v1.00.txt)

<details>

FATAL FRAME Maiden of Black Water_CUSA28986_v01 04

</details>

[FATAL FRAME Maiden of Black Water_CUSA28986_v01.04.txt](https://github.com/user-attachments/files/16752282/FATAL.FRAME.Maiden.of.Black.Water_CUSA28986_v01.04.txt)

<details>

Life_is_Strange_Before_the_Storm_CUSA08487_v01 05

</details>

[Life_is_Strange_Before_the_Storm_CUSA08487_v01.05.txt](https://github.com/user-attachments/files/16752284/Life_is_Strange_Before_the_Storm_CUSA08487_v01.05.txt)

<details>

Minecraft_ CUSA00265_v02 94

</details>

[Minecraft_ CUSA00265_v02.94.txt](https://github.com/user-attachments/files/16752285/Minecraft_.CUSA00265_v02.94.txt)

<details>

Red Dead Redemption_CUSA36843_v01 04

</details>

[Red Dead Redemption_CUSA36843_v01.04.txt](https://github.com/user-attachments/files/16752286/Red.Dead.Redemption_CUSA36843_v01.04.txt)

<details>

VA-11_Hall-A_CUSA15402_v01 00

</details>

[VA-11_Hall-A_CUSA15402_v01.00.txt](https://github.com/user-attachments/files/16752287/VA-11_Hall-A_CUSA15402_v01.00.txt)

squidbus:
> Breaks audio in Assassins Creed Unity (CUSA00663), and causes Grand Theft Auto V (CUSA00419) to hang on startup. Note that the memory error in Assassins Creed Unity's log is unrelated. These tests were performed on Windows, not sure if they behave differently on Mac OS or Linux.

Can you check again, I've made some correctness fixes.

> Tested a few games, I hope it might be helpful, not much changed for the games I've own at least I can't see change in logs.

I'm not sure about some of these games but RDR is a shader compile issue, I'm planning to look into some graphics related issues with MoltenVK after this.

StevenMiller123:
Seems like your update fixes the issue I had in Grand Theft Auto V. I'm not on a setup that can verify the Assassins Creed Unity issue, but as both use the same sound system, it should be fixed too.

StevenMiller123:
I've now confirmed that the audio issue with Assassins Creed Unity is fixed.

squidbus:
Regressions reported are fixed so this should be ready for review.

raphaelthegreat:
I'm curious what was the cause of the regressions?

squidbus:
> I'm curious what was the cause of the regressions?

Implemented timedwait as if it were a duration instead of a fixed time point.

raphaelthegreat:
Ah nice. PR looks good to me as well

PR #613, opened 2024-08-27
shader_recompiler: constant propagation bitwise operations + S_CMPK_EQ_U32 fix

0xsegf4ult:
- Implemented BitwiseAnd64 IR opcode and added it with BitwiseOr64 to the constant propagation pass
- Fixed S_CMPK_EQ_U32 to use the correct operand

Currently S_CMPK_EQ_U32 uses src[0] which is not defined for a SOPK instruction and this code always fails an assertion as the operand field is left undefined.
```c++
const IR::U32 src0{GetSrc(inst.src[0])};
```
I moved this to inst.dst[0] as specified in the ISA documentation and in line with how the decoder parses SOPK instructions
> S_CMPK_EQ_U32: SCC = (D.u == SIMM16).
```cpp
void GcnDecodeContext::decodeInstructionSOPK(u32 hexInstruction) {
    u32 sdst = bit::extract(hexInstruction, 22, 16);
    u32 op = bit::extract(hexInstruction, 27, 23);
        
    m_instruction.opcode = static_cast<Opcode>(op + static_cast<u32>(OpcodeMap::OP_MAP_SOPK));
    
    m_instruction.dst[0].field = getOperandField(sdst);
    m_instruction.dst[0].code = sdst;
    m_instruction.dst_count = 1;
        
    m_instruction.control.sopk = *reinterpret_cast<InstControlSOPK*>(&hexInstruction);
}   
```

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
@@ -286,6 +286,7 @@ Id EmitShiftRightLogical64(EmitContext& ctx, Id base, Id shift);
 Id EmitShiftRightArithmetic32(EmitContext& ctx, Id base, Id shift);
 Id EmitShiftRightArithmetic64(EmitContext& ctx, Id base, Id shift);
 Id EmitBitwiseAnd32(EmitContext& ctx, IR::Inst* inst, Id a, Id b);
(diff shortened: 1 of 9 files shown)


0xsegf4ult:
Should be fine now, I recommend squashing commits since I just woke up was a bit clumsy with the rebase :)

PR #618, opened 2024-08-27
video_core: Added copyGPUCmdBuffers option

vertver:
Because CUSA01593 crashes due to command buffer reuse, the workaround in the command buffer copy is required to resolve [this](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/50) issue.


src/video_core/amdgpu/liverpool.cpp
```diff
@@ -568,9 +569,47 @@ Liverpool::Task Liverpool::ProcessCompute(std::span<const u32> acb, int vqid) {
...
 }
 
+void Liverpool::CopyCmdBuffers(std::span<const u32>& dcb, std::span<const u32>& ccb) {
+    auto& queue = mapped_queues[GfxQueueId];
+
+    // This is fine because resize doesn't reallocate the buffer on shrink
+    queue.dcb_buffer.resize(
+        std::max(queue.dcb_buffer.size(), queue.dcb_buffer_offset + dcb.size()));
+    queue.ccb_buffer.resize(
+        std::max(queue.ccb_buffer.size(), queue.ccb_buffer_offset + dcb.size()));
```
psucien: I believe it should be `ccb.size()`

src/video_core/amdgpu/liverpool.cpp
```diff
@@ -568,9 +569,47 @@ Liverpool::Task Liverpool::ProcessCompute(std::span<const u32> acb, int vqid) {
...
+    if (!dcb.empty()) {
+        std::memcpy(queue.dcb_buffer.data() + queue.dcb_buffer_offset, dcb.data(),
+                    dcb.size_bytes());
+        queue.dcb_buffer_offset += dcb.size();
+    }
+
+    if (!ccb.empty()) {
+        std::memcpy(queue.ccb_buffer.data() + queue.ccb_buffer_offset, ccb.data(),
+                    ccb.size_bytes());
+        queue.ccb_buffer_offset += dcb.size();
```
psucien: The same as above

Issue #619, opened 2024-08-27
Integrate CCache for increase repeat buildings on Linux and CI


GermanAizek:
More info: https://ccache.dev/

LeDragoX:
@GermanAizek I'm already working to fix that, expect it to come Soon™, literally with CCache. You'll be shocked for how fast it is.
Also, look at #542, they already talked about it.

GermanAizek:
@LeDragoX fixed it on cmakelists https://github.com/shadps4-emu/shadPS4/pull/621. Is not affected actions CI.

LeDragoX:
@GermanAizek I didn't knew how to make it work from CMakeLists, nice, all I could do was modifying the CI, does it benefit other OSes too?
If you want to look at my modifications, I got it here: https://github.com/LeDragoX/shadPS4/tree/cache-actions/.github/workflows

Hermiten:
Fixed with this PR https://github.com/shadps4-emu/shadPS4/pull/985

PR #621, opened 2024-08-27
cmake: integrate ccache for decrease repeat compiling time on Linux and BSD

GermanAizek:
Fixing this issue https://github.com/shadps4-emu/shadPS4/issues/619


abouvier:
And what if `ccache` is installed but somehow you don't want to use it? It's an user preference, you should pass these variables in the cmake invocation.

georgemoralis:
close cause of #622 merged

Issue #542, opened 2024-08-23
Added caching in actions


georgemoralis:
Especially windows builds are slow maybe caching can be added in ci

aidenzzz:
Maybe something like [this](https://github.com/saxbophone/CPP20-Cross-Platform-Template/blob/6991098e7f243ae9c6477c601953dd132f499e53/.github/workflows/continuous-integration.yml#L31-L62)?

georgemoralis:
That could probably work somehow yes

LeDragoX:
> Maybe something like [this](https://github.com/saxbophone/CPP20-Cross-Platform-Template/blob/6991098e7f243ae9c6477c601953dd132f499e53/.github/workflows/continuous-integration.yml#L31-L62)?

I tried using your example and reading through the updated docs https://github.com/actions/cache and
I've only managed to improve the "Configure CMake" time, maybe I've done something wrong.

| macOS-Qt | Linux-Qt | Windows-Qt |
| ------------- | ------------- | ------------- |
| ![image](https://github.com/user-attachments/assets/7234b714-7e0d-4368-9212-1e7754097b86) | ![image](https://github.com/user-attachments/assets/d0a2645f-9c86-48a0-b910-e05af720e37d) | ![image](https://github.com/user-attachments/assets/30d94459-4084-4b22-a972-e49425c5604f) |


So, I'm thinking about switching it to https://github.com/hendrikmuhs/ccache-action and see if it helps with the build time (hoping the hash keys will match)
Is this a great option?

Artalus:
#622 now speeds up Linux and Mac builds, but @LeDragoX mentioned that they could not set up `ccache` for Windows builds. Speaking from my experience maintaining my company's CI, `ccache` indeed does not behave when CMake is configured to use MSBuild\*. The alternative for me was to build our CMake projects using Ninja\*\* instead. Moreover, Ninja on its own is already faster than MSBuild even without compiler caching.

There is a caveat though: some directives in `CMakeLists` are handled differently by the two generators\*\*\*. So if only Ninja were to be used in CI, a subtle bug in `CMakeLists.txt` could evade the automated checks and create issues for Visual Studio users relying on the default generator

---

I ran some very basic experiments with Github Actions in my fork of the shadPS4 repo. Using `cmake -G Ninja` alone reduces time of `cmake --build` from ~20min to ~5min . Throwing `ccache` into the mix boosts this to ~2min.
[...]

PR #622, opened 2024-08-27
ci: enable cmake cache on all platforms (SDL/Qt)

LeDragoX:
This PR aims to improve build times in CI, Linux and macOS building steps have improved by a LOT!
Unfortunately, Windows still can't cache, even using sccache with MSVC.
- Partially solves (#542)

## Changes:
- [x] uses [actions/cache@v4](https://github.com/actions/cache) and [hendrikmuhs/ccache-action@v1.2.14](https://github.com/hendrikmuhs/ccache-action)
- [x] Keeps the cache from CMake Cache configuration (Windows, Linux, macOS).
- [x] Keeps the cache from CMake build objects (Linux/macOS).
- [x] Linux is using **ccache** from hendrikmuhs/ccache-action@v1.2.14.
- [x] macOS is using **sccache** from hendrikmuhs/ccache-action@v1.2.14.
- [x] Windows is caching the **CMake Configuration** step only, I've tried most stuff to force the use of caching, but without success, if anyone is willing to try, I've written some instructions below. ~I'm done with it~.

## Known Issues:
- ~Using same cache name/hash for each OS will create new bugs like macOS-sdl trying to build with Qt.~ (Fixed)
- The first CI run is not cached, the first time is when the cache is stored in [here](https://github.com/shadps4-emu/shadPS4/actions/caches).
- There is a limit for how many cache GitHub actions can store, see [here](https://github.com/actions/cache?tab=readme-ov-file#cache-limits).
- Using CCache with Windows didn't work as expected, but did not broke the actions.
  - The only way that _may_ work (needs testing and time) is by using MSYS2 MINGW64 and installing the packages:
  - `mingw-w64-x86_64-ccache` or `mingw-w64-x86_64-sccache`
  - It's worth noticing that MSYS2-compiled versions will only work under the MSYS2 shell, on native Windows will crash, even if you've copied the DLLs. (see action [here](https://github.com/LeDragoX/shadPS4/actions?query=branch%3Amsys2-ci))
- ~Using CCache with macOS will break the actions.~ Fixed by using sccache instead.

~There are some regression for some tests, but the most results are beneficial to CI building time.~
**Update:** the regressions for Windows and macOS were fixed, and brings about a minute less of compile time.

Comparisons:

|                       | Before Caching | After Caching |
| :--------------: | :--------------: | :--------------: |
|  Linux            |  firefox_2NvYAWlE2N 10m 13s | firefox_M6CKQAD16m 2m 10s |
|  Linux-Qt       | firefox_QmHnVkSlaQ 11m 27s | firefox_4HC8x1XkWl 3m 5s |
|  macOS          | firefox_0Zl7SOm1Uu 5m 39s | Captura de tela 2024-09-01 175335 2m 58s |
|  macOS-Qt     | firefox_v43iL0Odi9 6m 54s | Captura de tela 2024-09-01 175403 4m 8s |
|  Windows       | firefox_b615qmUwBJ 17m 50s | firefox_7Zdfq2vBa0 14m 39s |
|  Windows-Qt | firefox_DRsdumDujr 18m 15s | firefox_ugdmtlIfuw 20m 11s (Fixed) |

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -25,8 +25,27 @@ jobs:
       run: >
        sudo apt-get update && sudo apt install libx11-dev libxext-dev libwayland-dev libfuse2 clang build-essential
 
+    - name: Cache CMake dependency source code
+      uses: actions/cache@v4
+      env:
+          cache-name: ${{ runner.os }}-sdl-cache-cmake-dependency-sources
+      with:
+          path: | 
+            ${{github.workspace}}/build
+          key: ${{ env.cache-name }}-${{ hashFiles('**/CMakeLists.txt', 'cmake/**') }}
+          restore-keys: |
+            ${{ env.cache-name }}-
+
+    - name: Cache CMake dependency build objects
+      uses: hendrikmuhs/ccache-action@v1.2.14
+      env:
+          cache-name: ${{ runner.os }}-sdl-cache-cmake-dependency-builds
+      with:
+        append-timestamp: false
+        key: ${{ env.cache-name }}-${{ hashFiles('**/CMakeLists.txt', 'cmake/**') }}
+
     - name: Configure CMake
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
+      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
 
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --parallel
--- a/.github/workflows/macos-qt.yml
+++ b/.github/workflows/macos-qt.yml
@@ -40,8 +40,29 @@ jobs:
         arch: clang_64
         archives: qtbase qttools
 
+    - name: Cache CMake dependency source code
+      uses: actions/cache@v4
+      env:
+          cache-name: ${{ runner.os }}-qt-cache-cmake-dependency-sources
+      with:
+          path: | 
+            ${{github.workspace}}/build
+          key: ${{ env.cache-name }}-${{ hashFiles('**/CMakeLists.txt', 'cmake/**') }}
+          restore-keys: |
+            ${{ env.cache-name }}-
+
+    - name: Cache CMake dependency build objects
+      uses: hendrikmuhs/ccache-action@v1.2.14
+      env:
+          cache-name: ${{runner.os}}-qt-cache-cmake-dependency-builds
+      with:
+        append-timestamp: false
+        create-symlink: true
+        key: ${{env.cache-name}}-${{ hashFiles('**/CMakeLists.txt', 'cmake/**') }}
+        variant: sccache
+
     - name: Configure CMake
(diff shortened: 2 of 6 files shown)


DanielSvoboda:
half the time, nice

.github/workflows/macos-qt.yml
```diff
@@ -40,6 +40,28 @@ jobs:
...
+    - name: Cache CMake dependency build objects
+      uses: actions/cache@v4
+      env:
+        cache-name: cache-cmake-dependency-builds
+      with:
+        path: |
+          ${{github.workspace}}/build/shadps4.app
+        key: ${{ env.cache-name }}-qt-${{ matrix.os }}-${{ matrix.cxx }}-${{ hashFiles('**/CMakeLists.txt', 'cmake/**') }}
+        restore-keys: |
+          ${{ env.cache-name }}-${{ matrix.os }}-${{ matrix.cxx }}-qt-
```
squidbus: Duplicated cache was removed from macOS SDL but not here. Is there some reason this needs to be cached separately when it's already in the build folder?

LeDragoX: Thanks! I'm not used to Mac, I will fix it.

LeDragoX:
Rebased and squashed commits.

.github/workflows/macos-qt.yml
```diff
@@ -49,13 +60,13 @@ jobs:
     - name: Package
       run: |
         mkdir upload
-        mv ${{github.workspace}}/build/shadps4.app upload
-        mv ${{github.workspace}}/build/translations upload
+        cp -R ${{github.workspace}}/build/shadps4.app upload
+        cp -R ${{github.workspace}}/build/translations upload
```
LeDragoX: Copying files at the end resulted in bigger cache files and a worst build time, I am considering reverting `cp -R` to `mv` in macOS/-qt and Windows/-qt.
```suggestion
        mv ${{github.workspace}}/build/shadps4.app upload
        mv ${{github.workspace}}/build/translations upload
```

.github/workflows/linux.yml
```diff
@@ -25,8 +25,26 @@ jobs:
...
+          key: ${{ env.cache-name }}-sdl-${{ hashFiles('**/CMakeLists.txt', 'cmake/**') }}
+          restore-keys: |
+            ${{ env.cache-name }}-sdl-
+
+    - name: Cache CMake dependency build objects
+      uses: hendrikmuhs/ccache-action@v1.2.14
+      env:
+          cache-name: cache-cmake-dependency-builds
+      with:
+        key: ${{ env.cache-name }}-${{ matrix.os }}-${{ matrix.cxx }}-sdl
```
LeDragoX: I've discovered a small mistake on the cache naming, it was trying to find `matrix.os` and `matrix.cxx` which are 2 variables set at the beginning as `jobs: build: matrix: os/cxx`, used only when multiple OSes are running in the same workflow file, but here they are unused.
To fix that just need to put `runner.os` over the rest. This need to be done on all files.
```suggestion
        key: ${{ env.cache-name }}-${{ runner.os }}-sdl
```

.github/workflows/linux.yml
```diff
@@ -25,8 +25,26 @@ jobs:
...
+          key: ${{ env.cache-name }}-${{ hashFiles('**/CMakeLists.txt', 'cmake/**') }}
+          restore-keys: |
+            ${{ env.cache-name }}-
+
+    - name: Cache CMake dependency build objects
+      uses: hendrikmuhs/ccache-action@v1.2.14
+      env:
+          cache-name: ${{ runner.os }}-sdl-cache-cmake-dependency-builds
+      with:
+        key: ${{ env.cache-name }}-
```
LeDragoX: Using `append-timestamp` as `false` will avoid the creation of unnecessary multiples caches for ccache, and it's better to remove the '-' from `${{ env.cache-name }}-` because the action always append the '-' at the end.
This is also required on `linux-qt.yml`

```suggestion
      with:
        append-timestamp: false
        key: ${{ env.cache-name }}
```

LeDragoX:
Ready for review, I couldn't make _build cache_ working for Windows' CI, but I hope you'll enjoy these results 😅.

psucien:
I'm a simple man: less time -> merge. Thank you

PR #625, opened 2024-08-28
Force PIE usage on linux

Random06457:
It seems the project currently has an implicit requirement that emulator code be located at an address > 40 bits. Otherwise the code gets overwritten when the PS4 memory is allocated, resulting in memory corruptions.

Some linux distributions (e.g. NixOS and fedora(?)) come with a GCC binary built without the `--enable-default-pie` configuration enabled, which means binaries produces by such systems do not have ASLR support by default.

This causes the compiler to emit relocations which aren't compatible with a high base address (e.g. `R_X86_64_32`) and the linker to map the base address at a fixed address, which happens to be `0x400000` according to GNU's default linker script.

To address this issue, this PR forces GCC to produce a PIE executable via the `CMAKE_POSITION_INDEPENDENT_CODE` cmake variable and adds the `-pie` linker flag to `UNIX` targets.

Note that I am not 100% sure that enabling PIE guarantees that the base address is going to be > 40 bits. I only looked at the linux source tree briefly and it seems to be very configuration-dependant. However in practice the 40 bit assumption seems to hold so far, so I guess worst case scenario it only happens once and relaunching the emulator will do just fine. And if it happens too frequently, a proper linker script can be added to force the base address.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,11 +6,6 @@ cmake_minimum_required(VERSION 3.16.3)
 set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 
-# Forcing PIE makes sure that the base address is high enough so that it doesn't clash with the PS4 memory.
-if(UNIX AND NOT APPLE)
-    set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
-endif()
-
 if(APPLE)
     enable_language(OBJC)
     set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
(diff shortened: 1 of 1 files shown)


Mixaill:
@Random06457 

Is it really necessary to manually add `pie` to the linker flags, since the specified minimum version of CMake is 3.16, so CMP0083 should be in NEW state.

https://cmake.org/cmake/help/latest/policy/CMP0083.html

Random06457:
Turns out you are right, I just updated my PR.

Mixaill:
Probably `if (UNIX)` can also be removed, since it just no-op on unsupported platforms.

Random06457:
The problem is that the issue only applies to linux, and MacOS actually requires it to be disabled.

Mixaill:
In this case it should be `if(UNIX AND NOT APPLE)` or `if(LINUX)` (requires CMake 3.25) because `UNIX` includes `LINUX`, `APPLE`, `CYGWIN` and other Unix-like systems.

Random06457:
Thanks for the info

Random06457:
Oh, actually I must have made a mistake when testing removing the `target_link_options(shadps4 PRIVATE -pie)` line because I just realize that it doesn't work. `CMAKE_POSITION_INDEPENDENT_CODE` and `POSITION_INDEPENDENT_CODE` are apparently two different things.

Random06457:
@Mixaill `set_property(TARGET shadps4 PROPERTY POSITION_INDEPENDENT_CODE ON)` doesn't seem to do the trick. Only `target_link_options(shadps4 PRIVATE -pie)` adds -pie. I'm no CMake export so do you have any idea what's the issue?

Mixaill:
@Random06457 

Made a little changes to perform a proper check for PIE support https://github.com/Random06457/shadPS4/pull/1

Should works fine.

```
# cmake .. -GNinja -DCMAKE_EXPORT_COMPILE_COMMANDS=ON

# grep -i fpic ./compile_commands.json | wc -l
267

# grep -i fpie ./compile_commands.json | wc -l
172

# grep -e 'LINK_FLAGS = -fPIE -pie' ./build.ninja | wc -l
1
```

raphaelthegreat (approved):

PR #626, opened 2024-08-28
pad: add support for configurable special pad class

InvoxiPlayGames:
this adds a new config section to set the controller class to report to the game

```
[Input]
useSpecialPad = true
specialPadClass = 1
```

no GUI option currently but this lets you configure using e.g. a wheel or guitar

this should allow getting past the title screen in Guitar Hero Live (`CUSA02410`) as well as enabling more gameplay in Rock Band 4 (`CUSA02901`)

both of those games require additional patches to boot so i have not been able to test this specific branch in any game. since this makes behaviour closer to a PS4 it would be good to regression test other games to make sure they can still take inputs when useSpecialPad is set to false


georgemoralis:
looking good but unfortunately pad interface is under heavy rewrite. The main idea of this is to support multiple players and multiple devices interface . I will open soon a pr so you we can add your stuff there as well

AussiePlick:
Also add support for Rocksmith 2014 please.

PR #633, opened 2024-08-28
Mouse lib implementation

georgemoralis:
Some games seems to use mouse among with pad (keyboard,gamepad). So tried to implemented this , but i miss something it seems.

So far added only mouse buttons , but doesn't seem to response . 

If someone has any idea (turok is one game that uses it , but i remember see it in a lot third person shooters ,or point and click games)


mailwl:
this PR is completely outdated.
@georgemoralis may you create a new one, over already stubbed mouse library?

georgemoralis:
that draft had a lot of unsolved issues unfortunately . If someone wants to work on this i can try rebase it

PR #635, opened 2024-08-28
Patch `insertq`

OFFTKP:


georgemoralis:
Surely need to add a check for cpu's (AMD) that has SSE4.2a

georgemoralis:
one game that uses it , is shadowman remaster

OFFTKP:
Reviewable

src/core/cpu_patches.cpp
```diff
@@ -108,9 +118,7 @@ static Xbyak::Reg AllocateScratchRegister(
     UNREACHABLE_MSG("Out of scratch registers!");
 }
 
-#ifdef __APPLE__
```
squidbus: This was very intentionally ifdef'd out on non-Apple as it relies on their conventions for TLS in the GS segment register.

OFFTKP: Hmm in that case I could push/pop instead of using the Allocate functions if that has no side effects, or use some different storage

squidbus: From my experiements writing this system you cannot reliably use the stack or any registers without saving them somewhere first. The stack because some games have functions that use stack area below the stack pointer, so you don't have any way to know for sure whether you're clobbering data or not.

My solution was to save registers to TLS since it needs to be per-thread storage (and I have some code that saves the stack to TLS and allows arbitrary use of a replacement stack to save registers), but it still relies on the Apple TLS convention as you can do a simple write to gs + offset without any scratch registers.

squidbus: Just to be clear, if there's a way to implement the save on other OSes without any scratch space that would resolve the issue. The rest of the code shouldn't have any OS-specific issues.

src/core/cpu_patches.cpp
```diff
@@ -534,6 +550,166 @@ static void GenerateTcbAccess(const ZydisDecodedOperand* operands, Xbyak::CodeGe
...
+
+    if (operands[0].type != ZYDIS_OPERAND_TYPE_REGISTER ||
+        operands[1].type != ZYDIS_OPERAND_TYPE_REGISTER) {
+        ASSERT_MSG("operands 0 and 1 must be registers.");
+    }
+
+    const auto dst = ZydisToXbyakRegisterOperand(operands[0]);
+    const auto src = ZydisToXbyakRegisterOperand(operands[1]);
+    Xbyak::Xmm xmm_dst = *reinterpret_cast<const Xbyak::Xmm*>(&dst);
+    Xbyak::Xmm xmm_src = *reinterpret_cast<const Xbyak::Xmm*>(&src);
```
squidbus, line 741: I assume this is guaranteed for this instruction but you may want to add an assert for isXmm just in case.

src/core/cpu_patches.cpp
```diff
@@ -534,6 +550,166 @@ static void GenerateTcbAccess(const ZydisDecodedOperand* operands, Xbyak::CodeGe
...
+
+    if (immediateForm) {
+        u8 length = operands[2].imm.value.u & 0x3F;
+        u8 index = operands[3].imm.value.u & 0x3F;
+        if (length == 0) {
+            length = 64;
+        }
+
+        if (length + index > 64) {
+            ASSERT_MSG("length + index must be less than or equal to 64.");
```
squidbus: You can put the condition directly into the assert:
```
ASSERT_MSG(length + index <= 64, "length + index must be less than or equal to 64.");
```

Same for a few other asserts around here.

OFFTKP:
Changed some stuff so it works with the illegal instruction handler, feel free to review

PR #638, opened 2024-08-29
Add patches for F16C instructions under Rosetta 2.

squidbus:
Adds patches for F16C instructions which are currently missing under Rosetta 2 on macOS.

This also enhances the stack saving capabilities in the patcher to allow saving entire contexts with fewer TLS slots. Before the patcher would allocate a TLS slot for each register it needed to save. Now, it only needs two slots for an arbitrary amount of context: one slot for a thread-specific patch stack and the other for the original function stack. When a patch needs to save context it can switch out the stacks and save registers, then do the reverse when done.

Additionally some code is added for using vector and immediate operands in patches.

I tested this with CUSA16429 which is the only game I have that uses these instructions. From basic debugging examining the inputs and outputs it seems to work as expected although the game still does not progress past a black screen on launch so I can't validate it in-depth. It gets a lot further a long though until it hits a missing primitive type.

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -142,11 +142,17 @@ if (WIN32)
     target_compile_options(sirit PUBLIC "-Wno-error=unused-command-line-argument")
 endif()
 
-# date
-if (APPLE AND NOT TARGET date::date-tz)
-    option(BUILD_TZ_LIB "" ON)
-    option(USE_SYSTEM_TZ_DB "" ON)
-    add_subdirectory(date)
+if (APPLE)
+    # half
+    add_library(half INTERFACE)
+    target_include_directories(half INTERFACE half/include)
+
+    # date
+    if (NOT TARGET date::date-tz)
+        option(BUILD_TZ_LIB "" ON)
+        option(USE_SYSTEM_TZ_DB "" ON)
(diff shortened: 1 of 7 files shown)


externals/CMakeLists.txt
```diff
@@ -142,6 +142,12 @@ if (WIN32)
     target_compile_options(sirit PUBLIC "-Wno-error=unused-command-line-argument")
 endif()
 
+# half
+if (NOT TARGET half)
```
abouvier: It should just be `if (APPLE)`

externals/CMakeLists.txt
```diff
@@ -142,6 +142,12 @@ if (WIN32)
     target_compile_options(sirit PUBLIC "-Wno-error=unused-command-line-argument")
 endif()
 
+# half
+if (APPLE AND NOT TARGET half)
```
abouvier: The non-imported target `half` will never exist at this point you know :p

squidbus: There's no reason not to check in case it may be imported externally at some point.

abouvier: The imported name will probably be something like `half::half`, if it ever exists.

PR #642, opened 2024-08-29
Cheats/patches Italian translation

bigol83:
Translation of the Cheats/Patches section


georgemoralis:
probably needs to updated since we added DLC as well

bigol83:
Ok, going to check.

DanielSvoboda:
apparently I confused it(italian) with id(indonesian), thanks for the correction

PR #657, opened 2024-08-30
Allow more streamline shader skipping for debug/testing

jnackmclain:
This makes it easier for local builders to skip shaders for debugging and testing. Big shout out to Efigr in the discord for helping out on this.
Was guided by @raphaelthegreat to make RefreshComputeKey and RefreshGraphicsKey return bools to check for skips or not.
Logged a warning when skipped for clarity in the log that shader hacks were used.
I tried to make it clear in the comment that this is by no means a fix, simply an easier way to keep up with upstream while testing through games with hacks here and there.
image


jnackmclain:
Updated to pull from the config file instead so you dont even need to rebuild to troubleshoot
image
image

PR #665, opened 2024-08-30
Remove from_compute check in texture cache invalidation

adjonesey:
This is the change that fixes the character rendering in Bloodborne. 

image


raphaelthegreat:
This looks okay to me. Could you also remove the parameter from the function?

bigol83:
It doesn't work for me on latest main with this PR merged 

Cattura

raphaelthegreat:
This won't work on main because of another issue you need bb-hacks

bigol83:
> This won't work on main because of another issue you need bb-hacks

Oh, ok.

PR #670, opened 2024-08-30
vulkan: Few fixes for MoltenVK

squidbus:
* Fix an apparent mis-map of Format2_10_10_10. The Unorm version would map to `A2B10G10R10UnormPack32` with an unused duplicate mapping to `A2R10G10B10UnormPack32`, and the Snorm version would map to `A2R10G10B10SnormPack32`. Now both only map to their respective A2B10G10R10 formats. This helps for games that use these formats under MoltenVK since it does not support the A2R10G10B10 versions for vertex buffers.
* Set SPIR-V version to 1.5 if not using Vulkan 1.3 to silence a validation error about supported versions.

### DIFF
--- a/src/video_core/renderer_vulkan/liverpool_to_vk.cpp
+++ b/src/video_core/renderer_vulkan/liverpool_to_vk.cpp
@@ -287,8 +287,8 @@ vk::BorderColor BorderColor(AmdGpu::BorderColor color) {
 
 std::span<const vk::Format> GetAllFormats() {
     static constexpr std::array formats{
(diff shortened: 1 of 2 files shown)


raphaelthegreat (approved):

nyueki:
any plans for bloodborne support on mac?

mavethee:
> any plans for bloodborne support on mac?

@nyueki most of Bloodborne fixes and hacks are part of PR (#465), **not part of main releases even not to mention stable** so I'd ask there not on random merged PRs.

PR #674, opened 2024-08-30
Allow graphics/compute shader list for skipping

jnackmclain:
This makes it easier for local builders to skip shaders for debugging and testing.
This is helpful for debugging troublesome shaders on forks/branches while merging upstream changes as it provides a specific place to put them to remain constant.

RefreshComputeKey and RefreshGraphicsKey now return bools to check for skips or not.
Logged a warning when skipped for clarity in the log when shaders are skipped and of type
image

### DIFF
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
@@ -64,9 +64,8 @@ const ComputePipeline* PipelineCache::GetComputePipeline() {
 }
 
 bool ShouldSkipShader(u64 shader_hash, const char* shader_type) {
-    static constexpr u64 skip_hashes[] = {0xdeadbeefULL};
-    if (std::ranges::find(std::begin(skip_hashes), std::end(skip_hashes), shader_hash) !=
-        std::end(skip_hashes)) {
+    static constexpr std::array<u64, 0> skip_hashes = {};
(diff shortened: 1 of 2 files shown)


src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
```diff
@@ -59,7 +63,17 @@ const ComputePipeline* PipelineCache::GetComputePipeline() {
     return pipeline;
 }
 
-void PipelineCache::RefreshGraphicsKey() {
+bool ShouldSkipShader(u64 shader_hash, const char* shader_type) {
+    static constexpr u64 skip_hashes[] = {0xdeadbeefULL};
```
wheremyfoodat: (This should probably be killed before merging)

jnackmclain: array cant be empty or it wont compile, and i figured it would be helpful as an indication of what to put. if there is a better way please let me know!

rafael-57: Maybe just comment the lines and say to uncomment if needed?

jnackmclain: it was suggested in the discord to tweak how the array was checked to allow for an empty array and is reflected in latest commit

PR #677, opened 2024-08-30
[macOS] Place `shadPS4/user` folder in `Application Support`

### LINKED ISSUE #509: [macOS] Feature Request: Place user folder in Application Support
Currently the `user` folder is created in the same directory as the app bundle. 

The preferred place for it would be:
`~/Library/Application Support/shadPS4/user`

Because some users might not know how to access their user library (it’s hidden by default), a menu option in the UI that opens the user folder directly would also be useful.

mavethee:
## Details:

Attempts to resolve #509
CleanShot 2024-09-01 at 12 00 18@2x

+ Would say **50% work is done**, as I have no clue what to do with compiled translation files.
+ added `.DS_Store` for `.gitignore`

## How it works?

If it's fresh install, **it will use new requested path**.
If user folder is present near executable, it will continue using **old path**.
To use new path, you can backup user folder, create new one and replace the files.

## TODO:
- [] translations

### DIFF
--- a/src/common/path_util.cpp
+++ b/src/common/path_util.cpp
@@ -82,11 +82,20 @@ static std::filesystem::path GetBundleParentDirectory() {
 
 static auto UserPaths = [] {
 #ifdef __APPLE__
-    std::filesystem::current_path(GetBundleParentDirectory());
+    // Start by assuming the base directory is the bundle's parent directory.
+    std::filesystem::path base_dir = GetBundleParentDirectory();
+    std::filesystem::path user_dir = base_dir / PORTABLE_DIR;
+    // Check if the "user" directory exists in the current path:
+    if (!std::filesystem::exists(user_dir)) {
+        // If it doesn't exist, use the new hardcoded path:
+        user_dir =
+            std::filesystem::path(getenv("HOME")) / "Library" / "Application Support" / "shadPS4";
+    }
+#else
+    const auto user_dir = std::filesystem::current_path() / PORTABLE_DIR;
 #endif
 
     std::unordered_map<PathType, fs::path> paths;
-    const auto user_dir = std::filesystem::current_path() / PORTABLE_DIR;
 
     const auto create_path = [&](PathType shad_path, const fs::path& new_path) {
         std::filesystem::create_directory(new_path);
--- a/.gitignore
+++ b/.gitignore
@@ -409,3 +409,6 @@ FodyWeavers.xsd
 /out/*
 /third-party/out/*
 /src/common/scm_rev.cpp
+
+# for macOS
+**/.DS_Store


src/common/path_util.cpp
```diff
@@ -82,11 +82,16 @@ static std::filesystem::path GetBundleParentDirectory() {
 
 static auto UserPaths = [] {
 #ifdef __APPLE__
-    std::filesystem::current_path(GetBundleParentDirectory());
```
squidbus, line 85: Rather than getting rid of the portable mode this should detect if the portable directory exists and use it.

mavethee: > Rather than getting rid of the portable mode this should detect if the portable directory exists and use it.

I was thinking of expanding a GUI like when user is being asked about games directory, to ask about that path as well and fill defaults with `Application Support` path but did that in a rush 😅 
Reason this being a PR already cuz I needed some advices.

Will improve it.

squidbus: I think for now it would be sufficient to just check if the portable path exists and set it to the `Application Support` path if not.

mavethee: > I think for now it would be sufficient to just check if the portable path exists and set it to the `Application Support` path if not.

Should work as proposed now.
If someone wants to move to standard Apple path, just delete or move it out of folder where emulator is.

src/common/path_util.cpp
```diff
@@ -82,11 +82,19 @@ static std::filesystem::path GetBundleParentDirectory() {
...
+    const auto base_dir = GetBundleParentDirectory();
+    const auto user_dir = base_dir / "user";
+    if (std::filesystem::exists(base_dir)) {
+        return std::filesystem::path{user_dir};
+    } else {
+        const auto base_dir = std::filesystem::path(getenv("HOME")) / "Library/Application Support/shadPS4";
+        const auto user_dir = base_dir / "user";
+            std::filesystem::create_directories(user_dir);
+            return user_dir;
```
squidbus: You can't return early here or in the other path, you're skipping all the extra logic for creating the subdirectories below.

src/common/path_util.cpp
mavethee: Should work now, even a few lines of code are not suitable on weekends for me it seems.
```sh
cd GitHub/shadps4/build                                                                         
rm -rf user
# No user folder in app location, goes for requested path:                                                                                     
./shadps4.app/Contents/MacOS/shadps4                                                               
Saving new configuration file /Users/marcinmitura/Library/Application Support/shadPS4/user/config.toml
2024-08-31 14:01:02.144 shadps4[11679:3467331] +[IMKClient subclass]: chose IMKClient_Modern
2024-08-31 14:01:02.144 shadps4[11679:3467331] +[IMKInputSession subclass]: chose IMKInputSession_Modern
rm -rf /Users/marcinmitura/Library/Application Support/shadPS4/user
clear                                                                                           
mkdir user
# Present user folder in app location, goes for old path:                                                                                         
./shadps4.app/Contents/MacOS/shadps4                                                               
Saving new configuration file /Users/marcinmitura/GitHub/shadPS4/build/user/config.toml
2024-08-31 14:01:56.192 shadps4[12000:3468403] +[IMKClient subclass]: chose IMKClient_Modern
2024-08-31 14:01:56.193 shadps4[12000:3468403] +[IMKInputSession subclass]: chose IMKInputSession_Modern
```

Now the clang to fix my terrible formmating.

mavethee: Should be everything, I have no clue how to deal with translations that are compiled with app not created after first ran like user folder.

abouvier:
You can ditch the `translations` directory, because its files are compiled inside the binary and are not used.

mavethee:
> You can ditch the `translations` directory, because its files are compiled inside the binary and are not used.

Yeah I'm well aware of that.
I could try do one thing if I'd ditch portable path, but whatever squidbus says stays ig.
I'm used to symlinks if I want portable paths XD
On other hand I could also check how DuckStation handles that 🤔

squidbus:
> I could try do one thing if I'd ditch portable path

Translations directory has nothing to do with portable path. It's not needed at all because the translations are compiled into the application itself.

mavethee:
> Translations directory has nothing to do with portable path. It's not needed at all because the translations are compiled into the application itself.

Well, then I guess that's all I can do for now 🤷🏻‍♂️

mavethee:
@squidbus could you review in your free time, should be everything we can do now in this case, not like its a lot 🤷🏻
Thanks in advance :D

src/common/path_util.cpp
```diff
@@ -82,11 +82,22 @@ static std::filesystem::path GetBundleParentDirectory() {
 
 static auto UserPaths = [] {
 #ifdef __APPLE__
-    std::filesystem::current_path(GetBundleParentDirectory());
+    // Start by assuming the base directory is the bundle's parent directory.
+    std::filesystem::path base_dir = GetBundleParentDirectory();
+    std::filesystem::path user_dir = base_dir / "user";
```
squidbus: Should use the existing `PORTABLE_DIR` constant.

src/common/path_util.cpp
```diff
@@ -82,11 +82,22 @@ static std::filesystem::path GetBundleParentDirectory() {
...
+
+    // Check if the "user" directory exists in the current path:
+    if (!std::filesystem::exists(user_dir)) {
+        // If it doesn't exist, use the hardcoded path:
+        base_dir = std::filesystem::path(getenv("HOME")) / "Library/Application Support/shadPS4";
+        user_dir = base_dir / "user";
```
squidbus: I don't think the user part is actually needed here since we're already in a shadps4-specific data folder, but if you want it there then it should at least probably be using the same constant here as well.

src/common/path_util.cpp
```diff
@@ -82,11 +82,22 @@ static std::filesystem::path GetBundleParentDirectory() {
...
 #ifdef __APPLE__
-    std::filesystem::current_path(GetBundleParentDirectory());
+    // Start by assuming the base directory is the bundle's parent directory.
+    std::filesystem::path base_dir = GetBundleParentDirectory();
+    std::filesystem::path user_dir = base_dir / "user";
+
+    // Check if the "user" directory exists in the current path:
+    if (!std::filesystem::exists(user_dir)) {
+        // If it doesn't exist, use the hardcoded path:
+        base_dir = std::filesystem::path(getenv("HOME")) / "Library/Application Support/shadPS4";
```
squidbus: Should probably use the `/` operator of filesystem path instead of hardcoding in the string, e.g. `std::filesystem::path(getenv("HOME")) / "Library" / "Application Support" / "shadPS4"`

squidbus (commented):
Left a few minor comments but it generally looks fine to me. But I don't have proper review/merge power here so you'll need someone else to look at it.

mavethee:
> Left a few minor comments but it generally looks fine to me. But I don't have proper review/merge power here so you'll need someone else to look at it.

I'm aware of that, just wanted your review as initially reviewed it :P

src/common/path_util.cpp
mavethee: Done, thanks for pointing out! :D

src/common/path_util.cpp
mavethee: Yeah I was thinking that too originally, for some reason was thinking `user` folder is part of PS4's FS, nvm 🤦🏻 
Corrected.

src/common/path_util.cpp
mavethee: @squidbus Using `PORTABLE_DIR` results in shadPS4 ignoring everything after, creates user folder near the executable no matter the scenario.

squidbus: `PORTABLE_DIR` is literally just `"user"`, you should be able to replace that part and it would be no different.

So: `std::filesystem::path user_dir = base_dir / PORTABLE_DIR;`

src/common/path_util.cpp
```diff
@@ -82,11 +82,20 @@ static std::filesystem::path GetBundleParentDirectory() {
...
 #ifdef __APPLE__
-    std::filesystem::current_path(GetBundleParentDirectory());
+    // Start by assuming the base directory is the bundle's parent directory.
+    std::filesystem::path base_dir = GetBundleParentDirectory();
+    std::filesystem::path user_dir = base_dir / PORTABLE_DIR;
+    // Check if the "user" directory exists in the current path:
+    if (!std::filesystem::exists(user_dir)) {
+        // If it doesn't exist, use the new hardcoded path:
+        base_dir =
+            std::filesystem::path(getenv("HOME")) / "Library" / "Application Support" / "shadPS4";
```
squidbus, line 92: I think the issue you were having with non-portable not working is that your assignment to `user_dir` has disappeared from here, you should probably be setting `user_dir` instead of `base_dir` to this.

squidbus:
Looks good to me now 👍

mavethee:
> Looks good to me now 👍

Thanks a lot then 🖤

PR #681, opened 2024-08-30
Implement V_READFIRSTLANE_B32

greggameplayer:
Needed by CUSA00006 (Knack) which falls in the case exec mask different from zero.


raphaelthegreat:
The exec check here will never trigger and doesn't make much sense to me (if exec is false than the invocation should not be execute vector ops at all)

PR #686, opened 2024-08-31
savedata: Handle a missing case in saveDataMount

Exhigh:
This should potentially fix the mount issue in Resistance : Retribution (CUSA32636)

Testing is welcome.


Exhigh:
Once the fix is confirmed to fix the issue in CUSA32636, this PR should be ready for review.

greggameplayer:
@Exhigh thanks for the fix, i test it this evening and i will tell the result here

greggameplayer:
it seems to progress further and error here : 
image

here the log : 
[shad_log.txt](https://github.com/user-attachments/files/16826014/shad_log.txt)

i try to pass that point as soon as i hit continue it keeps throwing that error but it writes a new line in the log of stub : 
[Core] <Error> stubs.cpp:CommonStub:42: Stub: sceRazorCpuPlotValue (nid: njGikRrxkC0) called

greggameplayer:
i don't know if it's really related to libSceRazorCpu because it seems to be a library use only for application debugging purposes for developers

lzardy:
> i don't know if it's really related to libSceRazorCpu because it seems to be a library use only for application debugging purposes for developers

Yes it is likely a game functionality is not behaving correctly and the game attempts to debug it.

PR #690, opened 2024-08-31
update zh_CN translation

bigQY:
Added Chinese translation for some texts.

### DIFF
--- a/.github/workflows/windows-qt.yml
+++ b/.github/workflows/windows-qt.yml
@@ -4,6 +4,7 @@
 name: Windows-Qt
 
 on:
+  workflow_dispatch:
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/windows-qt.yml
```diff
@@ -4,6 +4,7 @@
 name: Windows-Qt
 
 on:
+  workflow_dispatch:
```
abouvier, line 7: Why did you add that?

bigQY: I added this line of code to be able to run workflow manually, which is not necessary in this pr. I chose the wrong branch when I submitted this pr, I will close the current pr and submit a new one. #699 

Thank you.

PR #695, opened 2024-08-31
ci: Remove translations folder from Windows/macOS upload.

squidbus:
Removes the translations folder from the Windows and macOS Qt distribution as it is not needed; the translations resources are baked into the app.


LeDragoX:
> I didn't touch Windows as I can't verify it, someone else can do it if they know its not needed there either.

I tested removing the translations folder on **Windows-Qt** in the latest `main` from Actions, and indeed, it's not needed.

squidbus:
> I tested removing the translations folder on **Windows-Qt** in the latest `main` from Actions, and indeed, it's not needed.

Thanks, updated that as well.

PR #705, opened 2024-09-01
CI: include build date and git hash in build artifact name

tGecko:
Not sure if this is something y'all would want, but I think it's useful to differentiate builds.
 
Include build date and git short hash in build artifact name:
image

to avoid this
image


gandalfthewhite19890404:
Add date too, i.e. 2024-09-01 (YYYY-MM-DD) prior hash

georgemoralis:
> Add date too, i.e. 2024-09-01 (YYYY-MM-DD) prior hash

yah good idea

tGecko:
great idea, added! 👍🏻

PR #708, opened 2024-09-01
Replace recursive_mutex with mutex

OFFTKP:


georgemoralis:
some details pls?

OFFTKP:
My understanding is that recursive_mutex is only used here so that member functions can call other member functions without deadlocking, however this can be solved by making a private version of the function that doesn't lock and calling that instead

src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -231,6 +231,11 @@ u32 BufferCache::BindIndexBuffer(bool& is_indexed, u32 index_offset) {
 std::pair<Buffer*, u32> BufferCache::ObtainBuffer(VAddr device_addr, u32 size, bool is_written,
                                                   bool is_texel_buffer) {
     std::scoped_lock lk{mutex};
```
raphaelthegreat: The lock in ObtainBuffer is actually not necessary. Can be safely moved to SynchronizeBuffer and revert this rename (should also help perf a bit by locking mutex far less)

OFFTKP: Cool, done

raphaelthegreat (approved):

PR #710, opened 2024-09-01
Make AvPlayerQueue size atomic

OFFTKP:
std::queue::size is not thread safe, and also

```cpp
if (Size() == 0) {
    return std::nullopt;
}
std::lock_guard guard(m_mutex);
auto result = std::move(m_queue.front());
m_queue.pop();
```
doesn't look thread safe, because the Size() == 0 check can be false, but before the mutex is locked a different thread pops an element or clears the queue making size 0, which leads to UB

Added another size check after locking the mutex instead of locking it at the start of the function, so that Pop doesn't unconditionally lock but it also doesn't try to pop from an empty queue

### DIFF
--- a/src/core/libraries/avplayer/avplayer_common.h
+++ b/src/core/libraries/avplayer/avplayer_common.h
@@ -58,12 +58,13 @@ template <class T>
 class AvPlayerQueue {
 public:
     size_t Size() {
-        return m_queue.size();
+        return m_size;
     }
 
     void Push(T&& value) {
         std::lock_guard guard(m_mutex);
         m_queue.emplace(std::forward<T>(value));
+        ++m_size;
     }
 
     std::optional<T> Pop() {
@@ -71,18 +72,24 @@ public:
             return std::nullopt;
         }
         std::lock_guard guard(m_mutex);
+        if (Size() == 0) { // in case size changes while the mutex is being locked
+            return std::nullopt;
+        }
         auto result = std::move(m_queue.front());
         m_queue.pop();
+        --m_size;
         return result;
     }
 
     void Clear() {
         std::lock_guard guard(m_mutex);
         m_queue = {};
+        m_size = 0;
     }
 
 private:
     std::mutex m_mutex{};
+    std::atomic_size_t m_size{0};
     std::queue<T> m_queue{};
 };
 


roamic:
This is a SPSC queue, the described scenario can never happen. This mutex is redundant.

OFFTKP:
> This is a SPSC queue, the described scenario can never happen. This mutex is redundant.

What do you mean by "this mutex is redundant"? The mutex was already there and std::queue by itself is not thread safe even in SPSC.

roamic:
Please describe a scenario where this will be a problem in SPSC.

OFFTKP:
Is it SPSC? AvPlayerSource::GetVideoData calls ::Pop, and as far as I can see that gets called from sceAvPlayerGetVideoData, and also gets called from VideoDecoderThread for example. That's two places that call Pop is it not?

roamic:
`GetVideoData` call `Pop` on `m_video_frames`. `VideoDecoderThread` does not call `Pop` on this queue.

OFFTKP:
And only one thread can call sceAvPlayerGetVideoData at a time?

roamic:
> And only one thread can call sceAvPlayerGetVideoData at a time?

Correct.

OFFTKP:
Then in that case if these queues are externally synchronized everywhere the mutex can be removed and PR can be closed. If they aren't synchronized externally then the mutex can't be removed because std::queue is not thread safe in SPSC. This sort of middle state of "sorta MPMC safe but sorta not" is just very misleading if you decide to keep the mutex.

roamic:
The `std::queue` is "externally" synchronized by this `mutex`. `queue::size` has no need to be synchronized because it's just reading an integral variable from `std::queue`. If It's 0 in the consumer thread then it will just wait for producer to push values. I'm sorry but I do not understand your case.

OFFTKP:
> The `std::queue` is "externally" synchronized by this `mutex`. `queue::size` has no need to be synchronized because it's just reading an integral variable from `std::queue`. If It's 0 in the consumer thread then it will just wait for producer to push values. I'm sorry but I do not understand your case.

My response was to you saying "this mutex is redundant". This PR does not add a mutex, so I am assumed you meant we should remove the original mutex that was already there

roamic:
This was bad wording on my part, sorry. The `mutex` that synchronizes `Push` and `Pop` is not redundant. I meant that the change as a whole is redundant in SPSC scenario.

PR #712, opened 2024-09-01
Use standard user paths thanks to SDL

abouvier:
Use the old portable `user` directory if it exists (in the app directory), otherwise ask SDL to return a platform specific standard user directory. It should work on macOS too, but I'm not sure :p

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -855,6 +855,7 @@ if (ENABLE_QT_GUI)
     set_target_properties(shadps4 PROPERTIES
 #       WIN32_EXECUTABLE ON
         MACOSX_BUNDLE ON
+        MACOSX_BUNDLE_INFO_PLIST "${CMAKE_CURRENT_SOURCE_DIR}/cmake/MacOSXBundleInfo.plist.in"
         MACOSX_BUNDLE_ICON_FILE shadPS4.icns)
 
     set_source_files_properties(src/images/shadPS4.icns PROPERTIES
--- a/cmake/MacOSXBundleInfo.plist.in
+++ b/cmake/MacOSXBundleInfo.plist.in
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+SPDX-FileCopyrightText: 2024 shadPS4 Emulator Project
+
+SPDX-License-Identifier: GPL-2.0-or-later
+-->
+
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>English</string>
+	<key>CFBundleExecutable</key>
+	<string>${MACOSX_BUNDLE_EXECUTABLE_NAME}</string>
+	<key>CFBundleGetInfoString</key>
+	<string>${MACOSX_BUNDLE_INFO_STRING}</string>
+	<key>CFBundleIconFile</key>
+	<string>${MACOSX_BUNDLE_ICON_FILE}</string>
+	<key>CFBundleIdentifier</key>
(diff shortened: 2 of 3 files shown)


squidbus:
For portable path this will not work out of the box on macOS. The existing bundle path helper needs to be used instead of `SDL_GetBasePath()` for full compatibility.

Technically adding `SDL_FILESYSTEM_BASE_DIR_TYPE` with a value of `parent` to the `Info.plist` for the app would mostly address this although it would not work for a freshly downloaded app in the downloads folder due to translocation. I suppose with the pref path fallback it would be acceptable, I'm not sure people really need the portable directory on a fresh, un-moved download.

This can be done by adding this file: https://github.com/Kitware/CMake/blob/master/Modules/MacOSXBundleInfo.plist.in

Then add to the end:
```
	<key>SDL_FILESYSTEM_BASE_DIR_TYPE</key>
	<string>parent</string>
```

Then finally set this in CMake under `MACOSX_BUNDLE ON`: 
```MACOSX_BUNDLE_INFO_PLIST "${CMAKE_SOURCE_DIR}/(path to the file)"```

abouvier:
`MACOSX_BUNDLE_INFO_PLIST` should not be needed if the custom file is still named `MacOSXBundleInfo.plist.in` and is inside one of the `CMAKE_MODULE_PATH`.

squidbus:
> `MACOSX_BUNDLE_INFO_PLIST` should not be needed if the custom file is still named `MacOSXBundleInfo.plist.in` and is inside one of the `CMAKE_MODULE_PATH`.

As it is now I'm not seeing the key in the `Info.plist` of the GitHub Actions build.

abouvier:
Probably [because](https://github.com/Kitware/CMake/blob/ae6bd4d5b63e55b1f1e39d78406c65a9bb2b1b8d/Source/cmLocalGenerator.cxx#L4333) of [this](https://github.com/Kitware/CMake/blob/ae6bd4d5b63e55b1f1e39d78406c65a9bb2b1b8d/Source/cmMakefile.cxx#L3887-L3896). Also [this](https://gitlab.kitware.com/cmake/cmake/-/issues/19442)?

squidbus:
Yep seems to be working now, thanks.

Only other suggestion I have is maybe set the `std::filesystem::current_path` to the base path from SDL as if a user leaves the game directory blank it tries to scan from wherever the working directory was initialized to (which if you open from Finder seems to be the root of the filesystem).

squidbus:
This looks good, is it going to get merged at all soon? Would be nice to have some of the path stuff finalized.

georgemoralis:
i am not a big fan to depend on sdl for core stuff since that might change in the future , maybe at least can be done with a SDLBUILD def or something

abouvier:
It can be done with `QStandardPaths` too, but only `SDL` is common to both GUI and non-GUI version.
You will ditch `SDL` at some point? Many things will have to be rewritten then :p

georgemoralis:
we are not planning to ditch SDL but if it can be done with standard std we prefer this way . If it works better in apple with sdl implementation we can leave sdl implementation for mac only

squidbus:
> we are not planning to ditch SDL but if it can be done with standard std we prefer this way . If it works better in apple with sdl implementation we can leave sdl implementation for mac only

There is no C++ standard library utility for getting user data directories.

raphaelthegreat:
I think this is okay, but could it be limited to macos only for now? It would be pretty big change to move directory like that on all platforms and need some discussion first

abouvier:
If the compiled app bundle is distributed with an empty `user` directory, this will not change anything.

raphipod:
I just thought I'd chime in and say that we should respect the (more or less) standardized app save locations for each platform. 
In Linux for example, the current save location for shadPS4 is /home/$USER/user/, which seems like an oversight and doesn't help the clarity in the already cluttered home directory.

abouvier:
The current location is `~/user`, only if you launch shadps4 from your home directory.

Hermiten:
Can it be merge or are you still discussing adding it ?
Just to have an update, put labels , etc ..

abouvier:
Not needed anymore I guess.

PR #717, opened 2024-09-01
video_core: Add bounds checking for subspan use in liverpool functions

baggins183:
I have seen Nop packets that exceed the remaining submission length for the command list. When this happens, for me subspan underflows and returns a span with a very high, garbage size, which causes the ProcessCompute function to keep looping.
This commit adds an error message when this happens, and returns an empty span so at least the loop bails out.

Also adds stub cases for type 0 and 1 packets in ProcessCompute

Happens in CUSA01127 (PT)

### DIFF
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -20,6 +20,20 @@ static const char* acb_task_name{"ACB_TASK"};
 
 std::array<u8, 48_KB> Liverpool::ConstantEngine::constants_heap;
 
+static std::span<const u32> NextPacket(std::span<const u32> span, size_t offset) {
+    if (offset > span.size()) {
+        LOG_ERROR(
+            Lib_GnmDriver,
+            ": packet length exceeds remaining submission size. Packet dword count={}, remaining "
+            "submission dwords={}",
+            offset, span.size());
+        // Return empty subspan so check for next packet bails out
+        return {};
+    }
+
+    return span.subspan(offset);
+}
+
 Liverpool::Liverpool() {
     process_thread = std::jthread{std::bind_front(&Liverpool::Process, this)};
 }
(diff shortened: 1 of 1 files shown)


src/video_core/amdgpu/liverpool.cpp
```diff
@@ -20,6 +20,20 @@ static const char* acb_task_name{"ACB_TASK"};
 
 std::array<u8, 48_KB> Liverpool::ConstantEngine::constants_heap;
 
+static std::span<const u32> checked_subspan(std::span<const u32> span, size_t offset) {
```
raphaelthegreat:
```suggestion
static std::span<const u32> NextPacket(std::span<const u32> span, size_t offset) {
```
Shorter name and I feel more expressive of what function does

baggins183:
maybe should have kept the extra switch in a different commit because it look fucked
but I just found out you can add ?w=1 to the url to show a better diff

src/video_core/amdgpu/liverpool.cpp
```diff
@@ -544,90 +558,99 @@ Liverpool::Task Liverpool::ProcessCompute(std::span<const u32> acb, int vqid) {
...
-            UNREACHABLE_MSG("Invalid PM4 type {}", type);
-        }
 
-        const u32 count = header->type3.NumWords();
-        const PM4ItOpcode opcode = header->type3.opcode;
-        const auto* it_body = reinterpret_cast<const u32*>(header) + 1;
-        switch (opcode) {
-        case PM4ItOpcode::Nop: {
-            const auto* nop = reinterpret_cast<const PM4CmdNop*>(header);
+        switch (type) {
```
psucien: Please return it back. Packet type 2 is not used on ASC queue (even for alignment cases as GFX has)

psucien (changes requested):

PR #719, opened 2024-09-01
use ComPtr in gui_context_menus.h

oltolm:
see title


georgemoralis:
ok and why that is a better solution?

oltolm:
> ok and why that is a better solution?

Because the [`ComPtr`](https://learn.microsoft.com/en-us/cpp/cppcx/wrl/comptr-class?view=msvc-170) is a smart pointer that automatically does reference counting. It is best practice for writing COM code on Windows. That is what Microsoft recommends.

oltolm:
ping

georgemoralis:
@oltolm since there are a lot of changes do a rebase first just in case :)

raphaelthegreat:
You seem to have removed some pragma GCC wrappers around vma include. Can you restore/remove any unrelated changes so this can be merged

oltolm:
> You seem to have removed some pragma GCC wrappers around vma include. Can you restore/remove any unrelated changes so this can be merged

I removed the pragmas because of this discussion https://github.com/shadps4-emu/shadPS4/pull/749. They were unnecessary.

raphaelthegreat (approved):
Ah I see

PR #720, opened 2024-09-01
Preserve flags on some patched instructions

OFFTKP:


src/core/cpu_patches.cpp
```diff
@@ -218,7 +218,7 @@ static void SaveContext(Xbyak::CodeGenerator& c) {
         c.push(Xbyak::Reg64(reg));
     }
     for (int reg = 0; reg <= 7; reg++) {
-        c.sub(rsp, 32);
+        c.lea(rsp, ptr[rsp - 32]);
```
wheremyfoodat, line 221: (Fwiw I'd change this to lower rsp once at the start and then emit an offset in the vmovdqu instructions. Should be faster, likely smaller and preserves flags if done with a lea)

OFFTKP: Yeah that would be nice, I'll do it in a different PR as to not overcomplicate these changes

src/core/cpu_patches.cpp
```diff
@@ -363,6 +363,8 @@ static void GenerateVCVTPH2PS(const ZydisDecodedOperand* operands, Xbyak::CodeGe
 
     SaveContext(c);
 
+    c.pushfq(); // VCVTPH2PS shouldn't modify flags
```
squidbus: I saw this in your other PR and was planning to do this when I had the chance, thanks for this. Could you move the push/popfq to Save/RestoreContext though since that's part of the general intent of those helpers.

OFFTKP: That's what I thought of doing initially, however what if you wanna use Save/RestoreContext in some instruction that wants to modify flags in some way and you don't want to just keep the old ones? Or is Save/Restore only going to be used on instructions that must not modify flags?

squidbus: I think having a boolean flag into Save/RestoreContext for whether to preserve flags, defaulted to false, would work.

OFFTKP: Done

squidbus:
Tested a few games on macOS, LGTM 👍

PR #722, opened 2024-09-01
Set CF correctly on BLSI patch

OFFTKP:
BLSI needs to set CF if source is not zero and clear it otherwise:
```
temp := (-SRC) bitwiseAND (SRC);
SF := temp[OperandSize -1];
ZF := (temp = 0);
IF SRC = 0
    CF := 0;
ELSE
    CF := 1;
FI
DEST := temp;
```

Since unfortunately AND clears CF, we need to take the CF from NEG and apply it after the AND.
NEG sets CF like so:
```
IF DEST = 0
    THEN CF := 0;
    ELSE CF := 1;
FI;
DEST := [– (DEST)]
```

### DIFF
--- a/src/core/cpu_patches.cpp
+++ b/src/core/cpu_patches.cpp
@@ -308,21 +308,21 @@ static void GenerateBLSI(const ZydisDecodedOperand* operands, Xbyak::CodeGenerat
     SaveRegisters(c, {scratch});
 
     // BLSI sets CF to zero if source is zero, otherwise it sets CF to one.
-    Xbyak::Label carrySet, carryNotSet, end;
+    Xbyak::Label set_carry, clear_carry, end;
 
     c.mov(scratch, *src);
     c.neg(scratch); // NEG, like BLSI, clears CF if the source is zero and sets it otherwise
-    c.jc(carrySet);
-    c.jmp(carryNotSet);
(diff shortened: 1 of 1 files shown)


src/core/cpu_patches.cpp
```diff
@@ -307,9 +307,24 @@ static void GenerateBLSI(const ZydisDecodedOperand* operands, Xbyak::CodeGenerat
...
+    c.jmp(carryNotSet);
+
+    c.L(carrySet);
     c.and_(scratch, *src);
+    c.stc(); // setting/clearing carry needs to happen after the AND because that clears CF
+    c.jmp(end);
+
+    c.L(carryNotSet);
+    c.and_(scratch, *src);
+    c.clc();
```
wheremyfoodat: This CLC is redundant given that and has already cleared CF, but if you want to abuse this you should definitely put a comment about it here.

There's some easy ways to copy ZF to CF without any branches but the AND makes it fairly annoying

OFFTKP: Oh true thanks

Yeah I tried to think of a way to do it without branches and couldn't come up with something

squidbus:
Tested a few games on macOS, LGTM 👍

PR #724, opened 2024-09-01
Set CF correctly on BLSR/BLSMSK patch

OFFTKP:
Similarly to BLSI, BLSR needs to modify CF.
If source is zero CF is set, otherwise it's cleared:

```
temp := (SRC-1) bitwiseAND ( SRC );
SF := temp[OperandSize -1];
ZF := (temp = 0);
IF SRC = 0
    CF := 1;
ELSE
    CF := 0;
FI
DEST := temp;
```
Also doing the same on BLSMSK

### DIFF
--- a/src/core/cpu_patches.cpp
+++ b/src/core/cpu_patches.cpp
@@ -323,9 +323,26 @@ static void GenerateBLSMSK(const ZydisDecodedOperand* operands, Xbyak::CodeGener
 
     SaveRegisters(c, {scratch});
 
+    Xbyak::Label set_carry, clear_carry, end;
+
+    // BLSMSK sets CF to zero if source is NOT zero, otherwise it sets CF to one.
     c.mov(scratch, *src);
+    c.test(scratch, scratch);
+    c.jz(set_carry);
+    c.jmp(clear_carry);
+
+    c.L(set_carry);
     c.dec(scratch);
     c.xor_(scratch, *src);
+    c.stc();
+    c.jmp(end);
+
+    c.L(clear_carry);
+    c.dec(scratch);
+    c.xor_(scratch, *src);
+    // We don't need to clear carry here since XOR does that for us
+
+    c.L(end);
     c.mov(dst, scratch);
 
     RestoreRegisters(c, {scratch});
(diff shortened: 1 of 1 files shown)


squidbus:
Seems to work fine from my testing over a few games.

raphaelthegreat (approved):

src/core/cpu_patches.cpp
```diff
@@ -339,9 +356,26 @@ static void GenerateBLSR(const ZydisDecodedOperand* operands, Xbyak::CodeGenerat
 
     SaveRegisters(c, {scratch});
 
+    Xbyak::Label set_carry, clear_carry, end;
+
+    // BLSR sets CF to zero if source is NOT zero, otherwise it sets CF to one.
     c.mov(scratch, *src);
+    c.test(scratch, scratch);
+    c.jz(set_carry);
```
PatrickvL, line 364: Perhaps it's possible to this here?
`c.jnz(clear_carry);`

src/core/cpu_patches.cpp
```diff
@@ -339,9 +356,26 @@ static void GenerateBLSR(const ZydisDecodedOperand* operands, Xbyak::CodeGenerat
...
+    c.jmp(clear_carry);
```
PatrickvL, line 365: Then remove this line

src/core/cpu_patches.cpp
```diff
@@ -339,9 +356,26 @@ static void GenerateBLSR(const ZydisDecodedOperand* operands, Xbyak::CodeGenerat
...
+
+    c.L(set_carry);
```
PatrickvL, line 367: Then change this into;
`// set carry`

src/core/cpu_patches.cpp
```diff
@@ -339,9 +356,26 @@ static void GenerateBLSR(const ZydisDecodedOperand* operands, Xbyak::CodeGenerat
 
     SaveRegisters(c, {scratch});
 
+    Xbyak::Label set_carry, clear_carry, end;
```
PatrickvL, line 359: With the other suggestions, the `set_carry` variable could be removed from here.

PatrickvL (commented):
A bit late, but still one possible improvements on this.

PatrickvL:
Also the comments added in #722 would be appropriate here as well

OFFTKP:
Indeed, done in #770 thanks!

PR #738, opened 2024-09-02
Fix download Cheats/Patches on Linux

DanielSvoboda:


abouvier:
What distribution are you using? OpenSSL should already be linked to Qt6Network.

DanielSvoboda:
I only use Windows 10 Pro.
However, several people reported to me that the download was not working on Linux. And just adding Qt6Network doesn't work.

abouvier:
It works fine on Arch Linux.
It's weird to link a library when no file in the source tree uses this library directly.

abouvier:
I maintain that this is a bad solution for a problem that comes from elsewhere.
The library will not even be linked if you have `-Wl,--as-needed` in your `LDFLAGS`.

DanielSvoboda:
this worked for compilation with the github action. I don't know how to solve it any other way, if possible make a PR with a better solution, because I don't understand this part.

PR #739, opened 2024-09-02
shader_recompiler: Implement render target swizzles when no format is available

raphaelthegreat:
Dark Souls requests STD_ALT swap on R16G16B16A16 render target format. There is no vulkan format that directly emulates the requested order so instead it is emulated in the shader. Also fixed a bug with V_CVT_OFF_F32_I4 implementation. These fix the rainbow textures in Dark Souls


bigol83:
This also fixes Red Dead Redemption skybox


https://github.com/user-attachments/assets/fbfbbaf9-a009-40b2-93be-960694df23e2

Foul-Tarnished:
Might fix the rainbow issue on Bloodborne armor
but the bug seem random, don't know of a way to reproduce it

raphaelthegreat:
Latest commit includes a non-trivial refactor of shader recompiler interface to make passing state like this easier. Testing to ensure no regressions is welcome

bigol83:
No regressions for me with latest commits

PR #499, opened 2024-08-20
Amplitude (CUSA02480) former fixes and workarounds

xezrunner:
**Previous** fixes and temporary workarounds for Amplitude (2016) that were necessary to boot in-game.  
**NOTE:** these workarounds no longer apply with the latest upstream code!

Upstream now boots the game into the main menu with no workarounds. Using as temporary progress tracker, until below list is completed.

Unchecked items are workarounds that need investigation:
- [x] CFG fix (https://github.com/shadps4-emu/shadPS4/pull/497)
- [x] Handle SRGB images
- [x] Assertions for warp instructions during compute stage
- [x] **[crash]** `V_MOVREL`... shader instructions (missing implementation, pending: https://github.com/shadps4-emu/shadPS4/pull/745)
- [ ] Handle detiling `Display_MacroTiled` `SRGB` textures

image

PR #745, opened 2024-09-03
Implement V_MOVREL variants

baggins183:
shader_recompiler: Implement V_MOVRELS_B32

    Generates a ton of OpSelects to hardcode reading from the VGPR at src
    index if the index is equal to 1..max_allocated_vgpr. Then do that
    again to choose and write the dst vgpr

    Also add support for reading/writing m0 register.


baggins183:
Ugly cuz of all the OpSelects but I dont think theres any great solution. 
I think the best solution would be to generate OpSwitch during spirv
emitting, but it would need a case for each VGPR values to load it, then
a case to store back to each VPGR at emit time. You lose that ability after
ssa_rewrite.

IndecisiveTurtle also mentioned making the VGPRs an array in the IR so you have the ability to dynamically index.

baggins183:
helps #496

baggins183:
I need to check, but I think we can improve the src register indexing switch.
We should be able to eliminate any cases for VGPRs which are uninitialized. I am assuming the SSA pass will handle that part by replacing uninitialized register reads with %UndefU32.
Then we can pattern match and replace Select(cond, Undef, some_val) -> some_val (or the symmetrical version).

baggins183:
One comment about the option of turning the VGPRs into an array so we can dynamically index:
that could do more harm then good if the vk driver compiler (Nvidia, AMD, etc) sees the dynamic index and spills it to cache or memory, because I dont know if they can address registers dynamically.

If the shader doesnt contain MOVREL, the spirv shouldnt have dynamic indexing and it should be fine, and the vk compiler should be able to assign ssa values to individual array elements and promote them to registers. But if there is and it causes a spill, it will hurt the rest of the shader's perf, which might make it not worth it

baggins183:
for the switch cases that store a value to each possible VGPR, we can get rid of them if that VGPR isn't live after that. I think SSA + DCE should handle that automatically. Need to check.

xezrunner:
This PR crashes the GPU driver with Amplitude and Rock Band 4, which make use of the `movrel` instructions in some of their main shaders.

https://github.com/shadps4-emu/shadPS4/pull/499#issuecomment-2329549247 (Amplitude-specific PR/discussion source)

364469698-3c2ebd49-a0a0-4282-bad7-645f4095f984

[Brought/fixed up this PR to upstream privately](https://github.com/xezrunner/shadPS4/tree/pull/745_with_main) to gain some insight into the crash:

Vulkan validation layer crash (recent upstream feature, `config.toml`/`validation,validation_gpu,crashDiagnostic = true`):
[cdl_dump.yaml.txt](https://github.com/user-attachments/files/16875207/cdl_dump.yaml.txt)

Dump of the compute shader that causes the crash:
[cs_0x00000000c7f34c4f_0.bin.txt](https://github.com/user-attachments/files/16875228/cs_0x00000000c7f34c4f_0.bin.txt)

baggins183:
> Dump of the compute shader that causes the crash:
[cs_0x00000000c7f34c4f_0.bin.txt](https://github.com/user-attachments/files/16875228/cs_0x00000000c7f34c4f_0.bin.txt)

`.L3144_0:`
`/*000000000c48*/ v_readfirstlane_b32 s20, v22`
`/*000000000c4c*/ s_mov_b32       m0, s20`
`/*000000000c50*/ v_cmpx_eq_u32   s[20:21], s20, v22`
`/*000000000c58*/ v_movreld_b32   v0, v32`
`/*000000000c5c*/ v_movreld_b32   v1, v33`
`/*000000000c60*/ v_movreld_b32   v2, v34`
`/*000000000c64*/ s_andn2_b64     s[18:19], s[18:19], s[20:21]`
`/*000000000c68*/ v_movreld_b32   v3, v19`
`/*000000000c6c*/ s_mov_b64       exec, s[18:19]`
`/*000000000c70*/ s_cbranch_scc1  .L3144_0`
`/*000000000c74*/ s_mov_b64       exec, s[16:17]`
`/*000000000c78*/ v_sad_u32       v41, 1, 0, v41`
`/*000000000c80*/ s_mov_b32       m0, s22`
`.L3204_0:`

Seems like theres a consistent pattern with READLANE and MOVREL.
This MOVREL implementation will definitely be slow here, especially since it looks like MOVREL is called multiple times in a loop that could potentially diverge for every thread in the wave. It looks like it iterates once per distinct value of v22 among the threads in the wave
Would like to figure out what its trying to do here and optimize it

Im not sure if MOVREL is functionally wrong though, just very slow. The IR ive seen generated by MOVREL looks correct, so it could be hitting a different issue.

baggins183:
Hopefully fixed now. Like turtle was saying, we should do range analysis. Hopefully we can put an upper bound on m0 so there are fewer src or dst registers to consider

georgemoralis:
can you do a rebase as well ?:)

polybiusproxy:
New implementation fixes broken models on Persona 4 GOLDEN:
<details>
<summary>Before</summary>

</details>
<details>
<summary>After</summary>

</details>

PR #749, opened 2024-09-03
Fix warnings

oltolm:
fixes warnings like this
```
C:/src/shadPS4/externals/vma/include/vk_mem_alloc.h:16141:22: warning: pointer is missing a nullability type specifier (_Nonnull, _Nullable, or _Null_unspecified) [-Wnullability-completeness]
 16141 |     VmaAllocationInfo* pAllocationInfo)
       |                      ^
C:/src/shadPS4/externals/vma/include/vk_mem_alloc.h:16141:22: note: insert '_Nullable' if the pointer may be null
 16141 |     VmaAllocationInfo* pAllocationInfo)
       |                      ^
       |                        _Nullable
C:/src/shadPS4/externals/vma/include/vk_mem_alloc.h:16141:22: note: insert '_Nonnull' if the pointer should never be null
 16141 |     VmaAllocationInfo* pAllocationInfo)
       |                      ^
       |                        _Nonnull
```

```
C:/src/shadPS4/src/qt_gui/cheats_patches.cpp:1096:49: warning: implicit capture of 'this' with a capture default of '=' is deprecated [-Wdeprecated-this-capture]
 1096 |                             [=](bool checked) { applyPatch(patchName, checked); });
      |                                                 ^
C:/src/shadPS4/src/qt_gui/cheats_patches.cpp:1096:30: note: add an explicit capture of 'this' to capture '*this' by reference
 1096 |                             [=](bool checked) { applyPatch(patchName, checked); });
      |                              ^
      |                               , this
```

### DIFF
--- a/src/qt_gui/cheats_patches.cpp
+++ b/src/qt_gui/cheats_patches.cpp
@@ -147,13 +147,13 @@ void CheatsPatches::setupUI() {
     controlLayout->addWidget(downloadComboBox);
 
     QPushButton* downloadButton = new QPushButton(tr("Download Cheats"));
-    connect(downloadButton, &QPushButton::clicked, [=]() {
+    connect(downloadButton, &QPushButton::clicked, [this, downloadComboBox]() {
         QString source = downloadComboBox->currentData().toString();
         downloadCheats(source, m_gameSerial, m_gameVersion, true);
     });
 
     QPushButton* deleteCheatButton = new QPushButton(tr("Delete File"));
-    connect(deleteCheatButton, &QPushButton::clicked, [=]() {
+    connect(deleteCheatButton, &QPushButton::clicked, [this, CHEATS_DIR_QString]() {
         QStringListModel* model = qobject_cast<QStringListModel*>(listView_selectFile->model());
         if (!model) {
             return;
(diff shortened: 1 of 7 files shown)


abouvier:
`-Wnullability-completeness` is only supported by `clang` and produces more warnings with `gcc`.
Besides it should only be ignored once, with `target_compile_options` in `CMakeLists.txt`.

oltolm:
> `-Wnullability-completeness` is only supported by `clang` and produces more warnings with `gcc`. Besides it should only be ignored once, with `target_compile_options` in `CMakeLists.txt`.

I added a commit that fixes the warnings to https://github.com/shadps4-emu/shadPS4/pull/719. I didn't suppress the warning in the CMake script because I wanted to suppress it in `vk_mem_alloc.h` only.

abouvier:
You can add compile options to one [file](https://cmake.org/cmake/help/latest/command/set_source_files_properties.html) or one [target](https://github.com/flyinghead/flycast/blob/7174ef11948b0c58ca0ff8237141265f1ecc26c0/CMakeLists.txt#L1345) specifically.

Besides, do you really need to add the option everywhere while `vk_mem_alloc.h` is only compiled once in `src/video_core/renderer_vulkan/vk_common.cpp` where `VMA_IMPLEMENTATION` is defined?

oltolm:
You are right it is only needed in `src/video_core/renderer_vulkan/vk_common.cpp`.

PR #756, opened 2024-09-03
adds TR in all missing places

DanielSvoboda:
add TR for [PR 700](https://github.com/shadps4-emu/shadPS4/pull/700)
tr("Can't apply cheats before the game is started")
image

add: Icon,Name,Serial,Region,Firmware,Size,Version,Path
image

add: Restore Defaults, Save, Close, Apply
image

also added more information when downloading patches:
"If the patch does not appear, it may be that it does not exist for the specific serial and version of the game. It may be necessary to update the game."
image

*note, in fa_IR.ts I fixed the spacing, so it looks like I changed everything, but it was just the spaces and the separation of MainWindow.h and MainWindow.cpp


src/qt_gui/translations/pl_PL.ts
```diff
@@ -898,5 +898,10 @@
             <source>Failed to parse JSON:</source>
             <translation>Nie udało się przeanlizować JSON:</translation>
         </message>
+		<message>
+			<location filename="../cheats_patches.cpp" line="1163"/>
+			<source>Can't apply cheats before the game is started</source>
+			<translation>Nie można zastosować cheatów przed rozpoczęciem gry.</translation>
```
mavethee, line 904:
```suggestion
			<translation>Nie można zastosować kodów przed uruchomieniem gry.</translation>
```

mavethee (commented):
Correct polish translation

DanielSvoboda:
@mavethee is this right? I added your suggestion, but it seems to have disappeared

src/qt_gui/translations/pl_PL.ts
mavethee:
```suggestion
			<translation>Nie można zastosować kodów przed uruchomieniem gry.</translation>
```

Keep translation consistent to other translated strings

mavethee (changes requested):

mavethee:
> @mavethee is this right? I added your suggestion, but it seems to have disappeared

I've missclicked things, sorry, weird day 😅 
Current suggestion should be okay.

mavethee:
Also there's still some things missing in general, like Cancel, Apply, Restore Defaults or macOS exclusives, I wonder is it possible to add thoose to not make this PR oneline-ish?
**_Settings menu:_**
CleanShot 2024-09-03 at 18 50 34@2x
**_ShadPS4's macOS menubar_:**
CleanShot 2024-09-03 at 18 50 55@2x

DanielSvoboda:
There are actually some parts of the emulator that haven't been translated, but I think I can add them in a new PR. Because I will have to change the code to implement this, not just add the translation, from what I understand.

DanielSvoboda:
to take advantage of the PR :)
add in settings(settings_dialog.cpp)
Save,Apply,Restore Defaults,Close

mavethee:
@DanielSvoboda Just noticed, I didn't fully fixed my typo in previous PR

Line 254 in pl_PL.ts
Should be `Katalog zainstalowanych gier` not `Katalog zainstalowanych gry`, would be nice if you could correct that as I cannot suggest changes to something you havent touch 😅

DanielSvoboda:
line 254 ?

mavethee:
LGTM 😎

DanielSvoboda:
This is the only one that is pending, that is not being translated. And I couldn't get it to work.
"Directory to install games"
image

mavethee:
Ehh sometimes strings doesn't make sense in target language 😆 
Probably another my fuckup.

During installing updates:

For installed game version should be `Zainstalowana wersja gry:`
I assume source was just `Game installed:`

CleanShot 2024-09-04 at 10 27 10@2x

...aaaand confirmation window is missing translated strings.

PR #757, opened 2024-09-03
video_core: Improve handling of image buffer aliases

raphaelthegreat:
Tldr, these fixes allow Bloodborne character creator to show graphics on main, improve lighting on main game and fix the distorted motion blur

* Fixes the pitch being wrongly set to the width of the image for depth buffer specifically.

* Implements a new method of the buffer cache to validate buffers, SynchronizeBufferFromImage. This is specifically used for texture buffers as those are the only resources observed to rely on this aliasing so far. Bloodborne specifically uses compute shader to copy framebuffers (both depth and color) so with this change it will now work properly and without hacks.

* Use buffer cache for not just macro tiled images but for any type of image. This means if a game uses compute to write texture data it will work and detile shader will automatically use already available GPU buffer.

* Added a max invalidation distance heuristic. This avoid unnecessary copies by avoiding invalidating render targets far away from invalidation base address. If a shader intends to modify an image it is much more likely it will be bound exactly at or close to the image guest address.

Before this is merged the additional validation path needs to be expanded to allow for more copies inside the heuristic range, instead of looking of  1 match

image
image

### DIFF
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -1071,7 +1071,16 @@ ScePthread PThreadPool::Create() {
         }
     }
 
+#ifdef _WIN64
     auto* ret = new PthreadInternal{};
+#else
+    // TODO: Linux specific hack
+    static u8* hint_address = reinterpret_cast<u8*>(0x7FFFFC000ULL);
+    auto* ret = reinterpret_cast<PthreadInternal*>(
+        mmap(hint_address, sizeof(PthreadInternal), PROT_READ | PROT_WRITE,
+             MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0));
+    hint_address += Common::AlignUp(sizeof(PthreadInternal), 4_KB);
+#endif
     ret->is_free = false;
     ret->is_detached = false;
     ret->is_almost_done = false;
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -414,11 +414,6 @@ ScePthreadMutex* createMutex(ScePthreadMutex* addr) {
     if (addr == nullptr || *addr != nullptr) {
         return addr;
     }
-    static std::mutex mutex;
-    std::scoped_lock lk{mutex};
-    if (*addr != nullptr) {
-        return addr;
-    }
     const VAddr vaddr = reinterpret_cast<VAddr>(addr);
     std::string name = fmt::format("mutex{:#x}", vaddr);
     scePthreadMutexInit(addr, nullptr, name.c_str());
@@ -584,8 +579,7 @@ int PS4_SYSV_ABI scePthreadMutexLock(ScePthreadMutex* mutex) {
 }
 
 int PS4_SYSV_ABI scePthreadMutexUnlock(ScePthreadMutex* mutex) {
-    mutex = createMutex(mutex);
-    if (mutex == nullptr) {
+    if (mutex == nullptr || *mutex == nullptr) {
         return SCE_KERNEL_ERROR_EINVAL;
     }
 
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -1071,16 +1071,7 @@ ScePthread PThreadPool::Create() {
         }
     }
 
-#ifdef _WIN64
     auto* ret = new PthreadInternal{};
-#else
-    // TODO: Linux specific hack
-    static u8* hint_address = reinterpret_cast<u8*>(0x7FFFFC000ULL);
-    auto* ret = reinterpret_cast<PthreadInternal*>(
-        mmap(hint_address, sizeof(PthreadInternal), PROT_READ | PROT_WRITE,
-             MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0));
-    hint_address += Common::AlignUp(sizeof(PthreadInternal), 4_KB);
-#endif
     ret->is_free = false;
     ret->is_detached = false;
     ret->is_almost_done = false;
--- a/src/core/libraries/network/net.cpp
+++ b/src/core/libraries/network/net.cpp
@@ -10,7 +10,6 @@
 #include <arpa/inet.h>
 #endif
 
-#include <thread>
 #include "common/assert.h"
 #include "common/logging/log.h"
 #include "core/libraries/error_codes.h"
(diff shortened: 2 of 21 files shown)


rafael-57:
Tested, can confirm it works. Unlike latest BB-Hacks where the default hair colour is wrongly rendered brownish, using this build it's properly rendered as black, just like on console.

image

Crashes while loading up-ingame, but that's probably expected without hacks.
[shad_log.txt](https://github.com/user-attachments/files/16853032/shad_log.txt)
`[Render.Vulkan] <Error> tile_manager.cpp:TryDetile:390: Unsupported tiled image: D32SfloatS8Uint (Depth_MacroTiled)
[Render.Vulkan] <Warning> vk_graphics_pipeline.cpp:GraphicsPipeline:79: Rectangle List primitive type is only supported for embedded VS
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling fs shader 0x5e224cb6 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling vs shader 0x41ae600f 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling fs shader 0xb6e87124 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling vs shader 0x8e3f8dc4 
[Debug] <Critical> vector_memory.cpp:operator():366: Assertion Failed!
Non immediate offset not supported
`

raphaelthegreat:
Yeah the crashing is expected. This PR eliminates the need for the last non-shader skip hack, so now we can focus on properly implementing missing features to go in game without any hacks at all

bigol83:
Red Dead Redemption crashes seconds after going ingame with this PR

```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling fs shader 0x6c25319d 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling vs shader 0x4a7c8442 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling vs shader 0xb2f18afa 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling vs shader 0x65884639 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling fs shader 0x7ed2d200 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling vs shader 0xe3958f87 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling fs shader 0xe598865f 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling vs shader 0x690e1b83 
[Render.Vulkan] <Warning> vk_graphics_pipeline.cpp:GraphicsPipeline:79: Rectangle List primitive type is only supported for embedded VS
[Render.Vulkan] <Warning> vk_graphics_pipeline.cpp:GraphicsPipeline:79: Rectangle List primitive type is only supported for embedded VS
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:311: Compiling fs shader 0x21ed9df3 
[Debug] <Critical> image_info.cpp:ImageInfo:213: Unreachable code!
```

[shad_log.txt](https://github.com/user-attachments/files/16853218/shad_log.txt)


Tomb Raider looks like this while playing intro video 

Cattura

rafael-57:
Edit: Motion blur is also fixed!!!!!

Testing done with df743b9.

This build now goes in-game by itself with the temporary shader skips you've added.

Game is very GPU intensive at 1080p. Screenshot taken with 5800X3D + RTX 3080
image

I forgot to say in my earlier post that with inv-fixes lights render without nuclear explosions, but (I suppose) they will need #370  to render properly, since they're very bright now.

These shader skips are still needed for warping:
0xc0cbc309, 0x77d1c63,0xff7a6d7c, 0xddfbac23

Example of warp freeze due to these shaders:
[inv_fixes_warp_freeze.txt](https://github.com/user-attachments/files/16869092/inv_fixes_warp_freeze.txt)

Let me know if you need a log with validation layers enabled.

With inv-fixes builds blood splatter from hitting an enemy (which should be tessellation) crashes the game, even if merging with bb-hacks.
Here is a log with latest inv-fixes build (no bb-hacks)
[inv_fixes_ingame_crash_hit_enemy.txt](https://github.com/user-attachments/files/16869083/inv_fixes_ingame_crash_hit_enemy.txt)

Thanks! <3

bigol83:
It doesn't go ingame for me, even with the latest commit

[shad_log.txt](https://github.com/user-attachments/files/16869351/shad_log.txt)

```
[Debug] <Critical> vector_memory.cpp:operator():366: Assertion Failed!
Non immediate offset not supported
```

I have Intel i5 12400f rtx 3060 in case there is a difference with AMD gpu or cpu

raphaelthegreat:
@bigol83 I looked into the RDR device loss. It appears it is because of that view being created between BC1 and BC3 formats, which is something that should be fixed by surface work pt3.
The tomb raider problem I cannot reproduce but I think it could be due to different game version

bigol83:
> @bigol83 I looked into the RDR device loss. It appears it is because of that view being created between BC1 and BC3 formats, which is something that should be fixed by surface work pt3. The tomb raider problem I cannot reproduce but I think it could be due to different game version

Ok nice, it's great it will be fixed by that PR. About the Tomb Raider issue i have 1.00 version. Let me try to update and see if it goes away.

bigol83:
I updated Tomb Raider to latest patch 1.04 with latest PR commit and this is how it looks

Cattura
Cattura2
Cattura3

Subtitles keep being shown even when they shouldn't be there and ingame there is that trail effect when opening the pause menu. I checked and this doesn't happen on main branch build.

Dark Souls has a regression too 

Cattura

Final Fantasy VII main menu duplicated cursor

Cattura

bigol83:
With latest commit Bloodborne reaches ingame for me too.

raphaelthegreat:
@bigol83 Is the Dark Souls bug consistently reproducible or random?

bigol83:
> @bigol83 Is the Dark Souls bug consistently reproducible or random?

It is consistent, I took other camera angle screenshots, maybe this will help you 

Cattura

Cattura2

bigol83:
With latest commits Red Dead Redemption crash has been fixed but there are some random device lost crashes

Looks like graphics regressed in cutscenes 

Cattura


Dark Souls is still mostly green as before
Now Resident Evil 4 shows nothing, there is only audio

hgh32:
Bloodborne now booted to the title menu on the latest commit, and after selecting New Game I got that error again:

`[Debug] <Critical> vk_compute_pipeline.cpp:operator():182: Assertion Failed!`

[shad_log.txt](https://github.com/user-attachments/files/16883999/shad_log.txt)

gandalfthewhite19890404:
> Now Resident Evil 4 shows nothing, there is only audio

Because of merging surface-related PR. Do not mix issues

PR #761, opened 2024-09-03
[libSceAudio3d] Initialize Audio3d Library

lzardy:
This is simple initialization PR for beginning work on `sceAudio3d` library.


StevenMiller123:
Last I checked, libSceAudio3d is an unsafe LLE, as it tries to open a hardware device which does not exist. I could be wrong though.

lzardy:
> Last I checked, libSceAudio3d is an unsafe LLE, as it tries to open a hardware device which does not exist. I could be wrong though.

Yes I was curious if this LLE should be ignored, and only HLE used as it still have usage as audio output. This behavior will have to redirect to our hardware.

lzardy:
This should be good for merge now.

georgemoralis:
at least add the parameters in logging for each function , so we can have a meanfull output ;D

lzardy:
> at least add the parameters in logging for each function , so we can have a meanfull output ;D

This seem unnecessary as the debug will show for developers these parameter values. It will also be removed when the function is done, so this is unnecessary work for little. It is only needed to know the function is called, so we can know which to implement.

lzardy:
> at least add the parameters in logging for each function , so we can have a meanfull output ;D

Ok I added parameters to logs, should be good for merge.

PR #762, opened 2024-09-03
vulkan: Emulate depth clip control when extension is not available.

squidbus:
When the depth clip control extension is not available, emulate in the shader by reducing from [-1, 1] to [0, 1].

Fixes likely a number of rendering issues on MoltenVK. For example, Persona Dancing games will now show text and other UI graphics correctly.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
@@ -69,28 +69,32 @@ void GatherVertexOutputs(Shader::VertexRuntimeInfo& info,
                    : (ctl.IsCullDistEnabled(7) ? VsOutput::CullDist7 : VsOutput::None));
 }
 
-Shader::RuntimeInfo BuildRuntimeInfo(Shader::Stage stage, const GraphicsPipelineKey& key,
-                                     const AmdGpu::Liverpool::Regs& regs) {
+Shader::RuntimeInfo PipelineCache::BuildRuntimeInfo(Shader::Stage stage) {
     auto info = Shader::RuntimeInfo{stage};
+    const auto& regs = liverpool->regs;
     switch (stage) {
     case Shader::Stage::Vertex: {
         info.num_user_data = regs.vs_program.settings.num_user_regs;
         info.num_input_vgprs = regs.vs_program.settings.vgpr_comp_cnt;
         GatherVertexOutputs(info.vs_info, regs.vs_output_control);
+        info.vs_info.emulate_depth_negative_one_to_one =
+            !instance.IsDepthClipControlSupported() &&
+            regs.clipper_control.clip_space == Liverpool::ClipSpace::MinusWToW;
         break;
     }
     case Shader::Stage::Fragment: {
         info.num_user_data = regs.ps_program.settings.num_user_regs;
-        std::ranges::transform(key.mrt_swizzles, info.fs_info.mrt_swizzles.begin(),
+        std::ranges::transform(graphics_key.mrt_swizzles, info.fs_info.mrt_swizzles.begin(),
                                [](Liverpool::ColorBuffer::SwapMode mode) {
                                    return static_cast<Shader::MrtSwizzle>(mode);
                                });
+        const auto& ps_inputs = regs.ps_inputs;
         for (u32 i = 0; i < regs.num_interp; i++) {
             info.fs_info.inputs.push_back({
-                .param_index = u8(regs.ps_inputs[i].input_offset.Value()),
-                .is_default = bool(regs.ps_inputs[i].use_default),
-                .is_flat = bool(regs.ps_inputs[i].flat_shade),
-                .default_value = u8(regs.ps_inputs[i].default_value),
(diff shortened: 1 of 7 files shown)


src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
```diff
@@ -69,14 +69,18 @@ void GatherVertexOutputs(Shader::VertexRuntimeInfo& info,
                    : (ctl.IsCullDistEnabled(7) ? VsOutput::CullDist7 : VsOutput::None));
 }
 
-Shader::RuntimeInfo BuildRuntimeInfo(Shader::Stage stage, const GraphicsPipelineKey& key,
-                                     const AmdGpu::Liverpool::Regs& regs) {
+Shader::RuntimeInfo PipelineCache::BuildRuntimeInfo(Shader::Stage stage,
+                                                    const GraphicsPipelineKey& key,
```
raphaelthegreat: You can remove the GraphicsPipelineKey param now since its a member, its available as graphics_key

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -219,7 +219,16 @@ void EmitSetAttribute(EmitContext& ctx, IR::Attribute attr, Id value, u32 elemen
         return;
     }
     const Id pointer{OutputAttrPointer(ctx, attr, element)};
-    ctx.OpStore(pointer, ctx.OpBitcast(ctx.F32[1], value));
+    const Id valueF32{ctx.OpBitcast(ctx.F32[1], value)};
```
raphaelthegreat:
```suggestion
    const Id value_f32{ctx.OpBitcast(ctx.F32[1], value)};
```

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -219,7 +219,16 @@ void EmitSetAttribute(EmitContext& ctx, IR::Attribute attr, Id value, u32 elemen
...
+    if (ctx.runtime_info.vs_info.emulate_depth_negative_one_to_one &&
+        attr == IR::Attribute::Position0 && element == 2) {
+        // Adjust the Z value in the shader from [-1, 1] to [0, 1].
+        const Id valueZeroToTwo{ctx.OpFAdd(ctx.F32[1], valueF32, ctx.ConstF32(1.f))};
```
raphaelthegreat: I feel this assumes that the written z value is in [-1, 1] range. But games can output any z value together, where z / w is actually what is inside [-1, 1]

src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
squidbus: Done, I also removed `regs` since it's also a reference to member state.

raphaelthegreat: I think it would good to have an ```const auto& regs = liverpool->regs;``` on top since this change makes the code more verbose than before

squidbus: Also done.

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
squidbus: Hmm, is there any better way to account for this? Since it outputs both `z` and `w` as separate components not sure how I can transform `z / w` as a whole

squidbus: Actually since the clip spaces are `[0, w]` and `[-w, w]` couldn't I just transform `z` with something like:
```
z = (z + w) / 2
```

raphaelthegreat: Yeah I feel this transformation will work better

raphaelthegreat: You can also implement this as an epilogue to the shader so you have access to both z and w. There is EmitEpilogue IR function specifically for this purpose. You can add check there and access whole output_position vector
```
void ConvertDepthMode(EmitContext& ctx) {
    const Id type{ctx.F32[1]};
    const Id position{ctx.OpLoad(ctx.F32[4], ctx.output_position)};
    const Id z{ctx.OpCompositeExtract(type, position, 2u)};
    const Id w{ctx.OpCompositeExtract(type, position, 3u)};
    const Id screen_depth{ctx.OpFMul(type, ctx.OpFAdd(type, z, w), ctx.Constant(type, 0.5f))};
    const Id vector{ctx.OpCompositeInsert(ctx.F32[4], screen_depth, position, 2u)};
    ctx.OpStore(ctx.output_position, vector);
}
```

squidbus: Yep that works, thanks. Had something like that written up real quick but didn't know there was a dedicated epilogue instruction, that's useful.

raphaelthegreat (approved):

PR #763, opened 2024-09-03
Update renderdoc_app.h for larger MacOS Compatibility

N7z:


raphaelthegreat:
Any updates to renderdoc header should be made into renderdoc upstream

squidbus:
Can you explain what this is targeting? What macOS are you using that doesn't set `__APPLE__`?

N7z:
> Can you explain what this is targeting? What macOS are you using that doesn't set `__APPLE__`?

Added __MACH__ check to ensure compatibility with systems using the Mach kernel, such as some NeXTSTEP-derived systems like OPENSTEP and Rhapsody.

squidbus:
Well like was mentioned previously if you want to make a change to renderdoc you should go to the upstream, then pull from upstream into here.

PR #769, opened 2024-09-04
Mac OS ARM Native reaches ELF entry

niko1point0:
Reaches entry and then stops, the recompiler comes next


CMakeLists.txt
```diff
@@ -657,8 +657,11 @@ if (APPLE)
       target_link_libraries(shadps4 PRIVATE ${MOLTENVK})
   endif()
 
-  # Reserve system-managed memory space.
-  target_link_options(shadps4 PRIVATE -Wl,-no_pie,-no_fixup_chains,-no_huge,-pagezero_size,0x4000,-segaddr,TCB_SPACE,0x4000,-segaddr,GUEST_SYSTEM,0x400000,-image_base,0x20000000000)
+  # Using x86_64 (ARM by default if not set)
+  if (CMAKE_OSX_ARCHITECTURES)
```
squidbus: You will need to check more than just `CMAKE_OSX_ARCHITECTURES` because it may be compiled natively on an Intel Mac. Simplest option I think would be to move the logic added to `externals/CMakeList.txt` for ffmpeg to this file and use the `ARCHITECTURE` variable to see if `ARCHITECTURE STREQUAL "x86_64"`

src/common/assert.cpp
```diff
@@ -4,7 +4,11 @@
 #include "common/assert.h"
 #include "common/logging/backend.h"
 
+#ifdef __x86_64__
 #define Crash() __asm__ __volatile__("int $3")
+#elif __aarch64__
+#define Crash() ;
```
squidbus: This should probably have an actual implementation.

src/common/rdtsc.cpp
```diff
@@ -55,6 +58,12 @@ u64 EstimateRDTSCFrequency() {
     const u64 tsc_diff = tsc_end - tsc_start;
     const u64 tsc_freq = MultiplyAndDivide64(tsc_diff, 1000000000ULL, end_time - start_time);
     return RoundToNearest<100'000>(tsc_freq);
+
+    #else
+
+    return 0;
```
squidbus: Should have an alternate implementation or at least a TODO note.

src/common/rdtsc.h
```diff
@@ -32,6 +36,13 @@ static inline u64 FencedRDTSC() {
 }
 #endif
 
+#else
+// ARM
+static inline u64 FencedRDTSC() {
+    return 0;
```
squidbus, line 45: Same here for actual implementation or TODO.

src/core/address_space.cpp
```diff
@@ -382,13 +393,16 @@ struct AddressSpace::Impl {
...
-        ASSERT_MSG(ret != MAP_FAILED, "mmap failed: {}", strerror(errno));
-        return ret;
+        if (phys_addr != -1) {
+            const int handle = fd == -1 ? backing_fd : fd;
+            void* ret = mmap(reinterpret_cast<void*>(virtual_addr), size, prot, MAP_FIXED | MAP_SHARED,
+                             handle, phys_addr);
+            ASSERT_MSG(ret != MAP_FAILED, "mmap failed: {}", strerror(errno));
+        } else {
+            int ret = mprotect(reinterpret_cast<void*>(virtual_addr), size, prot);
+            ASSERT_MSG(ret == 0, "mprotect failed: {}", strerror(errno));
```
squidbus, line 403: This and the below in `Unmap` will need a closer look to make sure the behavior is correct, I put it together for you quick to get things going but I'm not sure it's good enough to go upstream.

src/core/linker.cpp
```diff
@@ -292,7 +306,9 @@ bool Linker::Resolve(const std::string& name, Loader::SymbolType sym_type, Modul
     return false;
 }
 
-void* Linker::TlsGetAddr(u64 module_index, u64 offset) {
+void* Linker::TlsGetAddr(u64 module_index, u64 offset) 
+{
+#ifdef __x86_64__
```
squidbus: You should probably just stub `GetTcbBase()` for now if anything, not the whole thing here.

src/core/linker.cpp
```diff
@@ -107,7 +119,9 @@ void Linker::Execute() {
         }
     }
 
+    #ifdef __x86_64__
     SetTcbBase(nullptr);
```
squidbus: Instead of stubbing out `SetTcbBase` calls it should probably just have an stub function to call for now, it's going to need to be set up somehow for ARM translated code to access. Ditto for the other cases of this.

squidbus (commented):
Overall this is good enough to get to ARM entry but a bit incomplete and untested, particularly the address space changes could have an impact or might need more modifications. I think you should really make some initial progress with getting an ARM translator running and then PR this stuff once its verified to be a sound base, just to validate that everything here is correct.

CMakeLists.txt
niko1point0: Done

src/core/address_space.cpp
niko1point0: Actually no, this part was @squidbus who has more experience with the memory, so I trust that it's correct. Feel free to test though

[Edit] I just realized it was you that said the message, this was in the mem.patch file you gave me

src/common/assert.cpp
niko1point0: Done

src/core/address_space.cpp
squidbus: That's... me though.

src/core/linker.cpp
niko1point0: Done

src/core/linker.cpp
niko1point0: Done

src/common/rdtsc.cpp
niko1point0: Done

src/common/rdtsc.h
niko1point0: Done

Hermiten:
Discontinued ?

squidbus:
This can probably be closed, some of the changes are already upstreamed and others are outdated and not that far in.

PR #778, opened 2024-09-04
video_core/renderer_vulkan: Ignore unsupported shader stages

polybiusproxy:
Avoids PSPHD games from compiling geometry shaders.


georgemoralis:
but that is a hack right , at least we probably can make an option for this

polybiusproxy:
> but that is a hack right , at least we probably can make an option for this

It's not a hack, unsupported stages are not supposed to be translated.

georgemoralis:
ignoring them will not fix the issue ;P

raphaelthegreat (approved):

polybiusproxy:
> ignoring them will not fix the issue ;P

Better that than making it compile them :p 
However I missed something on this PR and that is logging that there's an unsupported stage on the pipeline, like it did before.

PR #780, opened 2024-09-04
[09/04/2024] Added Discord Presence Integration.

morkev:
**Discord displays "In ShadPS4" along with elapsed time when a user is executing shadps4.exe in their computer.**

The presence displays the current status of the emulator (e.g., whether shadps4.exe is running) and updates it in real-time within the Discord client. All necessary files have been moved to a presence directory to keep it organized.

Changes:
- Added presence/discord.h, presence/discord.cpp, and presence/discord_rpc.h for handling Discord RPC integration.
- In testing, I modified ```main.cpp``` to initialize and run Discord Presence alongside the emulator as shown below:
```cpp
#include <fmt/core.h>
#include "emulator.h"
#include "presence/discord.h"

int main(int argc, char* argv[]) {
    if (argc == 1) {
        fmt::print("Usage: {} <elf or eboot.bin path>\n", argv[0]);
        return -1;
    }

    Discord discord("XXXXXXXXXXXXXXXXXXX");
    discord.Initialize();

    Core::Emulator emulator;
    emulator.Run(argv[1]);

    discord.Run();
    discord.Shutdown();

    return 0;
}
```

Let's talk the 19 digit Discord API key "XXXXXXXXXXXXXXXXXXX": I used my custom key and it worked, so the owner of the repository or someone with higher privileges needs to go to the Discord Developer Portal and get a key, I could otherwise configure this. That being said, all the functionality is there, and we are 1-key away. As for ```main.cpp``` I didn't want to mess with that directly in this merge request, therefore I am just showing what I did.

All files conform to the project's style guide:
- Functions and classes follow PascalCase.
- Variables follow lower_case_underscored naming.
- Line width limited to 100 characters.
- Used C++-style casts where needed (e.g., static_cast).
- Proper namespace usage.
- 4-space indentation with no tabs.

### DIFF
--- a/presence/discord.cpp
+++ b/presence/discord.cpp
@@ -1,0 +1,151 @@
+#include "discord.h"
+#include "discord_rpc.h"
+#include <iostream>
+#include <chrono>
+#include <thread>
+#include <filesystem>
+#include <fstream>
+
+/**
+ * @brief Constructor for the Discord class.
+ * @author Kevin Mora (morkev)
+ *
+ * Initializes the Discord class with the provided launch code (Discord application ID)
+ * and sets the application_running_ flag to false.
+ *
+ * @param launch_code The Discord application client ID.
+ */
+Discord::Discord(const std::string &launch_code)
+        : launch_code_(launch_code), application_running_(false), time_of_start_(0) {}
+
+/**
+ * @brief Initializes the Discord RPC connection.
+ *
+ * Sets up the event handlers for Discord Rich Presence and connects to Discord
+ * using the provided launch code.
+ */
+void Discord::Initialize() {
+    DiscordEventHandlers handlers{};
+    Discord_Initialize(launch_code_.c_str(), &handlers, 1, nullptr);
+}
+
+/**
+ * @brief Shuts down the Discord RPC connection.
+ *
+ * Terminates the connection to Discord and cleans up any active presence.
+ */
+void Discord::Shutdown() {
+    Discord_Shutdown();
+}
+
+/**
+ * @brief Checks if a process with the given name is running on the system.
+ *
+ * This function iterates over all running processes to check if a process
+ * named `process_name` (e.g., `shadps4.exe`) is running.
+ *
+ * @param process_name The name of the process to check for.
+ * @return true if the process is running, false otherwise.
+ */
+bool Discord::IsProcessRunning(const std::string &process_name) {
+    for (const auto &entry : std::filesystem::directory_iterator("/proc")) {
+        try {
+            std::string pid = entry.path().filename();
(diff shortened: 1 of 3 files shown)


presence/discord.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+    DiscordRichPresence presence{};
+    presence.state = state.c_str();
+    presence.large_image_key = large_image.c_str();
+    presence.large_image_text = large_text.c_str();
+    presence.start_timestamp = time_of_start_;
+
+    if (state != "Unknown") {
+        Discord_UpdatePresence(&presence);
+        std::cout << "Discord Rich Presence updated.\n";
```
polybiusproxy, line 110: Please use our log implementation for printing. The same goes for the rest of `std::cout` uses.

presence/discord.h
```diff
@@ -0,0 +1,64 @@
+#ifndef DISCORD_H
+#define DISCORD_H
```
polybiusproxy, line 2: I'd prefer the use of `#pragma once`. Same goes for the other header.

presence/discord.cpp
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * @param state Reference to the string where the current state will be stored (e.g., "In ShadPS4").
+ * @param large_image Reference to the string where the large image key will be stored.
+ * @param large_text Reference to the string where the large image's tooltip text will be stored.
+ */
+void Discord::GetData(std::string &state, std::string &large_image, std::string &large_text) {
+    if (IsProcessRunning("shadps4.exe")) {
+        state = "In ShadPS4";
+        large_image = "shadps4";
+        large_text = "ShadPS4";
```
polybiusproxy, line 83:
```diff
-        state = "In ShadPS4";
-        large_image = "shadps4";
-        large_text = "ShadPS4";
+        state = "In shadPS4";
+        large_image = "shadPS4";
+        large_text = "shadPS4";
```

polybiusproxy (changes requested):
I have quite a bunch of issues with this PR that should be resolved before merging. First of all, the builds are also failing the Reuse CI as there is no copyright header on the files so please fix that.

It's also worth noting that the discord-rpc submodule was removed as it went unused, so please add it back with this PR.

Finally, I have some issues with the actual implementation. You made the code check if the emulator process is running, and I don't understand why. If this is running from the emulator, rest assured the process is always running. You seem to have written this to run as an external program - ideally we would want this to be bundled with the emulator. For the main function you provided, looks good on a first glance but it's worth noting that `Emulator::Run` won't return so `Discord::Run` will likely never execute.

nogitsune-youkai:
It's pretty sad to see that a lot of people simply delete their repos when their PR is not approved and changes required.

PR #387, opened 2024-08-08
Implemented sceKernelMTypeProtect and sceKernelMProtect

menaman123:
Implemented sceKernelMTypeProtect and sceKernelMProtect for CUSA07402 to go past sceKernelBatchMap2. However, the game does not progress forward due to Invalid PM4 type error.
Screenshot_2024-08-08_165757

PR #782, opened 2024-09-04
Update sceBatchMap2

StevenMiller123:
Some of the sceBatchMap2 changes in https://github.com/shadps4-emu/shadPS4/pull/602 were reverted when https://github.com/shadps4-emu/shadPS4/pull/387 was merged. If there are any other changes I should make here, let me know.


georgemoralis:
good catch ;)

PR #809, opened 2024-09-06
Standardize game_data directory for game metadata.

squidbus:
Various references to the `game_data` folder are currently hardcoded throughout the GUI code for trophies and background images. As a result if that does not match the determined user directory in `path_util`, files will be placed in the wrong place.

This PR makes sure they are all going through `GetUserDir` so they have the right base directory, introducing a new path type for game metadata files like these.


georgemoralis:
ok that seems reasonable :D

PR #814, opened 2024-09-06
shader_recompiler: Implement data share append and consume operations

raphaelthegreat:
Allows Bloodborne to boot on main without any hacks. Also fixes Gravity Rush remastered crashing before title screen (though still wont go in-game due to a problematic shader).

Also implements remaining MRT swap modes so that crash should go away


rafael-57:
I can confirm getting in-game works with this build, out of the box!

Warping also works sometimes, but it's very unstable and not just because of the usual deadlocks. Sometimes warps hang while copmpiling a shader and sometimes that same shader gets compiled without issue. See logs:
[logs.zip](https://github.com/user-attachments/files/16909876/logs.zip)

Corruptions are still there
image

bigol83:
I still have crashes when enemies attack 

this is the log 
[shad_log.txt](https://github.com/user-attachments/files/16912145/shad_log.txt)

and i have Intel i5 12400f rtx 3060

illyazru:
Gravity Rush remastered doesn't boot anymore in the latest main
[shad_log.txt](https://github.com/user-attachments/files/16913215/shad_log.txt)

PR #815, opened 2024-09-06
Redefined ffmpeg's av_err2str macro to be c++ friendly

zmckevitt:
When compiling on Ubuntu 24.04, the definition of ```av_err2str``` in ```externals/ffmpeg-core/include/libavutil/error.h``` throws a compiler error when attempting to use the macro in a c++ environment (namely ```src/core/libraries/avplayer/avplayer_source.cpp```). 

This fix redefines the macro in a 'c++ friendly' manner without modifying the external library and generates no additional compiler warnings in my experience using g++13 or clang18.

More info: https://github.com/joncampbell123/composite-video-simulator/issues/5

### DIFF
--- a/documents/building-linux.md
+++ b/documents/building-linux.md
@@ -9,7 +9,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 #### Debian & Ubuntu
 ```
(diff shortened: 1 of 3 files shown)


zmckevitt:
I added some quality of life changes for building from scratch on Ubuntu systems. In addition to fixing the compiler error in modern gcc, this PR also displays both *.PKG and *.pkg files when importing games through the QT gui and updates the linux build documentation to include missing qt dependencies and options for building with QT.

PR #820, opened 2024-09-06
Typo fix

Pipi86:
I found and corrected a little typo.


LeDragoX:
If you don't want to break the building steps, you'll need to check where else `SignalFence` is used.
Like here:
https://github.com/shadps4-emu/shadPS4/blob/81f7c830be8b4595ade5aede5de0c4dcf8b2c6c2/src/video_core/amdgpu/pm4_cmds.h#L484
Change the **line 484** from the same file.

Pipi86:
@psucien is this correct? if so, it's ready for merge.

PR #767, opened 2024-09-04
Message Dialog library

viniciuslrangel:
~- based on the #598 branch, gonna rebase after merge~

> Check discord homebrew-downloads channel for a test binary

Gamepad as first-class

### Msg dialog
- [x] State
- [x] Basic text
- [x] Buttons
- [x] Progress bar
- [x] System error

~### Save Dialog~ I'll split it to another PR

## Demo
image
image
image
image

PR #824, opened 2024-09-07
Rewrite Save Data & Impl Save Data Dialog

viniciuslrangel:
~- based on the #767 branch, gonna rebase after merge~

## Extra
- [x] SFO file handling
        Better support for SFO binary payload & allow to encode (write) back to file
- [x] ImGui texture/image handling
         This uses a background thread to read & decode png then queues to upload in next frame

## Save Data
- [x] Save Data Library
- [x] Save Data Backup
- [x] Save Data Memory

## Save Data Dialog
- [x] Save Data Dialog listing
- [x] Save Data Dialog images
- [x] Save Data Dialog user msg
- [x] Save Data Dialog system msg
- [x] Save Data Dialog error
- [x] Save Data Dialog progress bar


viniciuslrangel:
No more corrupted save data due to crashes

StevenMiller123:
Kingdom Hearts Birth By Sleep no longer detects my previously dumped saves and crashes when trying to open the load game menu. There is no logged error, as the problem seems to be an exception of some kind

[shad_log.txt](https://github.com/user-attachments/files/17068996/shad_log.txt)

viniciuslrangel:
> Kingdom Hearts Birth By Sleep no longer detects my previously dumped saves and crashes when trying to open the load game menu. There is no logged error, as the problem seems to be an exception of some kind
> 
> [shad_log.txt](https://github.com/user-attachments/files/17068996/shad_log.txt)

Can you set the log filter to `Lib.SaveData:Debug Lib.SaveDataDialog:Debug` and make it sync? It must output some info in the log

The save will not show up, the format was changed to be the same as the real saves (so it must be capable of importing existing saves). But you can create a new save and copy the files

ElBread3:
> > Kingdom Hearts Birth By Sleep no longer detects my previously dumped saves and crashes when trying to open the load game menu. There is no logged error, as the problem seems to be an exception of some kind
> > [shad_log.txt](https://github.com/user-attachments/files/17068996/shad_log.txt)
> 
> Can you set the log filter to `Lib.SaveData:Debug Lib.SaveDataDialog:Debug` and make it sync? It must output some info in the log
> 
> The save will not show up, the format was changed to be the same as the real saves (so it must be capable of importing existing saves). But you can create a new save and copy the files

Just saw this as I went to test PDFT, it crashed at these three calls:

[Lib.SaveDataDialog] <Debug> savedatadialog.cpp:sceSaveDataDialogInitialize:50: called
[Lib.SaveData] <Debug> savedata.cpp:sceSaveDataDirNameSearch:753: called
[Lib.SaveDataDialog] <Debug> savedatadialog.cpp:sceSaveDataDialogOpen:79: called param->mode: LIST

But yeah PR pretty fkd rn, will test again tomorrow

StevenMiller123:
@viniciuslrangel The following lines are printed right before the crash (when choosing to load a save). 
```
[Lib.SaveDataDialog] <Debug> savedatadialog.cpp:sceSaveDataDialogInitialize:50: called
[Lib.SaveDataDialog] <Debug> savedatadialog.cpp:sceSaveDataDialogOpen:79: called param->mode: SYSTEM_MSG
```

[shad_log.txt](https://github.com/user-attachments/files/17069104/shad_log.txt)

StevenMiller123:
Can confirm, the saves not reading issue was just from them being modified to properly read in shad, as copying my original dumps of the saves fixed that issue. You should probably add some proper way to convert the saves though, as there will probably be tons of people complaining about deleted saves if you dont.

That said, fixing my saves did not fix the crashing issue.

viniciuslrangel:
> Can confirm, the saves not reading issue was just from them being modified to properly read in shad, as copying my original dumps of the saves fixed that issue. You should probably add some proper way to convert the saves though, as there will probably be tons of people complaining about deleted saves if you dont.
> 
> That said, fixing my saves did not fix the crashing issue. 

A just pushed a quick fix that I imagine is the problem. Can you try it?

I'll prob just write a guide how to migrate. (If you want to help 😁)

StevenMiller123:
Screenshot 2024-09-19 221147
@viniciuslrangel The new push did not fix the crash. I did get the build locally compiled though, and I see that item->titleId is null, causing the crash.

Screenshot 2024-09-19 221304

StevenMiller123:
@viniciuslrangel I think the problem is that you check for item->titleId being null AFTER storing the data from it. Makes sense why there'd be an issue with that.

viniciuslrangel:
> @viniciuslrangel I think the problem is that you check for item->titleId being null AFTER storing the data from it. Makes sense why there'd be an issue with that.

The fix is right below it, I forgot to delete that line 🤦

StevenMiller123:
Screenshot 2024-09-19 221832
Removing that line fixed the crash, but it seems that it's failing to find the save data when I go to load a save (and yes, this is with proper save files this time).

[shad_log.txt](https://github.com/user-attachments/files/17069356/shad_log.txt)

StevenMiller123:
Screenshot 2024-09-19 222827
@viniciuslrangel The issue is with how sceSaveDataDirNameSearch filters saves. The filter is removing all the saves from the list it finds.
Here's the game's search condition (or whatever cond means in this context), in case it helps
Screenshot 2024-09-19 222954

viniciuslrangel:
> Screenshot 2024-09-19 222827 @viniciuslrangel The issue is with how sceSaveDataDirNameSearch filters saves. The filter is removing all the saves from the list it finds. Here's the game's search condition (or whatever cond means in this context), in case it helps Screenshot 2024-09-19 222954

Can you ping me at Discord?

src/core/libraries/save_data/dialog/savedatadialog_ui.cpp
```diff
@@ -0,0 +1,787 @@
...
+        if (dir_name.empty()) {
+            continue;
+        }
+
+        auto dir_path = SaveInstance::MakeDirSavePath(user_id, title_id, dir_name);
+
+        PSF param_sfo;
+        param_sfo.Open(dir_path / "sce_sys" / "param.sfo");
+
+        auto utc_time = std::chrono::file_clock::to_utc(param_sfo.GetLastWrite());
```
baggins183: I'm getting a compiler error on linux, gcc 14.2.1
I can compile by changing to_utc to to_sys

This https://en.cppreference.com/w/cpp/chrono/file_clock
says
> file_clock provides exactly one of the following two pairs of static member functions:
>     to_utc and from_utc; or
>     to_sys and from_sys. 

seems weird

viniciuslrangel: Funny one. In Msvc there is no to_sys 😆

StevenMiller123:
As discussed on Discord, Kingdom Hearts Birth by Sleep and Kingdom Hearts 0.2 Birth by Sleep (note: is a UE4 game) are both able to properly detect save dumps from real HW. I've also verified that KH BBS is able to create new saves and load those successfully. 

Lastly, I've tested Hatsune Miku Project Diva Future Tone on the latest commit, and it's able to create saves and load them now too, though I haven't done any thorough testing with it yet.

src/core/libraries/save_data/dialog/savedatadialog_ui.cpp
maxphilippov: > I'm getting a compiler error on linux, gcc 14.2.1 I can compile by changing to_utc to to_sys
> 
> This https://en.cppreference.com/w/cpp/chrono/file_clock says
> 
> > file_clock provides exactly one of the following two pairs of static member functions:
> > to_utc and from_utc; or
> > to_sys and from_sys.
> 
> seems weird

I didn't know where to post this. But I got same error on Windows while building via MSYS MINGW64.
image
So I guess wrapping it into #ifdef _WIN32 isn't enough?

viniciuslrangel: Yeah, it would need to check if it is MinGW too
`#if defined(_WIN32) && !defined(__GNUC__) && !defined(__MINGW32__) && !defined(__MINGW64__)`

maxphilippov: Should I create an issue? I can make a PR but I'm no expert in building C++, so I'm literally not aware of what's the sufficient set of defines is.

viniciuslrangel: Can you check if this change solves your issue? If you don't want to PR, I can publish in my next PR

maxphilippov: image
Yeah, that one passed, now it can't find "visibility.h". cmake config issues? Should I just send it to Discord to investigate further? Cause it seems like it might be a chain of fixes

StevenMiller123: That one's a separate issue described here: https://github.com/shadps4-emu/shadPS4/issues/1195

hgh32:
Hi, can you take a look at this? https://github.com/shadps4-emu/shadPS4/issues/1517

PR #837, opened 2024-09-07
video_core: Add fallback path for pipelines with more than 32 bindings

raphaelthegreat:
Fixes crashes in games that bind many resources. This is done by maintaining a classic descriptor pool and allocating sets in batches to update them. It is expected that only a small subset of draws will hit this path, thus the update call should be fine perf wise. If there are games that consistently bind that many resources, we can explore options to batch updates.

### DIFF
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -2155,6 +2155,7 @@ int PS4_SYSV_ABI sceGnmSubmitCommandBuffersForWorkload() {
 
 int PS4_SYSV_ABI sceGnmSubmitDone() {
     LOG_DEBUG(Lib_GnmDriver, "called");
+    WaitGpuIdle();
     if (!liverpool->IsGpuIdle()) {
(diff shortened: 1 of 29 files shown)


bigol83:
Bloodborne crashes with this log after character creation

```
[Debug] <Critical> page_manager.cpp:operator():46: Assertion Failed!
Region protection failed
```

[shad_log.txt](https://github.com/user-attachments/files/16919708/shad_log.txt)

StevenMiller123:
Causes graphical artifacts in Kingdom Hearts Birth By Sleep (from the 1.5 + 2.5 remix). 
Before:
Screenshot 2024-09-07 143558

After:
Screenshot 2024-09-07 143731

bigol83:
Dark Souls remastered doesn't crash anymore reaching Undead Burg

Cattura
Cattura2
Cattura3

KITT1995:
With this commit Gravity Rush (CUSA01113) goes ingame but with main character not showing (same as US version) and Shantae Half-Genie Hero (CUSA03920) boots to intro but crash after that.

Issue #845, opened 2024-09-08
[Linux] Failing to compile on Arch after imgui implementation


Caliel666:
using 
`cmake -S . -B build/ -DENABLE_QT_GUI=ON`
`cmake --build . --parallel$(nproc)`

```shell
/shadPS4/src/imgui/renderer/imgui_core.cpp: In function ‘void ImGui::Core::Render(const vk::CommandBuffer&, Vulkan::Frame*)’:
/shadPS4/src/imgui/renderer/imgui_core.cpp:189:5: error: ambiguous overload for ‘operator=’ (operand types are ‘vk::Rect2D’ and ‘<brace-enclosed initializer list>’)
  189 |     };
      |     ^
In file included from /usr/include/vulkan/vulkan.hpp:8579,
                 from /shadPS4/src/video_core/renderer_vulkan/vk_common.h:17,
                 from /shadPS4/src/video_core/renderer_vulkan/vk_platform.h:11,
                 from /shadPS4/src/video_core/renderer_vulkan/vk_instance.h:9,
                 from /shadPS4/src/imgui/renderer/imgui_core.h:6,
                 from /shadPS4/src/imgui/renderer/imgui_core.cpp:9:
/usr/include/vulkan/vulkan_structs.hpp:8142:14: note: candidate: ‘vk::Rect2D& vk::Rect2D::operator=(const VkRect2D&)’
 8142 |     Rect2D & operator=( VkRect2D const & rhs ) VULKAN_HPP_NOEXCEPT
      |              ^~~~~~~~
/usr/include/vulkan/vulkan_structs.hpp:8124:10: note: candidate: ‘constexpr vk::Rect2D& vk::Rect2D::operator=(const vk::Rect2D&)’
 8124 |   struct Rect2D
      |          ^~~~~~
/usr/include/vulkan/vulkan_structs.hpp:8124:10: note: candidate: ‘constexpr vk::Rect2D& vk::Rect2D::operator=(vk::Rect2D&&)’
make[2]: *** [CMakeFiles/shadps4.dir/build.make:289: CMakeFiles/shadps4.dir/src/imgui/renderer/imgui_core.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:247: CMakeFiles/shadps4.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```
i made sure to update my system and install imgui from the AUR just in case

[...]

PR #846, opened 2024-09-08
Fix compilation on linux

N00byKing:
~~Also, add compile_commands.json to gitignore.~~

~~This file is generated when using the `CMAKE_EXPORT_COMPILE_COMMANDS` flag. It is used by clangd to search header files, and needs to be in the project root (tho I just symlink it).~~

Closes #845

PR #849, opened 2024-09-08
Fix building imgui on Linux

LNDF:
This fixex building after the introduction of ImGui on Linux (#845)

This PR also fixes GCC warning about a missimg newline before EOF in imgui_config.h.


roamic:
#846 had the same fix

georgemoralis:
done in #846

PR #852, opened 2024-09-09
cpu_patches: Patch just-in-time using signal handlers.

squidbus:
Reworks the patching system to patch on demand using a signal handler, instead of completely ahead of time.

The current system works on the assumption that just scanning down segments will generally work out without becoming misaligned by data or patching data as if it were code. In practice this generally works, but there are some games where it does not and code that needs patching will be missed.

Instead, we can discover instructions that need patching by actually running the code and trapping signals on bad instructions. I've introduced a platform-common way of registering multiple handlers for a signal, captured using either vectored exception handlers on Windows or signal handlers on POSIX systems, and wired up both this and the existing page manager to it.

One caveat is that when running under Rosetta 2 on macOS, for whatever reason patching an instruction at the beginning of a page does not work. It just acts like the instruction was not patched and raises SIGILL again. I've tried a few things to get around this but couldn't figure it out, so I left a workaround to patch instructions at the beginning of each code page ahead of time on macOS.

Tested this by running through a few games on macOS that need these patches. Needs testing on other OSes since this impacts TCB access patches and GPU page invalidation.

One game fixed by this is CUSA02799, which currently crashes on launch on ARM64 macOS due to a missed BMI1 instruction patch.

### DIFF
--- a/src/video_core/page_manager.cpp
+++ b/src/video_core/page_manager.cpp
@@ -12,6 +12,12 @@
 
 #ifndef _WIN64
 #include <sys/mman.h>
+#ifdef ENABLE_USERFAULTFD
+#include <fcntl.h>
+#include <linux/userfaultfd.h>
+#include <poll.h>
+#include <sys/ioctl.h>
+#endif
 #else
 #include <windows.h>
 #endif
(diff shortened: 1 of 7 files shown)


squidbus:
Updated with code for registering multiple handlers for a signal and using signal handlers for non-illegal instruction patches. This is now complete, and needs some testing on Windows and Linux.

Kawahosein (approved):

StevenMiller123:
Consistently causes a regression in Assassin's Creed Unity (at least on Windows). Instead of the normal memory crash, it now crashes on an access violation. As the log doesn't show any errors, let me know if there is any helpful information I can provide.

StevenMiller123:
In fact, this seems to cause similar errors in many games I'm testing. I'm also seeing a handful of games print the following error:
```
[Debug] <Critical> cpu_patches.cpp:operator():656: Assertion Failed!
Unable to find module for code at: 0x0
```

squidbus:
> In fact, this seems to cause similar errors in many games I'm testing. I'm also seeing a handful of games print the following error:

Any specific games? This means that your game is trying to run code at address 0x0 for some reason.

squidbus:
Actually please check again if you can on the latest, I had something flipped around on Windows 🤦

StevenMiller123:
I'm testing on Windows 11. I've yet to find a normally working game that hasn't regressed (only 11 games into my list of 54 tested currently), so far the access violations occur in:
- Assassin's Creed Unity
- Attack on Titan 2
- Clicker Heroes
- Digimon Story: Cyber Sleuth
- Digimon Story: Cyber Sleuth Hacker's Memory

The module at 0x0 one has appeared in a handful of Unity games, and Call of Duty Advanced Warfare (which calls the sceKernelDebugRaiseExceptionOnReleaseMode function).

I also see Danganronpa 1.2 Reload crashing on a similar error:
```
[Debug] <Critical> cpu_patches.cpp:operator():656: Assertion Failed!
Unable to find module for code at: 0x1404278d6
```

All of these that throw the module error normally crash with an access violation at that same (or similar) point in code execution on main builds, so I doubt it's too important. 

NOTE: Haven't tested on the new build yet, give me a moment.

StevenMiller123:
The new push fixes the `Unable to find module for code at:` errors, doesn't appear to fix the access violations though.

StevenMiller123:
Even basic games like Sonic Mania and We Are Doomed are crashing from these access violations.

StevenMiller123:
The newest push fixed all regressions I see when comparing to main.

squidbus:
> The newest push fixed all regressions I see when comparing to main.

Thanks for checking, turns out the last bit was just using the wrong parameter for the fault address so it wasn't detecting GPU page writes correctly.

squidbus:
Latest also fixes an issue deregistering the exception handler on exit, should be good to go now.

src/video_core/page_manager.cpp
```diff
@@ -27,207 +21,49 @@ namespace VideoCore {
...
-        uffdio_writeprotect wp;
-        wp.range.start = address;
-        wp.range.len = size;
-        wp.mode = allow_write ? 0 : UFFDIO_WRITEPROTECT_MODE_WP;
-        const int ret = ioctl(uffd, UFFDIO_WRITEPROTECT, &wp);
-        ASSERT_MSG(ret != -1, "Uffdio writeprotect failed with error: {}",
-                   Common::GetLastErrorMsg());
-    }
-
-    void UffdHandler(std::stop_token token) {
```
raphaelthegreat: Userfaultfd implementation seems to have been removed. Since its still useful should probably be re-added

squidbus: It doesn't really work for using in the new central signal handling component, but I've re-added the `PageManager` implementation so it can be used alongside.

squidbus:
Comments are addressed

raphaelthegreat (approved):
Thanks a lot, awesome work!

PR #854, opened 2024-09-09
support for unlocking trophies

CrazyBloo:
edit: this is ready for review, when Vinicius finishes the image support for imgui i will make another pr with trophy icons


src/core/libraries/np_trophy/trophy_ui.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+            TrophyInfo currentTrophyInfo = trophyQueue[0];
+            if (Begin("Trophy Window", nullptr,
+                      ImGuiWindowFlags_NoDecoration | ImGuiWindowFlags_NoSavedSettings |
+                          ImGuiWindowFlags_NoInputs)) {
+                Text("Trophy earned!");
+                TextWrapped(currentTrophyInfo.trophyName.c_str());
+
+                End();
+            }
```
viniciuslrangel: The End should be outside the conditional (I made the same mistake in my PR)

src/core/libraries/np_trophy/trophy_ui.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Libraries::NpTrophy {
+
+struct TrophyInfo {
+    int trophyId = -1;
+    std::string trophyName;
+};
+
+class TrophyUI final : public ImGui::Layer {
+    bool first_render{false};
```
viniciuslrangel: unused var

CrazyBloo: ah, i assumed this was needed for imgui_layer for some reason, thank you

viniciuslrangel: It's not! I used it to default select the button for gamepad navigation

src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -4,13 +4,20 @@
 #include <unordered_map>
 
 #include "common/logging/log.h"
+#include "common/path_util.h"
 #include "common/slot_vector.h"
 #include "core/libraries/error_codes.h"
 #include "core/libraries/libs.h"
+#include "externals/pugixml/src/pugixml.hpp"
```
abouvier, line 11: You should use `#include <pugixml.hpp>` for compatibility with the system library.

CrazyBloo: i will change it, thank you

src/core/libraries/np_trophy/trophy_ui.cpp
```diff
@@ -0,0 +1,74 @@
...
+            SetNextWindowCollapsed(false);
+            SetNextWindowPos(ImVec2(io.DisplaySize.x - 200, 50));
+            KeepNavHighlight();
+
+            TrophyInfo currentTrophyInfo = trophyQueue[0];
+            if (Begin("Trophy Window", nullptr,
+                      ImGuiWindowFlags_NoDecoration | ImGuiWindowFlags_NoSavedSettings |
+                          ImGuiWindowFlags_NoInputs)) {
+                Text("Trophy earned!");
+                TextWrapped(currentTrophyInfo.trophyName.c_str());
```
abouvier, line 69: Should be `TextWrapped("%s", currentTrophyInfo.trophyName.c_str());` to compile with `-Werror=format-security`.

Evilhelder:
After this commit, some games like Bloodborne are stuck on the trophy download screen.

[shad_log.txt](https://github.com/user-attachments/files/16966893/shad_log.txt)


image
image

CrazyBloo:
> After this commit, some games like Bloodborne are stuck on the trophy download screen.
> 
> [shad_log.txt](https://github.com/user-attachments/files/16966893/shad_log.txt)
> 
> image image

Are you sure this is from this commit? I have seen people with this issue before. It doesn't seem like any of the functions I modified are called in this log. It seems like this issue is caused by sceNpTrophyGetTrophyInfo not being implemented. I will try to implement this function today

Evilhelder:
> > After this commit, some games like Bloodborne are stuck on the trophy download screen.
> > [shad_log.txt](https://github.com/user-attachments/files/16966893/shad_log.txt)
> > image image
> 
> Are you sure this is from this commit? I have seen people with this issue before. It doesn't seem like any of the functions I modified are called in this log. It seems like this issue is caused by sceNpTrophyGetTrophyInfo not being implemented. I will try to implement this function today

You are correct, sorry, I tested some builds before its implementation and the error persists, but it is strange, because if I test an older build like the clean 0.2.1 the game works normally (Obviously it does not go beyond character creation, but it does not give this error with the trophies.) sorry for the inconvenience .v.

CrazyBloo:
> You are correct, sorry, I tested some builds before its implementation and the error persists, but it is strange, because if I test an older build like the clean 0.2.1 the game works normally (Obviously it does not go beyond character creation, but it does not give this error with the trophies.) sorry for the inconvenience .v.

No worries, i will try to fix it anyway

Issue #869, opened 2024-09-11
[macOS] Regression: Several games fail to boot after PR 837 (Bindings)


shinra-electric:
At least the following games no longer boot on macOS after PR #837:

Dead Cells - CUSA11253
Limbo - CUSA01369

They fail with
```
[Debug] <Critical> liverpool.h:unhandled_exception:1110: Unreachable code!
Unhandled exception: vk::Device::createComputePipelineUnique: ErrorInvalidShaderNV
```

squidbus:
For future reference, Vulkan Validation enabled in Dead Cells yields this:
```
[Render.Vulkan] <Error> vk_platform.cpp:DebugUtilsCallback:71: VUID-VkShaderModuleCreateInfo-pCode-08737: Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
Expected Image 'Sampled Type' to be the same as Texel components
  OpImageWrite %91 %92 %93
. The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
```
Validation error is not Mac specific but does cause a crash there.

shinra-electric:
Fixed by #871

PR #871, opened 2024-09-11
shader_recompiler: Use correct integer type for OpImageWrite.

squidbus:
PR https://github.com/shadps4-emu/shadPS4/pull/837 changed the integer type for OpImageWrite from U32 to S32, which introduced the following validation error when the image type is U32:
```
VUID-VkShaderModuleCreateInfo-pCode-08737: Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
Expected Image 'Sampled Type' to be the same as Texel components
  OpImageWrite %91 %92 %93
```

Instead, use `buffer.result_type` which is already either U32[4] or S32[4] depending on the image sampled type.

Fixes regressions reported in https://github.com/shadps4-emu/shadPS4/issues/869


squidbus:
Not sure what's going on with Linux-Qt build failing in SDL3 but it seems to be unrelated to this change as its also failing in main.

raphaelthegreat (approved):

PR #798, opened 2024-09-05
add stuff for dreams

0xcaff:
* more completeness for sceGnmAddEqEvent
* allow DirectMemoryQuery to return both allocated and unallocated chunks. i believe this is the expected behavior.
* adds support for V_CVT_PKNORM_U16_F32, S_XOR_B32 and S_LOAD_DWORDX2 op codes

PR #875, opened 2024-09-12
add gnm pipeline changes for dreams

0xcaff:
from https://github.com/shadps4-emu/shadPS4/pull/798

will implement the others separately

### DIFF
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -304,6 +304,7 @@ struct AscQueueInfo {
     VAddr map_addr;
     u32* read_addr;
     u32 ring_size_dw;
+    u32 pipe_id;
 };
 static Common::SlotVector<AscQueueInfo> asc_queues{};
 static constexpr VAddr tessellation_factors_ring_addr = Core::SYSTEM_RESERVED_MAX - 0xFFFFFFF;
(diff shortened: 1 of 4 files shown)


src/core/libraries/gnmdriver/gnmdriver.cpp
```diff
@@ -371,15 +372,28 @@ s32 PS4_SYSV_ABI sceGnmAddEqEvent(SceKernelEqueue eq, u64 id, void* udata) {
     kernel_event.event.udata = udata;
     eq->AddEvent(kernel_event);
 
+    Platform::InterruptId interruptId;
+    GnmEventIdents eventId;
+
+    if (id == 64) {
```
psucien, line 378: To avoid adding these checks, change `InteruptId` enum values to the same what `GnmEventIdents` has, and then simply use single `id` in `Register` function

src/core/libraries/gnmdriver/gnmdriver.cpp
```diff
@@ -540,15 +554,14 @@ void PS4_SYSV_ABI sceGnmDingDong(u32 gnm_vqid, u32 next_offs_dw) {
...
+    liverpool->SubmitAsc(vqid, acb_span, asc_queue.pipe_id);
 
     *asc_queue.read_addr += acb_size;
     *asc_queue.read_addr %= asc_queue.ring_size_dw * 4;
 }
 
-int PS4_SYSV_ABI sceGnmDingDongForWorkload() {
-    LOG_ERROR(Lib_GnmDriver, "(STUBBED) called");
-    return ORBIS_OK;
+void PS4_SYSV_ABI sceGnmDingDongForWorkload(u32 gnm_vqid, u32 next_offs_dw) {
```
psucien, line 563: The function takes 3 args with `u64 workloadId` being the last. It is not used anyhow, but better to be added

src/core/libraries/gnmdriver/gnmdriver.cpp
0xcaff: I don't think this is a safe operation as there are other usages of InterruptId

https://github.com/shadps4-emu/shadPS4/blob/00f201e81e7063773a24e9f46251c7ec0137457e/src/video_core/amdgpu/liverpool.cpp#L645

src/core/libraries/gnmdriver/gnmdriver.cpp
0xcaff: where did you obtain this information? how are you sure its correct?

src/core/libraries/gnmdriver/gnmdriver.cpp
psucien: 1. The code you refer to is based on an assumption that `pipeId` is a valid and with appropriate value member of `InterruptId`. Why it is safe then?
2. `InterruptId` is our enum and not based on any real values, so we free to change it
3. `GnmEventIdents` and `InterruptId` members for ASC are matching already. What you need to do is to change GfxEop member only. And add a comment for the `InterruptId` enum about assumptions made

georgemoralis:
@0xcaff are you planning on doing the neccesary fixes so we can merge?

0xcaff:
hey @georgemoralis , decided to spend time unblocking stuff deeper in the pipeline. will work on merging once i've got dreams in a launching state. feel free to pull in these changes manually if they are useful for games besides dreams.

PR #876, opened 2024-09-12
Fix copyGpuBuffers when resize invalidates commands in flight

baggins183:
now PT with copyGpuBuffers=true fails at the same place as =false :)

### DIFF
--- a/src/video_core/amdgpu/liverpool.h
+++ b/src/video_core/amdgpu/liverpool.h
@@ -1088,6 +1088,15 @@ public:
         submit_cv.notify_one();
     }
 
+    void reserveCopyBufferSpace() {
+        GpuQueue& gfx_queue = mapped_queues[GfxQueueId];
+        std::scoped_lock<std::mutex> lk(gfx_queue.m_access);
+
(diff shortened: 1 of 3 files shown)


Emulator-Team-2:
This PR fixes random crashes in (J-STARS Victory VS+) right after launch. The game now boots without any problems every time. When the copyGpuBuffers option is set to “true”.

src/video_core/amdgpu/liverpool.h
```diff
@@ -1088,6 +1088,15 @@ struct Liverpool {
         submit_cv.notify_one();
     }
 
+    void reserveCopyBufferSpace() {
+        GpuQueue& gfx_queue = mapped_queues[GfxQueueId];
+        std::scoped_lock<std::mutex> lk(gfx_queue.m_access);
+
+        constexpr size_t gfx_reserved_size = (2 << 21) / sizeof(u32); // 2MB
```
psucien: I would prefer simple `2_MB` here.  Also, we're using camel case for constants naming

PR #902, opened 2024-09-13
Implement S_ABSDIFF_I32 shader instruction

LukeWarnut:
Implements the S_ABSDIFF_I32 (Scalar ALU 44) shader instruction.

This fixes a crash in **Atelier Lulua \~The Scion of Arland~** after the opening movie, and is likely used in other Atelier games. Atelier Lulua now hangs on the loading screen after the opening.

[AMD documentation ](https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/rdna2-shader-instruction-set-architecture.pdf)was used as reference.

screenshot


polybiusproxy (approved):
Implementation looks correct. Thanks.

Issue #894, opened 2024-09-12
[Regression] Bloodborne (CUSA03173) fails to load after #828


lukaszarczynski:
System:
OS: Windows 10
CPU: AMD Ryzen 7 5800X3D
GPU: AMD Radeon RX 7900 XT (24.20.11.01-240902a-407173E-KB-AMD-Software-Adrenalin-Edition)
RAM: 32GB

After merging #828, Bloodborne crashes before entering menu or displaying any output.

I'm attaching the logs from #828 commit ([fc90f27](https://github.com/shadps4-emu/shadPS4/commit/fc90f27)) and the last working commit ([bc66fe8](https://github.com/shadps4-emu/shadPS4/commit/bc66fe8)). It still crashes on the current version ([1b6cc44](https://github.com/shadps4-emu/shadPS4/commit/1b6cc447)):

[shad_log_fc90f27.txt](https://github.com/user-attachments/files/16985940/shad_log_fc90f27.txt)
[shad_log_bc66fe8.txt](https://github.com/user-attachments/files/16985941/shad_log_bc66fe8.txt)

lukaszarczynski:
#906 fixed this issue, thanks!

PR #906, opened 2024-09-14
 vulkan: Disable exceptions for getImageFormatProperties and account for errors.

squidbus:
Potential fix for https://github.com/shadps4-emu/shadPS4/issues/894. Check for possible errors from `getImageFormatProperties` and fall back to single sampling.

Needs to be verified as I don't have a reproducible system/game. If it does not fix please provide more details on the crash for investigation.

### DIFF
--- a/src/video_core/texture_cache/image.cpp
+++ b/src/video_core/texture_cache/image.cpp
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#define VULKAN_HPP_NO_EXCEPTIONS
 #include "common/assert.h"
 #include "video_core/renderer_vulkan/liverpool_to_vk.h"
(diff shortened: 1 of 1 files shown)


0xZ3n0x:
I could not reproduce the error with this one. For me, it seems to error have been resolved.

System:
OS: Windows 10
CPU: AMD Ryzen 5 5600
GPU: AMD Radeon RX 7600 XT (24.20.11.01)
RAM: 16GB

raphaelthegreat:
I think a better way might be to disable exceptions just for this file

squidbus:
> I think a better way might be to disable exceptions just for this file

Can you elaborate on how to do this as its not something I've done before.

raphaelthegreat:
Vulkan Hpp has a macro VULKAN_HPP_NO_EXCEPTIONS that you can add to the top of this file to disable exceptions. Ideally we should enable it globally but will require a bit more work to adjust existing functions (The hpp repo has a list of all available macros as well https://github.com/KhronosGroup/Vulkan-Hpp)

squidbus:
Okay, done.

PR #928, opened 2024-09-15
semaphore: Fix semaphore wait list leaking

raphaelthegreat:
When a semaphore wait timed out we did not remove the thread from the wait list. This means the list could fill with garbage entries over time. May fix memory leaks in games that use orbis semaphores a lot like Bloodborne and The Last Guardian

### DIFF
--- a/src/core/libraries/kernel/threads/semaphore.cpp
+++ b/src/core/libraries/kernel/threads/semaphore.cpp
@@ -34,10 +34,14 @@ public:
 
         // Create waiting thread object and add it into the list of waiters.
         WaitingThread waiter{need_count, is_fifo};
-        AddWaiter(&waiter);
+        const auto it = AddWaiter(&waiter);
 
         // Perform the wait.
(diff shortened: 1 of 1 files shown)


rafael-57:
My system has 32GB RAM + 10GB VRAM. Not sure if memory usage changed meaningfully, went from 14GB to 22GB total PC RAM usage after walking + fighting enemies for 5 minutes. I went from Central Yharnam to the great bridge.

Interestingly, when RAM usage hit its peak the emulator hanged for a bit, then freed up lots of memory (went immediately down to 15GB) but then the screen started blinking and it had lots of vertex explosions. I could still move and play though. Usually when it hanged like this due to RAM I remember it just crashed in the past, so maybe that's progress.

https://youtu.be/QpvuHw_EWJI

[shad_log.txt](https://github.com/user-attachments/files/17006390/shad_log.txt)

raphaelthegreat:
This does not seem to fix memory leak in bloodborne. But merging since its a valid fix anyway

PR #915, opened 2024-09-14
shader_recompiler: Implement basic 64-bit floating point support

polybiusproxy:
This PR serves as a base for 64-bit floating point shader instructions.
- Implements `V_MAX_F64` needed by **Beach Buggy Racing (``CUSA02366``)**

PR #728, opened 2024-09-02
Implement flags for RDNA instructions

OFFTKP:

PR #930, opened 2024-09-15
Sort opcodes by their indices. Group them too when applicable

korenkonder:
Making this for future changes. Creating a "base" so to speak.

Conflicts with https://github.com/shadps4-emu/shadPS4/pull/728 and https://github.com/shadps4-emu/shadPS4/pull/915


korenkonder:
Broken as of now

PR #931, opened 2024-09-15
graphics_pipeline: Move some depth configuration to dynamic state

raphaelthegreat:
Some of these depth/stencil states can change very significantly between frames even and cased a lot of unnecessary pipeline recompilation. Move them to dynamic state to solve this (these states are supported by Vulkan 1.0 so should be good compatibility wise)

### DIFF
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -1069,7 +1069,16 @@ ScePthread PThreadPool::Create(const char* name) {
         }
     }
 
+#ifdef _WIN64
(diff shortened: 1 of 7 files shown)


raphaelthegreat:
Should fix memory leak problem in bloodborne

rafael-57:
Me and others in Discord tested and it's looking very good! No more memory increase when spinning camera in-place, memory increase while walking in areas is very manageable now.

Tested in intensive areas like Central Yharnam, Old Yharnam and others.

image

[shad_log_key.txt](https://github.com/user-attachments/files/17007176/shad_log_key.txt)

Issue #865, opened 2024-09-10
[Linux AMD] Crash when minimizing/going into fullscreen


LNDF:
On Linux with RADV (AMD open source drivers) vkAcquireNextImage returns VK_TIMEOUT. when minimizing/going into fullscreen and the emulator crashes.

Additionaly the following validation layer errors show up on the logs. 

```
[Render.Vulkan] <Error> vk_platform.cpp:DebugUtilsCallback:69: VUID-vkAcquireNextImageKHR-surface-07783: Validation Error: [ VUID-vkAcquireNextImageKHR-surface-07783 ] Object 0: handle = 0xa000000000a, type = VK_OBJECT_TYPE_SWAPCHAIN_KHR; | MessageID = 0xad0e15f6 | vkAcquireNextImageKHR():  Application has already previously acquired 2 images from swapchain. Only 2 are available to be acquired using a timeout of UINT64_MAX (given the swapchain has 4, and VkSurfaceCapabilitiesKHR::minImageCount is 3). The Vulkan spec states: If forward progress cannot be guaranteed for the surface used to create the swapchain member of pAcquireInfo, the timeout member of pAcquireInfo must not be UINT64_MAX (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkAcquireNextImageKHR-surface-07783)
[Render.Vulkan] <Error> vk_platform.cpp:DebugUtilsCallback:69: VUID-vkAcquireNextImageKHR-surface-07783: Validation Error: [ VUID-vkAcquireNextImageKHR-surface-07783 ] Object 0: handle = 0xa000000000a, type = VK_OBJECT_TYPE_SWAPCHAIN_KHR; | MessageID = 0xad0e15f6 | vkAcquireNextImageKHR():  Application has already previously acquired 3 images from swapchain. Only 2 are available to be acquired using a timeout of UINT64_MAX (given the swapchain has 4, and VkSurfaceCapabilitiesKHR::minImageCount is 3). The Vulkan spec states: If forward progress cannot be guaranteed for the surface used to create the swapchain member of pAcquireInfo, the timeout member of pAcquireInfo must not be UINT64_MAX (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkAcquireNextImageKHR-surface-07783)
[...]

PR #933, opened 2024-09-15
Swapchain recreation and window resizing

LNDF:
This PR makes the window resizable and fits the image inside the window without stretching (respecting the aspect ratio).

Acquired swapchain images are released even if recreation is needed so they can be acquired again, fixing #865.

There may be a better place than the present funcion to recreate the swapchain.


AsierBujedo:
Nice work!

gandalfthewhite19890404:
Please, restart CI to compile artifacts. Current - expired already, i want to check one thing

LNDF:
Now the frame is cleaned before blitting the image so that there is no artifacts.

PR #943, opened 2024-09-16
Patch `extrq`

OFFTKP:
Patches EXTRQ (SSE4A instruction) on cpus that don't support it.
If the EXTRQ is less than 5 bytes (can be 4 bytes sometimes), it handles it directly in the illegal instruction handler and doesn't patch it because 5 bytes at least are needed for a relative jump (trampoline)

Another method for dealing with this would be copying the next instruction to run after the trampoline injected code to make more space for the relative jump. This would be faster but less safe (what happens if something jumps to the area you clobbered and replaced with a jump) (also tedious because displacements in rip relative memory access/jumps would need to be changed)

Thus this method of handling smaller than jmp near instructions is employed here

Infamous First Light seems to get further with this patch and it doesn't seem to cause performance issues due to the signal handler.


src/core/signals.cpp
```diff
@@ -3,18 +3,161 @@
...
+
+    if (instruction.length < 5) {
+        // not enough bytes for a relative jump for the trampoline
+        return true;
+    }
+
+    return false;
+}
+
+bool handleIllegalInstruction(void* ctx, void* code_address) {
```
squidbus: This should all be part of the `cpu_patches` handler, this file is meant to be the generic dispatch. Adding a bunch of specific use cases here will clutter things up too much. The utilities for accessing the context in platform-specific ways should stay here though.

OFFTKP: I'll move the utilities to a separate file since they are shared

OFFTKP: Done, could you give this another read?

src/core/cpu_patches.cpp
```diff
@@ -714,6 +846,99 @@ static std::pair<bool, u64> TryPatch(u8* code, PatchModule* module) {
...
 
+#if defined(ARCH_X86_64)
+// We need to check, before patching, if there's enough space for a relative jump to the trampoline.
+// If there isn't, the instruction must be handled specially in the illegal instruction handler
+// itself.
+static bool ShouldNotBePatched(void* code_address) {
+    ZydisDecodedInstruction instruction;
+    ZydisDecodedOperand operands[ZYDIS_MAX_OPERAND_COUNT];
+    const auto status =
+        Common::Decoder::Instance()->decodeInstruction(instruction, operands, code_address);
```
squidbus: The way things are laid out currently we're decoding the instruction twice, once here and then again in either HandleIllegalInstruction or TryPatch. Could probably refactor this a bit so we only have to decode once.

src/core/signals.cpp
```diff
@@ -22,17 +23,17 @@ namespace Core {
 static LONG WINAPI SignalHandler(EXCEPTION_POINTERS* pExp) noexcept {
     const auto* signals = Signals::Instance();
 
-    auto* code_address = reinterpret_cast<void*>(pExp->ContextRecord->Rip);
+    auto* code_address = Common::GetRip(pExp);
```
squidbus: For this and the other parameters, if we're passing the context into the signal handler anyway we might as well just make it so the only parameter is the context, then the handler function can use these common helpers to get the parts it needs out. For both access violation and illegal instruction.

src/core/signals.cpp
```diff
@@ -43,37 +44,14 @@ static LONG WINAPI SignalHandler(EXCEPTION_POINTERS* pExp) noexcept {
...
-
     ZydisDecodedInstruction instruction;
     ZydisDecodedOperand operands[ZYDIS_MAX_OPERAND_COUNT];
-    static constexpr u64 max_length = 0x20;
     const auto status =
-        ZydisDecoderDecodeFull(&decoder, code_address, max_length, &instruction, operands);
+        Common::Decoder::Instance()->decodeInstruction(instruction, operands, code_address);
     if (ZYAN_SUCCESS(status)) {
         ZydisFormatter formatter;
         ZydisFormatterInit(&formatter, ZYDIS_FORMATTER_STYLE_INTEL);
```
squidbus, line 54: Should probably update the decoder to also expose a function to use its formatter, then we can get rid of this formatter initialization as well.

squidbus (commented):
Looks mostly good, just a few comments I have.

src/core/signals.cpp
OFFTKP: Sure but I'll do it in a different PR, the decoder is just me rebasing on top of #959 so I don't have to wait for it to get merged

src/core/signals.cpp
OFFTKP: Sure, except the access violation still needs the fault address param as it gets it from info->si_addr on linux

src/core/cpu_patches.cpp
OFFTKP: Done

src/core/cpu_patches.cpp
```diff
@@ -746,17 +950,19 @@ static void TryPatchAot(void* code_address, u64 code_size) {
     }
 }
 
-static bool PatchesAccessViolationHandler(void* code_address, void* fault_address, bool is_write) {
-    return TryPatchJit(code_address);
+static bool PatchesAccessViolationHandler(void* context, void* /* fault_address */) {
+    return TryPatchJit(context);
```
squidbus:
```suggestion
    return TryPatchJit(Common::GetRip(context));
```

OFFTKP: Yikes, 5:16 AM coding session not going great

inseyven:
Bloodborne crashes with this commit

StevenMiller123:
No regressions in my library of games, though I don't have Bloodborne to test, and I don't think any of my games use SSE4a instructions. @inseyven You should probably provide a log of your crash so that regression you ran into can be researched.

OFFTKP:
@StevenMiller123 Are you on Windows or Linux? I was told it has some problem on Windows, but I thought the latest commit might have fixed it but I didn't test because I'm not on windows

StevenMiller123:
@OFFTKP My tests were done on Windows. I can't say anything about previous commits, but the most recent commit was what I used in my testing, and there were no problems.

OFFTKP:
@StevenMiller123 Cool thanks for the info

dbz400:
Other than Infamous First Light, any other games may help with this PR at this moment ?

OFFTKP:
@dbz400 Some other games hit extrq, such as driveclub and *maybe* infamous second son although I haven't seen proof of the latter yet

dbz400:
> @dbz400 Some other games hit extrq, such as driveclub and _maybe_ infamous second son although I haven't seen proof of the latter yet

Thanks .

OFFTKP:
@inseyven Another user is telling me bloodborne works just fine, could you try the latest exe and see if it still crashes?

squidbus (approved):
LGTM 👍

PR #948, opened 2024-09-16
Patch CPUID

OFFTKP:
Untested and WIP since DS crashes for some unrelated reason

### DIFF
--- a/src/core/cpu_patches.cpp
+++ b/src/core/cpu_patches.cpp
@@ -798,4 +798,50 @@ void PrePatchInstructions(u64 segment_addr, u64 segment_size) {
 #endif
 }
 
+// We patch CPUID to handle it specially and simulate an AMD environment.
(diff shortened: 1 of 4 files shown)


korenkonder:
There's a better way of doing this. You can relocate instructions after that too

OFFTKP:
~~Needs to be reworked to instead patch it normally and copy the instruction at the end of the injected code since the trampoline jump is bigger than 2 bytes for cpuid~~

Nevermind, after some discussion it's not as easy as it sounds, because the next instruction may be a relative jump or some other code might want to jump to the instruction after cpuid that would get clobbered.

bigol83:
Tested with Windows, it doesn't crash for me, but Dark Souls still has the same issue of exploding geometry when moving the character

Cattura

OFFTKP:
No matter the combination of values returned the dark souls problem is not fixed so I am going to retire this PR, the method to patch the CPUID was pretty neat I think so maybe it can be useful in the future

PR #953, opened 2024-09-17
Pre-release | Unifies builds

DanielSvoboda:
- Unifies all workflows into a single build.yml file;
- At the end of the workflow, pre-release files are automatically created only when there is a push to the main branch;
- The other old pre-releases and their tags are deleted, keeping the repository clean with only the latest pre-release."

This will be necessary for the auto-update feature being developed...

# Images:
Action:
image

Artifacts:
image

Pre-build:
image


.github/workflows/build.yml
```diff
@@ -0,0 +1,334 @@
...
+      uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: latest
+
+    - name: Install MoltenVK
+      run: |
+        arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+        arch -x86_64 /usr/local/bin/brew install molten-vk
+
+    - name: Cache CMake dependency build objects
```
LeDragoX: Missing code from macos.yml above
https://github.com/shadps4-emu/shadPS4/blob/28ec489dbeafe2b56fd916e6205991a75eb59f01/.github/workflows/macos.yml#L34-L43

.github/workflows/build.yml
```diff
@@ -0,0 +1,334 @@
...
+        arch -x86_64 /usr/local/bin/brew install molten-vk
+    - uses: jurplel/install-qt-action@v4
+      with:
+        version: 6.7.2
+        host: mac
+        target: desktop
+        arch: clang_64
+        archives: qtbase qttools
+
+    - name: Cache CMake dependency build objects
```
LeDragoX: Missing code from macos-qt.yml above
https://github.com/shadps4-emu/shadPS4/blob/28ec489dbeafe2b56fd916e6205991a75eb59f01/.github/workflows/macos-qt.yml#L43-L52

.github/workflows/build.yml
```diff
@@ -0,0 +1,334 @@
...
+    needs: get-info
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+
+    - name: Install dependencies
+      run: sudo apt-get update && sudo apt install -y libx11-dev libxext-dev libwayland-dev libfuse2 clang build-essential
+
+    - name: Cache CMake dependency build objects
```
LeDragoX: Missing code from linux.yml above
https://github.com/shadps4-emu/shadPS4/blob/28ec489dbeafe2b56fd916e6205991a75eb59f01/.github/workflows/linux.yml#L28-L37

.github/workflows/build.yml
```diff
@@ -0,0 +1,334 @@
...
+    needs: get-info
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+
+    - name: Install dependencies
+      run: sudo apt-get update && sudo apt install -y libx11-dev libxext-dev libwayland-dev libfuse2 clang build-essential qt6-base-dev qt6-tools-dev
+
+    - name: Cache CMake dependency build objects
```
LeDragoX: Missing code from linux-qt.yml above
https://github.com/shadps4-emu/shadPS4/blob/28ec489dbeafe2b56fd916e6205991a75eb59f01/.github/workflows/linux-qt.yml#L28-L37

viniciuslrangel:
Awesome!

Isn't keeping old pre-releases a wanted behavior?

DanielSvoboda:
> Awesome!
> 
> Isn't keeping old pre-releases a wanted behavior?

I did this with the intention of keeping only the latest version, while removing the older pre-releases. Pre-releases are mainly intended for 'auto-update', while users who want to download a specific change can access the action artifacts, as is currently done. Pre-releases can still be manually downloaded by people who do not have a GitHub account, and these people will also check for the latest version, and it will make the releases screen cleaner, so I see no reason to keep all pre-releases. This is my opinion, but this decision is up to the DEVs, and they only need to remove the last part of the code for this.

viniciuslrangel:
Makes sense. Agree 😁

gandalfthewhite19890404:
Can you add backup folder for previous versions. I know that is hdd space consuming, but why not?

DanielSvoboda:
> Can you add backup folder for previous versions. I know that is hdd space consuming, but why not?

I think you are commenting on the other PR-1006 'Auto Update'
Do you think it is better to make a backup before updating, but which files would be saved? just the ones that will be replaced?

gandalfthewhite19890404:
> > Can you add backup folder for previous versions. I know that is hdd space consuming, but why not?
> 
> I think you are commenting on the other PR-1006 'Auto Update' Do you think it is better to make a backup before updating, but which files would be saved? just the ones that will be replaced?

I guess only ShadPS4.exe at current point.

PR #955, opened 2024-09-17
Draft: Exceptions

roamic:
What needs to be done:
* Any synchronization code that can be called from any game thread should be re-written to be interruptable/alertable where applicable (using either pthreads or native API).
* Investigate why raising an exception does nothing although signal callback is called on the desired thread. Probably something to do with context structure not having some info.

### DIFF
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -1,10 +1,6 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
-#include <mutex>
-#include <semaphore>
-#include <thread>
-
 #include "common/alignment.h"
 #include "common/assert.h"
 #include "common/error.h"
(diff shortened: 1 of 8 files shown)


src/core/libraries/kernel/thread_management.cpp
```diff
@@ -1373,97 +1375,90 @@ int PS4_SYSV_ABI posix_pthread_detach(ScePthread thread) {
...
-        SetPosixErrno(EINVAL);
-        return -1;
-    }
-    if (sem != nullptr) {
-        *sem = new PthreadSemInternal{
-            .semaphore = std::counting_semaphore<ORBIS_KERNEL_SEM_VALUE_MAX>{value},
-            .value = {static_cast<int>(value)},
-        };
+int PS4_SYSV_ABI posix_sem_init(sem_t* sem, int pshared, unsigned int value) {
+    int result = sem_init(sem, pshared, value);
```
squidbus, line 1379: This cannot be used on macOS, unnamed POSIX semaphores are not implemented.

roamic: Oh, this was just an experiment to make `evStart` from `pthreads` to alert the thread. This should be removed and re-worked either way. We cannot use `winpthreads` anyway as it does not provide the required functionality.

Randomuser8219:
Any updates?

Issue #675, opened 2024-08-30
Deleted


Emulator-Team-2:

roamic:
I have investigated and consulted with the team and it seems to be because of unstable TLS instructions patching on Windows.

Emulator-Team-2:
Tested with the main build - (v0.2.1 WIP) - https://github.com/shadps4-emu/shadPS4/commit/418274038433e674b5e793322742b9b286c12802

I found another game that doesn't work on Windows.
- Scott Pilgrim vs. the World: The Game

## Linux:

![1](https://github.com/user-attachments/assets/d67c83cf-b5e3-4e53-8e88-f9908df2ccca)

Log:
[shad_log.zip](https://github.com/user-attachments/files/16824828/shad_log.zip)

## Windows:

On Windows, the game remains on a black screen and doesn't progresses any further.

![1](https://github.com/user-attachments/assets/0a11a095-2532-47c3-bb49-a30535306269)

Log:
[shad_log.zip](https://github.com/user-attachments/files/16824832/shad_log.zip)

rafael-57:
These TLS instructions seem to be the cause of issue #687. I've had a person on Linux try and warp on BB to reproduce the issue. He warped 12 times and the emulator never hanged.

On Windows you can get hangs even on the first warp. You almost always get hang by the third or 4th warp. And the error is TLS malloc.

voguelike:
> These TLS instructions seem to be the cause of issue #687. I've had a person on Linux try and warp on BB to reproduce the issue. He warped 12 times and the emulator never hanged.
> 
> On Windows you can get hangs even on the first warp. You almost always get hang by the third or 4th warp. And the error is TLS malloc.

i tested my personal build just now and surprisingly found no TLS malloc errors and i'm still freezing indefinitely on warp. i've always had the TLS malloc error whenever this happened previously.

instead i now sometimes only have errors with 
`
[Lib.Rtc] <Error> rtc.cpp:sceRtcGetCurrentClockLocalTime:80: (STUBBED) called
[Lib.Rtc] <Error> rtc.cpp:sceRtcGetTick:122: (STUBBED) called
[Lib.Rtc] <Error> rtc.cpp:sceRtcSetTick:191: (STUBBED) called
`
and
`[Render.Vulkan] <Error> tile_manager.cpp:TryDetile:389: Unsupported tiled image: R32Sfloat (Depth_MacroTiled)`
[...]

PR #956, opened 2024-09-17
core: Replace Windows TLS code patches with exception handler.

squidbus:
Implements a new strategy for Windows TCB accesses using just an exception handler.

On Windows, FS segment accesses are generally guaranteed to cause an access violation. We currently take advantage of this to patch code on the fly, however instead we can actually just check if the instruction accessed FS and set the FS register. Windows doesn't seem to really care what's in the FS register, and it won't matter if the OS resets it since we can just set it back again any time a fault occurs.

By taking this approach we can eliminate the historically unreliable code patch that tries to calculate the location of the right TLS value directly in assembly. The hope is that directly populating FS base from `TlsGetValue` will be much more reliable. Plus this allows us to reliably patch any type of FS access instead of just `mov reg, fs:0x0`, although no other use patterns have been observed yet.

In the process of implementing this I moved the ahead-of-time TLS patching for Linux out of the main patching system into special patching code just for that OS, since that allows us to completely remove the access violation handler from the CPU patcher and just focus on illegal instructions at runtime. Eventually this too will hopefully go away as I want to experiment with swapping in and out the FS register on guest code entry/exit to eliminate that patch as well.

Tested this on a few games on a Windows machine I got access to, didn't find any issues.

### DIFF
--- a/src/core/cpu_patches.cpp
+++ b/src/core/cpu_patches.cpp
@@ -11,7 +11,6 @@
 #include "common/assert.h"
 #include "common/types.h"
 #include "core/signals.h"
-#include "core/tls.h"
 #include "cpu_patches.h"
 
 #ifdef _WIN32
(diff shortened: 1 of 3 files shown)


StevenMiller123:
Significantly regresses performance in Kingdom Hearts 0.2 (and probably other currently working UE4 titles I'd assume). The game performs hundreds of batchmap calls, and these calls are significantly slower on this build, while on main, the game speeds through these calls and initializes AvPlayer (where it currently crashes on Windows). The PR log doesn't have any crash, I stopped it after around 5 minutes of waiting. Main gets to the crash in around 6 seconds, for reference.

[shad_log Main.txt](https://github.com/user-attachments/files/17032707/shad_log.Main.txt)
[shad_log PR.txt](https://github.com/user-attachments/files/17032708/shad_log.PR.txt)

squidbus:
Please check again as I added some caching.

Emulator-Team-2:
Issue: https://github.com/shadps4-emu/shadPS4/issues/675.

With this PR `Disgaea 4 Complete +` doesn't crash before the opening video starts playing, but instead stands on a black screen.

##

Log Main Build:
[shad_log.zip](https://github.com/user-attachments/files/17033521/shad_log.zip)

PR:
[shad_log.zip](https://github.com/user-attachments/files/17033524/shad_log.zip)

StevenMiller123:
The new build seems faster than the prior build, but it's still significantly slower than main. Here's a log from this build. Like before, the memory calls became far too slow, so I stopped it around 3 minutes in.
[shad_log.txt](https://github.com/user-attachments/files/17034074/shad_log.txt)

squidbus:
This is probably not viable then, patch will need to stay.

raphaelthegreat:
Maybe you can do this, but permanently patch the instruction with the existing trampoline. So we get benefit of code discovery

squidbus:
> Maybe you can do this, but permanently patch the instruction with the existing trampoline. So we get benefit of code discovery

That's what the existing code does on Windows, so we already have that benefit. The goal here was to reduce the need for code patching and improve reliability over the trampoline equivalent of `TlsGetValue`.

raphaelthegreat:
I think the trampoline is okay reliability wise since it was based on kernel32/wine code andnd win32 is pretty stable ABI wise. But I see yeah

PR #959, opened 2024-09-17
Use a singleton for instruction decoding

OFFTKP:
We're going to need the decoder in more places soon and don't need to do ZydisDecoderInit everywhere

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -350,8 +350,8 @@ set(COMMON src/common/logging/backend.cpp
            src/common/config.cpp
            src/common/config.h
            src/common/debug.h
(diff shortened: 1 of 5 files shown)


polybiusproxy (commented):
Why not use the [Singleton](https://github.com/shadps4-emu/shadPS4/blob/main/src/common/singleton.h) class?

OFFTKP:
Ah sure missed it

polybiusproxy (approved):
Looks good

georgemoralis:
@OFFTKP is that absolute now?

OFFTKP:
Since extrq was merged this isn't needed

PR #961, opened 2024-09-17
Set High Priority for Windows

Xphalnos:
Inspired by Cemu source code, I added the feature that shadPS4 process is "above normal" instead of "normal".
This can surely improve performance. (Only available for Windows!)


Xphalnos:
@georgemoralis any news ?

Foul-Tarnished:
@Xphalnos Could you also add game mode support ? Or maybe it needs to be whitelisted by Windows itself

When you open xbox game bar (Win+G) and gonto its settings, you will see the checkbox "Remember as a game" is not ticked.

Game mode can slightly improve frame pacing and reduce latency.

(I also use SpecialK to double check if it's enabled)

PR #989, opened 2024-09-19
video_core: Implement sceGnmInsertPushColorMarker

korenkonder:


raphaelthegreat:
This looks good to me after comparing with ghidra, will merge this

PR #992, opened 2024-09-20
Improve sceSystemServiceReceiveEvent stub

StevenMiller123:
I've added both an EventType enum, and what should be the proper event struct, both based primarily off fpPS4 implementations. In the one game I've seen calling this function, the game runs a while loop until ReceiveEvent returns something other than 0, causing an infinite loop with the current implementation. 

This needs testing with other games that can call this function, as I would assume some games may regress from the error return. If any regressions are found, please provide a log. 

Confirmed improvements:
- Ghost of Tsushima
- Stranger of Paradise Final Fantasy Origin
- Hatsune Miku Project Diva X


gandalfthewhite19890404:
Adventures of Fat Princess uses this

gandalfthewhite19890404:
Sorry, Ghost of Tsushima uses this

gandalfthewhite19890404:
Now GoTs goes further
image

StevenMiller123:
@gandalfthewhite19890404 Could you provide a log? I've got the method set to log some extra details, so a log will tell me if there's anything the game is hardcoded to expect.

gandalfthewhite19890404:
@StevenMiller123 
[shad_log.txt](https://github.com/user-attachments/files/17077367/shad_log.txt)

StevenMiller123:
Thanks, looks like my stub is working as expected in Ghost of Tsushima.

dbz400:
Stranger of Paradise: Final Fantasy Origin also uses it and it now goes in-game graphic

polybiusproxy (approved):

PR #995, opened 2024-09-20
vulkan: Enable VULKAN_HPP_NO_EXCEPTIONS broadly.

squidbus:
This is a follow-up to some discussion in https://github.com/shadps4-emu/shadPS4/pull/906. Enables `VULKAN_HPP_NO_EXCEPTIONS` broadly, adjusting the necessary API calls to check the result value.

In some places a non-success result will fail gracefully, where it makes sense to or other checks on the output already do so. Others I have added asserts to basically mirror the current behavior but without exceptions; these can be adjusted in the future as needed. Feel free to comment if you have any specific preferences anywhere.


squidbus:
Rebased

PR #996, opened 2024-09-20
ci: Refresh CMake cache on build.

squidbus:
Passes an argument to `cmake` to refresh the cache when configuring. This means less of the CMake files will be reused from cache, but also makes it more tolerant to environment changes, e.g. the current issues in macOS builds from a recent GitHub actions image update.

Worth noting as well that sometimes GitHub will gradually phase in new actions images and you can get runs with both old and new versions for a while, so just clearing the cache doesn't really work reliably.

I tested this in my fork and still got 3-4 minute build times across all the targets with warmed up cache.


LeDragoX:
Very good addition, I saw that command before, but I'm new to cmake, so I didn't tested it out, CIs are pretty stable now with it, congrats @squidbus!!!.

PR #1001, opened 2024-09-20
vulkan: Add alternatives when nullDescriptor is not supported.

squidbus:
MoltenVK does not support the `nullDescriptor` feature of `VK_EXT_robustness2`. Instead, bind to the null image view or buffer as a workaround.


squidbus:
Extended to cover null buffers as well.

raphaelthegreat:
Can you also give those null resources a name with SetObjectName so its recognizable in renderdoc when they are bound

squidbus:
> Can you also give those null resources a name with SetObjectName so its recognizable in renderdoc when they are bound

Done.

PR #1011, opened 2024-09-21
SaveLib PR related fixes

viniciuslrangel:
- Fix SFO reading in QT, handling all cases when the key is not present
- Avoid backup error crashing the world
- Handle invalid param from old firmwares


viniciuslrangel:
Good to merge

PR #1007, opened 2024-09-21
vulkan: Use closest available equivalent to missing clamp modes.

squidbus:
Adding support for these clamp modes will likely require non-trivial sampler addressing emulation in shaders. For now, warn and use the closest available equivalent to enable games that use them to function.

Also set `samplerMirrorClampToEdge` in the Vulkan 1.2 features since it is required for using `vk::SamplerAddressMode::eMirrorClampToEdge`.

PR #1013, opened 2024-09-21
shader_recompiler: Additional scope handling and user data as push constants

raphaelthegreat:
The initial goal of this PR was to handle some additional cases of divergence spotted in some Bloodborne particle shaders which caused certain branches to be entirely missed.

However this fix by itself, actually regressed the game because some compute shaders used for copying data around had some bounds checks implemented that fell under this category. So the change made the branches work properly, but the limit was stored in user data regs. The compiler treated user data regs as compiler constants and expected they would not change between the same shader. However these shaders can be bound multiple times with different limits depending on how much game was copying.

So the control flow improvement is also bundled with a different change that moves user data regs to the push constant block. I've also added some minor additions from various tests that don't deserve their own PR. This may improve some particle effects in BB but I have not noticed any visual change

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -474,6 +474,7 @@ set(SHADER_RECOMPILER src/shader_recompiler/exception.h
                       src/shader_recompiler/params.h
                       src/shader_recompiler/runtime_info.h
                       src/shader_recompiler/specialization.h
+                      src/shader_recompiler/backend/bindings.h
                       src/shader_recompiler/backend/spirv/emit_spirv.cpp
                       src/shader_recompiler/backend/spirv/emit_spirv.h
                       src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
--- a/src/core/libraries/kernel/thread_management.cpp
+++ b/src/core/libraries/kernel/thread_management.cpp
@@ -1069,7 +1069,16 @@ ScePthread PThreadPool::Create(const char* name) {
         }
     }
 
+#ifdef _WIN64
     auto* ret = new PthreadInternal{};
+#else
+    // TODO: Linux specific hack
+    static u8* hint_address = reinterpret_cast<u8*>(0x7FFFFC000ULL);
+    auto* ret = reinterpret_cast<PthreadInternal*>(
+        mmap(hint_address, sizeof(PthreadInternal), PROT_READ | PROT_WRITE,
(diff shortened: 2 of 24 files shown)


raphaelthegreat:
This may also help improve random vertex explosions

Emulator-Team-2:
https://github.com/shadps4-emu/shadPS4/issues/496#issuecomment-2362079848

Fixes a crash in `Trine Enchanted Edition` related to a missing opcode (V_CVT_PKNORM_U16_F32). 
Now the game crashes with this error.

```
[Debug] <Critical> liverpool_to_vk.cpp:operator():680: Assertion Failed!
Unsupported component swap mode 3
```

Log:
[shad_log.zip](https://github.com/user-attachments/files/17085826/shad_log.zip)

ElBread3:
With this PR, Kyoei Toshi now currently crashes with this error.

```
[Debug] <Critical> emit_spirv_image.cpp:EmitImageQueryDimensions:189: Unreachable code!
SPIR-V Instruction
```

because Color2DMsaa is not implemented in the EmitImageQueryDimensions function

raphaelthegreat:
> EmitImageQueryDimensions

This should be added now

ElBread3:
> > EmitImageQueryDimensions
> 
> This should be added now

After fixing that, Kyoei Toshi's now here:
```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:351: Compiling vs shader 0x54f55417 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:351: Compiling vs shader 0x54f55417 (permutation)
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:351: Compiling fs shader 0x8aabfc44 
[Core] <Error> stubs.cpp:CommonStub:42: Stub: sceNpWebApiDeleteContext (nid: XUjdsSTTZ3U) called, returning zero to 0x9f005ee6f
[Debug] <Critical> liverpool_to_vk.cpp:SurfaceFormat:585: Unreachable code!
Unknown data_format=49 and num_format=4
```

squidbus:
FYI, the clamp mode added here is also in https://github.com/shadps4-emu/shadPS4/pull/1007

Moonlacer:
For the record this PR now allows RDNA3 to run Bloodborne completely vanilla without crashing after character creation

PR #1015, opened 2024-09-21
shader_recompiler: Use SetDst in more instructions.

squidbus:
Fix up some places where the destination register was hardcoded as a basic vector register by using `SetDst` instead, to make sure all the necessary transformations are applied.

Fixes incorrect screen rendering in CUSA16429 due to a missing multiply by 2 on a destination.

### DIFF
--- a/src/shader_recompiler/frontend/translate/vector_alu.cpp
+++ b/src/shader_recompiler/frontend/translate/vector_alu.cpp
@@ -359,14 +359,13 @@ void Translator::EmitVectorAlu(const GcnInst& inst) {
 // VOP2
 
 void Translator::V_CNDMASK_B32(const GcnInst& inst) {
-    const IR::VectorReg dst_reg{inst.dst[0].code};
     const IR::ScalarReg flag_reg{inst.src[2].code};
     const IR::U1 flag = inst.src[2].field == OperandField::ScalarGPR
                             ? ir.GetThreadBitScalarReg(flag_reg)
                             : ir.GetVcc();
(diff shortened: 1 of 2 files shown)


src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -1046,11 +1033,9 @@ void Translator::V_SAD_U32(const GcnInst& inst) {
 void Translator::V_LSHL_B64(const GcnInst& inst) {
     const IR::U64 src0{GetSrc64(inst.src[0])};
     const IR::U64 src1{GetSrc64(inst.src[1])};
-    const IR::VectorReg dst_reg{inst.dst[0].code};
     ASSERT_MSG(src0.IsImmediate() && src0.U64() == 0 && src1.IsImmediate() && src1.U64() == 0,
                "V_LSHL_B64 with non-zero src0 or src1 is not supported");
-    ir.SetVectorReg(dst_reg, ir.Imm32(0));
-    ir.SetVectorReg(dst_reg + 1, ir.Imm32(0));
+    SetDst64(inst.dst[0], ir.Imm64(0LL));
```
raphaelthegreat: Best to keep this as is to avoid redundant pack instruction

raphaelthegreat (approved):

src/shader_recompiler/frontend/translate/vector_alu.cpp
squidbus: Reverted this one.

PR #1016, opened 2024-09-21
shader_recompiler: Exclude non-float results from output modifiers.

squidbus:
* Exclude non-floating point result values from output modifiers as defined in the spec. This is also already how we treat source input modifiers and destination output modifiers for 64-bit values.
* Re-enable `SetDst` for `V_ADDC_U32` as this fixes the issue.


squidbus:
Confirmed previous regression from `V_ADDC_U32` change is no longer occurring.

PR #1018, opened 2024-09-22
Build error due to missing std::optional

amilamad:
- When building on Ubuntu 24.04 with gcc 13 build failed with error 

> Projects/shadPS4_dev/src/core/libraries/save_data/save_backup.h:55:6: error: ‘optional’ in namespace ‘std’ does not name a template type
   55 | std::optional<BackupRequest> PopLastEvent();
      |      ^~~~~~~~
Adding #include <optional> to src/core/libraries/save_data/save_backup.h fixed the issue


StevenMiller123:
This fix is already included in https://github.com/shadps4-emu/shadPS4/pull/1011

PR #1019, opened 2024-09-22
Better screenshots

Xphalnos:
I removed the "Sister Projects" part because it already exists in "Special Thanks".

Before:
Before

After:
After


polybiusproxy:
Might be best to only update screenshots once the new release is up.

PR #1026, opened 2024-09-22
Separate Updates from Game Folder

ElBread3:
Up until this point, you either had to keep a duplicate copy of a game (2x space requirements 😵‍) or uninstall and reinstall a game (keeping the pkg around somewhere) in order to play two versions of a game at the same time. I propose we meet in the middle: updates will be extracted to their own folder CUSAXXXXX-UPDATE, and the emulator will read files from this update directory instead of the game folder if present. Additionally, there is a right-click menu to delete the game or the update from the GUI.

I have tested by booting Limbo, but tests for other games are definitely needed as for obvious reasons I haven't tested with Bloodborne. I also haven't tested DLC, as none of my library has any, so DLC will need tested too.

### DIFF
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -56,7 +56,8 @@ std::filesystem::path MntPoints::GetHostPath(std::string_view guest_directory, b
     const auto rel_path = std::string_view(corrected_path).substr(pos);
     std::filesystem::path host_path = mount->host_path / rel_path;
 
-    std::filesystem::path patch_path = mount->host_path.string() + "-UPDATE";
+    std::filesystem::path patch_path = mount->host_path;
+    patch_path += "-UPDATE";
     if (corrected_path.starts_with("/app0/") && std::filesystem::exists(patch_path / rel_path)) {
         host_path = patch_path / rel_path;
     }
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include <algorithm>
+#include "common/config.h"
 #include "common/string_util.h"
 #include "core/file_sys/fs.h"
 
@@ -56,7 +57,7 @@ std::filesystem::path MntPoints::GetHostPath(std::string_view guest_directory, b
     std::filesystem::path host_path = mount->host_path / rel_path;
 
     std::filesystem::path patch_path = mount->host_path.string() + "-UPDATE";
-    if (std::filesystem::exists(patch_path / rel_path)) {
+    if (std::filesystem::exists(patch_path / rel_path) && Config::getSeparateUpdateEnabled()) {
         host_path = patch_path / rel_path;
     }
 
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -53,7 +53,14 @@ std::filesystem::path MntPoints::GetHostPath(std::string_view guest_directory, b
     // Remove device (e.g /app0) from path to retrieve relative path.
     pos = mount->mount.size() + 1;
     const auto rel_path = std::string_view(corrected_path).substr(pos);
-    const auto host_path = mount->host_path / rel_path;
+    std::filesystem::path host_path = mount->host_path / rel_path;
+
+    //Use file in update directory instead if it's there (e.g. CUSAXXXXX-UPDATE)
+    std::filesystem::path patch_path = mount->host_path.string() + "-UPDATE";
+    if (std::filesystem::exists(patch_path / rel_path)) {
+        host_path = patch_path / rel_path;
+    }
+
     if (!NeedsCaseInsensitiveSearch) {
         return host_path;
     }
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -57,7 +57,7 @@ std::filesystem::path MntPoints::GetHostPath(std::string_view guest_directory, b
     std::filesystem::path host_path = mount->host_path / rel_path;
 
     std::filesystem::path patch_path = mount->host_path.string() + "-UPDATE";
-    if (std::filesystem::exists(patch_path / rel_path) && Config::getSeparateUpdateEnabled()) {
+    if (corrected_path.starts_with("/app0/") && std::filesystem::exists(patch_path / rel_path)) {
         host_path = patch_path / rel_path;
     }
 
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -55,7 +55,6 @@ std::filesystem::path MntPoints::GetHostPath(std::string_view guest_directory, b
     const auto rel_path = std::string_view(corrected_path).substr(pos);
     std::filesystem::path host_path = mount->host_path / rel_path;
 
-    //Use file in update directory instead if it's there (e.g. CUSAXXXXX-UPDATE)
     std::filesystem::path patch_path = mount->host_path.string() + "-UPDATE";
     if (std::filesystem::exists(patch_path / rel_path)) {
         host_path = patch_path / rel_path;
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -60,6 +60,7 @@ static bool vkMarkers = false;
 static bool vkCrashDiagnostic = false;
 static s16 cursorState = HideCursorState::Idle;
 static int cursorHideTimeout = 5; // 5 seconds (default)
+static bool separateupdatefolder = false;
 
 // Gui
 std::vector<std::filesystem::path> settings_install_dirs = {};
@@ -207,6 +208,10 @@ bool vkCrashDiagnosticEnabled() {
     return vkCrashDiagnostic;
 }
 
+bool getSeparateUpdateEnabled() {
+    return separateupdatefolder;
+}
+
 void setGpuId(s32 selectedGpuId) {
     gpuId = selectedGpuId;
 }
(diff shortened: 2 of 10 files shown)


AselGamer:
Tested it for 30 mins seems to work fine for bloodborne, my save had DLC items and there was no issue loading them. Only had 1 crash, and it didn't have anything to do with the PR. Sending the report anyway just in case.
[shad_log.txt](https://github.com/user-attachments/files/17096649/shad_log.txt)

ElBread3:
Nice! Glad to see there doesn't seem to be any issues! Though I've got two questions, as I'm trying to verify saves version correctly:

1. Did you make sure that the original game folder is unpatched 1.0 and not a still updated copy?
2. If you remove the sce_sys folder in the original game folder does making new saves work correctly (and particularly does the param.sfo in the save folder have 1.09 as the version, but that might be too technical for you to answer)?

Curioso9998:
[shad_log.zip](https://github.com/user-attachments/files/17100283/shad_log.zip)
I updated the following games but apparently one shows that pace has been updated another is not them the Guilty Gear Xrd -Revelator- CUSA04112 and the ULTRA STREET FIGHTER IV CUSA01846 I say I updated but the behavior of the emulator a the fazelos seems strange to me because it closes and the Guilty Gear Xrd -Revelator does not show the update information but the ULTRA STREET FIGHTER IV even shows but I found it too fast when doing it apparently it didn't update it in fact no

AselGamer:
> Nice! Glad to see there doesn't seem to be any issues! Though I've got two questions, as I'm trying to verify saves version correctly:
> 
>     1. Did you make sure that the original game folder is unpatched 1.0 and not a still updated copy?
> 
>     2. If you remove the sce_sys folder in the original game folder does making new saves work correctly (and particularly does the param.sfo in the save folder have 1.09 as the version, but that might be too technical for you to answer)?

1. The original game folder appears to be 1.00, but the update is still applied I don't know how s
ince I removed the separate update folder, I also checked the param.sfo, and it says it's version 1.00. Also, I reinstalled the game without the update and I crash before the title screen, sending log just in case [shad_log.txt](https://github.com/user-attachments/files/17101695/shad_log.txt).
2. I can't even select the game in the emulator without the sce_sys folder. 
2.1 I just checked the param.sfo in the savedata folder, and it doesn't have the game version, either that or I don't know how to read hexadecimal

I just saw you updating the pr, I hope my testing hasn't been redundant I would have replied earlier, but I was busy

ElBread3:
To clarify why I'm asking, the way the system works is if there is a file in the update directory to load it in there instead of in the original game folder. I'm trying to find any missed place where it could load data from the original game folder instead of the update folder where it should.

The first question was to solve if you have a 1.09 game directory and then a 1.09 update directory. If it loads from the original game directory, which we don't want, I couldn't tell because no matter which it loaded they would both be 1.09.

The second question was to solve if the save system attempted to load from the original game sfo instead of the update sfo, resulting in the save file sfo to be 1.00 (again, don't want this). If there's no sfo in the original game, then if it errors that it can't find it there that's an issue, and if it doesn't error then it finds the sfo in the right place.

TLDR I'm trying to figure out if I need to redirect the save system to the update folder or if it's already good

ElBread3:
This PR seems tested enough, ready to review

georgemoralis:
I would prefer to have an option for user to decide if it want to use the same folder (overwrite) or separate . Also what happends if game has multiple updates?

Hermiten:
> I would prefer to have an option for user to decide if it want to use the same folder (overwrite) or separate . Also what happends if game has multiple updates?

The best things is to keep the original game and when you do multiple update, you can only play the original game or the game with the latest update and not every update.

Example for Bloodborne:
You can only have Bloodborne 1.00 and Bloodborne 1.09
If you only apply 1 update like bloodborne 1.07, you can but you cannot downgrade your update, you can only update it.

Is it clear ?
It's my opinion

ElBread3:
> I would prefer to have an option for user to decide if it want to use the same folder (overwrite) or separate . Also what happends if game has multiple updates?

I can definitely make using the separate update path a config option. Second, I agree with Hermiten, just like has it's been before installing another update will just overwrite the update. If you would prefer it not to, I can add the version number in the folder name and only use the latest, but selecting an update won't be possible to implement until per-game configs arrive.

ArtemVideoGames:
It's abandoned or just in stop?

ElBread3:
> It's abandoned or just in stop?

It's not abandoned, just haven't figured out how to solve the issue where after I updated submodules it's deleted ext-boost. Also needs a re-review after as I've since added the config option.

Edit: Finally compiles! So every time the submodule changelogs appear in the GUI I updated them but once updating on the dev branch it deleted ext-boost so I just didn't do that. Only thing is it appears in the git diff on the first commit, but it doesn't appear in the changed files tab so that may or may not be OK.

src/core/file_sys/fs.cpp
```diff
@@ -53,7 +54,13 @@ std::filesystem::path MntPoints::GetHostPath(std::string_view guest_directory, b
...
     pos = mount->mount.size() + 1;
     const auto rel_path = std::string_view(corrected_path).substr(pos);
-    const auto host_path = mount->host_path / rel_path;
+    std::filesystem::path host_path = mount->host_path / rel_path;
+
+    std::filesystem::path patch_path = mount->host_path.string() + "-UPDATE";
+    if (std::filesystem::exists(patch_path / rel_path) && Config::getSeparateUpdateEnabled()) {
+        host_path = patch_path / rel_path;
+    }
+
```
viniciuslrangel: It would be better to check if this starts with `/app0/`. The game could also be reading from a save mounted folder

ElBread3: Something like "rel_path.string().starts_with("/app0/")" in the if?

viniciuslrangel: Prob better to use `corrected_path`. Also, host_path.string() can crash when inside a folder with special characters. You could save the parent path and only do the `-UPDATE` concat to the directory name

par274:
Hello, although the game patches was downloaded after the build, it does not appear in the ui(completely empty). Also, the updates are not installed anywhere else. Are you sure it works?

ElBread3:
> Hello, although the game patches was downloaded after the build, it does not appear in the ui(completely empty). Also, the updates are not installed anywhere else. Are you sure it works?

Be careful about the word "download" for game updates. Does it change the version number on the game in the list to the update number? Also did you enable the config option beforehand?

par274:
> > Hello, although the game patches was downloaded after the build, it does not appear in the ui(completely empty). Also, the updates are not installed anywhere else. Are you sure it works?
> 
> Be careful about the word "download" for game updates. Does it change the version number on the game in the list to the update number? Also did you enable the config option beforehand?

Yes, I turned on the relevant setting from the settings. The version is upgrading, but after that when I say "delete update" I get the message "no update found". I think it's still writing the updates into the game's files.

ElBread3:
> Yes, I turned on the relevant setting from the settings. The version is upgrading, but after that when I say "delete update" I get the message "no update found". I think it's still writing the updates into the game's files.

Must be an issue with the install dir feature, I didn't account for it, I'll add it to my fix PR momentarily

par274:
> > Yes, I turned on the relevant setting from the settings. The version is upgrading, but after that when I say "delete update" I get the message "no update found". I think it's still writing the updates into the game's files.
> 
> Must be an issue with the install dir feature, I didn't account for it, I'll add it to my fix PR momentarily

By the way, I didn't say "update", I said "patch". I meant Widescreen patch, resolution patch etc. after this build, it downloads the patches but they don't appear in the interface, the screen goes blank. So, what I meant by "downloading" in that sentence was patch/mod. Not game/update.

resim

But it is visible if patches are installed before the build. So I didn't mention downloading the game or the update.

ElBread3:
@par274 could you test with the fixes PR and make sure there's a -UPDATE folder and see what happens? It should be grabbing the version info from the game info, it might be that BB has no patches on 1.00 and well it's not updated

par274:
> @par274 could you test with the fixes PR and make sure there's a -UPDATE folder and see what happens? It should be grabbing the version info from the game info, it might be that BB has no patches on 1.00 and well it's not updated

Ok i download this build: https://github.com/shadps4-emu/shadPS4/actions/runs/11350759439

By the way, another bug. It says the game is installed even though it is not.

resim

This build creates a "-UPDATE" folder but its content is missing.
resim

Only sce_sys

resim

I know that there is a dvdroot_ps4 folder in this update. And when I press the "delete update" button, the emulator says it's 1.0, but the updated version is still visible in the game. I think it's still loading it into the game files.

ElBread3:
@par274 those two bugs should now be fixed, you can retest
also could you tell me if the patches got fixed or not?

par274:
Build: https://github.com/shadps4-emu/shadPS4/actions/runs/11352777085

Patch bug continues. What I noticed is that the reason why the patches are not showing up is because the game has not been updated to 1.09.
resim

Update files are placed in the correct location but the game does not read them.
The version still appears as 1.0 in the bottom right.
resim

ElBread3:
It seems my fix PR got merged already so I'll work on this tomorrow

par274:
> It seems my fix PR got merged already so I'll work on this tomorrow

Should I test the main build?

ElBread3:
> Should I test the main build?

yep it should have the fix PR

ElBread3:
but if you are already on latest fix PR no changes have been made since merge obviously so no point

par274:
> but if you are already on latest fix PR no changes have been made since merge obviously so no point

Okay I tried the last "main" build and there is no change. This is actually a very nice PR and if it is developed it can be used for logic game modes as well. In this way, we can install mods without affecting the game files.

par274:
After your last update everything works as it should.

resim

ElBread3:
> After your last update everything works as it should.

This is with https://github.com/shadps4-emu/shadPS4/pull/1397?

par274:
> > After your last update everything works as it should.
> 
> This is with #1397?

Yes #1397

PR #1028, opened 2024-09-22
Add nullptr check in scePthreadSetprio function

tGecko:
Apparently sometimes scePthreadSetprio gets called with 0 as thread. Guard against that.

With this, I can get past first menu and up until intro cutscene in CUSA03365 (Dark Souls III).
Before this, it crashed within a few seconds of starting up.

DS3 tries to set prio to nullptr-threads 7 times

image

image

Log with this: now crashes with `vk::Device::waitForFences: ErrorDeviceLost` in intro cutscene
[shad_log.txt](https://github.com/user-attachments/files/17091219/shad_log.txt)


DanielSvoboda:
In CUSA05158 the scePthreadGetprio function also called thread null, I believe it also has to have this 'if'.

Kironte:
This also allows the DS3 network test to go in-game! 5-10 FPS but in-game nonetheless.

PR #1049, opened 2024-09-23
Save data: fix nullptr & concurrent file write

viniciuslrangel:
Fixes RDR crash reported by users


DanielSvoboda:
I tested it and now it is working correctly

PR #1050, opened 2024-09-24
 amdgpu: Fix buffer comparison by naming padding fields for initialization.

squidbus:
Comparing two buffers was not working because `memcmp` is not guaranteed to work properly for bitfield structs with anonymous padding, as the padding is not initialized.

This was causing some shaders to balloon in binding count since buffers were not properly being matched and de-duplicated. It also caused pipeline compilation failures in CUSA16429 since two different permutations could have different memcmp results in de-duplication and thus compile with different buffer bindings.

### DIFF
--- a/src/video_core/amdgpu/resource.h
+++ b/src/video_core/amdgpu/resource.h
@@ -23,7 +23,7 @@ enum class CompSwizzle : u32 {
 // Table 8.5 Buffer Resource Descriptor [Sea Islands Series Instruction Set Architecture]
 struct Buffer {
     u64 base_address : 44;
-    u64 : 4;
+    u64 _padding0 : 4;
     u64 stride : 14;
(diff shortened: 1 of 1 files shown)


raphaelthegreat:
Would zero initializing the inline_cbuf work as well? Using BitField is fine enough tho we specifically avoided it for the sharps as it reduces readability

squidbus:
> Would zero initializing the inline_cbuf work as well? Using BitField is fine enough tho we specifically avoided it for the sharps as it reduces readability

I tried using the zero initialize operator {} both on the field in BufferResource and where the BufferResource is created, and neither worked. Also BitField is already being used for Sampler sharps.

squidbus:
I'm pretty sure the root issue is that any bits that are not defined to a field are not zero-initialized, only the explicit fields.

PR #1051, opened 2024-09-24
fix lane inst decoding

baggins183:
these instructions were using dst[1] for some reason, even though the data_share.cpp implementations read from dst[0].
also fix how some of the src operands are treated
fixes issues with rb4


baggins183:
should probably be tested with other games with readlane, writelane, readfirstlane

georgemoralis:
red dead redemption uses it i think , someone should test if there are improvements with it
(or totaly broke :D)

bigol83:
> red dead redemption uses it i think , someone should test if there are improvements with it (or totaly broke :D)

I checked the initial area and the first cutscenes, doesn't seem to improve anything, but it also doesn't break anything

jnackmclain:
thank you once again 🙏

PR #1059, opened 2024-09-24
Don't attempt to promote eR8G8B8A8Unorm to depth

LukeWarnut:
Because eR8G8B8A8Unorm already has depth, it doesn't need to be converted. I also replaced the `else if` statement with a `switch`.

This fixes a crash in **Zombie Army Trilogy** (CUSA01837), but another error causes the game to crash right afterwards.


baggins183:
are you saying it should treat the B channel (z coord) as depth?

Dont we need some kind of image view that only reads from the z coord if its going to be used as a depth attachment?

baggins183:
I am curious about this too. I also saw this come up in P.T. @psucien would know more

PR #1064, opened 2024-09-25
Use fs::path::native whenever possible, avoid unnecessary fs->string conversions in GUI code

OFFTKP:
The problem is that `std::filesystem::path::string` throws an exception on Windows if the path contains special characters.

This causes the emulator to mysteriously crash with no immediately obvious error on Windows if the executable (or config files or other paths used by the GUI) are in some path that contains special characters (such as Cyrillic, Spanish etc.)

This has been a very annoying problem that gets a lot of users to report it in discord, and it's not easy to solve because a lot of times users don't know what a path is.

---

The native string format used on Windows for paths is wstring, which supports special characters. On POSIX it's regular old std::string. path::native returns the correct type based on OS.
This PR reduces some unnecessary uses of `path::string` by using the fs::path itself, and opts for `path::native` on the rest.

An annoying side effect is that fmt::format & co (for good reason) doesn't allow you to:
- Mix strings and wstrings
- Use wstrings without adding `L` in front of the format string

---

Emulator code that does path->string conversions was ignored because there was a lot of them in logging facilities which are the most annoying to deal with, and most likely there isn't going to be any paths with wide characters used in games


src/imgui/renderer/imgui_core.cpp
```diff
@@ -51,8 +51,14 @@ void Initialize(const ::Vulkan::Instance& instance, const Frontend::WindowSDL& w
...
-    io.IniFilename = SDL_strdup(config_path.string().c_str());
-    io.LogFilename = SDL_strdup(log_path.string().c_str());
+
+    using native_char = std::filesystem::path::value_type;
+    io.IniFilename = new char[config_path.native().size() * sizeof(native_char)];
+    io.LogFilename = new char[log_path.native().size() * sizeof(native_char)];
+    std::memcpy((void*)io.IniFilename, config_path.native().c_str(),
+                config_path.native().size() * sizeof(native_char));
+    std::memcpy((void*)io.LogFilename, log_path.native().c_str(),
+                log_path.native().size() * sizeof(native_char));
```
viniciuslrangel: Imgui expects to be UTF8 and internally converts to Wide Char.
```suggestion
    auto path = config_path.u8string();
    char* config_file_buf = new char[path.size() + 1];
    config_file_buf[path.copy((char8_t*)config_file_buf, path.size())] = '\0';
    io.IniFilename = config_file_buf;

    path = log_path.u8string();
    char* log_file_bug = new char[path.size() + 1];
    log_file_bug[path.copy((char8_t*)log_file_bug, path.size())] = '\0';
    io.LogFilename = log_file_bug;
```

Also change the Shutdown below here to use delete instead of SDL_free
```cpp
    const auto ini_filename = io.IniFilename;
    const auto log_filename = io.LogFilename;

    Vulkan::Shutdown();
    Sdl::Shutdown();
    DestroyContext();

    delete[] ini_filename;
    delete[] log_filename;
```

OFFTKP:
Needs some more stuff on Qt

viniciuslrangel (changes requested):

src/imgui/renderer/imgui_core.cpp
```diff
@@ -52,13 +52,15 @@ void Initialize(const ::Vulkan::Instance& instance, const Frontend::WindowSDL& w
...
-    using native_char = std::filesystem::path::value_type;
-    io.IniFilename = new char[config_path.native().size() * sizeof(native_char)];
-    io.LogFilename = new char[log_path.native().size() * sizeof(native_char)];
-    std::memcpy((void*)io.IniFilename, config_path.native().c_str(),
-                config_path.native().size() * sizeof(native_char));
-    std::memcpy((void*)io.LogFilename, log_path.native().c_str(),
-                log_path.native().size() * sizeof(native_char));
+    auto path = config_path.u8string();
+    char* config_file_buf = new char[path.size() + 1]();
+    std::memcpy(config_file_buf, path.c_str(), path.size());
```
viniciuslrangel, line 57: Still need to add the null terminator. Prob just use `path.size() + 1`

OFFTKP: Won't new char\[...](); initialize the array to zeroes?

viniciuslrangel: Oh, didn't see the `()`, mb

viniciuslrangel:
LGTM 👍

RedDog777777:
I am just learning lol so confused

RedDog777777 (approved):

OFFTKP:
Reviewable

PR #1068, opened 2024-09-25
shader_recompiler: Sample images using correct result type.

squidbus:
Similar to ImageFetch, update other image ops to use the correct result type and bitcast to F32.

Fixes validation errors and potential compile issues in some shaders in CUSA16429 where it was trying to sample from a uint texture with float result type.


raphaelthegreat:
Looks good. Can you make the bitcast conditional if the image is integer in a similar way to what texture buffers are doing? So for the most common case the shader wont have a useless bitcast

squidbus:
> Looks good. Can you make the bitcast conditional if the image is integer in a similar way to what texture buffers are doing? So for the most common case the shader wont have a useless bitcast

Sure, done.

PR #1071, opened 2024-09-25
Use UTF-8 in console

nishinji:
I don't know if it is correct to put the code here, but I think it is correct to put it right after the console initialisation.

# Before
image

# After
image


nishinji:
Sorry already in https://github.com/shadps4-emu/shadPS4/pull/1064

PR #1073, opened 2024-09-25
TLS Malloc Fixes

StevenMiller123:
In games like Attack on Titan 2, the current heap API implementation can cause issues specific to Linux. By disabling the heap API malloc calls, like we do on Windows, these games can progress further. If you're in the shadPS4 discord, this message will provide more context: https://discordapp.com/channels/1080089157554155590/1258386683691274361/1288493573200810005

Also fixes a minor oversight in the Windows TLS code, where no malloc calls were made if there was no heap_api provided by the game. This appears to fix inconsistent crashes when testing Uncharted 4 on Windows.


StevenMiller123:
So Mac doesn't like printing a pthread? Is there any better way to identify the thread that would work on all OSes? For now I've disabled printing the thread id. I will update the pr (and fix clang format) when I know of a better solution.

StevenMiller123:
Swapped to draft, as I need to revert this change for Mac specifically, as the game this issue occurs in cannot be tested on Mac yet. I will update this when I have the chance.

StevenMiller123:
Should be ready for testing now.

camdrit:
Tried this on Linux with BloodBorne. No notable change in behavior either for good or bad. Still hangs in Kernel same as main branch so I'm not sure how useful of a test this is, but the output is attached. Have yet to dump other games from my system so I haven't tried other games.

[shadps4-output.txt](https://github.com/user-attachments/files/17416521/shadps4-output.txt)

StevenMiller123:
@camdrit Thanks for the test. This PR isn't expected to help Bloodborne, as that game needs work on pthreads to progress further on Linux. If you do end up dumping any other games, feel free to test those.

StevenMiller123:
I'm closing this PR, as #1440 better resolves the issues this PR was designed to fix.

PR #1090, opened 2024-09-26
Config: Configurable User Path

ElBread3:
Implements configuration of user paths on emulator startup, and refactors to use a ConfigDir path type. To solve the issue of the configuration file's location not being determinable on startup anymore, the config file is placed in the app data folder.


ElBread3:
Closing as there is the issue of the user paths being created before the config file is able be loaded

PR #1091, opened 2024-09-26
Error reporting on failed memory allocation

OFFTKP:


OFFTKP:
Someone came to the discord and this assert was hitting, but the error wasn't very descriptive so this would be good for when it happens again

PR #1093, opened 2024-09-27
trophy icon + platinum fixes

CrazyBloo:
This should be ready to merge

edit: I have reworked how trophy unlock state is saved, so trophy progress has been reset. If you want to keep your trophy progress go into shadps4/user/gamedata/(your game serial)/trophyfiles/xml/trop.xml and change all trophies with "unlocked" to "true"

Had to rework how platinum trophies are unlocked, the previous method would unlock the platinum early on some games.
Still need to add support for games with multiple platinum trophies, but none boot yet.

image


src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -223,6 +223,14 @@ int PS4_SYSV_ABI sceNpTrophyGetGameIcon(OrbisNpTrophyContext context, OrbisNpTro
     return ORBIS_OK;
 }
 
+struct GameTrophyInfo {
+    uint32_t numGroups;
```
raphaelthegreat:
```suggestion
    u32 num_groups;
```
Ditto for all other cases of these changes

src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -249,19 +257,8 @@ int PS4_SYSV_ABI sceNpTrophyGetGameInfo(OrbisNpTrophyContext context, OrbisNpTro
...
-        numTrophiesByRarity[3] = 0;
-        numTrophiesByRarity[4] = 0;
-        uint32_t unlockedTrophies = 0;
-        uint32_t unlockedTrophiesByRarity[5];
-        unlockedTrophiesByRarity[1] = 0;
-        unlockedTrophiesByRarity[2] = 0;
-        unlockedTrophiesByRarity[3] = 0;
-        unlockedTrophiesByRarity[4] = 0;
+        GameTrophyInfo gameInfo;
+        memset(&gameInfo, 0, sizeof(GameTrophyInfo));
```
raphaelthegreat:
```suggestion
        GameTrophyInfo info{};
```

src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -249,19 +257,8 @@ int PS4_SYSV_ABI sceNpTrophyGetGameInfo(OrbisNpTrophyContext context, OrbisNpTro
 
     if (result) {
```
raphaelthegreat: This result branch can be made into early return
```
if (!result) {
    LOG_INFO(Lib_NpTrophy, "couldnt parse xml : {}", result.description());
    return ORBIS_OK;
}
```
However returning ORBIS_OK in this situation seems a bit risky, either appropriate error code or assert is probably better

src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -249,19 +257,8 @@ int PS4_SYSV_ABI sceNpTrophyGetGameInfo(OrbisNpTrophyContext context, OrbisNpTro
...
-        uint32_t unlockedTrophiesByRarity[5];
-        unlockedTrophiesByRarity[1] = 0;
-        unlockedTrophiesByRarity[2] = 0;
-        unlockedTrophiesByRarity[3] = 0;
-        unlockedTrophiesByRarity[4] = 0;
+        GameTrophyInfo gameInfo;
+        memset(&gameInfo, 0, sizeof(GameTrophyInfo));
 
         auto trophyconf = doc.child("trophyconf");
         for (pugi::xml_node_iterator it = trophyconf.children().begin();
```
raphaelthegreat:
```suggestion
        for (const auto& node : trophyconf.children()) {
```
No need for iterators here. Additionally I see a lot of unnecessary std::string construction here which results in a lot of heap allocations. Consider using sting_view which has no additional cost.
```
const auto name = std::string_view{node.name()};

File: src/core/libraries/np_trophy/np_trophy.cpp [hunk 5]
```diff
@@ -278,35 +275,35 @@ int PS4_SYSV_ABI sceNpTrophyGetGameInfo(OrbisNpTrophyContext context, OrbisNpTro
...
-                numGroups++;
+                gameInfo.numGroups++;
 
             if (std::string(it->name()) == "trophy") {
                 std::string currentTrophyUnlockState = it->attribute("unlockstate").value();
                 std::string currentTrophyGrade = it->attribute("ttype").value();
 
-                numTrophies++;
+                gameInfo.numTrophies++;
                 if (!currentTrophyGrade.empty()) {
```
raphaelthegreat: Can be made into early return
```
if (current_trophy_grade.empty()) {
    continue;
}
```

src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -350,26 +354,14 @@ int PS4_SYSV_ABI sceNpTrophyGetGroupInfo(OrbisNpTrophyContext context, OrbisNpTr
...
-        numTrophiesByRarity[2] = 0;
-        numTrophiesByRarity[3] = 0;
-        numTrophiesByRarity[4] = 0;
-        uint32_t unlockedTrophies = 0;
-        uint32_t unlockedTrophiesByRarity[5];
-        unlockedTrophiesByRarity[1] = 0;
-        unlockedTrophiesByRarity[2] = 0;
-        unlockedTrophiesByRarity[3] = 0;
-        unlockedTrophiesByRarity[4] = 0;
+        GroupTrophyInfo groupInfo;
```
raphaelthegreat: Ditto

src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -998,14 +975,14 @@ int PS4_SYSV_ABI sceNpTrophyUnlockTrophy(OrbisNpTrophyContext context, OrbisNpTr
...
                         }
 
-                        g_trophy_ui.AddTrophyToQueue(trophyId, currentTrophyName);
+                        g_trophy_ui.AddTrophyToQueue(currentIconPath, currentTrophyName);
                     }
                 }
             }
         }
 
         if (std::string(platinumIt->attribute("unlockstate").value()).empty()) {
```
raphaelthegreat:
```suggestion
        if (platinumIt->attribute("unlockstate").empty()) {
```
Is this a valid transformation? If not consider using std::string_view instead. And possibly a helper lambda to reduce some code duplication

src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -1022,12 +999,14 @@ int PS4_SYSV_ABI sceNpTrophyUnlockTrophy(OrbisNpTrophyContext context, OrbisNpTr
 
                 std::string platinumTrophyId = platinumIt->attribute("id").value();
                 std::string platinumTrophyName = platinumIt->child("name").text().as_string();
+                std::string platinumIconPath =
```
raphaelthegreat: Can this be kept as std::filesystem path and using append operator?

src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -1022,12 +999,14 @@ int PS4_SYSV_ABI sceNpTrophyUnlockTrophy(OrbisNpTrophyContext context, OrbisNpTr
 
                 std::string platinumTrophyId = platinumIt->attribute("id").value();
                 std::string platinumTrophyName = platinumIt->child("name").text().as_string();
```
raphaelthegreat:
```suggestion
                std::string_view platinumTrophyId = platinumIt->attribute("id").value();
                std::string_view platinumTrophyName = platinumIt->child("name").text().as_string();
```

src/core/libraries/np_trophy/np_trophy.cpp
```diff
@@ -394,30 +387,30 @@ int PS4_SYSV_ABI sceNpTrophyGetGroupInfo(OrbisNpTrophyContext context, OrbisNpTr
 
                 if (!currentTrophyGroupID.empty()) {
                     if (std::stoi(currentTrophyGroupID) == groupId) {
```
raphaelthegreat: I see you call std::stoi(currentTrophyGroupID) here. Can't the original value be retrieved as a number with as_int() ?

raphaelthegreat (changes requested):
This doesn't seem to have been reviewed last PR. The code is mostly good but doesn't match the style of the general codebase (numTrophies -> num_trophies). Also all if statements should have braces regardless if they are single statement or not and std::string_view should be preferred over std::string when possible

viniciuslrangel (changes requested):
Could you make static AddTrophyToQueue/RemoveThophyToQueue add trophy info to the queue and use a std::optional field to create a new TrophyUI for each one this way you don't have to track stuff like icon_loaded, then calling .reset() to destruct the last one when there is not more trophy left to allow RefCountedTexture free the image from memory and to remove from the layer list.

Also please use ImGui::GetIO()->DeltaTime to make the 5sec counter instead of the system clock, so we can control the time rate. You can make this timer call the RemoveThophy so it destruct itself.

RefCountedTexture is also async, so you should use the boolean operator to wrap the texture drawing to not draw an invalid ID, triggering vulkan validation errors. Use a filled rect in else branch as placeholder

src/core/libraries/np_trophy/trophy_ui.cpp
```diff
@@ -4,71 +4,86 @@
...
-using namespace Libraries::NpTrophy;
+namespace Libraries::NpTrophy {
 
-TrophyUI::TrophyUI() {
+std::optional<TrophyUI> current_trophy_ui;
+std::queue<TrophyInfo> trophy_queue;
+
+TrophyUI::TrophyUI(std::filesystem::path trophyIconPath, std::string trophyName)
+                   : trophy_icon_path(trophyIconPath), trophy_name(trophyName) {
+    trophy_start_time = std::chrono::system_clock::now().time_since_epoch().count();
```
viniciuslrangel: Instead of saving the start time, set the timer to 5.0f and each draw `timer -= ImGui::GetIO().DeltaTime;` so you don't depend on external clocks

src/core/libraries/np_trophy/trophy_ui.cpp
```diff
@@ -4,71 +4,86 @@
...
+    trophy_time_now = trophy_start_time;
+    if (std::filesystem::exists(trophy_icon_path)) {
+        trophy_icon = RefCountedTexture::DecodePngFile(trophy_icon_path);
+    } else {
+        LOG_ERROR(Lib_NpTrophy, "Couldnt load trophy icon at {}", trophy_icon_path.string());
+    }
```
viniciuslrangel: You can load here and only store the icon, as the decoding runs in the background and will not cause any slowdown

src/core/libraries/np_trophy/trophy_ui.cpp
```diff
@@ -4,71 +4,86 @@
...
+    if (Begin("Trophy Window", nullptr,
+              ImGuiWindowFlags_NoDecoration | ImGuiWindowFlags_NoSavedSettings |
+                  ImGuiWindowFlags_NoInputs)) {
+        if (trophy_icon) {
+            Image(trophy_icon.GetTexture().im_id, ImVec2(50, 50));
+            ImGui::SameLine();
+        } else {
+            // placeholder
+            ImGui::GetWindowDrawList()->AddRectFilled(ImVec2(1040, 60), ImVec2(1090, 110),
+                                                      GetColorU32(ImVec4{0.7f}));
```
viniciuslrangel: Instead of absolute position, it's better to use:
```suggestion
            const auto pos = GetCursorScreenPos();
            ImGui::GetWindowDrawList()->AddRectFilled(pos, pos + ImVec2{50.0f}, GetColorU32(ImVec4{0.7f}));
```

viniciuslrangel (changes requested):
I think that's all!

viniciuslrangel (changes requested):
Just remembered: here should have a mutex. And the pop queue too

src/core/libraries/np_trophy/trophy_ui.cpp
```diff
@@ -4,71 +4,85 @@
...
+            TrophyInfo next_trophy = trophy_queue.front();
+            trophy_queue.pop();
+            current_trophy_ui.emplace(next_trophy.trophy_icon_path, next_trophy.trophy_name);
+        } else {
+            current_trophy_ui.reset();
         }
     }
 }
+
+void AddTrophyToQueue(std::filesystem::path trophyIconPath, std::string trophyName) {
```
viniciuslrangel, line 81: Just remembered: here should have a mutex. And the pop queue too

src/core/libraries/np_trophy/trophy_ui.h
```diff
@@ -5,32 +5,37 @@
...
     ~TrophyUI() override;
 
-    void AddTrophyToQueue(int trophyId, std::string trophyName);
-
     void Finish();
 
     void Draw() override;
+
+private:
+    std::filesystem::path trophy_icon_path;
```
viniciuslrangel: Field unused

viniciuslrangel (changes requested):

src/core/libraries/np_trophy/trophy_ui.cpp
CrazyBloo: I'm glad you noticed this, I didn't think about a race condition at all here :P

viniciuslrangel (approved):

PR #1097, opened 2024-09-27
shader_recompiler: Define fragment output type based on number format.

squidbus:
* Expand fragment shader runtime info for render targets to include the number format.
* Declare fragment shader render target outputs with the correct number type.
* Cast to the correct output number type in `EmitSetAttribute`.
* Fix a misplaced brace, a wrong exception message, and the param pointer type in `OutputAttrPointer` (currently always the same as what was hardcoded, but use the right field in case it isn't in the future).

This fixes at least two shaders in CUSA16429 which were trying to output to R32Uint targets with float type, causing SPIR-V errors.

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
@@ -85,11 +85,11 @@ Id OutputAttrPointer(EmitContext& ctx, IR::Attribute attr, u32 element) {
     }
 }
 
-Id OutputAttrComponentType(EmitContext& ctx, IR::Attribute attr) {
+std::pair<Id, bool> OutputAttrComponentType(EmitContext& ctx, IR::Attribute attr) {
     if (IR::IsParam(attr)) {
         const u32 index{u32(attr) - u32(IR::Attribute::Param0)};
         const auto& info{ctx.output_params.at(index)};
-        return info.component_type;
+        return {info.component_type, info.is_integer};
     }
     switch (attr) {
     case IR::Attribute::Position0:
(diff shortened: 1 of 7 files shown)


rafael-57:
[shad_log - Copy (6).txt](https://github.com/user-attachments/files/17166653/shad_log.-.Copy.6.txt)

[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:358: Compiling fs shader 0x9e3394c8 
[Debug] <Critical> spirv_emit_context.cpp:GetAttributeInfo:172: Unreachable code!
Invalid attribute type Srgb

This breaks the Bloodborne, crashes on load

squidbus:
> This breaks the Bloodborne, crashes on load

The error here is fixed now.

rafael-57:
> > This breaks the Bloodborne, crashes on load
> 
> The error here is fixed now.

[shad_log.txt](https://github.com/user-attachments/files/17168223/shad_log.txt)

Glad to be of help, can confirm it's fixed

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -222,7 +257,8 @@ void EmitSetAttribute(EmitContext& ctx, IR::Attribute attr, Id value, u32 elemen
         return;
     }
     const Id pointer{OutputAttrPointer(ctx, attr, element)};
-    ctx.OpStore(pointer, ctx.OpBitcast(ctx.F32[1], value));
+    const Id component_type{OutputAttrComponentType(ctx, attr)};
+    ctx.OpStore(pointer, ctx.OpBitcast(component_type, value));
```
raphaelthegreat: Can this check if MRT number format is integer too so it doesn't emit an unnecessary bitcast on the most common case of float

squidbus: This was doing a bitcast to float previously, is it okay to skip that?

raphaelthegreat: From my understanding this bitcast is there to ensure type of value written matches the number type of destination MRT. SetAttribute always receives a float input so if MRT is also float which is most common case we can skip the bitcast. Though I guess this is fine because MRT outputs are not very common

squidbus: Okay, updated to skip the bitcast for float.

PR #1101, opened 2024-09-27
vulkan: Fix some common validation errors.

squidbus:
* When `VK_KHR_maintenance5` is not supported, fall back to the old logic of setting all flags on buffers since they cannot be added on to buffer views later.
* Add extension `SPV_EXT_demote_to_helper_invocation` when using capability `DemoteToHelperInvocationEXT` for cases when the extension is being used with Vulkan 1.2.

These are the two most common causes of validation error spam when I'm testing on MoltenVK, which makes it hard to spot other more useful validation errors.


rafael-57:
[shad_log.txt](https://github.com/user-attachments/files/17159888/shad_log.txt)

Tested with BB, doesn't seem to render any new effects or have any regressions

squidbus:
Added on a fix for validation error that buffer binding range cannot be zero.

PR #1103, opened 2024-09-27
Fix HLE Module loads

StevenMiller123:
With the current code, if someone provides improperly dumped firmware modules, or corrupted firmware modules, the HLE module variant won't load. As we already return an error in linker->LoadModule when these issues occur, the only change needed is to make the LoadSystemModules function check for an error return.


StevenMiller123:
Closing this, as it sounds like @georgemoralis plans to add this fix to their upcoming PR.

PR #1105, opened 2024-09-27
Misc fixes12

georgemoralis:
Some misc fixes as usual.

- improve some sceNetCtl calls
- Added NpGetStatus
- Loading of ngs2 lle is now done from sysloadmodule fuction . That prevents of loading it on startup if not neccesary. More modules will follow this


gandalfthewhite19890404:
Child of light refuses to boot with that PR:
image

src/core/libraries/system/sysmodule.cpp
```diff
@@ -6,6 +6,7 @@
 #include <magic_enum.hpp>
 
 #include "common/logging/log.h"
+#include "core/libraries//kernel/libkernel.h"
```
polybiusproxy, line 9: There's a double slash here

Randomuser8219:
What happened to this pull request?

PR #1107, opened 2024-09-27
documentation: fix Linux build instructions

LNDF:
Update the build instruction for Linux.
- Use clang when building just like in the github workflows
- Fix dependency package installation.

Tested on Ubuntu 24.04, Fedora 40, Arch Linux and OpenSUSE Tumbleweed.

In some cases, jack2 is used insteed of pipewire-jack. This should be corrected.


Phil-Nettekoven (approved):

tGecko:
fresh fedora KDE install here, I was still missing libXext-devel after following your instructions

PR #1116, opened 2024-09-27
menu description | rename: Release/Nightly

DanielSvoboda:
adds a field with text descriptions of what each option does
It's similar to what you have in PCSX2 and RPCS3

image

----

Change name from stable/unstable to Release/Nightly,
added goupbox with text 'Update Channel' in settings.
added text 'Update Channel' on update screen, to make it clearer which version will be downloaded.
also adjusted code so it works if updateChannel name is not Release/Nightly.
also updateChannel is set as default according to build.
image

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -3,6 +3,7 @@
 
 #include <fstream>
 #include <string>
+#include <common/version.h>
 #include <fmt/core.h>
 #include <fmt/xchar.h> // for wstring support
 #include <toml.hpp>
@@ -37,7 +38,7 @@ static s32 gpuId = -1; // Vulkan physical device index. Set to negative for auto
 static std::string logFilter;
 static std::string logType = "async";
 static std::string userName = "shadPS4";
-static std::string updateChannel = "stable";
+static std::string updateChannel;
 static bool useSpecialPad = false;
 static int specialPadClass = 1;
 static bool isDebugDump = false;
@@ -414,7 +415,11 @@ void load(const std::filesystem::path& path) {
         logFilter = toml::find_or<std::string>(general, "logFilter", "");
         logType = toml::find_or<std::string>(general, "logType", "sync");
         userName = toml::find_or<std::string>(general, "userName", "shadPS4");
-        updateChannel = toml::find_or<std::string>(general, "updateChannel", "stable");
+        if (Common::isRelease) {
+            updateChannel = toml::find_or<std::string>(general, "updateChannel", "Release");
+        } else {
+            updateChannel = toml::find_or<std::string>(general, "updateChannel", "Nightly");
+        }
         isShowSplash = toml::find_or<bool>(general, "showSplash", true);
         isAutoUpdate = toml::find_or<bool>(general, "autoUpdate", false);
     }
@@ -565,7 +570,11 @@ void setDefaultValues() {
     logFilter = "";
     logType = "async";
     userName = "shadPS4";
-    updateChannel = "stable";
+    if (Common::isRelease) {
+        updateChannel = "Release";
+    } else {
+        updateChannel = "Nightly";
+    }
     useSpecialPad = false;
     specialPadClass = 1;
     isDebugDump = false;
--- a/src/qt_gui/check_update.cpp
+++ b/src/qt_gui/check_update.cpp
@@ -22,6 +22,7 @@
 #include <common/config.h>
 #include <common/path_util.h>
 #include <common/scm_rev.h>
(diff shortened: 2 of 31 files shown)


ElBread3:
Here's my set of text for this feature:

Console Language: Sets the language that the PS4 game uses. It's recommended to set this to a language the game supports, which will vary by region.

Emulator Language: Sets the language of the emulator's user interface.

Enable Full Screen: Automatically puts the game window into full-screen mode. This can be toggled by pressing the F11 key.

Show Splash Screen: Shows the game's splash screen (a special image) while the game is starting.

Is PS4 Pro: Makes the emulator act as a PS4 PRO, which may enable special features in games that support it.

Username: Sets the PS4's account username, which may be displayed by some games.

Log Type: Sets whether to synchronize the output of the log window for performance. May have adverse effects on emulation.

Log Filter: Filters the log to only print specific information. Examples: "Core:Trace" "Lib.Pad:Debug Common.Filesystem:Error" "*:Critical" Levels: Trace, Debug, Info, Warning, Error, Critical - in this order, a specific level silences all levels preceding it in the list and logs every level after it in the list.

Update: Stable: Official versions released every month that may be very outdated, but are more reliable and tested. Unstable: Development versions that have all the latest features and fixes, but may contain bugs and are less stable.

Play Title Music: If a game supports it, enable playing special music when selecting the game in the GUI.

Graphics Device: On multiple GPU systems, select the GPU the emulator will use from the drop down list, or select "Auto Select" to automatically determine it.

Width/Height: Sets the size of the emulator window at launch, which can be resized during gameplay. This is different from the in-game resolution.

Vblank Divider: The frame rate at which the emulator refreshes at is multiplied by this number. Changing this may have adverse effects, such as increasing the game speed, or breaking critical game functionality that does not expect this to change!

Enable Shaders Dumping: For the sake of technical debugging, saves the games shaders to a folder as they render.

Enable Null GPU: For the sake of technical debugging, disables game rendering as if there were no graphics card.

Enable PM4 Dumping: For the sake of technical debugging, saves raw GPU instruction data to a folder as the emulator processes it.

f1amy:
Maybe show descriptions on hover?

DanielSvoboda:
> Maybe show descriptions on hover?

yes, it only appears if you hover the mouse over the element, and when nothing is selected it says:
'Point your mouse at an options to display a description in here'

ElBread3:
Here's the Debug tab text, I hope I haven't misunderstood the technical details:

Enable Debug Dumping: Saves the import and export symbols and file header information of the currently running PS4 program to a directory.

Enable Vulkan Validation Layers: Enables a system that validates the state of the Vulkan renderer and logs information about it's internal state. This will reduce performance and likely change the behavior of emulation.

Enable Vulkan Synchronization Validation: Enables a system that validates the timing of Vulkan rendering tasks. This will reduce performance and likely change the behavior of emulation.

Enable RenderDoc Debugging: If enabled, the emulator will provide compatibility with Renderdoc to allow capture and analysis of the currently rendered frame.

f1amy:
> > Maybe show descriptions on hover?
> 
> yes, it only appears if you hover the mouse over the element, and when nothing is selected it says: 'Point your mouse at an options to display a description in here'

I mean in a popup/tooltip instead of a section below, usually its done that way

f1amy:
Maybe something like this: https://doc.qt.io/qt-5/qml-qtquick-controls2-tooltip.html

DanielSvoboda:
> > > Maybe show descriptions on hover?
> > 
> > 
> > yes, it only appears if you hover the mouse over the element, and when nothing is selected it says: 'Point your mouse at an options to display a description in here'
> 
> I mean in a popup/tooltip instead of a section below, usually its done that way

I think it looks better with the text below, and I also did a poll, where the majority also vote.
It's similar to what you have in PCSX2 and RPCS3
image

Yaricheg1:
Can you add a progress bar to show how long it takes to download a new version of the emulator like in RPCS3?

kavika13:
Sorry for additional nitpicks. I randomly noticed the other spelling issue by chance in #1136.

I grepped and these two are only issues with `it's` I can find in the project. I figured I'd be thorough if I'm bothering to pipe up to begin with.

> Enable Vulkan Validation Layers:\nEnables a system that validates the state of the Vulkan renderer and logs information about it's internal state. This will reduce performance and likely change the behavior of emulation.

Should be:
> Enable Vulkan Validation Layers:\nEnables a system that validates the state of the Vulkan renderer and logs information about its internal state. This will reduce performance and likely change the behavior of emulation.

PR #1130, opened 2024-09-28
Add volume slider for title/background music

tGecko:
Added a volume slider for the title music.

image

The volume is stored as an integer ranging from 0 to 100 and is applied using a logarithmic scale to better match human hearing perception.


lukaszarczynski:
@tGecko These changes, as well as the previous MR related to this feature, are causing major regressions in Linux AppImages. Have you tested if AppImages generated by CI are working correctly?

PR #1137, opened 2024-09-28
Added sceKernelRmdir

ElBread3:
This PR adds sceKernelRmdir, a function often used by Unreal Engine 4 games such as Street Fighter V, as well as makes IOFile::Open() log the error message just like IOFile::Close()

### DIFF
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -310,35 +310,49 @@ int PS4_SYSV_ABI posix_mkdir(const char* path, u16 mode) {
     return result;
 }
 
-int PS4_SYSV_ABI sceKernelRmdir(const char* path, u16 mode) {
+int PS4_SYSV_ABI sceKernelRmdir(const char* path) {
     auto* mnt = Common::Singleton<Core::FileSys::MntPoints>::Instance();
     bool ro = false;
 
     const std::filesystem::path dir_name = mnt->GetHostPath(path, &ro);
 
     if (dir_name.empty()) {
+        LOG_INFO(Kernel_Fs, "Failed to remove directory: {}, permission denied", path);
         return SCE_KERNEL_ERROR_EACCES;
     }
 
     if (ro) {
+        LOG_INFO(Kernel_Fs, "Failed to remove directory: {}, directory is read only", path);
         return SCE_KERNEL_ERROR_EROFS;
     }
 
     if (!std::filesystem::is_directory(dir_name)) {
+        LOG_INFO(Kernel_Fs, "Failed to remove directory: {}, path is not a directory", path);
         return ORBIS_KERNEL_ERROR_ENOTDIR;
     }
 
     if (!std::filesystem::exists(dir_name)) {
-        return ORBIS_OK;
+        LOG_INFO(Kernel_Fs, "Failed to remove directory: {}, no such file or directory", path);
+        return ORBIS_KERNEL_ERROR_ENOENT;
     }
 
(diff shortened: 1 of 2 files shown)


src/core/libraries/kernel/file_system.cpp
```diff
@@ -310,6 +310,37 @@ int PS4_SYSV_ABI posix_mkdir(const char* path, u16 mode) {
...
+    if (ro) {
+        return SCE_KERNEL_ERROR_EROFS;
+    }
+
+    if (!std::filesystem::is_directory(dir_name)) {
+        return ORBIS_KERNEL_ERROR_ENOTDIR;
+    }
+
+    if (!std::filesystem::exists(dir_name)) {
+        return ORBIS_OK;
```
georgemoralis: that should return ORBIS_KERNEL_ERROR_ENOENT

src/core/libraries/kernel/file_system.cpp
```diff
@@ -310,6 +310,37 @@ int PS4_SYSV_ABI posix_mkdir(const char* path, u16 mode) {
...
+    }
+
+    int result = std::filesystem::remove_all(dir_name);
+    const int error = errno;
```
georgemoralis: why that??? that is needed only for posix functions

src/core/libraries/kernel/file_system.cpp
```diff
@@ -310,6 +310,37 @@ int PS4_SYSV_ABI posix_mkdir(const char* path, u16 mode) {
...
+
+    if (!std::filesystem::is_directory(dir_name)) {
+        return ORBIS_KERNEL_ERROR_ENOTDIR;
+    }
+
+    if (!std::filesystem::exists(dir_name)) {
+        return ORBIS_OK;
+    }
+
+    int result = std::filesystem::remove_all(dir_name);
```
georgemoralis: result is not used?

src/core/libraries/kernel/file_system.cpp
```diff
@@ -310,6 +310,37 @@ int PS4_SYSV_ABI posix_mkdir(const char* path, u16 mode) {
...
+    const int error = errno;
+
+    if (error == 0) {
+        return ORBIS_OK;
+    }
+    return ErrnoToSceKernelError(error); // error - 0x7ffe0000;
```
georgemoralis: that's not neccesary since it is not posix function

georgemoralis (changes requested):

src/core/libraries/kernel/file_system.cpp
StevenMiller123: Pretty sure sceKernelRmdir internally uses an rmdir posix function, so that might be where this came from.
image

src/core/libraries/kernel/file_system.cpp
ElBread3: Sorry, I interpreted this as a case where removing the directory was unnecessary as it already doesn't exist

src/core/libraries/kernel/file_system.cpp
ElBread3: forgot to edit that ty

src/core/libraries/kernel/file_system.cpp
ElBread3: should be resolved by new commit

src/core/libraries/kernel/file_system.cpp
ElBread3: should be resolved by new commit

raphaelthegreat (approved):

PR #1150, opened 2024-09-29
Fix fmt error

OFFTKP:


src/core/cpu_patches.cpp
```diff
@@ -1057,7 +1057,7 @@ static bool TryExecuteIllegalInstruction(void* ctx, void* code_address) {
                 // Undefined behavior if length + index is bigger than 64 according to the spec,
                 // we'll warn and continue execution.
                 LOG_WARNING(Core,
-                            "extrq at {:x} with length {} and index {} is bigger than 64, "
+                            "extrq at {} with length {} and index {} is bigger than 64, "
```
squidbus, line 1060: Shouldn't this just be `{:#x}` so it's still hex?

polybiusproxy (approved):

PR #1160, opened 2024-09-29
sdl: Fix use of functions that now return SDL_bool

squidbus:
Recent SDL update changed some functions to return `SDL_bool` instead of an integer return code. Adjust return value handling accordingly.

Fixes missing music audio in Project Diva Future Tone.

### DIFF
--- a/src/audio_core/sdl_audio.cpp
+++ b/src/audio_core/sdl_audio.cpp
@@ -101,14 +101,14 @@ s32 SDLAudio::AudioOutOutput(s32 handle, const void* ptr) {
         return 0;
     }
     // TODO mixing channels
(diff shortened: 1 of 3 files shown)


abouvier:
Even more recent SDL commits replaced `SDL_bool` by the native `bool` type. And `SDL_TRUE`/`SDL_FALSE` by `true`/`false`.

squidbus:
> Even more recent SDL commits replaced `SDL_bool` by the native `bool` type. And `SDL_TRUE`/`SDL_FALSE` by `true`/`false`.

Won't make a difference here since `SDL_bool` is an alias for `bool`, the way its handled in this PR would be unchanged.

PR #1164, opened 2024-09-30
vulkan: Use dynamic vertex buffer strides when dynamic bindings unavailable.

squidbus:
Currently when `VK_EXT_vertex_input_dynamic_state` is unsupported, the fallback fixed vertex bindings in the pipeline are not updated when user data affects the definition of the buffer by changing the stride or format. This is the root cause of a lot of vertex explosions and corrupted graphics on drivers like MoltenVK, across many games.

Instead, we can take a two-part approach:
* For changing stride, use dynamic vertex input stride from `VK_EXT_extended_dynamic_state`, which is supported on MoltenVK and any Vulkan 1.3 driver.
* For changing format, add the binding formats into the graphics key to create new pipelines.

This necessitates a minimum macOS version bump to 14, as dynamic vertex binding stride is not supported before Metal 3.1, which was introduced with macOS 14. Apple Silicon Macs already require 15, so this just affects some older Intel Macs that cannot upgrade to 14. I'm not sure how well this worked on those older models to begin with as I have never tested them, but guaranteeing higher Metal version support will make things a lot easier in general.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
@@ -247,6 +247,7 @@ bool PipelineCache::RefreshGraphicsKey() {
     key.blend_controls.fill({});
     key.write_masks.fill({});
     key.mrt_swizzles.fill(Liverpool::ColorBuffer::SwapMode::Standard);
+    key.vertex_buffer_formats.fill(vk::Format::eUndefined);
 
     // First pass of bindings check to idenitfy formats and swizzles and pass them to rhe shader
     // recompiler.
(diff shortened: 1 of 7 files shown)


src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
```diff
@@ -308,6 +309,26 @@ bool PipelineCache::RefreshGraphicsKey() {
         }
 
         std::tie(infos[i], modules[i], key.stage_hashes[i]) = GetProgram(stage, params, binding);
+
+        if (stage == Shader::Stage::Vertex && !instance.IsVertexInputDynamicState()) {
```
raphaelthegreat: Minor nit, this can be moved out of the loop and read info[1] directly

shinra-electric:
On macOS this fixes the vertex explosions in the animation just before the menu in P4G, and the 3d models are now visible in-game.

Main: 
Screenshot 2024-09-30 at 08 38 47
Screenshot 2024-09-30 at 08 39 08

This PR:
Screenshot 2024-09-30 at 08 35 27
Screenshot 2024-09-30 at 08 34 52

src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
squidbus: Done

PR #1165, opened 2024-09-30
Implement shader resource tables

baggins183:
Leaving this in draft for regression testing

Still needs a lot of refactoring. Also handling arbitrary readconst usage (not used in sharps).

### DIFF
--- a/src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
+++ b/src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
@@ -1,0 +1,267 @@
+
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <unordered_map>
+#include <boost/container/flat_map.hpp>
+#include <xbyak/xbyak.h>
+#include <xbyak/xbyak_util.h>
+#include "common/config.h"
+#include "common/io_file.h"
+#include "common/path_util.h"
+#include "common/singleton.h"
+#include "shader_recompiler/info.h"
+#include "shader_recompiler/ir/breadth_first_search.h"
+#include "shader_recompiler/ir/opcodes.h"
+#include "shader_recompiler/ir/passes/srt.h"
+#include "shader_recompiler/ir/program.h"
+#include "shader_recompiler/ir/reg.h"
+#include "shader_recompiler/ir/srt_gvn_table.h"
+#include "shader_recompiler/ir/value.h"
+#include "src/common/arch.h"
+#include "src/common/decoder.h"
+
+using namespace Xbyak::util;
+
+// TODO make sure no problems with identity and Insts being used in maps
+
+static void DumpSrtProgram(const Shader::Info& info, const u8* code, size_t codesize) {
+#ifdef ARCH_X86_64
+    using namespace Common::FS;
+
+    const auto dump_dir = GetUserPath(PathType::ShaderDir) / "dumps";
+    if (!std::filesystem::exists(dump_dir)) {
+        std::filesystem::create_directories(dump_dir);
+    }
+    const auto filename =
+        fmt::format("{}_{:#018x}_{}.srtprogram.txt", info.stage, info.pgm_hash, info.perm_idx);
+    const auto file = IOFile{dump_dir / filename, FileAccessMode::Write, FileType::TextFile};
+
+    u64 address = reinterpret_cast<u64>(code);
+    u64 code_end = address + codesize;
+    ZydisDecodedInstruction instruction;
+    ZydisDecodedOperand operands[ZYDIS_MAX_OPERAND_COUNT];
+    ZyanStatus status = ZYAN_STATUS_SUCCESS;
+    while (address < code_end && ZYAN_SUCCESS(Common::Decoder::Instance()->decodeInstruction(
+                                     instruction, operands, reinterpret_cast<void*>(address)))) {
+        std::string s =
+            Common::Decoder::Instance()->disassembleInst(instruction, operands, address);
+        s += "\n";
+        file.WriteString(s);
+        address += instruction.length;
+    }
+#endif
+}
+
+namespace {
+class SrtCodegen : public Xbyak::CodeGenerator {
+public:
+    SrtCodegen() : CodeGenerator(1_MB) {}
+};
+
+using namespace Shader;
+
+struct PassInfo {
+    // map offset to inst
+    using PtrUserList = boost::container::flat_map<u32, Shader::IR::Inst*>;
+
+    Optimization::SrtGvnTable gvn_table;
+    // keys are GetUserData or ReadConst instructions that are used as pointers
+    std::unordered_map<IR::Inst*, PtrUserList> pointer_uses;
+    // GetUserData instructions corresponding to sgpr_base of SRT roots
+    boost::container::small_flat_map<IR::ScalarReg, IR::Inst*, 1> srt_roots;
+
+    // pick a single inst for a given value number
+    std::unordered_map<u32, IR::Inst*> vn_to_inst;
+
+    // Bumped during codegen to assign offsets to readconsts
+    u32 dst_off_dw;
+
+    PtrUserList* GetUsesAsPointer(IR::Inst* inst) {
+        auto it = pointer_uses.find(inst);
+        if (it != pointer_uses.end()) {
+            return &it->second;
+        }
+        return nullptr;
+    }
+
+    // Return a single instruction that this instruction is identical to, according
+    // to value number
+    // The "original" is arbitrary. Here it's the first instruction found for a given value number
+    IR::Inst* DeduplicateInstruction(IR::Inst* inst) {
+        auto it = vn_to_inst.try_emplace(gvn_table.GetValueNumber(inst), inst);
+        return it.first->second;
+    }
+};
+} // namespace
+
+namespace Shader::Optimization {
+
+namespace {
+static IR::Value GetReadConstOff(const IR::Inst* inst) {
+    ASSERT(inst->GetOpcode() == IR::Opcode::ReadConst);
+    return inst->Arg(1);
+}
+
+static IR::ScalarReg GetUserDataSgprBase(const IR::Inst* inst) {
+    ASSERT(inst->GetOpcode() == IR::Opcode::GetUserData);
+    return inst->Arg(0).ScalarReg();
+}
+
+static inline void PushPtr(Xbyak::CodeGenerator& c, u32 off_dw) {
+    c.push(rdi);
+    c.mov(rdi, ptr[rdi + (off_dw << 2)]);
+    c.mov(r10, 0xFFFFFFFFFFFFULL);
+    c.and_(rdi, r10);
+}
+
+static inline void PopPtr(Xbyak::CodeGenerator& c) {
+    c.pop(rdi);
+};
+
+static void VisitPointer(u32 off_dw, IR::Inst* subtree, PassInfo& pass_info,
+                         Xbyak::CodeGenerator& c) {
+    PushPtr(c, off_dw);
+    PassInfo::PtrUserList* use_list = pass_info.GetUsesAsPointer(subtree);
+    ASSERT(use_list);
+
+    // First copy all the src data from this tree level
+    // That way, all data that was contiguous in the guest SRT is also contiguous in the
+    // flattened buffer.
+    // TODO src and dst are contiguous. Optimize with wider loads/stores
+    // TODO if this subtree is dynamically indexed, don't compact it (keep it sparse)
+    for (auto [src_off_dw, use] : *use_list) {
+        c.mov(r10d, ptr[rdi + (src_off_dw << 2)]);
+        c.mov(ptr[rsi + (pass_info.dst_off_dw << 2)], r10d);
+
+        use->SetFlags<u32>(pass_info.dst_off_dw);
+        pass_info.dst_off_dw++;
+    }
+
+    // Then visit any children used as pointers
+    for (const auto [src_off_dw, use] : *use_list) {
+        if (pass_info.GetUsesAsPointer(use)) {
+            VisitPointer(src_off_dw, use, pass_info, c);
+        }
+    }
+
+    PopPtr(c);
+}
+
+static void GenerateSrtProgram(Info& info, PassInfo& pass_info) {
+    Xbyak::CodeGenerator& c = *Common::Singleton<SrtCodegen>::Instance();
+
+    if (info.srt_info.srt_reservations.empty() && pass_info.srt_roots.empty()) {
+        return;
+    }
+
+    pass_info.dst_off_dw = NumUserDataRegs;
+
+    // Special case for V# step rate buffers in fetch shader
+    for (auto i = 0; i < info.srt_info.srt_reservations.size(); i++) {
+        PersistentSrtInfo::SrtSharpReservation res = info.srt_info.srt_reservations[i];
+        // get pointer to V#
+        c.mov(r10d, ptr[rdi + (res.sgpr_base << 2)]);
+
+        u32 src_off = res.dword_offset << 2;
+
+        for (auto j = 0; j < res.num_dwords; j++) {
+            c.mov(r11d, ptr[r10d + src_off]);
+            c.mov(ptr[rsi + (pass_info.dst_off_dw << 2)], r11d);
+
+            src_off += 4;
+            ++pass_info.dst_off_dw;
+        }
+    }
+
+    ASSERT(pass_info.dst_off_dw == info.srt_info.flattened_bufsize_dw);
+
+    for (const auto& [sgpr_base, root] : pass_info.srt_roots) {
+        VisitPointer(static_cast<u32>(sgpr_base), root, pass_info, c);
+    }
+
+    c.ret();
+    c.ready();
+
+    size_t codesize = c.getSize();
+    info.srt_info.walker = SmallCodeArray(c.getCode(), codesize);
+
+    if (Config::dumpShaders()) {
+        DumpSrtProgram(info, c.getCode(), codesize);
+    }
+
+    c.reset();
+
+    info.srt_info.flattened_bufsize_dw = pass_info.dst_off_dw;
+}
+
+}; // namespace
+
+void FlattenExtendedUserdataPass(IR::Program& program) {
+    Shader::Info& info = program.info;
+    PassInfo pass_info;
+
+    // traverse at end and assign offsets to duplicate readconsts, using
+    // vn_to_inst as the source
+    boost::container::small_vector<IR::Inst*, 32> all_readconsts;
+
+    for (auto r_it = program.post_order_blocks.rbegin(); r_it != program.post_order_blocks.rend();
+         r_it++) {
+        IR::Block* block = *r_it;
+        for (IR::Inst& inst : *block) {
+            if (inst.GetOpcode() == IR::Opcode::ReadConst) {
+                if (!GetReadConstOff(&inst).IsImmediate()) {
+                    continue;
+                }
+
+                all_readconsts.push_back(&inst);
+                if (pass_info.DeduplicateInstruction(&inst) != &inst) {
+                    // This is a duplicate of a readconst we've already visited
+                    continue;
+                }
+
+                IR::Inst* ptr_composite = inst.Arg(0).InstRecursive();
+
+                const auto pred = [](IR::Inst* inst) -> std::optional<IR::Inst*> {
+                    if (inst->GetOpcode() == IR::Opcode::GetUserData ||
+                        inst->GetOpcode() == IR::Opcode::ReadConst) {
+                        return inst;
+                    }
+                    return std::nullopt;
+                };
+                auto base0 = IR::BreadthFirstSearch(ptr_composite->Arg(0), pred);
+                auto base1 = IR::BreadthFirstSearch(ptr_composite->Arg(1), pred);
+                ASSERT_MSG(base0 && base1 && "ReadConst not from constant memory");
+
+                // TODO this probably requires some template magic to fix. BFS needs non-const
+                // variant. Needs to be non-const to change flags
+                IR::Inst* ptr_lo = base0.value();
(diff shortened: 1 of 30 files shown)


StevenMiller123:
> Leaving this in draft for regression testing

Seems like a lot of normal games are running the SRT code and regressing as a result. Clicker Heroes, Hatsune Miku Project Diva Future Tone, Persona 5 Royal, and Skullgirls 2nd Encore all seem to do this. 

Uncharted: The Nathan Drake Collection crashes in the same way as the prior titles, though it's a title that I believe is actually using SRTs, as it uses the S_LOAD_DWORD and S_LOAD_DWORDX2 instructions. 

Every one of these titles crash with the same exception, pictured below:
image

baggins183:
> Seems like a lot of normal games are running the SRT code

this is intentional. Any shader that uses indirection of a pointer stored in user data sgprs will take the same approach as SRT (generate a cpu program to flatten the data read by the shader via s_load_dword*)
Doing it 1 way simplifies the rest of the code.

But idk why the exceptions. Ill update this hopefully tomorrow to dump the generated cpu program so u can post that, unless u want to disas from the debugger.

Also if you turn on shader dumping it should dump some *.ir.txt files. Posting \<hash\>._pre_hoist.ir.txt and \<hash\>.pre_flatten.ir.txt would help.
Also the gcn

baggins183:
Thanks for testing

StevenMiller123:
@baggins183 I'm getting a slightly different exception on the newer builds. This is from running Uncharted: The Nathan Drake Collection.
image
image

Here's all the dumped files for the shader it's crashing on, compressed into a zip because I cannot directly send bin files.
[fs_0x000000007c219ffb_0.zip](https://github.com/user-attachments/files/17237700/fs_0x000000007c219ffb_0.zip)

baggins183:
the mov edi, dword ptr [rdi+30h] looks wrong (dst should be rdi).
also the rest of the movs should be dword, but these are probably harmess

Xbyak is doing the opposite of what i want when i say dword or qword explicitly lol

I added dumping for the srt program with zydis in the last commit. Doesnt seem to show width though (qword, dword, etc) in the disassembly. Gonna try to fix that

baggins183:
last commit might have fixed the edi thing

StevenMiller123:
Seems like Uncharted is still crashing on the same shader.
image

[fs_0x000000007c219ffb_0.zip](https://github.com/user-attachments/files/17239344/fs_0x000000007c219ffb_0.zip)

StevenMiller123:
Looks like we're back to that edi issue thing. I also see that the code shown in the debugger doesn't seem to align with the logged srtprogram, though I may just be missing something, as I'm not particularly familiar with this type of code.

Let me know if you'd like me to hold off on testing future commits. I don't want to spam this PR too much if you're already aware of the current issues.

baggins183:
nah i dont mind. Its definitely appreciated

baggins183:
~there might be a system dependent issue with xbyak for moving a 64 bit immediate to a register
because the initial PushPtr works on my os/cpu and should match the old path (readUD<>)
might need to do the mov(r10, 0xff...)  in 2 separate instructions for the lo and hi bits~

nvm i misread it. no idea why the edi is still there

baggins183:
The dumped srtprogram looks right. Im guessing the VS window looks wrong because it starts decoding from above the start of the function, which is garbage bytes. So a red herring.

Can u try dereferencing rdi when the crash happens?

So something like this (gdb)
p \*(AmdGpu::Image \*) $rdi

also p/x $rdi to show what the pointer in s[12:13] actually is

baggins183:
and then from the frame below (Info::RunSrtWalker)
can u post

\*(uint64_t\*) (user_data.data() + 12)
that will let me know if the hi bits are set. Might be useful.

    c.mov(r10, 0xFFFFFFFFFFFFULL);
    c.and_(rdi, r10);

is supposed to mask these off like Info::ReadUD

StevenMiller123:
Hmm, I can't get gdb working on Windows, and the crash doesn't seem to be occuring on Linux. If there's some other Windows debugger I could replicate this testing in, please let me know. For now, here are the CPU registers shown by the VSCode debugger when testing Uncharted.
image

As for the other request, I was able to log that to the shad log. The actual data in there seems to vary by game. 
Code for reference:
image

In Uncharted:
image
[Uncharted Log.txt](https://github.com/user-attachments/files/17251815/Uncharted.Log.txt)

In all the non-SRT games that crash, the value appears to always be 0. Here's a log from Clicker Heroes to validate that.
[Clicker Heroes Log.txt](https://github.com/user-attachments/files/17251816/Clicker.Heroes.Log.txt)

baggins183:
found it, the windows calling conventions are different 
https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170

"Integer arguments are passed in registers RCX, RDX, R8, and R9"

baggins183:
the problem was i assumed arg0 and arg1 go to rdi and rsi

baggins183:
think that will fix it

StevenMiller123:
Uncharted is now back to where it was on main, crashing on `[Debug] <Critical> data_share.cpp:operator():164: Assertion Failed!` after logging `Unknown opcode S_LOAD_DWORDX2`

Most of my non-SRT games are now crashing on `[Debug] <Critical> resource_tracking_pass.cpp:operator():349: Assertion Failed!` (so still regressed compared to main).

Let me know if you need any logs.

baggins183:
Weird, and theyre still regressed on linux? sorry to make u dual boot lol

StevenMiller123:
I believe the resource_tracking_pass assert was also getting hit on Linux, though I haven't retested on there with the latest pushes. Give me a moment to reboot, and I can confirm.

baggins183:
if u could zip all the shader dumps for a shader hash that happens on thatd help out
the .bin, .pre_hoist, .pre_flatten, and .srtprogram are the most important

baggins183:
> Unknown opcode S_LOAD_DWORDX2

This part is expected
Ill try it this weekend

StevenMiller123:
> if u could zip all the shader dumps for a shader hash that happens on thatd help out the .bin, .pre_hoist, .pre_flatten, and .srtprogram are the most important

Here's a shader from Clicker Heroes that's running into the assert.
[fs_0x00000000d66216a7_0.zip](https://github.com/user-attachments/files/17252735/fs_0x00000000d66216a7_0.zip)

baggins183:
This one looks like a phi problem with the ssa_rewrite pass

```
.L164_0:
/*0000000000a4*/ v_mov_b32       v2, s19
/*0000000000a8*/ s_load_dwordx4  s[12:15], s[12:13], 0x8
/*0000000000ac*/ v_mul_f32       v7, s18, v2
/*0000000000b0*/ s_waitcnt       lgkmcnt(0)
/*0000000000b4*/ s_buffer_load_dwordx8 s[20:27], s[12:15], 0x8
```

the s_load_dwordx4 writes to  s[12:13] which must be confusing the ssa pass.
Probably works on main because one of the BreadFirstSearch's must be brute forcing to find the original GetUserData that writes to SGPR 12

the walker program is wrong because of this 

definitely solvable though because theres nothing dynamic going on

baggins183:
`[0000562cee223808] %100   = Phi [ %27, {Block $0} ], [ %101, {Block $9} ] (uses: 2)`


```
[0000562cee216918] %207   = CompositeConstructU32x2 %205, %206 (uses: 8)
[0000562cee216988] %208   = ReadConst %207, #0 (uses: 2)
```


heres the bad phi
%205 is identity to %100
I should turn on extra id removal passes to make this less confusing

baggins183:
I think with this it should be at parity with main

baggins183:
actually there could be some problems, but might work

StevenMiller123:
I'll run my library through the latest version of this then, and I'll let you know if I find any remaining regressions.

baggins183:
u can hold off if u want, i think it will be broken. Will need to refactor to fix it

StevenMiller123:
Looks like the issue is still present in the new build, so I'll hold off on further tests.

baggins183:
thanks, will deal with this on the weekend

StevenMiller123:
Seems like I'm still hitting the same error. Here's an updated shader dump in case anything looks different.
[fs_0x00000000d66216a7_0.zip](https://github.com/user-attachments/files/17269857/fs_0x00000000d66216a7_0.zip)

squidbus:
Two issues I'm encountering testing this:
* Something about it is interfering with CPU instruction patches, I'm getting access violations when it tries to generate the trampoline for illegal instructions in guest code.
* Working around that by re-enabling some ahead-of-time patching code, a number of fragment shaders are hitting the unreachable in `ConvertImageType` with an invalid type 2.

ElBread3:
> a number of fragment shaders are hitting the unreachable in `ConvertImageType` with an invalid type 2.

Oh yeah I saw this before, in Lego City Undercover, I hope the fix for that could also fix Lego City Undercover as that is the only shader crash away from vanilla boot

squidbus:
> Oh yeah I saw this before, in Lego City Undercover, I hope the fix for that could also fix Lego City Undercover as that is the only shader crash away from vanilla boot

I'm not sure about that game, but just to be clear in my case the shaders with the issue are ones that worked before this.

baggins183:
The invalid sharp problems are related to phi instructions. I can probably hack around it but going to try to fix the underlying problem. That means fixing probably the control flow generation and the ssa_rewrite pass, might take a bit.

Main works because it uses breadth first search which brute forces through phi nodes to find the original readconst/getuserdata.

In the example I saw from Stephen, phis are being created even when theres 1 possible value for a given SGPR. The one i looked at showed a backedge being created from the end of the program (ENDPGM) to some label which shouldn't be possible.

Id guess the lego city ones are failing due to phi problems. But if BFS isn't working, there might be actual dynamic selection of sharps which would be hard to deal with.

baggins183:
> Something about it is interfering with CPU instruction patches, I'm getting access violations when it tries to generate the trampoline for illegal instructions in guest code.

sceGeoshadersfan posted about some other problem on windows, might be related. Right now i use new and delete wrong in the SmallCodeArray class. I need to alloc page-aligned memory to protect/unprotect it correctly and not interfere with other allocations. Turns out this is broken on windows, and also windows prevents an easy solution because they dont support std::aligned_allloc
So im currently using boost::aligned_alloc on my computer, but i think that needs to go in the boost submodule

raphaelthegreat:
Searching phi instructions when tracking sharps is fine, especially if the texture fetch is inside a loop. There are problems currently though due to control flow restructuring where multiple sources could be possible to find from phi nodes, even when in original assembly only 1 path is possible. Main picks the first possible it finds from BFS which in these cases at best results in a ReadConst crash or at worst picking a completely wrong sharp.

A relatively easy way to solve without messing with control flow is to perform a simple reachability test in the original CFG that mirrors the assembly control flow. I.e collect all source candidates from a complete search and test if the current block is reachable from the source block (the guest address of a ReadConst and be stored in its flags and used to lookup the block in the original CFG). In most cases this should resolve down to 1 possible access, but conditional sharp loads are possible to emit by the compiler and those need special handling for later

The long term solution to this would be to rework control flow to allow else clauses in the structured control flow. This limits the amounts of paths the ssa path deems as reachable and will prevent some of these cases. However I'm not very aware of many reliable algorithms with this feature. Red Prig has mentioned "Goto Elimination Strategies in the Migration of Legacy Code to Java" which might be interesting to investigate.

So I think for now to avoid doing too many invasive changes messing with control flow can be left for later probably

src/shader_recompiler/ir/passes/srt.h
```diff
@@ -26,15 +26,35 @@ namespace Shader {
...
+    }
+
+    void resize(size_t new_size_dw) {
+        buf.resize(new_size_dw);
+    }
+
+private:
+    // TODO aligned_alloc with ubo min alignment (dynamic)
+    // 256 for now to be conservative
+    boost::container::vector<u32, boost::alignment::aligned_allocator<u32, 256>> buf;
```
raphaelthegreat: Since you are copying it to the stream buffer alignment of the host memory probably doesn't matter. Its not bad idea though

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+    if (info.srt_info.srt_reservations.empty() && pass_info.srt_roots.empty()) {
+        return;
+    }
+
+    pass_info.dst_off_dw = NumUserDataRegs;
+
+    // Special case for V# step rate buffers in fetch shader
+    for (auto i = 0; i < info.srt_info.srt_reservations.size(); i++) {
+        PersistentSrtInfo::SrtSharpReservation res = info.srt_info.srt_reservations[i];
```
raphaelthegreat:
```suggestion
    for (const auto [sgpr_base, dword_offset, num_dwords] : info.srt_info.srt_reservations) {
```

src/shader_recompiler/ir/passes/srt.h
```diff
@@ -0,0 +1,126 @@
...
+
+#include <memory>
+#include "common/alignment.h"
+#include "common/assert.h"
+#include "common/types.h"
+#include "xbyak/xbyak.h"
+
+namespace Shader {
+
+class FlattenedUserDataBuffer {
```
raphaelthegreat: This abstraction probably is superfluous and can be a std::vector<u32> directly in info. Or a type alias ```using FlattenedUserDataBuffer = std::vector<u32>;``` if it makes sense

src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -290,6 +290,13 @@ void BufferCache::InlineDataToGds(u32 gds_offset, u32 value) {
     cmdbuf.updateBuffer(gds_buffer.Handle(), gds_offset, sizeof(u32), &value);
 }
 
+std::pair<Buffer*, u32> BufferCache::ObtainHostUBO(VAddr host_addr, u32 size) {
```
raphaelthegreat: If buffer is std::vector you can pass std::span\<const u32\> to this function

src/shader_recompiler/ir/passes/srt.h
```diff
@@ -0,0 +1,126 @@
...
+    void resize(size_t new_size_dw) {
+        buf.resize(new_size_dw);
+    }
+
+private:
+    std::vector<u32> buf;
+};
+
+typedef void(__attribute__((sysv_abi)) * PFN_SrtWalker)(const u32* /*user_data*/,
+                                                        u32* /*flat_dst*/);
```
raphaelthegreat: ```using PFN_SrtWalker = void PS4_SYSV_ABI (*)(const u32* /*user_data*/, u32* /*flat_dst*/);```

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -0,0 +1,267 @@
...
+            Common::Decoder::Instance()->disassembleInst(instruction, operands, address);
+        s += "\n";
+        file.WriteString(s);
+        address += instruction.length;
+    }
+#endif
+}
+
+namespace {
+class SrtCodegen : public Xbyak::CodeGenerator {
```
raphaelthegreat: This is superfluous abstraction I think, can use Xbyak::CodeGenerator directly. Also it can be constructed each time its used, doesn't need to be a global singleton if memory is managed externally. Or you could use a static Xbyak::Codegen in this file if the memory can be managed by xbyak during the duration of the program (I think easier overall)

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -0,0 +1,267 @@
...
+    // traverse at end and assign offsets to duplicate readconsts, using
+    // vn_to_inst as the source
+    boost::container::small_vector<IR::Inst*, 32> all_readconsts;
+
+    for (auto r_it = program.post_order_blocks.rbegin(); r_it != program.post_order_blocks.rend();
+         r_it++) {
+        IR::Block* block = *r_it;
+        for (IR::Inst& inst : *block) {
+            if (inst.GetOpcode() == IR::Opcode::ReadConst) {
+                if (!GetReadConstOff(&inst).IsImmediate()) {
```
raphaelthegreat: ```const IR::Value offset = inst.Arg(1);``` and can remove the GetReadConstOff function likely

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -0,0 +1,267 @@
...
+                ptr_lo = pass_info.DeduplicateInstruction(ptr_lo);
+
+                auto ptr_uses_kv =
+                    pass_info.pointer_uses.try_emplace(ptr_lo, PassInfo::PtrUserList{});
+                PassInfo::PtrUserList& user_list = ptr_uses_kv.first->second;
+
+                user_list[GetReadConstOff(&inst).U32()] = &inst;
+
+                if (ptr_lo->GetOpcode() == IR::Opcode::GetUserData) {
+                    IR::ScalarReg ud_reg = GetUserDataSgprBase(ptr_lo);
```
raphaelthegreat: This is only used once, can be replaced with ```ptr_lo->Arg(0).ScalarReg()```

src/shader_recompiler/ir/passes/srt.h
```diff
@@ -0,0 +1,126 @@
...
+
+// Utility for copying a simple relocatable function from a Xbyak code generator to manage memory
+// separately
+class SmallCodeArray {
```
raphaelthegreat: This class also raises some questions if its useful or not. If we manage memory externally we can emit to it directly or just let xbyak manage it with a static instance, either way copying can probably be avoided

src/shader_recompiler/ir/passes/srt.h
```diff
@@ -0,0 +1,126 @@
...
+    };
+
+    SmallCodeArray walker;
+    boost::container::small_vector<SrtSharpReservation, 2> srt_reservations;
+    u32 flattened_bufsize_dw;
+
+    // Special case for fetch shaders because we don't generate IR to read from step rate buffers,
+    // so we won't see usage with GetUserData/ReadConst.
+    // Reserve space in the flattened buffer for a sharp ahead of time
+    u32 reserve_sharp(u32 sgpr_base, u32 dword_offset, u32 num_dwords) {
```
raphaelthegreat: ReserveSharp

src/shader_recompiler/ir/passes/srt.h
```diff
@@ -0,0 +1,126 @@
...
+    // Special case when fetch shader uses step rates.
+    struct SrtSharpReservation {
+        u32 sgpr_base;
+        u32 dword_offset;
+        u32 num_dwords;
+    };
+
+    SmallCodeArray walker;
+    boost::container::small_vector<SrtSharpReservation, 2> srt_reservations;
+    u32 flattened_bufsize_dw;
```
raphaelthegreat: Can do ```u32 flattened_bufsize_dw = 16;``` and remove constructor

src/shader_recompiler/ir/srt_gvn_table.h
```diff
@@ -0,0 +1,157 @@
...
+        }
+        return iv;
+    }
+
+    // Temp workaround for something like this:
+    // [0000555558a5baf8] %297   = Phi [ %24, {Block $1} ], [ %297, {Block $5} ] (uses: 4)
+    // [0000555558a4e038] %305   = CompositeConstructU32x2 %297, %296 (uses: 4)
+    // [0000555558a4e0a8] %306   = ReadConst %305, #0 (uses: 2)
+    // Should probably be fixed in ssa_rewrite
+    std::optional<IR::Value> TryRemoveTrivialPhi(IR::Inst* phi) {
```
raphaelthegreat, line 125: A better workaround might be to add a post-pass to ssa rewrite that removes those dead phis after the pass has run. At least then this file doesn't need to deal with it and we can fix it in faster way when IR gains ability to track users

src/shader_recompiler/ir/passes/srt.h
baggins183: > let xbyak manage it with a static instance

I dont think theres a way to garbage collect old programs if we do this
I thought if u have permutation blow up, then this might become a problem
thats the main reason i did this

But SmallCodeArray is probably superfluous vs having 1 Xbyak::CodeGenerator : 1 Shader::Info.
The only pro to SmallCodeArray is being smaller (forgot exactly why i did it), so probably not worth

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
baggins183: you can make the call on this. Personally id either use SmallCodeArray or a single Xbyak::CodeGenerator per Shader::Info.
Although these are inneficient since they need a page at minimum per program, and programs need separate allocations for garbage collection purposes.
my last choice would be using a single static Codegenerator that never gets reset cuz of the permutation problem

baggins183:
windows build is failing due to some winpthreads problem with "MemoryBarrier", which windows defines as a macro. No idea how my PR is causing this, somehow it doesnt happen on main

vulkan.hpp #undefs MemoryBarrier but that shouldn't be included by thread_management.cpp anyways

AboMedoz:
> windows build is failing due to some winpthreads problem with "MemoryBarrier", which windows defines as a macro. No idea how my PR is causing this, somehow it doesnt happen on main
> 
> vulkan.hpp #undefs MemoryBarrier but that shouldn't be included by thread_management.cpp anyways


image
you need to replace MemoryBarrier() with __faststorefence in pthread.h to build on windows

src/shader_recompiler/ir/srt_gvn_table.h
baggins183: this is the reason the trivial phi is still there, right? (from the linked paper)
> _Moreover, if a φ function could be successfully replaced, other φ functions
using this replaced value might become trivial as well. For this reason, we apply
this simplification recursively on all of these users_

so we need to apply this on the users of the phi if TryRemoveTrivialPhi succeeds, and so on until we reach a fixed point?

baggins183: I prefer keeping this part of the PR the same until user tracking, otherwise I have to refactor ssa_rewrite which I dont want to break

PR #1173, opened 2024-10-01
fix wolf2022 cheats download

DanielSvoboda:
in short, this will download more wolf2022 cheats if available, not just 1

Before, it only downloaded 1 cheat file from the wolf2022 server,
but there may be more than one file, as in the print below, it only downloaded the file CUSA04885_01.00_wolf2022.json which was made by Tyler, but as you can see in the second print there is another cheat from another author TLH CUSA04885_1.00_wolf2022.json.
The version number does not have the 0 at the beginning and in a way this was ignored.

*this affects 79 games, not just the one I'm showing
<div style="display: flex;">
    Before
    After
</div>

The file name was also added to the message asking if you want to replace it because there is already a file with the same name.
image


gandalfthewhite19890404:
Yes, it was fixed, thanks!

PR #1176, opened 2024-10-01
config: Add option to change DLC install path.

squidbus:
Adds a config option to change the DLC install path. It's useful to be able to point this to a directory of DLC contents independent of your config and other user data, same as how the game path works.

For users with existing config files, the current DLC path is defaulted to. When a user is opening the emulator for the first time, a DLC directory option will appear next to the game directory option, with the path inside the user directory filled in as the default in case the user does not care.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -8,6 +8,7 @@
 #include <fmt/xchar.h> // for wstring support
 #include <toml.hpp>
 #include "common/logging/formatter.h"
+#include "common/path_util.h"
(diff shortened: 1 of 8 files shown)


AussiePlick:
Hello Squidbus. Should the DLC PKG's be extracted or left as PKG files?
And many thank you's for this...

squidbus:
> Should the DLC PKG's be extracted for left as PKG files?

The PKGs are extracted by the emulator into the install directory.

PR #1178, opened 2024-10-01
added discord rpc

delledev:


delledev:
I could use some help with linting. Its my first time with clang-format and im having trouble with it.

LNDF:
Do clang-format -i filepath to automatically correct it's formatting.

delledev:
Ty doing that now.

delledev:
Hopefully it's fixed now.  I've removed the old rpc files from common and added a new one to QT Gui, i think this is better than messing with the core emulator. 
Btw this is how it looks. 
img0
img1

delledev:
Before merging someone with higher clearence should change the application id inside discord_rpc_handler.cpp.

Hermiten:
Nice !
I'm not a dev to test it but it's look great

.gitmodules
```diff
@@ -94,4 +94,7 @@
 [submodule "externals/pugixml"]
 	path = externals/pugixml
 	url = https://github.com/zeux/pugixml.git
-	shallow = true
\ No newline at end of file
+	shallow = true
+[submodule "externals/discord-rpc"]
+	path = externals/discord-rpc
+	url = https://github.com/delledev/discord-rpc
```
abouvier: You should use https://github.com/shadps4-emu/ext-discord-rpc

delledev: Thank you for your input. I've changed it.

CMakeLists.txt
```diff
@@ -838,4 +840,10 @@ if (UNIX AND NOT APPLE)
         find_package(OpenSSL REQUIRED)
         target_link_libraries(shadps4 PRIVATE ${OPENSSL_LIBRARIES})
     endif()
+endif()
+
+if(ENABLE_QT_GUI)
+    add_subdirectory(externals/discord-rpc)
+    target_link_libraries(shadps4 PRIVATE discord-rpc)
+    include_directories(externals/discord-rpc/include)
```
abouvier: You should use `target_include_directories` and set `BUILD_EXAMPLES` to `OFF`. And also probably put all this, except `target_link_libraries`, in `externals/CMakeLists.txt`.

delledev: Sorry it's my first time with Cmake. Does ENABLE_QT_GUI check works inside externals/CmakeLists.txt ?

abouvier: Yes, variables are inherited in subdirectories by default.

src/qt_gui/discord_rpc_handler.cpp
```diff
@@ -0,0 +1,43 @@
...
+    std::string discordAppId = "1290207945476280360";
+    Discord_Initialize(discordAppId.c_str(), &handlers, 1, nullptr);
+    startTimestamp = time(nullptr);
+    enabled = true;
+}
+
+void RPC::setStatusIdling() {
+    DiscordRichPresence rpc{};
+    rpc.largeImageKey = "https://github.com/shadps4-emu/shadPS4/raw/main/.github/shadps4.png";
+    rpc.largeImageText = "ShadPS4 is a PS4 emulator";
```
polybiusproxy:
```diff
-    rpc.largeImageText = "ShadPS4 is a PS4 emulator";
+    rpc.largeImageText = "shadPS4 is a PS4 emulator";
```

src/qt_gui/discord_rpc_handler.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <cstring>
+#include <ctime>
+#include "src/qt_gui/discord_rpc_handler.h"
+
+namespace DiscordRPCHandler {
+
+void RPC::init() {
+    DiscordEventHandlers handlers{};
+    std::string discordAppId = "1290207945476280360";
+    Discord_Initialize(discordAppId.c_str(), &handlers, 1, nullptr);
```
polybiusproxy: App ID can simply be directly on the function argument, no need to initialize a std::string

polybiusproxy (commented):
Looks good on a first glance. Got a few requests though - might be better if this could be configurable (can be enabled/disabled on Qt settings), and make it so it can also be used on SDL builds.

src/qt_gui/discord_rpc_handler.cpp
delledev: I dont think it's a good idea to make it an argument. If for some reason we need to initalize it elsewhere it would be extra work to keep all the ids updated. Turning it into a string literal might be better, I think.

polybiusproxy: If that becomes the case later on, we can always change it back.

delledev:
> Looks good on a first glance. Got a few requests though - might be better if this could be configurable (can be enabled/disabled on Qt settings), and make it so it can also be used on SDL builds.

I'm heavily inexperienced with C++ and Qt, so I don't really understand what you mean by this. But if anyone else wants to help, their help is much appreciated.

CMakeLists.txt
```diff
@@ -838,4 +840,7 @@ if (UNIX AND NOT APPLE)
         find_package(OpenSSL REQUIRED)
         target_link_libraries(shadps4 PRIVATE ${OPENSSL_LIBRARIES})
     endif()
-endif()
\ No newline at end of file
+endif()
+
+#Discord RPC
```
polybiusproxy:
```diff
-#Discord RPC
+# Discord RPC
```

externals/CMakeLists.txt
```diff
@@ -182,4 +182,9 @@ add_subdirectory(tracy)
...
 if (NOT TARGET pugixml::pugixml)
     add_subdirectory(pugixml)
-endif()
\ No newline at end of file
+endif()
+
+#Discord RPC
+    set(BUILD_EXAMPLES OFF)
+    add_subdirectory(discord-rpc/)
+    target_include_directories(discord-rpc INTERFACE discord-rpc/include)
```
polybiusproxy: Proper formatting here please:
```diff
-#Discord RPC
-    set(BUILD_EXAMPLES OFF)
-    add_subdirectory(discord-rpc/)
-    target_include_directories(discord-rpc INTERFACE discord-rpc/include)
+# Discord RPC
+set(BUILD_EXAMPLES OFF)
+add_subdirectory(discord-rpc/)
+target_include_directories(discord-rpc INTERFACE discord-rpc/include)
```

src/common/config.cpp
```diff
@@ -422,6 +431,7 @@ void load(const std::filesystem::path& path) {
         isFullscreen = toml::find_or<bool>(general, "Fullscreen", false);
         playBGM = toml::find_or<bool>(general, "playBGM", false);
         BGMvolume = toml::find_or<int>(general, "BGMvolume", 50);
+        enableDiscordRPC = toml::find_or<bool>(general, "enableDiscordRPC", false);
```
polybiusproxy: Might be best to have it enabled by default here

polybiusproxy (changes requested):
Looks good now. Only a few things to take care of and this should be ready to merge.
It might be best to use our Singleton class for the RPC handler instead of making a `getInstance` function.

delledev:
> Looks good now. Only a few things to take care of and this should be ready to merge.
> It might be best to use our Singleton class for the RPC handler instead of making a `getInstance` function.

Could you elaborate on this. I needed to make it so the same instance can be accesible by MainWindow and Emulator. How can I do this without getInstance ?

polybiusproxy:
I talk about the common [Singleton](https://github.com/shadps4-emu/shadPS4/blob/main/src/common/singleton.h) class we have

polybiusproxy (approved):
Looks good to me, thanks!

par274:
Does not work for me

PR #1184, opened 2024-10-01
ci: add missing libs to enable Wayland backend for Qt and SDL

ngoquang2708:
Currently, only X11 is detected and enabled for SDL3 by CMake. This PR enables the Wayland backend for SDL.

### DIFF
--- a/.github/linux-appimage-qt.sh
+++ b/.github/linux-appimage-qt.sh
@@ -9,6 +9,8 @@ fi
 
 export Qt6_DIR="/usr/lib/qt6"
 export PATH="$Qt6_DIR/bin:$PATH"
+export EXTRA_QT_PLUGINS="waylandcompositor"
+export EXTRA_PLATFORM_PLUGINS="libqwayland-egl.so;libqwayland-generic.so"
 
 # Prepare Tools for building the AppImage
 wget -q https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -291,7 +291,7 @@ jobs:
         submodules: recursive
 
     - name: Install dependencies
-      run: sudo apt-get update && sudo apt install -y libx11-dev libxext-dev libwayland-dev libfuse2 clang build-essential libasound2-dev libpulse-dev libopenal-dev
+      run: sudo apt-get update && sudo apt install -y libx11-dev libxext-dev libwayland-dev libdecor-0-dev libxkbcommon-dev libglfw3-dev libgles2-mesa-dev libfuse2 clang build-essential libasound2-dev libpulse-dev libopenal-dev
 
     - name: Cache CMake Configuration
       uses: actions/cache@v4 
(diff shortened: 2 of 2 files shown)


ngoquang2708:
This still produces a black screen (same as previous) on the Wayland session because, I think, SDL defaults to Xwayland. So I have to force Wayland by unset the ```DISPLAY``` env var without rely on the use of ```gamescope```:
```
env -u DISPLAY ./shadps4
```

Caliel666:
tested on gnome wayland ~~after applying the patch for bb~~ it works almost perfectly!, it does sometimes crash when loading a game when that usually doesn't happen on x11 but not sure if that's relevant for this fix:
```shell
[Render.Vulkan] <Error> vk_platform.cpp:DebugUtilsCallback:68: VUID-vkQueueSubmit-pWaitSemaphores-03238: Validation Error: [ VUID-vkQueueSubmit-pWaitSemaphores-03238 ] Object 0: handle = 0x980f360000000011, type = VK_OBJECT_TYPE_SEMAPHORE; Object 1: handle = 0x5882a73c2700, type = VK_OBJECT_TYPE_QUEUE; | MessageID = 0xb50452b0 | vkQueueSubmit(): pSubmits[0].pWaitSemaphores[0] queue (VkQueue 0x5882a73c2700[]) is waiting on semaphore (VkSemaphore 0x980f360000000011[]) that has no way to be signaled. The Vulkan spec states: All elements of the pWaitSemaphores member of all elements of pSubmits created with a VkSemaphoreType of VK_SEMAPHORE_TYPE_BINARY must reference a semaphore signal operation that has been submitted for execution and any semaphore signal operations on which it depends must have also been submitted for execution (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkQueueSubmit-pWaitSemaphores-03238)
terminate called after throwing an instance of 'std::system_error'
  what():  Invalid argument```

**Comment by ngoquang2708** [comment]
Yeah , Wayland is not perfect in SDL3 at the moment.

**Comment by ngoquang2708** [comment]
But at least this enable us to use native Wayland which is not possible before.

**Comment by h16033** [comment]
> Yeah , Wayland is not perfect in SDL3 at the moment.
 work with gamescope

**Comment by h16033** [comment]
work with gamescope

**Comment by ngoquang2708** [comment]
You have to use Gamescope previously because SDL3 use XWayland by default under Wayland, causing game black screen.
This PR allows you to run the emulator in native Wayland without using Gamescope though. But you have to force Wayland by unset the ```DISPLAY``` environment variable because SDL3 select XWayland by default when run under Wayland.
```
env -u DISPLAY ./Shadps4-qt.AppImage
```

qurious-pixel:
@ngoquang2708 I recommend you test the changes for the Qt branch on non-wayland systems as well.
Usually the libs allow either Wayland or X, but not both. 
So you may need to include logic to determine which libs to load, with LD_PRELOAD depending on the windowing system.

ngoquang2708:
@qurious-pixel

IFAIK, SDL and Qt both support X11 and Wayland, and they can dynamically select which backend to use based on the current desktop session.

Without this PR, the current AppImage doesn't include:
* Qt's plugins required for Wayland support, to run the GUI as native Wayland apps.
* SDL support for Wayland to allow the emulator itself to show as a native Wayland window.
and will run on Wayland session via XWayland.

The Qt GUI ran fine on both X11 and Wayland (via XWayland) desktop session.
The game window however got blackscreen on Wayland (via XWayland) desktop session and require the use of Gamescope as a workround.

With this PR, the Qt GUI can run as a native Wayland app on Wayland session. The emulator window through its logic, however, still use XWayland on Wayland session. But we now have an option to force it to use native Wayland without the use of Gamescope.

7oxicshadow:
@ngoquang2708 nice work. I can confirm that it works for me on Fedora 40 (Wayland) with Nvidia 560.35.03

ngoquang2708:
@7oxicshadow Do you have to force Wayland by unset the ```DISPLAY``` env?

7oxicshadow:
@ngoquang2708 Yes. Without it I still get a black screen

PR #1196, opened 2024-10-02
core: Add support for Neo mode memory size.

squidbus:
Adds support for the larger PS4 Pro memory size when enabled. The value is not confirmed to be exact with hardware but aligns with expectations and at least gets games that utilize the extra memory running under Pro mode for now.

I also added an assert when the GPU page manager tries to track memory that is not mapped to the GPU, as I ran into a problem related to this while testing PS4 Pro mode. Currently when this happens it silently re-protects non-GPU memory, which can fail in confusing ways later as writes to that memory will be ignored by the signal handler and fail. Adding an assert makes it more obvious what happened.

### DIFF
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -25,7 +25,7 @@ asm(".zerofill GUEST_SYSTEM,GUEST_SYSTEM,__guest_system,0xFBFC00000");
 
 namespace Core {
 
-static constexpr size_t BackingSize = SCE_KERNEL_MAIN_DMEM_SIZE;
+static constexpr size_t BackingSize = SCE_KERNEL_MAIN_DMEM_SIZE_PRO;
 
 #ifdef _WIN32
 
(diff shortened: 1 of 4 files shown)


bigol83:
Red Dead Redemption with 1.04 patch enabling Ps4pro support and the option enabled in emulator hangs just before reaching logos

[shad_log.txt](https://github.com/user-attachments/files/17228078/shad_log.txt)

squidbus:
> Red Dead Redemption with 1.04 patch enabling Ps4pro support and the option enabled in emulator hangs just before reaching logos

This doesn't really do anything but fix memory size asserts for Pro mode, so if there are other issues they're not expected to change.

squidbus:
> I thought both had 8GB, just the pro having faster ram

It's less about the physical RAM size and more the amount made available to games.

Foul-Tarnished:
Yeah, I guess the system has less on pro, weird they didn't add this to
base model with software update

PR #1199, opened 2024-10-02
shader_recompiler: Shader param fixups

squidbus:
* Use `GetAttributeInfo` to construct attribute infos more to simplify correct construction.
* Allow passing `num_components` to `GetAttributeInfo`. Currently still always 4 but more correct this way.
* Add asserts to make sure input parameters array is not overrun.
* Fix missing boolean parameter and incorrect pointer type in one of the input attribute declarations.
* Use existing constants for params and fragment outputs array sizes.
* Misc style cleanup.


squidbus:
Fixes regression in CUSA05133

PR #1200, opened 2024-10-02
Refactor audio handling with range checks, buffer threshold, and lock…

Mikasa-san:
… fixes

- Added range checks for handle to avoid invalid index access in AudioOutOutput, AudioOutSetVolume, and AudioOutGetStatus.
 
- Added a constant AUDIO_STREAM_BUFFER_THRESHOLD for the buffer threshold (was previously a magic number).
 
- Set the freq parameter correctly in the SDL_AudioSpec structure in AudioOutOpen.

- Fixed locking issues in AudioOutOutput to avoid unlocking before it's locked.

Edit

- Ensure that port.stream and other large objects are being managed efficiently, either with smart pointers or explicit cleanup to avoid memory leaks.

- Removes tab spaces to fix clang-format error

### DIFF
--- a/src/audio_core/sdl_audio.cpp
+++ b/src/audio_core/sdl_audio.cpp
@@ -14,150 +14,168 @@
 
 namespace Audio {
 
+constexpr int AUDIO_STREAM_BUFFER_THRESHOLD = 65536; // Define constant for buffer threshold
+
 int SDLAudio::AudioOutOpen(int type, u32 samples_num, u32 freq,
-                           Libraries::AudioOut::OrbisAudioOutParamFormat format) {
-    using Libraries::AudioOut::OrbisAudioOutParamFormat;
-    std::unique_lock lock{m_mutex};
-    for (int id = 0; id < portsOut.size(); id++) {
-        auto& port = portsOut[id];
-        if (!port.isOpen) {
-            port.isOpen = true;
-            port.type = type;
-            port.samples_num = samples_num;
-            port.freq = freq;
-            port.format = format;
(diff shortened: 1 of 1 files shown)


src/audio_core/sdl_audio.cpp
```diff
@@ -92,33 +94,45 @@ int SDLAudio::AudioOutOpen(int type, u32 samples_num, u32 freq,
...
+    while (SDL_GetAudioStreamAvailable(port.stream) > AUDIO_STREAM_BUFFER_THRESHOLD) {
         SDL_Delay(0);
     }
 
     return result ? ORBIS_OK : -1;
 }
 
 bool SDLAudio::AudioOutSetVolume(s32 handle, s32 bitflag, s32* volume) {
+    if (handle < 1 || handle > portsOut.size()) { // Add handle range check
+        return ORBIS_AUDIO_OUT_ERROR_INVALID_PORT;
```
squidbus, line 126: I don't think this will work as intended because this returns `bool`.

src/audio_core/sdl_audio.cpp
```diff
@@ -152,6 +166,10 @@ bool SDLAudio::AudioOutSetVolume(s32 handle, s32 bitflag, s32* volume) {
 }
 
 bool SDLAudio::AudioOutGetStatus(s32 handle, int* type, int* channels_num) {
+    if (handle < 1 || handle > portsOut.size()) { // Add handle range check
+        return ORBIS_AUDIO_OUT_ERROR_INVALID_PORT;
```
squidbus, line 170: Same here.

PR #1203, opened 2024-10-02
Fix title music playing when it shouldn't

### LINKED ISSUE #1198: Title music always on - MacOS
Independent of on/off in the settings dialog the title music is always on.

tGecko:
Somehow I've missed that. Fixes #1198


Foul-Tarnished:
@tGecko 
It seems there's always an audio feed being sent regardless of the setting.

By that I mean bluetooth devices are always receiving empty audio, so they "hiss" with no sound, instead of getting into sleep mode when not used. (reported on discord by @Jojo1)

PR #1209, opened 2024-10-02
Stub return value of sceNpCreateRequest

mikusp:
`sceNpCreateRequest` should return a positive value, this PR makes it return increasing numbers.

Makes RAGE2 [CUSA10300] reach title screen
Screenshot From 2024-10-02 20-18-54


georgemoralis:
looking ok but probably we should look for a real implementation so i will invest it a bit

PR #1218, opened 2024-10-03
core/gnmdriver: Proper interrupt registering

polybiusproxy:
Improves Dreams and Unreal Engine games.

### DIFF
--- a/src/core/platform.h
+++ b/src/core/platform.h
@@ -17,31 +17,35 @@
 namespace Platform {
 
 enum class InterruptId : u32 {
-    Compute0RelMem = 0u,
-    Compute1RelMem = 1u,
-    Compute2RelMem = 2u,
-    Compute3RelMem = 3u,
-    Compute4RelMem = 4u,
-    Compute5RelMem = 5u,
-    Compute6RelMem = 6u,
-    GfxEop = 7u,
-    GfxFlip = 8u,
-    GpuIdle = 9u,
+    Compute0RelMem = 0x00,
+    Compute1RelMem = 0x01,
+    Compute2RelMem = 0x02,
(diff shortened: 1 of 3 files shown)


georgemoralis:
by improving we mean?

polybiusproxy:
Games calling `sceGnmAddEqEvent` with another event ID (e.g. `Compute6RelMem` on Unreal Engine), then GfxEop or vice versa would assert and crash due to register of GfxEop event twice

src/core/platform.h
```diff
@@ -81,7 +87,7 @@ struct IrqController {
         std::queue<IrqHandler> one_time_subscribers{};
         std::mutex m_lock{};
     };
-    std::array<IrqContext, magic_enum::enum_count<InterruptId>()> irq_contexts{};
+    std::unordered_map<InterruptId, IrqContext> irq_contexts{};
```
raphaelthegreat, line 90: The number of interrupts is still quite small so an array will be faster than unoredered_map

polybiusproxy: I don't see how to do this with an array here. What do you think would be the best approach?
I'm not sure of how much are these called so I don't know how vital performance is here

raphaelthegreat: Before this change the array was indexed with InterruptId, now it was changed to hash map, so unless there is a new behavior that relies on the hashmap I don't see why abandoning array is needed

polybiusproxy: I made it a map because I wasn't sure how to do indexing with the array without having a special if case for GfxEop

raphaelthegreat: Can InterruptIdMax be increased to 0x41 so it works

PR #1227, opened 2024-10-03
Using a more standard data directory for linux

robyn-dressler:
Currently, running the emulator will generate a "user" directory wherever it is being run from. This can cause issues when the program was installed as a package on linux, because the user folder will end up in the user's home directory. In my case, it caused a strange interaction with the dolphin emulator, because dolphin started using shadPS4's user directory to save data. Using `$HOME/.local/share/shadPS4` should avoid any clashes with other programs.

### DIFF
--- a/src/common/path_util.cpp
+++ b/src/common/path_util.cpp
@@ -95,6 +95,14 @@ static auto UserPaths = [] {
         user_dir =
             std::filesystem::path(getenv("HOME")) / "Library" / "Application Support" / "shadPS4";
     }
+#elif defined(__linux__)
+    auto user_dir = std::filesystem::current_path() / PORTABLE_DIR;
+    // Check if the "user" directory exists in the current path:
+    if (!std::filesystem::exists(user_dir)) {
+        // If it doesn't exist, use the new hardcoded path:
+        user_dir =
+            std::filesystem::path(getenv("HOME")) / ".local" / "share" / "shadPS4";
+    }
 #else
     const auto user_dir = std::filesystem::current_path() / PORTABLE_DIR;
 #endif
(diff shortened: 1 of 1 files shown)


squidbus:
This is already done by https://github.com/shadps4-emu/shadPS4/pull/712

robyn-dressler:
@squidbus Ah, that approach is probably better. Probably not good practice to hard-code file paths

robyn-dressler:
Just in case, I'm going to keep this PR open until a fix is merged. From the discussion on the other PR it seems like there is reluctance to use SDL as a dependency in core, and mentions of only using that approach for macOS. IMO, this is an issue that especially needs to be fixed in linux if users want to use the emulator as a regular desktop app.

If the devs don't want to change the default behavior on a fresh install, we could also add a configuration option in the form of a TOML file installed alongside the program, or even a cmake option to switch between portable and non-portable modes. This would allow us to easily package a desktop version of the emulator without patching the code directly.

src/common/path_util.cpp
```diff
@@ -95,6 +95,13 @@ static auto UserPaths = [] {
         user_dir =
             std::filesystem::path(getenv("HOME")) / "Library" / "Application Support" / "shadPS4";
     }
+#elif defined(__linux__)
+    auto user_dir = std::filesystem::current_path() / PORTABLE_DIR;
+    // Check if the "user" directory exists in the current path:
+    if (!std::filesystem::exists(user_dir)) {
+        // If it doesn't exist, use the new hardcoded path:
+        user_dir = std::filesystem::path(getenv("HOME")) / ".local" / "share" / "shadPS4";
```
abouvier: You're supposed to use `$XDG_DATA_HOME` first before fallbacking to `$HOME/.local/share`.

robyn-dressler: Fixed

Issue #1234, opened 2024-10-04
Updator not working for AppImage build: cannot create regular file, Text file busy


KrisCris:
```log
Starting Update...
Archive:  /home/deck/Applications/Sony PlayStation 4/user/temp_download_update/temp_download_update.zip
  inflating: /home/deck/Applications/Sony PlayStation 4/user/temp_download_update/Shadps4-qt.AppImage  
cp: cannot create regular file '/home/deck/Applications/Sony PlayStation 4/Shadps4-qt.AppImage': Text file busy
```

Hermiten:
Maybe something for you @DanielSvoboda ?

DanielSvoboda:
> ```
> Starting Update...
> Archive:  /home/deck/Applications/Sony PlayStation 4/user/temp_download_update/temp_download_update.zip
>   inflating: /home/deck/Applications/Sony PlayStation 4/user/temp_download_update/Shadps4-qt.AppImage  
> cp: cannot create regular file '/home/deck/Applications/Sony PlayStation 4/Shadps4-qt.AppImage': Text file busy
> ```

What is your operating system?

DanielSvoboda:
Please download this action and try to perform the update, to test if it solves your problem.
https://github.com/shadps4-emu/shadPS4/actions/runs/11184504826

KrisCris:
> Please download this action and try to perform the update, to test if it solves your problem. https://github.com/shadps4-emu/shadPS4/actions/runs/11184504826

Can confirm this build fixed the issue!

PR #1238, opened 2024-10-04
fix update linux console

DanielSvoboda:
This should solve the problem of not being able to update if the emulator is running through the console on Linux.

as reported in: issues https://github.com/shadps4-emu/shadPS4/issues/1234 great number :)


DanielSvoboda:
@georgemoralis this is ready to merge, basically it closes shaps4 on linux before replacing the update files when running via command line, solving the problem mentioned in issues #1234

PR #1240, opened 2024-10-04
Disable ffmpeg DLLs

Xphalnos:
I made sure to remove the ffmpeg DLLs which reduces the size of the artifact from 31MB to 23.7MB.
These DLLs are not necessary for shadPS4 to work.
This also speeds up the download speed of updates.


LeDragoX:
I've tested before if removing the FFmpeg DLLs would break something, and it was only required to **Play title music** option, because without it, no sound would be played when a game is selected.
image

Xphalnos:
You're right!

PR #1242, opened 2024-10-04
Fix list sorting for some categories

ElBread3:
When [the sort all menu items PR](https://github.com/shadps4-emu/shadPS4/pull/1183) was added I noticed the sorting being entirely out of order for some of the categories, so this should fix that. I've only confirmed they sort in the right order, so I'm not sure every category sorts in the right direction.

Additionally, I decided to add game info to the tool tip because I would always have to switch to list view from grid view just to see stuff like the game version, very annoying.


bigol83:
All is good, works as expected

PR #1249, opened 2024-10-05
Fix a Qt warning

Xphalnos:
I fixed two warnings that appeared since the Qt 6.7.3 update.

image


polybiusproxy:
I assume this has been tested

Xphalnos:
> I assume this has been tested

Yes I tested it and found nothing abnormal.

PR #1250, opened 2024-10-05
vulkan: Fix two more validation errors.

squidbus:
Small fixes to two more validation errors:
* Fix SPIR-V error when sharp has size of 0, as array declaration needs to be at least size 1. Since size is 0 it should just be bound to null anyway.
* Fix window resize destroying semaphores before commands referencing them have finished execution, by waiting for device to become idle before destroying.


raphaelthegreat:
When size is zero the array size should be set to maximum ubo size. If shader is used again with a valid buffer size we make sure it will be at least that. Permutations will only kick in if the size is large enough to need ssbo

PR #1251, opened 2024-10-05
shader_recompiler: Set array size to max UBO size when 0.

squidbus:
Per [comment on previous PR](https://github.com/shadps4-emu/shadPS4/pull/1250#issuecomment-2395153326), change the array size 0 behavior to use maximum UBO size.


squidbus:
Added in extra change requested by @psucien

PR #1252, opened 2024-10-05
DebugPrintf in shaders

baggins183:
Add a DebugPrint instruction to the IR that maps to SPIRV Nonsemantic Debugprintf
https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/docs/debug_printf.md
to view in renderdoc
https://renderdoc.org/docs/window/shader_messages.html


src/shader_recompiler/ir/ir_emitter.cpp
```diff
@@ -1553,4 +1558,40 @@ void IREmitter::ImageWrite(const Value& handle, const Value& coords, const Value
...
+
+    for (int i = format_args.size(); i < PRINT_MAX_ARGS; i++) {
+        args[i] = Inst(Opcode::Void);
+    }
+
+    Value val = Inst(Opcode::DebugPrint, args[0], args[1], args[2], args[3], args[4]);
+    DebugPrintFlags flags;
+
+    flags.string_idx.Assign(info.string_pool.size());
+    info.string_pool.emplace_back(format);
```
raphaelthegreat: Thinking about it more, if the format string is always constant you don't need a pool and can link to the const char* directly.

baggins183: i changed it to this way so i can fit a 5th argument by stashing a u16 index instead of using arg0 for a pointer

baggins183: i guess it would be simpler though to assume a constant string literal. I wouldn't need to pass Info around as much

raphaelthegreat: You could also add a const char* type into the IR::Value tagged union and sacrifice arg0 to pass the format string

baggins183: ok, i prefer that too, even if 4 arg limitation. ill change it

src/shader_recompiler/ir/value.cpp
```diff
@@ -69,6 +72,8 @@ bool Value::operator==(const Value& other) const {
     case Type::U64:
     case Type::F64:
         return imm_u64 == other.imm_u64;
+    case Type::StringLiteral:
+        return std::string(string_literal) == other.string_literal;
```
raphaelthegreat:
```suggestion
        return std::string_view(string_literal) == other.string_literal;
```

raphaelthegreat:
The debug printf instruction must also be marked as having side effects otherwise it will be removed as dead code by pass

baggins183:
> The debug printf instruction must also be marked as having side effects otherwise it will be removed as dead code by pass

that part is there in microinstruction.cpp

baggins183:
rebased

raphaelthegreat (approved):
Thanks

PR #1255, opened 2024-10-05
core/libraries: Initial fiber implementation

polybiusproxy:
Implements user-mode fibers used by games such as Persona 5 (Royal).

Missing:
- [ ] Cross-platform implementation
  - Look into Boost fibers

image


georgemoralis:
rayman also uses fibers maybe it should be tested it too

polybiusproxy:
I don't have Rayman so I'd appreciate testing.

PR #1257, opened 2024-10-06
Working touchpad support

fireph:
Tested on a PS5 controller plugged in via USB on Windows 11. It appears to be working as one would expect, but there can be some weirdness when you press the touchpad really fast (it doesn't register the correct side of the touchpad in that case sometimes). I filed an issue with SDL since it is a bug within that project that causes this behavior https://github.com/libsdl-org/SDL/issues/11085.

I attempted to also get the back button on other controllers to trigger the left side of the touchpad, but haven't gotten a working version of that yet (and there would likely need to be a setting in shadPS4 to set whether the user wanted it to be left/right/center/none of the touchpad for the back button) so I'll leave that for a future PR (https://github.com/shadps4-emu/shadPS4/pull/1258).

### DIFF
--- a/src/core/libraries/pad/pad.cpp
+++ b/src/core/libraries/pad/pad.cpp
@@ -368,7 +368,8 @@ int PS4_SYSV_ABI scePadReadState(s32 handle, OrbisPadData* pData) {
     pData->angularVelocity.x = 0.0f;
     pData->angularVelocity.y = 0.0f;
     pData->angularVelocity.z = 0.0f;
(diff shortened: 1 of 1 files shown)


Kyoskii:
tested on ps4 controller as well, also works.

gandalfthewhite19890404:
I suggest add separate window (by ImGui for example) as touchpad, so you can click and move there by mouse.

fireph:
> I suggest add separate window (by ImGui for example) as touchpad, so you can click and move there by mouse.

that's a good idea, but I would want to attempt that in a future PR since it will require these changes anyways.

rafael-57:
image
IT WORKS, I LOVE YOU

PR #1258, opened 2024-10-06
Add support to click touchpad using back button on non PS4/5 controllers

fireph:
Takes the back button and allows the user to change the behavior of how it clicks the touchpad. The current options are left, right, center, and none. It currently defaults to left, but I am open to changing this if "None" seems to be a better default. I could also change this to just 2 floats from 0->1 for x and y if we want something more generic in the settings, but I am open to any feedback on this.

image

Depends on https://github.com/shadps4-emu/shadPS4/pull/1257


fireph:
The only question I have now is if I need to add all the translations for each language, or just leave them blank until translations exist (If I do need to add them, I'll update the PR with them)

PR #1261, opened 2024-10-06
Fix settings descriptions

fireph:
They were broken by https://github.com/shadps4-emu/shadPS4/pull/1249, but this is the correct way to fix the qt warning


Xphalnos:
I already made a commit here https://github.com/shadps4-emu/shadPS4/pull/1259

fireph:
oh whoops, disregard then

PR #1267, opened 2024-10-06
ime_dialog: Initial implementation

LNDF:
This PR adds a basic implementation of ime_dialog. Text can be entered using a ImGui windw opened on the screen when the game invokes que keyboard.

Supports both single line and multiline, numeric input and text filter callbacks.

I don't have access neither a macOS or Windows system so I was not able to thest it there.


squidbus:
I think it would be simpler to use [Boost.Locale](https://www.boost.org/doc/libs/1_49_0/libs/locale/doc/html/charset_handling.html) instead of adding an additional dependency on iconv.

UTF-16 -> UTF-8 can be done with `boost::locale::conv::utf_to_utf<char>(start, end)` and the other direction with `boost::locale::conv::utf_to_utf<char16_t>(start, end)`

LNDF:
> I think it would be simpler to use [Boost.Locale](https://www.boost.org/doc/libs/1_49_0/libs/locale/doc/html/charset_handling.html) instead of adding an additional dependency on iconv.
> 
> UTF-16 -> UTF-8 can be done with `boost::locale::conv::utf_to_utf<char>(start, end)` and the other direction with `boost::locale::conv::utf_to_utf<char16_t>(start, end)`

I will try this later to avoid using platform specific code. 

I used icomv since it is installed by default on Linux and macos, bundled with glibc in Linux and as a standalone library on macos.

LNDF:
Sorry, I am not able to do that until the necessary boost components are added to https://github.com/shadps4-emu/ext-boost/

squidbus:
> Sorry, I am not able to do that until the necessary boost components are added to https://github.com/shadps4-emu/ext-boost/

You should probably open a PR to add them, then.

LNDF:
> > Sorry, I am not able to do that until the necessary boost components are added to https://github.com/shadps4-emu/ext-boost/
> 
> You should probably open a PR to add them, then.

Done. One drawback I notice with this approach is that a std::string or std::u16string is created each time a conversion occurs, resulting in dynamic memory allocation on the heap.

ElBread3:
Tested with Patapon and froze on tell name still, also could you please implement sceImeUpdate it gets used a lot

LNDF:
> Tested with Patapon and froze on tell name still, also could you please implement sceImeUpdate it gets used a lot

Does it freeze in main?

Edit: also, what os?

ElBread3:
It crashes on main by calling sceImeDialogInit sceImeDialogGetResult which main doesn't have resulting in engine crash. With PR it just freezes entirely without popping up anything. I'm on Windows.

Also FYI the only stub resolves for sceIme in Patapon looks to be sceImeKeyboardOpen (PMVehSlfZ94) sceImeUpdate (-4GCfYdNF1s) sceImeKeyboardClose (eaFXjfJv3xs) so would be great to implement those.

LNDF:
> It crashes on main by calling sceImeDialogInit sceImeDialogGetResult which main doesn't have resulting in engine crash. With PR it just freezes entirely without popping up anything. I'm on Windows.
> 
> Also FYI the only stub resolves for sceIme in Patapon looks to be sceImeKeyboardOpen (PMVehSlfZ94) sceImeUpdate (-4GCfYdNF1s) sceImeKeyboardClose (eaFXjfJv3xs) so would be great to implement those.

Tanks. I will test it later

ElBread3:
Would have tested Stardew Valley but mono games crash rn. Of course Souls works. Just not Patapon so far.
bbime
dsrime

ElBread3:
@LNDF Using debug markers, in Patapon when init creates the UI object the first debug marker hit is the destruction function, explains why I don't see it but how tf can it call destroy before create?
patapon-imedestruct

LNDF:
> @LNDF Using debug markers, in Patapon when init creates the UI object the first debug marker hit is the destruction function, explains why I don't see it but how tf can it call destroy before create? patapon-imedestruct

The UI object is created in the stack on sceImeDialogInit. Then, the object is moved into the global variable and the one on the stack is destroyed after the function ends. 

I have been testing patapon earlier andi got the keyboard UI rendered in one frame.

I have been reverse engineering the game and it literally loops until the result of sceImeDialogGetStatus is FINISHED.

```c
while (sceImeDialogGetStatus() != FINISHED) {
    sceKernelUsleep();
}
```

And imGui cannot receive input to finish the dialog so we are in a deadlock

There must be something else that we are missing.

LNDF:
@ElBread3 I was unable to determine what causes there game to lock on shadps4 and not the PS4.

I cannot say whether the lock is caused by an incorrect implementation of the ime dialog or it is coming from another place on shadps4.

I will take another look tomorrow.

ElBread3:
Do you happen to be on the Discord server?

LNDF:
> Do you happen to be on the Discord server?

Yes

LNDF:
This still doesn't fix patapon

CMakeLists.txt
```diff
@@ -768,6 +770,10 @@ if (APPLE)
 
   # Half float conversions for F16C patches
   target_link_libraries(shadps4 PRIVATE half)
+
+  # Link against the iconv in macOS
+  find_package(Iconv REQUIRED)
+  target_link_libraries(shadps4 PRIVATE Iconv::Iconv)
```
viniciuslrangel: Dont need to use this to convert between utf8/wchar

src/common/cstring.h
```diff
@@ -152,6 +152,24 @@ class CString {
...
+
+/**
+ * @brief A null-terminated 16-bit char string with a fixed maximum length
+ *        This class is not meant to be used as a general-purpose string class
+ *        It is meant to be used as `char[N]` where memory layout is fixed
+ * @tparam N Maximum length of the string
+ */
+template <size_t N>
+using CU16String = CString<N, char16_t>;
+
```
viniciuslrangel: Dont need to repeat this comment here

src/core/libraries/dialogs/ime_dialog.cpp
```diff
@@ -1,28 +1,80 @@
...
+#include <algorithm>
+#include <magic_enum.hpp>
 #include "common/logging/log.h"
 #include "core/libraries/error_codes.h"
 #include "core/libraries/libs.h"
 #include "ime_dialog.h"
+#include "ime_dialog_ui.h"
+
+static constexpr float MAX_X_POSITIONS[] = {3840.0f, 1920.0f};
+static constexpr float MAX_Y_POSITIONS[] = {2160.0f, 1080.0f};
```
viniciuslrangel: use std::array to avoid memory corruption

src/core/libraries/dialogs/ime_dialog.cpp
```diff
@@ -1,28 +1,80 @@
...
+        return Error::DIALOG_NOT_IN_USE;
+    }
+
+    if (g_ime_dlg_status != OrbisImeDialogStatus::RUNNING) {
+        LOG_INFO(Lib_ImeDialog, "IME dialog not running");
+        return Error::DIALOG_NOT_RUNNING;
+    }
+
+    g_ime_dlg_status = OrbisImeDialogStatus::FINISHED;
+    g_ime_dlg_result.endstatus = OrbisImeDialogEndStatus::ABORTED;
```
viniciuslrangel, line 53: Reset/Reconstruct UI here to remove it from layer list

src/core/libraries/dialogs/ime_dialog.cpp
```diff
@@ -45,26 +97,138 @@ int PS4_SYSV_ABI sceImeDialogGetPanelSizeExtended() {
...
+
+    if (result == nullptr) {
+        LOG_INFO(Lib_ImeDialog, "called with result (NULL)");
+        return Error::INVALID_ADDRESS;
+    }
+
+    if (!IsMemZero(result->reserved, 12)) {
+        LOG_INFO(Lib_ImeDialog, "Invalid result->reserved");
+        return Error::INVALID_RESERVED;
+    }
```
viniciuslrangel: Not really necessary. Better to not have this and avoid version compatibility

LNDF:
@ElBread3 Ok so I found what is happening with patapon. Shadps4 freezes the emulator window whenever a game goes to 0 fps,. When the on screen keyboard is present, the game goes to 0 fps (even on PS4) and the keyboard ui is shown. Because there are no new frames, the emulator freezes and cannot be interacted with the keyboard. In constrast, on ps4, the keyboard can be interacted with even if the game is 0 fps.

That i think is a separate issue and must be fixed in another PR, so I think this PR can be merged.

src/core/libraries/dialogs/ime_dialog.cpp
```diff
@@ -45,26 +97,138 @@ int PS4_SYSV_ABI sceImeDialogGetPanelSizeExtended() {
...
+
+    if (param->userId != 1) { // We only support user 1 for now
+        LOG_INFO(Lib_ImeDialog, "Invalid param->userId");
+        return Error::INVALID_USER_ID;
+    }
+
+    if (!IsMemZero(param->reserved, 16)) {
+        LOG_INFO(Lib_ImeDialog, "Invalid param->reserved");
+        return Error::INVALID_RESERVED;
+    }
```
viniciuslrangel: Isn't required

src/core/libraries/dialogs/ime_dialog.h
```diff
@@ -11,71 +12,116 @@ class SymbolsResolver;
...
-    ORBIS_IME_ENTER_LABEL_DEFAULT = 0,
-    ORBIS_IME_ENTER_LABEL_SEND = 1,
-    ORBIS_IME_ENTER_LABEL_SEARCH = 2,
-    ORBIS_IME_ENTER_LABEL_GO = 3
-};
-enum OrbiImeInputMethod { ORBIS_IME_INPUT_METHOD_DEFAULT = 0 };
-
-typedef int (*OrbisImeTextFilter)(wchar_t* outText, u32* outTextLength, const wchar_t* srcText,
-                                  u32 srcTextLength);
+enum class OrbisImeEnterLabel : u32 { DEFAULT = 0, SEND = 1, SEARCH = 2, GO = 3 };
```
viniciuslrangel: Put a trailing comma to let clang-format split each value in lines

src/core/libraries/dialogs/ime_dialog.h
```diff
@@ -11,71 +12,116 @@ class SymbolsResolver;
...
-    s8 reserved[16];
+DECLARE_ENUM_FLAG_OPERATORS(OrbisImeDialogOption)
+
+enum class OrbisImeInputMethod : u32 { DEFAULT = 0 };
+
+enum class OrbisImeHorizontalAlignment : u32 { LEFT = 0, CENTER = 1, RIGHT = 2 };
+
+enum class OrbisImeVerticalAlignment : u32 { TOP = 0, CENTER = 1, BOTTOM = 2 };
+
+enum class OrbisImePanelPriority : u32 { DEFAULT = 0, ALPHABET = 1, SYMBOL = 2, ACCENT = 3 };
```
viniciuslrangel: ditto

src/core/libraries/dialogs/ime_dialog_ui.cpp
```diff
@@ -0,0 +1,488 @@
...
+        return false;
+    }
+
+    int converted_size =
+        MultiByteToWideChar(CP_UTF8, 0, utf8_text, utf8_text_len,
+                            reinterpret_cast<wchar_t*>(orbis_text), orbis_text_len);
+
+    return converted_size != 0;
+#endif
+}
```
viniciuslrangel: Dear ImGui provides helpers for that: ImTextStrFromUtf8 & ImTextStrFromUtf8

src/core/libraries/dialogs/ime_dialog_ui.cpp
```diff
@@ -0,0 +1,488 @@
...
+    orbis_to_utf8 = iconv_open(IME_UTF8_ENCODING, IME_ORBIS_ENCODING);
+    utf8_to_orbis = iconv_open(IME_ORBIS_ENCODING, IME_UTF8_ENCODING);
+
+    ASSERT_MSG(orbis_to_utf8 != (iconv_t)-1, "Failed to open iconv orbis_to_utf8");
+    ASSERT_MSG(utf8_to_orbis != (iconv_t)-1, "Failed to open iconv utf8_to_orbis");
+#endif
+
+    if (param->title) {
+        std::size_t title_len = std::char_traits<char16_t>::length(param->title);
+        title = new char[title_len * 4 + 1];
```
viniciuslrangel: Avoid raw allocation, use a std::vector

src/core/libraries/dialogs/ime_dialog_ui.cpp
```diff
@@ -0,0 +1,488 @@
...
+        title[title_len * 4] = '\0';
+
+        if (!ConvertOrbisToUTF8(param->title, title_len, title, title_len * 4)) {
+            LOG_ERROR(Lib_ImeDialog, "Failed to convert title to utf8 encoding");
+        }
+    }
+
+    if (param->placeholder) {
+        std::size_t placeholder_len = std::char_traits<char16_t>::length(param->placeholder);
+        placeholder = new char[placeholder_len * 4 + 1];
```
viniciuslrangel: Ditto

src/core/libraries/dialogs/ime_dialog_ui.h
```diff
@@ -0,0 +1,91 @@
...
+    OrbisImeEnterLabel enter_label{};
+    OrbisImeTextFilter text_filter{};
+    OrbisImeExtKeyboardFilter keyboard_filter{};
+    u32 max_text_length{};
+    char16_t* text_buffer{};
+    char* title = nullptr;
+    char* placeholder = nullptr;
+
+    // A character can hold up to 4 bytes in UTF-8
+    char current_text[ORBIS_IME_DIALOG_MAX_TEXT_LENGTH * 4 + 1] = {0};
```
viniciuslrangel: Could be using Common::CString to avoid memcpy

src/core/libraries/dialogs/ime_dialog_ui.cpp
```diff
@@ -0,0 +1,488 @@
...
+    if (first_render) {
+        SetKeyboardFocusHere();
+    }
+    if (InputTextEx("##ImeDialogInput", state->placeholder, state->current_text,
+                    state->max_text_length, input_size, flags, InputTextCallback, this)) {
+        state->input_changed = true;
+    }
+}
+
+int ImeDialogUi::InputTextCallback(ImGuiInputTextCallbackData* data) {
```
viniciuslrangel, line 345: Would be nice to handle Enter as Ok button

viniciuslrangel (changes requested):

src/core/libraries/dialogs/ime_dialog_ui.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+    if (!ConvertUTF8ToOrbis(current_text, src_text_length, src_text,
+                            ORBIS_IME_DIALOG_MAX_TEXT_LENGTH)) {
+        LOG_ERROR(Lib_ImeDialog, "Failed to convert text to orbis encoding");
+        return false;
+    }
+
+    auto* linker = Common::Singleton<Core::Linker>::Instance();
+    int ret =
+        linker->ExecuteGuest(text_filter, out_text, &out_text_length, src_text, src_text_length);
```
viniciuslrangel, line 171:
```suggestion
    int ret = text_filter(out_text, &out_text_length, src_text, src_text_length);
```

viniciuslrangel (changes requested):

src/core/libraries/dialogs/ime_dialog_ui.cpp
squidbus, line 128: Why? All guest code is supposed to be called through `ExecuteGuest` now.

src/core/libraries/dialogs/ime_dialog.cpp
LNDF: An application should call sceImeDialogTerm after sceImeDialogAbort and resources are freed there. Also the application is not able to call sceImeDialogInit after sceImeDialogAbort because status is not NONE (it's FINISHED) so the UI will never be drawn twoice.

viniciuslrangel: Fair

src/core/libraries/dialogs/ime_dialog_ui.cpp
LNDF: The ImGUI function convert from UTF8 to UTF32 but PS4 uses UTF16. 
Wchar is different size on windows and linux/macos. We need a function to convert specificly from UTF8 to UTF16 and the other way arround.

viniciuslrangel: Remove the `IMGUI_USE_WCHAR32` from imgui_config.h

LNDF: I searched the imgui source code and, sadly, it only provides implementation for utf32. :(

I even found a coment in imgui that they stopped using ImTextStrFromUtf8 because they need to support both utf16 and 32.

https://github.com/shadps4-emu/ext-imgui/blob/636cd4a7d623a2bc9bf59bb3acbb4ca075befba3/imgui.cpp#L2268C1-L2272C81

viniciuslrangel: Yeah, but that definition in imgui_config.h controls whether ImWchar is 16 or 32 bits.
To ensure nobody changes, you can even do something like:
```Cpp
#ifdef IMGUI_USE_WCHAR32
#error "This project uses 16 bits wchar standard like Orbis"
#endif
```

LNDF:
I just tested DS: R after all fixes and it is working as spected.

Sorry @squidbus, I don't understand your review. What ExecuteGuest used incorrectly?

squidbus:
> I don't understand your review. What ExecuteGuest used incorrectly?

It wasn't a review, it was a response to a comment in another review. GitHub just marked it weird. You can ignore it.

PR #1276, opened 2024-10-07
Update sq translation

F1219R:


georgemoralis:
there are conflicts in the pr correct them so we can add it

PR #1290, opened 2024-10-08
shader_recompiler: Move sampling parameter resolution to tracking pass and support more derivative types.

squidbus:
The current code for the `IMAGE_SAMPLE` instruction does not work with derivatives for non-2D images as the type is not known when the IR is emitted, and thus it cannot determine how many derivative parameters there are.

Instead, we can emit a placeholder IR instruction with all of the addressing data, and then once the resource is known we can resolve all of the parameters using the type and replace it with the correct instruction.

With the emission of the exact sampling IR instruction moved to after the resource has been determined, non-2D image derivatives are now supported, and the parameter handling is cleaner since it no longer uses individual placeholder arguments that need to be resolved.

As part of this change I ended up needing to increase the max IR argument count to 6, in order to pass the full set of derivatives for the ImageGradient instruction. Technically you could probably pack one of the vectors in with the `lod_clamp`, but it would make things a lot messier having to place the `lod_clamp` at the right index depending on the derivative size, as well as unpack and pass the right data later on in SPIR-V. It is much cleaner if we can just have one more parameter slot to pass it in separately.

I also changed the sample dref instructions to have a F32x4 return type, with the right component extracted in the IR for the result. Otherwise the typing would be inconsistent between the placeholder and replacement instructions, and in the initial IR the type would be wrong compared to what the code expects. Some of the other argument types have also been fixed up in the sampling instructions to match what they must be.

Tested in a few games. Fixes missing cube derivative support in some shaders in CUSA16429.

### DIFF
--- a/src/shader_recompiler/ir/opcodes.h
+++ b/src/shader_recompiler/ir/opcodes.h
@@ -21,7 +21,7 @@ namespace Detail {
 struct OpcodeMeta {
     std::string_view name;
     Type type;
-    std::array<Type, 5> arg_types;
+    std::array<Type, 6> arg_types;
 };
 
 // using enum Type;
--- a/src/shader_recompiler/backend/spirv/emit_spirv_image.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_image.cpp
@@ -59,18 +59,21 @@ struct ImageOperands {
         }
     }
 
-    void AddDerivatives(EmitContext& ctx, Id derivatives) {
-        if (!Sirit::ValidId(derivatives)) {
+    void AddDerivatives(EmitContext& ctx, Id derivatives_dx, Id derivatives_dy) {
+        if (!Sirit::ValidId(derivatives_dx) || !Sirit::ValidId(derivatives_dy)) {
             return;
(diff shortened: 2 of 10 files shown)


src/shader_recompiler/ir/opcodes.h
```diff
@@ -21,7 +21,7 @@ namespace Detail {
 struct OpcodeMeta {
     std::string_view name;
     Type type;
-    std::array<Type, 5> arg_types;
+    std::array<Type, 6> arg_types;
```
raphaelthegreat, line 24: You also need to update args array in Inst class, see value.h

squidbus: Done

raphaelthegreat:
This looks okay to me. Would be good to get some testing first though as its pretty major change

squidbus:
> This looks okay to me. Would be good to get some testing first though as its pretty major change

I've been running it with all of my games and haven't seen any issues yet. Not sure if anyone else has tested.

raphaelthegreat:
Then let's do some testing in production

PR #1302, opened 2024-10-08
spirv: Flush denormals if possible and runtime info refactor

raphaelthegreat:
Most shaders provided by the guest are configured to flush 32-bit denormals by default. Implement this behavior in shaders by reading the float_mode member of program settings. This will only be applied if said capability is exposed by the driver (proprietary NVIDIA doesn't support denorm flushing for example).

Additionally done a small refactor to RuntimeInfo to allow members to be stored in a union to save some storage space. This required ensuring all member structs have trivial destructors. While at it, moved GS copy shader parsing to shader generation as it doesn't need to be done on every draw that uses GS (avoids constructing an std::unordered_map on every such draw)

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv.cpp
@@ -206,10 +206,7 @@ Id DefineMain(EmitContext& ctx, const IR::Program& program) {
     return main;
 }
 
-void DefineEntryPoint(const IR::Program& program, EmitContext& ctx, Id main) {
-    const auto& info = program.info;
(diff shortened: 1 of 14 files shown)


rainmakerv3:
image

Does not solve the problem with vertex explosions on Bloodborne completely, but hey might have been reduced. Am on NVidia GPU though, so not sure the denormal flushing is applied

raphaelthegreat:
This PR does nothing on NVIDIA so this is expected

Foul-Tarnished:
Will you add this feature for nvidia, or is it not needed ?

PR #1305, opened 2024-10-09
Play Time

DanielSvoboda:
**Records the playing time** in hours, minutes, seconds after closing and every 1 minute.

Saves in the file user/play_time.txt with serial and time 00:00:00.
In the menu it appears without seconds so as not to be cluttered.

image

----

Also fixes the checkbox that someone hid haha
image

And set the first tab (General) as default when opening the options, not the second one (Input)


GHU7924:
I'm sorry that I'm probably writing in the wrong place, but since you mentioned a problem in the interface, I also want to report a problem that occurs when you hover the cursor, which leads to a displacement of the interface
Interface

PR #1307, opened 2024-10-09
update boost submodule. Add boost includes in subproject instead of e…

baggins183:
…xternals/CMakeLists.txt


baggins183:
Main point is to move the include path setting to the ext-boost submodule so we dont need to change include paths there and here in separate commits

Im not sure if these are necessary
>     set(BOOST_ROOT "${CMAKE_SOURCE_DIR}/externals/ext-boost" CACHE STRING "")
>     set(Boost_NO_SYSTEM_PATHS ON CACHE BOOL "")
1st one causes a warning

PR #1324, opened 2024-10-10
Network libs fixes , stubs and more

georgemoralis:


georgemoralis:
@squidbus can you help a bit with macOS failures?

squidbus:
> can you help a bit with macOS failures?

`gethostname` is in `unistd.h`. This should be the case on Linux as well, I assume its being included transitively currently.

squidbus:
It looks like you will need both `unistd.h` and `netdb.h` to have both `gethostname` and `gethostbyname`.

georgemoralis:
> It looks like you will need both `unistd.h` and `netdb.h` to have both `gethostname` and `gethostbyname`.

doesnt' seem to work ;/

squidbus:
> doesnt' seem to work ;/

You removed `netdb.h` in your last commit

georgemoralis:
> > doesnt' seem to work ;/
> 
> You removed `netdb.h` in your last commit

thanks!

raphaelthegreat:
I don't see the point of the share play and remote play stubs. Unless you want to suppress their logging which isn't done here, we probably wont be implementing them soon

georgemoralis:
> I don't see the point of the share play and remote play stubs. Unless you want to suppress their logging which isn't done here, we probably wont be implementing them soon

all the functions i implemented are returning not available features , so it will probably prevent a game from asking them (at least that is what it does in re6)

Issue #1325, opened 2024-10-10
Game path duplicates upon each re-launch


KrisCris:
![image](https://github.com/user-attachments/assets/11888f94-aa0f-46dd-9349-20b71674b53d)

KrisCris:
And remove one remove them all

PR #1329, opened 2024-10-10
Slightly refactor the game install dirs code

KrisCris:
Well... I think I made this PR too late (originally as a fix of https://github.com/shadps4-emu/shadPS4/issues/1325), so some of the issues have already been fixed, sorry. But here it is.

The main reason #1308 duplicates installDir is that the config is not removed from the file. (https://github.com/shadps4-emu/shadPS4/commit/01e39a4e70f4a56d74f12a59dadf80e3f9a83528)

Meanwhile, I replaced setGameInstallDirs with addGameInstallDir and removeGameInstallDir so no more duplicated path-adding/removing code, and avoid passing the vectors around; The return value of getGameInstallDirs is made const& so it won't be copied and modified everywhere.


ElBread3:
Looks like a great refactor, should probably rename PR to reflect that. Only question is why use std ofstream instead of std filesystem like the rest of the code?

KrisCris:
> Only question is why use std ofstream instead of std filesystem like the rest of the code?

I just copied that code from the end of `Config::save` ;). But isn't std::ofstream << data the way to write? I am unsure about the alternative approach.

---

Btw, in `Config::save` do we really need to read from the file before saving, given we are reassigning every item? Perhaps we could just remove both pieces of code, so the file always reflects the updated config layout?

https://github.com/shadps4-emu/shadPS4/blob/01e39a4e70f4a56d74f12a59dadf80e3f9a83528/src/common/config.cpp#L542-L544

https://github.com/shadps4-emu/shadPS4/blob/01e39a4e70f4a56d74f12a59dadf80e3f9a83528/src/common/config.cpp#L574-L590

src/common/config.cpp
```diff
@@ -527,19 +537,17 @@ void load(const std::filesystem::path& path) {
 
         auto old_game_install_dir = toml::find_fs_path_or(gui, "installDir", {});
         if (!old_game_install_dir.empty()) {
-            settings_install_dirs.push_back(old_game_install_dir);
-            data.as_table().erase("installDir");
+            addGameInstallDir(old_game_install_dir);
+            gui.as_table().erase("installDir");
+            std::ofstream file(path, std::ios::out);
+            file << data;
+            file.close();
```
Exhigh: This needs to be moved to the save function and done such that it's for migration ( after a major release this should be removed). Having it in the load function isn't correct.
```suggestion
            gui.as_table().erase("installDir");
            std::ofstream file(path, std::ios::out);
            file << data;
            file.close();
```

src/common/config.cpp
```diff
@@ -325,9 +325,19 @@ void setMainWindowGeometry(u32 x, u32 y, u32 w, u32 h) {
...
-    settings_install_dirs.resize(dir.size());
-    settings_install_dirs = dir;
+bool addGameInstallDir(const std::filesystem::path& dir) {
+    if (std::find(settings_install_dirs.begin(), settings_install_dirs.end(), dir) ==
+        settings_install_dirs.end()) {
+        settings_install_dirs.push_back(dir);
+        return true;
+    }
+    return false;
+}
```
Exhigh, line 335: This should be changed such that its only called from addFolderButton since its the only place from where we add folders. Rather a check should be added to check if "installDir" has a path, if so then adding just installDir's path to installDirs then deleting installDir, such that we migrate away from calling installDir. Also, we will avoid spending time checking if a game folder exists within the game folders we saved before the current session.

Exhigh: And add a deprecation notice for the migrating code. (possibly after 1 major release)

Exhigh (changes requested):

src/common/config.cpp
Exhigh: Also this causes the rest of the settings to be not loaded if installDir exists.

src/common/config.cpp
ElBread3: There's already a migration function in config load, if you're talking about there not being one yet.

src/qt_gui/settings_dialog.cpp
Exhigh, line 1: Also line `ui->removeFolderButton->setEnabled(false);` (221) needs to be moved to LoadValuesFromConfig() with the bool checking whether the gamefolders are empty.

`ui->removeFolderButton->setEnabled(!Config::getGameInstallDirs().empty());`
should be as such.

src/common/config.cpp
Exhigh: Like i said we just need to add a comment saying that the code needs to be deprecated and i didn't say that there wasn't one. So that we remember to remove the code after a major release or so.

src/common/config.cpp
KrisCris: fixed in https://github.com/shadps4-emu/shadPS4/pull/1329/commits/13c1c34c82cb1b57d0316bc21c5e59a5fa1b59b4

src/qt_gui/settings_dialog.cpp
KrisCris: fixed via https://github.com/shadps4-emu/shadPS4/pull/1329/commits/fa53fb42afb6d25829a2156dd230e26eb1d69917. 

Instead of checking whether gamefolders are empty, I used the similar logic from the folder widget clicked code:
https://github.com/shadps4-emu/shadPS4/blob/fa53fb42afb6d25829a2156dd230e26eb1d69917/src/qt_gui/settings_dialog.cpp#L232-L234

src/common/config.cpp
KrisCris: > Like i said we just need to add a comment saying that the code needs to be deprecated and i didn't say that there wasn't one. So that we remember to remove the code after a major release or so.

Is this what you suggest? https://github.com/shadps4-emu/shadPS4/pull/1329/commits/4e294ddbf52348b6fb96a665dd5041806854a722

src/common/config.cpp
```diff
@@ -525,21 +535,16 @@ void load(const std::filesystem::path& path) {
...
-            if (not_already_included) {
-                settings_install_dirs.emplace_back(std::filesystem::path{dir});
-            }
+            settings_install_dirs.emplace_back(std::filesystem::path{dir});
+        }
+
+        // TODO Migration code, after a major release this should be removed.
+        auto old_game_install_dir = toml::find_fs_path_or(gui, "installDir", {});
+        if (!old_game_install_dir.empty()) {
+            addGameInstallDir(old_game_install_dir);
```
Exhigh: I would suggest moving this above and use an else condition to switch to the installDirs. Also, we don't need use `addGameInstallDir(old_game_install_dir);` rather just push the string to settings_install_dirs.
```suggestion
        // TODO Migration code, after a major release this should be removed.
        auto old_game_install_dir = toml::find_fs_path_or(gui, "installDir", {});
        if (!old_game_install_dir.empty()) {
            addGameInstallDir(old_game_install_dir);
```

src/common/config.cpp
```diff
@@ -639,6 +644,9 @@ void save(const std::filesystem::path& path) {
 
     data["Settings"]["consoleLanguage"] = m_language;
 
+    // TODO Migration code, after a major release this should be removed.
+    data.at("GUI").as_table().erase("installDir");
+
```
Exhigh, line 649: Ideally we should check if the string exists before we remove it.
Something like
```
// TODO Migration code, after a major release this should be removed.
auto old_game_install_dir =
        toml::find_or<std::string>(data.at("GUI"), "installDir", "doesn't exist");
    if (old_game_install_dir != "doesn't exist") {
        data.at("GUI").as_table().erase("installDir");
    }
```

KrisCris: I thought the table would not throw an error when attempting to erase a key that doesn't exist?

src/common/config.cpp
KrisCris: https://github.com/shadps4-emu/shadPS4/pull/1329/commits/2144b00de539490957d9e7437e5e2136bc5c267b

src/common/config.cpp
Exhigh: It doesn't but i guess its alright, since its going to be removed after some time. Ignore this then

src/common/config.cpp
```diff
@@ -535,16 +535,16 @@ void load(const std::filesystem::path& path) {
...
-            toml::find_or<std::vector<std::string>>(gui, "installDirs", {});
-        for (const auto& dir : install_dir_array) {
-            settings_install_dirs.emplace_back(std::filesystem::path{dir});
-        }
-
         // TODO Migration code, after a major release this should be removed.
         auto old_game_install_dir = toml::find_fs_path_or(gui, "installDir", {});
         if (!old_game_install_dir.empty()) {
-            addGameInstallDir(old_game_install_dir);
+            settings_install_dirs.emplace_back(std::filesystem::path{old_game_install_dir});
```
Exhigh, line 541: formatting should be fixed.

KrisCris: hmmm I have already formatted the file tho?

Exhigh: Hmm, GitHub is acting weird, it initially had no spacing. Ignore this.

bigol83:
I don't know if it's this that is causing it, but game list duplication bug is back

KrisCris:
> I don't know if it's this that is causing it, but game list duplication bug is back

I can't reproduce it on myside. Can you post your config.toml file, which should under the user folder? Might due to change that we are no longer checking duplicates on previously added dirs?

Exhigh:
> I don't know if it's this that is causing it, but game list duplication bug is back

It can possibly happen only on first boot with a config with the same path in 2 different locations, after which if you remove a duplicate entry in the path tabs in settings, it should be fine.

bigol83:
Right now it's not happening anymore, next time it happens (if it happens) will make sure to report how to reproduce it.

Issue #1226, opened 2024-10-03
Bloodborne (CUSA03173) physics engine sometimes does not render properly


par274:
The title may be a bit confusing, but Bloodborne (CUSA03173) physics engine sometimes gets crazy.

When this happens, everything in the game world usually starts to "shake", resulting in a 10-20 FPS drop in overall frame rate. And going to the menu and continuing again does not work, you need to close the emulator.

The problem usually occurs randomly and usually occurs on clothes, hair and plants.

Here is video:

https://github.com/user-attachments/assets/e888ea3f-a745-4c36-a3e6-625799a674db

IAmBatby:
Just a theory not an observation but is it possible this physics problem is the result of such a harse framedrop and not the cause of it?

par274:
> Just a theory not an observation but is it possible this physics problem is the result of such a harse framedrop and not the cause of it?

No. When this problem is not present, a constant 60 FPS is achieved, but if this problem occurs in a place where there are a lot of NPCs, for example, it drops to 20-30 FPS and is fixed by closing and reopening the emulator.

HeinrichVonStinkenfinger:
> The title may be a bit confusing, but Bloodborne (CUSA03173) physics engine sometimes gets crazy.
> 
> When this happens, everything in the game world usually starts to "shake", resulting in a 10-20 FPS drop in overall frame rate. And going to the menu and continuing again does not work, you need to close the emulator.
> 
> The problem usually occurs randomly and usually occurs on clothes, hair and plants.
> 
> Here is video:
> 
>  2024-10-04.01-52-50_1.mp4

Do you have the 60 FPS patch on?

par274:
> > The title may be a bit confusing, but Bloodborne (CUSA03173) physics engine sometimes gets crazy.
> 
> > 
> 
> > When this happens, everything in the game world usually starts to "shake", resulting in a 10-20 FPS drop in overall frame rate. And going to the menu and continuing again does not work, you need to close the emulator.
> 
> > 
> 
> > The problem usually occurs randomly and usually occurs on clothes, hair and plants.
> 
> > 
> 
> > Here is video:
> 
> > 
> 
> >  2024-10-04.01-52-50_1.mp4
> 
> 
> 
> Do you have the 60 FPS patch on?

Yes.

HeinrichVonStinkenfinger:
> > > The title may be a bit confusing, but Bloodborne (CUSA03173) physics engine sometimes gets crazy.
[...]

PR #1344, opened 2024-10-11
Use GetSystemTimePreciseAsFileTime to fix fps timing issues

adjonesey:
Improves frametime consistency at 60Hz/60fps on Windows platform.

I noticed the framerate jumping between 59fps and 60fps regularly, causing visible hitching. Some games seem to be affected worse than others. 

Here's a before and after of frametime graph:

image

Performance on my system with this change seems about the same - possibly even a couple of % faster (the CPU spends a little less time inside sceKernelGettimeofday)

### DIFF
--- a/src/core/libraries/kernel/time_management.cpp
+++ b/src/core/libraries/kernel/time_management.cpp
@@ -148,7 +148,7 @@ int PS4_SYSV_ABI sceKernelGettimeofday(OrbisKernelTimeval* tp) {
 
 #ifdef _WIN64
     FILETIME filetime;
-    GetSystemTimeAsFileTime(&filetime);
+    GetSystemTimePreciseAsFileTime(&filetime);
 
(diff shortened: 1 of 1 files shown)


par274:
Hello, does this fix, fix the physics issues in the Bloodborne 60 FPS patch?

https://github.com/shadps4-emu/shadPS4/issues/1226

adjonesey:
> Hello, does this fix, fix the physics issues in the Bloodborne 60 FPS patch?
> 
> #1226

Hey, I'm not aware of any change to physics as a result of this. The purpose of this change is to improve visual smoothness/reduce hitching

par274:
> > Hello, does this fix, fix the physics issues in the Bloodborne 60 FPS patch?
> > #1226
> 
> Hey, I'm not aware of any change to physics as a result of this. The purpose of this change is to improve visual smoothness/reduce hitching

I just tested it. It seems like it has definitely reduced the problem. And the drops are much less.

Issue #1348, opened 2024-10-11
Emulator freezes when using the fullscreen setting in Linux


robyn-dressler:
Environment: Linux / OpenSUSE Tumbleweed / X11 / Nvidia RTX 3080Ti

After turning on the Enable Fullscreen option in the configuration and starting the emulator, the screen completely freezes, and nothing gets drawn onto the screen. The program doesn't crash, so I had to kill the process manually. Moving windows across the screen glitches out like this:

![shadps4-fullscreen-bug](https://github.com/user-attachments/assets/fed35257-1817-4166-bb50-527b7cd06ee1)

I ran into this while trying to test a new feature for turning on fullscreen with an `-f` flag through the CLI, however this is reproducible through the normal TOML or GUI config as well.

Log:
[shad_log.txt](https://github.com/user-attachments/files/17348091/shad_log.txt)

Hermiten:
Can you try again ? I will close this because no update since 1 month but normally this is fixed.

Issue #1147, opened 2024-09-29
Fullscreen CLI linux


ghost:
Hello, 
Is there a flag to run the SDL-Appimage as fullscreen on linux? I'm trying to create a launcher for batocera?  Or better, yet when i tried a script to the qt app image /userdata/system/pro/Shadps4-Qt-AppImage "/path/to/.eboot" it doesn't launch with that.  (Would prefer launching via qt.)

Thanks.

qurious-pixel:
The quick way is to edit the `config.toml` in the user folder. 

```
[General]
Fullscreen = true
```

Then launch as normal `/userdata/system/pro/Shadps4-sdl-AppImage "/path/to/.eboot" `

This seems like a Feature Request that would make a good first issue for someone.

robyn-dressler:
> This seems like a Feature Request that would make a good first issue for someone.

If no one else is working on this, I'd be happy to take a look.

LordDaveTheKind:
If you are on Wayland, have you already considered Valve Gamescope?

qurious-pixel:
They are using Batocera, and want a switch to open the game window fullscreen when using SDL. 
There is already a patch switch `-p` so adding an `-f/--fullscreen` can be done similiarly.

kalaposfos13:
Can this be closed, as [my PR](https://github.com/shadps4-emu/shadPS4/pull/1356) has this implemented?

polybiusproxy:
It can be closed once it's merged

kalaposfos13:
Sorry, linked the wrong [PR](https://github.com/shadps4-emu/shadPS4/pull/1507)

PR #1349, opened 2024-10-12
Allowing Config class to resolve isFullscreen flag from CLI arguments

robyn-dressler:
Addresses feature request #1147

Changes:
- Move CLI argument parsing to `Config` class
- Added `-f` and `--fullscreen` flags that override `isFullscreen` in the config
- Added `patchFile` member variable to config (but excluding from TOML file, since it's game specific)
- Added second overload to `Emulator::Run`

I have unfortunately not been able to properly test fullscreen in my environment due to #1348, but I think the new CLI argument works.


src/emulator.cpp
```diff
@@ -256,6 +256,14 @@ void Emulator::Run(const std::filesystem::path& file) {
     std::exit(0);
 }
 
+void Emulator::Run(int& argc, char* argv[]) {
```
SachinVin: Why is `argc` a reference?

LNDF:
It might be better to use an arg parsing library like Boost program_options

qurious-pixel:
linux-sdl appimage works when entering path `./Shadps4-sdl.AppImage /path/to/eboot.bin -f` or `--fullscreen`.
Used to `./Shadps4-sdl.AppImage --fullscreen /path/to/eboot.bin` order convention, but this fails with case-sensitive error message `Eboot.bin file not found`. Whereas removing the switch, it boots into the game.

src/emulator.cpp
robyn-dressler: Good question, I don't know. I'll pass by value instead

robyn-dressler:
@LNDF I really would like to use a library for this, but I was hesitant to introduce a new dependency to `common`, because of this comment in the style documentation:
> Don't ever introduce new external dependencies into Core

If that doesn't apply here, then I could rework this to use the Boost library. That would also probably resolve the ordering issue @qurious-pixel mentioned.

LNDF:
Boost is already a dependency. You could try to open a PR in https://github.com/shadps4-emu/ext-boost adding boost/program_options.hpp using bcp.

georgemoralis:
done in #1507  among other things

DuckSleazzy:
I wasn't sure whether to open a new issue, or comment here.

On current release 0.5.0, running a game via CLI completely ignores `-f/--fullscreen`. I tried running Bloodborne with `shadps4.exe -g <path> -f true`. While the game itself boots fine, it is not fs. Later on pressing F11 makes it fs though.

Specs:
- Windows 11 24H2
- Ryzen 5800x
- RX 6650 XT
- 2x8GB 3600 CL18
- Aorus Gen4 1TB (OS), Samsung 860 EVO (Game and Emulator)

ngoquang2708:
@DuckSleazzy It is fixed in the pre-release/nightly build.

PR #1358, opened 2024-10-12
Devtools - Inspect regs/User data/Shader disassembly

viniciuslrangel:
image


georgemoralis:
Holy cow so many dev windows!

PR #1364, opened 2024-10-13
improved documentation + better toolbar icons

Xphalnos:
I improved the quickstart and the readme a bit.
I put better icons for the toolbar (Same thickness, same style...).
I also updated the submodules (Tested on Dark Souls Remastered, Tomb Raider Remastered and Red Dead Redemption).

Before:
Before

After:
After


Hermiten:
New icons are great. 
I don't know if it's a good idea to update every week the submodule without checking the change log of everyone if something important change. But I'm probably wrong

Xphalnos:
> New icons are great. I don't know if it's a good idea to update every week the submodule without checking the change log of everyone if something important change. But I'm probably wrong

Major changes come from glslang and xbyak, from what I tested nothing breaks the emulator

PR #1367, opened 2024-10-13
video_core: float clear values

LNDF:
Add float clear values when num_bits = 64. Apparently used in UE4 games. 

Draft because it needs testing.

@ElBread3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -771,7 +771,7 @@ endif()
 
 create_target_directory_groups(shadps4)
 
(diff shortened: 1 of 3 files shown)


LNDF:
Implemented SwapStandardReverse and SwapAlternativeReverse

LNDF:
Currently we are assuming that there are always 4 components. That may not be true.

LNDF:
Drafting since it is missing format and correct swap handling.

LNDF:
Closed and replaced by #1381

PR #1396, opened 2024-10-16
Disabling discord-rpc warnings

Xphalnos:
This disables all discord-rpc warnings going from +6000 lines of compilation log to less than 2000 (for Windows-Qt).


abouvier:
This will also set the flag for the other targets created later in this file. You should use [target_compile_options](https://cmake.org/cmake/help/latest/command/target_compile_options.html) instead.
And since `discord-rpc` is a [internal fork](https://github.com/shadps4-emu/ext-discord-rpc), you should put the flag there instead.

PR #1397, opened 2024-10-16
Load Eboot/Modules from Separate Update Folder

ElBread3:
Load eboot and sce_module modules from separate update folder. Should make separate updates finally complete, but this will need to be tested with lots of games with different module setups.

version-number


georgemoralis:
you probably mean sce_module not sys_modules , sys_modules are from fw and doesn't need to be different

ElBread3:
> you probably mean sce_module not sys_modules , sys_modules are from fw and doesn't need to be different

Yep I did, I meant the plural of "sce_module", has now been corrected in PR description too
Wow it is very easy to type sys instead of sce holy crap

PR #1416, opened 2024-10-19
cmake: Allow disabling discord rpc

raphaelthegreat:
Discord RPC is basically abandonware at this point and may not compile everywhere


georgemoralis:
probably need an #ifdef in the file that calls it as well

polybiusproxy:
What other alternatives do we have?

raphaelthegreat:
> probably need an #ifdef in the file that calls it as well

Done

raphaelthegreat:
> What other alternatives do we have?

I don't think there is an open alternative to this. Discord has some sdk stuff for integration but haven't done any research if they can be used, and nobody seems to use them anyway

georgemoralis:
there is a new interface for discord as i remember

PR #1419, opened 2024-10-20
Fix for new versions of magic_enum + cleanup

Xphalnos:
Since this commit **https://github.com/Neargye/magic_enum/pull/374**, the latest version of **magic_enum** requires changes in **externals/CMakeLists.txt**.
I took the opportunity to put the dependencies back in alphabetical order.

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -3,10 +3,7 @@
 
 set(BUILD_SHARED_LIBS OFF)
 set(BUILD_TESTING OFF)
-set_directory_properties(PROPERTIES
-    EXCLUDE_FROM_ALL ON
-    SYSTEM ON
-)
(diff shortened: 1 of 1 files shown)


abouvier:
The real fix is to use `#include <magic_enum/magic_enum.hpp>` instead of `#include <magic_enum.hpp>`, and exclude the current release with `find_package(magic_enum 0.9.7 CONFIG)`.

Xphalnos:
> The real fix is to use `#include <magic_enum/magic_enum.hpp>` instead of `#include <magic_enum.hpp>`, and exclude the current release with `find_package(magic_enum 0.9.7 CONFIG)`.

However, the latest version is [**0.9.6**](https://github.com/Neargye/magic_enum/releases).

abouvier:
> However, the latest version is [**0.9.6**](https://github.com/Neargye/magic_enum/releases).

That's why if you update the submodule and use an unreleased commit, then you need to require the next version, because the current one will not work with the change.

PR #1421, opened 2024-10-20
Implement Close Button

Xphalnos:
I implemented the Close button because before when you press it nothing happens.

Video Demo:

https://github.com/user-attachments/assets/0607c5c9-4e0f-4da9-b46f-086cf209bc0d

### DIFF
--- a/src/qt_gui/main_window.cpp
+++ b/src/qt_gui/main_window.cpp
@@ -4,6 +4,7 @@
 #include <QDockWidget>
 #include <QKeyEvent>
 #include <QProgressDialog>
+#include <SDL3/SDL_events.h>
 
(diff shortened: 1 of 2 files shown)


polybiusproxy:
I'm not sure if this should be merged. This is, at the moment, just an alternative to closing the game window, which results in a crash.
Ideally we'd want to have proper destruction of everything on the emulator, take for example, the renderer.

Xphalnos:
> I'm not sure if this should be merged. This is, at the moment, just an alternative to closing the game window, which results in a crash. Ideally we'd want to have proper destruction of everything on the emulator, take for example, the renderer.

I agree but for now it's a very basic implementation of this button.

PR #1431, opened 2024-10-22
recompiler: add use tracking

baggins183:
Add use tracking for instructions

Still want to add iterator adaptor for (distinct) users in addition to uses.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,10 @@ if (NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Release)
 endif()
 
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+    add_compile_definitions(_DEBUG)
+endif()
+
 project(shadPS4)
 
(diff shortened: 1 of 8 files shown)


baggins183:
on previous commit I stored uses (inst, index) in a list and used an adaptor to iterate over users.
That was broken though because apparently std:: or boost::unique modify the underlying container.
I changed it so per Inst, users are stored in a list, and nodes in the list hold a user + bitmask of the arg indices that use *this.
Needs a lot more lines of code. I wonder if there's to fix the old way so its simpler.

Still needs boost submodule update

georgemoralis:
i don't see a pr for it on https://github.com/shadps4-emu/ext-boost :/

georgemoralis:
@baggins183 if you can pr on boost the neccesary change so we can progress with that pr too

baggins183:
@georgemoralis here u go boss
https://github.com/shadps4-emu/ext-boost/pull/9

georgemoralis:
merge you need to update submodule here so we can see if it builds ;D

baggins183:
Closing since most of this made it in with ssa fixes

PR #1436, opened 2024-10-23
core/libraries: IME implementation

polybiusproxy:
Implementation of the IME library, used for opening an IME keyboard to interact with the game.


georgemoralis:
εικόνα

Seems ok :)

PR #1439, opened 2024-10-23
shader_recompiler: patch fmask access instructions

LNDF:
Patch some instructions that access fmask image bound to shader.

This PR patches those reads into constant values and removes the binding all together.

Probable we need to patch more opcodes and do some more testing.

Journey now doesn't assert on liverpool_to_vk.cpp. Insteed it crashes further on image.cpp::236 on the following line of code of the VideoCore::Image::Transit function:

```c++
ASSERT(subres_idx < subresource_states.size());
```

I don't think this is related to this PR bu I might be wrong.

More testing with more games that access fmask is needed.

### DIFF
--- a/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
+++ b/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
@@ -147,7 +147,7 @@ public:
     explicit Descriptors(Info& info_)
         : info{info_}, buffer_resources{info_.buffers},
           texture_buffer_resources{info_.texture_buffers}, image_resources{info_.images},
-          sampler_resources{info_.samplers} {}
+          sampler_resources{info_.samplers}, fmask_resources(info_.fmasks) {}
 
     u32 Add(const BufferResource& desc) {
         const u32 index{Add(buffer_resources, desc, [&desc](const auto& existing) {
@@ -193,6 +193,14 @@ public:
         return index;
     }
 
+    u32 Add(const FMaskResource& desc) {
+        u32 index = Add(fmask_resources, desc, [&desc](const auto& existing) {
+            return desc.sgpr_base == existing.sgpr_base &&
+                   desc.dword_offset == existing.dword_offset;
+        });
+        return index;
+    }
+
 private:
     template <typename Descriptors, typename Descriptor, typename Func>
     static u32 Add(Descriptors& descriptors, const Descriptor& desc, Func&& pred) {
@@ -209,6 +217,7 @@ private:
     TextureBufferResourceList& texture_buffer_resources;
     ImageResourceList& image_resources;
     SamplerResourceList& sampler_resources;
+    FMaskResourceList& fmask_resources;
 };
 
 } // Anonymous namespace
@@ -673,6 +682,19 @@ void PatchImageInstruction(IR::Block& block, IR::Inst& inst, Info& info, Descrip
         case IR::Opcode::ImageQueryLod:
             inst.ReplaceUsesWith(ir.Imm32(1));
             return;
+        case IR::Opcode::ImageQueryDimensions: {
+            IR::Value dims = ir.CompositeConstruct(ir.Imm32(static_cast<u32>(image.width)), // x
+                                                   ir.Imm32(static_cast<u32>(image.width)), // y
+                                                   ir.Imm32(1), ir.Imm32(1)); // depth, mip
+            inst.ReplaceUsesWith(dims);
+
+            // Track FMask resource to do specialization.
+            descriptors.Add(FMaskResource{
+                .sgpr_base = tsharp.sgpr_base,
+                .dword_offset = tsharp.dword_offset,
+            });
+            return;
+        }
         default:
             UNREACHABLE_MSG("Can't patch fmask instruction {}", inst.GetOpcode());
         }
--- a/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
+++ b/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
@@ -685,24 +685,24 @@ void PatchImageInstruction(IR::Block& block, IR::Inst& inst, Info& info, Descrip
 
     // Now that we know the image type, adjust texture coordinate vector.
     IR::Inst* body = inst.Arg(1).InstRecursive();
-    const auto [coords, arg] = [&] -> std::pair<IR::Value, IR::Value> {
+    const auto [coords, arg1, arg2] = [&] -> std::tuple<IR::Value, IR::Value, IR::Value> {
         switch (image.GetType()) {
         case AmdGpu::ImageType::Color1D: // x
-            return {body->Arg(0), body->Arg(1)};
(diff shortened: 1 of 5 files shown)


src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -656,6 +656,32 @@ void PatchImageInstruction(IR::Block& block, IR::Inst& inst, Info& info, Descrip
     }
     ASSERT(image.GetType() != AmdGpu::ImageType::Invalid);
     const bool is_storage = IsImageStorageInstruction(inst);
+
+    // Patch image instruction if image is fmask.
+    if (image.IsFmask()) {
+        // Ignore fmask writes. TODO: handle dimension queries.
```
psucien: I'm not sure if FMask will be written in shader (but not excluding such possibility), so will be better to make an assertion on `is_storage` there

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -656,6 +656,32 @@ void PatchImageInstruction(IR::Block& block, IR::Inst& inst, Info& info, Descrip
     }
     ASSERT(image.GetType() != AmdGpu::ImageType::Invalid);
     const bool is_storage = IsImageStorageInstruction(inst);
+
+    // Patch image instruction if image is fmask.
```
psucien: I would prefer to use `FMask` here and in other places as it is official terminology

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -656,6 +656,32 @@ void PatchImageInstruction(IR::Block& block, IR::Inst& inst, Info& info, Descrip
...
+    if (image.IsFmask()) {
+        // Ignore fmask writes. TODO: handle dimension queries.
+        if (is_storage && inst.GetOpcode() == IR::Opcode::ImageQueryDimensions) {
```
psucien: In case of dimensions query, better to put assert too or at least emit some error log message

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -685,24 +711,25 @@ void PatchImageInstruction(IR::Block& block, IR::Inst& inst, Info& info, Descrip
 
     // Now that we know the image type, adjust texture coordinate vector.
     IR::Inst* body = inst.Arg(1).InstRecursive();
-    const auto [coords, arg] = [&] -> std::pair<IR::Value, IR::Value> {
+    const auto [coords, arg1, arg2] = [&] -> std::tuple<IR::Value, IR::Value, IR::Value> {
         switch (image.GetType()) {
         case AmdGpu::ImageType::Color1D: // x
-            return {body->Arg(0), body->Arg(1)};
+            return {body->Arg(0), body->Arg(1), body->Arg(2)};
```
psucien: why 1d images are affected?

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -685,24 +711,25 @@ void PatchImageInstruction(IR::Block& block, IR::Inst& inst, Info& info, Descrip
...
-        case AmdGpu::ImageType::Color2DArray: // x, y, slice
             [[fallthrough]];
-        case AmdGpu::ImageType::Color2DMsaa: // x, y, frag
+        case AmdGpu::ImageType::Color2DMsaa: // x, y. (sample is passed on different argument)
+            return {ir.CompositeConstruct(body->Arg(0), body->Arg(1)), body->Arg(2), body->Arg(3)};
+        case AmdGpu::ImageType::Color2DArray: // x, y, slice
             [[fallthrough]];
         case AmdGpu::ImageType::Color3D: // x, y, z
-            return {ir.CompositeConstruct(body->Arg(0), body->Arg(1), body->Arg(2)), body->Arg(3)};
+            return {ir.CompositeConstruct(body->Arg(0), body->Arg(1), body->Arg(2)), body->Arg(3),
```
psucien: ..and 3d and cubes?

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -711,7 +738,12 @@ void PatchImageInstruction(IR::Block& block, IR::Inst& inst, Info& info, Descrip
 
     if (inst_info.has_lod) {
         ASSERT(inst.GetOpcode() == IR::Opcode::ImageFetch);
-        inst.SetArg(3, arg);
+        inst.SetArg(3, arg1);
+        if (image.GetType() == AmdGpu::ImageType::Color2DMsaa) {
```
psucien: MS textures don't have mips, so I would prefer to have an assertion for this case instead

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
LNDF: I will implement that before merge

LNDF:
I added patching for ImageQueryDimensions. BufferReadFormatF32 and BufferWriteFormatF32 are still unpatched but I don't know if we need to patch them.

hgh32:
This PR looks like added new assertions which caused regressions in bloodborne alpha, possibly only on specific hardware
[shad_log.txt](https://github.com/user-attachments/files/17695084/shad_log.txt)

LNDF:
> This PR looks like added new assertions which caused regressions in bloodborne alpha, possibly only on specific hardware [shad_log.txt](https://github.com/user-attachments/files/17695084/shad_log.txt)

Is this log in sync? I cannot see the assert at the end.

StevenMiller123:
@LNDF Their log has `[Debug] <Critical> resource_tracking_pass.cpp:operator():721: Assertion Failed!` logged in it.

LNDF:
> @LNDF Their log has `[Debug] <Critical> resource_tracking_pass.cpp:operator():721: Assertion Failed!` logged in it.

Thx I will take a look later

kalaposfos13:
Alien Isolation has the exact same assert right before main menu, commenting it out progresses the game to menus up until trying to enter the game, seemingly with no problems. From some testing, the image type is AmdGpu::ImageType::Color1D (edit: this is only with the kernel rewrite afaik, on main it crashes even sooner)

LNDF:
Can you provide the shaders that trigger the assert? Both on bb alpha and alien isolation (I cannot access the games).

Also @hgh32 what is the image type?

kalaposfos13:
[shad_log.txt.zip](https://github.com/user-attachments/files/17707624/shad_log.txt.zip)
The original log was 45 MB lol
~~I don't know how to get what shader triggered this, I'll do it if you can help me out~~ (edit: I found it)

kalaposfos13:
[dumps_alien_isolation.zip](https://github.com/user-attachments/files/17707707/dumps_alien_isolation.zip)
Enabling shaders dumping gave this, I guess this is what you're after

PR #1440, opened 2024-10-23
kernel: Rewrite pthread emulation

raphaelthegreat:
The current implementation of pthreads uses winpthreads on windows and pthread lib on linux. While it has served its purpose for quick prototyping, the code at the moment is very unorganized and messy. The emulator does not perform a lot of necessary checks to validate state. Note: This is linux only and draft at the moment

This has been in works for a while but aims to fully rewrite pthread primitives such as mutexes, condvars, rwlocks and threads to be accurate to the real freebsd libthr library. Thus we dont depend on an old library for it and have more control.

* Mutexes/Rwlocks now implement all static initializers and error checks as libthr. For mutexes recursion is handled manually and std::timed_mutex is used from STL as a sync primitive, while std::shared_timed_mutex for rwlocks. The PthreadMutex struct behaves like a lockable type so it can be passed to std::condition_variable directly to ensure correct ownership tracking. Adaptive mutexes have been implemented with spinloops.
* Condition variables are implemented on top of std::condtion_variable_any. It's a very slim wrapper at the moment but it works okay from testing. Before this PR is merged I want to experiment with implementing sleepqueues from libthr and using waitable atomics for sleeping threads. This will allow us to implement scePhtreadCondSignalTo properly.
* Thread local storage has been improved. A major bug in __tls_get_addr has been uncovered and fixed where an incorrect module TLS region could be loaded. On main this is masked because all DTVs are filled on thread creation but this isn't what real kernel does so we don't do that anymore either. May fix crashes on games that load custom prx modules with TLS regions. DTVs are now also freed when a thread is destoyed which may improve memory leaks on applications that create and destroy many threads.
* Thread management has been completely reworked to be more accurate. Cleanup functions are accurate to original library and thread-safe, thread joining/exiting are also performed manually with cancellation points in mind. Pthread recycling and garbage collection is implemented as well. Thread objects are allocated with a slab heap using memory from the system reserved region, which mimics how actual kernel uses sceLibcInternal to allocate objects. This fixes Bloodborne on Linux freezes and removed the need for the annoying hack.

TLS allocation is now performed by the caller thread, while the target thread simply sets the guest tcb. This means that calling guest code from host threads is a bit more tricky as we can't easily setup TLS without the help of the kernel. For this a Thread class has been added in kernel which mimics an std::jthread like interface and uses pthread call underneath. This is useful for avplayer HLE.

Thread stack allocation is also implemented but not wired up yet as it needs to more testing. This may help games like TLOU2 that query thread stacks from the memory system.


raphaelthegreat:
Another thing worth mentioning. Before a lot functions were duplicated as they had both posix and sce variants. Now there is an ORBIS() macro which wraps posix functions and adjusts the error code, which is how the actual libkernel works. This reduces code duplication significantly

Shoegzer:
Wow, nice PR.  ~Bloodborne still crashing on startup though, unfortunately.~ Actually nevermind this, it was a local issue that has been resolved.  The title runs, albeit without sound but I assume that's expected at this point.

pflyly:
Man, you are hero! Finally, Bloodborne now works great on my steamdeck with a stable 30 fps! I used a couple of mods for better performance of course. But still, it's impressive.

GHU7924:
I don't have the option to install any Linux distribution on my PC yet,
but I do have a Steam Deck OLED.

Steam Deck, like shadPS4, also has several update channels, so:
in the Stable version of Steam OS - shadPS4 linux-qt does not react in any way and does not even start,
in the Beta version of Steam OS and with the latest updates (I previously tried to run shadPS4 linux-qt),
the result was better, the emulator at least it was running, but when Bloodborne was running, I just had a green square on the screen and then nothing happened.

When I saw this "Pull request", I decided to try my luck again.
I have launched Build ae8ec44, a game without mods (only a mod was installed that replaced the PS4 buttons on Xbox), and WOW!!! 
For the first time, Bloodborne has launched on my Steam Deck.

I got to the main menu, selected "New Game", watched the video, but everything hung up on character creation (maybe it would have been worth waiting a little longer and everything would have continued, but I'm not sure),
so I just took my saves from shadPS4 windows-qt to see if anything would change,
and my save worked, I was able to boot into Hunter's Dream locations.
(Some textures were missing (or invisible) and so on.)

Outside of Hunter's Dream, I didn't have a chance to run much, usually there is a hang or the emulator closes, but I tried the New game again. I did create a character, I was even able to run through Iosefka's Clinic a little, but an invisible Scourge Beast interrupted my walk. (Some textures and stuff were also missing).

As a result, I can say that there is definitely progress, and even such a step I consider a job well done. 
Raphael the great - you are a legend!

Edit:

I returned from Beta Steam OS to Stable Steam OS, and it seems like I was able to run the emulator even so.
Steam OS - 3.6.19
30fps lock in Steam OS
Patches: Skip Intro, Disable Chromatic Aberration, Disable Motion Blur, Force Enable Old Hunters DLC, Resolution Patch (360p)







[shad_log.txt](https://github.com/user-attachments/files/17512129/shad_log.txt)

Inui-senpai:
Just tried the `windows-qt` artifact (build [9b142c3](https://github.com/shadps4-emu/shadPS4/pull/1440/commits/9b142c33f3ebd91955e965d4e1d00906a4c9dde4)), out of curiosity. DDLC+ crashed with this error:
```
[Debug] <Critical> exception.cpp:sceKernelInstallExceptionHandler:77: Unreachable code!
Missing exception implementation
```
[shad_log.txt](https://github.com/user-attachments/files/17562756/shad_log.txt)

ElBread3:
> ```
> [Debug] <Critical> exception.cpp:sceKernelInstallExceptionHandler:77: Unreachable code!
> Missing exception implementation
> ```

DDLC+ uses the Unity engine, which requires exception handling to run, which isn't supported on Windows currently. I assume Team Salvato didn't want to port Python (the original's engine is RenPy) to consoles, which is why they used Unity.

rafael-57:
Bloodborne, first time getting this:
[shad_log.txt](https://github.com/user-attachments/files/17686566/shad_log.txt)

Any idea what's causing it? Seems to crash very constantly when trying to load a savefile.

Shoegzer:
Having difficulties building with this PR merged, with errors such as:

```
In file included from /home/testing/Desktop/shadps4/shadPS4/src/core/libraries/kernel/threads/condvar.cpp:9:
/home/testing/Desktop/shadps4/shadPS4/src/core/libraries/kernel/threads/pthread.h:221:35: error: implicit instantiation of undefined template 'std::atomic<Libraries::Kernel::PthreadOnceState>'
  221 |     std::atomic<PthreadOnceState> state;
```

* OS: Linux Mint 22 (Ubuntu 24.04 base) / kernel 6.8
* Clang version: 18.1.3
* CMake version: 3.28.3
* Build tested: ce15829  (this PR merged)
* Last known good build: 75d2181 (this PR merged)
* Build log: [build.log](https://github.com/user-attachments/files/17779997/build.log)

jhauck2:
> Having difficulties building with this PR merged, with errors such as:
> 
> ```
> In file included from /home/testing/Desktop/shadps4/shadPS4/src/core/libraries/kernel/threads/condvar.cpp:9:
> /home/testing/Desktop/shadps4/shadPS4/src/core/libraries/kernel/threads/pthread.h:221:35: error: implicit instantiation of undefined template 'std::atomic<Libraries::Kernel::PthreadOnceState>'
>   221 |     std::atomic<PthreadOnceState> state;
> ```
> 
>     * OS: Linux Mint 22 (Ubuntu 24.04 base) / kernel 6.8
> 
>     * Clang version: 18.1.3
> 
>     * CMake version: 3.28.3
> 
>     * Build tested: [ce15829](https://github.com/shadps4-emu/shadPS4/commit/ce158290fbd7871a3f890323501e6fe3e8bec832)  (this PR merged)
> 
>     * Last known good build: [75d2181](https://github.com/shadps4-emu/shadPS4/commit/75d2181489d45dc72feba98c96ce9b2f065835a4) (this PR merged)
> 
>     * Build log: [build.log](https://github.com/user-attachments/files/17779997/build.log)
I was able to get this to build by adding `#include <atomic>` in pthread.h

Edit: OS: Fedora 41 / kernel 6.11.7-300

bigol83:
With latest commit Red Dead Redemption crashes with this error immediately after booting

```
[Debug] <Critical> address_space.cpp:operator():182: Assertion Failed!
Unmap operation on virtual_addr=0X18D800000 failed: (FormatMessageA failed to format error)

```
[shad_log.txt](https://github.com/user-attachments/files/17784290/shad_log.txt)

src/core/libraries/kernel/threads/pthread.h
```diff
@@ -0,0 +1,359 @@
...
+class SymbolsResolver;
+}
+
+namespace Libraries::Kernel {
+
+struct Pthread;
+
+enum class PthreadMutexFlags : u32 {
+    TypeMask = 0xff,
+    Defered = 0x200,
```
kalaposfos13, line 28: I believe this is a typo, but 'defered' is used for at least one other variable name, so I'm not sure

bigol83:
Latest commit fixed Red Dead Redemption crash 👍

Linear524:
I don't know for sure, but today something broke LocoRoco remastered PS4 title...
Logs doesn't show any error, just normal routines.
Yesterday LocoRoco was perfectly playable just before last update...

polybiusproxy:
image
LocoRoco has not regressed

Linear524:
I just get black screen and nothing more :(

Linear524:
And here is the logs from working build

polybiusproxy:
This does not give me enough information unfortunately

thecatontheceiling:
Regressed a few LEGO games (including LEGO Hobbit and LEGO Batman)

PR #1441, opened 2024-10-24
kernel: Fix new pthreads code on macOS.

squidbus:
PR is for the `more-kernel` branch, not `main`.

* Adds missing includes.
* Casts timespec duration to be supported by the clock type. Really I think this should be using `steady_clock` anyway since `system_clock` is sensitive to user time changes on all platforms, but at least as far as compatibility with macOS is concerned casting should be fine for now.
* Cleans up the TCB management code for macOS to work with the new thread code. Primarily there were two issues:
  * Relied on `SetTcbBase(nullptr)` when a thread finished running for cleanup, which was no longer being done. To make things simple it now just uses a `pthread_key_t` with a destructor to make sure the LDT allocations get cleaned up. It also guards against multiple `SetTcbBase` calls in a thread by reusing the allocated page.
  * The `ldt_entry` was not getting zero initialized and could contain incorrect data depending on the stack, and some of the threading changes exposed this problem.

Tested a few games and the changes are working on macOS now.

### DIFF
--- a/src/core/libraries/kernel/threads.h
+++ b/src/core/libraries/kernel/threads.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <functional>
+#include "common/polyfill_thread.h"
 #include "core/libraries/kernel/threads/pthread.h"
 
 namespace Core::Loader {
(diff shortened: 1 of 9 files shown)


georgemoralis:
of course MacOS fixes before windows . What world we are leaving at... ;/

PR #1442, opened 2024-10-24
kernel: Pthread rewrite touchups for Windows

polybiusproxy:
Fixes to make the recent kernel rewrite work on Windows. Might not compile on other platforms


src/core/libraries/kernel/threads/pthread.h
```diff
@@ -227,7 +238,8 @@ enum class ThreadListFlags : u32 {
     InGcList = 4,
 };
 
-using PthreadEntryFunc = void* (*)(void*);
+using PthreadEntryFunc = void* PS4_SYSV_ABI (*)(void*);
+using PthreadEntryHostFunc = void* (*)(void*);
```
squidbus: Feel like the use cases for this should just be made `PS4_SYSV_ABI`, having things take a function arg that can be two different ABIs does not sound like a good idea.

squidbus:
This should probably be marked draft since there are still things that are just arbitrarily commented out.

Randomuser8219:
Judging by GitHub Actions the Windows build for this PR doesn't compile due to errrors.

PR #1448, opened 2024-10-25
Load LLE modules from sys_modules/GAMEID folder

georgemoralis:
Since it appears that some games requires lle modules that other games doesn't seem to like you can place them in a specific folder so only that game load them

for gravity rush e.g you can create a folder under sys_modules folder like

sys_modules/CUSA01130  

and place the font libraries there

libSceFont.sprx
libSceFontFt.sprx
libSceFreeTypeOt.sprx

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -229,7 +229,7 @@ void Emulator::Run(const std::filesystem::path& file) {
     linker->LoadModule(eboot_path);
 
     // check if we have system modules to load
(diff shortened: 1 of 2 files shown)


leon22heart:
Not a bad idea .But wouldn't be better to incorporate this to a per game config system ? Since the wiki is not yet complete , the per game info changes during each commit , wouldn't we get filled with LLE folders in which we have to manually track ? Except if this is a debug only workaround as a pre-alpha game profile config.

PR #1451, opened 2024-10-26
texture_cache: Implement subresource specific uploads

raphaelthegreat:
It is a common pattern for games to clear specific subresources of an image (either a mipmap or a layer). Bloodborne renders to an texture cube by individually clearing the faces with compute dispatches and rendering to them. Red Dead Redemption  clears the first mipmap of images that have many. Without this handling subsequent usages of said image can either clear subresources that were rendered or trash the image with cpu data if a gpu buffer doesn't happen to cover the entire image size.

Before merging, support for images with more than 64 subresources needs to be added (will trigger an assert for now).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ endif()
 
 option(ENABLE_QT_GUI "Enable the Qt GUI. If not selected then the emulator uses a minimal SDL-based UI instead" OFF)
 option(ENABLE_DISCORD_RPC "Enable the Discord RPC integration" ON)
-CMAKE_DEPENDENT_OPTION(ENABLE_USERFAULTFD "Enable write tracking using userfaultfd on unix" ON "NOT LINUX OR APPLE" OFF)
+option(ENABLE_USERFAULTFD "Enable write tracking using userfaultfd on unix" OFF)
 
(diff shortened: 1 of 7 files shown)


bigol83:
Red Dead Redemption has random green textures, sometimes they are on buildings, other times they are on grass
Cattura

mizinin:
> Red Dead Redemption has random green textures, sometimes they are on buildings, other times they are on grass

Same in Bloodborne

raphaelthegreat:
There is a complement change to this PR in the works that should fix the RDR issue, as this PR exposes the preexisting problem. For BB not sure, we will see

MajorP93:
> There is a complement change to this PR in the works that should fix the RDR issue, as this PR exposes the preexisting problem. For BB not sure, we will see

Any updates on this PR?
Is the complement change done already?

v-fox:
For anyone interested, I've made a sloppy copy-paste+rename rebase of this PR for snapshot 81e7e4b2: [shadPS4-PR1451.patch.zip](https://github.com/user-attachments/files/18513341/shadPS4-PR1451.patch.zip)
It compiled but I don't guarantee that I did not screw it up in the process.

Xcedf:
@v-fox thanks for the patch, finally manged to try this PR, helps with some of texture corruption issues in GoW3

Randomuser8219:
Why did this PR become so inactive?

PR #1452, opened 2024-10-27
Publish to Release Repository

DanielSvoboda:
Stores the builds in 3 repositories, one for each operating system:

- shadps4-binaries-Windows
- shadps4-binaries-Linux
- shadps4-binaries-Mac 

This makes it possible to download previous builds, which will be used as intended by shadown...
image


polybiusproxy:
I personally don't believe having a repository for each platform is a good thing. If a user needs a previous build they can go to the commit's actions or build it themselves

PR #1453, opened 2024-10-27
AJM code improvements

roamic:
* Replaced a lot of AJM structures to generic chunks
* Improved the readability and structure of batch encoding/decoding


rafael-57:
Tested BB, seems way more stable but works the same as the source ajm branch. Most audio doesn't play or plays inconsistently for a little bit.

AJM branch without these changes seens much more prone to crash/hangs while getting ingame. I didn't get any crash after 4 boots with this build.
[shad_log.txt](https://github.com/user-attachments/files/17534071/shad_log.txt)

PR #1484, opened 2024-11-05
Videocodec implementation

georgemoralis:
Based on @polybiusproxy  videodec2 implementation . Seems to work , reply with tests thanks :D


bigol83:
Resident Evil 4 2005 fmv works without issues

Immagine 2024-11-05 124118

ElBread3:
think this could be rebased so I can test with ajm?

georgemoralis:
done

src/core/libraries/videodec/videodec_impl.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+        CopyNV12Data((u8*)pFrameBufferInOut.pFrameBuffer, *frame);
+
+        pPictureInfoOut.codecType = 1;
+        pPictureInfoOut.frameWidth = frame->width;
+        pPictureInfoOut.frameHeight = frame->height;
+        pPictureInfoOut.framePitch = frame->linesize[0];
+
+        pPictureInfoOut.isValid = true;
+        pPictureInfoOut.isErrorPic = false;
```
polybiusproxy: You're not outputting codec info, games may try to use info such as frame crop offsets.

src/core/libraries/videodec/videodec_impl.cpp
```diff
@@ -0,0 +1,201 @@
...
+        if (frame->format != AV_PIX_FMT_NV12) {
+            AVFrame* nv12_frame = ConvertNV12Frame(*frame);
+            ASSERT(nv12_frame);
+            av_frame_free(&frame);
+            frame = nv12_frame;
+        }
+
+        CopyNV12Data((u8*)pFrameBufferInOut.pFrameBuffer, *frame);
+
+        pPictureInfoOut.codecType = 1;
```
polybiusproxy: Codec type should be zero.

src/core/libraries/videodec/videodec_impl.cpp
```diff
@@ -0,0 +1,201 @@
...
```
polybiusproxy: Codec type should be zero.

polybiusproxy (changes requested):
Looks good at first glance, I do have a few nits though.

To start, this one is quite personal, I quite dislike the lack of empty newlines on some functions--to take one example:
```cpp
int PS4_SYSV_ABI sceVideodecFlush(OrbisVideodecCtrl* pCtrlIn,
                                  OrbisVideodecFrameBuffer* pFrameBufferInOut,
                                  OrbisVideodecPictureInfo* pPictureInfoOut) {
    LOG_INFO(Lib_Videodec, "called");

    if (!pFrameBufferInOut || !pPictureInfoOut) {
        return ORBIS_VIDEODEC_ERROR_ARGUMENT_POINTER;
    }
    if (pFrameBufferInOut->thisSize != sizeof(OrbisVideodecFrameBuffer) ||
        pPictureInfoOut->thisSize != sizeof(OrbisVideodecPictureInfo)) {
        return ORBIS_VIDEODEC_ERROR_STRUCT_SIZE;
    }
    VdecDecoder* decoder = (VdecDecoder*)pCtrlIn->handle;
    if (!decoder) {
        return ORBIS_VIDEODEC_ERROR_HANDLE;
    }
    return decoder->Flush(*pFrameBufferInOut, *pPictureInfoOut);
}
```

I believe it would look better this way:
```cpp
int PS4_SYSV_ABI sceVideodecFlush(OrbisVideodecCtrl* pCtrlIn,
                                  OrbisVideodecFrameBuffer* pFrameBufferInOut,
                                  OrbisVideodecPictureInfo* pPictureInfoOut) {
    LOG_INFO(Lib_Videodec, "called");

    if (!pFrameBufferInOut || !pPictureInfoOut) {
        return ORBIS_VIDEODEC_ERROR_ARGUMENT_POINTER;
    }
    if (pFrameBufferInOut->thisSize != sizeof(OrbisVideodecFrameBuffer) ||
        pPictureInfoOut->thisSize != sizeof(OrbisVideodecPictureInfo)) {
        return ORBIS_VIDEODEC_ERROR_STRUCT_SIZE;
    }

    VdecDecoder* decoder = (VdecDecoder*)pCtrlIn->handle;
    if (!decoder) {
        return ORBIS_VIDEODEC_ERROR_HANDLE;
    }

    return decoder->Flush(*pFrameBufferInOut, *pPictureInfoOut);
}
```
clang-format doesn't mind it, but I do.

Secondly, I ask you to change the names of function arguments and struct members to avoid using Hungarian notation which would go against the whole code style from the emulator, so rename things like `pPictureInfoOut` to `pictureInfo`. It wouldn't make sense either way since you're also using it on references.

Finally, since this class is named `VdecDecoder`, to avoid confusion I believe we should either change the Vdec2 implementation class's name to `Vdec2Decoder` or make a common class for both classes to inherit from.

src/core/libraries/videodec/videodec_impl.cpp
```diff
@@ -0,0 +1,211 @@
...
+    return av_make_error_string(errbuf, AV_ERROR_MAX_STRING_SIZE, errnum);
+}
+#define av_err2str(err) av_err2string(err).c_str()
+#endif // av_err2str
+
+namespace Libraries::Videodec {
+
+static inline void CopyNV12Data(u8* dst, const AVFrame& src) {
+    std::memcpy(dst, src.data[0], src.width * src.height);
+    std::memcpy(dst + (src.width * src.height), src.data[1], (src.width * src.height) / 2);
```
roamic: As far as we are aware of, Vdec1 aligns `width` and `height` to `16` so it needs the conversion logic from `avplayer_source.cpp` where the actual data returned by ffmpeg is laid out in the buffer correctly. This is an assumption though, so needs testing and finding an actual game that breaks if it is not implemented.

src/core/libraries/videodec/videodec_impl.cpp
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        if (frame->format != AV_PIX_FMT_NV12) {
+            AVFrame* nv12_frame = ConvertNV12Frame(*frame);
+            ASSERT(nv12_frame);
+            av_frame_free(&frame);
+            frame = nv12_frame;
+        }
+
+        CopyNV12Data((u8*)pFrameBufferInOut.pFrameBuffer, *frame);
```
roamic, line 107: While each packet usually contains only one frame, it can contain multiple. Judging from this code you re-write the previously written frame if it is the case. Again, not sure it's relevant, but good to keep in mind while testing. The artifacts it can cause is sped up video output.

src/core/libraries/videodec/videodec_impl.cpp
```diff
@@ -0,0 +1,211 @@
...
```
roamic, line 157: Same issue with possible multiple frames per packet.

src/core/libraries/videodec/videodec_impl.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+        pPictureInfoOut.codecType = 0;
+        pPictureInfoOut.frameWidth = frame->width;
+        pPictureInfoOut.frameHeight = frame->height;
+        pPictureInfoOut.framePitch = frame->linesize[0];
+
+        pPictureInfoOut.isValid = true;
+        pPictureInfoOut.isErrorPic = false;
+
+        u32 width = Common::AlignUp((u32)frame->width, 16);
```
roamic, line 167: If you align up you should also correct the `CopyNV12Data` function to account for it.

roamic (approved):

PR #1487, opened 2024-11-05
More Fixes for Separate Update

ElBread3:
Please do not merge quite yet as I would still like to use this PR to fix issues that come up.

- [x] Fix condition so that it should properly fall back on game folder eboot.
- [x] Fix UE4 not booting due to not getting the dir entries for both folders.
- [x] Added info about how separate updates are created for ItemzFlow users.
- [ ] Work on install directory selection features
- [x] Prx loading issues

### DIFF
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -186,4 +186,14 @@ void HandleTable::CreateStdHandles() {
     setup("/dev/stderr", new Devices::Logger("stderr", true));  // stderr
 }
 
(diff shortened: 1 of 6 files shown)


Hermiten:
I post this here, to follow :)

- PKG Extraction is forced, even if you cancel https://github.com/shadps4-emu/shadPS4/issues/1660
- Bulk PKG installation Prompt removal https://github.com/shadps4-emu/shadPS4/issues/1580

Maybe:
- https://github.com/shadps4-emu/shadPS4/issues/1684

PR #1490, opened 2024-11-06
Allow disable the auto updater

ngoquang2708:
Auto updater should not be enabled for local build.

Cc: @DanielSvoboda

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,7 @@ endif()
 
 option(ENABLE_QT_GUI "Enable the Qt GUI. If not selected then the emulator uses a minimal SDL-based UI instead" OFF)
 option(ENABLE_DISCORD_RPC "Enable the Discord RPC integration" ON)
+option(ENABLE_UPDATER "Enables the options to updater" ON)
 
 # First, determine whether to use CMAKE_OSX_ARCHITECTURES or CMAKE_SYSTEM_PROCESSOR.
 if (APPLE AND CMAKE_OSX_ARCHITECTURES)
@@ -732,6 +733,12 @@ set(EMULATOR src/emulator.cpp
 if(ENABLE_QT_GUI)
 qt_add_resources(RESOURCE_FILES src/shadps4.qrc)
 
+if (ENABLE_UPDATER)
+    set(UPDATER src/qt_gui/check_update.cpp
+                src/qt_gui/check_update.h
+    )
+endif()
+
 set(QT_GUI src/qt_gui/about_dialog.cpp
            src/qt_gui/about_dialog.h
            src/qt_gui/about_dialog.ui
@@ -739,8 +746,6 @@ set(QT_GUI src/qt_gui/about_dialog.cpp
            src/qt_gui/background_music_player.h
            src/qt_gui/cheats_patches.cpp
            src/qt_gui/cheats_patches.h
-           src/qt_gui/check_update.cpp
-           src/qt_gui/check_update.h
            src/qt_gui/main_window_ui.h
            src/qt_gui/main_window.cpp
            src/qt_gui/main_window.h
@@ -771,6 +776,7 @@ set(QT_GUI src/qt_gui/about_dialog.cpp
            ${EMULATOR}
            ${RESOURCE_FILES}
            ${TRANSLATIONS}
+           ${UPDATER}
 )
 endif()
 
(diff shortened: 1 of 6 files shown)


src/qt_gui/settings_dialog.cpp
```diff
@@ -153,6 +156,12 @@ SettingsDialog::SettingsDialog(std::span<const QString> physical_devices, QWidge
             auto checkUpdate = new CheckUpdate(true);
             checkUpdate->exec();
         });
+#else
+        // FIXME Can we do better than this?
```
jardon: i think it depends.

im of the opinion that we should likely disable the entire update management dialog unless we strategically rethink the update channels and how that is handled.  with it as is, for a package, it would likely load the diff from nightly, but idk if thats helpful since we cant guarantee that packages would track nightly at all

ngoquang2708: This is in the settings dialog though. I have disabled the whole update dialog and the check update from the menu.

ngoquang2708: My goal here is to somehow hide the update section in the settings, or make it look like be disabled.

ngoquang2708: Current code just make checkbox, combobox.. don't response to anything.

jardon: oh my mistake.  looks good :)

jardon:
i guess my question for this is would it be better to be opt-out on updates instead of opt-in?

ngoquang2708:
> i guess my question for this is would it be better to be opt-out on updates instead of opt-in?

You mean name the flag ```ENABLE_UPDATER``` or ```DISABLE_UPDATER```, right?

ngoquang2708:
I don't think anyone want to type more when build locally!
Only AppImages use the auto updater.

jardon:
does that not also impact windows and mac users?

ngoquang2708:
> does that not also impact windows and mac users?

The auto updater should only functional when files download from this repo. I already enable it in ```build.yaml```.

jardon:
i meant more of the "typing while building locally".

but either way is fine with me.

DanielSvoboda:
in the CMakeLists.txt file on line 34, you should add something like this to be able to select within vscode:
```cpp
option(ENABLE_UPDATER "Enables the options to updater" ON)
```
image

DanielSvoboda:
in the CMakeLists.txt file on line 34, you should add something like this to be able to select within vscode:
```cpp
option(ENABLE_UPDATER "Enables the options to updater" ON)
```

image

ngoquang2708:
@DanielSvoboda Do you think we should set it to ```OFF``` by default since the updater won't function correctly in local build? Of-course we can enable it to test.

DanielSvoboda:
> @DanielSvoboda Do you think we should set it to `OFF` by default since the updater won't function correctly in local build? Of-course we can enable it to test.

It works perfectly, when I compile it on my machine I can update it normally. What problem are you facing?

ngoquang2708:
Think about it this way. You compile it, then install it somewhere. Run it. Update it. You got the version from github not the version you compiled. Of course we dev want to test the update funtionality so se enable it to test.

DanielSvoboda:
The version that I compile from my machine I usually don't use AutoUpdater, because I update the repository and compile the new version, and I believe that few people use AutoUpdater in the local build. I use AutoUpdater with the version that I downloaded. But I think that this option can be ON, and if someone wants to change it, it is up to the person to change it.

ngoquang2708:
Yeah, that my point. We have to enable explicitly to use it because we know we gonna using it. And I already enable it in the ```build.yaml``` for this repo.

jardon:
I've actually changed my opinion on the opt-in vs opt-out nature of this.

The auto-updater assumes the build comes directly from GitHub as an app image/pkg/exe. and if shadps4 is distributed in any sort of way outside of builds from GitHub, the update system doesn't really apply or work.

Thats why I think it's actually better to opt in to the updater for the GitHub builds and that's it

CMakeLists.txt
```diff
@@ -843,8 +849,11 @@ else()
 endif()
 
 if (ENABLE_QT_GUI)
-   target_link_libraries(shadps4 PRIVATE Qt6::Widgets Qt6::Concurrent Qt6::Network Qt6::Multimedia)
-   add_definitions(-DENABLE_QT_GUI)
+    target_link_libraries(shadps4 PRIVATE Qt6::Widgets Qt6::Concurrent Qt6::Network Qt6::Multimedia)
```
jardon, line 852: This is nitpicky and probably unintentional, but did you mean to change the spacing here?

ngoquang2708: Yeah, some place use 2, some place use 4, this place use 3 :(

jardon: if only we had a ci/cd system that linted PRs :upside_down_face:

raphaelthegreat (approved):

PR #1491, opened 2024-11-06
Fix emulator name in desktop entry for appimages

ngoquang2708:
~~Warning: manual update of the AppImage is required for next update.~~

### DIFF
--- a/.github/linux-appimage-qt.sh
+++ b/.github/linux-appimage-qt.sh
@@ -30,4 +30,4 @@ cp -a "$GITHUB_WORKSPACE/build/translations" AppDir/usr/bin
 ./linuxdeploy-x86_64.AppImage --appdir AppDir -d "$GITHUB_WORKSPACE"/.github/shadps4.desktop  -e "$GITHUB_WORKSPACE"/build/shadps4 -i "$GITHUB_WORKSPACE"/.github/shadps4.png --plugin qt 
 rm AppDir/usr/plugins/multimedia/libgstreamermediaplugin.so
 ./linuxdeploy-x86_64.AppImage --appdir AppDir --output appimage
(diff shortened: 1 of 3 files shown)


.github/linux-appimage-qt.sh
```diff
@@ -30,4 +30,4 @@ cp -a "$GITHUB_WORKSPACE/build/translations" AppDir/usr/bin
 ./linuxdeploy-x86_64.AppImage --appdir AppDir -d "$GITHUB_WORKSPACE"/.github/shadps4.desktop  -e "$GITHUB_WORKSPACE"/build/shadps4 -i "$GITHUB_WORKSPACE"/.github/shadps4.png --plugin qt 
 rm AppDir/usr/plugins/multimedia/libgstreamermediaplugin.so
 ./linuxdeploy-x86_64.AppImage --appdir AppDir --output appimage
-mv Shadps4-x86_64.AppImage Shadps4-qt.AppImage
+mv shadPS4-x86_64.AppImage shadPS4-qt.AppImage
```
jardon: maybe we could not rename the appimage so that updates dont break?  i would assume most people arent executing the appimage directly so the name of the appimage file likely doesnt matter.

alternatively, since we're using bash, maybe we could adjust the update script to account for both cases?

ngoquang2708: I don't think breaking things in this early of development is that bad. At least we don't leave any teach debt.

jardon: pragmatically, i would agree.  but the solution is pretty straightfoward so i tend to wanna implement it

jardon (approved):
LGTM

raphaelthegreat (approved):
Looks good

Issue #1499, opened 2024-11-07
Feature request: skip GUI when launching game directly


fpiesche:
It'd be useful if the main Qt GUI executable didn't bother trying to read the entire game directory when it's launched directly into a game (by calling `shadps4 path/to/eboot.bin`). Some emulators just do this transparently, others (eg. RPCS3) have a command line flag to disable running the main GUI (in RPCS3's case, `--no-gui`).

This will make the emulator useful for configuring via the GUI when running it directly with no parameters, while also allowing the same executable to launch into a game directly without having to wait for the game list to build on startup first. This is a common use case with game launchers such as Steam's game mode or things like Launchbox and Playnite.

Ideally this should also launch in fullscreen mode, though doing this could also be handled by a separate command line parameter (that is, just running `shadps4 eboot.bin` will run the game in windowed mode, while running `shadps4 --fullscreen eboot.bin` will launch it in fullscreen).

DanielSvoboda:
There is the SDL version, which is different from the QT version which has an interface

fpiesche:
Yeah - thus in the current setup, a user would need both the SDL and the Qt version installed to be able to direct-launch games without waiting for the game list to populate first, but also have the GUI available for configuration. Making the Qt version skip the GUI, or at least the game list read, when launching an executable directly, would improve QOL for users of the Qt version.

I'll admit I haven't actually used the SDL version; does it meaningfully differ from the Qt version once it gets to actually running a game, e.g. in terms of Wayland support etc?

DanielSvoboda:
The QT version didn't even have a command, I remember that I was the one who added the ability to run the game through the command line in the QT version, just like in SDL, I did this just to be able to create [game shortcuts](https://github.com/shadps4-emu/shadPS4/pull/246) correctly. It's a good idea to have a command to run QT headless, similar to SDL.

Hermiten:
Can you try the last main version ?:)

DanielSvoboda:
> Can you try the last main version ?:)

It looks like the shortcut is broken...
then I do more tests

Nt0w:
I’m a newbie and had ai write the code.. hope it helps.. it’s for Linux and intel hope it help ✌️

#include <xf86drm.h>
#include <xf86drmMode.h>
#include <fcntl.h>
#include <unistd.h>
#include <cstring>
#include <iostream>
#include <cstdlib> // For strtol

// Function to parse CLI arguments
[...]

PR #1507, opened 2024-11-09
Skip GUI if launching a game directly

kalaposfos13:
Implements the suggestion in https://github.com/shadps4-emu/shadPS4/issues/1499


kalaposfos13:
Now instead of opening a game with shadps4 {path/to/eboot.bin}, you can now do this:
shadps4 --game CUSAXXXXX --show-gui --patch {path/to/patch/file}
Reading the source code or running shadps4 --help will give further info.
Opening game by ID is done by checking each game install folder, and seeing if there is a folder named CUSAXXXX in them, and if so, opens {installdir}/CUSAXXXXX/eboot.bin.

georgemoralis:
i wonder if that can somehow merge with #1349 and use boost program_options for flexability

kalaposfos13:
I'm guessing by "merging" you mean "rewriting my solution to also use Boost", right? Probably doable, from what I've seen on said PR, and it would be not that hard as that also uses maps, just differently, though I've never used that library, so I'll have to learn it.

georgemoralis:
it is not bad , but might be cleaner that way, i will pass it to @LNDF since he was the one than purpose boost :D

LNDF:
> it is not bad , but might be cleaner that way, i will pass it to @LNDF since he was the one than purpose boost :D

To be able to use boots, program_options needs to be added to ext-boots. As an alternativem, a header only more lightweight library can be used (such as tclap).

PR #1519, opened 2024-11-11
imgui: Dispatch SDL text input requests to main thread on macOS.

squidbus:
`SDL_StartTextInput` and `SDL_StopTextInput` utilize UI APIs that need to be executed from the main thread on macOS. SDL does not seem to handle dispatching itself, nor does it document that these should be executed from the main thread. To handle this without much disturbance, we can just dispatch these calls to the main thread on macOS.

I've also reported upstream to SDL to get either a fix or documentation clarification on this: https://github.com/libsdl-org/SDL/issues/11447

Fixes any game using IME input.


slouken:
FYI, you can now use SDL_RunOnMainThread() to call functions on the main thread in a cross-platform way.

PR #1520, opened 2024-11-11
ajm: handle single-frame decode jobs (+mp3 imrovements)

roamic:
* Break after a single frame if multi-frame decoding was not requested (fixes sound in Journey)
* Return `AJM_RESULT_NOT_ENOUGH_ROOM` if no frames fit the buffer
* mp3: implemented `GetCodecInfo`
* mp3: implemented FGH OFL header parser (used by Journey)
* Reworked gapless decoding to simplify the comprehension
* added error logs for outstanding cases
* minor code improvements


roamic:
Regressed some games, putting it in draft for now.

Issue #1362, opened 2024-10-13
FPS counter keybinding (F10) should be added to README


kalaposfos13:
As far as I know, it isn't mentioned anywhere, and it is a very useful thing to know
Also there is rdoc capture on F12 whatever that means

GHU7924:
I know:
F10 - FPS counter
Ctrl+F10
F11 - Fullscreen
there are more, but I don't remember them))

I think it's worth making a separate item in the "Help" section, which will display a list of available commands (which is similar to "About shadPS4"), only a window similar to something like this will be displayed:
(An image from the Internet for example)
![Screen-Shot-2014-05-16-at-7 22 04-AM-4109755284](https://github.com/user-attachments/assets/746f8154-443f-44e7-a65a-2ba4952c5e73)

It is also worth adding 2 buttons to the emulator itself:
1 - report a problem (you will be taken to the issues page in your browser - https://github.com/shadps4-emu/shadPS4/issues)
2 - report to discord (https://discord.com/invite/bFJxfftGW6)

Hermiten:
Good idea !
To follow

Hermiten:
@DanielSvoboda Hey, maybe for you ? :)

GHU7924:
I really hope that someone will take up the implementation.

At the moment, my personal vision is this (sketched in Paint):

Keyboards Shortcuts

![Sample 1](https://github.com/user-attachments/assets/f3a68de2-cb9c-476d-b79f-97c0711424a8)
![Sample 2](https://github.com/user-attachments/assets/e9fb0073-1c7a-443d-b82a-e63d40e1fcc9)

Send a report

![Report 1](https://github.com/user-attachments/assets/07467804-b2b3-4b1b-b281-346604934916)
![Report 2](https://github.com/user-attachments/assets/a1ddc455-fefc-4ed4-9423-e5e3f766976a)

If someone has their own ideas or options, then please suggest))

P.S. It's also worth changing the name "FPS counter keybinding (F10) should be added to README" to another one.

DanielSvoboda:
I put these buttons in the 'About' and I can make a separate screen as you mentioned for the keys and shortcuts

![ezgif-7-0f0a6ce22a](https://github.com/user-attachments/assets/a2945c6d-060b-47cf-943a-fef76eddc272)

GHU7924:
Your version looks very good too. And maybe that's what you should do with this window, but there should be general links there:

[...]

PR #1356, opened 2024-10-12
Added keyboard and mouse input remapping, mouse movement to joystick logic, GUI and more

kalaposfos13:
An updated overview of the PR as of 2 months later, since so much has changed that not a lot remains from the original version:

The current state of input handling
-
- Most controllers work out of the box, with Xbox controllers' back button emulating a touchpad press as well.
- However, these are the default bindings, and they are unchangeable.
- Keyboard inputs are technically supported, but only as single buttons, you can't rebind them and the configuration is basically unplayabe.
- You can't use the mouse at all.

I decided that since more and more games are finally becoming more and more playable, an update to this system is really needed, as a core part of PC gaming is the ability to rebind controls however you want, and a lot of people prefer using KBM over controller.

An overview of the changes
-
- I added the ability to rebind your mouse, keyboard and controller inputs, with up to 3 inputs per output.
- You can create a custom configuration for every game in your library.
- I added a GUI for this (accessed by the previously unused Controller button) where you can edit your bindings straight from the launcher. You can even reload the configuration mid-game, giving you the ability to change keybinds without restarting the emulator.
- I added an implementation of mouse-to-joystick emulation, making controlling the camera in games like Bloodborne with your mouse movement possible.
- A more detailed user manual can be found by clicking the Help button in the aforementioned config editor.

A technical overview of the control flow of the new input handling system
-
0. Upon launching a game, the config is read in from the correct file. Next, the main SDL event handling loop is started. A separate timer is also created for mouse polling, which runs on a completely different system than everything else.
1. The event handling loop waits for an event. If one is found, and it is a controller, keyboard or mouse (button or wheel) event, the main part of input handling starts.
2. Based on the event, a custom input ID is calculated (these are just the SDL ID-s but with an offset based on which device it is, as all of them start at 0). For analog inputs, the distance is also stored.
3. Next, we add or remove this ID based on the event type to the list of currently active inputs (for example, we add it if it's a keydown, but remove it if it's a keyup event). If that list changed (holding down a keyboard button will generate multiple keydown events, for example, so not every event changes the list), we recalculate the state of the emulated controller, otherwise, we finish the handling of this event here.
4. If the list changed, the following things happen: First, we reset the "new" state of the controller (from now on, this refers to the emulated one). It has two states: the actual and the "new", and the later is used by the algorithm, and when it finishes, changes the actual state to the newly calculated state.
[...]

PR #1522, opened 2024-11-11
Help - improvement

DanielSvoboda:
issue suggestions: https://github.com/shadps4-emu/shadPS4/issues/1362
- [X] Buttons in 'About' and clicking takes you to the respective links.

I removed the  submenu for 'Keys and Shortcuts' and left only the buttons for the external links of the sites.
I think it would be better to implement this in another PR so that it doesn't get stuck.

ezgif-5-c411d750db
ezgif-5-c20a1c4ad1


GHU7924:
1) I do not know what about YouTube (maybe it is not needed), but I know for sure that there are still links - https://shadps4.net  and https://x.com/shadps4
* Title: I'm not completely sure how "Keyboard shortcuts" or "Hotkeys" would be better.
** "Hotkey" - I drew the third line in the menu list (it was just more convenient for me), 
Sample 1
but in fact it should be the first or second, so it would be correct.

2) ShadPS4 has the ability to change the theme, 
Themes

so I think the color should be adjusted to the selected theme (I mean the link icons) .

3) 
>Maybe reduce the size of that logo on the left a little.

If all the links fit, then perhaps the logo on the left should not be touched.

4) I drew my version of Keyboard shortcuts on top of the "About shadPS4" window, so I don't know how to do it right, because new functions (keys) may still be added.

I think you can make a temporary option.
Make the window look like "About shadPS4" (or copy the old version), that is, you take the old version and just change the text.
I may have been a little hasty and pointed out some buttons that don't exist yet, because they haven't been merged with Main.
Something like this will do:
Interfeice456

Linear524:
Maybe F12 = Screenshot into dedicated screenshot folder inside ShadPS4 folder ?)))

DanielSvoboda:
> Maybe F12 = Screenshot into dedicated screenshot folder inside ShadPS4 folder ?)))

Screen capture has not been implemented yet, but if renderdoc is enabled and you press F12, it captures the information for debugging this system and saves it in the captures folder if I'm not mistaken. I'll continue this PR this weekend...

GHU7924:
https://github.com/shadps4-emu/shadPS4/pull/1356#issuecomment-2478201400 As soon as PR 1356 is merged with Main, we will have new functions.

F7 - Captures mouse and enables mouse movement input
F8 - Reparse keyboard input
F9 - Pause emulator (in Debug Menu)
F10 - FPS counter
LCtrl + F10 - Debug Menu
F11 - Fullscreen
F12 - Сaptures the information for debugging this system and saves it in the captures folder (with RenderDoc debugging enabled)
(I haven't checked and, to be honest, I don't know how to describe it more briefly)

@DanielSvoboda How many links are left in the end? 
GitHub, Discord, Ko-fi.com, YouTube???, X.com??, ShardPS4.net? 
If there are all 6, then you will have to reduce the PS4 icon and center the icons.

About links

If possible, it is worth implementing a screenshot function to make it easier to capture a moment or a graphical error, if one appears.

Move the F12 function to one of the F1-F6 keys, and make F12 a screenshot function (If there is a desire and an opportunity).

GHU7924:
Links with different themes look good.
Themes

I don't know about that, is it necessary here? I'm talking about the controller. 50/50
k

If you leave it, you need to improve it, and the background should be transparent in the image so that it looks uniform with the rest of it. But I think you wanted to do it anyway)) Example of changes:
k2

I think it's working out very well.

DanielSvoboda:
@GHU7924 I made some changes as you mentioned, now it's a little better. I also put the text inside a group.

image

kalaposfos13:
Just a quick note: This is neither my nor the current default keybinds, but I guess this would have to be rewritten anyway when my pull request gets merged

GHU7924:
@DanielSvoboda I think it looks amazing. Great job.

Edit:

With the implementation of the keyboard and mouse from @kalaposfos13 , it will be possible to do something like this:
k3

P.S. It's just a sketch, as I see it roughly, but it needs to be refined and brought to the same style.
P.S.S. L1 = LMB + LShift and R2 = RMB + LShift - in my opinion, it would be better, but as it is by default, so it is.

DanielSvoboda:
As this PR doesn't have much priority, I even left it as a draft because there are some things missing, I'll wait a bit for things to unfold and then I'll make the necessary changes later.

GHU7924:
You're right, the main thing here is not to rush. I just wrote (made a sketch) for an example, but I'm not saying that this is how it should be. I believe you, you will do well.

GHU7924:
Does it make sense to add system requirements?

Ab
( I didn't try very hard, but just quickly made a sketch, so the system requirements are identical )

Edit: (Additional information)
I suggested adding system requirements because I have seen such questions from users several times on some forums.
https://github.com/shadps4-emu/shadPS4/issues/1362#issuecomment-2469208491 Also, the second part of my comment can be implemented here. There will be general links and leads to the issues section. I think this is quite a good addition.

I also thought again about the implementation of the "Keys and shortcuts" window.
Here is a new view of the controller image, although when I did it, I entered my control settings for the keyboard, but I did not correct it anymore.

k4

But I didn't come up with anything about the "Share" button, because I don't know what is planned to be done in the emulator for this button and what functionality it will have, if at all.

I marked the PS button as (M = Menu), but I also don't know for its implementation, but let it be so for now.

DanielSvoboda:
I removed the keys shortcuts and left only the buttons for the external links of the sites.
I think it would be better to implement this in another PR so that it doesn't get stuck

GHU7924:
> I removed the keys shortcuts and left only the buttons for the external links of the sites. I think it would be better to implement this in another PR so that it doesn't get stuck

Yes, this makes sense, especially since we can mark the new PR as a continuation of this, and the ideas and improvements that will be here can be transferred to future PR. Also, a new PR can be started only after combining the keyboard PR.

EDIT: Future PR - Interface update.

PR #1525, opened 2024-11-13
Update Norwegian Bokmål translations

Martini-141:
Translation's for Bokmål, also renamed some stuff to make it even more clear its Bokmål and not Nynorsk translation.


Martini-141:
changed some wording I was unhappy with no further changes for this pr.

PR #1554, opened 2024-11-20
Don't read uninitialized memory in `sceSystemServiceReceiveEvent`

korenkonder:


polybiusproxy:
I believe it'll be better to just move the log after the nullptr check.

georgemoralis:
> I believe it'll be better to just move the log after the nullptr check.

agree

viniciuslrangel:
The `event` in this function is an out parameter, so we shouldn't read at all from that

polybiusproxy:
Oh, in that case...

PR #1557, opened 2024-11-20
Fixed false-positive image reuploads

roamic:
Should fix half of the rainbow artifacts in BloodBorne.
Big thanks to @raphaelthegreat and @psucien for helping debug and fix this.

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -635,7 +635,7 @@ bool BufferCache::SynchronizeBufferFromImage(Buffer& buffer, VAddr device_addr, 
                "Texel buffer aliases image subresources {:x} : {:x}", device_addr,
                image.info.guest_address);
     boost::container::small_vector<vk::BufferImageCopy, 8> copies;
-    u32 offset = buffer.Offset(image.cpu_addr);
+    u32 offset = buffer.Offset(image.info.guest_address);
     const u32 num_layers = image.info.resources.layers;
     const u32 max_offset = offset + size;
     for (u32 m = 0; m < image.info.resources.levels; m++) {
--- a/src/video_core/page_manager.cpp
+++ b/src/video_core/page_manager.cpp
@@ -115,7 +115,7 @@ struct PageManager::Impl {
             // Notify rasterizer about the fault.
             const VAddr addr = msg.arg.pagefault.address;
             const VAddr addr_page = Common::AlignDown(addr, PAGESIZE);
-            rasterizer->InvalidateMemory(addr_page, PAGESIZE);
+            rasterizer->InvalidateMemory(addr, addr_page, PAGESIZE);
         }
     }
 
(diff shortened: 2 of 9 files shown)


Missake212:
tried a full run in central yharnam, it is greatly reduced but I still got one (and it's way more purple than it used to be, also got it on one broken crate) great work !
Capture d'écran 2024-11-20 202650
Capture d'écran 2024-11-20 203833

AwesomeSoftware79:
Improved a lot, but still a couple of rainbow effects here and there (once in about 45 min.)-

image

GHU7924:
I made a circle in Central Yharnam, and during that time I had only one moment:


Edit: If something needs to be entered in the Log Filter, then write it, I will test it later if necessary.

TagoSMD:
NOPE
less frequent but still happen

roamic:
This fixes half of false-positive image re-uploads. The one we can easily detect. The other half is a lot more complicated to spot so will be out of scope of this PR.

Foul-Tarnished:
Will this get merged soon ?

I wonder if #1451 helps with the remaining buggy textures 🤔

roamic:
The latest commit should solve remaining issues in a non-hacky way.

rafael-57:
Thank you! Did some rounds of testing with Bloodborne in an heavy area that would produce rainbow corruptions after hitting enemies very fast, haven't seen any so far. If this doesn't regress other games I would merge soon 👍

Missake212:
Tested through Central Yharnam, killed every enemies like I did last time, not a single rainbow, will try a few more runs but usually I get at least 1, this is amazing work !

ZuwasieMan:
Played about ca. 45 minutes, I could not find any rainbow/metallic/grey textures. Looks very good.

ZuwasieMan:
After one hour I found one metallic/grey texture:

grafik

AxeKeeper:
Скриншот 24-11-2024 233426
Скриншот 24-11-2024 233440
Скриншот 24-11-2024 233523

Missake212:
Couldn't capture it since it was quick but while hitting an enemy he turned rainbow for a second until I hit him again, beside that didn't see any other rainbows

roamic:
> Couldn't capture it since it was quick but while hitting an enemy he turned rainbow for a second until I hit him again, beside that didn't see any other rainbows

I've seen this too, thought I fixed it with hash initialization. Seems not.

sam20112011:
> Скриншот 24-11-2024 233426 Скриншот 24-11-2024 233440 Скриншот 24-11-2024 233523

This has been going on for quite some time. Even before the blackness/rainbow fix and with mod from Nexus fix. It's just very rare and hard to pay attention to.

By the way, the rainbow effect on enemies, happens even in the version with the hack. Short. But still.
Perhaps it is really something to do with texture cache overflow, or 2gb buffer bug. Since it happens after a long gaming session.

Missake212:
Killed everyone in Central Yharnam like usual, not a single rainbow, this might be the one, I'll try a few more times to make sure but it's looking good !

PR #1565, opened 2024-11-21
Adding Optimizations to Compilation

Xphalnos:
I added some compilation optimizations.
I disabled the warnings "warning: argument unused during compilation: '/Zc:preprocessor' [-Wunused-command-line-argument]"
If this is not correct, maybe **@abouvier** can help me because he knows a lot about CMake ;)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -917,6 +917,9 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/src/imgui/renderer)
 add_dependencies(shadps4 ImGui_Resources)
 target_include_directories(shadps4 PRIVATE ${IMGUI_RESOURCES_INCLUDE})
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Qunused-arguments")
+target_compile_options(shadps4 PRIVATE $<$<CONFIG:Release>:-O3>)
+
 if (ENABLE_QT_GUI)
     set_target_properties(shadps4 PROPERTIES
 #       WIN32_EXECUTABLE ON


abouvier:
`-O3` is already [added by default](https://github.com/Kitware/CMake/blob/8a2e9f152a7c5c702f0ac080a7da28bd33b0af51/Modules/Compiler/GNU.cmake#L59) for `Release` build type. And compilation flags should be added with `target_compile_options`. Also `-Qunused-arguments` is a clang specific option and should be added conditionally.

Xphalnos:
> `-O3` is already [added by default](https://github.com/Kitware/CMake/blob/8a2e9f152a7c5c702f0ac080a7da28bd33b0af51/Modules/Compiler/GNU.cmake#L59) for `Release` build type. And compilation flags should be added with `target_compile_options`. Also `-Qunused-arguments` is a clang specific option and should be added conditionally.

Are you sure `-O3` is the default because the file you show me is `GNU.cmake` and I think on **Windows** it might not be that file (maybe I'm wrong). I noticed that without this flag, the application is **33184KB** and with it, **33170KB** (which shows the evidence of a change somewhere).

Foul-Tarnished:
> > `-O3` is already [added by default](https://github.com/Kitware/CMake/blob/8a2e9f152a7c5c702f0ac080a7da28bd33b0af51/Modules/Compiler/GNU.cmake#L59) for `Release` build type. And compilation flags should be added with `target_compile_options`. Also `-Qunused-arguments` is a clang specific option and should be added conditionally.
> 
> Are you sure `-O3` is the default because the file you show me is `GNU.cmake` and I think on **Windows** it might not be that file (maybe I'm wrong). I noticed that without this flag, the application is **33184KB** and with it, **33170KB** (which shows the evidence of a change somewhere).

GNU because it's for the GCC ; Gnu compiler collection

Idk what their github actions use, but you may add ``O3`` to msvc (and clang) to be sure windows builds gets it

PR #1574, opened 2024-11-22
savedatadialog_ui: Use fmt::localtime instead of formatting std::chrono::local_time

squidbus:
Formatting `std::chrono::local_time` is not compiling right on some Linux systems, probably to do with the standard library version check that fmt has for it.

Instead we can just convert to `time_t` and use `fmt::localtime`. This also means we can eliminate all the extra code using `std::chrono` time zones and the corresponding Mac compatibility code using the date library.

Tested in a few games and the save dialog times still look correct.


viniciuslrangel:
Oh, that's waaaay better

Issue #1566, opened 2024-11-21
ShadPS4 Game screen stretches to size of window rather than staying at 16:9


brad0demx:
Seems to have been introduced around commit 5054827 to main branch, not sure the exact commit but the ones from the day prior work correctly

![image](https://github.com/user-attachments/assets/a4c9ef69-89a8-4284-828f-d3a7d91b7d61)
![image](https://github.com/user-attachments/assets/8f1d2afc-58b4-4eed-836f-b1c3e252ebcd)

AwesomeSoftware79:
The same happens with CUSA03173.
The game stretches itself when in full screen. 
I think this is only noticeable when using a non-16:9 screen, i.e. 21:9.

psucien:
Please re-check on the latest main

brad0demx:
![image](https://github.com/user-attachments/assets/7dd520db-e292-4227-af2d-6d8a041fb6c7)
Issue is fixed in latest main, great work👏

willink87:
is it possible to get a stretch to window option? or can you confirm what i can modify on my end to make stretch to window originally described above?

PR #1583, opened 2024-11-23
video_core: restored presenter aspect calculations

psucien:
Fix for https://github.com/shadps4-emu/shadPS4/issues/1566


bigol83:
It is fixed 

Immagine 2024-11-23 112323

AwesomeSoftware79:
Yeah it is fixed.
Maybe could it be a good idea to add an "stretch to window" option? i.e. a checkbox in Settings/Emulator/"Stretch to window", and let user decide.
That is a typical option for many emulators. Not essential, but nice to have.

mRf1996:
> Yeah it is fixed. Maybe could it be a good idea to add an "stretch to window" option? i.e. a checkbox in Settings/Emulator/"Stretch to window", and let user decide. That is a typical option for many emulators. Not essential, but nice to have.

Second this! Would also appreciate if non-standard aspect ratios (i.e. 16:10, 21:9) gets implemented.

PR #1601, opened 2024-11-26
cmake: unbundle stb

abouvier:


src/core/libraries/libpng/pngdec.cpp
georgemoralis, line 1: actually why stb lib is needed here? we have replaced it with libpng

abouvier: I don't know, `stbi_load_from_memory` is called in both `src/core/file_format/splash.cpp` and `src/imgui/renderer/texture_manager.cpp`. You can convert these codes to use `libpng` if you want :p

abouvier: Never mind, it was a mistake.

PR #1604, opened 2024-11-27
[Libs] Ngs2

lzardy:
- Added all ngs2 related headers
- Reworked `ngs2_impl` to be more accurate, including structs (with byte comments, need confirmation)

I can remove the `.cpp` or `.h` that are not used or empty if needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,6 +315,24 @@ set(SYSTEM_LIBS src/core/libraries/system/commondialog.cpp
                 src/core/libraries/ngs2/ngs2_error.h
                 src/core/libraries/ngs2/ngs2_impl.cpp
                 src/core/libraries/ngs2/ngs2_impl.h
+                src/core/libraries/ngs2/ngs2_custom.cpp
+                src/core/libraries/ngs2/ngs2_custom.h
+                src/core/libraries/ngs2/ngs2_reverb.cpp
+                src/core/libraries/ngs2/ngs2_reverb.h
+                src/core/libraries/ngs2/ngs2_geom.cpp
+                src/core/libraries/ngs2/ngs2_geom.h
+                src/core/libraries/ngs2/ngs2_pan.cpp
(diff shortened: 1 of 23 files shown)


bigol83:
Is this supposed to work yet? I tried with Resident Evil and it crashes when calling the module 
[shad_log.txt](https://github.com/user-attachments/files/17952072/shad_log.txt)

`[Lib.SysModule] <Error> sysmodule.cpp:sceSysmoduleLoadModule:47: (DUMMY) called module = ORBIS_SYSMODULE_NGS2`

kalaposfos13:
First off, let me state that audio in GR2 is a bit broken, and needs a [skipped assert](https://github.com/shadps4-emu/shadPS4/blob/953fb1463850f071e5df2a25532c8e41eb198013/src/core/libraries/ajm/ajm_at9.cpp#L76) to get ingame, so it's possible that the problem comes from that, and not your PR.

Without the Ngs2 sysmodule, GR2 does output any sound, and with that on, the audio is way too loud and loopy. [Here's](https://www.youtube.com/watch?v=Tea-lcx2b5E) a video showcasing it.

As the game does not go ingame on the latest main yet, I merged this PR on my gr2-hacks branch to test it. It segfaults a few seconds into booting, with the following stack trace:
hostBufferInfo is NULL here:
https://github.com/shadps4-emu/shadPS4/blob/8d7d7815e2f4568bff0179ba9ff5ec2b2ad33b59/src/core/libraries/ngs2/ngs2_impl.cpp#L135
called from here:
https://github.com/shadps4-emu/shadPS4/blob/8d7d7815e2f4568bff0179ba9ff5ec2b2ad33b59/src/core/libraries/ngs2/ngs2.cpp#L192
called from the game code.

I'll attach the [log](https://github.com/user-attachments/files/18201412/shad_log.txt) as well, with the last few lines manually copied over from the terminal, as those lines got printed out to there, but not in the log.

georgemoralis:
that doesn't suppose to produce any sound , it is just a basic skeleton

PR #1606, opened 2024-11-27
libc_internal: Implement more functions.

squidbus:
Wanted to do some testing with HLE `libSceLibcInternal` to manipulate call results and encountered a bunch of missing functions. Adds a bunch of simple ones that were stubbed in my log output, mainly for string manipulation, math, and memory allocation.

With these I am able to get into the game I was testing in (CUSA04518) without LLE `libSceLibcInternal`.

### DIFF
--- a/src/core/libraries/libc_internal/libc_internal.cpp
+++ b/src/core/libraries/libc_internal/libc_internal.cpp
@@ -3,8 +3,8 @@
 
 #include <cmath>
 
+#include "common/assert.h"
 #include "common/logging/log.h"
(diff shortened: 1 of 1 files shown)


gandalfthewhite19890404:
Do I need libSceLibcInternal.sprx after that changes? Could a be a some kind of conflict if I kept that lib?

squidbus:
> Do I need libSceLibcInternal.sprx after that changes? Could a be a some kind of conflict if I kept that lib?

This helps if you don't have it but having it will probably still be better. No conflicts.

PR #1607, opened 2024-11-27
libraries: Add initial HLE JPEG encoder skeleton

squidbus:
Adds the initial skeleton of HLE `libSceJpegEnc`. Functions except encode are fully implemented, encode has some of the validity checks but is otherwise stubbed.

Tested in CUSA06093 with the PV screenshot functionality. Obviously no actual JPEG is encoded but the encoder is set up correctly and passes the validation checks.


src/core/libraries/jpeg/jpegenc.h
```diff
@@ -0,0 +1,59 @@
...
+
+struct OrbisJpegEncEncodeParam {
+    void* image;
+    void* jpeg;
+    u32 image_size;
+    u32 jpeg_size;
+    u32 image_width;
+    u32 image_height;
+    u32 image_pitch;
+    u16 unk1;
```
georgemoralis: u16 pixel_format;
	u16 encode_mode;
	u16 color_space;
	u8 sampling_type;
	u8 compression_ratio;
	s32 restart_interval;

squidbus: Updated, thanks for the info.

PR #1612, opened 2024-11-28
Define a data structure for MEM_SEMAPHORE PM4 command

mikusp:


ElBread3:
Ratchet and Clank requires MemSemaphore, and I discovered yesterday it boots upon skipping the compute shader pipeline and its unreachable (very hacky I know), but if this opcode is the only thing it needs to boot you may make a great many people happy if you finish MemSemaphore's implementation🙏(just letting you know)

polybiusproxy:
I think it'd be better to have a structure along with an implementation to make sure there aren't any mistakes.

PR #1615, opened 2024-11-29
semaphore: Use condvars with separate signaled flag to prevent races

squidbus:
* Revert back to using `std::conditional_variable` for waiting threads in semaphores.
* Add a flag `was_signaled` for the condvar to wait on, to know for sure if a thread was signaled once the lock is re-acquired.

Together these solve a potential race condition where `Wait` could time out during a `Signal` call. In this scenario the `Wait` call has determined whether a timeout has occurred from `try_acquire`, but still must wait to get the lock back from `Signal`. Meanwhile `Signal` could see the thread in the wait list and try to signal it, and the waiting thread will never know about it, will try to double-remove itself from the wait list, and will return a timeout error code.

We could solve this issue with flags and more careful lock handling using `std::binary_semaphore`, but its cleaner to do this with a condvar as its basically what they're meant for.

This fixes some semaphore-related instability in Hatsune Miku: Project DIVA X Update 1.02

### DIFF
--- a/src/core/libraries/kernel/threads/condvar.cpp
+++ b/src/core/libraries/kernel/threads/condvar.cpp
@@ -122,7 +122,7 @@ int PthreadCond::Wait(PthreadMutexT* mutex, const OrbisKernelTimespec* abstime, 
         SleepqUnlock(this);
 
         //_thr_cancel_enter2(curthread, 0);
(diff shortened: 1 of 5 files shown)


squidbus:
With everything here so far, seem to have stabilized 1.02 to the same extent as 1.00, but both still can crash during video playback, e.g. the opening movie.

PR #1621, opened 2024-11-29
Remove save migration code

Xphalnos:
Since a new major version is out, we can remove this piece of code used for data migration.
I had to add a lot of newlines because they are unevenly distributed.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -3,13 +3,14 @@
 
 #include <fstream>
 #include <string>
-#include <common/version.h>
 #include <fmt/core.h>
 #include <fmt/xchar.h> // for wstring support
 #include <toml.hpp>
-#include "common/logging/formatter.h"
+
 #include "common/path_util.h"
 #include "config.h"
+#include "logging/formatter.h"
+#include "version.h"
 
 namespace toml {
 template <typename TC, typename K>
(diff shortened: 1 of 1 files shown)


Xphalnos:
@ElBread3 It seems to me that you are the one who did the migration process. However, each time you start the application it asks for the games installation folder again.
Do you have any idea where this comes from?

ElBread3:
> @ElBread3 It seems to me that you are the one who did the migration process. However, each time you start the application it asks for the games installation folder again. Do you have any idea where this comes from?

No I do not, but this feature was just a small untested afterthought on the first PR before the code gotten all rewritten so its probably better gotten rid of yeah

src/common/config.cpp
```diff
@@ -545,18 +578,6 @@ void load(const std::filesystem::path& path) {
...
-        // TODO Migration code, after a major release this should be removed.
-        auto old_game_install_dir = toml::find_fs_path_or(gui, "installDir", {});
-        if (!old_game_install_dir.empty()) {
-            addGameInstallDir(std::filesystem::path{old_game_install_dir});
-        } else {
-            const auto install_dir_array =
-                toml::find_or<std::vector<std::string>>(gui, "installDirs", {});
-            for (const auto& dir : install_dir_array) {
-                addGameInstallDir(std::filesystem::path{dir});
-            }
```
squidbus, line 557: This part wasn't supposed to be removed and is why you're having an issue. This loads from the `installDirs` array in the config, without it there are no directories loaded.

PR #1622, opened 2024-11-30
sysmodule: Remove need for libSceRazorCpu

squidbus:
Uses of `libSceRazorCpu`, the primary current one being in `libSceFiber`, are gated behind checks that the sysmodule is loaded. Since it is only used for internal debugging and not useful to us as LLE, we can reduce the number of libraries that we need to load by always returning it as unloaded when queried.

Since I was modifying the sysmodule query functions, I also added the validity check for the sysmodule ID and related error codes.

Tested in CUSA17416 (Persona 5 Royal) and CUSA01130 (Gravity Rush: Remastered).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,6 +276,7 @@ set(SYSTEM_LIBS src/core/libraries/system/commondialog.cpp
                 src/core/libraries/save_data/dialog/savedatadialog_ui.h
                 src/core/libraries/system/sysmodule.cpp
                 src/core/libraries/system/sysmodule.h
+                src/core/libraries/system/system_error.h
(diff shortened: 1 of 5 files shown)


Dnonmi:
Is this PR related to the following?
"[Core] <Error> stubs.cpp:CommonStub:42: Stub: sceRazorCpuPlotValue (nid: njGikRrxkC0) called, returning zero to 0x90017ddbb"

CUSA33754

This is the log from this build:

[shad_log.txt](https://github.com/user-attachments/files/17964552/shad_log.txt)

squidbus:
Hmm I guess some versions of PSPHD use it to emit debug info? It shouldn't be an issue though, the stubs should suffice.

PR #1627, opened 2024-11-30
core/libraries: Misc. Ime fixes

polybiusproxy:
- Fixes crashes with Ime open.
- Do not return `ORBIS_IME_ERROR_CONNECTION_FAILED` on ImeKeyboard open—fixes softlocks on games that don't handle this error


georgemoralis:
fixed with #1655

PR #1632, opened 2024-11-30
ci: Limit build parallelism to number of processors.

squidbus:
Limits the build parallelism to the number of available processors (macOS CI already does this). May help prevent CI hangs.


ngoquang2708:
This should be tested with the cache be cleared.

squidbus:
> This should be tested with the cache be cleared.

After deleting all cache entries: https://github.com/squidbus/shadPS4/actions/runs/12098449916

PR #1635, opened 2024-11-30
libraries: Add stubs for libScePlayGoDialog

squidbus:
Adds basic stubs for libScePlayGoDialog. These stubs pretend that the dialog was opened and then finished successfully.

~~Allows CUSA16404 to progress further loading a save in PS4 Pro mode.~~ This was actually only needed because a recent commit broke PlayGo functionality, but this may be useful to have anyway.


isshininu:
PlayGo commits fixed usual error in CUSA00076 (The Order: 1886), now game crashed trying to call ORBIS_SYSMODULE_PLAYGO
[shad_log.zip](https://github.com/user-attachments/files/17968857/shad_log.zip)

PR #1636, opened 2024-12-01
 ci: Use link-time optimization for building

MajorP93:
This PR enables LTO if the respective build toolchain supports it.

During my tests LTO enabled builds of shadPS4 performed noticeably better in certain games (e.g. Bloodborne).
Especially 1% low FPS measurement results were better and perceived stutter was less.

This observation was backed by other testers that were given a test build by me.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ jobs:
         arch: amd64
 
     - name: Configure CMake
-      run: cmake --fresh -G Ninja -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+      run: cmake --fresh -G Ninja -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE=ON -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
 
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --parallel $env:NUMBER_OF_PROCESSORS
@@ -143,7 +143,7 @@ jobs:
         arch: amd64
 
     - name: Configure CMake
-      run: cmake --fresh -G Ninja -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DENABLE_QT_GUI=ON -DENABLE_UPDATER=ON -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+      run: cmake --fresh -G Ninja -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DENABLE_QT_GUI=ON -DENABLE_UPDATER=ON -DCMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE=ON -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
 
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --parallel $env:NUMBER_OF_PROCESSORS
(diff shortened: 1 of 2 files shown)


baggins183:
This shouldnt apply to debug builds

baggins183:
IMO it would be good to wrap this in some variable check so its off by default for anyone building locally but can be enabled by -Dwhatever for github builds

abouvier:
This is a user preference, you should just pass `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` to the cmake call if you want (if you compiler supports it).

ngoquang2708:
So this config it better set in GitHub CI instead?

abouvier:
> So this config it better set in GitHub CI instead?

Yes, like every other build preferences (compiler, optimization level, etc.), or in a `CMakePresets.json` file.

ngoquang2708:
Missing ``CMAKE_`` prefix.

MajorP93:
> Missing `CMAKE_` prefix.

Thanks for the thint.

Updated the PR: following your recommendations I enabled LTO in the GH action instead of CMakeLists.txt in order to leave the option of building with having it disabled for people compiling the project locally.

abouvier:
You can put `set(CMAKE_POLICY_DEFAULT_CMP0069 NEW)` at the top of `externals/CMakeLists.txt` to configure third-party projects. It won't do anything when LTO is disabled.

MajorP93:
@abouvier thanks for the hint.
I updated the PR accordingly.
I used `cmake_policy(SET CMP0069 NEW)` instead of `set(CMAKE_POLICY_DEFAULT_CMP0069 NEW)` as this seems to be the recommended way of setting the policy in a `CMakeLists.tx`t file.

Compiled locally, behaves as expected.

abouvier:
`cmake_policy` is for setting the policies of your own project, not the externals ones. The policy CMP0069 is already set to NEW for shadPS4 because of the `cmake_minimum_required(VERSION 3.24)` at the very beginning.

See https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_DEFAULT_CMPNNNN.html

MajorP93:
> `cmake_policy` is for setting the policies of your own project, not the externals ones. The policy CMP0069 is already set to NEW for shadPS4 because of the `cmake_minimum_required(VERSION 3.24)` at the very beginning.
> 
> See https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_DEFAULT_CMPNNNN.html

Can you give me a reference to documentation where they say cmake_policy is meant to be used for own projects?

Without the line I added / without setting CMP0069 at all I get warnings like this for externals:

```
CMake Warning (dev) at externals/discord-rpc/src/CMakeLists.txt:31 (add_library):
  Policy CMP0069 is not set: INTERPROCEDURAL_OPTIMIZATION is enforced when
  enabled.  Run "cmake --help-policy CMP0069" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.
```

The output mentions `cmake_policy`.
With the line I added / after using cmake_policy, these warnings are not being printed anymore so it behaves as expected.

MajorP93:
> `cmake_policy` is for setting the policies of your own project, not the externals ones. The policy CMP0069 is already set to NEW for shadPS4 because of the `cmake_minimum_required(VERSION 3.24)` at the very beginning.
> 
> See https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_DEFAULT_CMPNNNN.html

I read some documentation and was able to verify that CMAKE_POLICY_DEFAULT_CMPNNNN makes more sense here.

Updated accordingly. Thanks.

abouvier (approved):

baggins183:
should this use -DCMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE instead?
https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION_CONFIG.html

dont think there are non-release builds on this repo's actions but maybe good idea

MajorP93:
> should this use -DCMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE instead? https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION_CONFIG.html
> 
> dont think there are non-release builds on this repo's actions but maybe good idea

I also think that the action will only be used for building "release" config but it does not hurt to only enable LTO for specifically that config.
On debug builds compiler optimization is not wanted / needed.

Thanks for the hint, I updated the PR.

MajorP93:
Anything else that needs to be addressed?
I did not encounter any issues during my tests which is why I think this PR is ready to get merged.
If there is anything that should be addressed or tested I am happy to do so.

ZuwasieMan:
@georgemoralis may I ask about the status of this PR? Does it need further testing? Can be merged into main?
I tested a build with these compiler optimizations, and it seems like bb's fps are more stable compared to main, with less stuttering and smoother fps dips.
I think the main discussion now is to apply those optimizations only for release builds?

MajorP93:
> @georgemoralis may I ask about the status of this PR? Does it need further testing? Can be merged into main? I tested a build with these compiler optimizations, and it seems like bb's fps are more stable compared to main, with less stuttering and smoother fps dips. I think the main discussion now is to apply those optimizations only for release builds?

This PR in its current form will only enable LTO for release builds (as per cmake configuration).
This makes sense as cmake debug configurations also disable other optimization flags (e.g. -O3), so in order to ensure consistency (no optimization needed for debug builds) LTO should also be disabled for these configurations.

shadPS4's CI is currently always building release configuration. LTO would always be included for CI builds as of the current state of this PR. This should be fine as CMake flag "INTERPROCEDURAL_OPTIMIZATION" will opt for Thin-LTO and never use Fat/Full-LTO. Due to this the build time does only increase a little.

PR #1637, opened 2024-12-01
More devtools stuff

viniciuslrangel:
- Add memory map window
- Add loaded shader inspector
- Fix user data display bug
- Fix miss-highlight

image
image


georgemoralis:
awesome!

PR #1640, opened 2024-12-01
video_core: Add multipler to handle special cases of texture buffer stride mismatch

raphaelthegreat:


Xcedf:
TLG texel stride assert is gone, but graphics went flickering, Infamous still hits the assert

AboMedoz:
> TLG texel stride assert is gone, but graphics went flickering, Infamous still hits the assert

True

AboMedoz:
### InFAMOUS First Light - CUSA00897

```
[Debug] <Critical> vk_rasterizer.cpp:operator():511: Assertion Failed!
Texel buffer stride must match format stride
```
image

ngoquang2708:
userfaultfd cause BB to hang when load ingame for a couple of seconds:  [hang.txt](https://github.com/user-attachments/files/17973771/hang.txt).
Disable userfaultfd make BB work again with this PR: [work.txt](https://github.com/user-attachments/files/17973778/work.txt).

Xcedf:
Being combined with latest fixes PR feels quite well:
TLG has no flickering anymore and Infamous seems improved, have assert commented out in my build, will test later if it still hits
before

after
изображение_2024-12-06_141719880

PR #1643, opened 2024-12-01
shader: Specialize on vertex input number types if needed.

squidbus:
According to Vulkan spec the vertex inputs should have number type classes (float, uint, int) matching those in the shader. This adds specialization to match that.

For drivers with `VK_EXT_legacy_vertex_attributes` support, this can be skipped, as it allows mismatched types:
* Windows: NVIDIA with Proprietary 552.61.0.0+
* Linux: AMD/NVIDIA/Intel with Mesa 24.2.0+, NVIDIA with Proprietary 550.40.63.0+

Fixes validation errors like:
```
VUID-vkCmdDrawIndexed-Input-08734: Validation Error: [ VUID-vkCmdDrawIndexed-Input-08734 ] Object 0: handle = 0xc1fcef000000854c, name = vs_0xee67e28b10b611ad_0, type = VK_OBJECT_TYPE_SHADER_MODULE; | MessageID = 0x1c98f0b9 | vkCmdDrawIndexed():  vkCmdSetVertexInputEXT set pVertexAttributeDescriptions[5] (binding 5, location 5) with format VK_FORMAT_R16G16B16A16_SINT but the vertex shader input is numeric type vec4 of float32.
```

### DIFF
--- a/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
+++ b/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
@@ -155,18 +155,12 @@ void EmitContext::DefineInterfaces() {
 }
 
 const VectorIds& GetAttributeType(EmitContext& ctx, AmdGpu::NumberFormat fmt) {
-    switch (fmt) {
-    case AmdGpu::NumberFormat::Float:
-    case AmdGpu::NumberFormat::Unorm:
-    case AmdGpu::NumberFormat::Snorm:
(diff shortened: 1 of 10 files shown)


squidbus:
Draft as the path without `VK_EXT_legacy_vertex_attributes` still needs some work.

squidbus:
Superceded (at least for now) by https://github.com/shadps4-emu/shadPS4/pull/1656 as this may not be needed after all.

PR #1644, opened 2024-12-01
shader_recompiler: Implement manual barycentric interpolation path

raphaelthegreat:
Credits to @vladmikhalin for initially investigating this issue.

In some areas in Bloodborne some instanced objects can have noticeable flicker, specifically on Nvidia GPUs, when moving in relation to them. This is hard to show in plain screenshots so I've recorded a video.

https://github.com/user-attachments/assets/be057dc6-013b-4fc7-98a2-588710c83ddd

This is because the shader used for rendering those objects passes the gl_InstanceIndex to the fragment shader using varying float attributes instead of a flat integer attribute. Nvidia has a different interpolation method compared to AMD hw, so 3 identical floats aren't interpolated to exactly the same value, which causes the instance index to get truncated to zero for some pixels and to one for others. This problem is also noted on other emulators like [Xenia](https://github.com/xenia-project/xenia/issues/2012) and [RPCS3](https://github.com/RPCS3/rpcs3/issues/13755). To fix this we use the VK_KHR_fragment_shader_barycentric extension to manually interpolate the attributes according to the 2 step AMD formula (credit to [Triang3l](https://github.com/Triang3) for providing the appropriate formula in their issue). This is only applied on Nvidia drivers that support the extension, which means Turing and newer GPUs. If your GPU is older than that you are probably out of luck.

https://github.com/user-attachments/assets/f7259e86-5691-457f-ab4b-2ee7764ed4d1

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv.cpp
@@ -206,7 +206,7 @@ Id DefineMain(EmitContext& ctx, const IR::Program& program) {
     return main;
 }
 
-void SetupCapabilities(const Info& info, EmitContext& ctx) {
+void SetupCapabilities(const Info& info, const Profile& profile, EmitContext& ctx) {
     ctx.AddCapability(spv::Capability::Image1D);
     ctx.AddCapability(spv::Capability::Sampled1D);
     ctx.AddCapability(spv::Capability::ImageQuery);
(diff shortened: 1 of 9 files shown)


ZuwasieMan:
This flickering seems to happen with some **interactable** objects, like breakable objects (coffins in Yharnam, jars in Oedon Chapel) or even doors. This PR fixes those problems.

GHU7924:
> This flickering seems to happen with some **interactable** objects, like breakable objects (coffins in Yharnam, jars in Oedon Chapel) or even doors. This PR fixes those problems.

Yes, it is. I tested build 30c469b.

PR #1576, opened 2024-11-22
cmake: unbundle libpng

abouvier:
`zlib-ng` is not used anymore (apart from building `zlib`) and could be replaced by `zlib` repo directly.

PR #1645, opened 2024-12-02
cmake: fix build

abouvier:
This should really fix https://github.com/shadps4-emu/shadPS4/pull/1576#issuecomment-2510013059 for good.


Foul-Tarnished:
Still getting this...
@abouvier 
image

Just did a full git clone --recursive
then opened in VS 2022

The last error line appeared when I added Qt through CMake argument list, and is probably unrelated

PR #1650, opened 2024-12-02
Fix SDL version cannot launch game using game ID

ngoquang2708:
Missing from https://github.com/shadps4-emu/shadPS4/pull/1507.

Summary from @kalaposfos13:
  * config.toml now gets autogenerated when you run the sdl version for the first time
  
  * Added an --add-game-folder argument so you can have a CLI interface to do so, instead of having to manually edit the file
  
  * Added  a warning if you try to start a game without setting at least one game directory
  
  * Added the option to treat the last argument as the game input, if the previous arguments didn't specify it yet, because it was already in the SDL version, but not in the Qt one
  
  * Renamed gamePath to game_path to follow the project style
  
  * Updated the Help text with the new info

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4,12 +4,14 @@
 #include "functional"
 #include "iostream"
 #include "string"
+#include "system_error"
 #include "unordered_map"
 
 #include <fmt/core.h>
 #include "common/config.h"
 #include "common/memory_patcher.h"
 #include "common/path_util.h"
+#include "core/file_sys/fs.h"
 #include "emulator.h"
 
 #ifdef _WIN32
@@ -38,6 +40,7 @@ int main(int argc, char* argv[]) {
                           "  -p, --patch <patch_file>      Apply specified patch file\n"
                           "  -f, --fullscreen <true|false> Specify window initial fullscreen "
                           "state. Does not overwrite the config file.\n"
+                          "  --add-game-folder <folder>    Adds a new game folder to the config.\n"
                           "  -h, --help                    Display this help message\n";
              exit(0);
          }},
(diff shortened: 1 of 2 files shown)


kalaposfos13:
For this, you'd have to manually write the game directory path to the config file, right? Otherwise it'll still do nothing, as it won't have any directories to search, and while I never used the SDL version, there probably isn't a popup at the first startup to set it.

ngoquang2708:
This require preconfigured config of course.

kalaposfos13:
I opened a pull request on your fork to add some things to this:
- config.toml now gets autogenerated when you run the sdl version for the first time
- Added an --add-game-folder argument so you can have a CLI interface to do so, instead of having to manually edit the file
- Added  a warning if you try to start a game without setting at least one game directory
- Added the option to treat the last argument as the game input, if the previous arguments didn't specify it yet, because it was already in the SDL version, but not in the Qt one
- Renamed gamePath to game_path to follow the project style
- Updated the Help text with the new info
Now that I look at all this, maybe I should just open a PR to the main project, but I don't want to just invalidate your PR just because I copied and improved upon it.

kalaposfos13:
Can this be merged? I don't think this has any issues that prevent it from happening

Hermiten:
Have you tried it successfully? I think so, but it's always worth clarifying before merging.

ngoquang2708:
Rebase.

ngoquang2708:
Tested with:

* Qt (with ``ENABLE_QT_GUI=ON`` set):
```
shadps4 --game ID
shadps4 -g ID
shadps4 ID
```

* SDL (with ``ENABLE_QT_GUI=OFF`` set):
```
shadps4 --game ID
shadps4 -g ID
shadps4 ID

```

All work without any problems.

PR #1653, opened 2024-12-03
shader_recompiler: Read image format info directly from sharps instead of storing in shader info.

squidbus:
Bit of cleanup removing some of the format data stored in `Shader::Info` structures in favor of always reading it from the sharps.


squidbus:
Superceded by https://github.com/shadps4-emu/shadPS4/pull/1656

PR #1656, opened 2024-12-03
renderer_vulkan: Parse fetch shader per-pipeline

squidbus:
Fetch shader used by a vertex shader may change, so instead of parsing its data out to `Shader::Info` where it remains constant for all permutations, save just the fetch shader sharp location, and load it again for each permutation generated.

Along with this, some other pieces of `Shader::Info` data that save a fixed format were instead changed to load format data from the underlying sharps.

This fixes validation errors in games like CUSA26789 which were also crashes on some drivers like MoltenVK. This may also improve graphics in some games where the missing change in fetch shader had a larger impact.


squidbus:
Considering this complete now minus review.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -187,13 +187,23 @@ void Rasterizer::Draw(bool is_indexed, u32 index_offset) {
...
-    buffer_cache.BindVertexBuffers(vs_info);
+    const auto& fetch_shader = pipeline->GetFetchShader();
+    buffer_cache.BindVertexBuffers(vs_info, fetch_shader);
     const u32 num_indices = buffer_cache.BindIndexBuffer(is_indexed, index_offset);
 
     BeginRendering(*pipeline, state);
     UpdateDynamicState(*pipeline);
 
-    const auto [vertex_offset, instance_offset] = vs_info.GetDrawOffsets(regs);
+    u32 vertex_offset = regs.index_offset;
```
raphaelthegreat: Can this be made into a function member of fetch_shader but receive a std::span of user data, probably a bit cleaner

squidbus: Done.

raphaelthegreat: Passing both regs and info makes fetch_shader.h need to include info header which is quite heavy, thats why I proposed it should receive a std::span + an integer of index_offset

raphaelthegreat: But not too big of a deal however

raphaelthegreat: Oh I see you need to already include it for ReadUdReg, so yeah its fine

squidbus: It already includes info to be able to have a `GetSharp` function

PR #1661, opened 2024-12-04
renderer_vulkan: Add support for indexed QuadList draw.

squidbus:
Currently in the case of an indexed QuadList draw, the support code just ignores the index buffer and creates a new one with default vertex order. Instead, convert the existing index buffer to the new vertex order. Needed to move the index buffer generation/conversion to header functions to simplify templating.

Fixes vertex explosions from some effects in CUSA17416 - Persona 5 Royal, for example the pink splashes when walking.

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -236,7 +236,7 @@ bool BufferCache::BindVertexBuffers(const Shader::Info& vs_info) {
 u32 BufferCache::BindIndexBuffer(bool& is_indexed, u32 index_offset) {
     // Emulate QuadList primitive type with CPU made index buffer.
     const auto& regs = liverpool->regs;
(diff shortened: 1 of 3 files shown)


squidbus:
Seems to fix some visual effects in other GFD Engine games as well.

rafael-57:
image

Sadly this doesn't address vertex explosions in BB :P

Crispy81:
Screenshot 2024-12-04 210559

Or on Persona 5 menu. Everything else looks fine. May not even be related to this.

squidbus:
> Or on Persona 5 menu. Everything else looks fine. May not even be related to this.

Did this happen before this PR?

Crispy81:
Yes, it did. I tested this about a week ago and noticed it. And just tested again with this PR.

squidbus:
Okay, I know Royal also still has some other visual issues so its probably something else.

PR #1664, opened 2024-12-04
thread: Configure stack and guard on POSIX hosts.

squidbus:
Applies the stack and guard configurations from the guest `PthreadAttr` to the host `pthread_attr_t` on POSIX systems.

Fixes a crash on launch in CUSA08010 when the default host stack size is too small.

### DIFF
--- a/src/core/thread.cpp
+++ b/src/core/thread.cpp
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include "libraries/kernel/threads/pthread.h"
 #include "thread.h"
 
 #ifdef _WIN64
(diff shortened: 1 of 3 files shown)


src/core/thread.cpp
```diff
@@ -15,14 +16,16 @@ Thread::Thread() : native_handle{0} {}
...
+int Thread::Create(ThreadFunc func, void* arg, const ::Libraries::Kernel::PthreadAttr* attr) {
 #ifdef _WIN64
     native_handle = CreateThread(nullptr, 0, (LPTHREAD_START_ROUTINE)func, arg, 0, nullptr);
     return native_handle ? 0 : -1;
 #else
     pthread_t* pthr = reinterpret_cast<pthread_t*>(&native_handle);
     pthread_attr_t pattr;
     pthread_attr_init(&pattr);
+    pthread_attr_setstack(&pattr, attr->stackaddr_attr, attr->stacksize_attr);
+    pthread_attr_setguardsize(&pattr, attr->guardsize_attr);
```
raphaelthegreat: According to manual guardsize is ignored when setting user allocated stacks so this is probably not needed

squidbus: Ah right because it only matters for the one allocating the stack, okay removed.

PR #1665, opened 2024-12-04
Support for Vulkan 1.4

Xphalnos:
Added support for Vulkan 1.4

Interesting article: https://www.khronos.org/news/press/khronos-streamlines-development-and-deployment-of-gpu-accelerated-applications-with-vulkan-1.4


raphaelthegreat:
Unless we actually use vulkan 1.4, I don't think updating find_package is a good idea, as it will prevent some linux distros that are not cutting edge from using system headers. @abouvier probably knows more on this subject

gandalfthewhite19890404:
Hardware requirements gonna be raised? What GPU now we should have?

raphaelthegreat:
This doesn't raise minimum vulkan version emulator requires (that will still be 1.2), but rather updates library headers to _allow_ emulator to use vulkan 1.4 in the future. In summary this is mostly on cosmetic side so not much to worry about

externals/CMakeLists.txt
```diff
@@ -123,6 +123,7 @@ endif()
 # MagicEnum
 if (NOT TARGET magic_enum::magic_enum)
     add_subdirectory(magic_enum)
+    target_include_directories(magic_enum INTERFACE magic_enum/include/magic_enum)
```
abouvier: Headers are now installed in `/usr/include/magic_enum/magic_enum.hpp`. You should edit all *.cpp files, replacing `#include <magic_enum.hpp>` with `#include <magic_enum/magic_enum.hpp>` instead. Using system `magic_enum` lib will fail otherwise.

abouvier:
> Unless we actually use vulkan 1.4, I don't think updating find_package is a good idea, as it will prevent some linux distros that are not cutting edge from using system headers. @abouvier probably knows more on this subject

It's fine, old distros will fallback on external headers. Otherwise, without the source code change, it's the rolling distros that will not be able to compile with the latest system headers. Newer distros should be privileged :)

Xphalnos:
I have made the requested changes!

Xphalnos:
@raphaelthegreat Any news? The changes that were requested have been applied and approved.

raphaelthegreat:
Yes it looks okay now, thanks for the upgrade

abouvier:
Oops, you should have raised the version constraint of `magic_enum` too: `find_package(magic_enum 0.9.7 CONFIG)`.

PR #1671, opened 2024-12-05
The way to Unity, pt.2

polybiusproxy:
A big roadblock for Unity on Windows is the use of exceptions, and to implement those, we need to do some preparations.
The idea is to use APCs—for this, we must switch our code to use alertable waits with Win32 wrappers on:
- [x] Pthread condition waits
- [x] Pthread mutexes
- [x] Orbis and POSIX semaphores
- [x] Thread sleeps
- [ ] Event flags (was not seen on Unity, thus not a priority)

This will allow us to properly implement exceptions later on.

### DIFF
--- a/src/core/libraries/kernel/sync/semaphore.h
+++ b/src/core/libraries/kernel/sync/semaphore.h
@@ -1,0 +1,117 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include <atomic>
+#include <chrono>
+
+#include "common/assert.h"
+#include "common/types.h"
+
+#ifdef _WIN64
+#include <windows.h>
+#else
+#include <semaphore>
+#endif
+
+namespace Libraries::Kernel {
+
+template <s64 max>
+class Semaphore {
+public:
+    Semaphore(s32 initialCount)
+#ifndef _WIN64
+        : sem{initialCount}
+#endif
+    {
+#ifdef _WIN64
+        sem = CreateSemaphore(nullptr, initialCount, max, nullptr);
+        ASSERT(sem);
+#endif
+    }
+
+    ~Semaphore() {
+#ifdef _WIN64
+        CloseHandle(sem);
+#endif
+    }
+
+    void release() {
+#ifdef _WIN64
+        ReleaseSemaphore(sem, 1, nullptr);
+#else
+        sem.release();
+#endif
+    }
+
+    void acquire() {
+#ifdef _WIN64
+        for (;;) {
+            u64 res = WaitForSingleObjectEx(sem, INFINITE, true);
+            if (res == WAIT_OBJECT_0) {
+                return;
+            }
+        }
+#else
+        sem.acquire();
+#endif
+    }
+
+    bool try_acquire() {
+#ifdef _WIN64
+        return WaitForSingleObjectEx(sem, 0, true) == WAIT_OBJECT_0;
+#else
+        return sem.try_acquire();
+#endif
+    }
+
+    template <class Rep, class Period>
+    bool try_acquire_for(const std::chrono::duration<Rep, Period>& rel_time) {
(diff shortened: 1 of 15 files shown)


src/core/libraries/kernel/sync/semaphore.cpp
```diff
@@ -0,0 +1,113 @@
...
+        return try_acquire();
+    }
+
+    return WaitForSingleObjectEx(sem, timeout_ms, true) == WAIT_OBJECT_0;
+#else
+    return sem.try_acquire_for(timeout);
+#endif
+}
+
+CountingSemaphore::CountingSemaphore(s32 initialCount, s32 maxCount) {
```
squidbus: This seems redundant, why can't we just have a single semaphore implementation and use initial/max count = 1 for binary semaphore?

polybiusproxy: They use `binary_semaphore` and `counting_semaphore` from the STL on other platforms and their implementation might be different (it is on MSVC).

squidbus: They should still behave equivalent though, on other platforms `binary_semaphore` is just a typedef for `counting_semaphore<1>`

raphaelthegreat: At least on msvc stl binary_semaphore isnt a typedef but has actual special implementation iirc

squidbus: On MSVC we're using Win32 semaphores explicitly here

StevenMiller123:
This PR noticably regresses performance in UE4 titles. This is especially noticable in Dragon Ball Z Kakarot. A severe performance regression is also noticable in updated versions of Hatsune Miku Project Diva X.

src/core/libraries/kernel/sync/semaphore.h
```diff
@@ -0,0 +1,85 @@
...
+
+    bool try_acquire();
+
+    template <class Clock, class Duration>
+    bool try_acquire_until(const std::chrono::time_point<Clock, Duration>& abs_time) {
+#ifdef _WIN64
+        auto duration = abs_time.time_since_epoch();
+        auto milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(duration).count();
+
+        u64 res = WaitForSingleObjectEx(sem, static_cast<u64>(milliseconds), true);
```
squidbus: `BinarySemaphore` is passing a duration timeout but here you are just passing the plain epoch timestamp. Shouldn't this be subtracted from `now()`?

ElBread3:
And Touhou Luna Nights takes my CPU again

gandalfthewhite19890404:
You broke Flow.
Main:
image

PR:
image

polybiusproxy:
> This PR noticably regresses performance in UE4 titles. This is especially noticable in Dragon Ball Z Kakarot. A severe performance regression is also noticable in updated versions of Hatsune Miku Project Diva X.

> And Touhou Luna Nights takes my CPU again

This might be due to the fact that we're using WaitForSingleObjectEx now (just as we did previously with winpthreads), which uses quite the whole thread on infinite waits. This was removed with the pthread rewrite where we moved to STL semaphores which use atomics and WaitOnAddress and WakeByAddressAll/WakeByAddressSingle on Windows.

Unfortunately, I don't see a way of resolving this as, to my knowledge, there is no other documented method of having alertable waits.

StevenMiller123:
My reported performance regressions are solved after the latest pushes. Both Diva X v1.02 and Dragon Ball Z Kakarot are performing as they do on current main builds now.

GHU7924:
I found a couple of Unity games for the test, but they failed at startup.

Build e1ecfb8

CUSA05297 INSIDE

>[Debug] <Critical> exception.cpp:sceKernelInstallExceptionHandler:77: Unreachable code!
>Missing exception implementation

CUSA23081 Untitled Goose Game

>[Debug] <Critical> exception.cpp:sceKernelInstallExceptionHandler:77: Unreachable code!
>Missing exception implementation

PR #1674, opened 2024-12-06
shader_recompiler: Implement V_LSHL_B64 for immediate arguments.

squidbus:
Implements `V_LSHL_B64` when both arguments are immediates. Fixes an assert in CUSA00221.


raphaelthegreat (approved):
Thanks LGTM

PR #1689, opened 2024-12-07
Fix + documentation update

Xphalnos:
As reported in this comment https://github.com/shadps4-emu/shadPS4/pull/1665#issuecomment-2525261521, I have updated the version number of magic_enum.
I have also updated the build instructions for Windows.


raphaelthegreat:
Nice

PR #1692, opened 2024-12-08
memory: Handle 0 alignment in MemoryManager::Allocate

squidbus:
Similar to other memory functions, handle alignment of 0x0 to `MemoryManager::Allocate` with 16KB default, to satisfy memory mapping requirements.

Fixes crash on launch in CUSA28193 (SONIC FRONTIERS). Had some testing done and doesn't appear to be any regressions.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -96,12 +96,12 @@ PAddr MemoryManager::PoolExpand(PAddr search_start, PAddr search_end, size_t siz
 PAddr MemoryManager::Allocate(PAddr search_start, PAddr search_end, size_t size, u64 alignment,
                               int memory_type) {
     std::scoped_lock lk{mutex};
+    alignment = alignment > 0 ? alignment : 16_KB;
 
     auto dmem_area = FindDmemArea(search_start);
 
     const auto is_suitable = [&] {
-        const auto aligned_base = alignment > 0 ? Common::AlignUp(dmem_area->second.base, alignment)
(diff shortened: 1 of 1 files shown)


kvicken:
CUSA28194 - SONIC FRONTIERS crashes on boot

```
[Debug] <Critical> memory.cpp:operator():655: Assertion Failed!
Mapping cannot fit inside free region: size = 0x52000000
```

[shad_log.txt](https://github.com/user-attachments/files/18050583/shad_log.txt)

squidbus:
> CUSA28194 - SONIC FRONTIERS crashes on boot

This is the crash that happens on main, so no regression. I don't have this crash so it fixes it for some setups at least. Or maybe the USA version doesn't have the issue for some reason?

kvicken:
> > CUSA28194 - SONIC FRONTIERS crashes on boot
> 
> This is the crash that happens on main, so no regression. I don't have this crash so it fixes it for some setups at least. Or maybe the USA version doesn't have the issue for some reason?

What's your game update version? i'm on 01.00

squidbus:
> What's your game update version? i'm on 01.00

Ah, I'm using the latest update.

kvicken:
> Ah, I'm using the latest update.

That must be it then. I can't update the game, for some reason the emulator crashes when I try to install an update for this game.

C0rn3j:
@kvicken likely the same issue as #1701

PR #1696, opened 2024-12-08
Better title bar for Cheats/Patches menu

Xphalnos:
Currently the title bar only shows "Cheats / Patches". Now it has the game title in addition.

Before:
image1

After:
image

### DIFF
--- a/src/qt_gui/cheats_patches.cpp
+++ b/src/qt_gui/cheats_patches.cpp
@@ -39,7 +39,7 @@ CheatsPatches::CheatsPatches(const QString& gameName, const QString& gameSerial,
       m_gameSize(gameSize), m_gameImage(gameImage), manager(new QNetworkAccessManager(this)) {
     setupUI();
     resize(500, 400);
-    setWindowTitle(tr("Cheats / Patches"));
+    setWindowTitle(tr("Cheats / Patches for ") + m_gameName);
 }
 
 CheatsPatches::~CheatsPatches() {}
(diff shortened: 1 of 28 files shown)


GHU7924:
I apologize for the OFFTOP, but I have a question about improvement. In order not to create new topics once again, I decided to write a message here, I hope it will not cause you any inconvenience.

Patches

I think you've already understood everything from the content of the screenshots, but I'll explain a little bit... while cheats can be deleted, patches can only be saved, could you also add a button to delete patches so as not to do it manually through the folder? 
The save button would become common to the entire window.  Is such a change possible?

EDIT: Or do something like this
Patches2

Xphalnos:
@GHU7924, I understand the usefulness of this but I don't think I'll be able to implement it. Maybe @DanielSvoboda will be able to since he's the one who created the cheat manager.

src/qt_gui/cheats_patches.cpp
```diff
@@ -39,7 +39,7 @@ CheatsPatches::CheatsPatches(const QString& gameName, const QString& gameSerial,
       m_gameSize(gameSize), m_gameImage(gameImage), manager(new QNetworkAccessManager(this)) {
     setupUI();
     resize(500, 400);
-    setWindowTitle(tr("Cheats / Patches"));
+    setWindowTitle(tr("Cheats / Patches for ") + m_gameName);
```
Osyotr, line 42: This should have been `setWindowTitle(tr("Cheats / Patches for %1").arg(m_gameName));` for better translations.

DanielSvoboda:
putting 'download' and 'delete' on the left side and 'save' on the right side seems to be better. In addition to deleting the file, you have to update the files.json file. It shouldn't be too difficult to do this, I can take advantage and translate the 'for'

PR #1697, opened 2024-12-08
Displays "Never Played" if the game has never been played

Xphalnos:
In the game list, if the game has never been played, instead of showing 0, it shows "Never Played".

Before:
Before

After:
After


DanielSvoboda:
Could change 'pt_BR.ts', the translation would be:
Nunca jogado

Xphalnos:
I resolved the conflicts with git. I also updated pt_BR.ts thanks to @DanielSvoboda. Ready to be merged!

GHU7924:
@Xphalnos It seems that there is a small moment, insignificant, but it is there. If you start a game you've never played and it crashes after a few seconds, the value will be 0. And since you changed the value of "0" to "never played", it might also be worth adding something like "less than a minute" for such cases. This is not so important, but on the other hand, this moment does not fit into the concept that you have implemented.


EDIT: Sorry, it wasn't the "Main Build", it looks like everything is fine with you, so this is a problem with another build and is related to something else.

DanielSvoboda:
Although hours and minutes are displayed, seconds are also recorded. A possible improvement would be to compare the value less than 1 instead of comparing it to zero.

PR #1683, opened 2024-12-07
renderer_vulkan: Introduce shader HLE system with copy shader implementation.

squidbus:
Introduces a system for implementing more efficient HLE macros for simple shaders.

In this case, implements a macro for a copy shader used by Bloodborne to copy arbitrary blocks of memory from one buffer to another. Currently this shader is responsible for big spikes in memory usage from mapping 2GB+ buffers of memory space to arbitrarily copy between. On a real console this would not be an issue since it just provides a window into the memory range, but for us this results in buffer memory allocations to cover the entire range.

By implementing this shader in HLE we can avoid these giant buffer allocations, instead only mapping much smaller areas where the copy actually occurs. This should eliminate memory spikes, reduce overall VRAM usage, and be generally more performant. We hope to have a more low-level solution for this problem in the future, but pending that this is a good way to keep memory usage under control and improve performance.

This also allows Bloodborne to go in-game on macOS (if compiled using the latest dev build of MoltenVK) since this shader was the main blocker for that. Still needs some more improvements to be fully stable.

Thanks to @raphaelthegreat for implementing the algorithm to batch the copies together.

PR #1705, opened 2024-12-08
Devtools - Shader editing

viniciuslrangel:
image

### DIFF
--- a/src/common/string_util.cpp
+++ b/src/common/string_util.cpp
@@ -37,6 +37,10 @@ std::vector<std::string> SplitString(const std::string& str, char delimiter) {
     return output;
 }
 
+std::string_view U8stringToString(std::u8string_view u8str) {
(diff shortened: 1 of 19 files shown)


Foul-Tarnished:
Cpuld this be used with #1683 to make game mods ? I mean try stuff with this, then export the shader and load it with #1683 automatically

Sure it may be easier to edit the game files directly, if there's a modding scene and tools to edit these files already

squidbus:
> Could this be used with #1683 to make game mods ? I mean try stuff with this, then export the shader and load it with #1683 automatically
> 
> Sure it may be easier to edit the game files directly, if there's a modding scene and tools to edit these files already

Not sure if you got the correct PR there as it's not related to modifying shaders like this. We do have https://github.com/shadps4-emu/shadPS4/pull/1633 in place which allows loading patched shader files, presumably you could take the shader you made here, compile to SPIR-V, and insert it in as a patched shader file through there.

PR #1724, opened 2024-12-10
thread: Apply alternate signal stack to created threads.

squidbus:
With guest stacks in place, if the guest asks for a very small stack size we may overflow the stack in our signal handling. Notably this can happen to more threads on macOS, since we use signal handlers more heavily for patching a few instructions not supported by Rosetta 2.

To solve this, apply an alternate signal stack to each thread created. When signal handlers execute they will switch to this alternate stack instead of executing on the thread stack.

Fixes crashes launching CUSA04551 (NieR:Automata) on macOS, probably helps Linux and some other games as well.

### DIFF
--- a/src/core/thread.cpp
+++ b/src/core/thread.cpp
@@ -1,15 +1,15 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
-#include "libraries/kernel/threads/pthread.h"
-#include "thread.h"
-
+#include "common/alignment.h"
 #include "core/libraries/kernel/threads/pthread.h"
+#include "thread.h"
 
 #ifdef _WIN64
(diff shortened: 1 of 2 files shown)


C0rn3j:
```c++
/build/shadps4-git/src/shadps4/src/core/thread.cpp: 
In member function ‘void Core::NativeThread::Initialize()’:

/build/shadps4-git/src/shadps4/src/core/thread.cpp:147:5: 
error: designator order for field ‘stack_t::ss_flags’ does not match declaration order in ‘const stack_t’
```

Fails to build on Arch Linux...

I do have an extra patch with 1683 changes in there but those have nothing to do with threads it seems...

EDIT: so 1683 got merged as I was trying this, will retry master without patches just in case, but still doubt it was the issue.

EDIT2: Still an issue.

EDIT3: I think I fixed it? Will send a PR if the build passes...

PR #1726, opened 2024-12-10
cache: Invalidate pages for file reads.

squidbus:
* Updates the cache invalidation logic to accept an arbitrary region of memory and handle figuring out the pages to invalidate locally.
* Adds logic to file reads to invalidate the target memory before reading, as otherwise the read calls will fail instead of triggering the usual invalidation logic. This causes infinite loops on games like NieR:Automata.

### DIFF
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -1,21 +1,21 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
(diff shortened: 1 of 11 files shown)


smiRaphi:
This also let's persona 3 dancing load into gameplay :D

Emulator-Team-2:
The Legend of Korra goes Ingame now with this PR.

StevenMiller123:
Fixes the inconsistent memory crashes and loading crashes in Kingdom Hearts 0.2 Final Chapter Prologue. 
image

squidbus:
Regressions seem to be resolved.

PR #1729, opened 2024-12-10
Fix compiling due to typedefs varying across platforms

C0rn3j:
Fix the var ordering.

https://github.com/shadps4-emu/shadPS4/pull/1724 breaks it and CI does not catch it, as it is using clang ~~a compiler old enough to be complacent~~.

Note that I do not usually do C++ so the fix may be wrong, even though it works for me.

```c++
typedef struct
  {
    void *ss_sp;
    int ss_flags;
    size_t ss_size;
  } stack_t;
```
Fixes `error: designator order for field ‘stack_t::ss_flags’ does not match declaration order in ‘const stack_t’`

*Do squash the extra commit when merging please*

### DIFF
--- a/src/core/thread.cpp
+++ b/src/core/thread.cpp
@@ -141,8 +141,8 @@ void NativeThread::Initialize() {
                "Failed to allocate signal stack: {}", errno);
 
     const stack_t sig_stack = {
-        .ss_flags = 0,
         .ss_sp = sig_stack_ptr,
+        .ss_flags = 0,
         .ss_size = sig_stack_size,
     };
     ASSERT_MSG(sigaltstack(&sig_stack, nullptr) == 0, "Failed to set signal stack: {}", errno);
--- a/src/core/thread.cpp
+++ b/src/core/thread.cpp
@@ -141,9 +141,9 @@ void NativeThread::Initialize() {
                "Failed to allocate signal stack: {}", errno);
 
     const stack_t sig_stack = {
+        .ss_flags = 0,
         .ss_sp = sig_stack_ptr,
         .ss_size = sig_stack_size,
-        .ss_flags = 0,
     };
     ASSERT_MSG(sigaltstack(&sig_stack, nullptr) == 0, "Failed to set signal stack: {}", errno);
 #endif
--- a/src/core/thread.cpp
+++ b/src/core/thread.cpp
@@ -115,7 +115,7 @@ void NativeThread::Exit() {
 #else
     // Disable and free the signal stack.
     constexpr stack_t sig_stack = {
-        ss_flags: SS_DISABLE,
+        .ss_flags = SS_DISABLE,
     };
     sigaltstack(&sig_stack, nullptr);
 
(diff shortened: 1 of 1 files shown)


squidbus:
Two things:
* AFAIK this is a GCC thing, not a "modern C++ compiler" thing.
* The order of fields used currently is how they appear on my system, not how you have set them.

C0rn3j:
> The order of fields used currently is how they appear on my system, not how you have set them.

The typedef is taken from `signal.h` which includes `stack_t.h`:
image


EDIT: Even the docs have a mistmatching order with my system 2.40 glibc????

https://www.gnu.org/software/libc/manual/html_node/Signal-Stack.html

squidbus:
stack_t

C0rn3j:
That's fun... does this simple declaration need to have a platform dependent ifdef then, or is there a more reasonable way to solve this?

> AFAIK this is a GCC thing, not a "modern C++ compiler" thing.

I was under the impression that this here runs the same thing as my system - I suppose gcc - just an old version of it?

https://github.com/shadps4-emu/shadPS4/blob/e5e1aba24172b4cf41e1c4ea56adf0b0984b815b/.github/workflows/build.yml#L373-L374

raphaelthegreat:
Would probably change it to not use designated initializers to avoid platform ifdef

squidbus:
> I was under the impression that this here runs the same thing as my system - I suppose gcc - just an old version of it?

https://github.com/shadps4-emu/shadPS4/blob/main/.github/workflows/build.yml#L370-L371

We configure the build to use clang here.

C0rn3j:
image

I am beyond confused as to how my system version mismatches the upstream git version last touched in 2016 AND documentation also agrees with this git version and not what's installed on my system...

Even the commit that [Arch supposedly installs](https://gitlab.archlinux.org/archlinux/packaging/packages/glibc/-/blob/main/PKGBUILD?ref_type=heads#L12) has flags last.

https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/types/stack_t.h;h=21610f1d3d6ad004f10642fc5d8556688d961426;hb=aa533d58ff12e27771d9c960a727d74992a3f2a3

Looks like I will be taking this to my distribution first.

EDIT: Nope, glibc bs.

https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/types/stack_t.h;h=497e42b0ff6d19cdd287478d12ac4916d7889d89;hb=0bcec5321f7d31c31e4fd601d4e0922862d87b14

https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/types/stack_t.h;h=3cf0a4019a044f9235477606c45eb2642d0307a7;hb=0bcec5321f7d31c31e4fd601d4e0922862d87b14

EDIT2: And this VARIES between different Linux platforms on top of that - https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=0bcec5321f7d31c31e4fd601d4e0922862d87b14#patch2  
`bits/types/stack_t.h has three variants different in the order of the structure elements (generic = MIPS Linux, and other Linux)` Well that was a waste of time, time to figure out how to do it without designated initializers, no clue why you'd have the names if you'd then ignore them and use the order anyway, but now I am just ranting about C++ :)

C0rn3j:
That works, builds, passes CI and it's not a designated initializer, but not tagged constant either anymore.

Is this fine as is?

squidbus:
This looks fine to me yes.

Issue #1721, opened 2024-12-09
Deleting a game does not immediately remove it from the list


C0rn3j:
Arch Linux, Plasma, Wayland, Nvidia, built from AUR/shadps4-git on https://github.com/shadps4-emu/shadPS4/commit/0b59ebb22fd64bd4c60d2fe409966900b617428e.

Random log unrelated to this issue, with more system info: [shad_log.txt](https://github.com/user-attachments/files/18051640/shad_log.txt)

---

If I try to remove patched `Peggle 2 1.00 -> 1.01 (CUSA00743)`, (or presumably any other game), it seems to get removed, but the game list does not refresh.

Nothing is logged in STDOUT/STDERR.

Repeatedly removing seems to still do nothing (shouldn't it at least throw an error about a missing directory now?).

Restarting the emulator loads the game list without the deleted game fine.

Hermiten:
Fixed #1730

PR #1730, opened 2024-12-10
Remove game from list after deletion

DanielSvoboda:
Remove the game from the list after deleting it using the 'Delete Game' button, as mentioned in issues https://github.com/shadps4-emu/shadPS4/issues/1721


ghost:
Small Issue: Game Entry is removed when delete update only. Game ist still installed but the list must be reloaded manually.

Issue #1669, opened 2024-12-05
Save option settings on close


DanielSvoboda:
In the two codes below, you can see that when you close the emulator, the settings are saved.
This means that even when you press Cancel in the settings menu, the settings will be saved.

I had tried to adjust this in the AutoUpdater PR, but I removed this attempt because it had nothing to do with that PR.
If I'm not mistaken this happens to save the screen position when closing...

https://github.com/shadps4-emu/shadPS4/blob/main/src/emulator.cpp#L100
```cpp
Emulator::~Emulator() {
    const auto config_dir = Common::FS::GetUserPath(Common::FS::PathType::UserDir);
    Config::save(config_dir / "config.toml");
}
```
----

https://github.com/shadps4-emu/shadPS4/blob/main/src/qt_gui/main_window.cpp#L38
```cpp
MainWindow::~MainWindow() {
    SaveWindowState();
    const auto config_dir = Common::FS::GetUserPath(Common::FS::PathType::UserDir);
    Config::save(config_dir / "config.toml");
}
```
----

https://github.com/user-attachments/assets/c58aa35b-9650-4828-8e25-0383dd3d84b1

rainmakerv3:
I looked into this , and it seems there is also a related issue is actually not just caused by the Emulator or Mainwindow destructors mentioned above, but also because the function loading the values for widgets in settings_dialog.cpp uses the values currently in memory rather the values written to the config file.

Even after removing the function call to config::save in each destructor (there should probably be another function saving just window dimensions and such), the issue will still happen if you close and reopen the settings dialog after changing a setting

https://github.com/user-attachments/assets/8eb6fcc3-a67e-4ef6-aa22-cea5a59f3f19

I don't know if there is a better solution for this, but for example changing this in https://github.com/shadps4-emu/shadPS4/blob/main/src/qt_gui/settings_dialog.cpp#L323:

    ui->consoleLanguageComboBox->setCurrentIndex(
        std::distance(
            languageIndexes.begin(),
            std::find(languageIndexes.begin(), languageIndexes.end(), Config::GetLanguage())) %
        languageIndexes.size());

to this:

 ui->consoleLanguageComboBox->setCurrentIndex(
        std::distance(
            languageIndexes.begin(),
            std::find(languageIndexes.begin(), languageIndexes.end(), toml::find_or<int>(data, "Settings", "consoleLanguage", 6))) %
        languageIndexes.size());

(plus the toml parsing code)

[...]

PR #1747, opened 2024-12-12
Resubmit - Prevent settings from being saved when close button is pressed instead of save 

rainmakerv3:
Resubmitting PR#1679 because I accidentally deleted that branch instead of the one that was merged lol.

------------

Proposed fix for https://github.com/shadps4-emu/shadPS4/issues/1669, more detail on the issue can be seen on that page

There are actually two issues which cause settings to be saved when the close button is pressed:

MainWindow and Emulator destructor functions both call the Config::Save function, which saves every single config variable, including those changed in the config settings dialog box. To resolve this, I created another save function for just those two destructors, which save only the variables related to the main window itself, such as main window dimensions, recent files, and grid/list view settings.

Config variables are connected to each change in the config windows QT widgets, and the UI loads in these variables instead of the values saved to the config file. I rewrote this such that the config variables change value only upon saving, and that UI values are loaded from the config file

Not included in this PR is fixing how the translations are re-applied when changing emulator language.

### DIFF
--- a/src/common/config.h
+++ b/src/common/config.h
@@ -13,6 +13,7 @@ enum HideCursorState : s16 { Never, Idle, Always };
 
 void load(const std::filesystem::path& path);
 void save(const std::filesystem::path& path);
+void saveMainWindow(const std::filesystem::path& path);
 
 bool isNeoMode();
 bool isFullscreenMode();
@@ -67,6 +68,7 @@ void setNeoMode(bool enable);
 void setUserName(const std::string& type);
 void setUpdateChannel(const std::string& type);
 void setSeparateUpdateEnabled(bool use);
+void setGameInstallDirs(const std::vector<std::filesystem::path>& settings_install_dirs_config);
 
 void setCursorState(s16 cursorState);
 void setCursorHideTimeout(int newcursorHideTimeout);
--- a/src/common/config.h
+++ b/src/common/config.h
@@ -68,7 +68,7 @@ void setNeoMode(bool enable);
 void setUserName(const std::string& type);
 void setUpdateChannel(const std::string& type);
 void setSeparateUpdateEnabled(bool use);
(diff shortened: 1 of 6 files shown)


src/common/config.h
```diff
@@ -67,6 +68,7 @@ void setNeoMode(bool enable);
 void setUserName(const std::string& type);
 void setUpdateChannel(const std::string& type);
 void setSeparateUpdateEnabled(bool use);
+void setGameInstallDirs(const std::vector<std::filesystem::path>& settings_install_dirs_config);
```
raphaelthegreat, line 71:
```suggestion
void setGameInstallDirs(std::span<const std::filesystem::path> settings_install_dirs_config);
```

raphaelthegreat (approved):
Only minor nit but LGTM

src/common/config.h
rainmakerv3: The variable used in this function 

void setGameInstallDirs(std::span<const std::filesystem::path> settings_install_dirs_config) {
    settings_install_dirs = settings_install_dirs_config;
}

can be passed to the addGameInstallDir function, which is written for vector (as well as a similar function for removing install folders)

config.cpp: L354:
bool addGameInstallDir(const std::filesystem::path& dir) {
    if (std::find(settings_install_dirs.begin(), settings_install_dirs.end(), dir) ==
        settings_install_dirs.end()) {
        settings_install_dirs.push_back(dir);
        return true;
    }
    return false;
} 

My understanding is that the vector would have to be stored so that we can freely add and move directories, so at some point the span would have to be converted to vector to use these functions. If this understanding is correct, perhaps we can just leave the function as vector?

raphaelthegreat: Looking at the function again I think indeed this should be made back into a vector as it copies the entries into the config

rainmakerv3: ok will revert then

PR #1755, opened 2024-12-12
renderer_vulkan: Fix some color attachment indexing issues.

squidbus:
* For both of the cases where attachment bindings are skipped, due to `color_target_mask` and shader `mrt_mask`, make sure that:
  * The render target binding is `VK_NULL_HANDLE`.
  * The format is set to undefined and the attachment index is skipped over in color attachment info passes.
* Keep a counter of the number of color attachments in the pipeline key, as `std::ranges::find(key.color_formats, vk::Format::eUndefined)` is not sufficient when there are potentially undefined attachments in between defined attachments.
* Silence validation warning about stores to undefined output when masked out by `color_target_mask`.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
+++ b/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
@@ -229,17 +229,15 @@ GraphicsPipeline::GraphicsPipeline(const Instance& instance_, Scheduler& schedul
         });
     }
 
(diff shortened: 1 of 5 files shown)


Missake212:
Fixes the snow being gone on Intel iGPUs and AMD GPUs in Bloodborne.

Intel:
image

AMD:
image

raphaelthegreat (approved):

PR #1760, opened 2024-12-13
Fix uppercase config filenames for Linux

rainmakerv3:
Apologies for this, I used uppercase for the config file in some lines of my previous PR. This fixes it so it does not break Linux


ngoquang2708:
It works now. Thanks.

rainmakerv3:
thanks for the heads up as well @ngoquang2708 !

PR #1763, opened 2024-12-13
graphics: Improve handling of color buffer and storage image swizzles

raphaelthegreat:
* Removed some color buffer component swaps that were wrong since eR8G8B8A8Unorm and eA8B8G8R8UnormPack32 are technically the same format for little-endian GPUs, which both AMD and NVIDIA are. This lets the shader specialization mechanism handle it instead.

* Handle swizzles for storage image writes by swizzling incoming texel in the shader, as well as format buffer swizzles since BUFFER_LOAD_FORMAT and BUFFER_STORE_FORMAT use dst_sel from the sharp. ~For storage images specifically we only perform manual swizzling if the format isn't 8_8_8_8 since those can be swizzled with image view and is already mostly handled~

* Ignore mrtz stencil exports since those are unsupported for now

### DIFF
--- a/src/video_core/amdgpu/resource.h
+++ b/src/video_core/amdgpu/resource.h
@@ -208,14 +208,6 @@ struct Image {
         return dst_sel_x | (dst_sel_y << 3) | (dst_sel_z << 6) | (dst_sel_w << 9);
     }
 
-    bool CanSwizzleWithFormat() const {
-        // BGRA
-        if (DstSelect() == 0b111100101110 && GetDataFmt() == DataFormat::Format8_8_8_8) {
-            return true;
-        }
-        return false;
-    }
-
     CompSwizzle GetSwizzle(u32 comp) const noexcept {
         const std::array select{dst_sel_x, dst_sel_y, dst_sel_z, dst_sel_w};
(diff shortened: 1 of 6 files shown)


src/video_core/amdgpu/resource.h
```diff
@@ -204,6 +208,19 @@ struct Image {
         return dst_sel_x | (dst_sel_y << 3) | (dst_sel_z << 6) | (dst_sel_w << 9);
     }
 
+    bool CanSwizzleWithFormat() const {
+        // BGRA
+        if (DstSelect() == 0b111100101110 && GetDataFmt() == DataFormat::Format8_8_8_8) {
+            return true;
```
squidbus: One caveat is that, while I added some more BGRA formats recently, I learned that they aren't supported on NVIDIA for storage. Note the ones with low percents:
format

raphaelthegreat: Hmm, should we then remove the existing swizzle with view implementation and rely only on the shader system

squidbus: Might be a good idea yeah.

squidbus (approved):

Issue #1737, opened 2024-12-11
Background image when selecting a game doesn't scale properly when maximizing the window


Missake212:
When selecting a game in the list of games on the emulator, it displays an image that fills up the background of the emulator window, when clicking on the maximize button on the top right corner, the image doesn't take up the full screen but instead stays at the same size it was on the emulator window and creates copies of itself to fill up the rest of the background. 

https://github.com/user-attachments/assets/c3ef36cf-64d4-4f92-abe6-a35e44414e20

Hermiten:
Maybe for you @DanielSvoboda ? :)

DanielSvoboda:
I had never noticed this, only now I noticed that it is like this. It would be cool to have an option to disable the image in the settings, in case someone wants that. However, I wasn't the one who made that option, I don't know how it works.

DanielSvoboda:
@Missake212 Test this:
https://github.com/shadps4-emu/shadPS4/actions/runs/13346603636

PR #1769, opened 2024-12-14
Resize trophy pop up size based on window size

rainmakerv3:
Right now the current trophy window is hardcoded to a certain size that seems to be optimized for 720:

Main 720p window:

image

It looks very small however at larger window sizes:

1440p window:

image

This PR adjusts the trophy window to be the same at any window size.

PR 720p window:

image

PR 1440p window:

image

I also increased the text size by 20% and lengthened the window to accommodate this, for better visibility. Also tested to work on fullscreen mode

### DIFF
--- a/src/core/libraries/np_trophy/trophy_ui.cpp
+++ b/src/core/libraries/np_trophy/trophy_ui.cpp
@@ -38,21 +38,22 @@ void TrophyUI::Finish() {
 void TrophyUI::Draw() {
     const auto& io = GetIO();
 
+    float AdjustWidth = io.DisplaySize.x / 1280;
+    float AdjustHeight = io.DisplaySize.y / 720;
     const ImVec2 window_size{
-        std::min(io.DisplaySize.x, 250.f),
-        std::min(io.DisplaySize.y, 70.f),
+        std::min(io.DisplaySize.x, (300 * AdjustWidth)),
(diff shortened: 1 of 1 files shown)


Hermiten:
Nice fix !
Can I ask you for this ? https://github.com/shadps4-emu/shadPS4/issues/1737 seems to be something similar to fix :p

rainmakerv3:
That looks tricky I'll see first if I can manage it

ZuwasieMan:
@rainmakerv3 very nice job.
Btw I think the screen scale is not being taken into account here? If screen scale is not 100%, text might not be readable.

i.e. screen scale set to 100%:
grafik


i.e. screen scale set to 200%:
grafik

rainmakerv3:
You're right, it does not take into account system DPI settings, I could look into that in the future if I have time

rainmakerv3:
actually this may be a simpler fix than I thought. Can you try this build to see if it fixes the issue? @ZuwasieMan 

https://github.com/rainmakerv3/shadPS4/actions/runs/12328830836

rainmakerv3:
I thought it was an easy fix, but that windows screen scale setting is not based on DPI it seems as the DPI functions return the same value with it on or off so I don't know of a way to adjust for it o.O. I do know that the same problem is present before this PR, so at least I know this wasn't an issue introduced by these changes

PR #1770, opened 2024-12-14
shader_recompiler: Support VK_AMD_shader_image_load_store_lod for IMAGE_STORE_MIP

squidbus:
* Wires up the texture instruction parameters and argument filling to support `IMAGE_STORE_MIP` in IR.
  * Added parameters or swapped order slightly for `ImageRead/Write/Fetch` so LOD is always the same argument index, for easier patching in resource tracking pass.
* For SPIR-V, implements `ImageWrite` with LOD using `VK_AMD_shader_image_load_store_lod`.
  * This is a trivial implementation available only on AMD and Apple GPUs, will need emulated fallback for other GPUs using separate mip bindings.

Used by CUSA03694.


raphaelthegreat (approved):

kalaposfos13:
Are there any plans to make an implementation for Intel and Nvidia GPUs too?

PR #1775, opened 2024-12-14
DmaData and Recompiler fixes

roamic:
`DmaData`: handler now prefers copying GPU buffers over CPU. If readback attempt is encountered it will create a GPU buffer still until readback is properly implemented.

Recompiler: fix branch linker for `s_branch` instructions reading dead code. The issue resulted in `s_branch` instruction being ignored. We now insert a fake label for all code that follows this instruction so that the block ends on it.

This whole change should fix GS particle effects in bloodborne.

### DIFF
--- a/src/shader_recompiler/frontend/control_flow_graph.cpp
+++ b/src/shader_recompiler/frontend/control_flow_graph.cpp
@@ -80,6 +80,7 @@ void CFG::EmitLabels() {
         if (inst.IsUnconditionalBranch()) {
             const u32 target = inst.BranchTarget(pc);
             AddLabel(target);
+            AddLabel(pc + inst.length);
         } else if (inst.IsConditionalBranch()) {
             const u32 true_label = inst.BranchTarget(pc);
             const u32 false_label = pc + inst.length;
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -569,25 +569,39 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
                     break;
                 }
                 if (dma_data->src_sel == DmaDataSrc::Data && dma_data->dst_sel == DmaDataDst::Gds) {
+                    LOG_CRITICAL(Render_Vulkan, "DmaData Data ({}) -> Gds {:#x}, {} bytes",
+                                 dma_data->data, dma_data->dst_addr_lo, dma_data->NumBytes());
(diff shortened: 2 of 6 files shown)


raphaelthegreat (approved):
👀

setepenre:
I've spent a lot of time investigating which code path, testing with Bloodborne, provokes long stuttering (seconds to minutes of freeze frame with GPU usage at 99% and very low read/write speeds) ; it's the combination of memory to GDS and memory to memory copy buffer calls, using same src addresses, leading I guess to sync problems on the vulkan side when copying from the same source buffer. I have tried a whole lot of stage and access masks configurations managing only to somewhat mitigating those stutters but never actually eliminating them.
There's another class of slowdowns provoked by the particle effects enabled with this in Bloodborne when a vertex explosion happens to one or more particles.

roamic:
> I've spent a lot of time investigating which code path, testing with Bloodborne, provokes long stuttering (seconds to minutes of freeze frame with GPU usage at 99% and very low read/write speeds) ; it's the combination of memory to GDS and memory to memory copy buffer calls, using same src addresses, leading I guess to sync problems on the vulkan side when copying from the same source buffer. I have tried a whole lot of stage and access masks configurations managing only to somewhat mitigating those stutters but never actually eliminating them. There's another class of slowdowns provoked by the particle effects enabled with this in Bloodborne when a vertex explosion happens to one or more particles.

If you want theories, I've got a more convincing one.

Facts:
* The data copied by the `DmaData` packets is only 4 bytes in each call.
* It is generated on GPU by particle emit shader so it always resides in GPU memory.
* It is used as `u32 count` in a `for` loop in particle simulation compute shaders to generate the particles.

Theory:
It's the same issue as vertex explosions. The data on GPU gets overwritten by the garbage data from the CPU. Sometimes it's `0`s but sometimes it can be real garbage. Since it is used as a break count in `for` loop any garbage data that sets it to a large number will cause a long loop with 100% GPU consumption. With large numbers it can probably cause out-of-bounds access resulting in a crash.
I didn't get to proving this theory, but if you want to, be my guest. The solution though is memory readback which is very hard to implement without losing a lot of performance or significantly refactoring the renderer.

setepenre:
thanks for the insight!

Linear524:
About the idea of "significant refactoring the renderer" - how about switching from Vulkan to D3D12 + Rasterizer Ordered View (ROV) technique ?
I heard that this way of rendering can solve some difficult "Sync problems" from same source buffer or multiple sources...

P.S. - please don't kill me if my question is too stupid... :D

PR #1780, opened 2024-12-14
mac: Choose whether system Vulkan is needed at runtime.

squidbus:
Switches to using the system Vulkan loader when available, instead of deciding with a build flag, and alters how the included MoltenVK is bundled so that the system Vulkan loader will load it.

Should make it easy for any user to install the Vulkan SDK and enable debug layers without having to build from source, as layers cannot be loaded when using `libMoltenVK.dylib` directly.

Additionally, made some modifications to layer loading so that missing layers will error log instead of crashing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -892,11 +892,16 @@ if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
 endif()
 
 if (APPLE)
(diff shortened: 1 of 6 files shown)


squidbus:
Ready for review. Contains some fixes now to make sure we log and don't crash if the user enables a layer setting without it being installed.

PR #1787, opened 2024-12-14
GPU processor refactoring

psucien:
A bit of refactoring plus fixes for:
- ring read pointer update
- Gnm ASC events notification
- Coroutines CS context switching
- Tracy fiber markers

### DIFF
--- a/src/core/debug_state.cpp
+++ b/src/core/debug_state.cpp
@@ -15,6 +15,7 @@
 using namespace DebugStateType;
 
 DebugStateImpl& DebugState = *Common::Singleton<DebugStateImpl>::Instance();
+extern std::unique_ptr<AmdGpu::Liverpool> liverpool;
 
 static ThreadID ThisThreadID() {
 #ifdef _WIN32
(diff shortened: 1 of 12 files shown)


src/video_core/renderer_vulkan/vk_shader_hle.cpp
```diff
@@ -5,16 +5,16 @@
...
-#include "vk_rasterizer.h"
+extern std::unique_ptr<AmdGpu::Liverpool> liverpool;
 
 namespace Vulkan {
 
 static constexpr u64 COPY_SHADER_HASH = 0xfefebf9f;
 
-bool ExecuteCopyShaderHLE(const Shader::Info& info, const AmdGpu::Liverpool::Regs& regs,
-                          Rasterizer& rasterizer) {
+bool ExecuteCopyShaderHLE(const Shader::Info& info, Rasterizer& rasterizer) {
```
raphaelthegreat: You can pass a reference to ComputeProgram here instead of relying on extern liverpool

src/core/debug_state.cpp
```diff
@@ -142,8 +143,7 @@ void DebugStateImpl::PushQueueDump(QueueDump dump) {
     frame.queues.push_back(std::move(dump));
 }
 
-void DebugStateImpl::PushRegsDump(uintptr_t base_addr, uintptr_t header_addr,
-                                  const AmdGpu::Liverpool::Regs& regs, bool is_compute) {
+void DebugStateImpl::PushRegsDump(uintptr_t base_addr, uintptr_t header_addr, bool is_compute) {
```
raphaelthegreat: To avoid reliance on global state this can be split into separate functions for gfx and compute (PushComputeRegsDump) which receive regs ref and ComputeProgram ref respectively. I'm not actually sure if compute dump relies on full regs

src/video_core/amdgpu/liverpool.h
```diff
@@ -1277,6 +1279,7 @@ struct Liverpool {
     };
 
     Regs regs{};
+    std::array<ComputeProgram, NumComputeRings> asc_sh_regs{};
```
raphaelthegreat: An idea would be to store compute programs in mapped_queues and always access here with qid, so you don't need separate array and maintain a pointer

PR #1789, opened 2024-12-14
fix for "shadPS4" not being given on linux volume mixers

ConnorG512:
small patch to display the shadPS4 name inside of the volume mixer on Linux instead of the generic SDL_Application name

New:
image

Old:
image


polybiusproxy (approved):
Thanks for your contribution 🙂

PR #1795, opened 2024-12-15
renderer_vulkan: Various attachment cleanup and fixes.

squidbus:
* Add helper functions for depth/stencil buffer validity and use where these checks would happen.
* Move pipeline multisampling calculation to simple helper function and use where needed.
* Change out duplicated instances of buffer sample count logic with existing `NumSamples()` function.
* Make depth/stencil attachment formats and bindings more state-consistent.
* Do not bother attaching color buffers when mode is `Disable`, since none will be used.
* Handle resolve pass with single-sampled source by just doing a copy, as single-sample source resolve is not allowed in Vulkan.
* Skip scissor indices that do not have corresponding viewports.
* Misc typo fixes and dead code removal.

Fixes some validation errors in games like CUSA05637.

### DIFF
--- a/src/video_core/amdgpu/liverpool.h
+++ b/src/video_core/amdgpu/liverpool.h
@@ -428,6 +428,14 @@ struct Liverpool {
             BitField<0, 22, u32> tile_max;
         } depth_slice;
 
+        bool DepthValid() const {
+            return Address() != 0 && z_info.format != ZFormat::Invalid;
+        }
+
+        bool StencilValid() const {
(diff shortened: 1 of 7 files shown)


squidbus:
Rebased

squidbus:
Rebased (again, with the new stencil changes)

raphaelthegreat (approved):
Looks good to me, nice cleanup here

Linear524:
Wow ! :)
Gravity Rush 1 now renders all the text correctly ! :) :) :)

PR #1797, opened 2024-12-15
KBM-only remapping

kalaposfos13:
This is a modified version of my [remapping PR](https://github.com/shadps4-emu/shadPS4/pull/1356) that only contains the keyboard and mouse part, as this is already complete and has proven to be mostly bug-free, as it has been extensively tested by people using @diegolix29's fork. The controller part is still WIP and is not ready for merging yet, so until that's ready too, I'll just put this out here following @rainmakerv3's advice. The code still has some controller remapping elements left in, but they aren't hooked up to the main input loop. If I left in anything that I shouldn't have, please let me know.


rainmakerv3:
image

This was my attempt at designing a GUI for this. Since there's so many functions (ability to add up to 3 inputs per output is the most painful one to design), I can't seem to make it take a smaller space.

The intended use for this is that the dropdown boxes list down the possible inputs per output (since there can be up to three), and the user can press the corresponding input button, then input up to three keystrokes to fill in each combobox. Design wise this seems to be the only thing I can think of to support the things this program wants to do, but I am unable to code it this way.

As such, I'm at least putting the UI file here, and maybe someone else can code in the input saving and such.

(saved as txt file, rename to ui to use)

[kbm_config_dialog.txt](https://github.com/user-attachments/files/18144140/kbm_config_dialog.txt)

GHU7924:
@rainmakerv3 I think it's worth combining at least the keyboard, since some people can play on laptops or they don't have a controller, it should have been added a long time ago, and as for the graphical interface, we haven't made significant progress in this direction yet, let's leave it for later.

rainmakerv3:
Yes I agree. My thought is that this would be a good place to leave this for anyone who wants to use it. But like you, I also believe this is good enough to merge in its current state

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+#include "input_handler.h"
+
+#include "fstream"
+#include "iostream"
+#include "list"
+#include "map"
+#include "sstream"
+#include "string"
+#include "unordered_map"
+#include "vector"
```
raphaelthegreat:
```suggestion
#include <vector>
```
Ditto for all other STL includes

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+    std::vector<std::string> tokens;
+    std::stringstream ss(line);
+    std::string token;
+
+    while (std::getline(ss, token, ',')) {
+        tokens.push_back(token);
+    }
+
+    // Check and process tokens
+    for (const auto& t : tokens) {
```
raphaelthegreat:
```suggestion
    for (const auto token : std::views::split(line, ","sv)) {
```
This will avoid allocations

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+        }
+    }
+
+    return InputBinding(key1, key2, key3);
+}
+
+// function that takes a controlleroutput, and returns the array's corresponding element's pointer
+ControllerOutput* GetOutputPointer(const ControllerOutput& parsed) {
+    // i wonder how long until someone notices this or I get rid of it
+    int i = 0;
```
raphaelthegreat: Avoid writing "smart-code" if possible. I dont see any apparent benefit to this and will confuse readers.
I also wonder if this function is necessary, since std::ranges::find() can do the same

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+int mouse_joystick_binding = 0;
+float mouse_deadzone_offset = 0.5, mouse_speed = 1, mouse_speed_offset = 0.1250;
+Uint32 mouse_polling_id = 0;
+bool mouse_enabled = false, leftjoystick_halfmode = false, rightjoystick_halfmode = false;
+
+std::list<std::pair<u32, bool>> pressed_keys;
+std::list<u32> toggled_keys;
+std::list<BindingConnection> connections = std::list<BindingConnection>();
+
+ControllerOutput output_array[] = {
```
raphaelthegreat:
```suggestion
constexpr auto output_array = std::array{
```

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+        if (i[output_array] == parsed) {
+            return &output_array[i];
+        }
+    }
+    return &output_array[i];
+}
+
+void ParseInputConfig(const std::string game_id = "") {
+
+    const auto config_file = Config::GetFoolproofKbmConfigFile(game_id);
```
raphaelthegreat:
```suggestion
    const auto config_file = Config::GetFoolproofKbmConfigFile(game_id);
```

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+// Flags and values for varying purposes
+// todo: can we change these?
+int mouse_joystick_binding = 0;
+float mouse_deadzone_offset = 0.5, mouse_speed = 1, mouse_speed_offset = 0.1250;
+Uint32 mouse_polling_id = 0;
+bool mouse_enabled = false, leftjoystick_halfmode = false, rightjoystick_halfmode = false;
+
+std::list<std::pair<u32, bool>> pressed_keys;
+std::list<u32> toggled_keys;
+std::list<BindingConnection> connections = std::list<BindingConnection>();
```
raphaelthegreat:
```suggestion
static std::vector<BindingConnection> connections;
```
I dont see any advantage to using a std::list, please correct I'm wrong

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+                                "Syntax error: Please provide exactly 2 keys: "
+                                "first is the toggler, the second is the key to toggle: {}",
+                                line);
+                    continue;
+                }
+                ControllerOutput* toggle_out =
+                    GetOutputPointer(ControllerOutput((OrbisPadButtonDataOffset)KEY_TOGGLE));
+                BindingConnection toggle_connection =
+                    BindingConnection(InputBinding(toggle_keys.key2), toggle_out, toggle_keys.key3);
+                connections.insert(connections.end(), toggle_connection);
```
raphaelthegreat:
```suggestion
                connections.emplace_back(InputBinding(toggle_keys.key2), toggle_out, toggle_keys.key3);
```
Less lines and constructs object in-place. Ditto for all other cases of insert() like this

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+        return e.key.key;
+    case SDL_EVENT_MOUSE_BUTTON_DOWN:
+    case SDL_EVENT_MOUSE_BUTTON_UP:
+        return (u32)e.button.button;
+    case SDL_EVENT_MOUSE_WHEEL:
+    case SDL_EVENT_MOUSE_WHEEL_OFF:
+        return GetMouseWheelEvent(e);
+    case SDL_EVENT_GAMEPAD_BUTTON_DOWN:
+    case SDL_EVENT_GAMEPAD_BUTTON_UP:
+        return (u32)e.gbutton.button + 0x10000000; // I believe this range is unused
```
raphaelthegreat:
```suggestion
        return (u32)e.gbutton.button + 0x10000000; // I believe this range is unused
```
What is this magic value? Could we get a constexpr variable with a name instead?

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+    case SDL_EVENT_GAMEPAD_AXIS_MOTION:
+        // todo: somehow put this value into the correct connection
+        // solution 1: add it to the keycode as a 0x0FF00000 (a bit hacky but works I guess?)
+        // I guess in software developement, there really is nothing more permanent than a temporary
+        // solution
+        value_mask = (u32)((e.gaxis.value / 256 + 128) << 20); // +-32000 to +-128 to 0-255
```
raphaelthegreat: Ditto for above, this code is hard to understand at a glance

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+    ControllerOutput(OrbisPadButtonDataOffset::None, Axis::TriggerLeft),
+    ControllerOutput(OrbisPadButtonDataOffset::None, Axis::TriggerRight),
+
+    // the "sorry, you gave an incorrect value, now we bind it to nothing" output
+    ControllerOutput(OrbisPadButtonDataOffset::None, Axis::AxisMax),
+};
+
+// We had to go through 3 files of indirection just to update a flag
+void ToggleMouseEnabled() {
+    mouse_enabled ^= true;
```
raphaelthegreat:
```suggestion
    mouse_enabled = !mouse_enabled;
```
While this code is more optimal than the bitwise not, the compiler can already do this optimization and its less readable. So lets do the simpler thing methinks

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+        controller->Axis(0, axis_x, GetAxis(-0x80, 0x80, a_x));
+        controller->Axis(0, axis_y, GetAxis(-0x80, 0x80, a_y));
+    } else {
+        controller->Axis(0, axis_x, GetAxis(-0x80, 0x80, 0));
+        controller->Axis(0, axis_y, GetAxis(-0x80, 0x80, 0));
+    }
+}
+Uint32 MousePolling(void* param, Uint32 id, Uint32 interval) {
+    auto* data = (GameController*)param;
+    UpdateMouse(data);
```
raphaelthegreat: This separate UpdateMouse function is unnecessary, it can be merged into MousePolling

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+        bool active = false, analog_input_detected = false;
+        IsInputActive(it, active, analog_input_detected);
+        it.output->AddUpdate(active, analog_input_detected, it.parameter);
+    }
+    for (auto& it : output_array) {
+        it.FinalizeUpdate();
+    }
+}
+
+void UpdateMouse(GameController* controller) {
```
raphaelthegreat: I think all the mouse functions and state should be split into a separate mouse.h/cpp files

src/sdl_window.cpp
```diff
@@ -192,192 +203,63 @@ void WindowSDL::OnResize() {
...
-        axis = Input::Axis::TriggerRight;
-        if (event->type == SDL_EVENT_KEY_DOWN) {
-            axisvalue += 255;
-        } else {
-            axisvalue = 0;
+void WindowSDL::OnKeyboardMouseInput(const SDL_Event* event) {
+    using Libraries::Pad::OrbisPadButtonDataOffset;
+
+    // get the event's id, if it's keyup or keydown
+    bool input_down = event->type == SDL_EVENT_KEY_DOWN ||
```
raphaelthegreat:
```suggestion
    const bool input_down = event->type == SDL_EVENT_KEY_DOWN ||
```
Ditto for other cases of const correctness

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+        u32 value_to_search = value & 0xF00FFFFF;
+        for (auto& it = --pressed_keys.end(); (it->first & 0x80000000) != 0; it--) {
+            if ((it->first & 0xF00FFFFF) == value_to_search) {
+                it->first = value;
+                // LOG_DEBUG(Input, "New value for {:X}: {:x}", value, value);
+                return true;
+            }
+        }
+        // LOG_DEBUG(Input, "Input activated for the first time, adding it to the list");
+        pressed_keys.insert(pressed_keys.end(), {value, false});
```
raphaelthegreat: This inserts the key into pressed_keys without checking is_pressed, is it okay?
```suggestion
        pressed_keys.emplace_back(value, false);
```

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+}
+
+// Updates the list of pressed keys with the given input.
+// Returns whether the list was updated or not.
+bool UpdatePressedKeys(u32 value, bool is_pressed) {
+    // Skip invalid inputs
+    if (value == (u32)-1) {
+        return false;
+    }
+    if ((value & 0x80000000) != 0) {
```
raphaelthegreat: This value seems has special meaning depending some bits. Please make a bitfield struct describing its layout and use that. Manual bit parsing of random parts can be hard to memorize

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+    for (const auto& t : tokens) {
+        if (string_to_keyboard_key_map.find(t) != string_to_keyboard_key_map.end()) {
+            // Map to keyboard key
+            u32 key_id = string_to_keyboard_key_map.at(t);
+            if (!key1)
+                key1 = key_id;
+            else if (!key2)
+                key2 = key_id;
+            else if (!key3)
+                key3 = key_id;
```
raphaelthegreat: This can be a lambda defined in the function to reduce code duplication

src/input/input_handler.cpp
```diff
@@ -0,0 +1,694 @@
...
+    mouse_speed = 1;
+    mouse_speed_offset = 0.125;
+    int lineCount = 0;
+
+    std::ifstream file(config_file);
+    std::string line = "";
+    while (std::getline(file, line)) {
+        lineCount++;
+        // strip the ; and whitespace
+        line.erase(std::remove(line.begin(), line.end(), ' '), line.end());
```
raphaelthegreat:
```suggestion
        line.erase(std::ranges::remove_if(line, std::isspace), line.end());
```
std::isspace will handle more cases

kalaposfos13:
@raphaelthegreat I did the easy changes, and I'll do the rest tomorrow. So far I've only one comment on this:
`std::ranges::find` does indeed do what my function does, but `&*std::ranges::find(output_array, ControllerOutput(button_it->second))` just looks cursed.

raphaelthegreat:
You can use projections in ranges to simplify it to
```
const auto it = std::ranges::find(output_array, &ControllerOutput::button, button_it->second);
return &*it;
```

src/input/input_handler.cpp
kalaposfos13: This can't be done, as it's not a const, temporary variables get updated in the stored outputs every input frame

src/input/input_handler.cpp
kalaposfos13: It's an offset so the SDL ranges for different types of inputs don't overlap (they all start at 0, and I have to differentiate between them somehow) Keyboard and mouse got the no offset range, as they were the first to be implemented, then for controller buttons, I just chose this as it didn't overlap with anything, and for axes I chose a large offset (0x80000000) so when sorting them they will be at the end of the list, and that is important for reasons. I think a cleaner way to do it would be to finally make this a struct and just give up on storing everything in a single u32 (id, input type, analog value for axes)

src/input/input_handler.cpp
kalaposfos13:
```cpp
line.erase(std::remove_if(line.begin(), line.end(),
                          [](unsigned char c) { return std::isspace(c); }),
           line.end());
```
Your code was syntactically incorrect, but this should do the same

GHU7924:
How close is this PR to being merged?

kalaposfos13:
Well I finally resolved everything in the review, hopefully nothing got broken along the way.

README.md
```diff
@@ -86,32 +86,34 @@ F12 | Trigger RenderDoc Capture
...
+| Pad Up      | W, LAlt / Mouse Wheel Up    |
+| Pad Down    | S, LAlt / Mouse Wheel Down  |
+| Pad Left    | A, LAlt / Mouse Wheel Left  |
+| Pad Right   | D, LAlt / Mouse Wheel Right |
+| L1          | Right Button, LShift        |
+| R1          | Left Button                 |
+| L2          | Right Button                |
+| R2          | Left Button, LShift         |
+| L3          | X                           |
+| R3          | Q / Middle Button          |
```
ngoquang2708, line 106: Missing a space.

lifeiscontent, line 106:
```suggestion
| R3          | Q / Middle Button           |
```

kalaposfos13:
Closed and will only be moving forward with the main version.

PR #1798, opened 2024-12-15
renderer_vulkan: add support for Polygon draws

polybiusproxy:
The Polygon primitive type works by having a specified \**N\** amount of vertices and having the GPU figure out how to fill said primitive—this primitive type is obsolete by now so modern rendering APIs do not support this primitive.

To emulate this, we can use the same approach for quad lists; generate an index buffer on the CPU and draw as triangle lists. Specifically for polygons, we can triangulate these by anchoring the first vertex of the polygon and connecting it with the current and next vertices.

imagen

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -235,24 +235,43 @@ bool BufferCache::BindVertexBuffers(
 }
 
 u32 BufferCache::BindIndexBuffer(bool& is_indexed, u32 index_offset) {
-    // Emulate QuadList primitive type with CPU made index buffer.
+    // Emulate QuadList and Polygon primitive types with CPU made index buffer.
     const auto& regs = liverpool->regs;
-    if (regs.primitive_type == AmdGpu::PrimitiveType::QuadList && !is_indexed) {
-        is_indexed = true;
+    if (!is_indexed) {
+        bool needs_index_buffer = false;
+        if (regs.primitive_type == AmdGpu::PrimitiveType::QuadList ||
+            regs.primitive_type == AmdGpu::PrimitiveType::Polygon) {
+            needs_index_buffer = true;
+        }
+
+        if (!needs_index_buffer) {
+            return regs.num_indices;
+        }
 
         // Emit indices.
         const u32 index_size = 3 * regs.num_indices;
         const auto [data, offset] = stream_buffer.Map(index_size);
-        Vulkan::LiverpoolToVK::EmitQuadToTriangleListIndices(data, regs.num_indices);
+
+        switch (regs.primitive_type) {
+        case AmdGpu::PrimitiveType::QuadList:
+            Vulkan::LiverpoolToVK::EmitQuadToTriangleListIndices(data, regs.num_indices);
+            break;
+        case AmdGpu::PrimitiveType::Polygon:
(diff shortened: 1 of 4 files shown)


src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -251,6 +251,20 @@ u32 BufferCache::BindIndexBuffer(bool& is_indexed, u32 index_offset) {
...
+        const u32 index_size = 3 * regs.num_indices;
+        const auto [data, offset] = stream_buffer.Map(index_size);
+        Vulkan::LiverpoolToVK::EmitPolygonToTriangleListIndices(data, regs.num_indices);
+        stream_buffer.Commit();
+
+        // Bind index buffer.
+        const auto cmdbuf = scheduler.CommandBuffer();
+        cmdbuf.bindIndexBuffer(stream_buffer.Handle(), offset, vk::IndexType::eUint16);
+        return index_size / sizeof(u16);
+    }
```
squidbus, line 275: Think you should combine this into the previous QuadList case since all of the code is the same except the `EmitPolygonToTriangleListIndices` call, then just do a switch on the primitive type to call the right emit function.

src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -288,6 +302,9 @@ u32 BufferCache::BindIndexBuffer(bool& is_indexed, u32 index_offset) {
         cmdbuf.bindIndexBuffer(stream_buffer.Handle(), offset, index_type);
         return new_index_size / index_size;
     }
+    if (regs.primitive_type == AmdGpu::PrimitiveType::Polygon) {
+        UNREACHABLE();
```
squidbus, line 311: Not sure if you want to tackle this now but I think it should be trivial to implement, basically translating an indexed draw is the same except instead of outputting `i` for example you would output `input_indices[i]`. See how `ConvertQuadToTriangleListIndices` is very similar to `EmitQuadToTriangleListIndices`.

squidbus:
Should also be added to this case btw since these don't work for DrawIndirect: https://github.com/shadps4-emu/shadPS4/blob/0fd1ab674bbb6b41f5cc6d46ca5ed4bcd6d6052c/src/video_core/renderer_vulkan/vk_rasterizer.cpp#L249

src/video_core/buffer_cache/buffer_cache.cpp
polybiusproxy: I'd rather wait until I can reach this unreachable as not many games use this primitive

src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -235,25 +235,33 @@ bool BufferCache::BindVertexBuffers(
...
+
+        switch (regs.primitive_type) {
+        case AmdGpu::PrimitiveType::QuadList:
+            Vulkan::LiverpoolToVK::EmitQuadToTriangleListIndices(data, regs.num_indices);
+            break;
+        case AmdGpu::PrimitiveType::Polygon:
+            Vulkan::LiverpoolToVK::EmitPolygonToTriangleListIndices(data, regs.num_indices);
+            break;
+        default:
+            return regs.num_indices;
```
squidbus: I think the way this is laid out there's a bit of waste doing the stream buffer map, would be nice if we could terminate earlier if we don't need to generate a buffer.

squidbus (approved):

PR #1801, opened 2024-12-16
ir: Add heuristic based LDS barrier pass

raphaelthegreat:
Sometimes shaders can use shared memory without appropriate barriers in ISA level. This is probably because compiler optimized them away as it found them not needed (local workgroup size 64 for example makes barriers not necessary). This adds a new IR pass that attempts to insert such barriers to avoid device loss issues and graphics artifacts on NVIDIA.

Inserting barriers right after data share write instructions is not possible as we must have the barrier be outside of the non-uniform conditional block. The process is a bit naive at the moment, but it involves walking the generated AST and for shaders that use shared memory it inserts barriers after all zero-depth divergent conditional blocks.

The zero-depth clause prevents insertion of barriers already inside non-uniform blocks as that can cause issues since not all threads are executing that block. A block is deemed non-uniform if its condition contains gl_LocalInvocationId, which is simple and effective for now, but by far not exhaustive

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -671,6 +671,7 @@ set(SHADER_RECOMPILER src/shader_recompiler/exception.h
                       src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
                       src/shader_recompiler/ir/passes/ring_access_elimination.cpp
                       src/shader_recompiler/ir/passes/shader_info_collection_pass.cpp
+                      src/shader_recompiler/ir/passes/shared_memory_barrier_pass.cpp
                       src/shader_recompiler/ir/passes/ssa_rewrite_pass.cpp
                       src/shader_recompiler/ir/abstract_syntax_list.h
                       src/shader_recompiler/ir/attribute.cpp
--- a/src/shader_recompiler/ir/passes/ir_passes.h
+++ b/src/shader_recompiler/ir/passes/ir_passes.h
@@ -6,6 +6,10 @@
 #include "shader_recompiler/ir/basic_block.h"
 #include "shader_recompiler/ir/program.h"
 
+namespace Shader {
+struct Profile;
+}
+
 namespace Shader::Optimization {
 
 void SsaRewritePass(IR::BlockList& program);
@@ -21,6 +25,6 @@ void RingAccessElimination(const IR::Program& program, const RuntimeInfo& runtim
 void TessellationPreprocess(IR::Program& program, RuntimeInfo& runtime_info);
 void HullShaderTransform(IR::Program& program, RuntimeInfo& runtime_info);
 void DomainShaderTransform(IR::Program& program, RuntimeInfo& runtime_info);
-void SharedMemoryBarrierPass(IR::Program& program);
+void SharedMemoryBarrierPass(IR::Program& program, const Profile& profile);
 
 } // namespace Shader::Optimization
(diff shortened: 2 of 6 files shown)


bigol83:
I checked with Hollow Knight, i have an Nvidia gpu and was suffering from device lost when the game tried to reach gameplay.
Unfortunately it still happens with this PR. 

```
[Debug] <Critical> vk_presenter.cpp:operator():768: Assertion Failed!
```

[shad_log.txt](https://github.com/user-attachments/files/18142858/shad_log.txt)

raphaelthegreat:
What does log have with validation enabled in config? At least this is expected since such simple game isn't expected to use LDS compute shaders

bigol83:
> What does log have with validation enabled in config? At least this is expected since such simple game isn't expected to use LDS compute shaders

Here is the log 
[shad_log.txt](https://github.com/user-attachments/files/18142888/shad_log.txt)

StevenMiller123:
Causes device lost errors in Kingdom Hearts 0.2 when run with an Intel Iris Xe iGPU. Doesn't appear to change the AMD + Windows exclusive device lost in this title, and I can't test on Nvidia at the moment.

[Current main.txt](https://github.com/user-attachments/files/18142982/Current.main.txt)
[This PR.txt](https://github.com/user-attachments/files/18143000/This.PR.txt)
[cdl_dump.zip](https://github.com/user-attachments/files/18143001/cdl_dump.zip)
(The diag log doesn't show anything useful, probably due to Intel drivers).

EDIT: Also occurs with Intel + Linux (Mesa) too.

GHU7924:
I have an Intel processor + Nvidia GPU.

shadPS4 v0.4.1 WIP main 8a4e032 (Pre-release)

CUSA03745 The Last Guardian

>[Debug] <Critical> vk_presenter.cpp:operator():768: Assertion Failed!

[shad_log.txt](https://github.com/user-attachments/files/18143306/shad_log.txt)

build c5f065b (ir: Add heuristic based LDS barrier pass)


[shad_log.zip](https://github.com/user-attachments/files/18143432/shad_log.zip)

build ba82e3a (ir: Add heuristic based LDS barrier pass)



Further, if you skip the cut scene, the game does not load and ends without errors.

[shad_log.zip](https://github.com/user-attachments/files/18143359/shad_log.zip)

EDIT: 

>What does log have with validation enabled in config?

I do not know if I understood you correctly, but if I check the Enable Vulkan Validation Layers box, then I have an error in both Main and your PR.

[shad_log.txt](https://github.com/user-attachments/files/18149751/shad_log.txt)

bigol83:
> I do not know if I understood you correctly, but if I check the Enable Vulkan Validation Layers box, then I have an error in both Main and your PR.
> 
> [shad_log.txt](https://github.com/user-attachments/files/18149751/shad_log.txt)

Do you have Vulkan SDK? You need that to use Vulkan Validation Layers

GHU7924:
>Do you have Vulkan SDK? You need that to use Vulkan Validation Layers

Sorry, no.

I have now installed and tested it.

This is an option where I chose a new game and skipped the cutscene.

[shad_log.zip](https://github.com/user-attachments/files/18152029/shad_log.zip)

Next

This is an option when I chose a new game and did not miss the cutscene, even got into the game.
shadPS4 v0 4 1 WIP lds-barr c5f065b _ CUSA03745 - The Last Guardian™ _01 03_ Пн, 16 декабря 2024 г  17_38_46

I pressed the buttons, the number of these characters on the screen decreased, but then the game closed anyway.

[shad_log.zip](https://github.com/user-attachments/files/18152041/shad_log.zip)

P.S. I hope I've done everything right now.

PR #1804, opened 2024-12-16
style: add rounded borders and focus color styling to the search bar

richardhbtz:
before:
KzIXI35
after:
god0oGg
on focus:
aIAhV8F


Randomuser8219:
Looks a lot better than the older search bar.

PR #1805, opened 2024-12-16
Implement gamma correction in main config menu

rainmakerv3:
This PR adds a setting for Gamma correction in the settings dialog Graphics tab. While there is a gamma correction tool in the advanced debug menu, this one is intended is more accessible for regular users and the setting can also be saved (the devtools gamma setting is not currently saved).

Useful in the short term while the PR for gamma is being sorted out, in the long term also may be useful for games lacking a native Brightness setting

image

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -64,6 +64,7 @@ static bool vkCrashDiagnostic = false;
 static s16 cursorState = HideCursorState::Idle;
 static int cursorHideTimeout = 5; // 5 seconds (default)
 static bool separateupdatefolder = false;
+static int GammaValue = 1000;
 
 // Gui
 std::vector<std::filesystem::path> settings_install_dirs = {};
@@ -224,6 +225,10 @@ bool getSeparateUpdateEnabled() {
     return separateupdatefolder;
 }
 
+int getGammaValue() {
+    return GammaValue;
+}
+
 void setGpuId(s32 selectedGpuId) {
     gpuId = selectedGpuId;
 }
@@ -342,6 +347,10 @@ void setSpecialPadClass(int type) {
 
 void setSeparateUpdateEnabled(bool use) {
     separateupdatefolder = use;
+}
+
+void setGammaValue(int value) {
+    GammaValue = value;
 }
 
 void setMainWindowGeometry(u32 x, u32 y, u32 w, u32 h) {
@@ -566,6 +575,7 @@ void load(const std::filesystem::path& path) {
         shouldDumpShaders = toml::find_or<bool>(gpu, "dumpShaders", false);
         shouldPatchShaders = toml::find_or<bool>(gpu, "patchShaders", true);
         vblankDivider = toml::find_or<int>(gpu, "vblankDivider", 1);
+        GammaValue = toml::find_or<int>(gpu, "GammaValue", 1000);
     }
 
     if (data.contains("Vulkan")) {
@@ -668,6 +678,7 @@ void save(const std::filesystem::path& path) {
     data["GPU"]["dumpShaders"] = shouldDumpShaders;
     data["GPU"]["patchShaders"] = shouldPatchShaders;
     data["GPU"]["vblankDivider"] = vblankDivider;
+    data["GPU"]["GammaValue"] = GammaValue;
     data["Vulkan"]["gpuId"] = gpuId;
     data["Vulkan"]["validation"] = vkValidation;
     data["Vulkan"]["validation_sync"] = vkValidationSync;
@@ -775,6 +786,7 @@ void setDefaultValues() {
     m_language = 1;
     gpuId = -1;
     separateupdatefolder = false;
+    GammaValue = 1000;
 }
 
 } // namespace Config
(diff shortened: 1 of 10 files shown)


friyin:
It would be a good idea having a numeric field along the slider and clarify what side makes it brighter and darker.

rainmakerv3:
image

Sounds good, this is what I have now

Foul-Tarnished:
The default value should be at the center, and display 1.00

rainmakerv3:
Forgot to rename widgets from default names, will set as draft to clean up, but please feel free to comment on things other than widget naming

rainmakerv3:
- used decimal form for gamma label (ex. 1.234 instead of 1234), but I didn't label the default, I feel it's cluttered and unsure if needed
- properly named widgets and layouts
- fixed not saving on close as I did for the other config variables in a previous PR

GHU7924:
I'll add a few options, just in case you decide to finalize it.

EDIT: I meant the scale 0-100%, the middle = 50%, I did not notice that I wrote 0% instead of 50%.

rainmakerv3:
Whats the intent for the percentage boxes? What effect should they have?

GHU7924:
> Whats the intent for the percentage boxes? What effect should they have?

The current value, only as a percentage. When I saw the value 1533, I personally did not perceive it in any way, I immediately have a question - "What is it?" Perhaps interest is also an unsuccessful offer, I agree. Then it is worth making the values 1.0 or 1.00, etc.

Percentage boxes... This way, the value stands out more than just the numbers on the background.

rainmakerv3:
> > Whats the intent for the percentage boxes? What effect should they have?
> 
> The current value, only as a percentage. When I saw the value 1533, I personally did not perceive it in any way, I immediately have a question - "What is it?" Perhaps interest is also an unsuccessful offer, I agree. Then it is worth making the values 1.0 or 1.00, etc.
> 
> Percentage boxes... This way, the value stands out more than just the numbers on the background.

I believe gamma is usually represented in decimal values though, and I did change this in the previous commit, as such. This also makes it consistent with the values on the devtool widget. 

Will leave it like this for now pending review

image

Foul-Tarnished:
For end user it makes more sense to have 100% as the default

So if they want 20% brughter they'll go to 120%

Probably do 0-200% ?

Le mar. 17 déc. 2024 à 03:30, rainmakerv2 ***@***.***> a
écrit :

MajorP93:
@Foul-Tarnished 
In my opinion in this case -100% to 100% would make more sense.
Default value should be 0% in that case.
I think that would be more intuitive.
If a user wants to increase brightness by 20% they select gamma 20%.
If they want to make the image 20% darker they select -20% and so on.

GHU7924:
@Foul-Tarnished @MajorP93 In fact, this is not so important, in any case, one side is darker and the other is lighter.

@rainmakerv3 I just opened the NVIDIA Control Panel.

Values from and to: 0,30 - 2,80 and 1,00 as default.
NCP-Gamma

You can do the same.

EDIT: For example
Gamma

DemoJameson:
I think it's more customary to have the left side darker and the right side brighter

rainmakerv3:
> I think it's more customary to have the left side darker and the right side brighter

This is true, but if I read the code right, it's the system call itself that treats it the opposite way (having larger values be darker). My thought is that I want to stay as consistent as I can with current code (i.e. devtools widget, which treats it this way) and still be intuitive for users. And since lighter/darker is labeled anyway. I foresee no usability issues with this whatsoever in its current state.

Guys, I think everyone will have very different opinions on UI, that's always the case, but I think the remaining comments are getting very subjective in terms of whats better/worse, so I'm inclined to leave this as is, and maybe the reviewer can just have the final say.

rainmakerv3:
Added the following description text on hovering the mouse, but only for English

image

Hermiten:
Hello,
Where are we with this implementation?
Thanks you :)

rainmakerv3:
this is actually ready for review, no other changes are planned for this

polybiusproxy:
I think gamma should not be forcefully changed, they're only in a debug menu for reasons. Rather it should be configured via the game's settings.

rainmakerv3:
The possible benefits are adjustment for games without native brightness settings or more granular control over them. I cannot say though what the risks are of having gamma exposed to the end user, so if you feel this is detrimental, I understand and will close the PR

rafael-57:
> The possible benefits are adjustment for games without native brightness settings or more granular control over them. I cannot say though what the risks are of having gamma exposed to the end user, so if you feel this is detrimental, I understand and will close the PR

I agree, it's a useful feature since not all games have the slider. There is also a reset button to set it to default so I don't see any issues with this feature personally

GHU7924:
To be honest, I do not know if this PR is necessary. 
On the one hand, this is useful, on the other hand, it will be faster to get to the debugging menu, where you can change the gamma value during the game, immediately viewing the changes.
But even in the Debug menu, this setting is incomplete.
There is no return to the default value. It is not possible to set a value, but only to move the slider.

In addition, this option greatly affects the image, for example:

Default


Other values



It may be too early to implement such an option at this stage of development.
Or it is better to improve this option in the debugging menu and add information about this feature to the Wiki (how to change the gamma using an emulator for games without the built-in gamma change option).

georgemoralis:
closing as it will probably create more issues than solving

PR #1809, opened 2024-12-17
fs: Fix wrong mounts being matched by partial guest path.

squidbus:
Currently when finding the mount for a guest path, if one of the mounts is a prefix of another mount, the wrong mount may be matched. For example, if both `/addcont3` and `/addcont30` are mounted, the path `/addcont30/file` may match with `/addcont3` despite being the wrong mount.

To fix this:
* Added sanitization to make sure the `mount` guest path has no trailing slash, to help with the following change.
* Changed the mount find condition to check if either the paths are equal (path is exactly the mount path) or path starts with the mount path plus a trailing slash (either exactly the mount path with trailing slash or a sub-path of the mount).

Fixes loading DLC songs in CUSA12636 and CUSA12380.

### DIFF
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -10,16 +10,28 @@
 
 namespace Core::FileSys {
 
+std::string RemoveTrailingSlashes(const std::string& path) {
+    // Remove trailing slashes to make comparisons simpler.
+    std::string path_sanitized = path;
+    while (path_sanitized.ends_with("/")) {
+        path_sanitized.pop_back();
+    }
+    return path_sanitized;
+}
+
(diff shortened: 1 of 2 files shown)


Foul-Tarnished:
Imagine if this fix BB sound issue when loading after a crash
(loading after the message saying you didn't quit properly)

This is probaably a fs issue right ?

squidbus:
> Imagine if this fix BB sound issue when loading after a crash (loading after the message saying you didn't quit properly)

No idea but not likely, this is more likely to just fix problems with DLC.

Missake212:
> Imagine if this fix BB sound issue when loading after a crash (loading after the message saying you didn't quit properly)
> 
> This is probably a fs issue right ?

Can confirm it doesn't fix it.

PR #1810, opened 2024-12-17
vk_instance: Add additional fallback for missing D16UnormS8Uint.

squidbus:
For depth-stencil, if neither `D16UnormS8Uint` nor `D24UnormS8Uint` are supported, fall back to `D32SfloatS8Uint` as a last resort and hope the game doesn't depend on the interpretation too much.


squidbus:
Not sure what happened with the Windows CI but it looks like a GitHub issue.

PR #1819, opened 2024-12-18
shader_recompiler: Rework image read/write emit.

squidbus:
* Refactor ImageFetch and ImageRead IR together.
* Handle cube case and better handle mip case for ImageRead instruction.
  * For cube case, need to force storage image to use OpImageRead.
* Handle multisampling case for ImageWrite instruction.


squidbus:
Slightly simplified the texture cache changes.

squidbus:
Rebased on latest main

raphaelthegreat (approved):
Good work! This should fix Driveclub device loss on NVIDIA

PR #1822, opened 2024-12-18
Handle munmap over multiple VMAs

mikusp:
Ratchet&Clank CUSA01073 allocates multiple memory areas and then tries to unmap them in one go. According to the documentation of munmap that should be allowed.

The PR merges the VMA containing the starting address with the next one until the area reaches at least the size to unmap, and then performs the unmap as before.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -380,28 +380,26 @@ s32 MemoryManager::UnmapMemory(VAddr virtual_addr, size_t size) {
     return UnmapMemoryImpl(virtual_addr, size);
 }
 
-s32 MemoryManager::UnmapMemoryImpl(VAddr virtual_addr, size_t size) {
-    const auto it = FindVMA(virtual_addr);
-    const auto& vma_base = it->second;
-    ASSERT_MSG(vma_base.Contains(virtual_addr, size),
-               "Existing mapping does not contain requested unmap range");
-
+size_t MemoryManager::UnmapBytesFromEntry(VAddr virtual_addr, VirtualMemoryArea vma_base,
+                                          size_t size) {
     const auto vma_base_addr = vma_base.base;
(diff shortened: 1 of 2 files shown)


raphaelthegreat:
Merging VMAs before unmap is not accurate to freebsd unfortunately and its not guaranteed to succeed unless the mapping are contiguous in physical memory as well. In addition I'm not sure if windows placeholder mechanism allows you to call VirtualFree on multiple allocations.

I recommend to follow the freebsd logic instead https://github.com/freebsd/freebsd-src/blob/aa5d3dec9df9379ab610f426e9653d1a55b254e5/sys/vm/vm_map.c#L4031
Retrieve the first VMA iterator with FindVma and iterate entries to unmap. The existing logic can be moved to an UnmapEntry function similar to freebsd that works only on a single VMA

mikusp:
Thanks for the suggestion, I'll make the change

ElBread3:
With many games it instead crashes at a `[Debug] <Critical> memory.cpp:operator():643: Assertion Failed!
Adding a mapping to already mapped region` error. Don't know if this is progress or an issue with the implementation though.

mikusp:
@raphaelthegreat @ElBread3 I've pushed a different implementation

brad0demx:
Would it be possible for this to get a rebase? [CUSA05485](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/1678) boots using this PR and I heard the splash screen should be improved on 0.5.X

StevenMiller123:
@mikusp Could you rebase this? I made some changes to UnmapMemoryImpl in a recent PR that conflict with this.

squidbus:
Thank you, this has been rebased in https://github.com/shadps4-emu/shadPS4/pull/2233 and we will likely move forward with it there.

PR #1825, opened 2024-12-18
Handle more 64bit shifts in Translator

mikusp:


squidbus (approved):
This looks fine, I'll probably make a few touch-ups to V_LSHL_B64 though as a follow up.

PR #1826, opened 2024-12-18
Handle setting Vcc in Translator::SetDst64

mikusp:
I know 64bit instructions are kind of broken right now, if that is not correct let me know in the comments


raphaelthegreat:
I think this looks okay

PR #1828, opened 2024-12-19
style: Add options for system colors and system Qt widget styles

tomboylover93:
Taken from https://github.com/tomboylover93/shadPS4/releases/tag/v0.4.1_WIP_qt-style:

- Added "System (Light)" and "System (Dark)" options under View -> Themes. Since I couldn't figure out how to apply different icon colors based on whether or not your system theme is light or dark, they are separate options, so you should use them according to your system colors. Tested on Linux with KDE Plasma. Untested on Windows and macOS, but might be useful on Windows if you use custom system themes.
- Added "Widget Style" options on the Settings menu, with text that shows up on mouse hover. Defaults to "Fusion", as it is currently hardcoded to do so. "System" uses your system's default Qt widget style (example: Breeze, Oxygen, MS Windows 9x, Kvantum). Linux users can override this with the `QT_STYLE_OVERRIDE` environment variable, as long as it's set to "System". You can tweak this on config.toml with the line `widgetStyle = "Fusion"`. Since I couldn't figure out how to repaint all widgets after clicking on either "Save" or "Apply", this doesn't reflect in real time and you need to close and reopen shadPS4. Tested on Linux with KDE Plasma. Untested on Windows and macOS.

### Screenshots (taken on Linux with KDE Plasma):
- "System (Dark)" color theme, "Fusion" widget style

Screenshot_20241219_114710

- "System (Light)" color theme, "Fusion" widget style

Screenshot_20241219_114840

- "Dark" color theme, "System" widget style (set to Breeze in KDE System Settings)

Screenshot_20241219_115140

- "System (Dark)" color theme, "System" widget style (set to [Bali10050/Lightly](https://github.com/Bali10050/Lightly) in KDE System Settings)

Screenshot_20241219_115317

### Additional screenshots
- Settings menu

Screenshot_20241219_132609

This is purely cosmetical and should not affect the usability of the emulator in any way, but if you do have issues with it, such as the emulator not launching or the color/widget style changes not working, please let me know.


rainmakerv3:
For checking system dark mode, you can try 

if QGuiApplication::styleHints()->colorScheme() == Qt::ColorScheme::Dark

for refreshing widgets after a save/apply, i think you can have update->widget every widget and then QCoreApplication::processEvents after clicking apply/save (or if i remember right i put the same function for both, if so that would work better).

also i still think the widget style selection is better placed over/under where the theme selection is

tomboylover93:
I tried both of these. The first one broke all themes and their icon colors and the second has no effect. Unless I missed something?

As for placing the widget style selection on the View menu, I tried that before I decided to make a group box in the settings menu for it, and the emulator segfaulted immediately, so I went with the latter.

rainmakerv3:
It's very hard to discuss this without seeing how you tried to code it, and it's faster to do it myself. You can refer to it here:

https://github.com/rainmakerv3/shadPS4/commits/refs/heads/systemui/

seems working as expected on windows, sample vid:

https://github.com/user-attachments/assets/bf78c308-f161-4a3b-961a-0b0813b4cbb4

I used an old workaround for linux dark mode detection since apparently qt:colorscheme::dark does not work on linux, but I do not have a linux system to test.

I don't think widget style in the view menu can cause segfaults if coded properly, but I will leave this for now and can give my suggestions if/when the reviewer decides its necessary to move it to view menu.

tomboylover93:
My approach for widget updating was through a separate function, something like this:
```
void SettingsDialog::UpdateAllWidgets(const QList<QWidget*>& widgets) {
    for (auto widget : widgets) {
        widget->update();

        if (auto layout = widget->layout()) {
            QList<QWidget*> childWidgets;

            for (int i = 0; i < layout->count(); ++i) {
                QWidget* childWidget = layout->itemAt(i)->widget();
                if (childWidget) {
                    childWidgets.append(childWidget);
                }
            }

            UpdateAllWidgets(childWidgets);
        }
    }
}
```
I tried 2 other methods that were similar to this and they didn't work. I might've misread the documentation for QWidget::update.

In any case, your approach is cleaner and actually works so I'll merge it into my branch. I'll give another try at a single System color theme now that I have a reference for how to detect system dark or light mode.

rainmakerv3:
Sorry, i should have clarified that my code was not cleaned up yet, there are some things I would change if I were to finalize it (removing the prinft/log include, putting the extern in a namespace, removing processevents if unneeded, general cleanup) that I did not spend the time to do

tomboylover93:
That's fine, if you do make these changes I'll add them to this PR too.

By the way, if you have any ideas for how to set the icon color automatically, let me know. I haven't figured that out yet and that's something I need to do before calling this ready for review.

rainmakerv3:
tbh i wasn't planning on cleaning it up, just giving you a heads up cause it probably needs to be done hehe

to catch a system pallete change such as changing system theme and darkmode and change icon color every time thats done, you run the same kind of setuiicons code (or maybe it should have its own function at that point since there are a number of lines in it) on some event filter listening to event->type() == QEvent::ApplicationPaletteChange

GHU7924:
In my opinion, "Widget styles" should be in section "GUI Settings", but this may interfere with PR #1834.

tomboylover93:
I also think that it should be in "GUI Settings". I could place it under the volume slider so that it doesn't interfere with that PR's changes.

GHU7924:
The problem is, I'm not sure if all of this can be placed the way you say. In this case, you will have to redo the item related to the "title music", then you can say that everything is possible to accommodate.


I don't think anyone is going to redo this right now.

tomboylover93:
I don't know if I misread your reply, but this didn't require me to redo anything.

Screenshot_20241221_103603

GHU7924:
The difference is that if PR 1834 is merged (look, it's located above "Play title music"), then you probably won't have enough space.

tomboylover93:
The easiest way to avoid that is to put the widget selection above these buttons. Like so:

Screenshot_20241221_104920

GHU7924:
I want to say that you need to take into account the size of the window, that is, whether it will remain as compact as it is now, or whether it will have to be increased by default so that both your PR and that PR fit harmoniously.

rainmakerv3:
pushed the commit here to demonstrate the automatic coloring on system palette change, as discussed https://github.com/rainmakerv3/shadPS4/commits/refs/heads/systemui/ again though, I am not really able to take the time to clean up the code or check if theres a better way to implement it within the context of the themes code that I am not that familiar with, this is just to show how it works

tomboylover93:
This is actually what I was thinking of, I just didn't know how to implement it properly. The only downside I've noticed is that you have to start shadPS4 with your system theme on dark mode and then change it to light mode, because if you start it on light mode the icons don't render, but I don't know if this happens on Windows or macOS.

GHU7924:
build 4514dd1 (Windows 11 24H2)



P.S. Increase the width of the Settings window by about 1.5 cm so that the bars align slightly (System and Update).

tomboylover93:
I will. On another note, did the icons render properly when opening it with the Windows theme set to light mode?

GHU7924:
Current Widget Style = System 

I immediately showed you what my dark theme looks like.
Although there were no visible points, I just switched from dark to light from memory (View - Themes).
There are also problems with other topics.


In fact, the inscriptions are just very faintly visible, for example, theme Gruvbox, it is best visible on it.

tomboylover93:
I can't replicate this on my end. 
image
image

This could probably be fixed with `themePalette.setColor(QPalette::ButtonText, Qt::white);`, but that would break light mode.

GHU7924:
I switched from System to Fusion.



And after restarting, the theme looked fine.

GHU7924:
I deleted the messages because maybe I was doing the wrong thing.

Another such moment. You probably don't have a topic separation or anything like that. I chose the system version, but Fusion will be in the emulator settings. Although if I chose the system one, it should change in the settings, as I think.


That is, I have to choose a theme in two places, I think it shouldn't be like that.

Dark, Light, Green, Blue, Violet, Gruvbox = Fusion.
System = System (changes in the operating system).

Maybe I'm wrong, then it would be great if someone else tested and expressed their opinion.

tomboylover93:
I made them separate settings in separate places because:
1. I don't know how to add drop-down menus on the toolbar.
2. I made this with Linux in mind. Since Linux desktops let you have more granular control with your themes I thought it would be good to give the end user that exact level of control as opposed to what other Qt apps do (example: PCSX2, which has a single "Native" theme that combines both the system color and widget style options).

Although you aren't the first person to suggest something like that, so if a reviewer thinks it's best to combine them, I will.

As for the text issues on dark mode, I asked someone who also uses Windows and they didn't have that issue.

GHU7924:
When I selected System both ways, and then changed it in the operating system itself, then I was able to get this.


>As for the text issues on dark mode, I asked someone who also uses Windows and they didn't have that issue.

I'm sorry, I must have done something wrong.

GHU7924:
@tomboylover93 I figured out what was wrong.

Windows 11 has a color mode where you can choose dark or light. If you choose a dark theme and choose a light theme in the emulator, then the display will be incorrect. If you select a light theme and select a dark theme in the emulator, then the display will also be incorrect. 


And if you select the system theme in the emulator in both places and then change the theme in the operating system, then everything is fine.

rainmakerv3:
Yes,

> @tomboylover93 I figured out what was wrong.
> 
> Windows 11 has a color mode where you can choose dark or light. If you choose a dark theme and choose a light theme in the emulator, then the display will be incorrect. If you select a light theme and select a dark theme in the emulator, then the display will also be incorrect. 
> 
> And if you select the system theme in the emulator in both places and then change the theme in the operating system, then everything is fine.

This is probably the reason why system theme settings are usually packaged together

tomboylover93:
Correct me if I'm wrong, but using the system's colors while using a widget style that isn't the system default shouldn't cause issues like this? During my time using Windows I've used many Qt apps that use the Fusion style and I've never come across such an issue with either light or dark mode.

The only thing I can think of that would cause this is the "Custom" option on Windows' theme settings, which lets you choose between light and dark mode for system apps and non-system apps, respectively.

rainmakerv3:
The issue is the combination of system widgets with themes that enforce conflicting text colors. you can see it in the video I have above

tomboylover93:
To my knowledge, the Gruvbox theme is the only one that enforces a specific text color, every other theme uses `themePalette.setColor(QPalette::ButtonText, Qt::white);`. I'll try changing that to see if this issue stops happening.

rainmakerv3:
That would be forcing the text to white yes, most of the themes are darker background

tomboylover93:
Pushed it to my [qt-style-test](https://github.com/tomboylover93/shadPS4/tree/qt-style-test) branch. Can someone try [this action?](https://github.com/tomboylover93/shadPS4/actions/runs/12458469561)

GHU7924:
@tomboylover93 The problem is that if the system theme is selected in GUI Settings, then everything breaks. That's where the problem lies. This is where the distortion occurs due to the color mismatch. With Fusion, all 7 themes seem to be working fine.
Or at least I didn't notice any obvious problems if we are launching the emulator for the first time.

The fact is that Windows has always had a light theme.
Then at one point they added a dark one, which I wrote about above, this setting affects the explorer.
Now you can make it dark.

I see a possible solution in the following:
You just need to make sure that when you select a system theme in the GUI settings, 
the system theme (View - Themes) was also selected by default, 
and selecting other themes from this list would be prohibited.

It seems that if you select a system theme in the GUI settings, and then select one of the themes (View - Themes), then switch to Fusion and (View - Themes) select a light theme, then distortion also occurs.

GHU7924:
> Поместил его в свою [ ветку qt-style-test ](https://github.com/tomboylover93/shadPS4/tree/qt-style-test) . Может кто-нибудь попробовать [ это действие? ](https://github.com/tomboylover93/shadPS4/actions/runs/12458469561)

GUI Settings - Current Widget Style = System
View - Themes = Light
OS color mode = Dark



GUI Settings - Current Widget Style = System
View - Themes = Dark
OS color mode = Light

tomboylover93:
Just to check, does it work as intended if the widget and theme are set to System and with the Windows color mode set to either dark or light?

tomboylover93:
This is a strange case. Either I combine the System color theme and widget style into one (which I haven't figured out how to) or I grey out the System option on Windows, which neither is ideal, but the latter is better than the former as it is currently hardcoded to use Fusion because most of the UI was built around it, and Windows has the largest userbase out of all the 3 OSes supported by shadPS4.

Besides that, I also need to know how this works in macOS.

GHU7924:
I think that somehow you need to remove the "Not OK" option, leaving only both options OK.

tomboylover93:
I'd have to remove every theme other than System, which means this probably won't get merged. Unless I reworked the theme system to use the Fusion widget for other color themes, like what PCSX2 and Duckstation do.

GHU7924:
@tomboylover93 Look at PR 1855, something similar was done there. Yes, PR is completely different, (but if there are items in the menu, you cannot select them, because there is no checkbox in the checkbox). Maybe this can be somehow transferred to the Current Widget Style selection.
Something like, when the system theme is selected, you will not be able to select other themes other than the system theme.

tomboylover93:
Seems like a good alternative, I'll look into it.

GHU7924:
>Besides that, I also need to know how this works in macOS.

@C4ndyF1sh , you can help @tomboylover93  test this PR on MacOS?

C4ndyF1sh:
@GHU7924 
Works on Mac Screenshot 2024-12-27 at 16 56 19

C4ndyF1sh:
@GHU7924 I switched to light mode but shadps4 doesn't, I even restarted shad (Tried with system theme on and off) Screenshot 2024-12-27 at 16 59 20

GHU7924:
@C4ndyF1sh In General, try Current Widget Style = System.

You can see our comments above and compare them with your experience to see if you have similar problems.

Thank you for responding. 

P.S. I haven't tested the new build yet, I'll do a test in the near future.

GHU7924:
@tomboylover93 Perhaps it would be better to replace the style selection with Radio buttons?


Fusion - all themes are available.
System = System + System and click blocking "Themes".

tomboylover93:
I could do that but I don't know how to combine both Qt themes and color themes, nor how to block "Themes" from being clicked when selecting System.

GHU7924:
@tomboylover93 Then you should mark your PR as a draft.

Perhaps @f8ith or @DanielSvoboda they will give you some tips, or maybe not.
(The description of the problem begins with comments from about 5-7 days ago to today).

tomboylover93:
> I switched to light mode but shadps4 doesn't

@C4ndyF1sh Can you try the latest commit?

tomboylover93:
I forgot to put this here, but I did some testing with a Windows 10 virtual machine (I had to use Mesa's 'lavapipe' Vulkan drivers to get it to open, because it's a virtual machine and shadPS4 requires Vulkan 1.3 support).

Here's what it looks like with the Windows color theme set to "Dark", the app's color theme set to "System" and the widget style set to "Fusion":

Screenshot_20250107_122700

And with the widget style set to "System":

Screenshot_20250107_122709

In the 2 weeks since my last commit to this PR I have not been able to figure out the problems with Windows and the "System" widget style, and they only work fine with "Fusion" which is what the app is hardcoded to use. I think I'll gray out the widget style option on Windows, and possibly macOS.

rharish101:
I'm running Arch Linux with Kvantum (not KDE) to set my Qt theme for Qt apps (using `QT_QPA_PLATFORMTHEME=qt6ct`, and 'kvantum-dark' as the style in the Qt config utility). With this branch, when I launch shadPS4 after setting the widget style to 'System', all widgets use my system style, except for the background, which is in black:
shadps4-theme-before

However, if I open the settings and click on Save without making a single change, then the background changes to the correct system style:
shadps4-theme-after

tomboylover93:
I believe this has something to do with widget repainting when you click on "Save" or "Apply". I don't know why the background looks like that when launching the app, it looks just fine to me.
image

rharish101:
Is it possibly a Wayland issue? I'm running Hyprland.

tomboylover93:
I use a KDE Plasma Wayland session and don't have this issue, maybe it's something to do with Kvantum or qt6ct.

tomboylover93:
@GHU7924 @C4ndyF1sh Can you try the latest commit on Windows and macOS? From what I've seen (and tested) I'll have to disable widget style selection for Windows and disable the "System" color theme for macOS since it's broken with light mode, and I'm trying to see if there are any other options (I'm not very experienced with Qt for platforms other than Linux).

GHU7924:
I confirm that in Windows, the system theme is not selected in the emulator settings, but only Fusion. 
The System is reset to Fusion. Since System only works on Linux, it might be worth adding a label, such as System (Linux only). Also, as I understand it, there is no description of the element.

Qt1

Depending on the color mode, the system theme adjusts and switches between Dark and Light.
I also noticed this detail, when choosing a dark theme, shades are applied,
but for some reason they are not when choosing a light one.

Qt4

You should also update your PR first to apply the new changes that were made to PR #2213.
Now you have to transfer the elements of your PR to the GUI tab.

tomboylover93:
I merged it with main (checks are failing for a different reason), and I'll change the text for the System option later. I don't see any missing text, though. 
image

GHU7924:
@tomboylover93 I honestly don't know what the problem is, but I have it that way.

https://github.com/user-attachments/assets/dd2552fc-78bd-480d-8f77-8ebe8621e378

One more thing
Differences

And in my opinion, it's better to do this.

DanielSvoboda:
Putting the volume bar next to the name also seems like a good idea to me

image

GHU7924:
@DanielSvoboda I agree, but I suggested doing this: https://github.com/shadps4-emu/shadPS4/pull/1828#issuecomment-2558121231 (The volume icon would change every 33%)
But then there's more work to be done.

tomboylover93:
Does this look good?

image

GHU7924:
>Does this look good?

Don't worry, there may still be a few more edits in the future as more improvements, fixes, and new features are added ))

tomboylover93:
@DanielSvoboda Do you happen to know how I can grey out a QComboBox element based on what operating system the app was compiled for?

GHU7924:
This is what my Windows window looks like at the moment.
изображение

The problem of the light theme remains.
Qt4

@tomboylover93 At the next PR update, can you make the edit that @DanielSvoboda  suggested?
Vol

tomboylover93:
Like this?

image

I'm aware of the spacing between "Volume" and the volume slider but nothing I've tried had an effect on it.

rainmakerv3:
> @DanielSvoboda Do you happen to know how I can grey out a QComboBox element based on what operating system the app was compiled for?

I think it would be better to just ->hide() the groupbox entirely, and hardcode to fusion. I believe I showed you an example of #ifdef previously here, thats for selectively applying code to certain definitions (such as OS)

tomboylover93:
What I'm doing currently is using `#if defined(__linux__) || defined(__APPLE__)` so that the widget style selection is disabled on Windows (you can still see and interact with the widget however). I don't know how I could adapt this to `#if defined(Q_OS_WIN)`.

GHU7924:
@tomboylover93 To be honest, I hate to say this, but maybe you should close your PR or postpone it in order to return to it in the future as a new PR. You tried very hard, but it turned out that the systems are very different, and we can't put everything together. I really want to help you, but since I don't have any programming knowledge, unfortunately I can't offer you solutions.

I tested your latest PR and I also have a large distance between the label and the slider.

My last idea is to close this PR and open a new one containing all the changes at the moment from Main.
Add only this option, which, depending on the operating system, will use Fusion or System, without the code for the GUI tab, as @rainmakerv3  said.
изображение

Windows = System + Fusion
mac OS = System + Fusion
Linux = System only (System + System)

tomboylover93:
If anyone with more Qt experience than me (and especially for more than one platform, as I've learned that writing the best code you could write for the platform that you yourself use is not a good idea) is willing to do it, then that's fine. That said, I think the Linux part is pretty much done, so anyone could just use this as a base while improving it on the Windows and macOS side.

GHU7924:
I think that if someone wants to continue your work, it will probably be easier for them to do everything from scratch than to fix existing code. This way the author will have more control over the code and he will know what is implemented and what is not.
I think it would be better for you to close the PR.

@tomboylover93 No negativity towards you, you did well. If you will be doing any PR in the future and you need help with testing, then do not hesitate to contact me, I will be happy to help you (if my knowledge is enough to give you good feedback).

tomboylover93:
I think I'll do one last rebase to make sure it's in a more usable state (doesn't really do anything because this repo has new PRs and commits every day, but why not) and move it to an archive branch and then close this PR. Then I'll open one just for the widgets part and fix the issues with Windows (as I could confirm from other people that the system widgets work as they should on macOS). That should be simple to do. The color themes part, and specifically the issue you described where the accent color doesn't work with the Windows light theme, is not something I could fix considering my knowledge (or lack thereof) of how Qt works on other platforms.

DanielSvoboda:
> @DanielSvobodaVocê sabe como posso tornar um elemento QComboBox cinza com base no sistema operacional para o qual o aplicativo foi compilado?

I don't know


about the spacing between the 'volume' and the bar, you need to change this: label_Volume
image

PR #1829, opened 2024-12-19
gui: start the emulator in new thread

### LINKED ISSUE #882: [Linux] Once game starts, main UI stops responding
Once you launch the game, you can no longer interact with the main window. It stops painting. You can move it around, but that's it. If I hit the close button, it doesn't respond, and KDE asks if I want to terminate the app after a few seconds.

This makes it a little annoying to close the emulator, but it also makes it impossible right now to use the cheats feature.

Other people in discord seemed to have the same problem. This happens both if I build from source or if I run a recent AppImage from Actions (usually of someone else's fork or in a branch, because of problems with `new PthreadInternal` under Linux - but the problem still occurs exactly the same way, whether I apply that patch or not).

ngoquang2708:
Fix Qt GUI not responding in Linux.

Fix: https://github.com/shadps4-emu/shadPS4/issues/882.

### DIFF
--- a/src/qt_gui/main_window.cpp
+++ b/src/qt_gui/main_window.cpp
@@ -554,7 +554,6 @@ void MainWindow::CreateConnects() {
 }
 
 void MainWindow::StartGame() {
-    isGameRunning = true;
     BackgroundMusicPlayer::getInstance().stopMusic();
     QString gamePath = "";
     int table_mode = Config::getTableMode();
@@ -577,13 +576,12 @@ void MainWindow::StartGame() {
     }
     if (gamePath != "") {
         AddRecentFiles(gamePath);
-        Core::Emulator emulator;
         const auto path = Common::FS::PathFromQString(gamePath);
         if (!std::filesystem::exists(path)) {
             QMessageBox::critical(nullptr, tr("Run Game"), QString(tr("Eboot.bin file not found")));
             return;
         }
-        emulator.Run(path);
+        StartEmulator(path);
     }
 }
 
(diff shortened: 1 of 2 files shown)


ngoquang2708:
@DanielSvoboda Can you help with the Qt window crash when resizing.

GHU7924:
@ngoquang2708 There is a moment when the active window overlaps the previous one, just like now, when a PR with a status column was added, an empty rectangle appears, which freezes when closed and you have to stop the emulator.
In addition, when installing *.pkg, the progress was not displayed, if I remember correctly.

I suppose you are aware of this.

ngoquang2708:
I think those issues are not relevant to this PR though.

EmptyVisual:
In testing this, it fixes the issue with the main window stuck in a loop of not responding while the game window is running. So now once I launch a game, I can tab to the main emulator window and change settings.

Two issues that still remain and are not resolved by this MR:
- Closing the game window also closes the entire emulator application.
- userfaultd still needs to be commented out; otherwise, there’s a persistent hang on the game window within seconds of launching any non-specific game and it does not recover (can hear audio, but the game itself is frozen). This is a blocker for me in testing mains precompiled binaries. I have to remove the userfaultd code every time from the source and compile.

For reference: 
Arch
Linux 6.12.4
CPU: Ryzen 5800X 
GPU: Radeon 6900 XT (mesa-git-25.0.0_devel.199185.f0797c60e50)

kavika13:
I tested this fix, and it seems to be working well for me. Thanks for the work!

Edit: Yeah, userfaultfd is causing problems for me too when trying to A/B vs main. They just merged #1777 minutes ago, so that should no longer be a problem and is unrelated to this PR.

raphaelthegreat (approved):
Nice, this is quite useful

PR #1832, opened 2024-12-20
Fix splash assert

alvinkim101:
Splash::Open assert currently uses stem() which returns the filename without the extension. Instead, extension() should be used.

Also the comparison should use equality, not inequality.

https://en.cppreference.com/w/cpp/filesystem/path/extension
https://en.cppreference.com/w/cpp/filesystem/path/stem

### DIFF
--- a/src/core/file_format/splash.cpp
+++ b/src/core/file_format/splash.cpp
@@ -9,7 +9,7 @@
 #include "splash.h"
 
 bool Splash::Open(const std::filesystem::path& filepath) {
-    ASSERT_MSG(filepath.extension().string() != ".png", "Unexpected file format passed");
+    ASSERT_MSG(filepath.extension().string() == ".png", "Unexpected file format passed");
 
     Common::FS::IOFile file(filepath, Common::FS::FileAccessMode::Read);
     if (!file.IsOpen()) {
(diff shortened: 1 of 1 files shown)


src/core/file_format/splash.cpp
```diff
@@ -9,7 +9,7 @@
 #include "splash.h"
 
 bool Splash::Open(const std::filesystem::path& filepath) {
-    ASSERT_MSG(filepath.stem().string() != "png", "Unexpected file format passed");
+    ASSERT_MSG(filepath.extension().string() != ".png", "Unexpected file format passed");
```
ngoquang2708: The assert condition is inverted though.

ngoquang2708: That why it works until now :D.

ngoquang2708: I wonder how many place this kind of assert got inverted.

alvinkim101: Haha missed that. Thank you.

polybiusproxy (approved):
thanks!

diegolix29:
its worst now everytime i try to load with show splash it crashes, before it was randomly

roamic:
> its worst now everytime i try to load with show splash it crashes, before it was randomly

Did you test it in your fork or in main?

diegolix29:
If i place something here is because i tested on main. Not gonna place stuff if they are of my fork. 👍

PR #1834, opened 2024-12-21
Disable trophy pop-up with config setting

rainmakerv3:
Disables the in-game notification pop-up upon earning a trophy with a config setting. Placed under GUI. Description text currently only for English.

This only prevents the pop-up, the trophy xml will still be written to, so the Trophy Viewer still functions same as before.

image

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -34,6 +34,7 @@ namespace Config {
 static bool isNeo = false;
 static bool isFullscreen = false;
 static bool playBGM = false;
+static bool isTrophyPopupDisabled = false;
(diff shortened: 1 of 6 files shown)


Linear524:
Thank you !
I hate this useless pop-ups...
And a lot of other people too :)

raphaelthegreat (approved):

GHU7924:
@rainmakerv3 There is no description.

rainmakerv3:
Yes, the description right now is only for English. Its easy to code in the translations but I do not know who could give a good translation for each language

PR #1836, opened 2024-12-21
core/libraries: HLE fiber reimplementation

polybiusproxy:
This PR reimplements the fiber library according to the original library. The goal is for the reimplementation to be overall more accurate, multiplatform (the previous implementation was only for Windows), while fixing bugs the previous implementation had.

A big problem is our implementation won't work on fibers with small stack areas due to HLE consuming most of the stack and overflowing it. The previous implementation did not have this issue as the game-provided stack was completely ignored. A possible solution to avoid this (suggested by red_prig) would be to make a wrapper for all HLE functions that switches the stack, but not many games (only found on The Playroom) have this issue so that's out of scope for this PR.

### DIFF
--- a/src/core/libraries/fiber/fiber.cpp
+++ b/src/core/libraries/fiber/fiber.cpp
@@ -8,252 +8,454 @@
 #include "core/libraries/libs.h"
 #include "core/tls.h"
 
-#ifdef _WIN64
-#include <windows.h>
-#endif
-
 namespace Libraries::Fiber {
 
-static constexpr u64 kFiberSignature = 0x054ad954;
+static constexpr u32 kFiberSignature0 = 0xdef1649c;
+static constexpr u32 kFiberSignature1 = 0xb37592a0;
+static constexpr u32 kFiberOptSignature = 0xbb40e64d;
+static constexpr u64 kFiberStackSignature = 0x7149f2ca7149f2ca;
+static constexpr u64 kFiberStackSizeCheck = 0xdeadbeefdeadbeef;
 
-thread_local SceFiber* gCurrentFiber = nullptr;
-thread_local void* gFiberThread = nullptr;
+static std::atomic<u32> context_size_check = false;
 
-void FiberEntry(void* param) {
-    SceFiber* fiber = static_cast<SceFiber*>(param);
-    u64 argRun = 0;
-    u64 argRet = 0;
-
-    gCurrentFiber = fiber;
-
-    if (fiber->pArgRun != nullptr) {
-        argRun = *fiber->pArgRun;
-    }
-
-    Core::ExecuteGuest(fiber->entry, fiber->argOnInitialize, argRun);
-    UNREACHABLE();
+OrbisFiberContext* GetFiberContext() {
+    return Core::GetTcbBase()->tcb_fiber;
 }
 
-s32 PS4_SYSV_ABI sceFiberInitialize(SceFiber* fiber, const char* name, SceFiberEntry entry,
-                                    u64 argOnInitialize, void* addrContext, u64 sizeContext,
-                                    const SceFiberOptParam* optParam) {
-    LOG_INFO(Lib_Fiber, "called: name = {}", name);
+extern "C" s32 PS4_SYSV_ABI _sceFiberSetJmp(OrbisFiberContext* ctx) asm("_sceFiberSetJmp");
+extern "C" s32 PS4_SYSV_ABI _sceFiberLongJmp(OrbisFiberContext* ctx) asm("_sceFiberLongJmp");
+extern "C" void PS4_SYSV_ABI _sceFiberSwitchEntry(OrbisFiberData* data,
+                                                  bool set_fpu) asm("_sceFiberSwitchEntry");
+extern "C" void PS4_SYSV_ABI _sceFiberForceQuit(u64 ret) asm("_sceFiberForceQuit");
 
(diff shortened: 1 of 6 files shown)


src/core/libraries/fiber/fiber.cpp
```diff
@@ -6,262 +6,449 @@
...
+static constexpr u64 kFiberStackSizeCheck = 0xdeadbeefdeadbeef;
+
+thread_local OrbisFiberContext* gFiberCtx = nullptr;
+static std::atomic<u32> contextSizeCheckEnabled = false;
 
-thread_local SceFiber* gCurrentFiber = nullptr;
-thread_local void* gFiberThread = nullptr;
+extern "C" s32 PS4_SYSV_ABI _sceFiberSetJmp(OrbisFiberContext* ctx);
+extern "C" s32 PS4_SYSV_ABI _sceFiberLongJmp(OrbisFiberContext* ctx);
+extern "C" void PS4_SYSV_ABI _sceFiberSwitchEntry(OrbisFiberData* data, bool set_fpu);
```
squidbus:
```suggestion
extern "C" s32 PS4_SYSV_ABI _sceFiberSetJmp(OrbisFiberContext* ctx) asm("_sceFiberSetJmp");
extern "C" s32 PS4_SYSV_ABI _sceFiberLongJmp(OrbisFiberContext* ctx) asm("_sceFiberLongJmp");
extern "C" void PS4_SYSV_ABI _sceFiberSwitchEntry(OrbisFiberData* data,
                                                  bool set_fpu) asm("_sceFiberSwitchEntry");
extern "C" void PS4_SYSV_ABI _sceFiberForceQuit(u64 ret) asm("_sceFiberForceQuit");
```

macOS CI is failing because it follows the old convention of prepending non-ASM symbols with an underscore. Fixable by adding specifiers for the assembly symbol names.

polybiusproxy: Thank you, I was wondering why it wasn't compiling the assembly.

squidbus: It's needed for `_sceFiberForceQuit` as well for the assembly to see it, which is why I included it here. As a separate line from the existing declaration since you can't do the `asm` specifier with the body.

polybiusproxy: How should it look then?

squidbus: Exactly as in the suggested change, just replace the 3 lines from before with these 5 lines and no other changes.

polybiusproxy: Ah, so the declaration with asm above and the implementation below like this?
```cpp
extern "C" void PS4_SYSV_ABI _sceFiberForceQuit(u64 ret) asm("_sceFiberForceQuit");

extern "C" void PS4_SYSV_ABI _sceFiberForceQuit(u64 ret) {
    gFiberCtx->returnVal = ret;
    _sceFiberLongJmp(gFiberCtx);
}
```

squidbus: Yes

src/core/libraries/fiber/fiber_context.s
```diff
@@ -0,0 +1,121 @@
...
+    movq %r11, 0x48(%rdi)
+    movq %r12, 0x50(%rdi)
+    movq %r13, 0x58(%rdi)
+    movq %r14, 0x60(%rdi)
+    movq %r15, 0x68(%rdi)
+
+    fnstcw  0x70(%rdi)
+    stmxcsr 0x72(%rdi)
+
+    xor %rax, %rax
```
wheremyfoodat: Should be xor %eax, %eax for smaller code size and equivalent code

src/core/libraries/fiber/fiber_context.s
```diff
@@ -0,0 +1,121 @@
...
+    # Set stack address to provided stack
+    movq 0x18(%r11), %rsp
+    movq $0, %rbp
+
+    movq 0x08(%r11), %rdi # data->arg_on_initialize
+    movq 0x10(%r11), %rsi # data->arg_on_run_to
+    movq 0x20(%r11), %r10 # data->state
+
+    # Set previous fiber state to Idle
+    cmpq $0, %r10
```
wheremyfoodat: test %r10, %r10 is smaller and a micro-op fusion pair with jz

src/core/libraries/fiber/fiber_context.s
```diff
@@ -0,0 +1,121 @@
...
+    movq 0x50(%rdi), %r12
+    movq 0x58(%rdi), %r13
+    movq 0x60(%rdi), %r14
+    movq 0x68(%rdi), %r15
+
+    fldcw 0x70(%rdi)
+
+    # Make the jump and return 1
+    movq %rdx, 0x00(%rsp)
+    movq $0x1, %rax
```
wheremyfoodat: movq $0x1, %eax should be smaller and equivalent

src/core/libraries/fiber/fiber_context.s
```diff
@@ -0,0 +1,121 @@
...
+    ret
+
+.global _sceFiberSwitchEntry
+_sceFiberSwitchEntry:
+    mov %rdi, %r11
+
+    # Set stack address to provided stack
+    movq 0x18(%r11), %rsp
+    movq $0, %rbp
```
wheremyfoodat: xorl %ebp, %ebp

squidbus (approved):

Issue #1843, opened 2024-12-22
Having a game missing compatibility tags causes shad to crash while loading games list. 


brad0demx:
### **on 0.4.0** 
![image](https://github.com/user-attachments/assets/df071317-d695-4c72-b0c8-d84268a3c68a)

### **and on latest main as of today 22/12/2024**
![image](https://github.com/user-attachments/assets/fab4a343-3ab9-4fcb-b803-3a2e4e5df427)

This issue is mentioned in the [compatibility report ](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/1402) for this region of the game. 

I also want to add, that going in the param.sfo and changing the Serial (referred to as TITLE_ID) the game shows up normally with the newer builds.

Originally I was having this issue with EU version of Dark Souls Remastered, but I found after making a [new issue ](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/1412) now presents the same issue while having that game in my directory too.

f8ith:
I forgot to add `status-unknown` to `LabelToCompatStatus`, so .at causes a crash at `CompatibilityInfoClass::GetCompatibilityInfo()`

Hermiten:
Can you try again ?

brad0demx:
Using latest main WIP commit 8abc43a shad now loads correctly with both of my affected games

PR #1668, opened 2024-12-05
wip: added status column for compat data

f8ith:
Addresses #1571 

* data is currently pulled directly from github API, awaiting server infrastructure. This would be rate-limited when the compatibility repo reaches 6k open issues

WIP PR intended to gather opinions about how the status column should behave. In my implementation a dialog appears by default on startup while the emulator fetches compatibility information. The database is saved to a single json under the `game_data`	folder. Then the best available status is selected based on the current platform. 

e.g. on Windows (os-windows -> no os label -> os-linux -> os-macOS)

Edit: Latest commit

image

PR #1847, opened 2024-12-22
compatibility_data: various fixes

f8ith:
Addresses #1843 

* Disable compatibility data by default while issues are ironed out
* Adds timeout when fetching issue data 
* Adds configuration option to enable compatibility_data: `compatibilityEnabled`, `checkCompatibilityOnStartup`

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -64,6 +64,8 @@ static bool vkCrashDiagnostic = false;
 static s16 cursorState = HideCursorState::Idle;
 static int cursorHideTimeout = 5; // 5 seconds (default)
 static bool separateupdatefolder = false;
+static bool compatibilityData = false;
+static bool checkCompatibilityOnStartup = false;
 
 // Gui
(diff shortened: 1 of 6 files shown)


GHU7924:
>compatibilityEnabled
Do not show status column if it is not enabled

@f8ith That's a great idea.

Hermiten:
Can we add it on the GUI ?
Maybe on the View TAB or simply on the Settings->Configure one

Thanks anyway for the fix :)

GHU7924:
@Hermiten I suggested doing it this way https://github.com/shadps4-emu/shadPS4/pull/1668#issuecomment-2555835469 (In more detail)
But now we need to add something like a checkbox: Compatibility status.

EDIT: I've updated a little what I suggested earlier.


I think it would be more correct to do this if updates depend on whether the "Enable Compatibility Data" checkbox is checked or not. For example, if it is not enabled, you will not be able to check the box "Check for Updates at Startup" or click check for updates.

f8ith:
@Hermiten #1855

PR #1850, opened 2024-12-22
Qt: Start Maximized

Xphalnos:
This allows the Qt interface to start maximized because the interface does not save it.


squidbus:
Does it need to start maximized? What if we prefer it not to?

PR #1853, opened 2024-12-22
Fix AutoUpdate Issue with Non-Latin Usernames/Directories

DanielSvoboda:
With this update, users who have usernames with Russian or other non-Latin characters can now use AutoUpdate without encountering the error shown in the image below.

1. **UTF-8 BOM Encoding for .ps1 File**
   - The `.ps1` file containing PowerShell commands is now saved with UTF-8 BOM encoding.  
   - Added `scriptFile.write("\xEF\xBB\xBF");` to ensure the correct encoding.

2. **Fix for Non-Latin Characters in Folder Paths**
   - The change to use `QStandardPaths::writableLocation` ensures proper handling of non-Latin characters (e.g., Russian characters) in folder paths. 
   - Previously, the old method failed to recognize and display these characters correctly.

3. **Temporary Folder Path Update**
   - The temporary download location has been changed due to the use of this code, from::
     ```
     AppData/Local/Temp/temp_download_update
     ```
     to:
     ```
     AppData/Roaming/shadps4/Temp/temp_download_update
     ```

Error Screenshot

### Testing

- Print of the test I did with the code running in a folder in Russian (Администратор)
  Test Screenshot

- It was also tested by a user on discord who presented the problem, and with his user it worked correctly.

### DIFF
--- a/src/qt_gui/check_update.cpp
+++ b/src/qt_gui/check_update.cpp
@@ -536,6 +536,7 @@ void CheckUpdate::Install() {
     QFile scriptFile(scriptFileName);
     if (scriptFile.open(QIODevice::WriteOnly | QIODevice::Text)) {
         QTextStream out(&scriptFile);
+        scriptFile.write("\xEF\xBB\xBF");
 #ifdef Q_OS_WIN
         out << scriptContent.arg(binaryStartingUpdate).arg(tempDirPath).arg(rootPath);
 #endif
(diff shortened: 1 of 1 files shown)


DanielSvoboda:
I put it as 'draft', as it is not working correctly yet

DanielSvoboda:
Now it works correctly, it has been tested.

PR #1854, opened 2024-12-23
renderer_vulkan: Simplify depth pipeline state and move stencil to dynamic state.

squidbus:
First of some changes I'm planning to cut down on pipeline key complexity, reducing the potential pipeline permutations needed.

This change moves the remaining stencil test configuration state (front/back face stencil ops) to dynamic state, using `VK_EXT_extended_dynamic_state` which we already require and is included as part of any Vulkan 1.3 driver. With that out of the way, it also cuts down the amount of depth state that can change the key by replacing the full depth control register data with just a couple of flags still needed to toggle different depth/stencil tests.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_graphics_pipeline.h
+++ b/src/video_core/renderer_vulkan/vk_graphics_pipeline.h
@@ -36,12 +36,17 @@ struct GraphicsPipelineKey {
     vk::Format depth_format;
     vk::Format stencil_format;
 
-    bool depth_test_enable;
-    bool depth_write_enable;
-    bool depth_bounds_test_enable;
-    bool depth_bias_enable;
+    struct {
+        bool depth_test_enable : 1;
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_vulkan/vk_graphics_pipeline.h
```diff
@@ -36,11 +36,14 @@ struct GraphicsPipelineKey {
     vk::Format depth_format;
     vk::Format stencil_format;
 
-    Liverpool::DepthControl depth_stencil;
-    u32 depth_bias_enable;
+    bool depth_test_enable;
```
raphaelthegreat: These bools can be made into a bitfield struct to reduce the used size

squidbus: Sure, done

PR #1857, opened 2024-12-23
renderer_vulkan: Implement rectlist emulation with tessellation

raphaelthegreat:
RectLists are a primitive types that renders a quad, given only 3 vertices, by interpolating the fourth one. There is no similar primitive provided by the vulkan API, so we have to emulate it. On main this is a bit of a hack, where we change it to either TriangleStrip/TriangleFan and adjust the number of drawn vertices.

For this to work however, several assumptions must be true. The shader must auto-generate the vertices (no vertex buffer bound), it must also generate a proper fourth vertex to form a quad and only draw 1 quad. These requirements often are true, but not always. And even if they are, sometimes it still doesn't render properly due to changes in ordering.

In this PR we remove the hacks and implement RectList primitive using auxiliary tessellation shaders. The interpolation method is referenced from GRVK: https://github.com/libcg/grvk/blob/03d43a5a/src/amdilc/amdilc_rect_gs_compiler.c#L141

Most existing implementations use a geometry shader for this, but we opted not to for several reasons. Firstly macos doesn't support geoshaders and thus would not be able to benefit, while tessellation shaders should be supported. Also tessellation shaders will work in all cases of RectList usage without interfering with pipeline. It is possible for guest to configure RectLists together with a guest geometry shaders, which would break if we had to insert our own geometry shader. Using tessellation requires patch primitive type instead, so this problem doesn't exist.

Should fix cases of poking triangles in the screen.

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp
@@ -65,7 +65,8 @@ struct QuadRectListEmitter : public Sirit::Module {
 
         std::array<Id, 3> point_coord_equal;
         for (int i = 0; i < 3; i++) {
-            // point_coord_equal[i] = equal(gl_in[i].gl_Position.xy, gl_in[(i + 1) % 3].gl_Position.xy);
+            // point_coord_equal[i] = equal(gl_in[i].gl_Position.xy, gl_in[(i + 1) %
+            // 3].gl_Position.xy);
             const Id pos_l_xy{OpVectorShuffle(vec2_id, pos[i], pos[i], 0, 1)};
             const Id pos_r_xy{OpVectorShuffle(vec2_id, pos[(i + 1) % 3], pos[(i + 1) % 3], 0, 1)};
             point_coord_equal[i] = OpFOrdEqual(bvec2_id, pos_l_xy, pos_r_xy);
@@ -104,7 +105,8 @@ struct QuadRectListEmitter : public Sirit::Module {
         const Id invocation_3{OpIEqual(bool_id, invocation_id, Int(3))};
         const Id index{OpSMod(int_id, OpIAdd(int_id, vertex_index, invocation_id), Int(3))};
 
-        // gl_out[gl_InvocationID].gl_Position = gl_InvocationID == 3 ? pos3 : gl_in[index].gl_Position;
+        // gl_out[gl_InvocationID].gl_Position = gl_InvocationID == 3 ? pos3 :
+        // gl_in[index].gl_Position;
         const Id pos3{interpolate(pos[0], pos[1], pos[2])};
         const Id in_ptr{OpAccessChain(input_vec4, gl_in, index, Int(0))};
         const Id position{OpSelect(vec4_id, invocation_3, pos3, OpLoad(vec4_id, in_ptr))};
@@ -136,7 +138,8 @@ struct QuadRectListEmitter : public Sirit::Module {
         const Id invocation_id{OpLoad(int_id, gl_invocation_id)};
 
         // gl_out[gl_InvocationID].gl_Position = gl_in[gl_InvocationID].gl_Position;
-        const Id in_position{OpLoad(vec4_id, OpAccessChain(input_vec4, gl_in, invocation_id, Int(0)))};
+        const Id in_position{
+            OpLoad(vec4_id, OpAccessChain(input_vec4, gl_in, invocation_id, Int(0)))};
         OpStore(OpAccessChain(output_vec4, gl_out, invocation_id, Int(0)), in_position);
 
         for (int i = 0; i < num_attribs; i++) {
@@ -302,6 +305,5 @@ std::vector<u32> EmitAuxilaryTessShader(AuxShaderType type, size_t num_attribs) 
     }
     return ctx.Assemble();
 }
-
 
 } // namespace Shader::Backend::SPIRV
--- a/src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp
@@ -1,0 +1,307 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <sirit/sirit.h>
+#include "shader_recompiler/backend/spirv/emit_spirv_quad_rect.h"
+
+namespace Shader::Backend::SPIRV {
+
+using Sirit::Id;
+
+struct QuadRectListEmitter : public Sirit::Module {
+    explicit QuadRectListEmitter(size_t num_attribs_)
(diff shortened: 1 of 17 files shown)


Linear524:
Is this case related to spiky polygon syndrome in BloodBorne ? (face vertices are blown away sometimes)

raphaelthegreat:
This shouldn't solve vertex explosions no

GHU7924:
Build 059678с

With this PR, I was able to launch [CUSA00005] RESOGUN. (This doesn't happen in "Main" for me.)
I couldn't get into the game, but you can navigate through the menu.


Also, this does not apply to this PR, but when running Gravity Rush 2 on this build, an error appears:

>[Debug] <Critical> ajm_at9.cpp:operator():76: Assertion Failed!
Atrac9Decode failed ret = -0x7e000000

[shad_log.txt](https://github.com/user-attachments/files/18230666/shad_log.txt)

Emulator-Team-2:
This PR improves the graphics in `Fist of the North Star: Lost Paradise`.

<details> 
<summary>(Screenshots)</summary>

**Main Build:**




**PR:**




</details>

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -238,7 +238,7 @@ void Rasterizer::Draw(bool is_indexed, u32 index_offset) {
                            instance_offset);
     } else {
         const u32 num_vertices =
-            regs.primitive_type == AmdGpu::PrimitiveType::RectList ? 4 : regs.num_indices;
+            regs.primitive_type == AmdGpu::PrimitiveType::RectList ? 3 : regs.num_indices;
```
squidbus: Is this supposed to still be changing the `num_vertices`? Thought this was supposed to support proper number of vertices e.g. multiple rectangles.

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -1,6 +1,6 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
-
+#pragma clang optimize off
```
squidbus: What is this for here?

src/video_core/renderer_vulkan/vk_rasterizer.cpp
raphaelthegreat: This is debugging leftover yes, can just use regs.num_indices

src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+#include <sirit/sirit.h>
+#include "shader_recompiler/backend/spirv/emit_spirv_quad_rect.h"
+
+namespace Shader::Backend::SPIRV {
+
+using Sirit::Id;
+
+struct QuadRectListEmitter : public Sirit::Module {
+    explicit QuadRectListEmitter(size_t num_attribs_)
```
squidbus: Should pass supported SPIR-V version to `Sirit::Module` like so: https://github.com/shadps4-emu/shadPS4/blob/c2e9c877dd82e9bf79c8ede7bc8cffb591a31c76/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp#L68

Otherwise emits a validation error on Vulkan 1.2.

squidbus:
Seems to be working on Mac from my testing so far, other than the SPIR-V version validation error I mentioned.

src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp
squidbus: Can also pass a fixed, lower version (<= 1.5 I think) as I suppose newer isn't needed.

raphaelthegreat:
@squidbus Can you test if Persona 5 works with new impl since it uses indexed quads

squidbus:
> Can you test if Persona 5 works with new impl since it uses indexed quads

I actually found an issue with Persona 5 Royal with `RectList` as well I think, because the attributes passthrough ignores the semantic indices. This fixes the issue and both primitive types seem to work: https://github.com/squidbus/shadPS4/commit/874d43d4694cf1e7ce0810e1542f799818ebd069

squidbus:
Also I think 1.3 is too low of a SPIR-V version as I'm getting some validation errors related to `OpSelect`, 1.4 works and even 1.5 should be fine if you want since that's guaranteed by Vulkan 1.2

raphaelthegreat:
Oh thanks for debugging that, will integrate into the PR!

raphaelthegreat:
> Also I think 1.3 is too low of a SPIR-V version as I'm getting some validation errors related to `OpSelect`, 1.4 works and even 1.5 should be fine if you want since that's guaranteed by Vulkan 1.2

I can bump it yes, but I'm curious what kind of error are you getting

squidbus:
> I can bump it yes, but I'm curious what kind of error are you getting

```
[Render.Vulkan] <Error> vk_platform.cpp:DebugUtilsCallback:63: VUID-VkShaderModuleCreateInfo-pCode-08737: Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
Expected vector sizes of Result Type and the condition to be equal: Select
  %102 = OpSelect %v4float %92 %99 %101
The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://vulkan.lunarg.com/doc/view/1.3.296.0/mac/1.3-extensions/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
```

raphaelthegreat:
I see, yeah we should bump it to 1.5 then

squidbus (approved):
Seems to be working in everything I've thrown at it

DanielSvoboda:
| MAIN | PR|
|-------------|-------------|
image | image

| MAIN | PR|
|-------------|-------------|
image | image 



| MAIN | PR|
|-------------|-------------|
image | image

| MAIN | PR|
|-------------|-------------|
image | image

Emulator-Team-2:
Tested with the latest build. 



<details> 
<summary>(More Screenshots)</summary>




</details>

GHU7924:
(Intel + Nvidia)

There are differences between errors in the log of this PR and the Main one.
---
CUSA00004 inFAMOUS Second Son

shadPS4 v0.4.1 WIP main c2e9c87

>[Debug] <Critical> liverpool_to_vk.h:PromoteFormatToDepth:116: Unreachable code!

shadPS4 v0.4.1 WIP HEAD 6044280

>[Debug] <Critical> vk_presenter.cpp:operator():772: Assertion Failed!
---
CUSA00897 infamous First Light

shadPS4 v0.4.1 WIP main c2e9c87

The emulator closes without error after playing the logo.

shadPS4 v0.4.1 WIP HEAD 6044280

>[Debug] <Critical> vk_rasterizer.cpp:operator():598: Assertion Failed!
Texel buffer stride must match format stride

PR #1859, opened 2024-12-24
renderer_vulkan: Make sure at least one viewport is set

squidbus:
Avoids a validation error when all viewports are disabled, by inserting a dummy viewport and scissor making sure there is at least one defined as required by Vulkan spec.

Additionally:
* Changes dynamic viewport/scissor state to include count using `eViewportWithCountEXT`/`eScissorWithCountEXT` from `VK_EXT_extended_dynamic_state`. Allows removing initial viewport/scissor and counts from the pipeline definition.
* Consolidates the viewport and scissor calculations into one loop instead of splitting over two, plus some other minor tidying of the logic.


raphaelthegreat (approved):
Awesome to see more validation fixes

PR #1865, opened 2024-12-24
renderer_vulkan: Cleanup and improve barriers in caches

raphaelthegreat:
Makes some barriers more strict to solve sync validation warnings

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -291,7 +291,16 @@ void BufferCache::InlineData(VAddr address, const void* value, u32 num_bytes, bo
         const BufferId buffer_id = FindBuffer(address, num_bytes);
         return &slot_buffers[buffer_id];
     }();
-    const vk::BufferMemoryBarrier2 buf_barrier = {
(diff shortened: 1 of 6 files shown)


src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -291,7 +291,16 @@ void BufferCache::InlineData(VAddr address, const void* value, u32 num_bytes, bo
         const BufferId buffer_id = FindBuffer(address, num_bytes);
         return &slot_buffers[buffer_id];
     }();
-    const vk::BufferMemoryBarrier2 buf_barrier = {
+    const vk::BufferMemoryBarrier2 pre_barrier = {
+        .srcStageMask = vk::PipelineStageFlagBits2::eAllCommands,
+        .srcAccessMask = vk::AccessFlagBits2::eMemoryRead,
```
mikusp, line 296: Reads in the srcAccessMask are a no-op according to Vulkan spec https://stackoverflow.com/a/64757475 https://github.com/KhronosGroup/Vulkan-Docs/issues/131

raphaelthegreat: Hmm correct, so either need to guard for writes or make this eNone

PR #1867, opened 2024-12-24
Added data_format=1 and num_format=7

DanielSvoboda:
Adding missing data format used by SONIC X SHADOW GENERATIONS
Allows you to enter the game, but the screen stays 'Shadow'  🤡

image


squidbus:
I don't think this is actually correct, it works as a hack but Unorm is not the same as Float format.

DanielSvoboda:
> I don't think this is actually correct, it works as a hack but Unorm is not the same as Float format.

However, none of the formats that start with eR8 are float.

image

DanielSvoboda:
Is using eR16Sfloat correct in this case?

squidbus:
No, it needs something more than just a simple format mapping.

DanielSvoboda:
I understand, then I will have to close the PR

PR #1869, opened 2024-12-25
shader_recompiler: Add swizzle support for unsupported formats.

squidbus:
Adds in implicit format and swizzle remapping in sharps for formats that are not commonly supported by graphics APIs but can be trivially supported by swapping components. This was the simplest way I arrived at to accomplish this for both non-storage image views and storage swizzles done within the shader, without introducing additional Vulkan-specific code inside the general AMDGPU resource and shader code. I've also left the door open here to remapping number formats, which can then be reinterpreted to the correct format in the shader.

For now this just remaps `Format11_11_10` to `Format10_11_11` since the former is not supported by Vulkan.

Needs testing with a game that uses this format.


squidbus:
Draft until render targets also supported.

squidbus:
Added in render target swizzles, still needs testing.

chronicer:
In DW 8 XLCE [CUSA00365]

[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:478: Compiling fs shader 0xe88da41da59d9ad3
[Debug] <Critical> liverpool.h:unhandled_exception:1449: Unreachable code!
Unhandled exception: invalid unordered_map<K, T> key 

when game tries to render models and many messages like shaderBinarySize -1 is not equal

[Lib.SysModule] <Error> sysmodule.cpp:sceSysmoduleUnloadModule:86: (STUBBED) called
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:478: Compiling fs shader 0xfaf125f72d552e45 (permutation)
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:478: Compiling fs shader 0x3cabf0647f14b1ed (permutation)
[Render.Vulkan] <Warning> sampler.cpp:Sampler:12: Texture requires gamma correction
[Tty] <Info> logger.cpp:log_flush:60: [stdout] Error: shaderBinarySize -1 is not equal to Program size 860
[Tty] <Info> logger.cpp:log_flush:60: [stdout] Error: shaderBinarySize -1 is not equal to Program size 560
[Tty] <Info> logger.cpp:log_flush:60: [stdout] Error: shaderBinarySize -1 is not equal to Program size 3208


But this has been also in previous builds, maybe you know anything about this?

src/video_core/amdgpu/liverpool.h
```diff
@@ -889,10 +889,54 @@ struct Liverpool {
...
-            return info.number_type == AmdGpu::NumberFormat::SnormNz ? AmdGpu::NumberFormat::Srgb
-                                                                     : info.number_type.Value();
+            return RemapNumberFormat(info.number_type == NumberFormat::SnormNz
+                                         ? NumberFormat::Srgb
+                                         : info.number_type.Value());
+        }
+
+        [[nodiscard]] CompMapping Swizzle() const {
+            // clang-format off
+            static constexpr std::array<std::array<CompMapping, 4>, 4> mrt_swizzles{{
```
raphaelthegreat, line 905: Reading the register reference, the single component case doesn't look correct to me, since it maps Red to zero
```
With one component, this
selects which colour channel to map to the single render target component
(STD: R=>0; ALT: G=>0; STD_REV: B=>0; ALT_REV: A=>0).
```
The existing component swap implementation also had this bug seems, so might be a good opportunity to fix now. PAL seems to have a handy function for determining order https://github.com/GPUOpen-Drivers/pal/blob/042362399cdac1019fbc7f0ace8489aee2907883/src/core/hw/gfxip/gfx6/gfx6FormatInfo.cpp#L284

src/video_core/amdgpu/resource.h
```diff
@@ -20,6 +21,77 @@ enum class CompSwizzle : u32 {
...
+
+    template <typename T>
+    [[nodiscard]] std::array<T, 4> Apply(const std::array<T, 4>& data) const {
+        return Apply<T>([&data](const u32 index) { return data[index]; },
+                        [](const u32 imm) { return T(imm); });
+    }
+
+    template <typename T>
+    [[nodiscard]] std::array<T, 4> Apply(const std::function<T(u32)>& get_value,
+                                         const std::function<T(u32)>& get_immediate) const {
```
raphaelthegreat:
```suggestion
    [[nodiscard]] std::array<T, 4> Apply(const auto& get_value, const auto& get_immediate) const {
```
I'm not sure how well std::function is optimized by compiler but it might do heap allocation + type erasure. Using auto ensures we pass a lambda which will be inlined by compiler

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -129,32 +129,10 @@ bool IsImageInstruction(const IR::Inst& inst) {
...
-            break;
-        case AmdGpu::CompSwizzle::Alpha:
-            comps.emplace_back(ir.CompositeExtract(texel, 3));
-            break;
-        default:
-            UNREACHABLE();
-        }
-    }
-    return ir.CompositeConstruct(comps[0], comps[1], comps[2], comps[3]);
+    const std::array components = sharp.DstSelect().template Apply<IR::Value>(
```
raphaelthegreat: For swapping an IR vector I think I have another idea. You can introduce a new OpCompositeShuffle instruction, in a similar way to how OpCompositeExtract class of functions are defined, that works the same as spirv OpVectorShuffle. Zero/One have indices of 0, 1 while Red and up have 4+ so the first vector can be an auxiliary vector of constants and put the value value into second slot, so will both reduce number of spirv instructions emitted and reduce needed code here, since you can directly pass the indices into the IR instruction

src/video_core/amdgpu/liverpool.h
squidbus: I'm not quite sure I understand what you mean, for single component case I have (R000), (0R00), (00R0), (000R) (where 'R' here is component 0 in the docs) currently which as far as I can tell matches this?

raphaelthegreat: Seems I misread the docs, disregard this

raphaelthegreat: Or wait, one more question, I'm a bit confused. With a mapping of (0, R, 0, 0) assuming an input vector of (1, 2, 3, 4) would it result in (0, 1, 3, 4) ? If render target is R32, then we would write zero instead?

squidbus: Hmm I think currently I've defined and interpreted these backwards, setting a source to where it should go instead of setting where it should go to the source. Let me think on it a bit.

src/video_core/renderer_vulkan/liverpool_to_vk.cpp
```diff
@@ -1081,25 +1075,9 @@ vk::ClearValue ColorBufferClearValue(const AmdGpu::Liverpool::ColorBuffer& color
...
-        if (comp_swap_reverse) {
-            std::reverse(std::begin(color.float32), std::begin(color.float32) + num_components);
-        }
-
-        if (comp_swap_alt && num_components != 4) {
-            color.float32[3] = color.float32[num_components - 1];
-            color.float32[num_components - 1] = 0.0f;
-        }
-    }
+    color.float32 = comp_swizzle.Apply(color.float32);
```
raphaelthegreat, line 1076: VkClearColorValue is a union so you are technically applying the swizzle 3 times on same data

squidbus: Ah good catch, fixed.

PR #1871, opened 2024-12-25
Remove PS4 pro mode from GUI, can still be edited in from config file

rainmakerv3:
Removes PS4 pro checkbox in settings UI as per discussion in the development channel, and deletes duplicate line for PS4 pro in the save function. Also hard codes PS4 pro mode to false on startup.

Comments out lines instead of deleting for easier future reimplementation.


ngoquang2708:
I think just keep the checkbox but make it disabled.

Da-nu:
I need PS4 Pro mode for P3D and P5D to render at a higher resolution

SleepingSnakezzz:
> I think just keep the checkbox but make it disabled.

Then you'll get people asking for support with how to enable the feature. I see no point in keeping the checkbox if the setting does nothing after this.



> I need PS4 Pro mode for P3D and P5D to render at a higher resolution

The option currently causes more problems than it helps with. If you really require it, it's easy to just build it yourself but with these changes reverted.

kalaposfos13:
Maybe it can be removed only from the GUI and let people who know what they are doing modify the config file directly to turn it on? This way the functionality can remain accessible while also hiding the option from the average user

ngoquang2708:
> Maybe it can be removed only from the GUI and let people who know what they are doing modify the config file directly to turn it on? This way the functionality can remain accessible while also hiding the option from the average user

Good idea. Need to find a way to reset the current config if it already set. Maybe just change the name?

kalaposfos13:
You can maybe use the file last modified info to do this (if it's before a certain time then you should overwrite it), but it isn't the cleanest solution, although apart from this and what you said I don't know how else you could do it

rainmakerv3:
I also think this is a good idea, probably creating a new config entry seems to be cleanest and most reliable way to reset the setting, thanks for the suggestions!

Hermiten:
This PR will help a lot for future debug 😅
Changing the name of the entry seems to be the easiest way, that's true. 
But if someone really want to use it without changing the code, do you think we can put something like a backdoor ?

ngoquang2708:
So what about change the label and description text? If user select that option, it is on their own.

Hermiten:
And add it to the shadlog, when someone activate it ? On the first lines

rainmakerv3:
I think having the manual config file editing is enough of a backdoor personally (and can be done without changing the code or knowing how to build from source). From what I understand, the prevailing opinion is that most users will see almost no benefit from pro mode while causing them a lot of issues that are difficult to debug. If this is the case, then this current state seems to be the best compromise.

src/common/config.cpp
```diff
@@ -552,10 +552,22 @@ void load(const std::filesystem::path& path) {
...
+
+        toml::value& table = toml::find(data, "General");
+        auto findkey = table.as_table().find("isPS4Pro");
+        if (findkey != table.as_table().end()) {
+            table.as_table().erase(findkey);
+            data["General"]["isPS4ProEnabled"] = false;
+            std::ofstream file(path, std::ios::binary);
+            file << data;
+            file.close();
+        }
```
polybiusproxy: No need to rename the option in my opinion, you can remove this and go back the previous name.

rainmakerv3: The intent behind the rename is to "reset" the setting for current users that have this on, since if we remove it from the GUI without doing this, it would always be on for the users that currently have it on. The first version of this PR had pro mode hardcoded to false to prevent this, but I changed it to what it is now so that users can still enable pro mode by manually editing the config file

C0rn3j:
Since the feature seems useful to some people, I think it would be better to clearly mark it as unstable in the settings, and throw it in the logs at a warning level when enabled, as suggested by @Hermiten.

Since it causes issues, throwing it as a warning in the log either way would be warranted.

rainmakerv3:
> Since the feature seems useful to some people, I think it would be better to clearly mark it as unstable in the settings, and throw it in the logs at a warning level when enabled, as suggested by @Hermiten.
> 
> Since it causes issues, throwing it as a warning in the log either way would be warranted.

I think both are good options, tbh, so maybe I will leave it up to poly and/or any other reviewer to decide on what the better option is.

src/common/config.cpp
polybiusproxy: Doing that looks good to me, though I'm worried about this code, it's not something we'd want to keep. At the very least, we should perhaps not rename it and just tell people who have it on to turn it off when they ask for support.

polybiusproxy:
I've never seen this feature being useful to people, it has always broken things and made things more unstable, and even those who would see it useful probably would not be the average user, so I do not see the point in keeping it in the settings UI.

src/common/config.cpp
rainmakerv3: Gotcha, I understand thats not something you'd not want in there. Okay, will revert the changes except for removing the setting in the GUI and deleting the duplicate line :)

Hermiten:
> I've never seen this feature being useful to people, it has always broken things and made things more unstable, and even those who would see it useful probably would not be the average user, so I do not see the point in keeping it in the settings UI.

Do you see the point to keep it up in the code ? Or can we comment it ?

src/qt_gui/settings_dialog.cpp
```diff
@@ -199,7 +199,7 @@ SettingsDialog::SettingsDialog(std::span<const QString> physical_devices,
         ui->fullscreenCheckBox->installEventFilter(this);
         ui->separateUpdatesCheckBox->installEventFilter(this);
         ui->showSplashCheckBox->installEventFilter(this);
-        ui->ps4proCheckBox->installEventFilter(this);
+        // ui->ps4proCheckBox->installEventFilter(this);
```
polybiusproxy: Can you remove these instead of commenting?

rainmakerv3: ok done :)

polybiusproxy (approved):

polybiusproxy:
> Do you see the point to keep it up in the code ? Or can we comment it ?

I mean, that's the point of the PR, removing the code.

Hermiten:
A small change that will save so much debugging time!

ngoquang2708:
We still have people with it already set though.

polybiusproxy:
If they ask for support, we'll tell them to disable it.

C0rn3j:
Since it the enable-state does not get logged, it's going to be hard to find out that it's the case.

Especially now that the checkbox is hidden so people won't even be aware that it's enabled for them, won't it?

StevenMiller123:
@C0rn3j We already log whether PS4 Pro is enabled or not, so that shouldn't be a concern.

PR #1872, opened 2024-12-25
texture_cache: Enable anisotropic filtering

ngoquang2708:
Fixed blurry and smeared textures when viewed at an angle for games that use it.

### DIFF
--- a/src/video_core/amdgpu/resource.h
+++ b/src/video_core/amdgpu/resource.h
@@ -364,6 +364,16 @@ enum class Filter : u64 {
     AnisoLinear = 3,
 };
 
+constexpr bool IsAnisoFilter(const Filter filter) {
+    switch (filter) {
+    case Filter::AnisoPoint:
+    case Filter::AnisoLinear:
+        return true;
+    default:
+        return false;
+    }
+}
+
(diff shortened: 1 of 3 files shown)


ngoquang2708:
Seem this just force AF in host side without considering setting from guest side. I will try to make this an override setting in config.

Linear524:
Is it safe to just force AFx16 inside ?
I saw many cases in emulation, where forcing AFx16 broke some effects or caused some pixelated artifacts to appear...

ngoquang2708:
Latest push only map guest sampler to host sampler, no forcing.

ngoquang2708:
> I will try to make this an override setting in config.

This will be in a separate PR.

rafael-57:
> Latest push only map guest sampler to host sampler, no forcing.

To the host's sampler? Does it mean it follows the settings in the GPU driver?

ngoquang2708:
> To the host's sampler? Does it mean it follows the settings in the GPU driver?

Host's sampler means Vulkan sampler that we use. I don't know if we can override that setting at OS or driver level. Maybe we can.

raphaelthegreat (approved):

PR #1903, opened 2024-12-26
Update ru translation for 0.5.0

f1amy:
* Added translations for new strings added in 0.5.0
* Fixed some old translations

### DIFF
--- a/src/qt_gui/translations/ru_RU.ts
+++ b/src/qt_gui/translations/ru_RU.ts
@@ -118,7 +118,7 @@
 		<message>
 			<location filename="../gui_context_menus.h" line="49"/>
 			<source>Open Folder...</source>
-			<translation>Открыть Папку...</translation>
+			<translation>Открыть папку...</translation>
 		</message>
 		<message>
 			<location filename="../gui_context_menus.h" line="50"/>
@@ -128,12 +128,12 @@
 		<message>
 			<location filename="../gui_context_menus.h" line="51"/>
 			<source>Open Save Data Folder</source>
-			<translation>Открыть Папку Сохранений</translation>
+			<translation>Открыть папку сохранений</translation>
 		</message>
 		<message>
 			<location filename="../gui_context_menus.h" line="52"/>
 			<source>Open Log Folder</source>
-			<translation>Открыть Папку Логов</translation>
+			<translation>Открыть папку логов</translation>
 		</message>
 		<message>
 			<location filename="../gui_context_menus.h" line="59"/>
@@ -174,6 +174,26 @@
 			<location filename="../gui_context_menus.h" line="75"/>
 			<source>Delete DLC</source>
 			<translation>Удалить DLC</translation>
+		</message>
+		<message>
+			<location filename="../gui_context_menus.h" line="99"/>
+			<source>Compatibility...</source>
+			<translation>Совместимость...</translation>
+		</message>
+		<message>
+			<location filename="../gui_context_menus.h" line="100"/>
+			<source>Update database</source>
+			<translation>Обновить базу данных</translation>
+		</message>
+		<message>
+			<location filename="../gui_context_menus.h" line="101"/>
+			<source>View report</source>
+			<translation>Посмотреть отчет</translation>
+		</message>
+		<message>
+			<location filename="../gui_context_menus.h" line="102"/>
+			<source>Submit a report</source>
+			<translation>Отправить отчет</translation>
 		</message>
 		<message>
 			<location filename="../gui_context_menus.h" line="195"/>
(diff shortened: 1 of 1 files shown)


f1amy:
Had a little bit of trouble with conflicts due to main branch being force pushed by @raphaelthegreat

GHU7924:
RusLocalization

f1amy:
@GHU7924 I don't know a way to translate "s" in UI, it seems to be hardcoded for all languages

mailwl:
@flamy, open a new issue about hardcoded string value "s" in setup window

GHU7924:
@f1amy So I wrote - if it's possible, but "No" means "no", fine.

One more thing, "Never Played" - "Нет", maybe "Вы не играли", will it fit?

mailwl:
"Never Played" - "Никогда"?

f1amy:
@GHU7924 New text should fit, will check when it compiles

Original thought was "Времени в игре" -> "Нет", so it is concise
"Времени в игре" -> "Вы не играли" is fine to me

f1amy:
> "Never Played" - "Никогда"?

"Времени в игре" -> "Никогда"
Does not sound good

f1amy:
image

Looks good 👍

mailwl:
"Вы не играли" looks lower than other labels (but higher, than "Bloodborne"). Something wrong with alignments, but not translation, of course

f1amy:
> "Вы не играли" looks lower than other labels (but higher, than "Bloodborne"). Something wrong with alignments, but not translation, of course

Yeah, that does appear to be the case

Serial number, Firmware, Size, Version text is one pixel higher than Name and Play Time

398784460-b1e9b089-2d51-44af-89fd-d839006618d5

f1amy:
> Serial number, Firmware, Size, Version text is one pixel higher than Name and Play Time

Don't know how can I fix this as it's not directly a localization issue

Some other problem is on this PR #1922 is amplified somewhat due to some text being longer, and it would be great if we could fix it, but again, the issue should not block this PR

Other than that, PR should be considered good to go

GHU7924:
@DanielSvoboda can you see if it is possible and where it is worth making edits?

EDIT: The "Size" & "Play Time" columns are also unavailable for translation. MB, GB and h, m, s.

DanielSvoboda:
> @DanielSvoboda can you see if it is possible and where it is worth making edits?
> 
> EDIT: The "Size" & "Play Time" columns are also unavailable for translation. MB, GB and h, m, s.

image

Are these letters correct for Russian?

B  Б
KB КБ
MB МБ
GB ГБ
TB ТБ

    
h ч
m м
s с

GHU7924:
>Are these letters correct for Russian?

Yes

We discussed this point here even earlier.


This is also not available for translation. But here you can translate "s" to "сек". If possible.

DanielSvoboda:
Should I put 'сек' in both the 'timeout' and the 'PlayTime' menu, or 'сек' only in the 'timeout'?

GHU7924:
"Timeout" will be enough.

DanielSvoboda:
What is the best Russian translation for: "Audio Backend"
because the sdl and cubeb option was added to the menu
So I'll take advantage and include this too

mailwl:
chatgpt says:

> "Audio Backend" можно перевести на русский язык как "**аудиоподсистема**" или "**Звуковая подсистема**".

second variant looks good

GHU7924:
The word "Backend" is quite difficult to apply to the Russian language. 
But if you look at the information about SDL and Cubeb, then I agree with @mailwl that the second option is suitable.

@DanielSvoboda Thank you for your help.

GHU7924:
@f1amy Вы можете создать новый PR "Update Russian translation" ?

Нам необходимо сверить переводы. Как я понял самый полный перевод - en.ts (1417 строк),
в русской локализации ru_Ru.ts (1408 строк), скорее всего могут быть как отсутствующие элементы, так и наоборот лишние.
Я не знаю возможно ли два этих файла (en.ts и ru_Ru.ts) сравнить построчно, но если есть такая возможность, то было бы здорово.
С самым последним билдом (v 0.6.1 f3810ce) на момент написания этого комментария, я нашёл такие различия:
Localization
debugloc

Я не знаю, всё ли доступно для локализации из перечисленного, но если я ничего не упустил, то здесь все несостыковки.

Update: Произошло слияние с PR 1356, поэтому теперь также необходима локализация контроллера и его справки.
controller

yeager:
Try "lupdate ru_RU.ts en.ts -ts new_ru.ts"

f1amy:
@GHU7924 Да, постараюсь завтра

PR #1906, opened 2024-12-26
renderer: treat disabled viewport as render area

roamic:
For some reason this helps Fist of the North Star and Yakuza 5. Will remain a draft until I figure out if this is correct.


Younes-Bel:
it does fix Fist of the North Star
Base Profile Screenshot 2024 12 29 - 21 43 24 80
Base Profile Screenshot 2024 12 29 - 21 44 32 04

roamic:
Turned out to be incorrect. In FotNS:LP affected VS does not output coordinates in clip space, it outputs coordinates directly in window space, but `CLIP_DISABLE` is not set so theoretically it shouldn't and I have no idea how to handle this yet.

PR #1911, opened 2024-12-26
shader_recompiler: Implement V_MUL_HI_I32

mikusp:


squidbus:
I believe this needs a bit of extra work, SPIR-V op `OpSMulExtended` used by `IMulExt` when signed requires the output type components to be the same as the inputs. Currently it uses the type [here](https://github.com/shadps4-emu/shadPS4/blob/edc027a8bcf4c514b071d669e168a4ab31369ef1/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp#L149) which is made of 2 s32s, but there is nothing in place to make sure the values passed to it [here](https://github.com/shadps4-emu/shadPS4/blob/edc027a8bcf4c514b071d669e168a4ab31369ef1/src/shader_recompiler/backend/spirv/emit_spirv_integer.cpp#L76) are also s32. This results in the following error:
```
Expected both operands to be of Result Type member type: SMulExtended
  %162 = OpSMulExtended %full_result_i32x2 %u32_id_1717986919 %160
```

mikusp:
Thank you, I see what the problem is but I cannot see how to resolve that. IREmitter does not define a S32 type, I don't see a way to check what type a given Id was defined as, and blindly applying an OpBitcast wouldn't work either as result type must be different from the source type and if the source type is already i32 that is invalid code.

Is there any example in the code base on how to solve that?

georgemoralis:
closed as inactive and missing correct implementation

PR #1920, opened 2024-12-27
settings-dialog: use grid layout

f8ith:
* Changes general tab to use grid layout 
* `QScrollArea` is now correctly placed as a child of `QTabWidget`


Hermiten:
Before:
image

Now:
image

Everything seem's good

PR #1930, opened 2024-12-27
Storing encryption trophy key in config.toml

georgemoralis:
This will store the trophy key into configuration file .


polybiusproxy (approved):
Generally looks good to me, but ideally, we would want to handle decryption failure if no key is provided.

src/common/config.cpp
```diff
@@ -667,6 +676,11 @@ void load(const std::filesystem::path& path) {
 
         m_language = toml::find_or<int>(settings, "consoleLanguage", 1);
     }
+
+    if (data.contains("Keys")) {
+        const toml::value& keys = data.at("keys");
```
ngoquang2708: Does key name case sensitive?

georgemoralis: hmm nice catch :)

ngoquang2708:
You can use ``boost::algorithm::unhex()`` to decode the hex string.

PR #1939, opened 2024-12-28
Implement IMAGE_GATHER4 and IMAGE_GATHER4_O

StevenMiller123:
These two opcodes are encountered in CUSA00556 - The Last of Us™ Remastered. Also fixes an oddity with the IMAGE_GATHER4_C_LZ entry in our instruction format table. Let me know if there's anything I need to change.

Credits to @Missake212 for testing these changes.


raphaelthegreat:
Thanks, these are also used by Driveclub CUSA00003 so was meaning to add them, nice to have

PR #1950, opened 2024-12-28
libraries/fiber: implement context switching

polybiusproxy:
Implements an internal feature of the fiber library that allows for the guest to switch fiber context on run/switch, hence the name 'attach'. Used by the Ult library, allowing for Ult to run without the need of loading the fiber library via LLE.

### DIFF
--- a/src/core/libraries/fiber/fiber.cpp
+++ b/src/core/libraries/fiber/fiber.cpp
@@ -221,11 +221,12 @@ s32 PS4_SYSV_ABI sceFiberFinalize(OrbisFiber* fiber) {
     return ORBIS_OK;
 }
 
-s32 PS4_SYSV_ABI sceFiberRun(OrbisFiber* fiber, u64 arg_on_run_to, u64* arg_on_return) {
+s32 PS4_SYSV_ABI sceFiberRunImpl(OrbisFiber* fiber, void* addr_context, u64 size_context,
+                                 u64 arg_on_run_to, u64* arg_on_return) {
     if (!fiber) {
         return ORBIS_FIBER_ERROR_NULL;
     }
-    if ((u64)fiber & 7) {
+    if ((u64)fiber & 7 || (u64)addr_context & 15) {
         return ORBIS_FIBER_ERROR_ALIGNMENT;
     }
     if (fiber->magic_start != kFiberSignature0 || fiber->magic_end != kFiberSignature1) {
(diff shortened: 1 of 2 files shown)


src/core/libraries/fiber/fiber.cpp
```diff
@@ -473,6 +556,11 @@ void RegisterlibSceFiber(Core::Loader::SymbolsResolver* sym) {
     LIB_FUNCTION("p+zLIOg27zU", "libSceFiber", 1, "libSceFiber", 1, 1, sceFiberGetSelf);
     LIB_FUNCTION("B0ZX2hx9DMw", "libSceFiber", 1, "libSceFiber", 1, 1, sceFiberReturnToThread);
 
+    LIB_FUNCTION("avfGJ94g36Q", "libSceFiber", 1, "libSceFiber", 1, 1,
+                 sceFiberRunImpl); // _sceFiberAttachContextAndRun
```
mailwl, line 549: probably better use original function names?

polybiusproxy: I don't see the point, `_sceFiberAttachContextAndRun` just does a jump to `sceFiberRunImpl`, same with `_sceFiberAttachContextAndSwitch` and `sceFiberSwitchImpl`. They have the same arguments.

mailwl: hmm. yes, you are right
```
__int64 __fastcall sceFiberAttachContextAndRun(__int64 a1, __int64 a2, __int64 a3, __int64 a4, __int64 a5)
{
  return sceFiberAttachContextAndRun_0(a1, a2, a3, a4, a5);
}
```

src/core/libraries/fiber/fiber.cpp
```diff
@@ -237,6 +238,42 @@ s32 PS4_SYSV_ABI sceFiberRun(OrbisFiber* fiber, u64 arg_on_run_to, u64* arg_on_r
...
+        if (fiber->addr_context) {
+            return ORBIS_FIBER_ERROR_INVALID;
+        }
+
+        fiber->addr_context = addr_context;
+        fiber->size_context = size_context;
+
+        fiber->context_start = addr_context;
+        fiber->context_end =
+            reinterpret_cast<void*>(reinterpret_cast<u64>(addr_context) + size_context);
```
roamic:
```suggestion
            reinterpret_cast<u8*>(addr_context) + size_context;
```

src/core/libraries/fiber/fiber.cpp
```diff
@@ -304,6 +342,42 @@ s32 PS4_SYSV_ABI sceFiberSwitch(OrbisFiber* fiber, u64 arg_on_run_to, u64* arg_o
...
```
roamic:
```suggestion
            reinterpret_cast<u8*>(addr_context) + size_context;
```

src/core/libraries/fiber/fiber.cpp
```diff
@@ -304,6 +342,42 @@ s32 PS4_SYSV_ABI sceFiberSwitch(OrbisFiber* fiber, u64 arg_on_run_to, u64* arg_o
         return ORBIS_FIBER_ERROR_PERMISSION;
     }
 
+    /* Caller wants to attach context and switch. */
+    if (addr_context != nullptr || size_context != 0) {
```
roamic, line 349: Is it possible to move this to a separate function to avoid code duplication in `sceFiberRunImpl` and `sceFiberSwitchImpl`?

roamic (approved):

PR #1952, opened 2024-12-28
Make buffer thread synchronization more granular

hspir404:
This improves performance for me.
Ryzen 5950x
nVidia 3080
64gb ram.

In the Bloodborne DLC, in the Nightmare Grand Cathedral, looking down the hill, I get 30fps before this change, and 40fps after.


zofrea1:
Significantly increased CPU bound performance on my system (5800X3d, 3090, 64gb RAM)
Pinning the game to 7 cores made it even better. Usually I only pin to 5, but with these changes the performance continued to scale up to 7 cores.

raphaelthegreat:
Looking at this code, it feels like we are over-complicating things a bit. I think you could investigate also using a read-write lock instead so multiple threads can read from cache but the slot vector is also protected from writes. Also can you try using a spin-lock instead of normal lock, that might prevent more expensive kernel sleeps.

The slot vector used for buffer storage is designed for memory locality so stores all buffers in a single array, and when that is full it allocates a new one and moves the buffers there. This means that if you store a pointer to an object and then call insert(), if that insert causes a reallocation, the stored pointer will be invalid and will crash. Using rwlock should prevent this tho. If that design becomes cumbersome a more std::dequeue like approach can be investigated that maintains that partial memory locality but allows for growing the container without invalidating or moving allocated elements.

Also would love to see some profiling data that inspired this solution, so we can better fine tune it

raphaelthegreat:
I suspect a lot of performance is lost by locking in InvalidateMemory as that can spam on a per-page basis. If you simply remove the lock on that function, what happens?

Missake212:
This build absolutely kills my performance, making the game stutter every 2/3 seconds, fps drop and making my gpu usage jump to 2% at the same interval.

image

diegolix29:
for me it incresede the performance on aareas that were having low fps due to large portions of the map loading on the distance. 

> This improves performance for me. Ryzen 5950x nVidia 3080 64gb ram.
> 
> In the Bloodborne DLC, in the Nightmare Grand Cathedral, looking down the hill, I get 30fps before this change, and 40fps after.


same as here but i was down from 90 to 50 now im down from 90 to 65 so 15 fps increase

hspir404:
> I think you could investigate also using a read-write lock instead so multiple threads can read from cache but the slot vector is also protected from writes. Also can you try using a spin-lock instead of normal lock, that might prevent more expensive kernel sleeps.

Will do. Will need a bit of time to investigate that.

> This means that if you store a pointer to an object and then call insert(), if that insert causes a reallocation, the stored pointer will be invalid and will crash.

Yeah I guess any buffers being created or destroyed while any are being read could cause a race condition on invalid slot_buffers entries (after they've been found, but before they get dereferenced). Or if multiple threads are modifying the container itself, they could dirty read and corrupt the container. That might explain why some people have run into regressions with this PR so far, and others haven't.

> Using rwlock should prevent this tho.

Did you mean a lock for the container itself? Right now inserts and deletes from the container aren't protected by any lock. I had figured you might mean instead of the mutexes on the buffers themselves. But maybe both need to happen (?).

> it feels like we are over-complicating things a bit

This change isn't super complicated from where I'm sitting. It is just a map of buffer ids to locks, so that if different buffers are being accessed, that they don't contend with one-another. This should reduce contention from there just being one big lock on the entire thing.

There's a little bit of complication around the null buffer, but other than that it should be pretty straight forward. I might have been able to make the code more clear on that fallback case.

I did some changes with renaming some functions because I initially tried using `FindBuffer` inside `InvalidateMemory` and found out it also created buffers. And creating a buffer at that spot causes problems, and also doesn't make sense to do there. So I renamed it to something that makes it clear it mutates, and created a non-mutating sibling.

I added a lock in `DeleteBuffer` because I was getting a regression of a lot of poly corruption, and adding that lock fixed it. I suspect the previous less granular mutex was sort of guarding against some things like that (though I think it was probably an imperfect defense, mitigated by the fact that it's likely to be locked a LOT).

>Also would love to see some profiling data that inspired this solution, so we can better fine tune it

Can do. The lock in `SynchronizeBuffer` was sticking out like a sore thumb. Specifically in the area of the game I mentioned.

I can provide a saved trace from my computer, and a savegame for the game state I'm sampling.

I had problems when using the sampling profiler with Tracy. I did get it sort of working, but the flame graph was looking a bit wrong, so I gave up on it and used my normal profiler (Superluminal). I can try again with Tracy, and if I have problems, I can try with Very Sleepy. Probably best to try to stick to FOSS with an open source project.

I suspect it might take a few days for me to get all the requested stuff ready.

> I suspect a lot of performance is lost by locking in InvalidateMemory as that can spam on a per-page basis. If you simply remove the lock on that function, what happens?

I'll try thinking through this a bit more and try variants like that as well. Will be good to see which locks actually are important.

hspir404:
...or did you mean rw/spinkocks INSTEAD of this granular locking?
I'll try that too, and of course I'm profiling and measuring all this stuff, and will do more play testing.
I still think the container probably needs a lock it doesn't yet have, but I'll think through that more.

raphaelthegreat:
I meant this yes

hspir404:
Didn't mean to close this PR, but I think #1973 will probably supercede it anyway, so might as well. Can always reopen this later if it for some reason ends up becoming the right approach.

I suspect this closed itself due to my force push on my fork, cause I switched this to using a branch instead of pushes to main.

PR #1959, opened 2024-12-28
image: Add Depth_MicroTiled and Display_MicroTiled tiling modes.

squidbus:
* Adds `Display_MicroTiled` (mode `0x9`), and gives `Depth_MicroTiled` (mode `0x5`) its own constant for completeness instead of mapping to `Texture_MicroTiled`.
* Adds these new tiling modes to size calculations and detiling, handled same as `Texture_MicroTiled`.

Needs validation this is the correct approach to handle these, I could not effectively debug the use of `Display_MicroTiled` in Knack Chapter 13-1 on my system.

### DIFF
--- a/src/video_core/amdgpu/resource.h
+++ b/src/video_core/amdgpu/resource.h
@@ -122,7 +122,9 @@ constexpr std::string_view NameOf(ImageType type) {
 
 enum class TilingMode : u32 {
     Depth_MacroTiled = 0u,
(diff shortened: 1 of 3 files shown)


niko1point0:
Here's the Knack savedata, so you can use chapter select to skip to 13-1
https://discord.com/channels/1080089157554155590/1258386683691274361/1322403350112632922

PR #1960, opened 2024-12-28
pthread: Change minimum stack for HLE to additional stack.

squidbus:
May fix some remaining stack space issues with HLE.

@Emulator-Team-2 Can you check with Gintama Rumble and cubeb audio with this?

### DIFF
--- a/src/core/libraries/kernel/threads/pthread.cpp
+++ b/src/core/libraries/kernel/threads/pthread.cpp
@@ -244,10 +244,9 @@ int PS4_SYSV_ABI posix_pthread_create_name_np(PthreadT* thread, const PthreadAtt
     new_thread->tid = ++TidCounter;
 
     if (new_thread->attr.stackaddr_attr == 0) {
-        /* Enforce minimum stack size of 128 KB */
-        static constexpr size_t MinimumStack = 128_KB;
-        auto& stacksize = new_thread->attr.stacksize_attr;
-        stacksize = std::max(stacksize, MinimumStack);
+        /* Add additional stack space for HLE */
+        static constexpr size_t AdditionalStack = 128_KB;
+        new_thread->attr.stacksize_attr += AdditionalStack;
     }
 
     if (thread_state->CreateStack(&new_thread->attr) != 0) {


raphaelthegreat:
Should we also consider moving audio processing into a separate host thread to alleviate stack issues from cubeb?

squidbus:
> Should we also consider moving audio processing into a separate host thread to alleviate stack issues from cubeb?

The actual processing done by cubeb is already in another thread but I suppose we could move the initialization to a thread as well, yes. The only part at issue here is the `cubeb_stream_init` call from `sceAudioOpen`, and only on Windows release builds for some reason. Although there's no guarantee we won't face this issue with other HLE so would like to have a more robust solution in place (at least until we can do something better like a separate HLE stack).

raphaelthegreat:
If its only the initialization code that is an issue, you can wrap it with std::async I think for easier effect
```cpp
std::async([&] {
// init
}).get();
```

squidbus:
> If its only the initialization code that is an issue, you can wrap it with std::async I think for easier effect

Yes that looks fairly simple but are we going to litter this pattern all around the codebase whenever this happens again?

raphaelthegreat:
No that would look kinda ugly. If this happens again we should look into HLE stacks I think

raphaelthegreat:
Note I agree with this PR as a change, this isnt against it

squidbus:
> Note I agree with this PR as a change, this isnt against it

Okay thanks, I think if this doesn't work (it could be a passed stack pointer as well) I will go with `std::async` or similar as you suggested.

Emulator-Team-2:
Fixes issue in `Gintama Rumble` with selected (cubeb) audio backend.

##

`Disgaea 4 Complete+` still hangs, when the opening video starts playing with selected (cubeb) audio backend. This doesn't happen with (sdl).

Log:
[shad_log.zip](https://github.com/user-attachments/files/18267854/shad_log.zip)

squidbus:
> `Disgaea 4 Complete+` still hangs, when the opening video starts playing with selected (cubeb) audio backend. This doesn't happen with (sdl).

This is a different and I'll have to look into it for another PR.

PR #1962, opened 2024-12-29
recompiler: handle reads of output variables in hull shaders

baggins183:
Adds Inst to handle reads from output control points in hull shaders.

Also clean up some random stuff like EmitGetAttribute which is mostly unused for hull/domain other than reading tess coords.


baggins183:
I just realized that barriers are present in the ps4 shaders, but based on the spec, S_BARRIER is meant to synchronize across waves in a threadgroup, not within waves. Ive only seen one hull shader so far that reads output control points, where it happens in the patchconstantfunc (see https://learn.microsoft.com/en-us/windows/win32/direct3d11/direct3d-11-advanced-stages-hull-shader-design) as it has been lowered to HW.
There's a barrier separating the per-control-point and patchconst phases. If there was an output control point read in the per-control-point phase, I dont know if there would be a barrier in the gcn.

To be extra safe, we might need to insert extra barriers. This is probably good enough for now though. I dont know if output control point reads are allowed in PSSL during the per-control-point function (based on what HLSL looks like)

PR #1965, opened 2024-12-29
config: Don't load config in the Emulator class

ngoquang2708:
Allows overriding of configs in frontends.

Fix set fullscreen not working when specified in the CLI.

Cc: @kalaposfos13.

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -45,10 +45,6 @@ Frontend::WindowSDL* g_window = nullptr;
 namespace Core {
 
 Emulator::Emulator() {
-    // Read configuration file.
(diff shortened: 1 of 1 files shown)


georgemoralis:
but where config is loaded now then?  Does it load to main ?

ngoquang2708:
Configs are already loaded in ``main`` of both frontend version.

kalaposfos13:
--fullscreen true wasn't working because it loaded the config in (one of the) main.cpp, then the fullscreen flag was overwritten with the argument, but then Emulator::Run also independently called Config::Load which re-overwritten the flag, so this PR just removes that to make CLI args work

PR #1969, opened 2024-12-29
renderer_vulkan: Render polygons using triangle fans.

squidbus:
Remembered today that polygons and triangle fans are exactly the same except in lines mode, so use that instead of a generated triangle list index buffer, which allows us to also easily support polygons for indexed and indirect draws and completely remove the remaining re-indexing code.

For lines mode, our current workaround doesn't work for that either anyway since we convert it to triangles. If we find a game using polygons in lines mode for some reason we will need an alternate solution for that.


squidbus:
Confirmed working

polybiusproxy (approved):

PR #1970, opened 2024-12-29
build: Update MoltenVK and fix missing add_dependencies for copy.

squidbus:
Missed this before and caught it in a CI run in my fork today without cache. Apparently this target dependency needs to be added for it to know how to make the dylib file, adding the file dependency to the copy command isn't enough alone.


squidbus:
Added in an update to MoltenVK while I'm at it, includes a fix for an incorrect texture atomics error on M1 family GPUs.

PR #1982, opened 2024-12-30
Speed up LiverpoolToVK::SurfaceFormat

hspir404:
In Bloodborne this shows up as the function with the very highest cumulative "exclusive time". This is true both in scenes that perform poorly, and scenes that perform well.

I took (approximately) 10s samples using an 8khz sampling profiler.

In the Nightmare Grand Cathedral (looking towards the stairs, at the rest of the level):
- Reduced total time from 757.34ms to 82.61ms (out of ~10000ms).
- Reduced average frame times by 2ms (though according to the graph, the gap may be as big as 9ms every N frames).

[00-savedata.zip](https://github.com/user-attachments/files/18276958/00-savedata.zip)
Before:
00-before
After:
00-after
Before:
00-before-profile
After:
00-after-profile

In the Hunter's Dream (in the spawn position):
- Reduced the total time from 486.50ms to 53.83ms (out of ~10000ms).
- Average frame times appear to be roughly the same.

[01-savedata.zip](https://github.com/user-attachments/files/18276964/01-savedata.zip)
Before:
01-before
After:
01-after
Before:
01-before-profile
After:
01-after-profile

These are profiles of the change vs the version currently in the main branch. These improvements also improve things in the `locking` branch. They might improve them even more in that branch, but I didn't bother keeping track of my measurements as well in that branch. I believe this change will still be useful even when that branch is stabilized and merged.

It could be there are other bottlenecks in rendering on the main branch that are preventing this code from being the critical path in places like the Hunter's Dream, where the frame times aren't as high. That might explain why the reduction in call times isn't resulting in a higher frame rate in every scene.


bigol83:
This improves Resident Evil 4 and Yakuza 0 too by 2 fps.

squidbus:
This is a nice perf improvement but I would prefer a solution that does not now involve maintaining two different copies of format mappings.

squidbus:
One suggestion would be if we could bitwise combine data and number formats into an indexed lookup to replace the `find_if`, instead of breaking out into a whole extra switch-based mapping.

hspir404:
Changed to a lookup table.

It adds a 4kb table (6 bits * 4 bits * 4 bytes per vulkan format), but seems to result in a smaller executable than when the switch statement was there (even though I verified the switch was only 975 bytes of code). Confused how that works, but apparently this is a 35MB executable anyway so I guess we're not sweating such things.

It also runs a gnat's hair faster than the switch statement did on my computer.

In the Nightmare Grand Cathedral from the save above:

after

Guess I should do a clang format to tidy this now.

src/video_core/renderer_vulkan/liverpool_to_vk.cpp
```diff
@@ -665,16 +665,36 @@ std::span<const SurfaceFormatInfo> SurfaceFormats() {
...
+static const size_t amd_gpu_number_format_bit_size = 4; // All values are under 16
+
+static size_t GetSurfaceFormatTableIndex(AmdGpu::DataFormat data_format, AmdGpu::NumberFormat num_format) {
+    DEBUG_ASSERT(data_format < 1 << amd_gpu_data_format_bit_size);
+    DEBUG_ASSERT(num_format < 1 << amd_gpu_number_format_bit_size);
+    size_t result = static_cast<size_t>(num_format) | (static_cast<size_t>(data_format) << amd_gpu_number_format_bit_size);
+    return result;
+}
+
+static auto surface_format_table = []() constexpr {
```
squidbus, line 681: Why do we need a separate table, can we not just make the existing table above indexed like this?

squidbus: Actually I guess this is fine, probably easier to define and iterate the format table with it sequential...

hspir404: This is not the only place the list is referenced, and I didn't want to mess with existing code. But I'll look into that.

squidbus: If something like a unordered_map where we can both lookup by this index and easily iterate the values is performant enough, that would be ideal for a single table. But otherwise we can probably just leave it as is for now, at least the lookup table is generated from it.

hspir404: Looks like std::unordered_map isn't something that can be spit out by a constexpr and live in rodata. It would have to be initialized at runtime. If it is built at runtime, it's just going to be two copies of the table anyway. Just one would live in the heap, and the other would live in rodata, instead of both living in rodata as the PR does as of my last change (I think).

The runtime std::unordered_map might be smaller. But this seems to be running quickly despite its size, and 4kb on disk seems like absolute peanuts compared to the 35MB state of shadps4.exe.

There are libraries that can do it at compile time (and also give perfect hashing), but I don't know if anything like it is currently in the project. One example: https://github.com/mapbox/eternal

src/video_core/renderer_vulkan/liverpool_to_vk.cpp
```diff
@@ -665,16 +665,40 @@ std::span<const SurfaceFormatInfo> SurfaceFormats() {
...
+    DEBUG_ASSERT(num_format < 1 << amd_gpu_number_format_bit_size);
+    size_t result = static_cast<size_t>(num_format) |
+                    (static_cast<size_t>(data_format) << amd_gpu_number_format_bit_size);
+    return result;
+}
+
+static auto surface_format_table = []() constexpr {
+    std::array<vk::Format, 1 << amd_gpu_data_format_bit_size * 1 << amd_gpu_number_format_bit_size>
+        result;
+    for (auto& entry : result) {
```
mailwl, line 684: maybe better use `result.fill(vk::Format::eUndefined);`

hspir404: hmm. Looks like it gives me errors saying it can't be a constexpr anymore if I do that. Tried the member and free function varieties and was getting errors. Think this may be a no-go as well.

mailwl: https://en.cppreference.com/w/cpp/container/array/fill
constexpr since C++20, strange it still not working on c++23. but ok

squidbus (approved):

PR #1984, opened 2024-12-30
Motion controls

kalaposfos13:
Very WIP, needs testing for both different games and different controllers.
If a game uses gyro and acceleration data, then we just pass those without problem, but so far, I haven't found how to handle orientation correctly, so for now it is a very basic implementation using the accelerometer to figure out which way is "down".
The code is still in a "just make it work" stage, I'll have to clean it up later. I mostly put this out here so people can test it and give feedback for now.


kalaposfos13:
I finally found something that actually works for calculating orientation and could be easily modified for this use case, but it still requires some tuning to get it right. Apart from that one thing, everything that I wanted to with this PR is ready.

georgemoralis:
looks fine
Fix the clang format issue and i think we can merge this

georgemoralis (approved):

kalaposfos13:
Done

Issue #1994, opened 2024-12-31
[APP BUG]: Build broken on gcc due to issues in some of the amdgpu files


C0rn3j:
### Checklist

- [X] I have searched for a similar issue in this repository and did not find one.
- [ ] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

I am guessing this was introduced by #1869

```cpp
[ 35%] Building CXX object CMakeFiles/shadps4.dir/src/qt_gui/elf_viewer.cpp.o
In file included from /home/c0rn3j/.cache/paru/clone/shadps4-git/src/shadps4/src/video_core/amdgpu/liverpool.h:25,
                 from /home/c0rn3j/.cache/paru/clone/shadps4-git/src/shadps4/src/video_core/renderer_vulkan/vk_presenter.h:8,
                 from /home/c0rn3j/.cache/paru/clone/shadps4-git/src/shadps4/src/imgui/renderer/imgui_core.cpp:17:
/home/c0rn3j/.cache/paru/clone/shadps4-git/src/shadps4/src/video_core/amdgpu/resource.h: In function ‘AmdGpu::CompMapping AmdGpu::RemapComponents(DataFormat, CompMapping)’:
/home/c0rn3j/.cache/paru/clone/shadps4-git/src/shadps4/src/video_core/amdgpu/resource.h:88:9: error: could not convert ‘{components.AmdGpu::CompMapping::b, components.AmdGpu::CompMapping::g, components.AmdGpu::CompMapping::r, components.AmdGpu::CompMapping::a}’ from ‘<brace-enclosed initializer list>’ to ‘AmdGpu::CompMapping’
   88 |         };
      |         ^
      |         |
      |         <brace-enclosed initializer list>
/home/c0rn3j/.cache/paru/clone/shadps4-git/src/shadps4/src/video_core/amdgpu/resource.h:96:9: error: could not convert ‘{components.AmdGpu::CompMapping::a, components.AmdGpu::CompMapping::b, components.AmdGpu::CompMapping::g, components.AmdGpu::CompMapping::r}’ from ‘<brace-enclosed initializer list>’ to ‘AmdGpu::CompMapping’
   96 |         };
      |         ^
      |         |
      |         <brace-enclosed initializer list>
[...]

PR #2000, opened 2024-12-31
Fix amdgpu & other issues

liberodark:
upstream some fixes from nixos package.


squidbus:
Can you explain what these changes are for?

tomboylover93:
It likely refers to issue #1994, which is about shadPS4 failing to compile with GCC on one of the AMDGPU related files.

squidbus:
Okay. I don't really understand the need for the swizzle init change, but for the `DataFormat()` change could you please change `DataFormat()` -> `GetDataFmt()` and `NumberFormat()` -> `GetNumberFmt()` to match what we have in other resources?

liberodark:
In the CompMapping structure, we had initialization issues with brace initialization lists. The fix uses explicit member construction to avoid compilation errors.

Also this fix issue on GCC14

squidbus (approved):

PR #2005, opened 2025-01-01
Fix for trophy decryption

psucien:
Fix for memory corruption and incorrect decryption sequence.
~~NOTE: Now user should store the key as a toml array, e.g. `TrophyKey = [1, 2, 3, 4, .. ]`~~

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -67,7 +67,7 @@ static int cursorHideTimeout = 5; // 5 seconds (default)
 static bool separateupdatefolder = false;
 static bool compatibilityData = false;
 static bool checkCompatibilityOnStartup = false;
-static std::string trophyKey = "";
+static std::vector<u8> trophyKey;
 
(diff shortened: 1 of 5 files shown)


psucien:
It is still not clear what to do if no key provided (some games may crash if no .xml is available). Also, additional handling for separate update folder is required

ElBread3:
> Also, additional handling for separate update folder is required

The path for the separate update is handled in Emulator::Run, where it will pass in the separated update folder's eboot assuming it exists, though I believe the trophy viewer isn't handled yet (which I will add to my next PR)

psucien:
> The path for the separate update is handled in Emulator::Run, where it will pass in the separated update folder's eboot assuming it exists

The problem happens when a patch doesn't have `trophy` folder, so we need to fall back to base game version

georgemoralis (approved):
it looks fine for now :)

PR #2011, opened 2025-01-01
gnmdriver: Implement neo mode differences.

squidbus:
Implements Neo mode differences for `gnmdriver`. Init sequences have a few differences, and a few command buffer writes have changed. For example, `IA_MULTI_VGT_PARAM` is now (conditionally) used as a uconfig register instead of context, and many draws have a bit added to the command from the flags.

All init sequences have been verified to equal that of the real gnm driver using both regular and neo modes, both with and without an extra ioctl condition passing.

Draft for now because there's still two `ioctl` based conditions I don't know the meaning of, currently using `bool` constants.


squidbus:
As far as I can tell the ioctl condition should be constant for retail consoles but I've left the other init sequences path in gated behind a constant in case we need it in the future. Should be ready for review.

PR #2014, opened 2025-01-02
Save Fixes VI

viniciuslrangel:


bigol83:
This fixes Elden Ring save creation too

PR #2016, opened 2025-01-02
Add support for true fullscreen

kalaposfos13:
SDL by default sets borderless fullscreen when calling SDL_SetWindowFullscreen, but you can manually set it to true fullscreen, if you want to. Now you can set it to true fullscreen in the config:
image
Ingame you can't switch between the modes yet, only in the launcher settings. I also didn't set up translations for it yet, as I don't have much experience with that.


src/sdl_window.cpp
```diff
@@ -92,8 +92,22 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameController* controller_
         UNREACHABLE_MSG("Failed to create window handle: {}", SDL_GetError());
     }
 
-    SDL_SetWindowMinimumSize(window, 640, 360);
```
ngoquang2708, line 95: Why is this need to be removed?

kalaposfos13: It didn't have to be, it was a mistake

DanielSvoboda:
I think I'll rearrange this screen...
edit: I opened a pr on your branch

Pipi86:
Shouldn't Fullscreen settings be moved into "Graphics" tab? as it affects game window and not emulator main window.

Foul-Tarnished:
Does borderless stays the default ?

Maybe add to the tooltip that borderless is better in every way, on modern OS
At least on windows 10/11

kalaposfos13:
@Foul-Tarnished yes, borderless is the default option.

Hermiten:
> Shouldn't Fullscreen settings be moved into "Graphics" tab? as it affects game window and not emulator main window.

I think so

georgemoralis:
this need rebase too as well :D

kalaposfos13:
@georgemoralis Done

PR #2019, opened 2025-01-02
Reduce USBD Log Spamming

Xphalnos:
Drastically reduces spam in the log of `[Lib.Usbd] <Error> usbd.cpp:sceUsbdHandleEventsTimeout:218: (STUBBED)called` which can be spammed more than 461,000 times!
Which loses a lot of performance and makes the startup of some games longer.
It also makes **big log files** (**34MB for 1 minute**).
I notably corrected an error because it says `(STUBBED)called` whereas in all the other libraries it says `(STUBBED) called`.

image1
image2

### DIFF
--- a/src/core/libraries/usbd/usbd.cpp
+++ b/src/core/libraries/usbd/usbd.cpp
@@ -10,327 +10,327 @@
 namespace Libraries::Usbd {
 
 int PS4_SYSV_ABI sceUsbdAllocTransfer() {
-    LOG_ERROR(Lib_Usbd, "(STUBBED)called");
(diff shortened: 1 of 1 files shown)


polybiusproxy:
I should add that the performance hit will still be present since the log still happens, it just doesn't get output. Given it does a bunch of allocations per log, games like Gran Turismo Sport or Driveclub can have quite a big performance drop.

Xphalnos:
> I should add that the performance hit will still be present since the log still happens, it just doesn't get output. Given it does a bunch of allocations per log, games like Gran Turismo Sport or Driveclub can have quite a big performance drop.

At least that will avoid having large log files

PR #2023, opened 2025-01-02
Fix returns on sceNpGetNpId and sceNpGetOnlineId

StevenMiller123:
Through some hardware testing, I've confirmed that both of these functions should return ORBIS_NP_ERROR_SIGNED_OUT on a signed out console.

Handling this properly avoids libSceHttp related exceptions in The Last of Us Remastered.

### DIFF
--- a/src/core/libraries/np_manager/np_manager.cpp
+++ b/src/core/libraries/np_manager/np_manager.cpp
@@ -972,11 +972,8 @@ int PS4_SYSV_ABI sceNpGetGamePresenceStatusA() {
 }
 
 int PS4_SYSV_ABI sceNpGetNpId(OrbisUserServiceUserId user_id, OrbisNpId* np_id) {
-    LOG_INFO(Lib_NpManager, "user_id {}", user_id);
-    const auto name = Config::getUserName();
-    std::memset(np_id, 0, sizeof(OrbisNpId));
-    name.copy(np_id->handle.data, sizeof(np_id->handle.data));
-    return ORBIS_OK;
+    LOG_DEBUG(Lib_NpManager, "user_id {}", user_id);
+    return ORBIS_NP_ERROR_SIGNED_OUT;
 }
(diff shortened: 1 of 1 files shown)


src/core/libraries/np_manager/np_manager.cpp
```diff
@@ -972,11 +972,8 @@ int PS4_SYSV_ABI sceNpGetGamePresenceStatusA() {
 }
 
 int PS4_SYSV_ABI sceNpGetNpId(OrbisUserServiceUserId user_id, OrbisNpId* np_id) {
-    LOG_INFO(Lib_NpManager, "user_id {}", user_id);
-    const auto name = Config::getUserName();
-    std::memset(np_id, 0, sizeof(OrbisNpId));
-    name.copy(np_id->handle.data, sizeof(np_id->handle.data));
-    return ORBIS_OK;
+    LOG_DEBUG(Lib_NpManager, "user_id {}", user_id);
+    return ORBIS_NP_ERROR_SIGNED_OUT;
```
georgemoralis, line 976: still in am not sure if not filling np_id can cause data corruption . Doesn't only return the SIGN_OUT code works without removing the rest of the code?

StevenMiller123: I ran a quick test on real hardware, np_id was garbage on there.
Additionally, this seems to work fine in the games I've tested.

georgemoralis: ok if you have tested it :D

chronicer:
Similar problem with NFS Hot Pursuit Remastered - many logs like this

[Lib.Net] <Error> net.cpp:sceNetEpollCreate:554: (STUBBED) called
[Lib.Net] <Error> net.cpp:sceNetEpollControl:549: (STUBBED) called


After 5 mins - 100 mb log

PR #2028, opened 2025-01-03
kernel: Check PSF for neo mode support.

squidbus:
* Check game PSF for neo mode support in `sceKernelIsNeoMode()`
* Make sure other parts of the code that check neo mode are going through `sceKernelIsNeoMode()` for consistency.
* Rename `Config::isNeoMode()` to `Config::allowNeoMode()` to avoid confusion of its purpose.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -100,7 +100,7 @@ void setTrophyKey(std::string key) {
     trophyKey = key;
 }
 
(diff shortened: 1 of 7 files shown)


squidbus:
Updated to have full PSF attributes struct cached at init, and fixed some separate update path issues with reading `param.sfo` and `pic1.png`.

PR #2032, opened 2025-01-03
renderer_vulkan: Don't crash if Depth Format is not recognized

Xphalnos:
While it can't override Depth Format, this allows games to not crash due to this error.
Fixes **PC Building Simulator** and probably other games.

**PC Building Simulator** with this PR:
Capture d'écran 2025-01-03 180659
Capture d'écran 2025-01-03 180723


polybiusproxy:
This is not the right thing to do and will likely crash drivers since you're returning the original format to use for depth.

Xphalnos:
> This is not the right thing to do and will likely crash drivers since you're returning the original format to use for depth.

The original format is not necessarily bad, it can be processed correctly

polybiusproxy:
> The original format is not necessarily bad, it can be processed correctly

That's on your end. It can crash for someone else with a different GPU, with a different driver. The unreachable is placed there for a reason, the depth image must have the correct format, otherwise there can be issues scaling from rendering errors to device losses (which will be difficult to debug). Because of the reasons mentioned, I will be closing this PR.

PR #2036, opened 2025-01-03
texture_cache: slight detilers refactoring

psucien:
- Meaningful detilers names;
- Got rid of monstrous switch-case and format conversion;
- Image's `guest_size_bytes` renamed to just `guest_size`


squidbus (approved):
Nice, this looks a lot cleaner

PR #1896, opened 2024-12-26
core: better memory configuration

polybiusproxy:
This PR aims to resolve the issues from https://github.com/shadps4-emu/shadPS4/issues/825 based on @red-prig's work on kernel reverse engineering.
May also fix games that check for direct memory sizes, like The Last of Us Part II and Farming Simulator 22.

Marking as draft as this needs to be checked for regressions.

PR #2037, opened 2025-01-03
Improved error handling in sceKernelAllocateDirectMemory

StevenMiller123:
Some games, most notably the Call of Duty series of games, intentionally make `sceKernelAllocateDirectMemory` calls that fail. To match hardware behavior, we should return `ORBIS_KERNEL_ERROR_EAGAIN` instead of throwing an assert.

This should fix games that still hit `Unable to find free direct memory area` asserts after https://github.com/shadps4-emu/shadPS4/pull/1896.

Credits to @red-prig and @raphaelthegreat for helping with the error cases.

This needs testing, as it could hide deeper issues in some games.


Missake212:
Pretty sure we already tested this on Discord but since these commits are more recent I thought I'd give them a try and send the logs here.

NBA2K15
[shad_log.txt](https://github.com/user-attachments/files/18335372/shad_log.txt)

NBA2K17
[shad_log.txt](https://github.com/user-attachments/files/18335390/shad_log.txt)

squidbus:
What is the current status of this PR?

StevenMiller123:
This does resolve the issues I saw in Call of Duty games, but leads to hangs or crashes in other games that hit the `Unable to find free direct memory area` asserts, like Hatsune Miku Project Diva X and the NBA games Missake212 tested. 
I'd like to see if I can figure out why those NBA games are behaving weirdly before marking this PR as ready, since the issues in those games seem to be related to `sceKernelAllocateDirectMemory`.

Missake212:
> This does resolve the issues I saw in Call of Duty games, but leads to hangs or crashes in other games that hit the `Unable to find free direct memory area` asserts, like Hatsune Miku Project Diva X and the NBA games Missake212 tested. I'd like to see if I can figure out why those NBA games are behaving weirdly before marking this PR as ready, since the issues in those games seem to be related to `sceKernelAllocateDirectMemory`.

If you want me to I could try to merge this PR locally and check through VS why it's hanging

Missake212:
Not sure if I did this properly but here's what I got, if you want to we can try continuing the debugging on Discord. First screen is 2k15 second one is 2k17
Capture d'écran 2025-01-18 152911
Capture d'écran 2025-01-18 152936

StevenMiller123:
With Hatsune Miku Project Diva X's issue resolved, the only remaining "blocker" is figuring out the issue with NBA 2K games. I should be able to look into that issue next week.

Missake212:
In case you're actually going to buy a 2K game for testing, I'm just reminding you that this error happens only on 15 and 17 for me, 14 doesn't behave the same and I didn't get the chance to test 16/18/19 (I own them but digitally) ... so they could all behave differently.

StevenMiller123:
Through some tests, it's become clear that the NBA 2K games are facing an issue unrelated to this, caused by internally considering a `physAddrOut` of 0 the result of a failed `sceKernelAllocateDirectMemory` call. Properly fixing this is, in my opinion, outside the scope of this PR since it's caused by other issues with shadPS4.

That said, I'm keeping this as a draft while I investigate some issues I've learned of in `DEAD OR ALIVE Xtreme 3 Scarlet`, which appear to unrelated to the NBA issues.

StevenMiller123:
The issue with `DEAD OR ALIVE Xtreme 3 Scarlet` is a result of some libSceHmd stubs. I'll be releasing a separate PR to add those in and fix the issue.

As far as I can tell, I've covered all the games I've been informed of that still had issues during `sceKernelAllocateDirectMemory` calls, so I think this PR is finally ready for review.

Randomuser8219:
> The issue with `DEAD OR ALIVE Xtreme 3 Scarlet` is a result of some libSceHmd stubs. I'll be releasing a separate PR to add those in and fix the issue.
> 
> As far as I can tell, I've covered all the games I've been informed of that still had issues during `sceKernelAllocateDirectMemory` calls, so I think this PR is finally ready for review.

VR exclusive games like Astro Bot Rescue Mission spam the libsceHmd stubs.

src/core/libraries/kernel/memory.cpp
```diff
@@ -26,26 +26,42 @@ u64 PS4_SYSV_ABI sceKernelGetDirectMemorySize() {
 
 int PS4_SYSV_ABI sceKernelAllocateDirectMemory(s64 searchStart, s64 searchEnd, u64 len,
                                                u64 alignment, int memoryType, s64* physAddrOut) {
-    if (searchStart < 0 || searchEnd <= searchStart) {
+    if (searchStart < 0 || searchEnd < 0) {
```
georgemoralis, line 29: hmm range issue?

PR #2044, opened 2025-01-04
input: Add missing poll lock

ngoquang2708:


Randomuser8219:
What does this fix?

ngoquang2708:
This function access data protected by ``m_mutex`` without locking it.

Issue #1263, opened 2024-10-06
Invalid attribute type Ubnorm


AboMedoz:
```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:418: Compiling fs shader 0xb585d534
[Debug] <Critical> spirv_emit_context.cpp:GetAttributeType:167: Unreachable code!
Invalid attribute type Ubnorm
```
here is the Shader Disassembly 
[ShaderDisasm.txt](https://github.com/user-attachments/files/17268909/ShaderDisasm.txt)

Hermiten:
Thanks for this but sometimes better to put it on discord aswell.
Maybe you can also report with your OS version, the game used, etc ...

AboMedoz:
> Thanks for this but sometimes better to put it on discord aswell.
> Maybe you can also report with your OS version, the game used, etc ...

The issue isn't solved why did you close it ?

Hermiten:
Lack of information, so nobody took the problem into account.
I open it again if you want to update it.

AboMedoz:
> Lack of information, so nobody took the problem into account.
> I open it again if you want to update it.

Issue reported and information were given on discord very long ago

PR #2047, opened 2025-01-04
video_core: Implement conversion for uncommon/unsupported number formats.

squidbus:
Extends previously added format remapping logic for number conversions. Initially supports Uscaled, Sscaled, and Ubnorm.

As part of this, resource tracking pass is slightly reorganized again to do a first pass of sharp tracking, then a second pass of all argument patching and reinterpretation. This is so that a) we still know whether images are storage, similar to why it was split up before, and b) we can emit the final instructions from tracking pass with conversion applied rather than having to use tricks to patch it in later.

### DIFF
--- a/src/shader_recompiler/frontend/translate/export.cpp
+++ b/src/shader_recompiler/frontend/translate/export.cpp
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "shader_recompiler/frontend/translate/translate.h"
+#include "shader_recompiler/ir/reinterpret.h"
 #include "shader_recompiler/runtime_info.h"
 
 namespace Shader::Gcn {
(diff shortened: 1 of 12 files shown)


AboMedoz:
Tested it with InFAMOUS which used Ubnorm now the Ubnorm error is solved with no Visual Difference
image
image

Hermiten:
No visual difference ?

AboMedoz:
> No visual difference ?

Sadly no, There isn't beside Ubnorm is only present in One Shader iirc

squidbus:
> Sadly no, There isn't beside Ubnorm is only present in One Shader iirc

Can check again as I added something missing, but like you said if it's just one shader it may be a harder to spot difference.

squidbus:
Confirmed the conversion is working now, ready for review. Further optimization left as a future exercise; working but sub-optimal is at least better than not working, and this is a fairly rare case. For example we could do conversion operations on whole vectors instead of components, but I was having trouble figuring out how to add this into the current IR in a clean way.

Hermiten:
Does this solve https://github.com/shadps4-emu/shadPS4/issues/1263 ?

AboMedoz:
> Does this solve #1263 ?

yes

PR #2052, opened 2025-01-04
CI: Switch to Windows 2025 & Reduce Warnings

Xphalnos:
- Upgrade to **Windows 2025** because `windows-latest` uses `windows-2022` by default
- Added `-DCMAKE_C_FLAGS_RELEASE=-Ofast` `-DCMAKE_CXX_FLAGS_RELEASE=-Ofast`  because by default it uses the `-O3` optimization (**-Ofast** uses the **-O3** optimizations but with additional ones)

Main:
Before1

This PR:
After1

- 3 Warnings less by replacing `ubuntu-latest` with `ubuntu-24.04`
CI-Warnings

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,14 +14,14 @@ env:
 
 jobs:
   reuse:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     continue-on-error: true
     steps:
     - uses: actions/checkout@v4
     - uses: fsfe/reuse-action@v5
 
   clang-format:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     continue-on-error: true
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


polybiusproxy:
Is there any specific reason to move to a runner image in public preview?

Xphalnos:
> Is there any specific reason to move to a runner image in public preview?

It contains more up-to-date packages for example LLVM and also it uses the latest Windows SDK

- Windows 2022 => 10.0.20348
- Windows 2025 => 10.0.26100

MajorP93:
There are multiple reasons why adding DCMAKE_C_FLAGS_RELEASE=-Ofast and DCMAKE_CXX_FLAGS_RELEASE=-Ofast is not a good idea:
- -Ofast compiler flag is deprecated in Clang 20
- Can introduce issues since some mathematical operations behave differently when this flag is added
- Setting DCMAKE_C_FLAGS_RELEASE (and the respective CXX variant) in this case will overwrite other compiler flags. The option is not being appended to compiler flags but rather replacing the whole line of compiler flags during compilation.

Xphalnos:
I trust you. So let's remove the optimizations.

PR #2068, opened 2025-01-05
Package windows symbols with actions builds

hspir404:
This commit packages up the symbol files (for Windows builds), so it is possible to debug a crash in official builds.

I packaged it up with the corresponding package. It increases the zip file size a lot, so maybe folks would prefer it to be packaged separately. Creating the PR to open up that discussion.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'push' }}
 
 env:
-  BUILD_TYPE: Release
+  BUILD_TYPE: RelWithDebInfo
 
 jobs:
   reuse:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,11 +94,18 @@ jobs:
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --parallel $env:NUMBER_OF_PROCESSORS
 
+    - name: Deploy and Package
+      run: |
+        mkdir upload
+        move build/shadPS4.exe upload
+        move build/shadPS4.pdb upload
+        Compress-Archive -Path upload/* -DestinationPath shadps4-win64-sdl-${{ needs.get-info.outputs.date }}-${{ needs.get-info.outputs.shorthash }}.zip
+
     - name: Upload Windows SDL artifact
       uses: actions/upload-artifact@v4
       with:
         name: shadps4-win64-sdl-${{ needs.get-info.outputs.date }}-${{ needs.get-info.outputs.shorthash }}
-        path: ${{github.workspace}}/build/shadPS4.exe
+        path: upload/
 
   windows-qt:
     runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


georgemoralis:
i think that anyone who knows to debug should be able to compile a debug build on his own

hspir404:
> i think that anyone who knows to debug should be able to compile a debug build on his own

It's not for that problem. It's for this problem:

"My game crashed"
"Cool, send me your crash dump"
"Here you go"
**dev opens the dump**

```
Shad.exe 0x498021f321
Shad.exe 0x4er81920
Shad.exe 0xbadfood
Shad.exe 0xdeadbeef
kernel32.dll 0x498021f321
nvidiadrivers.lol 0x498021f321
```

Consistent repros of game crashes can be difficult to come by. Players often don't know exactly what they did (and often aren't terribly technically savvy), and games often behave in non-deterministic ways that makes a crash dump the easiest and fastest way to actually get to the root of the problem.

I also have personally found github actions builds to be convenient sometimes, and then later wished I'd synced locally and built myself instead when I hit a crash and wanted to debug it. So even if it's not used for player crash dumps, it can be useful to (some) devs at times.

There are of course some negatives. Bigger download size if they're in the same zip. Having it in another separate zip might confuse people downloading files ("do I need the "symbols.zip" file?"). Or maybe github actions has limited storage space (I don't know the answer to this). It might also slightly slow down build times. And the project might not be in a phase that focusing on more rare and harder to repro crash fixes makes sense.

georgemoralis:
this will destroy updater and also it produces a lot bigger download .

hspir404:
@georgemoralis makes sense, with respect to it being in the same zip file.

how about a separate zip?
I think this will be useful. I think it will make crash repros a lot easier.
if you don't mind that idea, I can make a PR for that (or edit this PR to do that).

hspir404:
Sorry to spam you in replies on a closed PR, George.

I investigated a bit more, and the `RelWithDebInfo` CMake mode that I used in this PR adds some debugging flags and metadata to the executable as well as spitting out the PDB, when used with Clang. That's not really acceptable, IMO. I believe it doesn't do any of that junk with the MSVC compiler (which is why I suggested it in the first place). But MSVC is not the compiler this project uses or is going to use. So I think that will never be the right CMake mode.

I found some flags that can be passed to Clang that seem to give something close to a release build, and also spit out PDBs, but I have yet to verify that it's fully the same as a regular Release build. If shad were ever going to support shipping PDBs (even in a separate zip), that problem would have to be solved.

So even if you could be convinced that shipping PDBs in a separate zip were a good idea, the solution in this PR is still the wrong one, and it is still blocked on more investigation.

Here's the flags I found:
```
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /Zi")
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zi")

set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /DEBUG /OPT:REF /OPT:ICF")
set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} /DEBUG /OPT:REF /OPT:ICF")
set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /DEBUG /OPT:REF /OPT:ICF")
```

^ I have tested that those flags produce a binary pretty close (but not identical) to the one I get from the github automation build. I haven't yet done a local A/B comparison to verify that it's the exact same as a Release build.

(my local builds aren't binary identical to github's for some reason I have yet to figure out, despite using the same CMake flags. but I think a fully local A/B is going to be sufficient)

PR #2080, opened 2025-01-07
Unmap Fixes

StevenMiller123:
This PR fixes several memory issues I've noticed.
1. When games try to unmap reserved memory, this would cause address_space asserts on Windows. This is because reserved memory was never mapped in address_space to begin with. My fix is to only unmap in address_space if the unmapped VMA had `type != VMAType::Reserved`.
2. Some games also attempt to unmap free memory. This causes address_space asserts on Windows, and also messes up later memory use calculations. My solution is adding an early return if the VMA to unmap has `type == VMAType::Free`.
3. `UnmapMemoryImpl` doesn't properly handle pooled memory. PoolReserved memory can either be reserved or decommitted memory, either case isn't mapped in address_space. Additionally, Pooled memory is GPU mapped. My solution is adding those relevant checks so we don't run into any strange Windows-specific issues.
4. When games overwrite memory with reserved pages, our check for `vma.type == VMAType::Free` would fail. This is because, while the `UnmapMemoryImpl` call modifies vma_map, the vma retrieved before that call is not modified. To fix this, I add an extra FindVMA call after the unmap.

This should fix cases of:
```
[Debug] <Critical> address_space.cpp:operator():268: Assertion Failed!
Invalid address/size given to unmap.
```
And some cases of:
```
[Debug] <Critical> memory.cpp:operator():217: Assertion Failed!
```

Credits to @red-prig for providing relevant information.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -395,12 +395,16 @@ s32 MemoryManager::UnmapMemoryImpl(VAddr virtual_addr, size_t size) {
     ASSERT_MSG(vma_base.Contains(virtual_addr, size),
                "Existing mapping does not contain requested unmap range");
 
+    const auto type = vma_base.type;
+    if (type == VMAType::Free) {
+        return ORBIS_OK;
+    }
+
     const auto vma_base_addr = vma_base.base;
     const auto vma_base_size = vma_base.size;
     const auto phys_base = vma_base.phys_base;
     const bool is_exec = vma_base.is_exec;
     const auto start_in_vma = virtual_addr - vma_base_addr;
-    const auto type = vma_base.type;
     const bool has_backing = type == VMAType::Direct || type == VMAType::File;
     if (type == VMAType::Direct) {
         rasterizer->UnmapMemory(virtual_addr, size);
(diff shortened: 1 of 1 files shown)


Missake212:
The Last of Us Remastered goes from crashing on launch to now hanging on a black screen.
[shad_log.txt](https://github.com/user-attachments/files/18335498/shad_log.txt)

StevenMiller123:
@Missake212 Pretty sure TLOU Remastered hangs on main too, this PR shouldn't impact it. Could you test main to verify?

Missake212:
> @Missake212 Pretty sure TLOU Remastered hangs on main too, this PR shouldn't impact it. Could you test main to verify?

TLOU Remastered crashes on main for me, here's the log
[shad_log.txt](https://github.com/user-attachments/files/18335802/shad_log.txt)

StevenMiller123:
Well, an improvement is an improvement, I won't complain about that.

Inui-senpai:
DDLC+ now crashes on `Existing mapping does not contain requested unmap range`.
[shad_log.txt](https://github.com/user-attachments/files/18338247/shad_log.txt)

georgemoralis:
let's get dangerous :D

PR #2081, opened 2025-01-07
Updated Swedish translation with additional strings

yeager:


DanielSvoboda:
You removed <source>ps4proCheckBox</source> , I understand that this option has currently been removed so that laymen do not press it without knowing what it does, but I believe that this will be reactive in the future, so I think it should not be removed from the translation file.
image

Several tags were changed places, for what reason? Now they are not in the same order as the other translation files.
I also noticed that the 2 tab space formatting was removed.
image

You added these 2 new texts: Release, Nightly, but they should not be translated into any language, as the current code does not support other words, I know it was translated into the same word, but these lines should not exist.
image

I would recommend you download this version below which is correct and just make the changes to the texts with these additional strings.
https://github.com/yeager/shadPS4/blob/3224d581054d90000952a0dd04f2fcff2f628e2a/src/qt_gui/translations/sv.ts

yeager:
You should always treat TS files as automatically generated files from lupdate or lconvert.
Never manually edit these. Use lrelease to create the QM file.
Formatting is not an issue. Use Qt Linguist to translate the TS


Updated the Swedish translations using lupdate to found additional strings
Remove "-no-obsolete" to keep old strings.

cd src/qt_gui/translations

lupdate ../../../../shadPS4/ -tr-function-alias QT_TRANSLATE_NOOP+=TRANSLATE,QT_TRANSLATE_NOOP+=TRANSLATE_SV,QT_TRANSLATE_NOOP+=TRANSLATE_STR,QT_TRANSLATE_NOOP+=TRANSLATE_FS,QT_TRANSLATE_N_NOOP3+=TRANSLATE_FMT,QT_TRANSLATE_NOOP+=TRANSLATE_NOOP,translate+=TRANSLATE_PLURAL_STR,translate+=TRANSLATE_PLURAL_FS -no-obsolete -locations none -source-language en -ts sv.ts

PR #2082, opened 2025-01-07
shader_recompiler: Fix V_CMP_U_F32

squidbus:
Spotted this in CUSA28193, which has shader code like `V_CMP_U_F32(value, value) ? 0 : value`. There is a typo in the AMD ISA docs for GCN; `V_CMP_U_F32` is supposed to be`D.u = (isNaN(S0) || isNaN(S1))`, not `D.u = (!isNaN(S0) || !isNaN(S1))`. This is supported by checking LLVM and Mesa compilers instruction usage, as well as docs for future ISA revisions listing it as `D[threadId] = (isNan(S0) || isNan(S1))`.


georgemoralis:
nice catch!

red-prig:
Isn't this a direct comparison to the `OpUnordered` SPIRV instruction?

squidbus:
> Isn't this a direct comparison to the `OpUnordered` SPIRV instruction?

`OpUnordered` requires compute kernel execution model.

PR #2083, opened 2025-01-07
shader_recompiler: Improvements to array and cube handling.

squidbus:
Enhances handling of array and cube images:
* Image sharps may have an array type, but in the actual shader the image is accessed with the array flag not set. In this case we should override the image to the proper non-array variant and avoid indexing as an array.
  * Fixes missing rendering in CUSA28193. Multiple shaders would have a 2D texture array bound but address it as a single layer, with the register used for the array index being invalid.
* Handles array flag for all MIMG instructions instead of just sampling instructions.
* Converts cubes to 2D arrays. This helps deal with cases where layers are not a multiple of 6 for whatever reason, and also enables derivatives and offsets on cube images, which are not possible with Vulkan cube images.

### DIFF
--- a/src/shader_recompiler/frontend/translate/vector_alu.cpp
+++ b/src/shader_recompiler/frontend/translate/vector_alu.cpp
@@ -1042,20 +1042,75 @@ void Translator::V_MAD_U32_U24(const GcnInst& inst) {
     V_MAD_I32_I24(inst, false);
 }
 
+IR::F32 Translator::SelectCubeResult(const IR::F32& x, const IR::F32& y, const IR::F32& z,
+                                     const IR::F32& x_res, const IR::F32& y_res,
+                                     const IR::F32& z_res) {
+    const auto abs_x = ir.FPAbs(x);
+    const auto abs_y = ir.FPAbs(y);
+    const auto abs_z = ir.FPAbs(z);
+
+    const auto z_face_cond{
+        ir.LogicalAnd(ir.FPGreaterThanEqual(abs_z, abs_x), ir.FPGreaterThanEqual(abs_z, abs_y))};
+    const auto y_face_cond{ir.FPGreaterThanEqual(abs_y, abs_x)};
+
+    return IR::F32{ir.Select(z_face_cond, z_res, ir.Select(y_face_cond, y_res, x_res))};
+}
+
 void Translator::V_CUBEID_F32(const GcnInst& inst) {
-    SetDst(inst.dst[0], GetSrc<IR::F32>(inst.src[2]));
+    const auto x = GetSrc<IR::F32>(inst.src[0]);
+    const auto y = GetSrc<IR::F32>(inst.src[1]);
+    const auto z = GetSrc<IR::F32>(inst.src[2]);
+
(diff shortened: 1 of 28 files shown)


squidbus:
Fixed a cube array crash, so no longer draft.

StevenMiller123:
Improves Final Fantasy XV Episode Duscae (previously crashed on `[Debug] <Critical> resource_tracking_pass.cpp:operator():570: Assertion Failed!`, now crashes on `[Debug] <Critical> tile_manager.cpp:operator():262: Assertion Failed!`)
[PR log.txt](https://github.com/user-attachments/files/18336096/PR.log.txt)
[main log.txt](https://github.com/user-attachments/files/18336097/main.log.txt)

src/shader_recompiler/info.h
```diff
@@ -73,9 +73,44 @@ struct ImageResource {
     bool is_read{};
     bool is_written{};
 
+    [[nodiscard]] AmdGpu::ImageType GetBoundType(const AmdGpu::Image& image) const noexcept {
```
raphaelthegreat: Maybe these functions are better placed in AmdGpu::Image and pass is_array as a bool argument?

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -1042,20 +1042,75 @@ void Translator::V_MAD_U32_U24(const GcnInst& inst) {
...
+    const auto z_face_cond{
+        ir.LogicalAnd(ir.FPGreaterThanEqual(abs_z, abs_x), ir.FPGreaterThanEqual(abs_z, abs_y))};
+    const auto y_face_cond{ir.FPGreaterThanEqual(abs_y, abs_x)};
+
+    return IR::F32{ir.Select(z_face_cond, z_res, ir.Select(y_face_cond, y_res, x_res))};
+}
+
 void Translator::V_CUBEID_F32(const GcnInst& inst) {
-    SetDst(inst.dst[0], GetSrc<IR::F32>(inst.src[2]));
+    const auto x = GetSrc<IR::F32>(inst.src[0]);
```
raphaelthegreat, line 1061: Might be good to add these as black-box IR opcodes and move this calculation to the backend, so SPV_AMD_gcn_shader can be used to emit the HW cube instructions on appropriate gpus. Should also help having something to compare to in case this has bugs

squidbus: I can do that, but it will mean that any backend will have to repeat implementing these calculations.

raphaelthegreat: Maybe you can do the check in IR level then and have these special IR ops only work when the extension is available?

squidbus: Done.

DanielSvoboda:
This also makes it possible to see in the games Elden Rind and SONIC X SHADOW GENERATIONS, as currently everything is black.

ryzendew:
This fixes lighting on the Guardian 
image

Here is before 
image

PR #2085, opened 2025-01-07
Move release info into metainfo.xml

fpiesche:
This works around a problem where Flatpak has incomplete support for external `releases.xml` files. `flatpak-builder` will correctly accept these being present as long as they match the FreeDesktop specification but spits out a warning about the presence of the file, while removing it will cause `flatpak-builder-lint` to fail (correctly) if the `metainfo.xml` specifies an external release.


georgemoralis:
and what does is serves as change?

jardon:
this relates to https://github.com/flathub/flathub/pull/5997

jardon:
> and what does is serves as change?

this is required by flathub as flatpak currently doesnt support splitting the releases out of the metainfo file.

fpiesche:
> and what does is serves as change?

sorry, I'd added this to the original commit message but then overwrote that. Added in the description.

ngoquang2708:
Need to remove from CMake too.

Issue #2086, opened 2025-01-07
[Feature Request]: Minor interface improvements for Steam Deck users.


GHU7924:
### Checklist

- [X] I have searched for a similar issue in this repository and did not find one.

### Description

Resizing the window and improving the functions of the "Save" button in the "Patches" window for ease of use by Steam Deck users.

### Reason

For more convenient use of the emulator in game mode on Steam Deck.

### Examples

At the moment, hovering over different interface elements causes the window to change its size depending on the description. It might be worth adding 5 lines to the default description (I don't think I've seen more). In game mode, the windows look like this.

![sdi](https://github.com/user-attachments/assets/22b1db4a-6e79-45d9-977b-d41f798840c3)

As for the "Patches" window, it looks like this.

![20250107153739_1](https://github.com/user-attachments/assets/97b51011-47dd-4ea3-b7d6-cae32c1b1cf5)

As you may have noticed, the window frame is cut off in game mode. I can close this window through the Steam menu, but it would be more convenient if the "Close" button was added or the "Save" button was changed to "Save & Exit" (that is, the window was closed after saving).

P.S. If anyone else wants to comment on improving the emulator interface in Steam Deck game mode, then please add this information, perhaps I don't know something or haven't come across it.

tomboylover93:
I use Bazzite on my main computer, which includes Steam Deck Gaming Mode, so I could test these issues for myself and fix them. That, and the Settings window needs some reorganizing.

GHU7924:
I also noticed that the AutoUpdater window also has a slight size change. It seems to happen after clicking the "Update" button.

tomboylover93:
Can you send a screenshot for comparison?

GHU7924:
@tomboylover93 I'm apologize.
In fact, it's perfectly acceptable there, you can ignore it. I took screenshots of all the steps to make sure of this.
I remember that there were some changes, but they turned out to be completely insignificant.
![20250108165304_1](https://github.com/user-attachments/assets/3015a0ba-19f1-41b5-8a63-5dd3e03c8d18)
![20250108165316_1](https://github.com/user-attachments/assets/d7caacde-8998-414d-ab1b-4d80cc2c0438)
![20250108165442_1](https://github.com/user-attachments/assets/82c3a4b0-e431-4c5b-a9f9-dcf238473c86)
![20250108165755_1](https://github.com/user-attachments/assets/8f739dab-fde5-473c-892d-4151ae7d3528)

[...]

PR #2103, opened 2025-01-09
qt: Improve user experience on Steam Deck and window managers

tomboylover93:
This aims to fix issues with the Qt interface on (mostly Linux) desktops that hide the app's titlebar and buttons, such as the Steam Deck's Gaming Mode, which uses Gamescope (see #2086), and the aforementioned window managers. They range from minor quirks to issues that make it outright impossible to perform some actions solely using the GUI, and some examples are:

- the lack of a "Close" button for the Cheats/Patches window, and
- the Settings window resizing itself when hovering over something whose description text occupies more screen real estate than 70px:

https://github.com/user-attachments/assets/4804a2e6-7746-4408-8b68-caecf5b3e816

The latter causes weird behavior outside of those desktop interfaces, as when maximizing the Settings window on any desktop environment the text box resizing mechanism does not respect the actual window size, which leads to this:

image

I'll leave it as a draft for now because I want more input from people using said desktops and other operating systems than Linux, since there might be something I missed or that needs to be changed.


DanielSvoboda:
Changing bottomMargin from 11 to 190 in UpdateLayout makes it not so stretched

https://github.com/shadps4-emu/shadPS4/blob/0c78d0c1c7e7100d0cb7e79318c8cf9f6fb2604e/src/qt_gui/settings_dialog.ui#L437

image

GHU7924:
Now the window does not change size and works properly.
The Patches window closes after saving.
It has clearly improved, but there is still such a moment (it's not critical, but it is):
Cheats

@tomboylover93 To test and get an approximate experience of using the emulator on Steam Deck,
you should change the resolution for the virtual machine.
The Steam Deck has a 16:10 aspect ratio and a resolution of 1280x800.

tomboylover93:
I don't think I could change much about the drop-down menu resizing the window, that's just how Gamescope is. Though I could make it open upwards instead of downwards.

GHU7924:
How much and what other edits do you want to make in this PR?

tomboylover93:
Most of the issues I had were already fixed with the initial commits, I'm leaving it as a draft because maybe someone noticed something that needs to be fixed that I didn't.

GHU7924:
Since no one else is writing, I think we can delete the draft and request a review.

roamic (approved):

PR #2104, opened 2025-01-09
renderer_vulkan: Simplify vertex binding logic and properly handle null buffers.

squidbus:
* Simplifies vertex binding logic by using common code for creating the required attribute and binding structures, for both the dynamic vertex input and pipeline vertex input paths.
* Handles unbound vertex buffers using null buffer instead of skipping binding, which could lead to missing vertex attribute bindings for the shader.

Fixes a crash in CUSA05637 halfway through the 7th stage.

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -10,13 +10,13 @@
 #include "video_core/amdgpu/liverpool.h"
 #include "video_core/buffer_cache/buffer_cache.h"
 #include "video_core/renderer_vulkan/liverpool_to_vk.h"
+#include "video_core/renderer_vulkan/vk_graphics_pipeline.h"
 #include "video_core/renderer_vulkan/vk_instance.h"
 #include "video_core/renderer_vulkan/vk_scheduler.h"
 #include "video_core/texture_cache/texture_cache.h"
 
 namespace VideoCore {
 
-static constexpr size_t NumVertexBuffers = 32;
 static constexpr size_t GdsBufferSize = 64_KB;
 static constexpr size_t StagingBufferSize = 1_GB;
 static constexpr size_t UboStreamBufferSize = 64_MB;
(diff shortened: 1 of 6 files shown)


kvicken:
Graphical regression in Bloodborne started with this build.

Player shadows are broken at the beginning of the game, it's fine in Central Yharnam.

With this build https://github.com/shadps4-emu/shadPS4/actions/runs/12706265313

https://github.com/user-attachments/assets/714dd2d0-6bfd-4a90-8177-a68970f8a89e

Wiith the previous main build: 

https://github.com/user-attachments/assets/fc052765-bcd2-4deb-91eb-15889c8ced70

Log: 
[shad_log.txt](https://github.com/user-attachments/files/18386906/shad_log.txt)

squidbus:
> Graphical regression in Bloodborne started with this build.

Your first video does not play for me.

squidbus:
Also there was an error in this build fixed by https://github.com/shadps4-emu/shadPS4/commit/6ec68f66a9d3c5c9f3124373fffcf5cfe6a65910 so it cannot be used for testing.

kvicken:
Still happens in the latest main, does this one work for you: 


https://github.com/user-attachments/assets/98e7e8d9-5271-4447-84bd-f5df51e78c10

squidbus:
I think you are looking at the wrong PR here. I reverted to an older main and then applied only this change and it works fine.

kvicken:
hmm that video above is from the latest main, maybe is something on my end?

squidbus:
Yes it happens on the latest main, but not from this PR

squidbus:
@kvicken Should be fixed by https://github.com/shadps4-emu/shadPS4/commit/5810c88c00390026ecc33a4244190843afe19b94 (at least on your GPU)

kvicken:
Yes, using https://github.com/shadps4-emu/shadPS4/commit/5810c88c00390026ecc33a4244190843afe19b94 fixed the shadows!

Captura de tela 2025-01-11 172204

Issue #2048, opened 2025-01-04
[GAME BUG]: Performance drop on Steam Deck on pre release, not present on v0.5.0


DiGik05:
### Checklist (we expect you to perform these steps before opening the issue)

- [X] I have searched for a similar issue in this repository and did not find one.
- [X] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.
- [X] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [X] I have disabled all patches and cheats and the issue is still present.
- [X] I have all the required [system modules](https://github.com/shadps4-emu/shadps4-game-compatibility?tab=readme-ov-file#informations) installed.

### Describe the Bug

There is a significant drop in performance on Steam Deck that is present on the pre-release version and not present on v0.5.0 - codename squidwars. Bloodborne used to run at 25-30 fps on V0.5.0 and now with the latest pre release the game runs at sub 20fps and stutters a lot, fps can get to 10fps. Running game again with Release version 0.5.0 fixes all issues. 

### Reproduction Steps

On Steam Deck open game, run around central Yharnam. Game runs at around 30fps on V0.5.0 but now with latest pre release it can get to 10fps, stutters a lot.

### Specify OS Version

SteamOS 3.6.20

### CPU

Zen 2 4c/8t, 2.4-3.5GHz

### GPU

8 RDNA 2 CUs, 1.6GHz (1.6 TFlops FP32)

### Amount of RAM in GB

16GB

### Amount of VRAM in GB

4GB

### Logs

NA

JvictorVentura:
I think you have the same issue i have, before the #1973 PR, I also had higher fps than now. I tested the builds before that PR and all of them have better performance. I don't have the same problem on Windows 10, only on my Linux Mint 22. But I've seen people with Arch Linux that didn't had this problem, so I don't think this is an issue on all Linux distros.

mlgprorektm8:
same issue, any updates?

SleepingSnakezzz:
Please attach log files of both versions (pre-release and regular release).

DiGik05:
> Please attach log files of both versions (pre-release and regular release).

Hello, can you let me know where the log file can be found? Thank you.

SleepingSnakezzz:
shadPS4 installation folder -> `user` folder -> `log` folder.

kalaposfos13:
[...]

PR #2105, opened 2025-01-09
video_core: Use adaptive mutex on Linux

ngoquang2708:
Fix performance regression with #1973 on SteamDeck (https://github.com/shadps4-emu/shadPS4/issues/2048).

Tested with reduced resolution to reduce GPU load and vblank divider set to 2.

# SteamDeck

main

pm+wm

# Laptop

https://github.com/user-attachments/assets/6bb79808-4dea-4c1e-b596-8d11d5a8652d

https://github.com/user-attachments/assets/ede4ebbf-bd15-4ed7-b024-7487cae36d05

### DIFF
--- a/src/common/adaptive_mutex.h
+++ b/src/common/adaptive_mutex.h
@@ -1,0 +1,27 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#ifdef __linux__
+#include <pthread.h>
+#endif
+
+namespace Common {
+
+#ifdef PTHREAD_ADAPTIVE_MUTEX_INITIALIZER_NP
+class AdaptiveMutex {
+public:
+    void lock() {
+        pthread_mutex_lock(&mutex);
+    }
+    void unlock() {
+        pthread_mutex_unlock(&mutex);
+    }
+
+private:
+    pthread_mutex_t mutex = PTHREAD_ADAPTIVE_MUTEX_INITIALIZER_NP;
+};
+#endif // PTHREAD_ADAPTIVE_MUTEX_INITIALIZER_NP
+
+} // namespace Common
--- a/src/video_core/buffer_cache/word_manager.h
+++ b/src/video_core/buffer_cache/word_manager.h
@@ -8,6 +8,9 @@
 #include <span>
 #include <utility>
 
+#ifdef __linux__
+#include "common/adaptive_mutex.h"
+#endif
 #include "common/spin_lock.h"
 #include "common/types.h"
 #include "video_core/page_manager.h"
@@ -272,7 +275,11 @@ private:
         }
     }
 
+#ifdef PTHREAD_ADAPTIVE_MUTEX_INITIALIZER_NP
+    Common::AdaptiveMutex lock;
+#else
     Common::SpinLock lock;
+#endif
     PageManager* tracker;
     VAddr cpu_addr = 0;
     WordsArray cpu;
--- a/src/video_core/page_manager.h
+++ b/src/video_core/page_manager.h
@@ -5,6 +5,9 @@
 
 #include <memory>
 #include <boost/icl/interval_map.hpp>
+#ifdef __linux__
+#include "common/adaptive_mutex.h"
+#endif
 #include "common/spin_lock.h"
 #include "common/types.h"
 
@@ -36,7 +39,11 @@ private:
     std::unique_ptr<Impl> impl;
     Vulkan::Rasterizer* rasterizer;
     boost::icl::interval_map<VAddr, s32> cached_pages;
+#ifdef PTHREAD_ADAPTIVE_MUTEX_INITIALIZER_NP
+    Common::AdaptiveMutex lock;
+#else
     Common::SpinLock lock;
+#endif
 };
 
 } // namespace VideoCore


kalaposfos13:
I tested in 4 different games: Bloodborne, GRR, GR2 and Nier: Automata, and this PR had no major change on the last 3, but it made Bloodborne run at 15-20 FPS in the Hunter's Dream, down from a stable 60 on main (tested 3 times, so it's consistent). It has to be noted, though, that even on main, the framerate also deteriorates after a few minutes to the same level (and it happens for all games, as well)
Specs: 3060 Laptop GPU (6GB), 12700h laptop CPU, 16 GB RAM, Linux Mint 22 with KDE Plasma 5

GHU7924:
I tested your PR on Steam Deck. 
I must say right away that I used patches, but according to the idea, these patches should not greatly affect the game (except for one thing, the resolution patch).
I do not know how it will be in the future, but now playing without patches on the Steam Deck is spoiling your impression.

Играл я на 12W. Vblank = 2. Profile 30 fps (90 Hz). Play Time = 1 h 30 min.
I used the following patches:


Here are some screenshots:





In general, the game is trying to go at about 30 frames, but due to delays, the fps is drop.
The performance has improved and become more stable.

DemoJameson:
I got worse performance on the PR
```
CPU: Intel 13600kf
GPU: NVIDIA 4080S
RAM: 32GB
OS : Windows 11 24H2
```

Main:
main

PR:
pr

JvictorVentura:
This PR fixed the performance regression for me. My FPS has increased to 50 from 17.
CPU: Intel i5-9400F (6) @ 4.100GHz 
GPU: GTX 1050 Ti 4GB VRAM
OS: Linux Mint 22 MATE
RAM: 16GB

diegolix29:
For me also reduce the performance by 8-10 fps overrall
intel 12900k i9
3080 12gb vram
64 RAM
windows 10

smiRaphi:
I went from 125-135fps to 120-125 with this pr
Ryzen 9 7950X3D
RTX 4090
96GB RAM
win 11

ngoquang2708:
Some interesting test results. So basically:
 * Using mutex instead of spinlock help weaker CPU, especially APU, allow them to "rest" more without taking too much power, which is shared with the GPU.
 * More powerful CPU can just "brutefore" its way to get more frames with spinlock.
 
I wonder if we can get a middle ground of this, something that can spin lock for a number of times, then sleep, like a adaptive mutex?

ngoquang2708:
I have change this PR to use adaptive mutex on Linux so this should not affect other OSes. Need retest from Linux users.

DiGik05:
> I have change this PR to use adaptive mutex on Linux so this should not affect other OSes. Need retest from Linux users.

I would like to test it on SteamOS, but I am not sure how to get the AppImage to do that. Can you let me know?

JvictorVentura:
> This PR fixed the performance regression for me. My FPS has increased to 50 from 17. CPU: Intel i5-9400F (6) @ 4.100GHz GPU: GTX 1050 Ti 4GB VRAM OS: Linux Mint 22 MATE RAM: 16GB

I didn't see any difference in the performance, or at least I didn't notice any.

JvictorVentura:
> > I have change this PR to use adaptive mutex on Linux so this should not affect other OSes. Need retest from Linux users.
> 
> I would like to test it on SteamOS, but I am not sure how to get the AppImage to do that. Can you let me know?

On the "actions" tab look for "video_core: Use adaptive mutex on Linux", click on the most recent one, then download the linux-qt version.

DiGik05:
[shad_log.txt](https://github.com/user-attachments/files/18382453/shad_log.txt)
I can confirm performance improvement on Steam deck.

nr1971:
While I'm not seeing any increase in FPS this really helps smooth out my frame rate by getting rid of the low dips. For example while standing at the Central Yarnham lampost my fps would hover around 65 but drop to 52-53 every 5 seconds or so. This is gone now.
Ryzen 5600x, RX 6700xt, 32GB ram, Ubuntu 24.04

Missake212:
Tested a bit in Central Yharnam comparing latest main with this, didn't notice any performance difference on Windows like other people said, think with a little more testing this might be good to go.

ngoquang2708:
> Tested a bit in Central Yharnam comparing latest main with this, didn't notice any performance difference on Windows like other people said, think with a little more testing this might be good to go.

This PR now only affect Linux.

roamic:
@ngoquang2708 AFAIK we have adaptive mutex implementation for windows in `src/core/libraries/kernel/threads/mutex.cpp` You can try to use it.

ngoquang2708:
> @ngoquang2708 AFAIK we have adaptive mutex implementation for windows in `src/core/libraries/kernel/threads/mutex.cpp` You can try to use it.

Isn't that for guest side use? This PR use adaptive mutex for host side, only for Linux though, since I don't see any complains about performance regression from Windows users.

roamic:
> > @ngoquang2708 AFAIK we have adaptive mutex implementation for windows in `src/core/libraries/kernel/threads/mutex.cpp` You can try to use it.
> 
> Isn't that for guest side use? This PR use adaptive mutex for host side, only for Linux though, since I don't see any complains about performance regression from Windows users.

Ah, you're right, my bad.

ngoquang2708:
@kalaposfos13 Can you help retest new changes on your system?

kalaposfos13:
The performance is now yet again similar to main. Side note, I was able to significantly lessen the performance degradation and lag spikes by manually installing the 565 Nvidia drivers (Mint/apt only comes with 550), but it didn't completely solve the problem.

ngoquang2708:
@kalaposfos13 Thanks.

squidbus (approved):

PR #2102, opened 2025-01-08
GUI: Open Log Location - Button

DanielSvoboda:
As suggested in [issues-2098](https://github.com/shadps4-emu/shadPS4/issues/2098).

image

It is also possible to open by right-clicking on a game, implemented in [PR-1513](https://github.com/shadps4-emu/shadPS4/pull/1513)
image

PR #2107, opened 2025-01-09
Qt: Open shadPS4 Folder

Xphalnos:
The idea comes from **yuzu** and the code is extremely inspired by **https://github.com/shadps4-emu/shadPS4/pull/2102**. So thanks @DanielSvoboda!

Before:
img

After:
img

### DIFF
--- a/src/qt_gui/main_window.cpp
+++ b/src/qt_gui/main_window.cpp
@@ -248,11 +248,10 @@ void MainWindow::CreateConnects() {
     });
 
     connect(ui->shadFolderAct, &QAction::triggered, this, [this]() {
-            QString userPath;
-            Common::FS::PathToQString(userPath,
-                                      Common::FS::GetUserPath(Common::FS::PathType::UserDir));
-            QDesktopServices::openUrl(QUrl::fromLocalFile(userPath));
-        });
+        QString userPath;
+        Common::FS::PathToQString(userPath, Common::FS::GetUserPath(Common::FS::PathType::UserDir));
+        QDesktopServices::openUrl(QUrl::fromLocalFile(userPath));
+    });
 
     connect(ui->playButton, &QPushButton::clicked, this, &MainWindow::StartGame);
     connect(m_game_grid_frame.get(), &QTableWidget::cellDoubleClicked, this,
--- a/src/qt_gui/main_window.cpp
+++ b/src/qt_gui/main_window.cpp
@@ -247,6 +247,13 @@ void MainWindow::CreateConnects() {
         }
     });
 
+    connect(ui->shadFolderAct, &QAction::triggered, this, [this]() {
+            QString userPath;
+            Common::FS::PathToQString(userPath,
+                                      Common::FS::GetUserPath(Common::FS::PathType::UserDir));
+            QDesktopServices::openUrl(QUrl::fromLocalFile(userPath));
+        });
+
     connect(ui->playButton, &QPushButton::clicked, this, &MainWindow::StartGame);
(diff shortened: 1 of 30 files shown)


GHU7924:
I think it's enough just to add one more point here.


If we talk about the menu you specified, then in order not to inflate it, you can make a general "Open..." in which there will be a choice of "Open/Add Elf Folder" and "Open shadPS4 Folder".

I do not know how important access to the entire emulator folder is, but most often we access the User folder, so we have access to all the saves, logs, and system modules. I think it's worth adding quick access to the system modules.

squidbus:
I'm in favor of having the open user folder button in the main menu as it is now in this PR. Easy to find and access for people who need to go to their user folder, especially on Linux/macOS where it may be somewhere else in a standard user path like `.local/share` or `Library/Application Support`.

I don't think it would make sense to put this in the game right-click menu as it is not per-game, in fact I don't know why there's a log folder option there in the first place for the same reason.

DanielSvoboda:
Maybe this order would be better:
install
Boot
Recent
Open/add
Open shadps4
Exit

DanielSvoboda:
> I'm in favor of having the open user folder button in the main menu as it is now in this PR. Easy to find and access for people who need to go to their user folder, especially on Linux/macOS where it may be somewhere else in a standard user path like `.local/share` or `Library/Application Support`.
> 
> I don't think it would make sense to put this in the game right-click menu as it is not per-game, in fact I don't know why there's a log folder option there in the first place for the same reason.

I placed the button to open the 'log' folder on the right button so that it is easily accessible, without having to open menus, although it is not something individual to each game.

Foul-Tarnished:
> > I'm in favor of having the open user folder button in the main menu as it is now in this PR. Easy to find and access for people who need to go to their user folder, especially on Linux/macOS where it may be somewhere else in a standard user path like `.local/share` or `Library/Application Support`.
> > I don't think it would make sense to put this in the game right-click menu as it is not per-game, in fact I don't know why there's a log folder option there in the first place for the same reason.
> 
> I placed the button to open the 'log' folder on the right button so that it is easily accessible, without having to open menus, although it is not something individual to each game.

Would make more sense to be in the file menu instead, and as easily accessible

GHU7924:
>I don't think it would make sense to put this in the game right-click menu as it is not per-game, in fact I don't know why there's a log folder option there in the first place for the same reason.

I think that since we have a common log for all games, it can be removed from the context menu altogether.

>I'm in favor of having the open user folder button in the main menu as it is now in this PR. Easy to find and access for people who need to go to their user folder, especially on Linux/macOS where it may be somewhere else in a standard user path like .local/share or Library/Application Support.

I forgot that Linux has such a folder separation. Then you should add another folder that leads to User.

Xphalnos:
> Maybe this order would be better: install Boot Recent Open/add Open shadps4 Exit

Like this ?

image

GHU7924:
Personally, it seems to me that Recent Games should be located above the exit button.

P.S I think we should also add "Open User Folder", because for Linux and macOS users this is an important improvement.

@Xphalnos Your PR is really useful, you even inspired me a little to watch all sorts of menus from other programs. Now I want us to redo Help again.

Xphalnos:
@GHU7924 I changed the order and I find it much better.

image

DanielSvoboda:
in pt_BR.ts, could already be translated to, please: Abrir pasta shadPS4

Xphalnos:
Done and ready for merge!

Xphalnos:
Any news for this PR?

squidbus (approved):

Issue #2112, opened 2025-01-10
[APP BUG]: PKG Viewer problem


GHU7924:
### Checklist

- [X] I have searched for a similar issue in this repository and did not find one.
- [X] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

![PKGV](https://github.com/user-attachments/assets/046c21f5-328a-41fa-8a3b-8f7aaac2e8d3)

Open - does not see PKG files.
Install - if there are no files and you click install, the emulator closes.

### Reproduction Steps

Open - does not see PKG files.
Install - if there are no files and you click install, the emulator closes.

### Specify OS Version

Windows 11

DanielSvoboda:
PR-2113 only fixes the issue of the install button appearing without anything selected, perhaps this 'issues' should remain open...

GHU7924:
The PKG Viewer does not see the PKG files (or rather, the PKG files are not displayed), but if you click File to Open the folder and select the folder containing the PKG files, the PKG Viewer will display all the PKG files that were in the added folder.

It would also be great if we could make some improvements.:

1) Add the "Close PKG Viewer" button to the "File"
2) Add another line "Remove from the list" to "InstallPKG" 
3) Make it possible to add multiple folders to the PKG Viewer

We only have the option to add one folder at a time, meaning I can't select a folder containing folders with PKG files.

PR #2113, opened 2025-01-10
Fix -PKG Viewer -Button install

DanielSvoboda:
[issues 2112 ](https://github.com/shadps4-emu/shadPS4/issues/2112)


GHU7924:
I also wrote that the PKG Viewer does not see the PKG files. The PKG itself is not displayed, but the folder is added and the added is displayed in the PKG Viewer.

I can't select the entire folder with other folders containing PKG for multiple installations.

And if PKG can be added and installed, then why is there no way to remove PKG (not from the computer, but from shadPS4), and there is no way to clear this list?

And in the "file" section, rename "Open folder" to something like "Add a folder with PKG" (or something like that), and also add the "Close PKG Viewer" button for convenience (so that "Open" is not lonely XD).

PR #2114, opened 2025-01-10
Implement sceNpCmp functions

StevenMiller123:
This PR implements sceNpCmpNpId, sceNpCmpNpIdInOrder, and sceNpCmpOnlineId based on decompiled behavior. Since libSceNpCommon hadn't been added yet, I've also included auto generated stubs for the rest of the library's functions.
This probably won't improve compatibility for anything, but it should decrease stub spam in some games.


georgemoralis (approved):

src/core/libraries/np_common/np_common.h
```diff
@@ -0,0 +1,1248 @@
...
+#include "common/types.h"
+
+namespace Core::Loader {
+class SymbolsResolver;
+}
+
+namespace Libraries::NpCommon {
+
+constexpr int ORBIS_NP_ERROR_INVALID_ARGUMENT = 0x80550003;
+constexpr int ORBIS_NP_UTIL_ERROR_NOT_MATCH = 0x80550609;
```
polybiusproxy: Errors should be moved to a separate header.

src/core/libraries/np_common/np_common.h
```diff
@@ -0,0 +1,1248 @@
...
+
+constexpr int ORBIS_NP_ONLINEID_MAX_LENGTH = 16;
+
+struct OrbisNpOnlineId {
+    char data[ORBIS_NP_ONLINEID_MAX_LENGTH];
+    char term;
+    char dummy[3];
```
polybiusproxy: Use custom types (`s8` in this case).

polybiusproxy (changes requested):

src/core/libraries/np_common/np_common.cpp
```diff
@@ -0,0 +1,7920 @@
...
+    if (np_id1 == nullptr || np_id2 == nullptr) {
+        return ORBIS_NP_ERROR_INVALID_ARGUMENT;
+    }
+
+    // Compare data
+    if (std::strncmp(reinterpret_cast<char*>(np_id1->handle.data),
+                     reinterpret_cast<char*>(np_id2->handle.data),
+                     ORBIS_NP_ONLINEID_MAX_LENGTH) != 0) {
+        return ORBIS_NP_UTIL_ERROR_NOT_MATCH;
+    }
```
polybiusproxy: Why are you casting?

StevenMiller123: std::strncmp requires the parameters to be char*, it doesn't compile without this change.

polybiusproxy: I see. I didn't realize `s8` was `std::int8_t`. You can change the type of `data` back to a char array.

StevenMiller123: Seems like the compiler is treating it as a signed char, but signed/unsigned char are different enough from char that I needed to cast.

polybiusproxy: Most compilers treat chars as signed unless you specify it's unsigned.

polybiusproxy (approved):

Issue #1858, opened 2024-12-23
CUSA03173 Bloodborne: Baldness issue


Missake212:
Some people reported that their character have some type of baldness issue, they're not completely bald but they do have some bald spots on their head depending on the haircut they chose during character creation, I tested in main on latest nightly and this is how it looks like on the emulator:

https://github.com/user-attachments/assets/f3d30c14-8402-42d5-b9bd-ad65829e54b0

And this is how it looks like on PS4:

https://github.com/user-attachments/assets/059f6572-6a6c-4365-bb3d-019ec4cf118c

rafael-57:
It's a bit of an old issue, I'm trying to isolate the guilty commit. Here's some more screenshots in the meantime:

0.5.1 latest
![image](https://github.com/user-attachments/assets/6e037e46-5f56-4f5c-b0f5-e3f15d16d84b)

0.4 release
![image](https://github.com/user-attachments/assets/181b4a17-6d7d-4bc0-b88f-469a7ecc4e92)

diegolix29:
im debuging which commit causes this as we cant look into old actions rn, as far i could get its between these commits im currently doing local testing as some people help me narrow it down on my fork but testing main shad old actions re pushed 
![image](https://github.com/user-attachments/assets/8acbc62e-6342-4ecc-b446-b07914a4709c)

diegolix29:
found it is this commit that broke the hair 
![image](https://github.com/user-attachments/assets/ffa15528-3889-4a0e-beff-228d8e47dc62)

raphaelthegreat:
Could you check if this PR fixes the issue https://github.com/shadps4-emu/shadPS4/pull/1869

diegolix29:
yes

diegolix29:
![image](https://github.com/user-attachments/assets/fd5dc59d-e15c-4725-94fc-295619d5cbdc)
 tried this action and is still there 
 
 https://github.com/squidbus/shadPS4/actions/runs/12497571513

raphaelthegreat:
Ok just to make sure, could you go back to latest commit and revert just https://github.com/shadps4-emu/shadPS4/commit/722a0e36be3486d2084bae557bc6722d7b895b3d to confirm. If it is that, we can try to more selectively revert things to see which exact change broke it

diegolix29:
sure give me a minute will do locally rn

diegolix29:
[...]

PR #2123, opened 2025-01-11
shader_recompiler: Improvements to buffer addressing implementation.

squidbus:
Adds a new common buffer addressing implementation, accounting for all of the different components:
* Fixes handling of `index_stride` and `element_size` which need to be converted into their actual value from bit field.
* Texture buffers still have some limitations, asserted on specially before calling this new function.
* Parameters from sharp needed for `swizzle_enable` added to specialization.

And misc related improvements:
* Add multiply by 1 case to constant propagation pass to help optimize new buffer address logic.
* Clean up buffer load/store instructions to pass through all the MUBUF/MTBUF flags and let the tracking pass take care of what is and isn't supported in one place.

Fixes some shaders in CUSA05637 that use swizzle addressing.

### DIFF
--- a/src/shader_recompiler/frontend/translate/vector_memory.cpp
+++ b/src/shader_recompiler/frontend/translate/vector_memory.cpp
@@ -164,8 +164,8 @@ void Translator::EmitVectorMemory(const GcnInst& inst) {
 }
 
 void Translator::BUFFER_LOAD(u32 num_dwords, bool is_typed, const GcnInst& inst) {
-    const auto& mtbuf = inst.control.mtbuf;
-    const bool is_ring = mtbuf.glc && mtbuf.slc;
+    const auto& mubuf = inst.control.mubuf;
+    const bool is_ring = mubuf.glc && mubuf.slc;
     const IR::VectorReg vaddr{inst.src[0].code};
     const IR::ScalarReg sharp{inst.src[2].code * 4};
     const IR::Value soffset{GetSrc(inst.src[3])};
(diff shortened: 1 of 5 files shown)


rafael-57:
Since I saw swizzle changes I tested for #1858, no changes for that issue.
image

squidbus:
> Since I saw swizzle changes I tested for #1858, no changes for that issue.

Different kind of swizzle.

diegolix29:
im having like 3 fps less on dlc BB with this pr but dont now it maybe load to pc but tested on main and the fps where more by 3-4 i know is not a big difference but its the only thing i noticed changed with this pr on bb

i9 intel 12900k
nvidia 3080  12gb
64 Ram 
windows 10

squidbus:
> im having like 3 fps less on dlc BB with this pr but dont now it maybe load to pc but tested on main and the fps where more by 3-4 i know is not a big difference but its the only thing i noticed changed with this pr on bb

If I had to guess any differences to performance should be either down to shader recompiles because of the changes which would be temporary as your GPU driver shader cache updates, or due to specialization fixes making more variants which is unavoidable for correctness.

squidbus:
I removed the uniform size fix for specialization for now as I'm still having some issues with excessive shader recompiles from buffer sizes, now it's just the buffer addressing changes.

meijjaa:
Hi Squidbus, thanks for the hard word so far! 
I was just testing with CUSA03745 v1.03 and noticed the 
"Unsupported texture buffer address mode." Assert due to swizzle_enable == 1. Is this something you'll be working on? Or can I help in any way? I have setup a debug environment for easy testing.

squidbus:
> Hi Squidbus, thanks for the hard word so far! I was just testing with CUSA03745 v1.03 and noticed the "Unsupported texture buffer address mode." Assert due to swizzle_enable == 1. Is this something you'll be working on? Or can I help in any way? I have setup a debug environment for easy testing.

It's going to require a lot more work to implement the rest of addressing for texture buffers, since we can't offset them at the sub-texel level.

Randomuser8219:
we should rebase again
the hedgehog engine 2 post processing fix has been added to main

squidbus:
> we should rebase again

Sure, done

PR #2129, opened 2025-01-11
shader_recompiler: Remove AMD native CubeFaceCoord.

squidbus:
AMD's SPIR-V `CubeFaceCoord` does not seem to be equivalent to `v_cubesc_f32`/`v_cubetc_f32`, rather it seems to include some of the additional transformations typically done on those instructions' outputs after. So we can't use it as a native implementation for those instructions.

Should fix some Bloodborne shadow issues in first area remaining on AMD GPUs, need someone with AMD to test and confirm.


squidbus:
Confirmed to be fixed

diogoaduartereis:
Hi there just out of curiosity, what kind of shadow issues were fixed in the first area? @squidbus

StevenMiller123:
@diogoaduartereis This fixed the Bloodborne shadow issues reported in https://github.com/shadps4-emu/shadPS4/pull/2104#issuecomment-2585372728

diogoaduartereis:
Thanks!

PR #2130, opened 2025-01-11
GUI: Speed up GUI loading by caching game sizes

kalaposfos13:
Are you tired of waiting for the emulator to go through tens, if not hundreds of gigabytes of your games for the nth time just to calculate how big each game is? Well, if you also want your loader to load everything in less than a second and don't really care how big your games are, then just turn the feature off!
When the option is disabled (it is on by default), the entire Size colunm is also hidden.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -72,6 +72,7 @@ static bool checkCompatibilityOnStartup = false;
 static std::string trophyKey;
 
 // Gui
+static bool load_game_size = true;
 std::vector<std::filesystem::path> settings_install_dirs = {};
 std::filesystem::path settings_addon_install_dir = {};
 u32 main_window_geometry_x = 400;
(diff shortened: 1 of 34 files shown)


kalaposfos13:
One known problem is that the Settings tab with this new line is yet again too tall and therefore acquired a scrollbar, but I don't know how to fix that

DanielSvoboda:
I made this PR to adjust https://github.com/kalaposfos13/shadPS4/pull/7
image

squidbus:
Why not just add a cache for this information? Most other emulators I've seen use a game list cache to keep loading times low.

DanielSvoboda:
Disabling 'size' seems like a great way to get around the problem, but caching is even better. Would this information be updated after installation and would it be compared with the serial and file size? Something similar to the time played. What would be stored? Just the 'size'?

squidbus:
> Would this information be updated after installation and would it be compared with the serial and file size? Something similar to the time played. What would be stored? Just the 'size'?

We have caching already to store stuff like trophies and the banner that displays as the background when a game is selected, would suggest a similar sort of caching here as well. When we determine the info for a game's listing we can save details that take a long time to determine to a cache file in the same `game_data/(game id)` directory as the other cached files, and reference that in the future.

kalaposfos13:
Caching is done by storing size_cache.txt in game_data/CUSAXXXXX/ for every game, which includes one line: the cached size. If the cache doesn't exist, or the file's last modified date is older than the game folder's last modified date, the size is recalculated, otherwise the cached string is used.

PR #2131, opened 2025-01-11
Fix

DanielSvoboda:
I removed the gameSizeCheckBox from the 'Emulator' group and put it in 'GUI settings'
hLayoutTrophy which contains the Trophy information was inside the GUIMusicLayout, so I fixed that too.
And I added the translation for the phrase: Show Game Size In List


DanielSvoboda:
This PR was for another branch, I ended up opening it in the wrong place, so I closed it...

PR #2134, opened 2025-01-12
renderer_vulkan: Handle depth-stencil copies through depth render overrides.

squidbus:
As noted in the AMD docs, `DB_RENDER_OVERRIDE` bits may be used to perform copies to alternate surfaces. Detect and implement this case by looking for these bits set along with a valid depth/stencil read/write base and disabled color buffer mode.

Fixes a lot of post-processing issues in Hedgehog Engine 2 games, such as persistent screen overlays (supposed to be an effect over objects far in the distance) and incorrect lighting effects (although there are still some other lighting problems remaining). Probably helps other games that do this as well.

### DIFF
--- a/src/core/devtools/widget/reg_popup.cpp
+++ b/src/core/devtools/widget/reg_popup.cpp
@@ -105,7 +105,8 @@ void RegPopup::DrawDepthBuffer(const DepthBuffer& depth_data) {
             "DEPTH_SLICE.TILE_MAX",           depth_buffer.depth_slice.tile_max,
             "Pitch()",                        depth_buffer.Pitch(),
             "Height()",                       depth_buffer.Height(),
-            "Address()",                      depth_buffer.Address(),
+            "DepthAddress()",                 depth_buffer.DepthAddress(),
+            "StencilAddress()",               depth_buffer.StencilAddress(),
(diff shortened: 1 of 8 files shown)


dbz400:
Sonic Forces no more blurred in-game and combine with https://github.com/shadps4-emu/shadPS4/pull/2123 , it is fully playable.

image

jas0n098:
Breaks Fist of the North Star: Lost Paradise by making nearly everything transparent/invisible

roamic:
FotNS:LP is not working without a hack so you can't really break it.

roamic:
That being said, FotNS:LP crashes on my system now (NV 4080 12 Gb, i7 8700, 32 Gb DDR4). It doesn't crash on main, so it's a regression.

tomboylover93:
This fixes the visibility issue in the Radical Highway section of the first stage in SONIC X SHADOW GENERATIONS (CUSA44422), making it fully playable from start to finish.

Screenshot_20250112_103730

roamic:
FotNS:LP crash regression if fixed with the latest commit.

squidbus:
Removed changes to depth buffer read/write base and just implements the special depth-stencil copy now.

roamic (approved):

PR #2135, opened 2025-01-12
CLI: Add argument to pass an argument to the game

kalaposfos13:
Syntax: `shadps4 <options> <game> -- <game arg 1> <game arg 2> ...`
Works for both SDL and Qt versions, but there's no GUI for it.
You can pass up to 32 arguments to a game (untested)
Tested with Knack 2 (needs only one argument): `./build/shadps4 CUSA08014 -- "-w=10 -cursoroff -region=a -2160p"`


ngoquang2708:
There is a common pattern used in Linux, that everything after a ``--`` get passed to the child/guest app.

kalaposfos13:
Now the above example looks like this:
`./build/shadps4 CUSA08014 -- "-w=10 -cursoroff -region=a -2160p"`
The automatic "last argument is the game" still works, but now it is the last argument before --, if there are game arguments. I can't verify that multiple arguments work, as I don't have a game that uses that that I also inspected with Ghidra.

ngoquang2708:
It should not have the double quote though. But that seem hard to do.

kalaposfos13:
The quotes are intentional, the game expects one string argument with all the options in it. If there weren't quotes, they would be split into multiple arguments, of which the game will ignore all but the first one.

ArjixWasTaken:
I suggest updating the description of the PR.

Foul-Tarnished:
Does Knack actually accept a custom resolution as an argument and use it ?

Is there any documented games launch arguments ?

ngoquang2708:
This doesn't work with ``-g`` or ``--game``:
```
bash$ ./Shadps4-sdl.AppImage -f true --game CUSA03173 -- "abc"
Unknown argument: --, see --help for info.

```

If you use ``--`` without passing args for game, it will crash:
```
bash$ ./Shadps4-sdl.AppImage -f true --game CUSA03173 -- 
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string: construction from null is not valid
Aborted (core dumped)

```

kalaposfos13:
@ngoquang2708 The above errors have been fixed.
Tested with:
`./build/shadps4 -g CUSA03173 -- abc def`
and
`./build/shadps4 -g CUSA03173 --`

src/main.cpp
```diff
@@ -37,6 +38,10 @@ int main(int argc, char* argv[]) {
              std::cout << "Usage: shadps4 [options] <elf or eboot.bin path>\n"
                           "Options:\n"
                           "  -g, --game <path|ID>          Specify game path to launch\n"
+                          "  -ga, --game-with-arg <path|ID> <arg>\n"
```
ngoquang2708: This is not needed anymore.

ngoquang2708:
Look in to this, I think we can do it better by extract CLI parsing logic to a common file share between Qt and CLI versions. In the future, we can easily port it to use boost 's ``program_options``. But that should be left for another PR I suppose.

ngoquang2708 (approved):

src/emulator.h
```diff
@@ -25,7 +25,8 @@ class Emulator {
     Emulator();
     ~Emulator();
 
-    void Run(const std::filesystem::path& file);
+    void Run(const std::filesystem::path& file,
+             const std::vector<std::string> args = std::vector<std::string>());
```
roamic:
```suggestion
             const std::vector<std::string> args = {});
```

src/core/linker.h
```diff
@@ -143,7 +143,7 @@ class Linker {
     void Relocate(Module* module);
     bool Resolve(const std::string& name, Loader::SymbolType type, Module* module,
                  Loader::SymbolRecord* return_info);
-    void Execute();
+    void Execute(const std::vector<std::string> args = std::vector<std::string>());
```
roamic:
```suggestion
    void Execute(const std::vector<std::string> args = {});
```

src/emulator.cpp
```diff
@@ -151,6 +151,15 @@ void Emulator::Run(const std::filesystem::path& file) {
         if (const auto raw_attributes = param_sfo->GetInteger("ATTRIBUTE")) {
             psf_attributes.raw = *raw_attributes;
         }
+        if (!args.empty()) {
+            int argc = args.size() > 32 ? 32 : args.size();
```
roamic:
```suggestion
            int argc = std::min<int>(args.size(), 32);
```

src/main.cpp
```diff
@@ -126,6 +130,28 @@ int main(int argc, char* argv[]) {
...
+            }
+            for (int j = i + 1; j < argc; j++) {
+                game_args.push_back(argv[j]);
+            }
+            break;
+        } else if (std::string(argv[i + 1]) == "--") {
+            if (!has_game_argument) {
+                game_path = argv[i];
+                has_game_argument = true;
+            }
```
roamic, line 146: I think that the code added below this point in this block is an unnecessary duplication of the previous `if` statement that will be executed in the next iteration anyway.

src/main.cpp
```diff
@@ -126,6 +130,28 @@ int main(int argc, char* argv[]) {
...
+        } else if (std::string(argv[i]) == "--") {
+            if (i + 1 == argc) {
+                std::cerr << "Warning: -- is set, but no game arguments are added!\n";
+                break;
+            }
+            for (int j = i + 1; j < argc; j++) {
+                game_args.push_back(argv[j]);
+            }
+            break;
+        } else if (std::string(argv[i + 1]) == "--") {
```
roamic: Would be good to add a sanity check that `i + 1 < argc`

src/qt_gui/main.cpp
```diff
@@ -131,6 +135,28 @@ int main(int argc, char* argv[]) {
             // Assume the last argument is the game file if not specified via -g/--game
             game_path = argv[i];
             has_game_argument = true;
+        } else if (std::string(argv[i]) == "--") {
```
roamic, line 138: All comments from the same block in SDL main also apply to this block.

roamic (approved):

Issue #1303, opened 2024-10-08
[Feature Request] Implement output resampling (upscaling/downscaling) methods


Foul-Tarnished:
**Let us choose the algorithm used to upscale/downscale the image from the game window.**

Some algorithm that could be great:
- Bilinear
- Bicubic
- Nearest neighbor
- Area (the best for downscaling, see https://github.com/dolphin-emu/dolphin/pull/11999)

Even better if we can choose a different algorithm for upscaling and downscaling in the GUI.

**Why:**
Considering most games are 1080p, would be useful for 1440p/4K monitors.
But would also help for game patches that change the game resolution (like BB 4K patch, etc.) on lower res monitors

MaxMo20:
What about Lanczos? Isn't that the sharpest and fairly smooth looking algorithm? I am not an expert but wouldn't supersampling with Lanczos algorithm would be the best, Essentially anti aliasing with a Lanczos upscaling Algorithm to mimic what native would look like. I know with DgVoodoo2 can do this with DirectX 9 games.

vardin23:
Given how motion vectors through Motion Blur and TAA have been used to mod DLSS into games using reshade , is that a possibility on ShadPS4 on a per games basis for games that support Motion Blur?
It seems like a really long shot to me but I think there's a possibility.

Foul-Tarnished:
How are game mods related to this feature request ?

The emulator translate gnm to vulkan, so I don't see how it could do it
besides modifying shadps4 vk itself, plus a game code patch to somehow
forward stuff needes for it

Le sam. 8 févr. 2025 à 21:18, Arvind Mishra ***@***.***> a
écrit :

DemoJameson:
we have fsr1 now #2624

Foul-Tarnished:
Still not what's been asked

Le jeu. 13 mars 2025 à 10:29, DemoJameson ***@***.***> a
écrit :

MajorP93:
+1
Would be awesome to have the option to configure this.
At a bare minimum FSR should be toggleable.
Right now FSR is enabled by default which alters the "look and feel" of a game quite substantially.
To preserve the original aesthetic of a game we should have the option to enable/disable FSR upscaling and -if possible- choose the desired upscaling/downscaling algorithm.

Foul-Tarnished:
What really ?

This is stupid, FSR1 looks deep fried with ghosting..

PR #2136, opened 2025-01-12
Allow setting of post processing filter in config

ngoquang2708:


Foul-Tarnished:
Great!

I imagine "linear" is bilinear ?
**Could you add bicubic, lanczos and other asked in #1303 ?** 🙏 

GUI in another PR ?
Would be cool if we could choose one filtrr used when downscaling and another when upscaling.
Would need to check the internal resolution compared to the window actual resolution and use the right filter accordingly

ngoquang2708:
This PR only allow changing of simple Vulkan filters. I don't know about those custom filters.

Foul-Tarnished:
I see

I found this but it seems ratger complex to add
https://github.com/DTolm/VkFFT
https://github.com/DTolm/VkResample

ngoquang2708:
Added setting dialog GUI from @rainmakerv3.

ngoquang2708:
Closed due to recently changes on main so this is not applicable anymore.

PR #2137, opened 2025-01-12
video_core: detiler: display micro 64bpp

psucien:
A rare 3d tiling case found in irradiance probes in Knack


DanielSvoboda:
I think `Minecraft CUSA00744` also makes use of this

PR #2146, opened 2025-01-14
renderer: handle disabled clipping

roamic:
Adds handling to the disabled clipping by creating a vertex shader permutation that converts vertex coordinates from clip space to screen space and fits them inside a huge (`[0..16383, 0..16383]`) viewport. One shortcoming of this method is that it does not support multiple different viewport transformations but nothing can be done about this without hardware and API support.

Also implicitly disables clipping for the `RECTLIST` primitive because this primitive should not be clipped, rather it should be culled.

Other minor changes:
* Added `primitiveTopologyPatchListRestart` if it is supported by the device to silence the validation errors.
* Changed empty viewport to one outside of the render space.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
@@ -125,6 +125,7 @@ const Shader::RuntimeInfo& PipelineCache::BuildRuntimeInfo(Stage stage, LogicalS
         info.vs_info.emulate_depth_negative_one_to_one =
             !instance.IsDepthClipControlSupported() &&
             regs.clipper_control.clip_space == Liverpool::ClipSpace::MinusWToW;
+        info.vs_info.clip_disable = graphics_key.clip_disable;
         if (l_stage == LogicalStage::TessellationEval) {
             info.vs_info.tess_type = regs.tess_config.type;
             info.vs_info.tess_topology = regs.tess_config.topology;
@@ -261,6 +262,15 @@ bool PipelineCache::RefreshGraphicsKey() {
     auto& regs = liverpool->regs;
     auto& key = graphics_key;
 
+    const auto& vp_ctl = regs.viewport_control;
+
+    // TODO(roamic): the statement below needs verification with a sample on the real HW.
+    // If there is no defined transform to convert from clip space to screen space we assume that
+    // clipping is also disabled.
+    const bool viewport_disabled = !vp_ctl.xoffset_enable && !vp_ctl.xscale_enable &&
+                                   !vp_ctl.yoffset_enable && !vp_ctl.yscale_enable &&
+                                   !vp_ctl.zoffset_enable && !vp_ctl.zscale_enable;
+    key.clip_disable = regs.clipper_control.clip_disable || viewport_disabled;
     key.depth_test_enable = regs.depth_control.depth_enable;
     key.depth_write_enable =
         regs.depth_control.depth_write_enable && !regs.depth_render_control.depth_clear_enable;
--- a/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
@@ -210,6 +210,8 @@ PipelineCache::PipelineCache(const Instance& instance_, Scheduler& scheduler_,
                                       instance.GetDriverID() == vk::DriverId::eNvidiaProprietary,
         .needs_lds_barriers = instance.GetDriverID() == vk::DriverId::eNvidiaProprietary ||
                               instance.GetDriverID() == vk::DriverId::eMoltenvk,
+        .max_viewport_width = instance.GetMaxViewportWidth(),
+        .max_viewport_height = instance.GetMaxViewportHeight(),
     };
     auto [cache_result, cache] = instance.GetDevice().createPipelineCacheUnique({});
     ASSERT_MSG(cache_result == vk::Result::eSuccess, "Failed to create pipeline cache: {}",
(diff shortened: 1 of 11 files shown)


src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
```diff
@@ -261,6 +262,15 @@ bool PipelineCache::RefreshGraphicsKey() {
...
 
+    const auto& vp_ctl = regs.viewport_control;
+
+    // TODO(roamic): the statement below needs verification with a sample on the real HW.
+    // If there is no defined transform to convert from clip space to screen space we assume that
+    // clipping is also disabled.
+    const bool viewport_disabled = !vp_ctl.xoffset_enable && !vp_ctl.xscale_enable &&
+                                   !vp_ctl.yoffset_enable && !vp_ctl.yscale_enable &&
+                                   !vp_ctl.zoffset_enable && !vp_ctl.zscale_enable;
+    key.clip_disable = regs.clipper_control.clip_disable || viewport_disabled;
```
Triang3l: I think disabling all offsets/scales while keeping clipping enabled would still produce a valid viewport with the bottom-right quadrant covering one pixel at (0, 0) with clipping in place — you can probably also use pixel coordinates meaningfully this way even with clipping enabled if you provide a wide scissor and guard band 🙃

Triang3l (changes requested):

src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
roamic: I'm not sure I understand you correctly. I have a game (FotNS:LP) that outputs vertex coordinates as if they were pixel coordinates while setting VTE pass-through (disabling viewport and disabling W0) and NOT disabling clipping. On the other hand I have a game (Bloodborne) that disables clipping and has a vertex shader that outputs x and y in NDC space (-1..1). I don't really know what to make of it.

roamic: I understand what you mean now, thank you for the insight. I'll check the guard band settings the game sets, but I remember now that it's just `1.0` for each adjust register. Maybe HW offset it set to something else though.

Triang3l: If the game draws rectangles, it's possible that they always bypass clipping, but I've never checked this on real GPUs so I'm not sure.

red-prig: By the way, an interesting question is whether the RECTLIST or QUADLIST topology is clipping, and we have also seen POLYGON in practice

Triang3l: Quads and polygons probably match the respective OpenGL geometry types, so they likely should be clipped as usual (after expansion to triangles likely — though they may be different from triangles when it comes to wireframe rendering in regard to inner edges, also not sure). Rectangles, however, as far as I understand, simply cover the AABB of their vertices on the screen, and just like for points, I'd imagine that there may be some special behavior of the clipper for rectangles, such as having clipping implicitly disabled completely — but again, I've never tried that so I'm not sure. Also not obvious whether DISC_ADJ of the guard band (rather than CLIP_ADJ) is applicable to them.

roamic: You were correct that clipping should be disabled for RECTLIST. I've disabled clipping for them, applied viewport transfrom and it renders as it should. The olny thing that is bothering me now is that on PS4 it is possible to have 16 viewports while this emulation method can only support one. But I guess there's nothing we can do apart from adding an error log or assertion.

Also, I've tried `VK_NV_fill_rectangle` and it didn't work although it shouldn't clip the rectangle. Couldn't even capture it with rdoc, it seems it doesn't support this extension.

Triang3l: The 16 hardware viewports need to be emulated as 16 separate Vulkan viewports/scissors, but in this case, when clipping is disabled, all Vulkan viewports/scissors need to cover the entire screen coordinate range, and the constants for remapping vertex positions from the guest to the host in host shaders also need to be passed as an array indexed by what goes to `gl_ViewportIndex` (also need to ensure that out-of-bounds index behavior on the host consistently matches what happens on the console in all places where it's relevant).

roamic: Turns out our recompiler does not support `ViewportIndex` yet so this will have to wait until it's implemented.

Also I think that vertices should be remapped in the geometry shader instead of vertex if it's present, but this also will have to wait until we encounter a game that disables clipping with geo shaders.

Triang3l: Viewport and render target array layer index outputs need to be supported regardless of whether clipping is disabled. Some devices may support the `VK_EXT_shader_viewport_index_layer` extension (or the Vulkan 1.2 features `shaderOutputViewportIndex` and `shaderOutputLayer`), which allows specifying the viewport/layer index from VS or TES, similarily to how, as far as I understand, they are finally emitted from the hardware VS on GCN — when supported, it's desirable to use that rather than emission from GS when GS is not needed, as geometry shaders often can be slow in hardware (though geometry shaders that emit the same vertex count as they receive may be faster on some drivers/hardware, but in general they aren't very optimal). If this functionality isn't supported, you can emit the viewport and/or the layer index from the Vulkan VS/TES as a generic output, and then use a GS to just copy the inputs to the actual outputs for the rasterizer/PS.

squidbus (approved):

PR #2160, opened 2025-01-16
sceKernelAio* implementation

georgemoralis:
Heavily based on https://github.com/GoldHEN/GoldHEN_Plugins_Repository

It doesn't actually implements Aio but it fake it with sync I/O . Should be enough for start . At least @StevenMiller123 will need this :D


georgemoralis:
known games that uses it  : 

- Cyberpunk 2077
- Need for speed heat
- Session: Skate Sim
- ????

polybiusproxy:
And all UE4 games.

PR #2161, opened 2025-01-16
network: Remove firing Np callbacks from check stubs.

squidbus:
Np callbacks are intended to be called when an event occurs and probably should not be spammed every time the game requests checking for events. Removing the calls from these stubs seems to fix general instability I have where the thread `Toolkit::NP::MainThread` from `libSceNpToolkit2` would crash randomly in `sceLibcMspaceFree`.

### DIFF
--- a/src/core/libraries/network/netctl.cpp
+++ b/src/core/libraries/network/netctl.cpp
@@ -93,7 +93,7 @@ int PS4_SYSV_ABI sceNetCtlUnregisterCallbackV6() {
 }
 
 int PS4_SYSV_ABI sceNetCtlCheckCallback() {
-    netctl.CheckCallback();
+    LOG_DEBUG(Lib_NetCtl, "(STUBBED) called");
     return ORBIS_OK;
 }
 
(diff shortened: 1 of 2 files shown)


georgemoralis:
Hmm? if not firing for checkcallback where should we check for it?

squidbus:
I don't think we need to fire it at all as long as it remains stubbed. There are no events occurring.

squidbus:
Basically `libSceNpToolkit2` spawns a thread with a loop that continuously does this to check for events:
```
    sceNetCtlCheckCallbackForNpToolkit();
    sceNpCheckCallbackForLib();
    if (DAT_010aedb8 == '\0') {
      sceNpCheckCallback();
    }
    sceKernelUsleep(33000);
```

If there are no events then the checks shouldn't do anything. Right now our behavior is basically to spam their callback saying that we keep signing out / disconnecting over and over.

ghost:
CUSA19104 is playable again with this build.

georgemoralis:
ok let's give it green light for now

PR #2162, opened 2025-01-16
Return error codes to game from map/unmap memory functions

mailwl:
Looks like Unity 2020+ games expects errors from map/unmap functions, this PR feeds it.
This should help with games like https://github.com/shadps4-emu/shadps4-game-compatibility/issues/2015 with crash like:

> [Kernel.Vmm] memory.cpp:sceKernelMapDirectMemory:183: called, redirected to sceKernelMapNamedDirectMemory
> [Debug] memory.cpp:operator():284: Assertion Failed!

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -287,13 +287,14 @@ int MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, size_t size, M
 
     // Fixed mapping means the virtual address must exactly match the provided one.
     if (True(flags & MemoryMapFlags::Fixed)) {
-        // This should return SCE_KERNEL_ERROR_ENOMEM but shouldn't normally happen.
         const auto& vma = FindVMA(mapped_addr)->second;
         const size_t remaining_size = vma.base + vma.size - mapped_addr;
         if (vma.IsMapped()) {
+            LOG_ERROR(Kernel_Vmm, "Attempted to map already mapped memory at {:#x}", virtual_addr);
             return ORBIS_KERNEL_ERROR_EBUSY;
         }
         if (remaining_size < size) {
+            LOG_ERROR(Kernel_Vmm, "Could not map memory with size {:#x}", size);
             return ORBIS_KERNEL_ERROR_ENOMEM;
         }
     }
@@ -398,6 +399,7 @@ s32 MemoryManager::UnmapMemoryImpl(VAddr virtual_addr, size_t size) {
     const auto it = FindVMA(virtual_addr);
     const auto& vma_base = it->second;
     if (!vma_base.Contains(virtual_addr, size)) {
+        LOG_ERROR(Kernel_Vmm, "Attempted to unmap not mapped memory at {:#x}", virtual_addr);
         return ORBIS_KERNEL_ERROR_EINVAL;
     }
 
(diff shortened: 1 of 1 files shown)


georgemoralis:
i would suggest to have an error log before these returns so if it hits we should know

mailwl:
my english is bad, hope error texts explains what's going on correctly

thecatontheceiling:
"Attempted to unmap already unmapped memory" maybe, just like the first log :p

StevenMiller123:
I've performed hardware tests on various Unity games, none of them hit any error returns during memory calls on real hardware. While it's possible a game is "expecting" errors here, I would think it's unlikely.
I'd recommend performing some decompilation or hardware tests of your own before making assumptions like this.

mailwl:
look, for example CUSA34417:
at place of crash Unity trying to get memory in loop:
```
while(1) {
// some code
            LODWORD(rc) = sceKernelAllocateMainDirectMemory(v17 - v8, 0LL, 0LL, &v36);
            if ( (unsigned int)(rc + 0x7FFDFFF3) < 2 )
                goto LABEL_40;
            v16 = v30;
            v17 = v32;
            if ( (_DWORD)rc )
                goto LABEL_23;
            v35 = v8;
            LODWORD(rc) = sceKernelMapDirectMemory(&v35, v29, v31, 16LL, v36, 0LL); // <-------- here shadPS4 chash
            if ( (unsigned int)(rc + 0x7FFDFFF3) < 2 )
            {
LABEL_40:
// some code
}
```
Whole function enough hard, but looks like it trying to find not mapped memory, and do not crash itself.
i checked some games, CUSA13806 GIGAWRECKER ALT, CUSA27632 Unspottable goes ingame
CUSA34417 Alice Gear Aegis CS - black screen, but not crashed

mailwl:
BTW, how to debug games on hw? is there any tutorials? I have retail ps4, not devkit

StevenMiller123:
With a retail PS4, you can compile and use GoldHEN plugins to test hardware behavior. That's what I've used for my hardware tests. I don't think there's any easy tutorials you can follow though.

As for the memory crash, just because a game handles errors doesn't mean it's expecting them. I'd recommend testing your build locally to see if this change fixes the problem you encountered, without causing any new errors.

mailwl:
i used wrong word, sorry. Doesn't directly expect errors, but knows how to handle them.

And you was right, I used wrong error code, if game cannot map direct memory to fixed virtual address, it should return NOMEM. now CUSA13806 GIGAWRECKER ALT fully playable

mailwl:
any other games, except unity 2019+ (i was thinking it appears in 2020 version only), doesn't crash at this place, so it should not break compatibility

StevenMiller123:
@mailwl There are actually many games crashing on that MapMemory assert. Some of them intentionally hit it (they hit it on actual PS4's too), while others hit it due to issues with our emulation. 
As for the UnmapMemoryImpl assert, your error return is a hack, and doesn't align with hardware behavior. There's a PR opened that properly fixes that one.

isshininu:
this PR allowed to progress only one game with cpp memory error further
Снимок экрана (1790)

mailwl:
didn't you read? CUSA13806 GIGAWRECKER ALT fully playable
https://github.com/shadps4-emu/shadps4-game-compatibility/issues/2017

PR #2165, opened 2025-01-16
hotfix: keep framerate stable even without vsync

viniciuslrangel:


diogoaduartereis:
Hi there, I'm trying to find out what this exactly did to improve the game? Can you clarify?

Missake212:
> Hi there, I'm trying to find out what this exactly did to improve the game? Can you clarify?

This fixed an issue where the fps counter would display really high fps when vblank was set to more than 1, but the game wasn't running at those frame rates at all

diogoaduartereis:
Ah that explains it then. My rivatuner was all messed up, displaying wrong frames and not locking the frames properly, which also messed up frame generation. Thanks for the quick reply!

PR #2170, opened 2025-01-17
input: Don't use old input state in GameController::ReadState()

ngoquang2708:
Fix input lag when there is multiple user inputs happen at the same time. SDL input events can only fire one at a time and game may get some input events later than other.

### DIFF
--- a/src/input/controller.cpp
+++ b/src/input/controller.cpp
@@ -260,17 +260,16 @@ void GameController::SetTouchpadState(int touchIndex, bool touchDown, float x, f
     }
 }
 
-void GameController::SetEngine(Engine* engine) {
+void GameController::SetEngine(std::unique_ptr<Engine> engine) {
     std::scoped_lock _{m_mutex};
-    delete m_engine;
-    m_engine = engine;
+    m_engine = std::move(engine);
     if (m_engine) {
         m_engine->Init();
     }
 }
 
 Engine* GameController::GetEngine() {
-    return m_engine;
+    return m_engine.get();
 }
 
 u32 GameController::Poll() {
--- a/src/input/controller.cpp
+++ b/src/input/controller.cpp
@@ -10,6 +10,55 @@
 
 namespace Input {
 
+using Libraries::Pad::OrbisPadButtonDataOffset;
+
+void State::OnButton(OrbisPadButtonDataOffset button, bool isPressed) {
+    if (isPressed) {
+        buttonsState |= button;
+    } else {
+        buttonsState &= ~button;
+    }
+}
+
+void State::OnAxis(Axis axis, int value) {
+    const auto toggle = [&](const auto button) {
+        if (value > 0) {
+            buttonsState |= button;
+        } else {
+            buttonsState &= ~button;
+        }
+    };
+    switch (axis) {
+    case Axis::TriggerLeft:
+        toggle(OrbisPadButtonDataOffset::L2);
+        break;
+    case Axis::TriggerRight:
+        toggle(OrbisPadButtonDataOffset::R2);
+        break;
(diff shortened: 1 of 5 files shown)


ngoquang2708:
Main:

https://github.com/user-attachments/assets/94a0b2c6-63fe-41ca-8dc0-bda627ba36c2

PR:

https://github.com/user-attachments/assets/92fc2711-8022-49f8-9778-b986ce1084f9

DemoJameson:
it seems you accidentally committed the `shadps4.flatpak` file

ngoquang2708:
> it seems you accidentally committed the `shadps4.flatpak` file

Nice catch. Thanks.

roamic:
Can someone check Gravity Rush Remastered with this PR?

georgemoralis:
hmm i would prefer no SDL code in controller.cpp since it gonna be generic (as i imagine it at least) and have devices for inputs (SDL_keyboard ,SDL_Gamepad , XInput etc)

GamerzHell9137:
Hoped this would fix Disgaea 5 but nope

https://github.com/user-attachments/assets/10e83f96-e835-44d7-9dcc-3ec23e1c6cb7

ngoquang2708:
Ended up refactor the whole input subsystem.

ngoquang2708:
> Can someone check Gravity Rush Remastered with this PR?

Confirmed motion controls is still working.

src/input/controller.cpp
```diff
@@ -211,62 +233,49 @@ void GameController::CalculateOrientation(Libraries::Pad::OrbisFVector3& acceler
...
-                LOG_ERROR(Input, "Failed to initialize accel controls for gamepad");
-            }
-        }
-
-        SDL_free(gamepads);
-
-        SetLightBarRGB(0, 0, 255);
+void GameController::SetEngine(Engine* engine) {
+    std::scoped_lock _{m_mutex};
+    delete m_engine;
```
roamic: `std::unique_ptr` can be used instead of manual lifetime management.

roamic (commented):
LGTM otherwise

Randomuser8219:
This could also fix Minecraft input lag

ngoquang2708:
> This could also fix Minecraft input lag

Could you test it?

src/input/controller.cpp
ngoquang2708: Done.

src/sdl_window.cpp
```diff
@@ -20,47 +21,200 @@
...
+    }
+    int gamepad_count;
+    SDL_JoystickID* gamepads = SDL_GetGamepads(&gamepad_count);
+    if (!gamepads) {
+        LOG_ERROR(Input, "Cannot get gamepad list: {}", SDL_GetError());
+        return;
+    }
+    if (gamepad_count == 0) {
+        LOG_INFO(Input, "No gamepad found!");
+        SDL_free(gamepads);
```
ngoquang2708, line 105: I would use ``goto`` here to jump to the cleanup code at the end (Linux practice) but I don't know if everyone like it or not?

roamic: Please don't use `goto`

ngoquang2708: It is not bad for C style cleanup codes (SDL) to be honest.

roamic: For `C`-based projects, sure. But this project is primarily `C++` so you can use its best practices like RAII wrapper around SDL objects instead of `goto`s.

ngoquang2708: Ok. Fair enough.

roamic (approved):

PR #2095, opened 2025-01-08
renderer_vulkan: Clear blank frame

ngoquang2708:
Fix display of garbage images on startup on some drivers.

<details>
  <summary>Epilepsy warning</summary>
  

https://github.com/user-attachments/assets/7b2a8363-6fc5-4fc0-8d27-7c7ed9145d27

</details>

PR #2176, opened 2025-01-18
renderer_vulkan: Remove swapchain image reinterpretation.

squidbus:
Instead of reinterpreting swapchain image as unorm in the image view, just select a unorm swapchain format to begin with.

Fixes blank or flickering screen on macOS.


ngoquang2708:
Hope this would supersede this PR https://github.com/shadps4-emu/shadPS4/pull/2095 but it doesn't 🙁.

PR #2177, opened 2025-01-18
texture_cache: Fix image mip overlap.

squidbus:
A check in `IsMipOf` was inverted to return a negative result if tiling is compatible, when it should return a negative result if tiling is not compatible. Additionally, an assert for rebinding color attachments seems to be invalid, as it is not necessarily true that the current state width/height will equal the old image size, if it was not previously the smallest bound attachment.

Additionally, I added level and layer info to image view debug names to help my debugging in this process.

Fixes passes in Hedgehog Engine 2 games where an image's base level would be used as a render target, and then subsequent levels would be rendered to from that base to generate mips.


Randomuser8219:
Are you saying this improves graphics or just improves some stuff internally?

squidbus:
> Are you saying this improves graphics or just improves some stuff internally?

It probably improves some effect, but I didn't notice much myself without looking in a graphics debugger for the specific passes.

PR #2187, opened 2025-01-19
Devtools: Close Button ( X )

DanielSvoboda:
Some people have accidentally entered the magical world of Devtools and don't know how to close it.
It is still possible to use Ctrl+F10.
However, I added an X button to close it.

image

### DIFF
--- a/src/core/devtools/layer.cpp
+++ b/src/core/devtools/layer.cpp
@@ -93,6 +93,12 @@ void L::DrawMenuBar() {
             }
             ImGui::EndMenu();
         }
+
+        SameLine(ImGui::GetWindowWidth() - 30.0f);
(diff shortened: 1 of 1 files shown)


kalaposfos13:
While you're there already, can you make it off by default? It doesn't make much sense for it to be on initially

DanielSvoboda:
> While you're there already, can you make it off by default? It doesn't make much sense for it to be on initially

It shouldn't start automatically, maybe there's a bug, and maybe the problem isn't that people are pressing the button to open the screen. I tested it with the last pre-release without config slavas, and this screen didn't open when starting the game, so I still don't know if it's really a bug, or if people are pressing ctrl+F10 by accident, but in any case an X button should help....

GHU7924:
It's a small thing, but it's nice.

PR #2190, opened 2025-01-19
Fix build error caused by having vulkan-headers > 1.4.303 installed

PowellNGL:
The build fails on my machine because I have vulkan-headers 1.4.304 installed and vk_platform.cpp requires 1.4.303. I've added `EXACT` to the find_package call for vulkan-headers to make cmake fall back to the expected version in externals/vulkan-headers instead

The error I get when it tries to build with 1.4.304:
```
[build] /home/nick/shadPS4/src/video_core/renderer_vulkan/vk_platform.cpp:391:28: error: cannot initialize a member subobject of type 'vk::PFN_DebugUtilsMessengerCallbackEXT' (aka 'unsigned int (*)(vk::DebugUtilsMessageSeverityFlagBitsEXT, Flags<DebugUtilsMessageTypeFlagBitsEXT>, const vk::DebugUtilsMessengerCallbackDataEXT *, void *)') with an lvalue of type 'VkBool32 (VkDebugUtilsMessageSeverityFlagBitsEXT, VkDebugUtilsMessageTypeFlagsEXT, const VkDebugUtilsMessengerCallbackDataEXT *, void *)' (aka 'unsigned int (VkDebugUtilsMessageSeverityFlagBitsEXT, unsigned int, const VkDebugUtilsMessengerCallbackDataEXT *, void *)'): type mismatch at 1st parameter ('vk::DebugUtilsMessageSeverityFlagBitsEXT' vs 'VkDebugUtilsMessageSeverityFlagBitsEXT')
[build]   391 |         .pfnUserCallback = DebugUtilsCallback,
[build]       |                            ^~~~~~~~~~~~~~~~~~
[build] 1 error generated.
```


georgemoralis:
i think it is more proper to update vulkan headers instead

georgemoralis:
fixed in #2197

PR #2188, opened 2025-01-19
Fix SurfaceFormat Format1_5_5_5

DanielSvoboda:
Fix `Castlevania: Symphony of the Night`

Format1_5_5_5
From:  eR5G5B5A1UnormPack16
To:       eA1B5G5R5UnormPack16

| Before | After |
|-------------|-------------|
image | image

[Render.Vulkan] <Error> vk_platform.cpp:DebugUtilsCallback:52: VUID-VkImageViewCreateInfo-usage-08931: Validation Error: [ VUID-VkImageViewCreateInfo-usage-08931 ] Object 0: handle = 0x9675f10000000107, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0xca93720b | vkCreateImageView(): pCreateInfo->

format VK_FORMAT_R5G5B5A1_UNORM_PACK16 with tiling VK_IMAGE_TILING_OPTIMAL doesn't support VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT.
(supported features: VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_BIT|VK_FORMAT_FEATURE_2_BLIT_SRC_BIT|VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_FILTER_LINEAR_BIT|VK_FORMAT_FEATURE_2_TRANSFER_SRC_BIT|VK_FORMAT_FEATURE_2_TRANSFER_DST_BIT|VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_FILTER_MINMAX_BIT|VK_FORMAT_FEATURE_2_HOST_IMAGE_TRANSFER_BIT_EXT).
The Vulkan spec states: If usage contains VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT, then the image view's format features must contain VK_FORMAT_FEATURE_COLOR_ATTACHMENT_BIT or VK_FORMAT_FEATURE_2_LINEAR_COLOR_ATTACHMENT_BIT_NV (https://vulkan.lunarg.com/doc/view/1.3.283.0/windows/1.3-extensions/vkspec.html#VUID-VkImageViewCreateInfo-usage-08931)

PR #2191, opened 2025-01-19
Fix SurfaceFormat Format1_5_5_5 - Format5_5_5_1

DanielSvoboda:
As discussed in: https://github.com/shadps4-emu/shadPS4/pull/2188

Format1_5_5_5 was changed to eA1R5G5B5UnormPack16
and the r-b, b-r components swapped.
Because eA1B5G5R5UnormPack16 was only added with Vulkan 1.4

add a mapping for Format5_5_5_1 as eR5G5B5A1UnormPack1

*I was unable to use eR5G5B5A1UnormPack16, even changing the components to the correct order, the colors were bad.

In the end the result was the same/good:
image

### DIFF
--- a/src/video_core/amdgpu/types.h
+++ b/src/video_core/amdgpu/types.h
@@ -292,6 +292,14 @@ inline CompMapping RemapSwizzle(const DataFormat format, const CompMapping swizz
         result.a = swizzle.r;
         return result;
     }
+    case DataFormat::Format1_5_5_5: {
+        CompMapping result;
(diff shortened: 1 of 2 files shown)


squidbus:
Could you also remove the `Format5_5_5_1` swizzle and add a mapping for it to `eR5G5B5A1UnormPack16`? Since we now know that should be the bit order for that format.

squidbus (approved):
LGTM, thanks will merge when the CI finishes

Issue #2154, opened 2025-01-15
[Feature Request]: Customisable save data path (for NAS or cloud drive)


shinra-electric:
### Checklist

- [X] I have searched for a similar issue in this repository and did not find one.

### Description

Request to add a customisable path for save data in the UI. 

### Reason

I have several computers. Currently, in order to sync saves I need to keep copying from one to another and determining which is the latest. If the save data was remote such as on a cloud drive, it could be accessed from any of them.

This request is **just** for the save data folder, not the whole shadPS4 data folder, as that can grow very large in size.

GHU7924:
Issue 1909 ([Feature request] Automatic backup of savedata) was recently released, I didn't like the idea of automation, but I liked the idea of creating a backup copy of the save, even a PR 1967 was created for implementation (it has already been closed), and in general I had the same thoughts as you, set paths for saves https://github.com/shadps4-emu/shadps4-qtlauncher/issues/58 and has multiple saves https://github.com/shadps4-emu/shadPS4/pull/1967#issuecomment-2565260489.

You can do something like this

![Image](https://github.com/user-attachments/assets/49ce571c-80d2-41a6-bea5-5e67fda69259)

shinra-electric:
This has been implemented in shadps4-emu/shadPS4#2199 

I've tested it by setting the save folder to an iCloud Drive, and accessing it using two different Macs. 

It has worked without issue so far, so I'll close this FR.

PR #2199, opened 2025-01-21
Added ability to change save data path

rootexpm:
This is for issue #2154 

Saving to SMB shares should work directly, but I have not tested this as I don't have access to any SMB shares.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -113,7 +113,7 @@ bool GetLoadGameSizeEnabled() {
 
 std::filesystem::path GetSaveDataPath() {
     if (save_data_path.empty()) {
-        return Common::FS::GetUserPath(Common::FS::PathType::UserDir) / "savedata";
+        return Common::FS::GetUserPath(Common::FS::PathType::SaveDataDir);
     }
     return save_data_path;
 }
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -78,6 +78,7 @@ static std::string trophyKey;
 static bool load_game_size = true;
 std::vector<std::filesystem::path> settings_install_dirs = {};
 std::filesystem::path settings_addon_install_dir = {};
+std::filesystem::path save_data_path = {};
 u32 main_window_geometry_x = 400;
 u32 main_window_geometry_y = 400;
 u32 main_window_geometry_w = 1280;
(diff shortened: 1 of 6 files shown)


Foul-Tarnished:
What happen if the network connection drop during the save is being made/modified ?
Corrupted file ?

Or disconnection between two saves, will the new one be made in a temporary folder ?

rootexpm:
I'm not sure honestly, I could try implementing temporary files saved locally , try saving them and maybe having a retry if unsucessful

squidbus:
> What happen if the network connection drop during the save is being made/modified ? Corrupted file ?

I would think that should be the responsibility of whatever network share or sync service you use, not the emulator.

Foul-Tarnished:
I don't think Windows handle gracefully when you disconnect  when copying
over a SMB share.

Iirc you just get an error message with a retry button.
Probably not what happens when you use Windows API

src/common/config.cpp
```diff
@@ -110,6 +111,13 @@ bool GetLoadGameSizeEnabled() {
     return load_game_size;
 }
 
+std::filesystem::path GetSaveDataPath() {
+    if (save_data_path.empty()) {
+        return Common::FS::GetUserPath(Common::FS::PathType::UserDir) / "savedata";
```
roamic: Shouldn't this be `Common::FS::GetUserPath(Common::FS::PathType::SaveDataDir)`?

rootexpm: Yeah, ill change that.

GHU7924:
Personally, I think it's better not to change the path for saving (although this option expands the user's capabilities).
What you added should be the path for a COPY of the original saves.

I do not know if it is possible to make the emulator do 2 saves at once instead of one save, that is, the same data is in 2 different folders, something like HDD with Raid. And you can do anything with this copy.

You also need two buttons that will allow you to move the saves so that you don't have to do it manually.
For example, I can copy original saves along the path that is configured in Path or vice versa, copy saves from that path to the original save folder.

roamic (approved):

GHU7924:
After merging, clicking on Settings opens the Paths tab instead of General.

P.S. In that case, can someone do a PR so that you can set the default tab? So that you can choose which tab will be displayed when opening Settings.

src/qt_gui/settings_dialog.ui
```diff
@@ -53,7 +53,7 @@
       </sizepolicy>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>3</number>
```
rootexpm, line 56: This is the issue, should be 0 instead of 3.

I'll see if i can do the default window that opens

PR #2203, opened 2025-01-22
tile: Fix some tile thickness calculation errors.

squidbus:
Fix some mistakes from updates to tile calculations related to volume thickness:
* Typo in arg order to `ImageSizeMicroTiled` in `UpdateSize`.
* `ImageSizeMicroTiled` should not return size multiplied with thickness, only use it for pitch alignment padding condition. Returned value will be multiplied with full texture depth after.


squidbus:
One more fix: the mip height should not be padded to tile height as this is handled with appropriate mip size value calculation. Height should still just be the actual mip image height. This fixes some of the remaining issues I was seeing with a 3D image where one of the slices would be garbage.

PR #2209, opened 2025-01-22
Update Japanese translation

creeper-0910:


src/qt_gui/translations/ja_JP.ts
```diff
@@ -182,35 +182,35 @@
...
 		</message>
 		<message>
 			<source>Delete %1</source>
-			<translation>Delete %1</translation>
+			<translation>%1 を削除</translation>
 		</message>
 		<message>
 			<source>Are you sure you want to delete %1's %2 directory?</source>
-			<translation>Are you sure you want to delete %1's %2 directory?</translation>
+			<translation>%1 内の %2 ディレクトリを本当に削除しますか?</translation>
```
ElBread3: Should be more like "%1 の %2 ディレクトリを本当に削除しますか". In use this can be formatted as "Are you sure you want to delete Bloodborne's game directory?", so phrasing it as "the game inside of Bloodborne" would be inaccurate.

src/qt_gui/translations/ja_JP.ts
```diff
@@ -790,71 +790,71 @@
 		</message>
 		<message>
 			<source>TrophyKey</source>
-			<translation>Trophy Key:\nKey used to decrypt trophies. Must be obtained from your jailbroken console.\nMust contain only hex characters.</translation>
+			<translation>トロフィーキー:\nトロフィーの暗号化を解除するためのキーです。脱獄済みのコンソールから取得することができます。\n16進数のみを受け入れます。</translation>
```
ElBread3: The phrase "暗号化を解除する" to me sounds like we are circumventing the encryption, which would be very concerning! "復号する" or "認証する" may work better, or perhaps even just calling it a "トロフィーの解読キー"

ElBread3 (changes requested):

src/qt_gui/translations/ja_JP.ts
creeper-0910: Thank you very much! Fixed!

ElBread3: Actually now that I think about it, using "認証" would describe "a key that authorizes trophy encryption" which doesn't sound right, so you'd have to do "トロフィーの解読キーです" or use "復号する"

ElBread3 (approved):

PR #2210, opened 2025-01-22
update extension vector capacity

panzone91:
Updated the capacity for the vector used to keep track of enable Vulkan extensions


georgemoralis:
was there an issue with old size?

panzone91:
Yup, on my hardware it would caused an overflow because of commit 80092b6367e1aa230fb6e72fb304a0828061d409

PR #2211, opened 2025-01-22
Update Linux building documentation

tomboylover93:
This aims to:

- Make the instructions more direct;
- Add some much needed information about building on distros other than the ones specified;
- Add other ways to build shadPS4 from source based on what the user finds to be easier.


kalaposfos13:
Can you make the
```bash
cmake -S . -B build/ -DENABLE_QT_GUI=ON -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
cd build/
cmake --build . --parallel$(nproc)
./shadps4
```
commands into:
```bash
cmake -S . -B build/ -DENABLE_QT_GUI=ON -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
cmake --build ./build --parallel$(nproc)
./build/shadps4
```
One can argue that this is mostly just personal preference, but this makes it easier to use, as it doesn't include folder switching, and is one less command.

Also, you should make a note about the fact that you can now open games by providing the CUSA IDs as an argument. (only if the game folder is already set in config.toml)

tomboylover93:
@georgemoralis I forgot to change this after changing the lines as kalaposfos mentioned, can you do it for me?

```
diff --git a/documents/building-linux.md b/documents/building-linux.md
index 2672c70f..8621d8ac 100644
--- a/documents/building-linux.md
+++ b/documents/building-linux.md
@@ -99,7 +99,7 @@ Open `cmake-gui` and specify the source code and build directories. If you clone
 
 Click on Configure, select "Unix Makefiles", select "Specify native compilers", click Next and choose `clang` and `clang++` as the C and CXX compilers. Usually they are located in `/bin/clang` and `/bin/clang++`. Click on Finish and let it configure the project.
 
-Now every option should be displayed in red. Change anything you want, then click on Generate to make the changes permanent, then open a terminal window and do steps 2 and 3 of Option 1.
+Now every option should be displayed in red. Change anything you want, then click on Generate to make the changes permanent, then open a terminal window and do step 2 of Option 1.
 
 #### Option 3: Visual Studio Code
```

georgemoralis:
done

abouvier:
`shadps4-git` doesn't force any compiler, but uses the one configured by the user (which is `GCC` by default). `PKGBUILD` should strive to respect the configuration of the user not the preference of the developer. Some chose to compile everything with `clang` or without any optimization flag.

tomboylover93:
In that case, @georgemoralis could you change this?

```
-**Note**: The `shadps4-git` AUR package is not maintained by any of the developers, and it uses GCC as the compiler as opposed to Clang. Use at your own discretion.
+**Note**: The `shadps4-git` AUR package is not maintained by any of the developers, and it uses the default compiler, which is often set to GCC. Use at your own discretion.
```

georgemoralis:
done

tomboylover93:
Markdown isn't displaying properly, there needs to be a space before that first hashtag. 
image

georgemoralis:
ok..

PR #2212, opened 2025-01-22
Optimize workflows by caching `apt install`

Virgula0:
This PR:

- Attempts to gain a few seconds by using `awalsh128/cache-apt-pkgs-action` for caching apt installs on `linux` building workflows

It is not that much but at least half minute per job, after the first push, is gained


Virgula0:
For some reasons `libfuse2` seems having problems with caching mechanism. I moved it outside.

EDIT:

qt packages seems to have the same problem

PR #2214, opened 2025-01-23
Fix showing debug menu bar / Devtools

DanielSvoboda:
After analyzing with @viniciuslrangel , we concluded that the issue occurs when the emulator crashes and the value of 'showing_debug_menu_bar' is saved incorrectly, causing the DevTools screen to appear the next time the game is started.

### DIFF
--- a/src/core/debug_state.cpp
+++ b/src/core/debug_state.cpp
@@ -17,6 +17,8 @@ using namespace DebugStateType;
 
 DebugStateImpl& DebugState = *Common::Singleton<DebugStateImpl>::Instance();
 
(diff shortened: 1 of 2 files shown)


viniciuslrangel:
* THIS IS A TEMPORARY SOLUTION UNTIL WE HAVE A PROPER GRACEFUL SHUTDOWN

ngoquang2708:
> * THIS IS A TEMPORARY SOLUTION UNTIL WE HAVE A PROPER GRACEFUL SHUTDOWN

You mean permanent.

PR #2217, opened 2025-01-23
texture_cache: Do not overwrite overlap hit with a miss.

squidbus:
When resolving overlaps, don't overwrite an overlap hit with a miss.

Fixes fog effect in battles in CUSA16404 incorrectly covering the entire screen. The depth buffer was not being matched and instead a new, empty color image was being created.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -345,8 +345,13 @@ ImageId TextureCache::FindImage(BaseDesc& desc, FindFlags flags) {
             view_slice = -1;
 
             const auto& merged_info = image_id ? slot_images[image_id].info : info;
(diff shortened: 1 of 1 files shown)


georgemoralis:
omg squidwars are back again!

bigol83:
i checked this with Red Dead Redemption and it always crashes ingame a bit before starting the first mission with this log error

```
[Debug] <Critical> vk_scheduler.cpp:operator():155: Assertion Failed!
Device lost during submit
```

squidbus:
Marking as draft, I think this should actually already be solved but is creating and returning the wrong image for some reason.

squidbus:
This should be correct now.

bigol83:
> This should be correct now.

It doesn't crash anymore 👍

PR #2220, opened 2025-01-23
Revert "Optimize workflows by caching `apt install`"

georgemoralis:
Reverts shadps4-emu/shadPS4#2212


georgemoralis:
revert since it appears i breaks main

PR #2224, opened 2025-01-24
Fix AutoUpdate - Changelog

DanielSvoboda:
It should solve this print problem, before it used the shortened commit hash, now it will use the complete one to compare the changelog, below is the old and new link

https://api.github.com/repos/shadps4-emu/shadPS4/compare/cef92fb...fb738bc
https://api.github.com/repos/shadps4-emu/shadPS4/compare/cef92fbcaad34e0caa90bc5afbce5e5b30c43636...fb738bc

image


se6799 (commented):
Yelp

PR #2226, opened 2025-01-24
shader_recompiler: Implement shader export formats.

squidbus:
Adds support for handling the shader export format register when performing MRT exports. The format can decide which components are exported and what number format they are provided to the export instruction in.


Randomuser8219:
In Knack, the unreachable that it hits now has less invalid types:
```
[Debug] <Critical> value.cpp:operator():150: Unreachable code!
Invalid type Attribute|U1|U16|F32|F64|U32x2|F32x2|F32x3
```

squidbus:
> In Knack, the unreachable that it hits now has less invalid types:
> 
> ```
> [Debug] <Critical> value.cpp:operator():150: Unreachable code!
> Invalid type Attribute|U1|U16|F32|F64|U32x2|F32x2|F32x3
> ```

I don't think this has anything to do with that? It would be easier to know what this error is if you launched in a debugger and found a stack trace for where the assert is hit from.

Randomuser8219:
```
(179c.339c): C++ EH exception - code e06d7363 (first chance)
(179c.339c): C++ EH exception - code e06d7363 (first chance)
(179c.339c): C++ EH exception - code e06d7363 (first chance)
(179c.339c): Security check failure or stack buffer overrun - code c0000409 (!!! second chance !!!)
Subcode: 0x7 FAST_FAIL_FATAL_APP_EXIT 
ucrtbase!abort+0x4e:
00007ffe`a79a502e cd29            int     29h
```
I used WinDBG

squidbus:
That output doesn't really mean anything, I need to know where in the code the function with the assert is being called from.

polybiusproxy (approved):
Tested, works well. The code looks good to me as well.

Randomuser8219:
> That output doesn't really mean anything, I need to know where in the code the function with the assert is being called from.

@squidbus 

```
namespace std {
std::size_t hash<Shader::IR::Value>::operator()(const Shader::IR::Value& v) const {
    using namespace Shader::IR;

    u64 h = HashCombine(static_cast<u64>(v.type), 0ULL);

    switch (v.type) {
    case Type::Void:
        return h;
    case Type::Opaque:
        return reinterpret_cast<u64>(v.InstRecursive());
    case Type::ScalarReg:
        return HashCombine(static_cast<u64>(v.sreg), h);
    case Type::VectorReg:
        return HashCombine(static_cast<u64>(v.vreg), h);
    case Type::Attribute:
        return HashCombine(static_cast<u64>(v.attribute), h);
    case Type::U1:
        return HashCombine(static_cast<u64>(v.attribute), h);
    case Type::U8:
        return HashCombine(static_cast<u64>(v.imm_u8), h);
    case Type::U16:
    case Type::F16:
        return HashCombine(static_cast<u64>(v.imm_u16), h);
    case Type::U32:
    case Type::F32:
        return HashCombine(static_cast<u64>(v.imm_u32), h);
    case Type::U64:
    case Type::F64:
        return HashCombine(static_cast<u64>(v.imm_u64), h);
    case Type::U32x2:
    case Type::U32x3:
    case Type::U32x4:
    case Type::F16x2:
    case Type::F16x3:
    case Type::F16x4:
    case Type::F32x2:
    case Type::F32x3:
    case Type::F32x4:
    case Type::F64x2:
    case Type::F64x3:
    case Type::F64x4:
    default:
        break;
    }
    UNREACHABLE_MSG("Invalid type {}", v.type);
}
} // namespace std
```

PR #2232, opened 2025-01-24
Fix heap read after free in UnmapMemoryImpl

hspir404:
This doesn't fix any known behavior, but it fixes a read after free.

`new_it` can become invalidated in `MergeAdjacent`, and the known safe iterator is returned from the function. The old code was discarding that new iterator, and continuing to access memory of the (potentially merged and potentially deleted) data.

The read of `vma.prot` was hanging my debugger (w/ the aggressive debug assertion settings I had enabled in Application Verifier).

I'm not sure what significance the `readonly_file` variable/parameter has here, but it had a chance of being inaccurate before this change.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -418,8 +418,9 @@ u64 MemoryManager::UnmapBytesFromEntry(VAddr virtual_addr, VirtualMemoryArea vma
     vma.phys_base = 0;
     vma.disallow_merge = false;
     vma.name = "";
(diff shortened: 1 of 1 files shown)


squidbus (approved):

squidbus:
There seems to be a conflict with the recent unmap fix change, if you could resolve that.

hspir404:
> There seems to be a conflict with the recent unmap fix change, if you could resolve that.

was just a whitespace change. I rebased on my end and resolved it

PR #2233, opened 2025-01-24
Rebase of "Handle munmap over multiple VMAs"

kalaposfos13:
Manually updated, since the original author hasn't been active lately.
Link to the original: https://github.com/shadps4-emu/shadPS4/pull/1822

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -389,8 +389,7 @@ s32 MemoryManager::UnmapMemory(VAddr virtual_addr, size_t size) {
     return UnmapMemoryImpl(virtual_addr, size);
 }
 
-size_t MemoryManager::UnmapBytesFromEntry(VAddr virtual_addr, VirtualMemoryArea vma_base,
-                                          size_t size) {
+u64 MemoryManager::UnmapBytesFromEntry(VAddr virtual_addr, VirtualMemoryArea vma_base, u64 size) {
     const auto vma_base_addr = vma_base.base;
     const auto vma_base_size = vma_base.size;
     const auto type = vma_base.type;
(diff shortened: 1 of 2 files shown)


src/core/memory.cpp
```diff
@@ -529,8 +538,8 @@ int MemoryManager::VirtualQuery(VAddr addr, int flags,
     info->is_flexible.Assign(vma.type == VMAType::Flexible);
     info->is_direct.Assign(vma.type == VMAType::Direct);
     info->is_stack.Assign(vma.type == VMAType::Stack);
-    info->is_pooled.Assign(vma.type == VMAType::PoolReserved || vma.type == VMAType::Pooled);
-    info->is_committed.Assign(vma.IsMapped());
+    info->is_pooled.Assign(vma.type == VMAType::PoolReserved);
+    info->is_committed.Assign(vma.type != VMAType::Reserved);
```
squidbus: I don't think these changes are supposed to be here.

src/core/memory.cpp
```diff
@@ -389,32 +389,29 @@ s32 MemoryManager::UnmapMemory(VAddr virtual_addr, size_t size) {
...
-    ASSERT_MSG(vma_base.Contains(virtual_addr, size),
-               "Existing mapping does not contain requested unmap range");
-
+size_t MemoryManager::UnmapBytesFromEntry(VAddr virtual_addr, VirtualMemoryArea vma_base,
+                                          size_t size) {
+    const auto vma_base_addr = vma_base.base;
+    const auto vma_base_size = vma_base.size;
     const auto type = vma_base.type;
     if (type == VMAType::Free) {
         return ORBIS_OK;
```
squidbus: I think this should return `adjusted_size` as this function was changed by PR to return a size, not a result code.

src/core/memory.cpp
kalaposfos13: No, if you check main, it's like this there
It's possible that I'm wrong, but I tried to get the changes from both versions as accurately as I could do
https://github.com/shadps4-emu/shadPS4/blob/main/src/core/memory.cpp#L532-L533

squidbus: If it was matching with main there would be no changes here. You've undone what's in main. The original PR didn't even change this code.

kalaposfos13: I see now, for some reason I was comparing it backwards, but yes, this does undo the changes in main.

squidbus:
Your code style changes to the function definitions at the end need the corresponding changes in `memory.h` as well

georgemoralis:
ok some reviews on this?

squidbus (commented):
I've tested this and it seems to work, approach looks fine to me as well. Would like a second opinion on this though as it's a part of the code I'm less familiar with.

georgemoralis:
looking good i vote for this

PR #1815, opened 2024-12-18
Handle MEM_SEMAPHORE PM4 packets

mikusp:
Implements MEM_SEMAPHORE as a global barrier in the pipeline, that is probably not optimal from the performance perspective, but allows games to process further.

If you have any comments let me know and I'll try to fix

PR #2235, opened 2025-01-24
liverpool: Implement PM4 MEM_SEMAPHORE.

squidbus:
Implements the `MEM_SEMAPHORE` PM4 packet, which signals or yields for wait on a memory address semaphore value.

Used by CUSA01047

### DIFF
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -636,6 +636,18 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
                 }
                 break;
             }
+            case PM4ItOpcode::MemSemaphore: {
+                const auto* mem_semaphore = reinterpret_cast<const PM4CmdMemSemaphore*>(header);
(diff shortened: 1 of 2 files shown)


StevenMiller123:
Is this different from how https://github.com/shadps4-emu/shadPS4/pull/1815 did it?

squidbus:
> Is this different from how #1815 did it?

Yes, this implements what I believe should be the correct behavior while that PR just emitted a global barrier into the Vulkan command buffer.

KITT1995:
Ratchet & Clank to menu
image

ElBread3:
@squidbus Shouldn't the mem_semaphore->Decrement(); call be inside the while loop instead of outside of it?

squidbus:
> Shouldn't the mem_semaphore->Decrement(); call be inside the while loop instead of outside of it?

No, semaphore waits for value to be greater than 0 (signalling either adds 1 or sets to 1) and then decrements it by 1.

ElBread3:
Got it, the code made it look slightly more overcomplicated than it really was, mb

PR #2236, opened 2025-01-24
shader_recompiler: Fix image write swizzles.

squidbus:
Component mapping for swizzle needs to be inverted for image writes. For example, if the swizzle is [000R], it means that alpha should be written to red, not that red should be written to alpha.

Fixes videos in CUSA01047 intro. I believe this fixes baldness in CUSA00900 as well (https://github.com/shadps4-emu/shadPS4/issues/1858).


StevenMiller123:
image
image
image
Fixes copyright screens in Marvel's Spider-Man.

bigol83:
> Component mapping for swizzle needs to be inverted for image writes. For example, if the swizzle is [000R], it means that alpha should be written to red, not that red should be written to alpha.
> 
> Fixes videos in CUSA01047 intro. I believe this fixes baldness in CUSA00900 as well (#1858).

Yes, it fixes Bloodborne baldness

image

rafael-57:
Thank you!!!!! Looks great
photo_2025-01-25_23-20-31

PR #2238, opened 2025-01-24
system_service: Add simple event queue and push an EntitlementUpdate event to it when app content is initialized

Sazpaimon:
Fist of the North Star: Lost Paradise expects an `EntitlementUpdate` system event to be fired before it checks if DLC is installed. It does call `sceAppContentInitialize` beforehand, so I added a simple queue to hold the events and pushed the `EntitlementUpdate` event to it when `sceAppContentInitialize` is called. This allows installed DLC to show up ingame now.

I'm not sure if this is 100% the correct behavior, but it made the most sense to me.

image

### DIFF
--- a/src/core/libraries/app_content/app_content.cpp
+++ b/src/core/libraries/app_content/app_content.cpp
@@ -12,6 +12,7 @@
 #include "core/file_sys/fs.h"
 #include "core/libraries/app_content/app_content_error.h"
 #include "core/libraries/libs.h"
+#include "core/libraries/system/systemservice.h"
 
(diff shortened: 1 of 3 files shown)


red-prig:
For ps4 there are several types of events that come during initialization, so apparently this is the correct behavior

georgemoralis:
looks ok but are we sure the event process in FIFO order?

squidbus:
> looks ok but are we sure the event process in FIFO order?

I think it's probably okay until proven otherwise, not like we have much of an event implementation to speak of currently.

PR #2241, opened 2025-01-26
memory_patcher: Remove hardcoded repositories when loading patches

Sazpaimon:
The GUI already supports showing and toggling patches existing in arbitrary directories under the patches folder, so long as it has a files.json, so the loader should do it as well.


squidbus:
Needs fix for clang-format errors.

Sazpaimon:
done

Issue #2242, opened 2025-01-26
[APP BUG]: Unknown game entry on grid view that crashes the app, can crash the os if deleted


mAmineChniti:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

there's an imageless game entry when you select grid view, it crashes the app when you have it selected and change game list icons size, if you delete that game entry it deletes everything in ~/
i'm using the official flatpak

### Reproduction Steps

![Image](https://github.com/user-attachments/assets/acaa8c7b-e6b3-4798-8509-abdd874af374)

- Game list mode -> grid view
- select the imageless nameless game entry
- change the game list icons size, or delete that game entry

### Specify OS Version

Debian 12

roamic:
Can you attach `config.toml` from the shadPS4 user directory? Also, what's the contents of shadPS4 games directory?

mAmineChniti:
i'm not sure how to access the config.toml, this is the flatpak version btw, and the games dir only has that deadpool game only

mAmineChniti:
```toml
[General]
logType = "async"
separateUpdateEnabled = false
showSplash = false
Fullscreen = false
checkCompatibilityOnStartup = false
updateChannel = "Release"
logFilter = ""
enableDiscordRPC = false
compatibilityEnabled = true
autoUpdate = false
userName = "shadPS4"
isTrophyPopupDisabled = false
playBGM = false
isPS4Pro = true
BGMvolume = 50

[Input]
cursorState = 1
cursorHideTimeout = 5
backButtonBehavior = "left"
specialPadClass = 1
useSpecialPad = false

[GPU]
screenWidth = 1280
screenHeight = 720
nullGpu = false
copyGPUBuffers = false
vblankDivider = 1
dumpShaders = false
patchShaders = true

[Vulkan]
gpuId = 1
validation = false
rdocMarkersEnable = false
validation_sync = false
validation_gpu = false
rdocEnable = false
crashDiagnostic = false

[Debug]
DebugDump = false
CollectShader = false

[GUI]
geometry_y = 36
sliderPosGrid = 0
geometry_w = 1920
mw_width = 1920
gameTableMode = 0
theme = 0
recentFiles = []
pkgDirs = []
[...]

PR #2243, opened 2025-01-26
qt: Prevent interacting with empty grid cells.

squidbus:
Prevent interacting with empty cells in the game grid view, to prevent actions on out-of-bounds game list reads with potentially destructive results. Unfortunately I am not familiar enough with Qt to figure out how to disable selecting the empty cells that pad out rows to the number of grid columns, but this will at least prevent it from causing harm.

Should prevent https://github.com/shadps4-emu/shadPS4/issues/2242


georgemoralis:
@DanielSvoboda can you take a look to that issue?

PR #2245, opened 2025-01-26
renderer_vulkan: use LDS buffer as SSBO on unsupported shared memory size

### LINKED ISSUE #2181: [GAME BUG]: CUSA03220 Gran Turismo®Sport - [Debug] <Critical> spirv_emit_context.cpp:operator():861: Assertion Failed!
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [x] I have disabled all patches and cheats and the issue is still present.
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadps4-game-compatibility?tab=readme-ov-file#informations) installed.

### Describe the Bug

Crashes on Display Settings with latest nightly (like https://github.com/shadps4-emu/shadPS4/issues/2062 but different error)

`[Lib.Kernel] <Error> time.cpp:sceKernelClockGettime:114: unsupported = 16 using CLOCK_REALTIME`

`[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:488: Compiling cs shader 0x58d7a804` 
`[Render.Vulkan] <Warning> resource_tracking_pass.cpp:operator():416: Inline sampler detected`
`[Render.Vulkan] <Warning> resource_tracking_pass.cpp:operator():416: Inline sampler detected`
`[Render.Vulkan] <Warning> resource_tracking_pass.cpp:operator():416: Inline sampler detected`
`[Render.Vulkan] <Warning> resource_tracking_pass.cpp:operator():416: Inline sampler detected`
`[Render.Vulkan] <Info> vk_pipeline_cache.cpp:CompileModule:488: Compiling cs shader 0x170f8dc0 `
`[Debug] <Critical> spirv_emit_context.cpp:operator():861: Assertion Failed!`

### Reproduction Steps

Start the game, go through display settings to HDR/SDR selection and press SDR. The game will crash

### Specify OS Version

Windows 11

### CPU

AMD Ryzen 7 7840HS

### GPU

AMD Radeon 780M Graphics iGPU

### Amount of RAM in GB

16

### Amount of VRAM in GB

4

### Log File

[shad_log.txt](https://github.com/user-attachments/files/18464669/shad_log.txt)

polybiusproxy:
To emulate LDS, a workgroup shared variable is made, however, this can't be done if the GPU has insufficient amounts of shared memory to meet the needs of the shader. Before PR https://github.com/shadps4-emu/shadPS4/pull/2175, if this was the case, there would be device losses.

To avoid this, if the GPU doesn't have enough shared memory size, LDS is emulated using a device local buffer bound as SSBO. This is not as optimal as a workgroup shared variable would be, but it is the only way.

This PR mainly aims to resolve one of the issues present on Gran Turismo Sport (https://github.com/shadps4-emu/shadPS4/issues/2181).

### DIFF
--- a/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
+++ b/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
@@ -65,17 +65,17 @@ void Name(EmitContext& ctx, Id object, std::string_view format_str, Args&&... ar
 
 } // Anonymous namespace
 
-EmitContext::EmitContext(const Profile& profile_, const RuntimeInfo& runtime_info_,
-                         const Info& info_, Bindings& binding_)
+EmitContext::EmitContext(const Profile& profile_, const RuntimeInfo& runtime_info_, Info& info_,
+                         Bindings& binding_)
     : Sirit::Module(profile_.supported_spirv), info{info_}, runtime_info{runtime_info_},
       profile{profile_}, stage{info.stage}, l_stage{info.l_stage}, binding{binding_} {
     AddCapability(spv::Capability::Shader);
     DefineArithmeticTypes();
     DefineInterfaces();
+    DefineSharedMemory();
     DefineBuffers();
     DefineTextureBuffers();
     DefineImagesAndSamplers();
-    DefineSharedMemory();
 }
 
 EmitContext::~EmitContext() = default;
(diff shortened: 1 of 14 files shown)


red-prig:
Binding a global buffer instead of a local one is an idea that is in the air, but in this case I do not see dividing the global buffer into local groups, because without this division shaders will contention for the same data, which is not true because the memory should be local to the shader call group.

jimbigjim1:
If this is not optimal why not just add separate min system requirements for different games? Its obvious you need a more powerful pc to run Gran Turismo instead of some 2D indie game, but how powerful?

squidbus:
> If this is not optimal why not just add separate min system requirements for different games? Its obvious you need a more powerful pc to run Gran Turismo instead of some 2D indie game, but how powerful?

As far as I can tell there is no NVIDIA GPU that has a high enough `maxComputeSharedMemorySize` for cases like the one here. So what you are suggesting is basically that no NVIDIA user can run this game.

If the GPU has enough shared memory it will still be used, and even falling back to an alternative the performance will probably be fine on plenty of GPUs, so I don't see why not.

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
```diff
@@ -852,20 +852,45 @@ void EmitContext::DefineSharedMemory() {
     if (!info.uses_shared) {
         return;
     }
+    const u32 max_shared_memory_size = runtime_info.cs_info.max_shared_memory_size;
```
squidbus: This is technically more of a left over from the last PR on this, but I think `max_shared_memory_size` should be in the `Profile` instead of the `RuntimeInfo`.

5jd:
Still getting `scalar_memory.cpp:operator():49: Unreachable code!` on a AMD RX 7800XT after seletecting SDR
Gran Turismo™SPORT CUSA02168 01.69
OS: Windows 11
[shad_log.txt](https://github.com/user-attachments/files/18558089/shad_log.txt)

polybiusproxy:
That error is unrelated to the issue this PR aims to fix.

PR #2247, opened 2025-01-26
linker: handle relocation for exported modules

panzone91:
In some games the module filename and module exported name are not the same. This PR changes how we decide if there is the need for a symbol relocation based on the export module info.

### DIFF
--- a/src/core/linker.h
+++ b/src/core/linker.h
@@ -109,10 +109,13 @@ public:
 
     void RelocateAnyImports(Module* m) {
         Relocate(m);
-        for (auto& module : m_modules) {
-            const auto imports = module->GetImportModules();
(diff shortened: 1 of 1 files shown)


georgemoralis:
you got any game that uses that ?

panzone91:
I've noticed the same pattern in a lot of M2 Shottriggers releases, but if you need a single game, Dodonpachi Daioujou (https://github.com/shadps4-emu/shadps4-game-compatibility/issues/534) uses a module with filename emu_integration_01700 but export it as libemu_integration (which is the symbol used as the import in the main game), so the relocate wasn't able to work correctly.

georgemoralis:
that seems ok

PR #2248, opened 2025-01-26
qt: fix grid

DanielSvoboda:
This fixes the problem presented in https://github.com/shadps4-emu/shadPS4/pull/2243, but causes another problem.
I'll leave it as a draft in the hope that someone can help to understand this better. 
I believe this cannot be merged because it becomes worse than before...

this should solve the selection of invisible grids:
image

It's a bit complicated to describe the new problem, so it's easier to watch the video, but in short:
after selecting outside the grid, the background images do not load, only if I click again outside the games on the grid

https://github.com/user-attachments/assets/cf11e8a3-38e8-4b6a-84cb-150b10d7bbb8

### DIFF
--- a/src/qt_gui/game_grid_frame.cpp
+++ b/src/qt_gui/game_grid_frame.cpp
@@ -43,12 +43,17 @@ void GameGridFrame::onCurrentCellChanged(int currentRow, int currentColumn, int 
     columnCnt = this->columnCount();
 
     auto itemID = (crtRow * columnCnt) + currentColumn;
+    static QString originalStyle = this->styleSheet();
     if (itemID > m_game_info->m_games.count() - 1) {
         cellClicked = false;
         validCellSelected = false;
         BackgroundMusicPlayer::getInstance().stopMusic();
+        this->setStyleSheet(originalStyle +
+                            "QTableWidget {outline: auto;}"
+                            "QTableWidget::item:selected {background-color: transparent;}");
         return;
     }
+    this->setStyleSheet(originalStyle);
     cellClicked = true;
     validCellSelected = true;
     SetGridBackgroundImage(crtRow, crtColumn);


GHU7924:
I haven't tried your PR, but here's what I noticed in Main. I currently have 11 games, but if I switch between them, then at the end I have a couple of empty objects. At first I didn't understand, but when I went into the folder with the games, I saw that I didn't have 11, but 13 objects. Two additional objects are separate Update folders for games. I believe that they were visible in the game list.


I changed the size of the icons through this menu and you may notice a discrepancy between the selected sizes and the displayed ones.



The emulator starts with a tiny icon, then I change Small - Medium - Large and Tiny again.

DanielSvoboda:
last line will have 'squares' until the end of it
image

GHU7924:
In that case, it's strange that they look different.

https://github.com/user-attachments/assets/77e3c7a3-81e5-42ed-a6cf-90b17de62b28

I have now checked and this problem of icon size mismatch has been present from the very beginning, in shadps4 v0.1.0 it also exists. Also shadps4-emu/shadps4-qtlauncher#36

DanielSvoboda:
yes, their size is based on a calculation of the available size and the number of games, try to make your window smaller on the sides to have a result similar to mine, more 'square'

GHU7924:
For some reason, when selecting tiny icons, the size of the square does not change, and yet I got this.

https://github.com/user-attachments/assets/e0510954-4530-441e-acbe-f7d984365193

With your PR, the squares don't disappear anywhere, they just become invisible. I also have the problem you mentioned.

I was thinking, maybe then we should make a separate background image for these invisible squares? I wonder if this will solve the problem? I suggest something like this.

GHU7924:
The selection is not saved.

https://github.com/user-attachments/assets/184887e1-92bb-4dac-bfa3-ab65d3357e96

I also compared the changes in file config.toml
toml

Some lines are swapped, maybe it's not that important, but this behavior is there.
Two spaces might not be a mistake either, but it caught my eye.

PR #2254, opened 2025-01-28
Add Vulkan debug options to the Debug tab

tomboylover93:
This adds Crash Diagnostics, Collect Shaders, Copy GPU Buffers, Host Debug Markers and Guest Debug Markers to an "Advanced" section in the Debug tab, all with proper descriptions. It mainly helps developers who might want to get more info about an error beyond a simple log from a user, as now these options are more easily accessible.

This comes with one caveat: I couldn't make the config values work with a QCheckBox element as they were, so I changed them (example: vkCrashDiagnosticEnabled is now getVkCrashDiagnosticEnabled, and I added setVkCrashDiagnosticEnabled). As such, I changed every instance where they're used. This might cause issues with how these options are supposed to work, so please let me know if there are any.

Other than that, you now have to manually enable Host and Guest Debug Markers when using Crash Diagnostics, whereas currently if you have Crash Diagnostics enabled these are automatically enabled. This is because they did not automatically disable when disabling Crash Diagnostics from the Settings menu.

image

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -267,18 +267,28 @@ bool vkValidationGpuEnabled() {
     return vkValidationGpu;
 }
 
-bool vkCrashDiagnosticEnabled() {
+bool getVkCrashDiagnosticEnabled() {
     return vkCrashDiagnostic;
 }
 
-bool vkHostMarkersEnabled() {
-    // Forced on when crash diagnostic enabled.
-    return vkHostMarkers || vkCrashDiagnostic;
+bool getVkHostMarkersEnabled() {
+    return vkHostMarkers;
 }
 
-bool vkGuestMarkersEnabled() {
-    // Forced on when crash diagnostic enabled.
-    return vkGuestMarkers || vkCrashDiagnostic;
+bool getVkGuestMarkersEnabled() {
+    return vkGuestMarkers;
+}
+
+void setVkCrashDiagnosticEnabled(bool enable) {
+    vkCrashDiagnostic = enable;
+}
+
+void setVkHostMarkersEnabled(bool enable) {
+    vkHostMarkers = enable;
+}
+
+void setVkGuestMarkersEnabled(bool enable) {
+    vkGuestMarkers = enable;
 }
 
 bool getSeparateUpdateEnabled() {
--- a/documents/building-linux.md
+++ b/documents/building-linux.md
@@ -29,7 +29,7 @@ sudo dnf install clang git cmake libatomic alsa-lib-devel pipewire-jack-audio-co
 sudo pacman -S base-devel clang git cmake sndio jack2 openal qt6-base qt6-declarative qt6-multimedia sdl2 vulkan-validation-layers
 ```
 
-**Note** : The `shadps4-git` AUR package is not maintained by any of the developers, and it uses the default compiler, which is often set to GCC. Use at your own discretion.
+**Note**: The `shadps4-git` AUR package is not maintained by any of the developers, and it uses the default compiler, which is often set to GCC. Use at your own discretion.
 
 #### OpenSUSE
 
--- a/src/common/config.h
+++ b/src/common/config.h
@@ -106,9 +106,12 @@ void setRdocEnabled(bool enable);
 bool vkValidationEnabled();
 bool vkValidationSyncEnabled();
 bool vkValidationGpuEnabled();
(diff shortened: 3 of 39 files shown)


DanielSvoboda:
I will include the translation into Portuguese, and the other languages ​​in English to make it standardized.

GHU7924:
>You need Vulkan Validation Layers enabled and the Vulkan SDK for this to work.

These are the correct lines, because external software is used.
It's good that you added this line to the description.

Actually, I once thought about making such an edit, but I still haven't decided whether it would be superfluous or not.


And a few more things.


Excellent PR.

tomboylover93:
Fixed that last one. As for the first one, I don't think the emphasis on the fact that you need the Vulkan SDK to use Validation Layers is needed, as it is already implied by other options that depend on it.

GHU7924:
>Fixed that last one.

You haven't fixed the issues I've pointed out yet.

https://github.com/user-attachments/assets/6e0d79a4-ceba-4072-95f3-3f655a7c062d

There may be a mistake here, if I understood correctly.

1) src/qt_gui/settings_dialog.cpp - lines 552 and 553 there was no need to delete it. 
- This is how you will return the description for "Enable RenderDoc Debuggin", if I understood correctly.
2) src/qt_gui/translations/en.ts - Line 713 does not match line 712. 
- As I understand it, this problem applies to all translations.

>As for the first one, I don't think the emphasis on the fact that you need the Vulkan SDK to use Validation Layers is needed, as it is already implied by other options that depend on it.

If you think so, then I don't mind.

tomboylover93:
Not sure why Enable RenderDoc Debugging is in the settings twice but I'll fix that and the description issue.

tomboylover93:
@DanielSvoboda Seems like you set the translation to "Copy GPU Buffers" as "Enable RenderDoc Debugging" by mistake. Can you open a PR on my branch to fix that?

GHU7924:
@tomboylover93 You can do it yourself, you need to replace line 713 in all translations with this:

`<translation>Copy GPU Buffers</translation>`

Only in sv.ts it is line 1201 ... and pt_BR.ts doesn't need it.

tomboylover93:
Both should be fixed now.

GHU7924:
I confirm that everything has been fixed, but there is one more big fix left.

You have corrected the first line for the description, but now you need to add these changes for translations as well.
These lines are in en.ts, but not in the rest. You need to add them to the `<translation></translation>` for all translations.

`Collect Shaders:\n`

`Crash Diagnostics:\n`

`Copy GPU Buffers:\n`

`Host Debug Markers:\n`

`Guest Debug Markers:\n`

tomboylover93:
That should be fixed now too.

GHU7924:
@tomboylover93 As for the interface, everything seems to be in order now. I think @DanielSvoboda can review everything additionally, if necessary.
As for the technical part, it is necessary to make a request for verification. 
Good job.

@DanielSvoboda After combining this PR, it may also be worthwhile (to create a PR) to conduct a full reconciliation of the translations. I will try to help with the search.
Are all the lines present everywhere, are there any extra ones, and so on. 
It is also worth checking if there are elements for which you need to add a translation.

At the moment, there doesn't seem to be any PR that can affect the interface and translations, so this is a good opportunity.

src/common/config.cpp
```diff
@@ -267,18 +267,28 @@ bool vkValidationGpuEnabled() {
...
+bool getVkHostMarkersEnabled() {
+    return vkHostMarkers;
 }
 
-bool vkGuestMarkersEnabled() {
-    // Forced on when crash diagnostic enabled.
-    return vkGuestMarkers || vkCrashDiagnostic;
+bool getVkGuestMarkersEnabled() {
+    return vkGuestMarkers;
+}
```
polybiusproxy, line 280: Why was the crash diagnostic logic removed?

tomboylover93: Quoting the description of the PR:

"This is because they did not automatically disable when disabling Crash Diagnostics from the Settings menu."

I couldn't find any workaround for this, and it was very visually jarring to have them enabled even when Crash Diagnostics is disabled.

polybiusproxy: I'm pretty sure that the force-enable here was made to make things easier to show when there's a device loss, so it's not a hard requirement.

polybiusproxy (approved):

PR #2007, opened 2025-01-01
libraries: Add zlib hle skeleton/stub

Nenkai:
Added Zlib library skeleton, while researching the module.
(for a potential HLE implementation, while looking more into it)

PR #2256, opened 2025-01-28
libraries: Implement libSceZlib.

squidbus:
Fully implements `libSceZlib` using the skeleton from https://github.com/shadps4-emu/shadPS4/pull/2007. The implementation uses a thread and queues to manage moving inflate tasks and results back and forth.

Allows CUSA14836 to go in-game.


src/core/libraries/zlib/zlib.cpp
```diff
@@ -0,0 +1,183 @@
...
+    void* dst;
+    u32 dst_length;
+};
+
+struct InflateResult {
+    u32 length;
+    s32 status;
+};
+
+Kernel::Thread task_thread;
```
Osyotr, line 33: These variables should be static or moved to an unnamed namespace.

PR #2263, opened 2025-01-28
Update README.md: Add note to keyboard mapping that some operating systems may require you to hold the Fn button

C4ndyF1sh:
On macOS the fn button needs to be used to use the F-buttons like F10 for example because they are mapped to a different function on default, from my testing only the fps counter can be activated this way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ F11 | Fullscreen |
 F12 | Trigger RenderDoc Capture |
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


C4ndyF1sh:
Screenshot 2025-01-28 at 23 39 17

squidbus:
I don’t think this should be framed this way, this is true of any laptop that uses F keys for functions by default. I would just add a note that some keyboards may require you to hold Fn.

C4ndyF1sh:
@squidbus I think I am done image

C4ndyF1sh:
End product (I hope) image

C4ndyF1sh:
Done image

squidbus:
Statement about macOS is still not correct. Debug info and fullscreen are working with command key.

PR #2265, opened 2025-01-29
Libraries: Update libcInternal

kalaposfos13:
Added all stubs, added back all things that were already implemented, and added functions based on what my games called. I have some basic implementation of printf functions, but I'm not really satisfied with the implementation, so I only did the ones that are actually called by games, and will update the rest when I find a better solution (Also, the non string variants should interface with shadPS4's tty system as well). The new things are:
More math functions
More str functions
setjmp, longjmp (Driveclub needs it to work)
memmove

The main thing that needs implementing for Gravity Rush 2 (and probably other games as well) is the sceLibcMspace functions, which are their own memory management sublibrary, from what I can see.

Logs filtered by `*:Critical Lib.LibcInternal:Info` are welcome if they contain stubs

This probably won't help a lot of games go further without sceLibcInternal as LLE (though implementing sceLibcMspace might do that), and my only game that reached further was Driveclub.


squidbus:
This is a really large file, it should probably at least be split by the type of function.

Also to be clear, games don't use this directly, it is used by other system modules only. So if your game needs it it's just that a LLE system module you have loaded for the game needs it.

squidbus:
And lastly as you can see from the build results, defining the symbols as is will conflict with the host standard library, which is why we added `internal_` before.

georgemoralis:
libc_internal_math , libc_interna_memory files as well

georgemoralis:
fix that clang already ;D

kalaposfos13:
I thought first I'd need to separate the important things like memory and math functions into their own file? Clang is less than a minute, so I'll just do that when I'm done

georgemoralis:
ok one suggestion , move the implemented functions on the top of the file , it will be easier to review them that searching all that big files

src/core/libraries/libc_internal/libc_internal.cpp
```diff
@@ -2,306 +2,16258 @@
...
+    return ORBIS_OK;
+}
+
+s32 PS4_SYSV_ABI internal_cbrtl() {
+    LOG_ERROR(Lib_LibcInternal, "(STUBBED) called");
+    return ORBIS_OK;
+}
+
+s32 PS4_SYSV_ABI internal_clearerr() {
+    LOG_ERROR(Lib_LibcInter
```
georgemoralis: gmtime_s something missing here...

src/core/libraries/libc_internal/libc_internal_str.cpp
```diff
@@ -0,0 +1,529 @@
...
+}
+
+const char* PS4_SYSV_ABI internal_strrchr(const char* str, int c) {
+    LOG_DEBUG(Lib_LibcInternal, "called");
+    return std::strrchr(str, c);
+}
+
+char* PS4_SYSV_ABI internal_strsep(char** strp, const char* delim) {
+    LOG_DEBUG(Lib_LibcInternal, "called");
+#ifdef _GNU_SOURCE
```
georgemoralis, line 290: ????

georgemoralis (commented):
LOG_DEBUG will spam a lot , at least remove them from the ones that proven to be correct (1:1 matches from std ones)

src/core/libraries/libc_internal/libc_internal_memory.cpp
```diff
@@ -0,0 +1,318 @@
...
+        // Size of 1 is used if 0 is provided.
+        size = 1;
+    }
+    void* ptr = std::malloc(size);
+    ASSERT_MSG(ptr, "Failed to allocate new object with size {}", size);
+    return ptr;
+}
+
+void* PS4_SYSV_ABI internal_malloc(size_t size) {
+    return std::malloc(size);
```
mailwl: i don't like this. Guest ps4 uses own memory allocating, no need to alloc on host

squidbus: For what it's worth (at least in my opinion) this HLE is basically just useful if you want to debug something by intercepting the calls. I don't think there's much interest in actually implementing HLE for this library because a) it has a very large API surface and b) by HLE-ing other modules you no longer need this one.

kalaposfos13: Does that mean that there are plans to implement libraries that come with the games (in sce_modules)? I know that there are some things that use it that are not manually loaded system libraries, like libSceNpToolkit for GR2

squidbus: Maybe, although in all likelihood anything Np related would likely be a stub. NpToolkit definitely doesn't need all this stuff added here anyway.

georgemoralis: It is not that easy to get rid lle modules for example ngs2 will require a lot of effort. Also there are gamemodules that might need it so it is not a useless efford

squidbus: > It is not that easy to get rid lle modules for example ngs2 will require a lot of effort.

Correct, but if you have LLE sysmodules there's no reason to not just put libcInternal with them.

georgemoralis: The less the better 😌

squidbus: Well if you're actually looking for a serious libcInternal HLE replacement then the original comment stands here, passing to host `malloc` won't be acceptable.

georgemoralis: Ah yes it isn't.  I thought it was general talk didn't see it is under a review 🤔

PR #2268, opened 2025-01-29
texture_cache: Make sure left-overlapped mips get marked for rebind.

squidbus:
Make sure a left-overlapped mip level gets marked for rebind, even if the replacement image has not been allocated yet.

Fixes some cases where, instead of rendering to the mip level, it will render to the base image since it wasn't rebound.


squidbus:
Turned out to just be a case where left-overlap mip could miss being marked for rebind, adjusted accordingly.

PR #2286, opened 2025-01-30
Fix game title sorting - List view

DanielSvoboda:
PR https://github.com/shadps4-emu/shadPS4/pull/2284 corrects the sorting when clicking on the header, but the list was already disorganized.
Now it starts in the correct order.

### DIFF
--- a/src/qt_gui/game_list_frame.cpp
+++ b/src/qt_gui/game_list_frame.cpp
@@ -1,8 +1,6 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
-#include <algorithm> // std::transform
-#include <cctype>    // std::tolower
 #include <QToolTip>
 #include "common/config.h"
 #include "common/logging/log.h"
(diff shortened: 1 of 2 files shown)


src/qt_gui/game_list_frame.cpp
```diff
@@ -103,14 +103,22 @@ void GameListFrame::PlayBackgroundMusic(QTableWidgetItem* item) {
...
     this->setColumnHidden(6, !Config::GetLoadGameSizeEnabled());
 
     this->setRowCount(m_game_info->m_games.size());
     ResizeIcons(icon_size);
 
+    if (isInitialPopulation) {
+        std::sort(m_game_info->m_games.begin(), m_game_info->m_games.end(),
+                  [this](const GameInfo& a, const GameInfo& b) {
+                      return CompareStringsAscending(a, b, 1); // Column 1 = Name
+                  });
```
squidbus: I think you can just reuse `SortNameAscending` here.

DanielSvoboda: yes I can use it, I will change it

squidbus (approved):
Thanks, this had been bothering me for a while.

PR #2295, opened 2025-01-31
sceAvPlayerAddSourceEx - error validation

DanielSvoboda:
sceAvPlayerAddSourceEx looks quite complex, however this validation allows it to continue on 'CUSA23185 Biped'
as reverse engineering this seems to be necessary.
image

sceAvPlayerAddSourceEx seems similar to sceAvPlayerAddSource, perhaps using it in some way:
handle->AddSource(sourceDetails->uri.name);
as it contains the address of a local file, but in my tests it caused an error...


DanielSvoboda:
just clarifying that other hacks are necessary to start Biped, something like this:
https://github.com/shadps4-emu/shadPS4/compare/main...DanielSvoboda:shadPS4:memory

roamic:
image
As far as I can see from RE it's just
```
if (uriType != SceAvPlayerUriType::Source || handle == nullptr || details == nullptr) {
    return ORBIS_AVPLAYER_ERROR_INVALID_PARAMS
}
```
There is no special type 99.

DanielSvoboda:
I think there is something wrong with my dump

PR #2298, opened 2025-01-31
Revert "libkernel: handle special case in path for load module"

georgemoralis:
Reverts shadps4-emu/shadPS4#2269

red_prig mentioned that you should do multiple attempts to open the file by the path list . Your implementation will probably break some unity games


panzone91:
Ok, I will try to be more in line with the implementation.

Hermiten:
This will break games that will be tested on 0.6.0 ? 😧

panzone91:
No, it was merged after

georgemoralis:
yah that was the purpose of the release to avoid stuff that might break :D

panzone91:
Hi, quick question: what was the problem here? What was breaking this? You make an example of Unity games, but why should this one break one of them?

PR #2302, opened 2025-01-31
Remapping: Documentation and defaults update + add option to use a unified config

kalaposfos13:
Adds an example of deadzone configuration to the default config, and adds the ability to use only one config file (which will be default.ini if the option is set to true, which it is by default) instead of one config per every game.


Caner452 (commented):
[kalaposfos13:remap-doc-update](url)

PR #2307, opened 2025-01-31
Add Slovak language to emulator

SharkStudiosSK:
Add Slovak language to the language selection in the settings dialog.

* Update the `languages` map in the `InitializeEmulatorLanguages` function to include "Slovak".
* Update the `languageIndexes` vector to include the index for Slovak.


DanielSvoboda:
Some points that might be worth reviewing:

1. You can't remove the 'reuse' header, add it back.
image

2. Two spaces were added there, you must format the 'clang' correctly for it to be correct.
image

3. 'Slovak' should be alphabetized
image

4. I believe that the number 31 should also be added to line 314
image

5. f you also want to add translation to the interface, you need to add a .ts file in the src/qt_gui/translations folder.

PR #2311, opened 2025-02-01
add controller remapping GUI

rainmakerv3:
Add a full GUI for controller remapping. This does not cover KBM, the old text editor can still be accessed through a button on this interface.

Translation text not included in the scope.

image


gandalfthewhite19890404:
PS button design is not copyrighted?

rainmakerv3:
> PS button design is not copyrighted?

Good point, I removed the logo

gandalfthewhite19890404:
> > PS button design is not copyrighted?
> 
> Good point, I removed the logo

What if add Shadps4 logo? let discuss it

rainmakerv3:
> > > PS button design is not copyrighted?
> > 
> > 
> > Good point, I removed the logo
> 
> What if add Shadps4 logo? let discuss it

that'd be kinda nice :) i'm not that good at image editing, maybe some can make it nicer than this

image

kalaposfos13:
Some thoughts: 
You can switch what config you're editing, but can't change the new per-game or unified config toggle, unless you go into the text editor.
I also think that instead of opening the old GUI in a new window, it might be better to add it as a new tab (see the image below)
image
Saving doesn't work on Linux with the default config due to the file being named Default.ini instead of default.ini
There is no way to unbind a button
According to the git commit names, there is a feature to allow multiple outputs to a single input which I neither found nor will it work, as the engine doesn't support it, only the other way around (multiple inputs to one output). Looking over the code, just mixed it up in the commit name, and it's fine otherwise
The Defaults button doesn't have the same functionality as the defaults button in the text editor, you should change one of them

src/qt_gui/control_settings.cpp
```diff
@@ -0,0 +1,419 @@
...
+        if (std::find(controlleroutputs.begin(), controlleroutputs.end(), input_string) !=
+                controlleroutputs.end() ||
+            output_string == "analog_deadzone") {
+            line.erase();
+            continue;
+        }
+        lines.push_back(line);
+    }
+
+    lines.resize(lines.size() - 6);
```
kalaposfos13: This is not the best way to handle whitespace, as it can lead to unintended side effects. What this does is strip out the empty lines after the `# Controller bindings` line, but it won't work if the user added custom bindings in this range, because then for every new manually added line, the first few lines after the section separator comment get preserved, regardless of if they are empty lines or bindings. Also, if a user for some reason deletes the newlines, this will get rid of even more things, possibly even affecting the Keyboard section.

rainmakerv3:
> Some thoughts: You can switch what config you're editing, but can't change the new per-game or unified config toggle, unless you go into the text editor. I also think that instead of opening the old GUI in a new window, it might be better to add it as a new tab (see the image below)

- I didn't even notice the new toggle :despair:. For the additional tab, my concern is that it is not visible enough, especially since the window is pretty big. I'm actually considering initially considering the same separate keyboard/controller buttons like what I did for the fork. Honestly I'm still not sure which is the best option.

 Saving doesn't work on Linux with the default config due to the file being named Default.ini instead of default.ini 
- will fix, thanks for the heads up

There is no way to unbind a button 
- Unbind a button, meaning it just has no output?

According to the git commit names, there is a feature to allow multiple outputs to a single input which I neither found nor will it work, as the engine doesn't support it, only the other way around (multiple inputs to one output). Looking over the code, just mixed it up in the commit name, and it's fine otherwise 
- Yes, I meant multiple inputs to an output

The Defaults button doesn't have the same functionality as the defaults button in the text editor, you should change one of them
- What specifically should be the same?

kalaposfos13:
> What specifically should be the same?

Mine does two things based on what config is currently open:
Default: Revert to the builtin default
Everything else: Revert to the default that the user set for themself

Yours just reverts everything to the builtin default currently

rainmakerv3:
> > What specifically should be the same?
> 
> Mine does two things based on what config is currently open: Default: Revert to the builtin default Everything else: Revert to the default that the user set for themself
> 
> Yours just reverts everything to the builtin default currently

I do feel reverting to the original default values is more intuitive (for controller most especially), though maybe someone else should decide :kek:

kalaposfos13:
You're right, for controllers it does make more sense to revert to the builtin default, but I feel like the same is not true for KBM, especially after the recent rollback to the original keybindings. I guess a popup to select which one to do when clicking the Default button might be the best choice here?

Also:
> Unbind a button, meaning it just has no output?

Yes, exactly that. You could maybe add something like a `-` to represent it in the drop down menu (and you can even save it like that, my parser will throw a warning about syntax errors but will just skip the line, or alternatively you can suppress that warning if the input is that, should be an easy change). It should also be the default option if the binding isn't in the file (which is a very nitpicky edge case, but still), so things aren't randomly bound to cross after a file is saved with your save function.

src/qt_gui/control_settings.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+    deadzonevalue = std::to_string(ui->RightDeadzoneSlider->value());
+    lines.push_back("analog_deadzone = rightjoystick, " + deadzonevalue);
+
+    std::ofstream output_file(config_file);
+    for (auto const& line : lines) {
+        output_file << line << '\n';
+    }
+
+    Input::ParseInputConfig(game_id);
```
kalaposfos13: This function call does nothing, you should remove it (it will only be relevant once a game is started up, not while the user is still in the menu)
```suggestion
```

rainmakerv3:
> You're right, for controllers it does make more sense to revert to the builtin default, but I feel like the same is not true for KBM, especially after the recent rollback to the original keybindings. I guess a popup to select which one to do when clicking the Default button might be the best choice here?
>

I think as far as this PR is concerned the behavior of the default button is fine.

As for the behavior of the existing editor’s default button, you may want to consider changing that regardless of what happens with this PR.

The controller portion I think we agree that the defaults should be based on the original defaults, so KBM should probably follow as well. If you want some sort of editable template (its a great feature, I agree), maybe you can add the ability to save/load presets instead. I think the current behavior of that default button will be confusing for a good number of people.

> Also:
> 
> > Unbind a button, meaning it just has no output?
> 
> Yes, exactly that. You could maybe add something like a `-` to represent it in the drop down menu (and you can even save it like that, my parser will throw a warning about syntax errors but will just skip the line, or alternatively you can suppress that warning if the input is that, should be an easy change). It should also be the default option if the binding isn't in the file (which is a very nitpicky edge case, but still), so things aren't randomly bound to cross after a file is saved with your save function.


Ok sounds fine

src/qt_gui/control_settings.cpp
rainmakerv3: Ok, but is there no way for the parsing to be triggered after save is pressed without having to press f8 or whatever? It feels kinda clunky that way

kalaposfos13: Ah right, I forgot that you can open this while ingame, it is fine then like this, my bad

src/qt_gui/control_settings.cpp
rainmakerv3: good point, changed this to remove  all consecutive empty lines instead

rainmakerv3:
Comments addressed, except I am keeping the KBM editor button as is for now. I am unsure as to what is the best approach, so I will leave it up to the reviewer.

As I see it, there are three options where the KBM button can be put:

1) Same as current, in the middle of the UI (can be seen fairly easily, but perhaps does not look as clean)

2) Put in an another tab (would look cleaner, but I am concerned it would be hard to see in a big window where users are generally looking at other UI elements). will also require a fairly large rework of the old text editor UI and related code since that one is not a QT designer UI

3) Have two main menu buttons, one for this GUI and the other for the KBM editor, like this (good visibility, if we're okay with adding another button to the main menu UI):

image

src/qt_gui/control_settings.cpp
kalaposfos13: Actually this is still incorrect, because if the user saves a config that's not the currently opened game's config, then that should not update the loaded keybinds ingame

rainmakerv3: oh yeah, I'll remove it for now

PR #2322, opened 2025-02-01
qt: Add more options to the "Copy info..." section + update en.ts/de.ts

C4ndyF1sh:
This pr adds the ability to copy the game version and the size of a game via the gui individually + adds the buttons to the English and German translation file (with german translation ofc). It also implements an other pr for German translations about game status.

The ability to copy the size and the version was already implemented but only was used when the option "Copy All" was used


C4ndyF1sh:
need to fix something rq

C4ndyF1sh:
Finally fixed the clang error

C4ndyF1sh:
Should be 100% ready to be merged now

PR #2330, opened 2025-02-02
Fix VideoOut events

StevenMiller123:
The data for events with the VideoOut filter is packed in a very specific way, containing an event id, timestamp, and the flip argument. Our prior code only passed along the flip argument, and didn't fill the data bytes accurately. As a result, games checking the event data on VideoOut events weren't seeing the expected data.

This PR adds a new function that fills the data for VideoOut events based on some decompilation work shared by @red-prig. Handling this properly fixes the early hang in The Last of Us Remastered (CUSA00552).

### DIFF
--- a/src/core/libraries/kernel/equeue.cpp
+++ b/src/core/libraries/kernel/equeue.cpp
@@ -84,7 +84,11 @@ bool EqueueInternal::TriggerEvent(u64 ident, s16 filter, void* trigger_data) {
         std::scoped_lock lock{m_mutex};
         for (auto& event : m_events) {
             if (event.event.ident == ident && event.event.filter == filter) {
-                event.Trigger(trigger_data);
+                if (filter == SceKernelEvent::Filter::VideoOut) {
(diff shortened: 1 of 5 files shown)


red-prig:
I will also add that the `event_id` passed to the trigger does not correspond to the `OrbisVideoOutEventId` enumeration, but specifically I know the following constants:

```
 EVENTID_FLIP      0x0006;
 EVENTID_VBLANK    0x0007;
 EVENTID_SETMODE   0x0051;
 EVENTID_POSITION  0x0058;
 EVENTID_PREVBLANK 0x0059;

 //0x0061 sys_update_scaler
 //0x0062 sys_prevblank wZI4fmUJMlw / vXMpe7Murfc
 //0x0063 sys_vblank
```

StevenMiller123:
Not sure if all the checks are fully accurate, but this code works, and now the event identifier should be set correctly on VideoOut events.
This will probably need some more testing before any merges can happen though.

georgemoralis:
looks fine to me

PR #2334, opened 2025-02-02
Change Background Image for games

### LINKED ISSUE #2258: [Feature Request]: Alternate list preview background
### Checklist

### Description

So the current implementation where the emu scrapes pic1.png from the game files and then applies the blur looks kinda weird in some situations, especially with brighter images.
I think that just taking the image and applying a simple 90% opacity full black overlay atop the image or just taking the image and generating a version of it that only has 10% opacity can yield better results.

Current:
Image

Proposed black overlay:
Image

Proposed 10% opacity:
Image

### Reason

The current implementation looks strange, as if there isn't just gaussian blur applied, but also has a low resolution look to it.

### Examples

_No response_

pablodrake:
This pr aims to fix [this](https://github.com/shadps4-emu/shadPS4/issues/2258) issue, by changing the way the background image for games is computed and also adding a checkbox to be able to hide the background image and a slider to change the opacity of the displayed image.

Also while using grid view I came across a bug that would cause the emulator to crash if a cell was selected and the refresh button was pressed, this caused an integer overflow on onCurrentCellChanged, this was easily fixed by some bounds checking.

List view example:
output

Grid view example:
output1

Added base English translations for the new UI elements and Spanish translations aswell.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -673,8 +673,6 @@ void load(const std::filesystem::path& path) {
         checkCompatibilityOnStartup =
             toml::find_or<bool>(general, "checkCompatibilityOnStartup", false);
         chooseHomeTab = toml::find_or<std::string>(general, "chooseHomeTab", "Release");
-        backgroundImageOpacity = toml::find_or<int>(general, "backgroundImageOpacity", 50);
-        showBackgroundImage = toml::find_or<bool>(general, "showBackgroundImage", true);
     }
 
     if (data.contains("Input")) {
@@ -751,6 +749,8 @@ void load(const std::filesystem::path& path) {
         m_recent_files = toml::find_or<std::vector<std::string>>(gui, "recentFiles", {});
         m_table_mode = toml::find_or<int>(gui, "gameTableMode", 0);
         emulator_language = toml::find_or<std::string>(gui, "emulatorLanguage", "en");
+        backgroundImageOpacity = toml::find_or<int>(gui, "backgroundImageOpacity", 50);
+        showBackgroundImage = toml::find_or<bool>(gui, "showBackgroundImage", true);
     }
 
     if (data.contains("Settings")) {
@@ -803,8 +803,6 @@ void save(const std::filesystem::path& path) {
     data["General"]["separateUpdateEnabled"] = separateupdatefolder;
     data["General"]["compatibilityEnabled"] = compatibilityData;
     data["General"]["checkCompatibilityOnStartup"] = checkCompatibilityOnStartup;
-    data["General"]["backgroundImageOpacity"] = backgroundImageOpacity;
-    data["General"]["showBackgroundImage"] = showBackgroundImage;
     data["Input"]["cursorState"] = cursorState;
     data["Input"]["cursorHideTimeout"] = cursorHideTimeout;
     data["Input"]["backButtonBehavior"] = backButtonBehavior;
@@ -843,6 +841,8 @@ void save(const std::filesystem::path& path) {
     data["GUI"]["addonInstallDir"] =
         std::string{fmt::UTF(settings_addon_install_dir.u8string()).data};
     data["GUI"]["emulatorLanguage"] = emulator_language;
+    data["GUI"]["backgroundImageOpacity"] = backgroundImageOpacity;
+    data["GUI"]["showBackgroundImage"] = showBackgroundImage;
     data["Settings"]["consoleLanguage"] = m_language;
 
     std::ofstream file(path, std::ios::binary);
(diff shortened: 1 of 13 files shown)


src/qt_gui/game_grid_frame.cpp
```diff
@@ -136,43 +155,60 @@ void GameGridFrame::PopulateGameGrid(QVector<GameInfo> m_games_search, bool from
...
+    if (!Config::getShowBackgroundImage()) {
+        backgroundImage = QImage();
         RefreshGridBackgroundImage();
+        return;
     }
+
+    const auto& game = (*m_games_shared)[itemID];
+    const int opacity = Config::getBackgroundImageOpacity();
+    const auto cache_path = Common::FS::GetUserPath(Common::FS::PathType::MetaDataDir) /
+                            game.serial / fmt::format("pic1_{}.png", opacity);
```
squidbus: This seems a bit wasteful having to cache a version of the image for every opacity, surely Qt has a way to just display the image with an opacity value?

pablodrake: Yes I only included this because the implementation before used the caching, but I'm down to removing it and just calculating the image everytime, I tested this approach and it was instant in my machine.

As for displaying the image with an opacity value I didn´t find anything that's why I went with the function to change the alpha channel if there is a better approach I would like to implement it.

squidbus: It's fine to cache the image when we get it from the game, because that takes a while. But I don't think we need a separate cached version for every opacity. Qt seems to have this, if it works: https://doc.qt.io/qt-6/qgraphicsopacityeffect.html

pablodrake: The image that was being cached before was the blurred image not the original image from the game, it was just being saved to metadatadir/gameserial/pic1.png but if you were to save the original image there what's the difference between getting the png file from that path or just getting it from the game files?

I will take a look at that Qt class.

squidbus: There was always a delay for me displaying the image the first time, but if that was because of the blur filter then yes should be fine to do without the cache.

pablodrake: That was most likely from saving the blurred image to disk so yes if we remove the caching entirely it should be instant.

Also I read the Qt documentation that you suggested and it seems that it functions by affecting at the rendering opacity of an entire widget and right now the background image is being drawn directly on the Qtable so maybe
it could work with some refactoring and if it is wanted I could do it, but in my machine with current custom function that just changes the alpha level it is instant.

squidbus: As long as it loads quick and doesn't need to save a separate png file for each opacity level, it's fine with me.

pablodrake: output

With no png saving whatsoever it is instant for me, even if it is the first time that the game is loaded.
I will commit the changes if this is acceptable.

squidbus: Would be nice if it could update the background automatically when you save the settings, but overall looks good to me.

pablodrake: Ok I also though about this, could maybe take a look at it tomorrow don´t merge this pr yet then.

pablodrake:
List view:
output

Grid view:
output1


Got it to work, I will commit the changes now, the time between selecting different games is the same as main branch without this change, at least on my machine, so this is not the bottleneck for that.

Also added some checks to make sure we only calculate the new background image when necessary.

pablodrake:
There was an unnecessary include so I amended the commit.

pablodrake:
Ok I found a way to crash the emulator if you hit the refresh button and then try to change the opacity.

I will take a look do not merge the pr.

pablodrake:
All seems fine in my machine now, maybe someone else could test it aswell.

Sorry for that, some leftover code so I amended.

src/common/config.cpp
```diff
@@ -783,6 +803,8 @@ void save(const std::filesystem::path& path) {
     data["General"]["separateUpdateEnabled"] = separateupdatefolder;
     data["General"]["compatibilityEnabled"] = compatibilityData;
     data["General"]["checkCompatibilityOnStartup"] = checkCompatibilityOnStartup;
+    data["General"]["backgroundImageOpacity"] = backgroundImageOpacity;
+    data["General"]["showBackgroundImage"] = showBackgroundImage;
```
squidbus: Think these should be under `GUI` settings category.

pablodrake: Yes my bad I got confused because I saw the BGM settings there and also the compatibility checks that control the GUI are in General maybe that needs changing aswell but not in this pr I think.

squidbus:
Gave it a test, looks good to me.

PR #2335, opened 2025-02-03
Reduce redundant calls to sceKernelIsNeoMode, improve BB frame times

hspir404:
I found this call to `sceKernelIsNeoMode` while poking around in the profiler. I saw that calls are very frequent (especially inside the conversion of texture descriptors to the Vulkan renderer). They are taking up some 750+ ms out of a 10 second sample. I sampled in several different places, and these calls were often only taking up as much as 400ms, depending on the scene being rendered, but still a not small chunk.

This change makes the code very slightly less appealing, but it's reducing 6040 calls to 65 (at least, those are the ones the sample-based profiler caught in a 10 second sample). Specifically, that sample was taken at the top of the DLC Nightmare Catherdral, notably the worst performing part of the game I've been able to find.

Before:
image

After:
image

I made an automation build here and tested both EXEs side by side in several parts of the game, and have screenshots from those tests. I loaded the emulator and loaded the save game, straight into these areas, and did not move the camera. I let the frame rate settle before taking screenshots, and attempted to capture the screenshot that got the highest frame rate I was seeing. If I missed the screenshot, I let it settle a bit again before taking another one.

Test Configuration -

Game Mods:
- Vertex explosions fix (modifies base game files)
- Skip intro
- Disable motion blur
- Disable chromatic aberration
- Disable HTTP calls
- 60fps patch
- vsync off
- vblank divider 2

Box:
- Windows 10
- 64GB RAM
- nVidia RTX 3080 10GB VRAM
- AMD 5950x (16-core processor)

In almost every scene there is a slight performance improvement.

Some other folks in chat also did some A/B testing, took some screenshots, and were getting a bit of a performance increase.

Before:
1-before

After:
1-after

Before:
2-before

After:
2-after

Before:
3-before

After:
3-after

Before:
4-before

After
4-after

Before:
5-before

After:
5-after

Before:
6-before

After:
6-after

Before:
7-before

After:
7-after

Before:
8-before

After:
8-after

### DIFF
--- a/src/core/libraries/kernel/process.cpp
+++ b/src/core/libraries/kernel/process.cpp
@@ -13,7 +13,6 @@
 namespace Libraries::Kernel {
 
 int PS4_SYSV_ABI sceKernelIsNeoMode() {
-    LOG_DEBUG(Kernel_Sce, "called");
     return Config::isNeoModeConsole() &&
            Common::ElfInfo::Instance().GetPSFAttributes().support_neo_mode;
 }


squidbus:
Looking at this it seems we now have to pass this flag around a lot, and will probably increase more in the future as more neo-mode differences are implemented. I would prefer to see one of the following:
* Optimize `sceKernelIsNeoMode` itself; could help even more as it is also spammed by games. Although I'm not sure why exactly it would take that much time to begin with as it basically is just two boolean checks. Maybe remove the debug log?
* Set a flag for neo mode in the `Liverpool` instance on creation, and use that around the rendering code instead of requiring passing around this boolean.

Randomuser8219:
Will this change anything with log filter set to Debug?

hspir404:
> Will this change anything with log filter set to Debug?

If you enable `*:Debug` on nightly, you get about 0.5fps. Unplayable.
If you enable `*:Debug` on this PR, you get just about full perf (with a lot of log spam).

> * Maybe remove the debug log?

I pushed another branch in my fork to test this out. Seems to do the trick. The performance is the same as this PR:
https://github.com/hspir404/shadPS4/actions/runs/13107421178

I also profiled after commenting out the log, and `sceKernelIsNeoMode` no longer shows up on the profiler. I looked at the disassembly, and the function call gets inlined and it's just a simple boolean check.

Sounds like deleting that log line would make sense, or some of the other suggestions Squidbus has.

squidbus:
> Sounds like deleting that log line would make sense, or some of the other suggestions Squidbus has.

Removing the log line here should be fine, or if you want to be safe downgrade it to `LOG_TRACE` and then it gets completely nulled out on non-debug builds.

hspir404:
> Removing the log line here should be fine

done

PR #2342, opened 2025-02-04
vk_instance: Clean up extension management.

squidbus:
* Remove some unused extensions.
* Remove some unused properties, features structures, and extension flags.
* Remove use of dynamic color write enable, as there is no functional difference from setting dynamic color write mask to 0 which we also do.
* Misc organization improvements for extension and feature managment code.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
+++ b/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
@@ -131,8 +131,7 @@ GraphicsPipeline::GraphicsPipeline(
         vk::DynamicState::eStencilOpEXT,
     };
 
-    if (instance.IsColorWriteEnableSupported()) {
-        dynamic_states.push_back(vk::DynamicState::eColorWriteEnableEXT);
+    if (instance.IsDynamicColorWriteMaskSupported()) {
         dynamic_states.push_back(vk::DynamicState::eColorWriteMaskEXT);
     }
     if (instance.IsVertexInputDynamicState()) {
@@ -241,7 +240,7 @@ GraphicsPipeline::GraphicsPipeline(
                                 ? LiverpoolToVK::BlendOp(control.alpha_func)
                                 : color_blend,
             .colorWriteMask =
-                instance.IsColorWriteEnableSupported()
+                instance.IsDynamicColorWriteMaskSupported()
                     ? vk::ColorComponentFlagBits::eR | vk::ColorComponentFlagBits::eG |
                           vk::ColorComponentFlagBits::eB | vk::ColorComponentFlagBits::eA
                     : key.write_masks[i],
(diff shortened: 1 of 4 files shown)


hgh32:
Crashes immediately on Intel GPUs, not sure if hardware specific

squidbus:
> Crashes immediately on Intel GPUs, not sure if hardware specific

Need crash details.

Missake212:
Tested on my Intel iGPU in Bloodborne and instantly crashed too, here's the log let me know if you need anything else.

[shad_log.txt](https://github.com/user-attachments/files/18699631/shad_log.txt)

squidbus:
> Tested on my Intel iGPU in Bloodborne and instantly crashed too, here's the log let me know if you need anything else.

I need the crash trace from debugger.

Missake212:
Hope this is what is needed

image

diegolix29:
also on my end 
image

squidbus:
Are there any validation errors?

voguelike:
this pr regressed the fix from pr [#1282](https://github.com/shadps4-emu/shadPS4/pull/1282) by removing the check for 
`if (Config::vkValidationEnabled() || Config::isRdocEnabled()) `
before loading 
`tooling_info = add_extension(VK_EXT_TOOLING_INFO_EXTENSION_NAME);`

squidbus:
> this pr regressed the fix from pr [#1282](https://github.com/shadps4-emu/shadPS4/pull/1282)

This seems like something you should report to Reshade, I would prefer to know in logs if tools are attached other than just validation and RenderDoc (and even with the latter it’s only reported if enabled in our config).

voguelike:
would it be possible to instead have a condition that checks for any kind of tooling that isn't used in regular play? maybe implementing a flag that toggles on when any kind of tooling is attached to allow for a simple condition. 

this hang affects a bunch of non-technical users which aren't using any tooling so just discriminating between regular use vs developer usage would let us workaround this without any issues.

also it would be great if contacting reshade to have them address this was actually feasible...

squidbus:
> would it be possible to instead have a condition that checks for any kind of tooling that isn't used in regular play? maybe implementing a flag that toggles on when any kind of tooling is attached to allow for a simple condition.

I’m not sure what exactly you have in mind here, when the extension you want to disable *is* the one that allows us to detect if any tooling is attached.

squidbus:
> also it would be great if contacting reshade to have them address this was actually feasible...

They appear to have a forum: https://reshade.me/forum

rainmakerv3:
how about I add a new debug setting to load this extension or not? the vast majority of regular users will not benefit from it, and it does cause crashes

squidbus:
> how about I add a new debug setting to load this extension or not? the vast majority of regular users will not benefit from it, and it does cause crashes

At that point you might as well remove it because there will be no point in detecting attached tools if its disabled by default.

Issue #2340, opened 2025-02-03
[APP BUG]: Controller touchpad not responding properly


morten71:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

After commit [c4bfaa6](https://github.com/shadps4-emu/shadPS4/commit/c4bfaa60312c6e224c26eca9a00b7b573253aa42) with the new input mapping system, my controller touchpad stopped responding in Bloodborne, ie. clicking on it now did nothing, either on the left side or the right side.

This happens both when I build the main branch myself, and when I use the latest AppImage (I'm using linux).

After messing around with the code a bit, and inserting some debug output, it looks like the problem is somewhere around this:

In the WaitEvent function in [src/sdl_window.cpp line 304](https://github.com/shadps4-emu/shadPS4/blob/8ad650582a947fecd4308d29d83955de9be9f11a/src/sdl_window.cpp#L304l) for every SDL event of the types `SDL_EVENT_GAMEPAD_BUTTON_DOWN`, `SDL_EVENT_GAMEPAD_BUTTON_UP`, or `SDL_EVENT_GAMEPAD_AXIS_MOTION`, 23 (or sometimes 24) new State objects get added to the controller's state object ring buffer. I believe that is only supposed to be one state object for each of those events?

The 23 state objects come from the iteration over `output_array` in [src/input/input_handler.cpp line 669](https://github.com/shadps4-emu/shadPS4/blob/8ad650582a947fecd4308d29d83955de9be9f11a/src/input/input_handler.cpp#L669). It results in 17 (or sometimes 18) calls to `GameController::CheckButton`, and 6 calls to `GameController::Axis`, both of which add a state object.

Since my PS4 controller triggers constant `SDL_EVENT_GAMEPAD_AXIS_MOTION` events, with small variations inside the dead zones, that ends up being a huge number of state objects, with the ring buffer being entirely taken up by just 3 of those SDL events.

If I add these lines to the beginning of the WaitEvent function at [line 315 in src/sdl_window.cpp](https://github.com/shadps4-emu/shadPS4/blob/8ad650582a947fecd4308d29d83955de9be9f11a/src/sdl_window.cpp#L315) just before the switch block:

```
if (event.type == SDL_EVENT_GAMEPAD_AXIS_MOTION) {
[...]

PR #2346, opened 2025-02-04
Fix issues with input detection due to overloaded ring buffer and fix buffer size being incorrect

kalaposfos13:
Adds an axis noise filter and only updates outputs if a change happened to that output, meaning that there are now both less axis movement event spam, and now every input event only updates one output, instead of updating all of them with all but the updated one holding duplicate data from the previous entry, which was making the ring buffer holding input events being overloaded in two different ways. These features were once in my main remap PR, but I had to revert them, although it looks like whatever was the issue that led to that no longer applies, as from my testing, I couldn't find any significant regressions, but just to be safe, I'll open this as a draft and wait until it's tested on more games.


sam20112011:
This PR fix this issue for me: https://github.com/shadps4-emu/shadPS4/issues/2340

Hermiten:
Does this fix input lag visible on these games https://github.com/shadps4-emu/shadps4-game-compatibility/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22input%20lag%22  ?

ElBread3:
Tested Minecraft, this PR appears to have gotten rid of menu input lag, but not any game input lag.

kalaposfos13:
I think I found a possible fix for Minecraft's (and maybe for other games') input lag, which is reducing the buffer size to 32, which is the number that's seen in Ghidra. It can be tested here: https://github.com/kalaposfos13/shadPS4/tree/mc-input-lag-hack

kalaposfos13:
Since this is a draft, and the ring buffer size change is at least partially related to this PR anyway, I figured I might as well add the Minecraft fix to here.

Hermiten:
I think you can and if there is a big problem, you will split it
Now we need some tests on games with and without this problem to be sure everything is good

PR #2348, opened 2025-02-04
Add outer deadzone config

kalaposfos13:
This currently is only editable from the original text  editor
Syntax: analog_deadzone = device, inner_deadzone, outer_deadzone
What was just simply deadzone before, is now called inner deadzone, to differentiate the two types
Both values still go from 1 to 127

### DIFF
--- a/src/qt_gui/control_settings.cpp
+++ b/src/qt_gui/control_settings.cpp
@@ -220,10 +220,10 @@ void ControlSettings::SaveControllerConfig(bool CloseOnSave) {
     lines.push_back("# Range of deadzones: 1 (almost none) to 127 (max)");
 
     std::string deadzonevalue = std::to_string(ui->LeftDeadzoneSlider->value());
-    lines.push_back("analog_deadzone = leftjoystick, " + deadzonevalue);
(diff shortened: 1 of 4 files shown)


src/qt_gui/control_settings.cpp
```diff
@@ -220,10 +220,10 @@ void ControlSettings::SaveControllerConfig(bool CloseOnSave) {
     lines.push_back("# Range of deadzones: 1 (almost none) to 127 (max)");
 
     std::string deadzonevalue = std::to_string(ui->LeftDeadzoneSlider->value());
-    lines.push_back("analog_deadzone = leftjoystick, " + deadzonevalue);
+    lines.push_back("analog_deadzone = leftjoystick, " + deadzonevalue + "127");
 
     deadzonevalue = std::to_string(ui->RightDeadzoneSlider->value());
-    lines.push_back("analog_deadzone = rightjoystick, " + deadzonevalue);
+    lines.push_back("analog_deadzone = rightjoystick, " + deadzonevalue + "127");
```
rainmakerv3: needs another ", "

PR #2349, opened 2025-02-04
Auto update of english translation file based on sources

georgemoralis:
This add auto - update of en.ts file based an latest changes on UI.

It creates a pull request with the update file every 12a.m or in manual run.


ElBread3:
Could theoretically spam PRs, so this would need to not run if the previous PR hasn't been merged yet and if there hasn't been any changes since last update. And the removing-copyright thing would need looked at too. I hope this leads into integrating Crowdin or a similar service, would be amazing!

georgemoralis:
yah crowdin is planned

georgemoralis:
@DanielSvoboda had any time to check this?

DanielSvoboda:
> Could theoretically spam PRs, so this would need to not run if the previous PR hasn't been merged yet and if there hasn't been any changes since last update. And the removing-copyright thing would need looked at too. I hope this leads into integrating Crowdin or a similar service, would be amazing!

I made some changes that are in https://github.com/shadps4-emu/shadPS4/pull/2392 , where the reuse text is added, instead of line 4 I put it on line 2, it is the maximum above that can be inserted. If you want it can be put back on line 4. It will also fix all the language files, not just en.ts. If there is another execution of the BOT it uses the same PR to make the changes, it does not spam PRs. But I think this should be done once a week, not once a day.

bot using own PR:
image

GHU7924:
Could you briefly explain for those who are not well versed in how these PR work (2349 and 2392)?

en.ts be created from scratch or will existing ones be expanded? The rest of the translations will be based on the updated en.ts?
I'm asking because at the moment translations differ in the number of lines and there may be such moments :
In ru_Ru.ts, lines 57-64 differ from the same lines in en.ts (as if they don't exist or are located elsewhere).

Personally, I would like all translations to be identical, of course, with the exception of the text in `<translation> </translation>`, but so that the number of lines matches and there are no duplicates in the lines, and all the elements are available for translation. 
This way, it will be clear that everything is in order, everything is in place and nothing is missing anywhere. 
en.ts should be as complete as possible.

DanielSvoboda:
> Could you briefly explain for those who are not well versed in how these PR work (2349 and 2392)?
> 
> en.ts be created from scratch or will existing ones be expanded? The rest of the translations will be based on the updated en.ts? I'm asking because at the moment translations differ in the number of lines and there may be such moments : In ru_Ru.ts, lines 57-64 differ from the same lines in en.ts (as if they don't exist or are located elsewhere).
> 
> Personally, I would like all translations to be identical, of course, with the exception of the text in `<translation> </translation>`, but so that the number of lines matches and there are no duplicates in the lines, and all the elements are available for translation. This way, it will be clear that everything is in order, everything is in place and nothing is missing anywhere. en.ts should be as complete as possible.

The translation itself will not be changed nor based on en.ts.
The comparison will be made from the source code, for example:
if a text is in the .tr translation file but is not being used in the emulator, 'vanished' will appear, see the print.
image
If the emulator has a text and it is not in the translation file, it is added as 'undefinished', see the print.

<br><br>
image
<br><br>
And several other types of corrections and standardizations are made.
The number of lines may vary because of 'vanished', it is possible to put a code to exclude these, since they are no longer used.

DanielSvoboda:
I applied 'no-obsolete' (remove o 'vanished'), so all files will have the same number of lines, and in the same order.

GHU7924:
@DanielSvoboda Thanks for the explanation.

PR #2360, opened 2025-02-05
emulator: crash faster

Osyotr:
By disabling Windows Error Reporting.

If a game triggers some assert then there's ~10s delay before emulator is closed. With this PR it's down to ~2s.


gandalfthewhite19890404:
Can we set it in config? By default = Windows error reporting set to disabled

DanielSvoboda:
> Can we set it in config? By default = Windows error reporting set to disabled

Why would you want this enabled? In my opinion this should definitely be disabled.

gandalfthewhite19890404:
> > Can we set it in config? By default = Windows error reporting set to disabled
> 
> Why would you want this enabled? In my opinion this should definitely be disabled.

I'm not against it, backwards.
By default - Windows error reporting system set to disabled. But user can change it in config.

georgemoralis:
Make it configurable in config (but not in gui)

georgemoralis:
no

Osyotr:
But you didn't explain how's WER helpful for debugging and why would shadPS4 want to notify Microsoft about every crash in emulated games?

kalaposfos13 (approved):

src/emulator.cpp
```diff
@@ -58,10 +58,11 @@ Frontend::WindowSDL* g_window = nullptr;
 namespace Core {
 
 Emulator::Emulator() {
-    // Initialize NT API functions and set high priority
+    // Initialize NT API functions, set high priority and disable WER
 #ifdef _WIN32
     Common::NtApi::Initialize();
     SetPriorityClass(GetCurrentProcess(), ABOVE_NORMAL_PRIORITY_CLASS);
+    SetErrorMode(SetErrorMode(0) | SEM_NOGPFAULTERRORBOX);
```
Niram7777, line 65: SetErrorMode(GetErrorMode() | SEM_NOGPFAULTERRORBOX);

I dont see a reason for calling SetErrorMode(0)

Niram7777 (commented):
Microsoft recommends to use SetThreadErrorMode/GetThreadErrorMode instead

src/emulator.cpp
Osyotr: https://devblogs.microsoft.com/oldnewthing/20040727-00/?p=38323

The article dated a bit since there *is* `GetErrorMode` but whatever.
The intent was to set error mode for the entire process, so using `SetThreadErrorMode` is not a good recommendation.

PR #2363, opened 2025-02-06
shader_recompiler: Replace texel buffers with in-shader buffer format interpretation

squidbus:
Replaces the current implementation of formatted buffer loads and stores based on texel buffers with one that performs the format interpretation manually in-shader, lowering it to plain buffer access. This may lower performance, but should help get around many limitations faced with texel buffers, such as the assert on buffer stride not equaling format stride, or unsupported texel buffer formats on drivers like MoltenVK.

The basic implementation strategy here is to define several pack/unpack instructions for different format types on the back-end side, and utilize those in IR passes to perform simple lowering. Only some of these pack/unpack instructions have native equivalents in SPIR-V, but other graphics APIs may be able to benefit in the future if supported (I know MSL has a native 2_10_10_10 pack/unpack function, for example).

This currently implements support for all formats that were both supported before via native Vulkan formats and marked as buffer readable/writable in the GPU docs. If there are any other formats actually possible to encounter with buffers, they can be implemented as needed.

Needs a lot of testing for any regressions.


Hermiten:
What kind of games it can be useful to test ? And do you have a protocol or something to inspect difference ? (And no unrelated things)

squidbus:
> What kind of games it can be useful to test ?

Anything is useful for regression testing since it's a wide-reaching change, but in particular anything that hit texel format assert, or R32G32B32 format error on macOS, may be improved.

shinra-electric:
on macOS 

**Sekiro**
Main: Crash on boot
PR: Intro cutscene

**Armored Core 6**
Main: Crash after splash screens
PR: Menus

chronicer:
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:494 CompileModule: Compiling fs shader 0xe88da41da59d9ad3
[Debug] <Critical> liverpool.h:1517 unhandled_exception: Unreachable code!
Unhandled exception: invalid unordered_map<K, T> key

Not fixes load low-quality models for weapons and crash when i tries to load character model in CUSA00365 Dynasty Warriors 8 Xtreme Legends(((

StevenMiller123:
Fixes the `[Debug] <Critical> vk_rasterizer.cpp:652 operator(): Assertion Failed!` asserts on Intel GPUs, bringing Dragon Ball Xenoverse, Dragon Ball Xenoverse 2 (v1.00), and Minecraft Playstation 4 Edition (v1.00) ingame for users of these GPUs.
Doesn't appear to cause any regressions.
image
image
image

C4ndyF1sh:
Improvements through this PR on Mac that I noticed:
- "Cuphead" goes to menu now
-  "Agartha Christie - The ABC Murders" now goes past the cutscenes and goes ingame
- "Minecraft" now goes past the menu and goes ingame

Didn't saw any regressions in all of my dumped games, performance seems to be also the same

src/shader_recompiler/backend/spirv/emit_spirv_bitwise_conversion.cpp
```diff
@@ -76,30 +60,311 @@ Id EmitUnpackSnorm2x16(EmitContext& ctx, Id value) {
...
+    const auto y{ctx.OpBitFieldSExtract(ctx.U32[1], value, ctx.ConstU32(8U), ctx.ConstU32(8U))};
+    const auto z{ctx.OpBitFieldSExtract(ctx.U32[1], value, ctx.ConstU32(16U), ctx.ConstU32(8U))};
+    const auto w{ctx.OpBitFieldSExtract(ctx.U32[1], value, ctx.ConstU32(24U), ctx.ConstU32(8U))};
+
+    const auto unpacked{ctx.OpCompositeConstruct(ctx.U32[4], x, y, z, w)};
+    return ctx.OpBitcast(ctx.F32[4], unpacked);
+}
+
+Id Float32ToUfloatM5(EmitContext& ctx, Id value, u32 mantissa_bits) {
+    const auto raw_value{ctx.OpBitcast(ctx.U32[1], value)};
```
raphaelthegreat: I think this should be defined as a spirv function in emit context and called here with OpFunctionCall. Should make shaders that use it much cleaner to read. Also would be a good idea to link some references on the implementation for future reference. Personally found this that mesa uses, might be good to double check https://github.com/chaotic-cx/mesa-mirror/blob/21cbe000949ad0a9cd1268088b0e1a5d8b1eeed4/src/util/format_r11g11b10f.h#L274

src/shader_recompiler/backend/spirv/emit_spirv_bitwise_conversion.cpp
```diff
@@ -76,30 +60,311 @@ Id EmitUnpackSnorm2x16(EmitContext& ctx, Id value) {
...
+        ctx.OpFAdd(ctx.F32[1],
+                   ctx.OpSelect(ctx.F32[1], norm_overflow,
+                                ctx.OpFAdd(ctx.F32[1], exponent, ctx.ConstF32(1.f)), exponent),
+                   ctx.ConstF32(15.f)))};
+    const auto norm{ctx.OpBitFieldInsert(ctx.U32[1], norm_final_mantissa, norm_final_exponent,
+                                         ctx.ConstU32(mantissa_bits), ctx.ConstU32(5U))};
+
+    return ctx.OpSelect(
+        ctx.U32[1], is_zero, ctx.ConstU32(0U),
+        ctx.OpSelect(ctx.U32[1], is_nan, ctx.ConstU32(31u << mantissa_bits | 1U),
```
raphaelthegreat: More nit but given the big indentation going on here it might good idea to break it into 2 variables

src/shader_recompiler/backend/spirv/emit_spirv_bitwise_conversion.cpp
```diff
@@ -76,30 +60,311 @@ Id EmitUnpackSnorm2x16(EmitContext& ctx, Id value) {
...
+        ctx.OpShiftRightLogical(ctx.U32[1], denorm_mantissa, ctx.ConstU32(mantissa_bits)),
+        ctx.ConstU32(0U))};
+    const auto denorm{ctx.OpSelect(ctx.U32[1], denorm_overflow, ctx.ConstU32(1U << mantissa_bits),
+                                   denorm_mantissa)};
+
+    const auto norm_mantissa{ctx.OpConvertFToU(
+        ctx.U32[1],
+        ctx.OpRoundEven(
+            ctx.F32[1],
+            ctx.OpFMul(
```
raphaelthegreat: x * 2^y can be done with ldexp()

src/shader_recompiler/backend/spirv/emit_spirv_bitwise_conversion.cpp
```diff
@@ -76,30 +60,311 @@ Id EmitUnpackSnorm2x16(EmitContext& ctx, Id value) {
...
+    return ctx.OpBitcast(ctx.F32[4], unpacked);
+}
+
+Id EmitPackSint4x8(EmitContext& ctx, Id value) {
+    return EmitPackUint4x8(ctx, value);
+}
+
+Id EmitUnpackSint4x8(EmitContext& ctx, Id value) {
+    // No SPIR-V instruction for this, do it manually.
+    const auto x{ctx.OpBitFieldSExtract(ctx.U32[1], value, ctx.ConstU32(0U), ctx.ConstU32(8U))};
```
raphaelthegreat: This pattern of bitfield extract at start of Unpack* instruction is common, maybe it can be separated into a function. I came up with small example can be adapted easily into spirv
```cpp
struct R {R(int a,int b) : offset(a), size(b){}int offset; int size;};
template <bool is_signed, typename... Args>
std::array<int, 4> Extract(int num, Args... args) {
    std::array<int, 4> result{};
    int i = 0;
    ([&]
    {
        result[i++] = (num >> args.offset) & ((1 << args.size) - 1);
    } (), ...);
    return result;
}

int main()
{
const auto [x, y, z, w] = Extract<true>(0x01020304, R(0, 8), R(8, 8), R(16, 8), R(24, 8));
std::printf("%d %d %d %d\n", x,y,z,w);
}
```
I think it could also be done for the BitFieldInsert calls too and save some code

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -469,25 +469,12 @@ bool Rasterizer::BindResources(const Pipeline* pipeline) {
...
         if (!meta_read) {
             for (const auto& desc : info.buffers) {
                 if (IsMetaUpdate(desc)) {
                     return false;
                 }
             }
-
-            for (const auto& desc : info.texture_buffers) {
-                if (IsMetaUpdate(desc)) {
-                    return false;
```
raphaelthegreat, line 488: The lambda IsMetaUpdate is not needed anymore and can be inlined into the loop for simplicity sake

src/shader_recompiler/backend/spirv/emit_spirv_bitwise_conversion.cpp
squidbus: Oh nice, for some reason I thought that wasn't in SPIR-V.

src/shader_recompiler/backend/spirv/emit_spirv_bitwise_conversion.cpp
squidbus: I already referenced the Mesa conversion functions, I'll look into making this a function.

src/shader_recompiler/backend/spirv/emit_spirv_bitwise_conversion.cpp
squidbus: I don't think there's a very good way to break this up, it's basically just an `OpSelect` chain checking the different result conditions.

hgh32:
f0afcfae1b0a97c9bc489306cd00c5082dcf5a1e Crashes immediately after running

squidbus:
> [f0afcfa](https://github.com/shadps4-emu/shadPS4/commit/f0afcfae1b0a97c9bc489306cd00c5082dcf5a1e) Crashes immediately after running

It is rebased on the PR that broke Intel, so yes it would do that.

squidbus:
Review comments addressed, issue with other PR breaking Intel is fixed.

raphaelthegreat (approved):
LGTM, feel free to merge once its tested to your satisfaction

hgh32:
Fixed alignment assertions for some games, still has graphical issues

Issue #2364, opened 2025-02-06
[APP BUG]: Network Error on checking updates (Nightly 0d498f1)


cvertee:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

![Image](https://github.com/user-attachments/assets/06225507-ce64-4997-8b3e-a52ba7f0ad91)

API in browser returns this
```json
{"message":"API rate limit exceeded for <IP>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

### Reproduction Steps

Open Nightly 0d498f1 version, try checking for updates

### Expected Behavior

Maybe display message in error dialog box? 

### Specify OS Version

Windows 10

DanielSvoboda:
This happens if you make more than 60 requests in less than 1 hour. I believe that if autoUpdate is turned on at startup, every time you open the emulator it makes a query. What could be done is a different message box for these cases.

DanielSvoboda:
I think it should be better to understand this way
![Image](https://github.com/user-attachments/assets/ff2f34b6-a28c-4b54-8756-ea0266a1d890)
I'm going to do a PR...

cvertee:
Thanks! But still it's weird, i did not make 60 requests on this day, opened emulator maybe like 5-6 times..

DanielSvoboda:
Do you use this game compatibility option? It should make requests too, this should add up to the total number of requests, not just the autoUpdater. I don't know how it's done, but maybe it makes more than one request when entering...
![Image](https://github.com/user-attachments/assets/acac9f74-08b2-4388-b77b-bfb00320aeb1)

DanielSvoboda:
According to my calculations, 'Game Compatibility' currently makes 24 requests to get the information, this will increase over time, it could be a future problem... one solution would be to process this information on a website, and the emulator get the information already summarized with one request.

cvertee:
Yeah, i have it enabled. Disabled it for now, didn't know about so much requests.
Also idk maybe it's case for another issue, when i disable Display Compatibility Data checkbox it can randomly crash, after 3 crashes finally disabled it

DanielSvoboda:
[...]

PR #2365, opened 2025-02-06
QT: AutoUpdate - improvement message limit exceeded

DanielSvoboda:
issues: https://github.com/shadps4-emu/shadPS4/issues/2364

| Before | After |
|-------------|-------------|
image | image


DanielSvoboda:
I'm making another PR to reduce the requests made for 'game compatibility', which ends up affecting the autoUpdate requests.

PR #2366, opened 2025-02-07
vulkan: Use more supported 4444 format.

squidbus:
Use more supported 4444 format, as the one currently used is not supported on some Intel GPUs.

Someone with Intel GPU please check if this fixes the recent crash on boot.


squidbus:
Seems Intel GPU does indeed work again.

PR #2367, opened 2025-02-07
Game-compatibility - improved

DanielSvoboda:
Let's go...
In short, it improves file loading at startup and makes fewer requests.

If `Display Compatibility Data` is checked, it will load the information already downloaded locally every time you start the emulator. And when you click on it, in addition to adding the columns, it will fill in if you have the file. Because before it would only load if the `Update Compatibility Database On Startup` option was checked.

And speaking of updating, it made more than 20 requests to GitHub to get this data, because it is only possible to request 100 items per page and we currently have more than 2000 issues. https://github.com/shadps4-emu/shadPS4/issues/2364
And this excess of requests ended up affecting the autoUpdate, because there is a limit of 60 requests per hour that GitHub allows without authentication. To solve this I made [this other PR](https://github.com/shadps4-emu/shadps4-game-compatibility/pull/2409), but in the `shadps4-game-compatibility` repository, where the `github action` itself creates the `compatibility_data.json` and no longer the emulator, updating it `every 3 hours`, creating the file in a 'release' and deleting the old ones. this way the emulator will only need to make a request to update the compatibility data.

This PR is in draft status for the following reasons:

- [X] [shadps4-game-compatibility/pull/2409](https://github.com/shadps4-emu/shadps4-game-compatibility/pull/2409) needs to be approved;
- [X] Create token with access to 'repo' `SHADPS4_TOKEN_COMPATIBILITY` in repository 'shadps4-game-compatibility';
- [X] Press 'Run workflow' once to create the .json file;
- [x]  Fix the problem that for some reason sometimes creates a release as a draft
- [x] Changing the repository link in this PR as it is being tested with my fork
- [x] And it's also good for someone to check if everything is working correctly 

*I also removed version_number, I didn't find it necessary...

### DIFF
--- a/src/qt_gui/compatibility_info.cpp
+++ b/src/qt_gui/compatibility_info.cpp
@@ -34,7 +34,7 @@ void CompatibilityInfoClass::UpdateCompatibilityDatabase(QWidget* parent, bool f
     if (!WaitForReply(reply)) {
         reply->deleteLater();
         QMessageBox::critical(parent, tr("Error"),
-                              tr("Tempo limite ao baixar os dados de compatibilidade."));
+                              tr("Timeout while downloading compatibility data."));
         return;
     }
 
--- a/src/qt_gui/compatibility_info.cpp
+++ b/src/qt_gui/compatibility_info.cpp
@@ -19,109 +19,48 @@ CompatibilityInfoClass::CompatibilityInfoClass()
 CompatibilityInfoClass::~CompatibilityInfoClass() = default;
 
 void CompatibilityInfoClass::UpdateCompatibilityDatabase(QWidget* parent, bool forced) {
-    if (!forced)
-        if (LoadCompatibilityFile())
-            return;
-
-    QNetworkReply* reply = FetchPage(1);
-    if (!WaitForReply(reply))
+    if (!forced && LoadCompatibilityFile())
         return;
+
+    QUrl url("https://github.com/DanielSvoboda/shadps4-game-compatibility/releases/latest/download/"
+             "compatibility_data.json");
+    QNetworkRequest request(url);
+    QNetworkReply* reply = m_network_manager->get(request);
 
     QProgressDialog dialog(tr("Fetching compatibility data, please wait"), tr("Cancel"), 0, 0,
                            parent);
     dialog.setWindowTitle(tr("Loading..."));
 
-    int remaining_pages = 0;
-    if (reply->hasRawHeader("link")) {
-        QRegularExpression last_page_re("(\\d+)(?=>; rel=\"last\")");
-        QRegularExpressionMatch last_page_match =
-            last_page_re.match(QString(reply->rawHeader("link")));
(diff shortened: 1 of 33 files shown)


Hermiten:
The PR is approved on the Compat list. Do we need to wait you to update the repository link to do the test ?
Thanks for this update because it will help a lot to reduce requests from user side and I think it will be faster like this. 

Do you think it can be good to save a cache of the result so if it's fail to retrieve , it will not pu "unknow" for every games but the last status on cache ?

GHU7924:
I don't see the point in frequently updating the compatibility database for the user, especially now that the emulator is in an early stage of development.
Therefore, I suggest doing so if possible.

GHU7924:
@DanielSvoboda I don't know how to describe it correctly, but in my opinion, statuses are reset after a certain period of time. I've updated the compatibility database, and the statuses have been applied. I restarted the emulator about 10 times and the statuses were saved, but if you don't run the emulator for a while (I don't know for sure if it's half an hour or an hour, maybe), then after running the emulator all the statuses will be unknown, updating the compatibility database again, everything repeats, the statuses are there, then the statuses are unknown.

This happens when the check mark "Update Compatibility Database On Startup" is removed.

DanielSvoboda:
@Hermiten yes, now I have updated, you can test it if you want.
@GHU7924 I couldn't repeat what you mentioned
The PR is practically ready, but first I'm going to do some tests to see if everything is ok.

DanielSvoboda:
> I don't see the point in frequently updating the compatibility database for the user, especially now that the emulator is in an early stage of development. Therefore, I suggest doing so if possible.

It seems like a good option, but I'm not going to do it in this PR, as it will affect the interface, so maybe another day, when I'm messing with the interface I can change it, or someone else.

GHU7924:
>@GHU7924 I couldn't repeat what you mentioned

Unfortunately, I haven't fully understood how this happens myself, so I can't 100% tell you that these are the right steps to reproduce. I didn't keep track of the time. For this reason, I cannot say how long it will take before this happens. Maybe you'll test it again. 

Perhaps you should download your build from GitHub, unpack, run, specify the games, check the box "Display Compatibility Data" and press "Update Compatibility Database On Startup", the statuses should apply, then restart the emulator, make sure that the statuses are still there, then close the emulator again. After about 40 minutes, run this build again and see if the statuses have been saved or if they have become unknown.

Maybe it's something else, but I can't say for sure. I've had this happen 3 times.

>It seems like a good option, but I'm not going to do it in this PR, as it will affect the interface, so maybe another day, when I'm messing with the interface I can change it, or someone else.

I'm just offering ideas that, in my opinion, will make the emulator more convenient and user-friendly, or provide more features to the user. When and who will be involved in the implementation does not depend on me (if anyone will take it at all). I'm not rushing anyone or insisting. If you approved the idea and decided that you would take over the implementation, then for me it means that the idea was really good.

PR #2377, opened 2025-02-08
Fix duplicated translations

Martini-141:


Martini-141:
Should be good now, removed some duplicates and fixed warning I got when compiling

Martini-141:
@DanielSvoboda It seems like the double mention of the **Unknown** string at **CompatibilityInfoClass**, causes a compile warning when compiling a translation change on the first compile. Do you want me to go through and remove all the duplicates for all of the other files?

DanielSvoboda:
It is necessary to remove duplicates from the other files, if you can do this I would appreciate it

Martini-141:
Fixed all of the duplicates included some extra ones like **GUI** duplicate in russian and swedish .ts files, 

It also seems like #2367 translated **boots**  into boots as in rainboots, so I tried to use the other translations there as much as possible.

yeager:
I saw that.. Boots (the ones you wear on your feet) is not correct for an emulator :) I prefer not to change any other languages than my own. Better to set them untranslated.

PR #2384, opened 2025-02-09
renderer_vulkan: Skip tessellation isolines if not supported.

squidbus:
If portability subset features declares tessellation isolines output is not supported, skip tessellation shaders that use it.


squidbus:
Fixes some crashes on macOS in Dark Souls III.

PR #2386, opened 2025-02-09
cpu_patches: Lower extrq/insertq log to trace.

squidbus:
If a game uses `extrq`/`insertq` instructions in a way with undefined behavior according to spec, this log can spam a lot and slow things down. Since this is an expected and handled case, lower it to trace to keep logs out of the hot path.


mailwl:
TRACE too low level, may you change to DEBUG? it doesn't include to common log without set explicit in settings
TRACE level too open to spam log

squidbus:
> TRACE too low level, may you change to DEBUG? it doesn't include to common log without set explicit in settings TRACE level too open to spam log

Debug logs still consume CPU time in the logging backend and are not good to use in a hot path like this emulating a CPU instruction. These messages aren't really important enough to warrant that IMO.

mailwl:
good point

PR #2381, opened 2025-02-09
video_out: HDR support

psucien:
This adds support for HDR output in games.

Requirements are:
- game should have HDR support
- `allowHDR` is set `true` in the config
- monitor has support for BT2020 PQ color space and rgb10a2 swapchain format

PR #2387, opened 2025-02-09
Add HDR option to settings menu

tomboylover93:
Adds an option to the settings dialog window for the HDR option introduced in #2381. Also adds `setAllowHDR` to the config.h and config.cpp files, so that there are no compilation errors.

image


ghost:
i know it is already merged but why not add it to graphics?

tomboylover93:
I was considering that but since it's defined under "General" instead of "GUI" in config.cpp I decided not to.

PR #2388, opened 2025-02-09
fix: lower UBO max size to account buffer cache offset

psucien:


src/shader_recompiler/info.h
```diff
@@ -21,11 +21,13 @@
 #include "video_core/amdgpu/liverpool.h"
 #include "video_core/amdgpu/resource.h"
 
+namespace Vulkan {
+extern size_t MaxUboSize;
```
raphaelthegreat: This should be stored in Profile struct, the recompiler interface should not have vulkan specific code

squidbus (approved):

PR #2391, opened 2025-02-09
Improve parameter checks for posix_pthread_rename_np

StevenMiller123:
Based on libkernel decomp, when `posix_pthread_rename_np` is called with a null name, the function effectively does nothing and returns 0.

This fixes a `fmt::v11::format_error` thrown by CUSA14619 during a `posix_pthread_rename_np` call.


squidbus (approved):
Took a quick look myself and seems correct

Issue #2230, opened 2025-01-24
[Feature Request]: Show changelog automatically, possibly with a preference to do so or not


larsberntrop:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

I love the auto update!  I allways check the Changelog, could a preference be added to allways show the Changelog when an available update is detected?

### Reason

Save me clicking a button on every update, and I am all for people actually getting to see the Changelog.  Makes me appreciate the industrious developers even more.

### Examples

Sorry, have no example.  Having this would be an excellent example in itself!

GHU7924:
Did I understand you correctly?

Before

![20250108165304_1](https://github.com/user-attachments/assets/3015a0ba-19f1-41b5-8a63-5dd3e03c8d18)
![Image](https://github.com/user-attachments/assets/f8983312-2772-4350-88ac-1ab247e1298a)

After

![Image](https://github.com/user-attachments/assets/305c4979-10ce-4b21-8f6d-5606247d5ede)
![Image](https://github.com/user-attachments/assets/14093ce6-d8b2-49b8-af14-fadb76ce61ab)

Instead of two subsequent windows (Before), we will immediately have only one window (After), it will be slightly different depending on whether the automatic update checkmark is set or not.
Instead of the "Show Changelog" and "Hide Changelog" buttons, we will always display the Changelog and also the progress bar (if possible).

@DanielSvoboda Is it possible to do this? It would also improve the interface for Steam Deck, because now it looks like this https://github.com/shadps4-emu/shadPS4/issues/2086#issuecomment-2577785173 .

But in this case, there is a problem (for Steam Deck in game mode), as soon as 100% progress is achieved, the following window will appear, and you may not have enough time to review all the changes, so we may need to make additional changes and add the "OK" button (the button becomes active only after update downloads).

![Image](https://github.com/user-attachments/assets/cb338c11-c332-4e5f-9145-d470217b19e8)

GHU7924:
@DanielSvoboda Can you comment? I'm not talking about doing PR right now, but I'm asking about the possibility of such an implementation. Is it possible to do this, as I showed in the last screenshot?

DanielSvoboda:
interesting, but I'll look into the possibilities first.

larsberntrop:
Your proposal is excellent.  I was originally proposing to have the showing
of the Changelog switchable by a preference.  For me always showing it is
great.  But perhaps other users will object.  I suppose they can post a
[...]

PR #2401, opened 2025-02-11
QT: AutoUpdate -Formatting/Always Show Changelog

DanielSvoboda:
- Align content after 'Current Version' and 'Latest Version', regardless of language;
- The question was on the buttons side;
- Added +20 to changelog field size, should work for very large texts.
- Add option: Always Show Changelog

image

Added +20 to changelog field size
image

Always Show Changelog
image


GHU7924:
How about adding solution #2230 to this PR? At least remove the "Show Changelog" and "Hide Changelog" buttons so that Changelog is always visible.

DanielSvoboda:
> How about adding solution #2230 to this PR? At least remove the "Show Changelog" and "Hide Changelog" buttons so that Changelog is always visible.

I think that if the changelog were to be visible, it would be very polluted, and not everyone would like to read them.

That way, the button would remain, and there would be a progress bar below it, as mentioned, so it would also look a bit strange to have it and the button together. I think it looks better as it is, which only appears when the download is in progress.

*edited in paint
image

If I'm not mistaken, I had put the 'OK' button at the end to give time for a file to be copied or something like that, I remember that was on purpose, if this button remained fixed at the bottom it would be quick to click, it would be good to have it together with the text, replacing the dialog box, but I would have to analyze if it works then, it's been a while since I made it, so I'm not sure if this problem still exists.

GHU7924:
>I think that if the changelog were to be visible, it would be very polluted, and not everyone would like to read them.

Someone wants to see changes, and someone doesn't, so you can do this:

If the user checks the box, they will always see an update with the changelog shown. You can call it something else.

As for my idea, you can tell me that it doesn't fit because... (arguments) and that's it, you don't have to do it. These are just ideas that can be discussed, then abandoned or improved.

GamerzHell9137:
Tbh if im upgrading something i'd like to know what changed. At that point I dont understand why we need to know the version number either if we are not interested in the changelog? Would be nice to see it togleable and make a way to always show changelogs and the version without changelog can remove the "show changelogs" button too.

DanielSvoboda:
I added the option to 'Always Show Changelog'

DanielSvoboda:
The updater options are in "General" in the configuration file, but I think there should be a separate one just for this, since it already has all of these:
updateChannel = "Nightly"
autoUpdate = false
alwaysShowChangelog = false

However, I noticed that several other options are "messed up" in the configuration file, this should be fixed in another PR, to make the config.toml file more readable and organized.

DanielSvoboda:
I think everything is fine now

GHU7924:
@GamerzHell9137 

>if im upgrading something i'd like to know what changed.

It's a matter of taste. For some, this is important, but for others, the latest = the best version. Now the option has been added, you can always see the open changelog.

>At that point I dont understand why we need to know the version number either if we are not interested in the changelog?

Those who are not interested in this will not pay attention. Changing this moment (if at all possible) will obviously cause a headache for the one who will write the code, and these changes won't affect the user much, so here I suggest going the simpler way and leaving everything as it is.

>the version without changelog can remove the "show changelogs" button too.

At first, I also thought it would be better to remove the buttons associated with the changelog, but then I thought about it and realized that such a change deprives the user of additional features. If the user still wants to view the changes, they can do so by clicking one button.

Then I thought that maybe the "Hide Changelog" button is not needed... but for the same reasons described above, I think it can also be left.

@DanielSvoboda I apologize for for adding work to you in this PR. As for the changes for AutoUpdater, I would add a choice for the user:

I really do not know if it is worth doing this. (You said that there was a reason why you added the OK button, which most likely starts the update process itself, so I don't know if such a change will be possible).

As for my idea in #2230, I was just thinking about how to improve Auto Updater so that it works within a single window, because I don't really like how it looks on the Steam Deck. https://github.com/shadps4-emu/shadPS4/issues/2086#issuecomment-2577785173
I can quite accept the way it is now.

PR #2402, opened 2025-02-12
shader_recompiler: Use shared memory buffer for non-compute shared memory.

squidbus:
Wires up the shared memory buffer emulation to also be used for non-compute shared memory, instead of the current lower-to-registers pass, as it does not take into account the possibility of shared memory stores within branches. I've also fixed a few issues I ran into related to binding the shared memory buffer.

Should improve a few shaders in CUSA28193 that use shared memory in fragment shaders with branching write logic. I didn't see much in the way of visual changes, although a validation error at the title screen about non-dominating variable usage that resulted from the lowering pass is now gone.


squidbus:
Seems this approach may not be very practical as, with shared memory buffer emulation fixed to properly split by thread group, it will be difficult to figure out how many groups there should be for a non-compute shader. Going to replace this with a different approach that continues to be more tailored to how shared memory is actually being used in these shaders.

PR #2403, opened 2025-02-12
shader_recompiler: Lower non-compute shared memory into spare VGPRs.

squidbus:
Improves the shared memory lowering pass for non-compute by lowering into spare VGPRs instead of directly referencing the value source. This allows it to work with branching shared memory logic. The previous implementation would ignore control flow and just take the last write instruction in order before the read, but by replacing with VGPR load/store instead, we can preserve control flow.

Fixes some fragment shaders in CUSA28193 that use branching logic for writing to shared memory then read the values back later. A few shaders that would emit a validation error about non-dominating variable access are now valid.

I also fixed a few issues in the shared memory buffer binding that I spotted, with it not being accounted for in `Info::AddBindings` or incrementing the buffer count in `Rasterizer::BindBuffers`.


squidbus:
Added an additional optimization: shared memory use pattern in non-compute is typically one "register" every 256 bytes, to allow for 4 byte values across 64 threads per group. As long as this pattern holds, we can divide the offsets, and thus the size, of the lowered private memory by 64, since our private memory is per-thread and thus only one of those 64 slots will be used.

raphaelthegreat:
Are the new IR instructions really necessary here? They look basically the same to me with the existing shared memory ones. Would prefer to avoid duplication unless necessary

squidbus:
> Are the new IR instructions really necessary here? They look basically the same to me with the existing shared memory ones. Would prefer to avoid duplication unless necessary

They aren’t strictly necessary but you will then have to detect and special-case shared memory in the backend on top of the frontend IR rewriting.

raphaelthegreat:
Well given the only difference is that one uses ctx.private_u32, ctx.private_memory_u32 and other ctx.shared_u32, ctx.shared_memory_u32, you could theoretically just special case DefineSharedMemory and it would work the same, wouldnt it?

squidbus:
> Well given the only difference is that one uses ctx.private_u32, ctx.private_memory_u32 and other ctx.shared_u32, ctx.shared_memory_u32, you could theoretically just special case DefineSharedMemory and it would work the same, wouldnt it?

Yes, that’s what I’m referring to. I’m concerned with having stuff like this cross frontend and backend boundaries with no strong hint to the backend but if you’re fine with this I can change it.

raphaelthegreat:
Also I just had an idea, not saying its better or has to be implemented here. We could also have a IR pass like the old one, but instead of just linking loads and stores, it keeps a hash map that maps data-share offset -> VGPR and rewrite stores into a vgpr store and loads into vgpr load. Available vgprs can be determined by simply looking at runtime_info.num_allocated_vgprs and incrementing it for every addition into hash map (this should guarantee we dont trample over used VGPRs. Theoretically shader might use all vgprs and none is available but I think that is pretty rare). The main advantage is that it needs no new IR ops and keeps the shader simple like before

squidbus:
> Theoretically shader might use all vgprs and none is available but I think that is pretty rare

Isn’t the main use of this that registers ran out and the compiler spills over into shared memory? At least the shaders I looked at it seemed to be the case.

raphaelthegreat:
Yes but for the shader I had initially looked I don't recall it using up to V255 for example, if that is the case with your shader then we can scrap this idea. If the shader doesnt reach that highest limit, the compiler may be enforcing a lower limit for reason but no idea. Could you check what value num_allocated_vgprs has for your shader?

squidbus:
Now just lowers into unused VGPRs, and removed unused U128 shared instructions.

squidbus:
~~Also if someone could check with TLG and Dark Souls Remastered, would be appreciated as those games use this to my understanding.~~ (update: both have been checked)

PR #2405, opened 2025-02-12
Fix S_LSHR_B32

DanielSvoboda:
The shift value should be extracted from the 5 least significant bits of the second operand (S1.u[4:0]), to ensure that the shift is limited to values ​​from 0 to 31, suitable for 32-bit operations

Instruction S_LSHR_B32
Description D.u = S0.u >> S1.u[4:0]. SCC = 1 if result is non-zero.


polybiusproxy (approved):
Looks good to me

PR #2407, opened 2025-02-12
CI: Update Workflows

Xphalnos:
Updated some sub-modules from CI to main to have the latest fixes and not have to update them with each major version.


squidbus:
I don't have a big problem with it, but I don't think it's generally recommended to tag workflow versions to main.

georgemoralis:
yah i think it is not neccesary closing

PR #2409, opened 2025-02-12
Qt: Use Qt 6.8.2

Xphalnos:
Update to Qt 6.8.2 and migration to msvc2022 for Windows.


kalaposfos13:
You should also update the version numbers in documents/building_windows.md and in other places where it's needed to as well

Xphalnos:
Thanks @kalaposfos13!

PR #2411, opened 2025-02-12
Input: configurable controller light setting

kalaposfos13:
Adds a feature to configure your controller's lightbar. Example:
`override_controller_color = false, 0, 0, 255`
This means that the colors aren't overridden, and the game can still update them, however, if you make the first parameter `true`, your custom set color will be the color of your controller. The other 3 parameters are a standard RGB value.
Also adds a dummy implementation of scePadResetLightBar, by making that change the color of your controller to the value that is set.
By default, the game can update this, and the default color is blue.
Draft because I still have some things to do, for example at booting something sets the color to blue and it doesn't get updated to the correct user set color for a few seconds

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -71,6 +71,8 @@ static bool rdocEnable = false;
 static s16 cursorState = HideCursorState::Idle;
 static int cursorHideTimeout = 5; // 5 seconds (default)
 static bool useUnifiedInputConfig = true;
+static bool overrideControllerColor = false;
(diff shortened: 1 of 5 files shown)


kalaposfos13:
The thing that resets the color is `SDL_InitSubSystem(SDL_INIT_GAMEPAD);`, which I can't do anything about, so you'll just have to deal with it as this can't be removed, so this PR is complete if there are no issues (the GUI problems will be a separate PR)

georgemoralis:
let it light :D

PR #2416, opened 2025-02-13
Properly implement sceVideoOutGetEventData

StevenMiller123:
Implements sceVideoOutGetEventData based on decompilation. Several Unity games need this properly implemented for their flip threads to actually flip frames.

This brings PowerWash Simulator (CUSA29983) ingame.

PR #2423, opened 2025-02-13
kernel: dce event code cleanup

polybiusproxy:
General cleanup of the code introduced by #2330 and #2416 to (mostly) get rid of the manual bit arithmetic.

### DIFF
--- a/src/qt_gui/kbm_gui.cpp
+++ b/src/qt_gui/kbm_gui.cpp
@@ -608,21 +608,28 @@ void KBMSettings::CheckMapping(QPushButton*& button) {
     MappingTimer -= 1;
     button->setText(tr("Press a key") + " [" + QString::number(MappingTimer) + "]");
 
+    if (pressedKeys.size() > 0) {
+        QStringList keyStrings;
+
+        for (const QString& buttonAction : pressedKeys) {
+            keyStrings << buttonAction;
+        }
+
+        QString combo = keyStrings.join(",");
+        SetMapping(combo);
+        MappingCompleted = true;
+        EnableMapping = false;
+
+        MappingButton->setText(combo);
+        pressedKeys.clear();
+        timer->stop();
+    }
     if (MappingCompleted) {
         EnableMapping = false;
         EnableMappingButtons();
         timer->stop();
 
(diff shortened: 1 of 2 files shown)


src/core/libraries/kernel/equeue.h
```diff
@@ -85,19 +99,23 @@ struct EqueueEvent {
...
-                    mask = (static_cast<u32>(event.data) + 0x1000) & 0xF000;
-                }
-                event.data = (mask | static_cast<u64>(static_cast<u32>(time) & 0xFFF) |
-                              (hint & 0xFFFFFFFFFFFF0000));
+        if (hint == 0) {
+            return;
+        }
+
+        DceHint dce_hint = *reinterpret_cast<DceHint*>(&hint);
+        DceData dce_data = *reinterpret_cast<DceData*>(&event.data);
```
raphaelthegreat:
```suggestion
        DceHint dce_hint = std::bit_cast<DceHint>(hint);
        DceData dce_data = std::bit_cast<DceData>(event.data);
```

src/main.cpp
```diff
@@ -154,7 +154,7 @@ int main(int argc, char* argv[]) {
     // If no game directory is set and no command line argument, prompt for it
     if (Config::getGameInstallDirs().empty()) {
         std::cout << "Warning: No game folder set, please set it by calling shadps4"
-                     " with the --add-game-folder <folder_name> argument";
+                     " with the --add-game-folder <folder_name> argument\n";
```
squidbus: Use `<< std::endl` instead of adding `\n` to the string.

src/core/libraries/kernel/equeue.h
```diff
@@ -85,19 +99,23 @@ struct EqueueEvent {
...
+
+        const auto hint_h = static_cast<u32>(hint >> 8) & 0xffffff;
+        const auto ident_h = static_cast<u32>(event.ident << 8);
+        if (dce_hint.event_id == event.ident && event.ident != 0xfe &&
```
red-prig: It's not quite correctly interpreted, it should be more like this:
```
if (dce_hint.event_id == ((event.ident >> 48) 0xffff) && ((event.ident >> 48) 0xffff) != 0xfe &&
```

red-prig: oops

PR #2428, opened 2025-02-14
shader_recompiler: Remove special case buffers and add support for aliasing

raphaelthegreat:
Commit messages are mostly self explanatory. This wont fix any bugs, but needs some testing to ensure no regressions. After its merged 64-bit buffer atomics can be implemented as well


squidbus:
Ran through some of my games with this, no issues so far.

rafael-57:
Ran Bloodborne a bit, didn't see any issues

PR #2429, opened 2025-02-14
Qt: Resizing Font Size and Icon Grid Size

Xphalnos:
This avoids having the name very small when the icons are very large.

Before:

https://github.com/user-attachments/assets/ea476caf-9134-4677-8f0b-1f476edda43a

After:

https://github.com/user-attachments/assets/a1bfb870-191a-4ebe-9445-fd961cb40fec

### DIFF
--- a/src/qt_gui/game_grid_frame.cpp
+++ b/src/qt_gui/game_grid_frame.cpp
@@ -117,7 +117,12 @@ void GameGridFrame::PopulateGameGrid(QVector<GameInfo> m_games_search, bool from
         layout->addWidget(image_label);
         layout->addWidget(name_label);
 
-        name_label->setStyleSheet("color: white; font-size: 12px; font-weight: bold;");
+        // Resizing of font-size.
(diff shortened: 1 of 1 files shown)


ghost:
i know its already merged but this should be optional

Xphalnos:
> i know its already merged but this should be optional

Why? What is the point of having very small writing?

ghost:
2 reason that comes to mind

1.  if the text is to small the user should use the feature the OS provides to increase the size of all text if they have a small screen or they need it for accessibility issues.
2.  the huge text in contrast of other texts already in the app like the header, context menus and the rest does not look great.

and i am not saying that it shouldn't be there just that making this optional would be better 

edit: now that i think about it why is it the serial number and not the name of the game for grid view? i feel the game name would make more sense

ElBread3:
I also agree, at Medium+ size it's kinda awful imo, the text filling the same width as the icon looks bad to me, though I say we should find a middle ground and at least keep the same ratio between text width and icon width as before while still generally increasing the size a bit.

PR #2430, opened 2025-02-14
Adding KBM icon for kbm remaps to the gui. 

diegolix29:
Adding KBM icon to the Gui

Edit: Added cause makes easier for people to access the remaps of the kbm.


diegolix29:
image

GHU7924:
I have nothing against it, although, in my personal opinion, it would be better to leave everything as it was before this PR. Whatever the input device is, you still adjust the controls so that they match the buttons of the original gamepad.

But I am against the icon (keyboard), which does not match the style of the neighboring icons.
@Xphalnos can you find something suitable for the style?

diegolix29:
Yeah i can try to find another icon, the issue for me is that u need to go through 2 more windows to find kbm remaps, and that people ask still alot for it cause they dont know its available too, this way u can acces them from both controller and keyboard icon and will let people know natively the emulator supports the kbm and the remap too. If xphalnos finds one and share it here i can add it. Thanks for the feedback.

GHU7924:
I searched a bit, and I kind of applied this icon - https://id.pinterest.com/pin/358247345378799941/ (in Paint). 
I didn't observe the distance and dimensions, so it looks a bit crooked.
keyboard icon
You can also view and attach your options for comparison.

Maybe there will be an icon where there will be both a keyboard and a mouse and it will suit the style.

Xphalnos:
@GHU7924 Something like this could be good. (With Github's dark mode you can't see much)

keyboard_icon

GHU7924:
@Xphalnos It looks good too. But the "Tail" must be shortened (or not?).
keyboard icon

There is a dilemma)) In my opinion, my version is close to the image of the gamepad, and your version is closer to the rest of the buttons.

Xphalnos:
> @Xphalnos It looks good too. But the "Tail" must be shortened (or not?). keyboard icon
> 
> There is a dilemma)) In my opinion, my version is close to the image of the gamepad, and your version is closer to the rest of the buttons.

Your is good! Let's keep it!

diegolix29:
nice yeah looks good thanks i will update the icon then thanks

diegolix29:
does it looks good like this ?
image

diegolix29:
made it thicker
image
image

diegolix29:
image
zoomed

diegolix29:
comparison 
image

GHU7924:
The fact that the icon of the gamepad and keyboard is slightly different from the rest of the icons is obvious. It's also noticeable that the images are compressed from a larger format, but overall it looks good. If everyone is satisfied with everything, then you can opt for this.

diegolix29:
Ok yeah i think i maybe doing another pr later on to fix both icons so they will be less pixelated. thanks for the suggestions

PR #2431, opened 2025-02-14
Qt: Disables the Pause and Stop button when no game is launched

Xphalnos:
When no game is running, the Pause and Stop buttons are unclickable.

https://github.com/user-attachments/assets/e74ce856-34a1-42c3-a046-e4532d465c15

### DIFF
--- a/src/qt_gui/main_window.cpp
+++ b/src/qt_gui/main_window.cpp
@@ -86,6 +86,10 @@ bool MainWindow::Init() {
     CheckUpdateMain(true);
 #endif
 
+    // Makes the Pause and Stop buttons non-activatable (No game is launched).
+    ui->pauseButton->setEnabled(false);
+    ui->stopButton->setEnabled(false);
+
     auto end = std::chrono::steady_clock::now();
     auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
     statusBar.reset(new QStatusBar);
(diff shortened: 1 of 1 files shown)


georgemoralis:
i think it is better to be disabled not unchickable

Xphalnos:
> i think it is better to be disabled not unchickable

What do you mean? That the buttons are always disabled?

georgemoralis:
pause and reset should be disable if a game isn't running

Xphalnos:
> pause and reset should be disable if a game isn't running

You mean they only appear when a game is launched?
If so, I'm afraid they'll break the interface when they appear.

GHU7924:
I suggest switching to Pause for the Play button.
That is, you press Play, the game starts, and the Play button changes its appearance and functionality to Pause. (Play -> Pause)
If you press Pause, the game pauses, and the Pause button changes its appearance and functionality to Play. (Pause -> Play)
One button switches its functionality between Play and Pause. (Play -> Pause -> Play)

Reset (Restart), if possible, so that it works in two modes:
1) If you are in the game, it restarts the game.
2) If you are not in the game, then it runs the last game you played.

You can read some of my thoughts about these buttons in this https://github.com/shadps4-emu/shadPS4/pull/2441#issuecomment-2661604346 .

georgemoralis:
this doesn't have much meaning atm since Qt UI is not connected to game process

PR #2437, opened 2025-02-14
imgui: Displays FPS color based on FPS

Xphalnos:
This allows the color to change depending on the performance.
I put colors without thinking too much so I am ready to make changes if you want to change the colors, the threshold at which such color is displayed...

60 FPS:

< 30 FPS:

< 15 FPS (Bloodborne ahh performance 🙏):

PS: The limits have been changed. See https://github.com/shadps4-emu/shadPS4/pull/2437#issuecomment-2660236727

### DIFF
--- a/src/core/devtools/layer.cpp
+++ b/src/core/devtools/layer.cpp
@@ -259,7 +259,19 @@ void L::DrawAdvanced() {
 
 void L::DrawSimple() {
     const float frameRate = DebugState.Framerate;
+    if (Config::fpsColor()) {
+        if (frameRate < 10) {
+            PushStyleColor(ImGuiCol_Text, ImVec4(1.0f, 0.0f, 0.0f, 1.0f)); // Red
+        } else if (frameRate >= 10 && frameRate < 20) {
+            PushStyleColor(ImGuiCol_Text, ImVec4(1.0f, 0.5f, 0.0f, 1.0f)); // Orange
+        } else {
+            PushStyleColor(ImGuiCol_Text, ImVec4(1.0f, 1.0f, 1.0f, 1.0f)); // White
+        }
+    } else {
+        PushStyleColor(ImGuiCol_Text, ImVec4(1.0f, 1.0f, 1.0f, 1.0f)); // White
+    }
     Text("%d FPS (%.1f ms)", static_cast<int>(std::round(frameRate)), 1000.0f / frameRate);
+    PopStyleColor();
 }
 
 static void LoadSettings(const char* line) {
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -68,6 +68,7 @@ static bool vkCrashDiagnostic = false;
 static bool vkHostMarkers = false;
 static bool vkGuestMarkers = false;
 static bool rdocEnable = false;
+static bool isFpsColor = true;
 static s16 cursorState = HideCursorState::Idle;
 static int cursorHideTimeout = 5; // 5 seconds (default)
 static bool useUnifiedInputConfig = true;
@@ -280,6 +281,10 @@ bool patchShaders() {
 
 bool isRdocEnabled() {
     return rdocEnable;
+}
+
+bool fpsColor() {
+    return isFpsColor;
 }
 
 u32 vblankDiv() {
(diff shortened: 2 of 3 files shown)


bigol83:
Is it based on fps or game speed? There are plenty of games locked at 30 fps on Ps4 and having an orange fps counter for a game that runs at 29 fps is not ideal i think

Xphalnos:
> Is it based on fps or game speed? There are plenty of games locked at 30 fps on Ps4 and having an orange fps counter for a game that runs at 29 fps is not ideal i think

That's right. I set the white color limit to 20fps?

bigol83:
> > Is it based on fps or game speed? There are plenty of games locked at 30 fps on Ps4 and having an orange fps counter for a game that runs at 29 fps is not ideal i think
> 
> That's right. I set the white color limit to 20fps?

that would be good yeah

Xphalnos:
@bigol83 Thanks for the comment!

GHU7924:
Can you add green at 30-45?

And move the frames a bit:
0-19 = red
20-29 = orange
30-45 = green
46+ = white

Xphalnos:
> Can you add green at 30-45?
> 
> And move the frames a bit: 0-19 = red 20-29 = orange 30-45 = green 46+ = white

The problem is that games running at 30 fps will not be able to take advantage of the same type of colors.

GHU7924:
While I was writing the message, I did not see your correspondence. So in summary, what are the possible options?

Xphalnos:
> While I was writing the message, I did not see your correspondence. So in summary, what are the possible options?

You can enble or disable the FPS color on config.
- Under 10 FPS = Red
- Above 10 FPS and Under 20 FPS = Orange
- Above 20 FPS = White

leonskenedy1:
less than half of game locked fps = orange
less than a third of game locked fps = red

src/core/devtools/layer.cpp
```diff
@@ -259,7 +259,19 @@ void L::DrawAdvanced() {
...
+        } else if (frameRate >= 10 && frameRate < 20) {
+            PushStyleColor(ImGuiCol_Text, ImVec4(1.0f, 0.5f, 0.0f, 1.0f)); // Orange
+        } else {
+            PushStyleColor(ImGuiCol_Text, ImVec4(1.0f, 1.0f, 1.0f, 1.0f)); // White
+        }
+    } else {
+        PushStyleColor(ImGuiCol_Text, ImVec4(1.0f, 1.0f, 1.0f, 1.0f)); // White
+    }
     Text("%d FPS (%.1f ms)", static_cast<int>(std::round(frameRate)), 1000.0f / frameRate);
+    PopStyleColor();
```
viniciuslrangel, line 274: You should only Pop if you're sure it did Push (wrap with the same conditional as the previous stmt)

viniciuslrangel (changes requested):

ghost:
if a game has a 30 fps limit i feel it somehow should be detected first and imo its should be >55 white and the rest a simple yellow (edit: or orange). at super low fps like 20 or 10 the user can visibly see the drops i feel it should be at 55 to make the user aware that hey there is some hiccups here i don't really a need to over complicate things.

src/core/devtools/layer.cpp
Xphalnos: I don't know if it's me who misunderstood the meaning of the sentence but when you look at the code, all the Push are connected to the Pop. Don't hesitate to tell me if you want to correct something

viniciuslrangel: Oh, sure. Didn't see the outer white part correctly

viniciuslrangel:
It's ready to go?

Xphalnos:
> It's ready to go?

Yeah

diegolix29:
just updated my main build and dosent show. does it needs a toggle?
image

Xphalnos:
> just updated my main build and dosent show. does it needs a toggle? image

It is from below 20 FPS that we see a difference in color (Orange)

PS: It's enabled by default

diegolix29:
ohh ok ok sorry i didnt finshed the comments 
can i suggest something like 
0-10 - red
11-25 - orange
26-35 - white
36-and up -  green, 

green seems more like when u have the best performance while white seems to me like the normal game pace fps

PR #2441, opened 2025-02-15
Initial Support for Stop and Restart Game

diegolix29:
For the moment the buttons only triggers a relaunch of the emulator i still need to figure out how to add the function to pass the last eboot path to the relaunch argument to trigger the last played game on restart button press, the stop button now closes the game and restart the emulator. Any suggestions for that last part is appreciated.


diegolix29:
Seems like the sdl builds are fixed now. Just need to figure how to start again the last booted game, after  the restart button relaunches the executable.

diegolix29:
changed all qt additions to std for better merge, still need suggestions to fetch lastgamepath

diegolix29:
last step to figure this one out and should be working ok 
image

diegolix29:
Rebase and clean up

GHU7924:
I haven't fully figured out this PR yet.
What I'm going to propose now is somewhat beyond the scope of this PR.
But I want to express my opinion on this matter.

buttons

1) I think the "Pause" and "Refresh" buttons are superfluous here.
"Pause" - It would be much better if it happened like in music players. You press "Play" and the music starts playing, and the icon changes to "Pause", and then vice versa.
"Refresh" - it belongs to the list, so you can either delete it and do it, as in browsers - press the F5 key to refresh the page, or, the second option, put it in the place where the icons are resized.

2) The Restart button can be redrawn based on the Refresh button. https://github.com/shadps4-emu/shadPS4/blob/main/src/images/refresh_icon.png

3) The order of the buttons (for an example). 
buttons2

4) Once again, about the "Play" and "Pause" buttons... In the debug menu (Ctrl + F10), you can pause the game and then continue it, that is, the emulator (game) is paused, and then cancel the pause and continue playing the game.. If you can transfer this functionality to buttons in the GUI, it will be excellent.

5) When you press the "Stop" button, I personally don't really like two things:
a) The fact that a message about a successful stop is displayed is better without it, but the game would simply close.
b) Restarting the emulator... Is this really necessary? 

6) The "Stop" and "Restart" buttons... I do not know if this is correct from the point of view of the emulator resources involved, but in my opinion, if the emulator (game) was automatically paused when clicking on "Stop" or "Restart", that is, "Stop" - pause and close the game with the memorization of the game itself (to restart) and "Restart" is pausing and closing a game, and then restarting it. Or does "Restart" work like starting the last game from the game selection menu?

7) I do not know if I have said everything I wanted to say or not, but for now I will just give you one piece of advice))
You need to talk to the person who developed the Debug Menu (or Dev Tools), perhaps they will help you with the implementation or at least making edits.

Foul-Tarnished:
This should never close the emulator, but cleanly exit the game

The thing is the emulator usually freeze when trying to cleanly exit a game

diegolix29:
currently theres no stop button this one creates a quit event that also relaunches the emulator so it will kinda be a close one. The intended behaviur should be just closing the game window, but as i remember soon the game window will be reworked, so maybe i will wait for it to be implemented so i can rewrite the pr thanks

rainmakerv3:
it's also just confusing having 2 very similar looking buttons, with no indication visually of which button does what.

Better label like RPCS3:
image

Or transfer one of the functions to the menu bar similar to PCSX2
image

GHU7924:
>Or transfer one of the functions to the menu bar similar to PCSX2

Refresh
This has been around for a long time, so I suggested removing the "Refresh" icon or moving it closer to the search.

Hiding the buttons "Pause", "Reset" in the menu is a bad idea in my opinion.

>Better label like RPCS3

It's a matter of taste. I can agree with you that a "Full Screen" button could be added, but an "Open" button would be redundant in shadPS 4 in my personal opinion, and as for the reboot icon, I personally am more accustomed to seeing it like this than the one in RPCS3.

As a result, you can do something like this (this is just an example):
buttons3

And the option when you have launched the game.
buttons4

rainmakerv3:
I was not referring to the button design or choice of buttons, just the fact that they have a text label underneath each button. that could be a solution for the fact that there are two buttons that look mostly the same

GHU7924:
>I was not referring to the button design or choice of buttons, just the fact that they have a text label underneath each button. that could be a solution for the fact that there are two buttons that look mostly the same

I think it would be easier to remove the "Refresh" button and add the ability to update the list using F5.

As for the inscriptions under the buttons... I honestly don't know, it also looks pretty good.

buttons3

buttons5

But as you may have noticed, this increases the area below the buttons and also affects the functions on the right (changing the size of icons and search), they will also have to be lowered so that everything looks harmonious.

You can add such a change as an option. In "GUI Settings" add the checkbox "Explanatory captions for function buttons".

P.S. I don't know English well, so I use an online translator, which sometimes makes it difficult to understand the correctness of the translation and what is written. So I apologize for not immediately understanding your idea.

rainmakerv3:
> > I was not referring to the button design or choice of buttons, just the fact that they have a text label underneath each button. that could be a solution for the fact that there are two buttons that look mostly the same
> 
> I think it would be easier to remove the "Refresh" button and add the ability to update the list using F5.
> 
> As for the inscriptions under the buttons... I honestly don't know, it also looks pretty good.
> 
> buttons3
> 
> buttons5
> 
> But as you may have noticed, this increases the area below the buttons and also affects the functions on the right (changing the size of icons and search), they will also have to be lowered so that everything looks harmonious.
> 
> You can add such a change as an option. In "GUI Settings" add the checkbox "Explanatory captions for function buttons".
> 
> P.S. I don't know English well, so I use an online translator, which sometimes makes it difficult to understand the correctness of the translation and what is written. So I apologize for not immediately understanding your idea.

This is fine as well. The general idea is just that there should be something done so that there aren't two buttons that look the same having no visual cue to tell them apart. Your proposed solutions can also solve that

georgemoralis:
We should be able to stop and restart game process properly in order to do that

PR #2446, opened 2025-02-15
Qt: Use Qt::SmoothTransformation for Background Image

Xphalnos:
Minor Improvement of the Qt GUI.
Makes the background less pixelated.

Before:
Before 
Before1

After:
After
After1


Osyotr:
Why ignore the aspect ratio though?

Xphalnos:
> Why ignore the aspect ratio though?

Otherwise it makes stretching images

Xphalnos:
Close this because it's added on this PR https://github.com/shadps4-emu/shadPS4/pull/2451

PR #2448, opened 2025-02-15
shader_recompiler: Implement AMD buffer bounds checking behavior.

squidbus:
Implements AMD buffer bounds checking behavior, in one of two ways:
* Enables `robustBufferAccess2` Vulkan feature where supported. This should provide equivalent buffer bounds checking behavior on its own. Have also added some logs for the presence of extension features we consider optional.
* When the above is unsupported (MoltenVK is the only example I currently know of relevant to us), emulates buffer bounds checks in shader.
  * Buffer sizes are read from sharps in the SRT flatbuf, and buffer loads/stores/atomics are wrapped in comparisons checking indices against the size.
  * Vertex fetch shaders are inlined as normal buffer accesses instead of being converted to vertex bindings, so that they can be wrapped with bounds checks. As far as I know this part should not have much of a performance impact as Apple GPUs have no vertex fetch hardware similar to AMD, so they convert to regular buffers at the driver level.

I have found at least one game that needs this behavior, CUSA16404, in order to draw some 2D UI effects, particles, and flat geometry. It seems the game has a bug where it does an instanced draw with vertex buffers sized to the number of instances, and for some reason a single one of the vertex attributes is indexed by vertex ID instead of instance ID like the others. This may not have been caught in development as the value is usually zero, so with bounds checking it works.

This may also help reduce some device losses and other weird GPU corruption behavior on macOS, as previously there was no buffer robustness at all, allowing misbehaving games to trash things and cause faults.


raphaelthegreat:
First note after a quick read, to get buffer sizes in shader you can also use the srt_flatbuf UBO that is provided with every shader. The sharp_idx for every buffer resource is known at compile time so you can emit a buffer fetch and bitfield extract to get it. You  can even do it from IR after my PR

squidbus:
Updated to pull sizes from srt_flatbuf as suggested.

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
```diff
@@ -192,7 +192,25 @@ EmitContext::SpirvAttribute EmitContext::GetAttributeInfo(AmdGpu::NumberFormat f
...
+    const auto stride{OpBitFieldUExtract(U32[1], rsrc1, ConstU32(16u), ConstU32(14u))};
+    const auto num_records{rsrc2};
+
+    const auto stride_zero{OpIEqual(U1[1], stride, u32_zero_value)};
+    const auto stride_size{OpIMul(U32[1], num_records, stride)};
+    return OpSelect(U32[1], stride_zero, num_records, stride_size);
+}
+
+void EmitContext::DefineBufferProperties() {
     for (BufferDefinition& buffer : buffers) {
```
raphaelthegreat: info.buffers and ctx.buffers have the same order so you could iterate here by index and avoid storing a desc reference in BufferDefinition

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -402,104 +409,177 @@ void EmitSetPatch(EmitContext& ctx, IR::Patch patch, Id value) {
...
+        const auto result_type = is_float ? ctx.F32[N] : ctx.U32[N];
+        auto compare_index = index;
+        auto zero_value = is_float ? ctx.f32_zero_value : ctx.u32_zero_value;
+        if (N > 1) {
+            compare_index = ctx.OpIAdd(ctx.U32[1], index, ctx.ConstU32(N - 1));
+            boost::container::static_vector<Id, N> zero_ids;
+            for (u32 i = 0; i < N; i++) {
+                zero_ids.push_back(zero_value);
+            }
+            zero_value = ctx.OpCompositeConstruct(result_type, zero_ids);
```
raphaelthegreat: Could this be OpConstantComposite?

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -402,104 +409,177 @@ void EmitSetPatch(EmitContext& ctx, IR::Patch patch, Id value) {
...
+static Id EmitLoadBufferBoundsCheck(EmitContext& ctx, Id index, Id buffer_size, Id result,
+                                    bool is_float) {
+    if (Sirit::ValidId(buffer_size)) {
+        // Bounds checking enabled, wrap in a select.
+        const auto result_type = is_float ? ctx.F32[N] : ctx.U32[N];
+        auto compare_index = index;
+        auto zero_value = is_float ? ctx.f32_zero_value : ctx.u32_zero_value;
+        if (N > 1) {
+            compare_index = ctx.OpIAdd(ctx.U32[1], index, ctx.ConstU32(N - 1));
+            boost::container::static_vector<Id, N> zero_ids;
```
raphaelthegreat:
```suggestion
            std::array<Id, N> zero_ids;
            zero_ids.fill(zero_value);
```
A bit shorter

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
squidbus: Thanks, forgot this was a separate thing in SPIR-V side.

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
raphaelthegreat: Can also make the loop below a .fill() call

squidbus: Sure, done that as well.

raphaelthegreat (approved):

Issue #1880, opened 2024-12-25
Qt, GUI: "PKG Extraction" Dialog  screen centered, next Dialog "Extraction Finished" - window centered


mailwl:
Windows 11.
If main shadPS4 window not centered on screen itself, then Dialog  "PKG Extraction" - in screen center, "Extraction Finished" - in center of window.
I think, both dialogs should be in same position

Hermiten:
@DanielSvoboda Simple things to do I think :p

DanielSvoboda:
I've already helped with some things in the interface, but I'm not an expert on the subject, I'll have to leave it to someone else.

mailwl:
chatgpt suggest to add code
```
                dialog.setGeometry(QStyle::alignedRect(Qt::LeftToRight, Qt::AlignCenter,
                                                          dialog.size(), this->geometry()));
```
here:
https://github.com/shadps4-emu/shadPS4/blob/ee974414d28b4da47fda57024987057c928e2872/src/qt_gui/main_window.cpp#L870

and it works

Hermiten:
@mailwl Open a PR ? :)

Xphalnos:
@mailwl I created a PR for your issue. https://github.com/shadps4-emu/shadPS4/pull/2453
Thanks for the code idea!

Xphalnos:
Fixed by https://github.com/shadps4-emu/shadPS4/pull/2453.

PR #2453, opened 2025-02-15
Qt: Center Installation Dialog

Xphalnos:
Fixes **https://github.com/shadps4-emu/shadPS4/issues/1880**
Thanks @mailwl!

Before

After:
After


src/qt_gui/main_window.cpp
```diff
@@ -965,6 +965,9 @@ void MainWindow::InstallDragDropPkg(std::filesystem::path file, int pkgNum, int
                 dialog.setAutoClose(true);
                 dialog.setRange(0, nfiles);
 
+                dialog.setGeometry(QStyle::alignedRect(Qt::LeftToRight, Qt::AlignCenter,
```
Osyotr, line 968: Just pass parent to the dialog?

PR #2458, opened 2025-02-16
renderer_vulkan: Bind descriptors to specific stages in layout.

### LINKED ISSUE #1820: [macOS] [BB] MoltenVK Crash (possibly tessellation related)
Using latest main (ccfb1bb) Bloodborne [CUSA03173] will reproducibly crash with two errors when trying to enter the Forbidden Woods: 

```
[mvk-error] VK_ERROR_INITIALIZATION_FAILED: Vertex shader requires output buffer, but there is no free slot to pass it.
[mvk-error] VK_ERROR_INITIALIZATION_FAILED: Tessellation control shader requires tessellation level output buffer, but there is no free slot to pass it.
[Debug] <Critical> vk_graphics_pipeline.cpp:operator():339: Assertion Failed!
Failed to create graphics pipeline: ErrorInitializationFailed
```

Unfortunately I can't confirm if the tessellation branch was the cause, as all previous GitHub actions have expired. 

Edit: I should note that on a previous build I was  able to enter this area and reach the door at the end, although at that time I did not have the password to proceed further.

Terminal Log:
[BB Terminal Output.txt](https://github.com/user-attachments/files/18184324/BB.Terminal.Output.txt)

Save (just move forward through the doorway to cause the crash):
[CUSA03173_save.zip](https://github.com/user-attachments/files/18184326/CUSA03173_save.zip)

squidbus:
When defining descriptor set layouts, instead of attaching descriptors to all stages, attach only to the stage they came from.

This reduces the per-stage binding pressure, which can help MoltenVK in particular when using push descriptors to have enough binding slots for its own internal buffers for purposes like tessellation data.

Should fix https://github.com/shadps4-emu/shadPS4/issues/1820

### DIFF
--- a/src/video_core/renderer_vulkan/vk_pipeline_common.h
+++ b/src/video_core/renderer_vulkan/vk_pipeline_common.h
@@ -15,7 +15,7 @@ class BufferCache;
 
 namespace Vulkan {
 
-static constexpr auto gp_stage_flags =
+static constexpr auto AllGraphicsStageBits =
     vk::ShaderStageFlagBits::eVertex | vk::ShaderStageFlagBits::eTessellationControl |
     vk::ShaderStageFlagBits::eTessellationEvaluation | vk::ShaderStageFlagBits::eGeometry |
     vk::ShaderStageFlagBits::eFragment;
--- a/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
+++ b/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
@@ -19,6 +19,15 @@ namespace Vulkan {
 
 using Shader::Backend::SPIRV::AuxShaderType;
 
+static constexpr std::array LogicalStageToStageBit = {
+    vk::ShaderStageFlagBits::eFragment,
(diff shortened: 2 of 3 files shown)


src/video_core/renderer_vulkan/vk_pipeline_common.h
```diff
@@ -15,10 +15,18 @@ class BufferCache;
 
 namespace Vulkan {
 
-static constexpr auto gp_stage_flags =
+static constexpr auto AllGraphicsStageBits =
     vk::ShaderStageFlagBits::eVertex | vk::ShaderStageFlagBits::eTessellationControl |
     vk::ShaderStageFlagBits::eTessellationEvaluation | vk::ShaderStageFlagBits::eGeometry |
     vk::ShaderStageFlagBits::eFragment;
+static constexpr std::array LogicalStageToStageBit = {
```
raphaelthegreat: This can probably be moved to vk_graphics_pipeline.cpp since its not needed in common

squidbus: Done.

raphaelthegreat (approved):

PR #2464, opened 2025-02-17
Emulator: Create a Log Backup

Xphalnos:
Like yuzu and other emulators, it is good to have a log backup.
When starting shadPS4, the log of the previous session is renamed to `shad_log.old.txt`.
This can be useful to compare logs when making changes.
I tried to make the code as understandable as possible.

image

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -44,11 +44,20 @@ Frontend::WindowSDL* g_window = nullptr;
 namespace Core {
 
 Emulator::Emulator() {
-    // Initialize NT API functions and set high priority
+    // Initialize NT API functions and set high priority.
 #ifdef _WIN32
     Common::NtApi::Initialize();
     SetPriorityClass(GetCurrentProcess(), ABOVE_NORMAL_PRIORITY_CLASS);
 #endif
+
+    // Renaming shad_log.txt to shad_log.old.txt when booting.
+    const auto LogDir = Common::FS::GetUserPath(Common::FS::PathType::LogDir);
+    const auto CurrentLog = LogDir / "shad_log.txt";
+    const auto NewLogName = LogDir / "shad_log.old.txt";
+    // Avoid crash due to shad_log.txt not existing.
+    if (std::filesystem::exists(CurrentLog)) {
+        rename(CurrentLog, NewLogName);
+    }
 
     // Start logger.
     Common::Log::Initialize();
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -50,14 +50,8 @@ Emulator::Emulator() {
     SetPriorityClass(GetCurrentProcess(), ABOVE_NORMAL_PRIORITY_CLASS);
 #endif
 
-    // Renaming shad_log.txt to shad_log.old.txt when booting.
-    const auto LogDir = Common::FS::GetUserPath(Common::FS::PathType::LogDir);
-    const auto CurrentLog = LogDir / "shad_log.txt";
-    const auto NewLogName = LogDir / "shad_log.old.txt";
-    // Avoid crash due to shad_log.txt not existing.
-    if (std::filesystem::exists(CurrentLog)) {
-        rename(CurrentLog, NewLogName);
-    }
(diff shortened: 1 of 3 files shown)


src/emulator.cpp
```diff
@@ -44,12 +44,21 @@ Frontend::WindowSDL* g_window = nullptr;
...
-    // Initialize NT API functions and set high priority
+    // Initialize NT API functions and set high priority.
 #ifdef _WIN32
     Common::NtApi::Initialize();
     SetPriorityClass(GetCurrentProcess(), ABOVE_NORMAL_PRIORITY_CLASS);
 #endif
 
+    // Renaming shad_log.txt to shad_log.old.txt when booting.
+    const auto LogDir = Common::FS::GetUserPath(Common::FS::PathType::LogDir);
+    const auto CurrentLog = LogDir / "shad_log.txt";
```
squidbus: We have a constant `LOG_FILE` in the path util for this.

squidbus (commented):
Seems fine, although one general thought is maybe we should do this as part of the log system init alongside defining the new log file instead of out in emulator init.

src/emulator.cpp
Xphalnos: > We have a constant `LOG_FILE` in the path util for this.

I tried but it seems like `rename` ignores it when I change. Maybe it wants a "raw" filename.

Xphalnos:
@squidbus I placed the code to rename the log outside the emulator init. Is it ok now?

src/emulator.cpp
squidbus: With `const auto CurrentLog = LogDir / LOG_FILE;` right? There shouldn't be any difference since it's just a string macro.

squidbus:
> I placed the code to rename the log outside the emulator init. Is it ok now?

What I had more in mind was, inside the log system where we initialize the file output to `shad_log.txt`, insert the rename of the previous log before that. But it's not something I feel particularly strongly about, just a suggestion.

src/emulator.cpp
Xphalnos: Yes, you were right it works.

thunderstatic: Would it be more convenient if we defined the log file name without the extension? That way, we could easily change the base name in one place.
```cpp
constexpr auto LOG_FILE = "shad_log";`
```
Then construct the full paths like this:
```cpp
const auto CurrentLog = LogDir / (std::string(LOG_FILE) + ".txt");
const auto NewLogName = LogDir / (std::string(LOG_FILE) + ".old.txt");
```
What do you think about this approach?

Xphalnos: > Would it be more convenient if we defined the log file name without the extension? That way, we could easily change the base name in one place.
> 
> ```c++
> constexpr auto LOG_FILE = "shad_log";`
> ```
> 
> Then construct the full paths like this:
> 
> ```c++
> const auto CurrentLog = LogDir / (std::string(LOG_FILE) + ".txt");
> const auto NewLogName = LogDir / (std::string(LOG_FILE) + ".old.txt");
> ```
> 
> What do you think about this approach?

The problem is that `LOG_FILE` is `shad_log.txt`, if we do what you showed, it would make `shad_log.txt.txt` and `shad_log.txt.old.txt`

thunderstatic: Yes, now I see that the constant is already used in the code. My bad.

Xphalnos:
> > I placed the code to rename the log outside the emulator init. Is it ok now?
> 
> What I had more in mind was, inside the log system where we initialize the file output to `shad_log.txt`, insert the rename of the previous log before that. But it's not something I feel particularly strongly about, just a suggestion.

Tell me if I should migrate it to the log system.

gandalfthewhite19890404:
Compressing was planned? Logs may be VERY BIG.

Xphalnos:
> Compressing was planned? Logs may be VERY BIG.

Probably later but it requires a lot of extra work.

Hermiten:
I think this PR need to be put as a draft because of this one: #2504

georgemoralis:
this probably needs update

georgemoralis:
closed due to inactivity

PR #2469, opened 2025-02-17
shader_recompiler: Proper support for inst-typed buffer format operations.

squidbus:
Implements support for the buffer load format instructions that specify the type in the instruction itself. Two main parts to this:
* Refactor `BUFFER_LOAD`/`BUFFER_STORE` functions into one function each for load and store, where the formatted IR instruction is emitted based on if it's a typed operation or not, whether that is from instruction type or buffer type.
* Add IR instruction info for passing along the format from typed buffer instruction, and use this for lowering format load/store if present instead of the buffer type.

### DIFF
--- a/src/shader_recompiler/frontend/translate/translate.h
+++ b/src/shader_recompiler/frontend/translate/translate.h
@@ -277,10 +277,9 @@ public:
 
     // Buffer Memory
     // MUBUF / MTBUF
-    void BUFFER_LOAD(u32 num_dwords, bool is_typed, const GcnInst& inst);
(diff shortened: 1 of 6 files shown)


georgemoralis:
any specific game that uses this?

squidbus:
> any specific game that uses this?

It's in NBA 2K18 and at least one other game that @StevenMiller123 was testing. I don't think these games were far enough along to tell how well it's working but it should be a pretty straightforward implementation reusing the current typed buffer logic.

StevenMiller123:
NBA 2K18 and updated versions of Monster Hunter World were hitting it. Both needs hacks to actually reach that point though.

PR #2473, opened 2025-02-18
fix: remove incorrect use of std namespace

jardon:
- remove std namespace from random time refs
- remove std namespace from libc internal math

this may not be needed or could be implemented differently, but these were issues i hit building

```
/var/home/jardon/Projects/shadPS4/src/core/libraries/libc_internal/libc_internal_math.cpp:491:13: error: no member named 'cosf' in namespace 'std'; did you mean simply 'cosf'?
  491 |     *cosp = std::cosf(x);
      |             ^~~~~~~~~
      |             cosf
```

### DIFF
--- a/src/core/devices/random_device.cpp
+++ b/src/core/devices/random_device.cpp
@@ -8,7 +8,7 @@
 namespace Core::Devices {
 
 std::shared_ptr<BaseDevice> RandomDevice::Create(u32 handle, const char*, int, u16) {
-    std::srand(std::time(nullptr));
+    std::srand(time(nullptr));
     return std::shared_ptr<BaseDevice>(
         reinterpret_cast<Devices::BaseDevice*>(new RandomDevice(handle)));
 }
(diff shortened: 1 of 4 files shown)


squidbus:
What toolchain are you using? It's passing on all our clang and gcc targets in CI and according to stdlib references these should be under `std` namespace.

squidbus:
Actually I think I've seen issues like this before, you may just need to add includes to these files for the C++ version headers like `#include <cmath>` for the math functions.

rainmakerv3:
I had these issues before on ubuntu (mint) using GCC but updating to gcc 14 fixed them, looks like it was this https://bugs.launchpad.net/ubuntu/+source/gcc-8/+bug/1831385

jardon:
that was with `clang(++)` at version `18.1.3`.  i dont doubt its my environment.  ill give my builder an update and see if it persists.

jardon:
looks like updating to clang 19 fixed my issues.  closing

PR #2476, opened 2025-02-18
Handle compute packets that are split between the ends of two command buffers

kalaposfos13:
For some reason P.T. submits compute command buffers that are not aligned at instruction ends, so some instructions end up having data in two buffers. This PR fixes the issue, by doing the following:
- If an instruction is longer than the remaining buffer size, we copy the incomplete part out to a buffer
- When the next instrucion buffer arrives, we check if there were any "leftovers" from the previous one, and if so, do the following:
- Check that if we took the appropriate amount of commands from the start of the buffer, the next data is a valid instruction
- If so, we finish setting up the command by copying the remaining data to the end of the previous one, then execute this command
- We then finish handling this new buffer, starting from the appropriate offset.

If any of these fail, we proceed to handle the new buffer as normal.

### DIFF
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -63,21 +63,6 @@ static std::span<const u32> NextPacket(std::span<const u32> span, size_t offset)
     return span.subspan(offset);
 }
 
-// didn't want to modify this everywhere, just in the relevant part for testing
-static std::optional<std::span<const u32>> NextComputePacket(std::span<const u32> span,
-                                                             size_t offset) {
-    if (offset > span.size()) {
-        LOG_ERROR(
-            Lib_GnmDriver,
-            "Packet length exceeds remaining submission size. Packet dword count={}, remaining "
-            "submission dwords={}",
-            offset, span.size());
-        return std::nullopt;
-    }
-
-    return span.subspan(offset);
-}
-
 Liverpool::Liverpool() {
     process_thread = std::jthread{std::bind_front(&Liverpool::Process, this)};
 }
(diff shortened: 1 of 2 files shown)


raphaelthegreat:
I dont really like this approach too much, it feels very hacky

kalaposfos13:
Any ideas as to how else could it be done? I agree that it's a bit too hacky, that's why I initially didn't release it

raphaelthegreat:
It depends on the details of the problem. So first submit ends at the middle of a PM4 packet and next one starts right after the previous? Is there some padding in-between? If you could give a real example with addresses would be helpful

kalaposfos13:
Continuing their trend of providing a fix and then passing it off to someone else to take credit, raphaelthegreat took this PR, and made a better implementation that accomplishes the same thing, only less hacky, and told me to just add it to here. Joke's on them, I know how to use `git-blame-someone-else`, so they will get credited for it regardless :). I think that with this, the PR can finally be closed.

src/video_core/amdgpu/liverpool.cpp
```diff
@@ -726,20 +726,39 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
...
-    const auto& queue = asc_queues[{vqid}];
+    auto& queue = asc_queues[{vqid}];
+
+    auto base_addr = reinterpret_cast<VAddr>(acb);
+    while (acb_dwords > 0) {
+        auto* header = reinterpret_cast<const PM4Header*>(acb);
+        u32 next_dw_off = header->type3.NumWords() + 1;
+
+        // If we have a buffered packet, use it.
+        if (queue.tmp_dwords > 0) {
```
raphaelthegreat:
```suggestion
        if (queue.tmp_dwords > 0) [[unlikely]] {
```

src/video_core/amdgpu/liverpool.cpp
```diff
@@ -726,20 +726,39 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
...
+            queue.tmp_dwords = 0;
+        }
 
-    auto base_addr = reinterpret_cast<uintptr_t>(acb.data());
-    while (!acb.empty()) {
-        const auto* header = reinterpret_cast<const PM4Header*>(acb.data());
-        const u32 type = header->type;
-        if (type != 3) {
+        // If the packet is split across ring boundary, buffer until next submission
+        if (next_dw_off > acb_dwords) {
```
raphaelthegreat:
```suggestion
        if (next_dw_off > acb_dwords) [[unlikely]] {
```

raphaelthegreat (approved):

src/video_core/amdgpu/liverpool.cpp
kalaposfos13: Ah sorry forgot to add these

Issue #2260, opened 2025-01-28
[APP BUG]: Sorting order broken when game title is in ALL CAPS


MuntKaicho:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

It looks like there's an issue when sorting by title name and the game title is in all caps. Check attached screenshot for ordering issue with Gravity Rush 2

![Image](https://github.com/user-attachments/assets/42f91024-ffce-4a4f-9bb7-4d184656b020)

### Reproduction Steps

Have Gravity Rush Remastered and Gravity Rush 2 in your library. Possibly same issue with other franchises where only one of the entries has an all caps title.

### Specify OS Version

Windows 11

zaid-ismail031:
I'll take care of this. It looks like a minor issue with case-sensitive string comparisons in the `CompareStringsAscending` and `CompareStringsDescending` functions of the `GameListFrame` class.

DanielSvoboda:
This has been fixed for `list view`, but `grid view` still needs adjustments on startup.
If you use the `refresh` button in `grid view` it is `fixed`

zaid-ismail031:
Yes, I see. The exact same case-sensitive string string comparison bug is also present in the `CompareStrings` function in `GameInfoClass`, which results in `QVector<GameInfo> m_games` being incorrectly sorted after it is initially populated

DanielSvoboda:
That was exactly it @zaid-ismail031 , I opened a https://github.com/shadps4-emu/shadPS4/pull/2341 to fix it, thanks

Hermiten:
Fixed in #2341

PR #2480, opened 2025-02-19
Refactor string comparison helper function for case-insensitive compa…

colinmendoza121:
## Summary

Refactor case-insensitive string comparison functionality into a helper function.

## Changes

- Moved the string comparison logic for case-insensitive sorting from the `GameListFrame` class into a dedicated helper function, `CompareStringsCaseInsensitive`.
- Used `std::transform` to convert both strings to lowercase before performing the comparison.
- Ensured consistent code spacing for readability and alignment with the project's coding style guidelines.

## Motivation

The refactor improves code readability and modularity by encapsulating the case-insensitive comparison logic in a reusable helper function. This also reduces redundancy and ensures that the string comparison logic is easier to maintain and extend in the future.

## Testing

- The string comparison function was tested with various game titles to ensure the sorting works correctly regardless of case (uppercase or lowercase).
- No regressions were observed in sorting behavior, and the UI displays the game list in the correct order.

## Related Issues

Fixes Issue #2260: Sorting Order Broken When Game Title Is in ALL CAPS.

### DIFF
--- a/src/qt_gui/game_list_frame.h
+++ b/src/qt_gui/game_list_frame.h
@@ -18,6 +18,26 @@
 #include "game_list_utils.h"
 #include "gui_context_menus.h"
 
+namespace GameListFrameHelper {
+
+/**
+ * Helper function to perform a case-insensitive comparison of two strings.
+ * This function converts both strings to lowercase and then compares them.
+ * 
+ * @param a First string for comparison.
+ * @param b Second string for comparison.
+ * @return true if a < b, false otherwise.
+ */
+inline bool CompareStringsCaseInsensitive(const std::string& a, const std::string& b) {
+    std::string lower_a = a;
+    std::string lower_b = b;
+    std::transform(lower_a.begin(), lower_a.end(), lower_a.begin(), ::tolower);
+    std::transform(lower_b.begin(), lower_b.end(), lower_b.begin(), ::tolower);
+    return lower_a < lower_b;
+}
+
+} // namespace GameListFrameHelper
+
 class GameListFrame : public QTableWidget {
     Q_OBJECT
 public:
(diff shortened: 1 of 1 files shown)


raphaelthegreat:
Why not simply use std::lexicographical_compare

src/qt_gui/game_list_frame.h
```diff
@@ -18,6 +18,26 @@
...
+
+/**
+ * Helper function to perform a case-insensitive comparison of two strings.
+ * This function converts both strings to lowercase and then compares them.
+ * 
+ * @param a First string for comparison.
+ * @param b Second string for comparison.
+ * @return true if a < b, false otherwise.
+ */
+inline bool CompareStringsCaseInsensitive(const std::string& a, const std::string& b) {
```
raphaelthegreat, line 31: I think having a dedicated namespace for it is a bit overkill, also I think the function is simple enough to not need a detailed comment like this. Also it performs a bunch of extra heap allocations that are unnecessary. The following I didn't test but should have same behavior without any additional overhead
```cpp
bool CompareStringsCaseInsensitive(std::string_view a, std::string_view b) {
    auto lhs = a | std::views::transform(::tolower);
    auto rhs = b | std::views::transform(::tolower);
    return std::ranges::lexicographical_compare(lhs, rhs);
}
```

src/qt_gui/game_list_frame.h
```diff
@@ -102,10 +119,7 @@ public Q_SLOTS:
     static bool CompareStringsDescending(GameInfo a, GameInfo b, int columnIndex) {
         switch (columnIndex) {
         case 1: {
-            std::string name_a = a.name, name_b = b.name;
-            std::transform(name_a.begin(), name_a.end(), name_a.begin(), ::tolower);
-            std::transform(name_b.begin(), name_b.end(), name_b.begin(), ::tolower);
-            return name_a > name_b;
+            return GameListFrameHelper::CompareStringsCaseInsensitive(a.name, b.name);
```
raphaelthegreat, line 122: The order is reverse here compared to other place. Probably need to swap the a and b argument order

thecatontheceiling:
IMO this PR should be closed since 90% of it is written by AI and can be implemented in a much better way

PR #2481, opened 2025-02-19
input: adding ability to overwrite lightbar while on bluetooth for dualsense

diegolix29:
Now lightbar overwrite works on dualsense while using it on bluetooth havent tested on dualshock if anyone can do a quick test would be great.


diegolix29:
also noticed that once the build with my pr get the light to be overwrited i can use the overwrite on other builds that dosent have the pr, but once the controller is re pair like turned off and back the bluetooth it dosent work on the other builds again, i need to trigger it on my local build in order to get picked up on the normal builds.

diegolix29:
tested also on a dualshock bluetooth and it works ok

PR #2489, opened 2025-02-20
Fix V_FFBL_B32

DanielSvoboda:
Added check in `V_FFBL_B32` to return `0xFFFFFFFF` when S0 is zero, similar to `V_FFBH_U32`.
Otherwise, returns the position of the first low bit (bit 1) from LSB.

Instruction V_FFBL_B32
Description Find first bit low.
D.u = position of first 1 in S0 from LSB; D=0xFFFFFFFF if S0==0.


raphaelthegreat:
The existing function is correct as OpFindILsb has this behaviour builtin already. Refer to FindILsb in https://registry.khronos.org/SPIR-V/specs/1.0/GLSL.std.450.html

PR #2490, opened 2025-02-21
QT: Fix search in Grid mode

DanielSvoboda:
The search works correctly in list mode. However, when using search in grid mode, the results are correct, but the itemID values in gui_context_menus.h were not accurate.

What was happening? After searching for a game, it would show up correctly in the list. However, when trying to open or interact with it in any way (such as using cheats/patches, opening or deleting the folder, or in any other context), it was using the ID from the table without the search filter.

Watch the video, it's easier to understand:

https://github.com/user-attachments/assets/1c494e51-f9ad-4ed8-8c8f-1fd7a0ab06b3

This PR fixes this issue, and also when switching from Grid to List or vice versa the search field is reset, as this would cause problems.

### DIFF
--- a/src/qt_gui/game_info.cpp
+++ b/src/qt_gui/game_info.cpp
@@ -49,6 +49,9 @@ void GameInfoClass::GetGameInfo(QWidget* parent) {
                   return readGameInfo(Common::FS::PathFromQString(path));
               }).results();
 
+    // used to retrieve values after performing a search
+    m_games_backup = m_games;
+
     // Progress bar, please be patient :)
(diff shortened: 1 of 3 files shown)


DanielSvoboda:
To improve performance I had put this, but if I search for example 'sonic mania' and I had other sonic games, when deleting the word 'mania', the list was not updated, only when deleting everything...
```cpp
    if (text.isEmpty()) {
        m_game_info->m_games = m_game_info->m_games_backup;
        m_game_grid_frame->PopulateGameGrid(m_game_info->m_games, false);
        m_game_list_frame->PopulateGameList()
    }
```

DanielSvoboda:
This should solve the performance issue 🐌

PR #2491, opened 2025-02-21
libcInternal HLE fixes

georgemoralis:
-internal_vprintf was wrong (commented for now i will provide a fix one)


squidbus (approved):

georgemoralis:
I can now play project diva future tone with only need ngs2 lle prx

εικόνα

Issue #2461, opened 2025-02-16
[Feature Request]: better achievmnet pop ups and viewer


ghost:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

![Image](https://github.com/user-attachments/assets/257588f9-9940-4f87-bcab-b4748fe9d45a)

the current implementation while works well there are some small things that can improve the experience even more

1. the biggest issue imo is that its not centered and while small issue but can make it looks better
2. maybe a small sound to make our brain release some dopamine while at it?
3. showing the rarity of it by having the gold bronze or silver trophies can be a nice touch

![Image](https://github.com/user-attachments/assets/5423c1da-c883-44b2-9587-42c40e159217)

the viewer have some small things that can make it look better

1. the locked and unlocked while does the job i feel a lock icon and unlocked icon can be nicer to the eyes, or maybe doing it like how steam does it making the unlocked ones grayed out or darker 
2. the PID, i don't know what it does tbh and i feel removing it (unless its important?) can make the page less full of text.
3. having icons for bronze, silver, gold and plat would be awesome
4. its personal opinion here but making the images smaller and having some free space between them would make it cleaner.
5. because the space for the description is fixed, descriptions have different amount of space between the text and the edge. i feel i can look better if they are all aligned.
i mean this

![Image](https://github.com/user-attachments/assets/5a54c6c8-bf99-4194-8e17-44944d1db62d)

thank you so much for the great work by the way you guys rock

### Reason

looks better in term of UI and UX

ghost:
oh and having the images have rounded corners would be awesome too

aminenr-hub:
adding the ability to add custom audio file for the notification like rpcs3 would be awesome

Hermiten:
Some of your requests have already been made and others will follow. I take the liberty of closing this issue, which has been taken into account.

RyuSeisuke1992:
I would like to briefly comment on this. Why are trophy pop-ups generally in terms of presentation, appearance, and sound so unspectacular on the Emulator eg RPCS3, Vita3K, and now also on ShadPS4? The emotional impact of trophy pop-ups on original consoles such as the PS3, Vita, or PS4 is actually an important part of the gaming experience. Can't we work towards something equivalent? I know that a lot of this has to do with copyrights. Maybe we could take a cue from the Steam Deck, which also has very good trophy pop-ups.

shadps4:

[...]

PR #2493, opened 2025-02-21
Trophy pop-up and viewer enhancements

rainmakerv3:
Partially addresses https://github.com/shadps4-emu/shadPS4/issues/2461

Trophy pop-up
1) show icon for trophy rarity next to the trophy art
2) center all elements vertically
3) minor spacing adjustments

Trophy viewer
1) show icon for trophy rarity instead of text
2) remove the hardcoded newlines that were being imported into the viewer from the XML
3) align left instead of center for trophy descriptions

Images, though patterned after the PS4 trophy icons, are originally modelled by discord user Tlarok. They are packaged inside the AppImage for Linux, inside the app bundle for Mac, and in a separate Resources folder which is packaged in the other builds, and copied by CMake command in the binary folder for locally compiled builds

sample photos:

Trophy pop-up (main):
plat main

Trophy pop-up (PR):
plat PR

Trophy viewer (main):
viewer main

Trophy viewer (PR):
viewer PR


georgemoralis:
seems fancy  :D

.github/workflows/build.yml
```diff
@@ -94,11 +94,14 @@ jobs:
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --parallel $env:NUMBER_OF_PROCESSORS
 
+    - name: Move Files and Resources
+      run: "mkdir upload\n\t move build/Resources upload\nmove build/shadPS4.exe upload\n"
```
squidbus: You can do multiline commands with `|`, just look for anywhere in the file with `run: |` for an example.

diegolix29:
Tbh i dont like the two icons one next to the other. Seems odd. Should be one icon then the info and then the other one. If not looks too busy for the eye.

CMakeLists.txt
```diff
@@ -1144,3 +1148,19 @@ if (ENABLE_QT_GUI AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
...
     install(FILES "src/images/net.shadps4.shadPS4.svg" DESTINATION "share/icons/hicolor/scalable/apps")
 endif()
+
+
+# copy Resource folder (trophy images) inside Mac App bundle resources folder
+if (ENABLE_QT_GUI AND APPLE)
+add_custom_command(TARGET shadps4 POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy_directory
+    "${CMAKE_SOURCE_DIR}/Resources"
+    "$<TARGET_FILE_DIR:shadps4>/../Resources"
```
squidbus: Typically files are added to the bundle resources like so:
```
    set_source_files_properties("your_source_file_path_here" PROPERTIES
        MACOSX_PACKAGE_LOCATION Resources/your_path_to_put_in_resources_here)
```

GHU7924:
I agree with @diegolix29 about the icons. I suggest options (for example):

Version 1
V1

Version 2
V2
If anyone has any other options, then suggest it.
##
Or, if possible, do something like a double animation, but so far I do not know if it will look good.

1) First, a sound is heard and such an icon is displayed for about a second, and then disappears.
Part1
or
Part1v2

2) Then the achievement is displayed, but without sound.
Part2

P.S. Although actually, I think we can limit ourselves to just the changes for the Trophy Viewer.

aminenr-hub:
the second version looks better to me . also adding the ability to add custom audio file for the notification sound like rpcs3 would be awesome too

ghost:
OMG, you are a god.

version 2 looks better imo as well.

thanks a lot

rainmakerv3:
Applied squidbus' review comments regarding formatting of build workflow and Cmake command, adjusted layout as per suggestions to the ff:
new

Sound for trophy pop up will not part be of this PR, have to read up on that first and then will make another PR if its something I can do

squidbus:
Only other thing I have to add is, I'm not a big fan of having these extra resources next to the executable file, especially since until now the SDL builds were basically standalone files. I would prefer if we could maybe embed them in and reference as a byte array variable similar to how we do the `host_shaders` [here](https://github.com/shadps4-emu/shadPS4/blob/main/src/video_core/host_shaders/CMakeLists.txt#L25-L41).

rainmakerv3:
> Only other thing I have to add is, I'm not a big fan of having these extra resources next to the executable file, especially since until now the SDL builds were basically standalone files. I would prefer if we could maybe embed them in and reference as a byte array variable similar to how we do the `host_shaders` [here](https://github.com/shadps4-emu/shadPS4/blob/main/src/video_core/host_shaders/CMakeLists.txt#L25-L41).

This seems to embed each file as a header file in the exe if I'm understanding right. It is a bit beyond my current knowledge, but I'll try my best to apply it

squidbus:
> This seems to embed each file as a header file in the exe if I'm understanding right. It is a bit beyond my current knowledge, but I'll try my best to apply it

Yes, basically it results in headers with a variable for each file's contents, then if you want to load that file you can just reference that in-memory array instead of reading from file.

rainmakerv3:
I have embedded the files in the exe as asked, using this handy cmake script: https://github.com/vector-of-bool/cmrc

This should address all the comments and suggestions so far.

squidbus:
Will let others weigh in on the UI, but looks good to me now as far as my concerns went.

jardon:
a couple things:

1. what are those changes in the github actions?  theyre just version bumps but are they related to the changes in the PR
2. where did those trophy icons come from?

jardon:
whats in your PR looks good.  theres also the option to go with a layout more similar to an actual PS4 notification

image

georgemoralis:
without the ps logo pls

ghost:
> whats in your PR looks good. theres also the option to go with a layout more similar to an actual PS4 notification
> 

call me crazy but i thing the PR looks better than the way PS4 looks

aminenr-hub:
> > whats in your PR looks good. theres also the option to go with a layout more similar to an actual PS4 notification
> 
> call me crazy but i thing the PR looks better than the way PS4 looks

my thought exactly , they could refine it more to look like the ps5 one which looks pretty good ( by also making the notification box a bit longer for trophy names that are a bit longer )
image

ghost:
> my thought exactly , they could refine it more to look like the ps5 one which looks pretty good ( by also making the notification box a bit longer for prophy names that are abit longer ) 

i feel if we continue it might get into subjective preference territory :)  but i agree with the idea of making it a little longer just in case since that seems to be the reason why ps4 and ps5 notifications have that much free space

aminenr-hub:
> > my thought exactly , they could refine it more to look like the ps5 one which looks pretty good ( by also making the notification box a bit longer for prophy names that are abit longer )
> 
> i feel if we continue it might get into subjective preference territory :) but i agree with the idea of making it a little longer just in case since that seems to be the reason why ps4 and ps5 notifications have that much free space

true i m just happy we gettihng a trophy makeover for the emulator lol , since i do like to trophy hunt 
image
( this is playnite + successstory-plugin that has support for shadps4 )

rainmakerv3:
> a couple things:
> 
> 1. what are those changes in the github actions?  theyre just version bumps but are they related to the changes in the PR
> 2. where did those trophy icons come from?

1. It's actually identical to this https://github.com/shadps4-emu/shadPS4/pull/2487. When I began the PR I was using the build doc prior to this version bump, then I made certain changes in the original verison of the PR, but I reverted them to address squidbus comments. The easiest way for me to revert was to just grab the build.yml from the repo, which is updated with the version bumps. therefore, this has no effect on main, it just shows up as changes to the PR

2. As I said in the initial description, though they are patterned after the ps icons, these are originally 3D-modelled by Tlarok, a regular in the shadps4 discord

In response to the comments, some trophy names are really long like this one for grim fandango (It's like I'm not happy unless I'm breathing in the thick, black, nauseating fumes...), so the text box would have to really long to fit everything in one line. Right now, it's at a certain width, but the text will automatically wrap to the next line if the title is too long, so it will always fit

I like the PR a bit better than the PS layout as well though I would admit there could a bit of bias in there :p

aminenr-hub:
image
just downloaded the latest nightly build which has this pr merged as the change-log mentioned but i m still getting the old trophies list with no trophy icons .

ghost:
same for me on windows

jardon:
why are those trophy icons not in `src/images`?

rainmakerv3:
Something was wrong with the embedded icons for a lot of people so I need to work on it a bit more.

the icons were put in another folder because of a method I used previously which depended on a function that iterates over a certain directory (which I might have to use again since the method I ended up with is not working). I suppose we can put them in src/images/Resources of something similar but personally, I think it's okay to have them in their own folder since they are not really intended to be used by anything other than the embedding function. Either way should be functionally fine though (once I get it to actually work that is)

PR #2496, opened 2025-02-21
QT: Various interface fixes

DanielSvoboda:
- The 'Title Music' volume slider updates in real-time.

https://github.com/user-attachments/assets/ded77c81-d058-4748-9203-1687081b3a46

----

- The 'Background Image' opacity slider updates in real-time.

https://github.com/user-attachments/assets/e0c3d014-289e-4810-9a0b-a3f865e9f6d6

----

When you close the settings screen, the information will be updated. For example, if you enable/disable the "Show Game Size in List" option and close the settings screen, you will immediately see the changes without needing to manually refresh the list.

If "Show Game Size in List" is disabled, the "Copy Size" option will not appear, and 'Size' will also be missing from the "Copy All" option.

### DIFF
--- a/src/qt_gui/cheats_patches.cpp
+++ b/src/qt_gui/cheats_patches.cpp
@@ -26,8 +26,10 @@
 #include <QTextEdit>
 #include <QVBoxLayout>
 #include <QXmlStreamReader>
-#include <common/logging/log.h>
+
 #include "cheats_patches.h"
+#include "common/config.h"
+#include "common/logging/log.h"
 #include "common/memory_patcher.h"
 #include "common/path_util.h"
 #include "core/module.h"
@@ -92,7 +94,7 @@ void CheatsPatches::setupUI() {
     gameVersionLabel->setAlignment(Qt::AlignLeft);
     gameInfoLayout->addWidget(gameVersionLabel);
 
-    if (m_gameSize.left(4) != "0.00") {
+    if (Config::GetLoadGameSizeEnabled()) {
         QLabel* gameSizeLabel = new QLabel(tr("Size: ") + m_gameSize);
         gameSizeLabel->setAlignment(Qt::AlignLeft);
         gameInfoLayout->addWidget(gameSizeLabel);
--- a/src/qt_gui/gui_context_menus.h
+++ b/src/qt_gui/gui_context_menus.h
@@ -84,7 +84,9 @@ public:
         copyMenu->addAction(copyName);
         copyMenu->addAction(copySerial);
         copyMenu->addAction(copyVersion);
-        copyMenu->addAction(copySize);
+        if (Config::GetLoadGameSizeEnabled()) {
+            copyMenu->addAction(copySize);
+        }
         copyMenu->addAction(copyNameAll);
 
         menu.addMenu(copyMenu);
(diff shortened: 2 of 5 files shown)


ghost:
if its not a problem, can i make some comments about some stuff that can be improved? 
i just felt it might be a good place since it seems you want to fix some different UI stuff

DanielSvoboda:
> if its not a problem, can i make some comments about some stuff that can be improved? i just felt it might be a good place since it seems you want to fix some different UI stuff

Yes, you can comment, depending on what it is I can try to do it, or leave it for another PR, feel free to give your opinion

ghost:
thanks, i have a things i want to give my opinion on, almost all of them are more like consistency of the ui and not issues but still


the first one is the icons used for the keyboard and the controller the icons used for them are thin lines and pretty detailed and look out of place with the other icons beside them. in short my idea is that these 2 icons are too detailed for their own good.

{CA6DC29D-18E1-4DE0-8EC2-16E894B0EE71}
in my opinions simple icons with a little thicker lines would make it more consistent with the rest
maybe something like these
https://www.flaticon.com/free-icon/game_13838536?term=game+controller&page=1&position=48&origin=tag&related_id=13838536

or something really simpler to keep it minimal like this one
https://www.flaticon.com/free-icon/console_686589?related_id=686589


and for keyboard maybe something simpler like this
https://www.flaticon.com/free-icon/keyboard_2507349?term=pc+keyboard&page=9&position=23&origin=search&related_id=2507349


another thing that i think can be changed is the spacing or padding above and below the options, while this page looks nice because they are the same height, the other tabs like GUI, graphics and users dont follow the same logic, i feel making them all follow the same shape would be better, all of them uses the least space they need or making all rows have the same height  to make them look consistent.  
{4512CCF3-A544-467A-B83D-DA0214A5118A}


another one is that (just in my opinion) the borders of the search is brighter than it should be? it looks fine in the light theme but in other themes the pure white can be tweaked a little maybe
{D048F9A2-9872-4076-B782-77CE891B5ED6}
by the way i do feel the theme options should be in the GUI section of the settings as well? or maybe just in the settings and not in the View menu only.


these ones are mostly things i dont understand why they exist
by right clicking the toolbar 
{EE84F26E-9F21-4B10-95D6-1CD00EF16533}
you can use this game list toggle to disables the entire panel of game list, and you can do it with pressing that X in the corner of the game list as well
{C8582E37-F5FF-4640-B0A1-7BF41B4909C6}
and for some reason its different than the toggle that exist in the view options. maybe it can be removed?
and this is my opinion but i don't think anyone is going to have their game list as a separate windows so this option can be removed perhaps? i mean this option
Untitled



i want to say all of these are subjective maybe someone can have a better idea than mine and im just being dumb. thank you for the great work tho, really appreciate it

edit: some formatting changes

GHU7924:
Interface

I think that the items highlighted in yellow can actually be deleted, because: 
1) I also don't see the point in separating the list from the emulator. 
2) Showing these two elements by right-clicking, especially since these options can feel great in the "View" section.

DanielSvoboda:
There are quite a few changes, I think it would be good to record them in an 'issues'

ghost:
> There are quite a few changes, I think it would be good to record them in an 'issues'

alrighty then, ill make an issue later for it

DanielSvoboda:
And about the first video, yes it is possible to add 'custom' sounds, just copy the mp3 file to the game folder, example: CUSA00000\sce_sys and rename the mp3 to snd0.at9

PR #2497, opened 2025-02-21
Adding stop function to help with check update while running a game

diegolix29:
this pr helps with the lock error while updating if u running a game 
image

https://github.com/user-attachments/assets/97ae5abe-1eab-49a7-ad4a-7130b224d16d

now if u update while running a game it shows a box with a warning and then stop the game to let the update installation continue without issues

image

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -289,14 +289,6 @@ void Emulator::Run(const std::filesystem::path& file, const std::vector<std::str
     std::quick_exit(0);
 }
 
-void Emulator::saveLastEbootPath(const std::string& path) {
-    lastEbootPath = path;
-}
-
-std::string Emulator::getLastEbootPath() const {
-    return lastEbootPath;
-}
-
 Emulator& Emulator::GetInstance() {
     static Emulator instance;
     return instance;
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -102,6 +102,7 @@ Emulator::~Emulator() {
 }
 
 void Emulator::Run(const std::filesystem::path& file, const std::vector<std::string> args) {
+    is_running= true;
     const auto eboot_name = file.filename().string();
     auto game_folder = file.parent_path();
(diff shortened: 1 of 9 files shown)


georgemoralis:
that is not a clean way to do this and will probably cause more issues ;/

diegolix29:
You think maybe just the stop game on update and not the stop button could work? So it will not crash and not update when updating having a game running?

georgemoralis:
why you should update emu while running a game in first place :D

diegolix29:
Haha practicality like i play for hours and if an update rolls up i have to quit close reopen then update wait and then replay. With this just quit inside the game and hit update. It will update and not mess savings haha

rainmakerv3:
Force closing a running game is probably not a good practice. And the emulator update pop-up should only appear on start-up no?

I think there should just be some check to prevent updating during a running game, or disable the button when the game is running

diegolix29:
Like i say it adds alot of steps to continue playing when an update rolls up. For me this is useful as the update force closes the emulator cause it needs it. This (without the stop button) adds a flag for people to quit the game or save properly and does the update as intended. Without having to close open update close open and play again.

rainmakerv3:
i still can't see the value. the auto update check will pop during startup, at which point the user will update, or they will not. I don't know what you mean when you say "when an update rolls up" because users are not notified of available updates while playing. this will occur only the next startup, when the update can be done seamlessly and safely. As I told you before when you asked me to look at your other PR, I did not want to do so because the fact the game is being forced closed means the PR has good chance to be closed, which indeed it now has, and this one as for for the same reason.

I don't have anything further to add. You're free to resubmit the PR ig but I don't see how that would change anything

diegolix29:
As i stated when i play for hours and an update rolls up this is useful.  Im no longer talking about stop function or restart. But thanks for the feedback and opinion. 🙏 I  wont summit more stuff in awhile i will just do one last thing i was working since my last pr and thats about it.

PR #2499, opened 2025-02-22
vk_instance: Re-enable getToolPropertiesEXT on AMD proprietary.

squidbus:
Issue with Reshade should be fixed in new version 6.4.0, need someone to test.


larfy298:
Tested on windows qt w/ 7900XTX and reshade 6.4.1 and just got crashes on launch, can't get to menus

squidbus:
That's disappointing, do you happen to have a trace or anything of where it crashes? And a log.

georgemoralis:
since  this reported to be still buggy let's leave it as it is for now

PR #2502, opened 2025-02-22
host_shaders: Simplifying a calculation

Xphalnos:
Micro optimization of a calculation. Is there a reason for doing this or is it an oversight?

PS: I also removed some trailing whitespaces


georgemoralis:
no :D

PR #2504, opened 2025-02-22
Add option to save logfiles seperate for each game

Fire-Cube:
With this pull request, an option is added to the settings to save a separate log file for each game. For example as "CUSA19120.log". If not activated, everything stays as before. This can be very useful for debugging.


ghost:
why not make it the default behavior instead of it being optional?

Fire-Cube:
> why not make it the default behavior instead of it being optional?

I can do this when it is desired

GHU7924:
@Xphalnos Will it interfere with your PR #2464? Maybe you should combine your ideas and improve the log?

georgemoralis:
now that is an interesting idea indeed :) (the pr)

GHU7924:
@georgemoralis I agree, but what about a situation where some logs reach large sizes (like 100 MB) due to spam from a single team (cases like http in the Bloodborne log). Theoretically, you can have logs and even backup logs for each game, but this needs to be regulated somehow. We will either have to manually delete them if they are not needed, or we will need a mini-editor, something similar to what is on the Paths tab, to view all your logs.

Fire-Cube:
@GHU7924 I think we can create a certain number of backup files that will be compressed. Too old files should be deleted

GHU7924:
@Fire-Cube I mentioned another PR above. Two logs are created there, the previous one and the current one. I think that other PR is a good idea, and I think it can be useful for testing, where you can also compare the logs of two launches of the same game. With your PR, it will be easier for users to navigate magazines. Let's say you have 10 games and each one has been launched twice, which means you will have 20 logs. 

Deletion is already happening there, if I remember correctly, or overwriting. On a new start, the old log will be deleted, the second log will be renamed to the old one and a new one will also be created, which, as you understand, does not change the number of logs.

To be honest, I don't know how to improve this moment yet. You were right to make it optional.

Fire-Cube:
@georgemoralis Could we merge this? I think the problem with backups should be worked on the other PR

GHU7924:
If I understand correctly, then to merge, you need to fix Clang and note that the PR is ready for review.

src/common/config.cpp
```diff
@@ -69,6 +69,7 @@ static bool vkHostMarkers = false;
 static bool vkGuestMarkers = false;
 static bool rdocEnable = false;
 static bool isFpsColor = true;
+static bool isSeperateLogFilesEnabled = false;
```
kalaposfos13: Minor nitpick, but you misspelled the word "separate" on all occasions.

src/qt_gui/settings_dialog.ui
```diff
@@ -1,6 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
-     SPDX-License-Identifier: GPL-2.0-or-later -->
```
kalaposfos13: You need to add this back, so the file is REUSE compliant.

src/common/config.cpp
Fire-Cube: oh sorry
english is not my native language

Fire-Cube:
i think it should be okay now

Fire-Cube:
clang error fixed now
can i run this check locally?

kalaposfos13:
You can set up your fork to run the actions automatically, or alternatively you can run `./.ci/clang-format.sh`, but only on Linux

PR #2510, opened 2025-02-23
Implement some RDNA flags

OFFTKP:
Originally in #728, was postponed because of some concerns with ADDC, which I see are now resolved so the rest of the instructions with TODO: Carry out can be implemented


Hermiten:
Did you find a game that use it ?

raphaelthegreat (approved):

PR #2521, opened 2025-02-23
Fix capitalization mistake

Randomuser8219:
Change "Gui" to "GUI"


jardon (approved):
LGTM

that is the only relevant reference that i see in the source

PR #2522, opened 2025-02-23
Fix and add back Trophy type icons for both the trophy pop-up and viewer

rainmakerv3:
Fixes and adds back the type icons in the trophy viewer. Does not include the trophy pop-up icons yet, that one is a bit more complicated.

*Should be working now, but would like some tests from other people to be sure

image

### DIFF
--- a/src/core/libraries/np_trophy/trophy_ui.cpp
+++ b/src/core/libraries/np_trophy/trophy_ui.cpp
@@ -11,8 +11,6 @@
 #include "imgui/imgui_std.h"
 #include "trophy_ui.h"
 
-CMRC_DECLARE(res);
-
 using namespace ImGui;
 namespace Libraries::NpTrophy {
 
--- a/src/core/libraries/np_trophy/trophy_ui.cpp
+++ b/src/core/libraries/np_trophy/trophy_ui.cpp
@@ -11,6 +11,8 @@
 #include "imgui/imgui_std.h"
 #include "trophy_ui.h"
 
+CMRC_DECLARE(res);
+
 using namespace ImGui;
 namespace Libraries::NpTrophy {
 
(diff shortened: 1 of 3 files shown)


rainmakerv3:
Pop-up icons should be fixed as well.

**Still need other people to test to make sure this time maybe @Missake212 ? :D @brad0demx ? :D

image

brad0demx:
Seems to be working fine for me 

image
image

brad0demx:
Tested on linux too and its also working on here

image
image

rainmakerv3:
Great, thanks very much!

rainmakerv3:
With the tests of brad, myself, and another from discord, I believe this is now working properly.

Added another thing I was planning to do since it's related, which is to adjust the size of the the text based on number of lines,

If the trophy name is one line (probably the large majority), text is centered vertically:


If 2 lines, align it to the top of the trophy icon, make text slightly smaller


If 3 lines, still aligned to the top, and text is even smaller. It's a bit cramped, but very few trophies have names this long

GHU7924:
My opinion is about two things in your PR.

1) Platinum icon. It's not the same in color, and the trophy itself should be different in shape from the rest - it's platinum.
Something similar in form and style would be - [this is just for an example](https://static.vecteezy.com/system/resources/previews/047/451/222/non_2x/3d-golden-luxury-style-award-show-trophy-elegant-style-3d-golden-award-show-trophy-photo.jpg) .

2) The inscription "Trophy earned" should be larger than the text of the trophy received.

I'm not insisting, the decision is entirely yours, I just expressed my point of view.
If everyone is happy with what they have now, then I don't mind.

Hermiten:
If it's easy to fix, I'm agree with a text for Trophy Earned larger than the rest or in bold

rainmakerv3:
I am ok with changing the icon, though I am not an icon designed myself, I just got this from a discord user. If you or anyone else can come up with a new icon or even icon set, I would gladly consider it for use.

Two text string with different sizes is a bit complicated though (at least for my skill level hehe). We would have two strings that each have to be drawn independently on top of each other, with each element scaling with window size from a different base size. It's a fair bit of work for minimal benefit I feel.

GHU7924:
>I am ok with changing the icon, though I am not an icon designed myself, I just got this from a discord user. If you or anyone else can come up with a new icon or even icon set, I would gladly consider it for use.

Three trophies in the form of cups, and the platinum one should be different, it could be like what I gave as an example, or a crown, or something else, it needs to be considered.

You need to set up a contest in Discord so that craftsmen can draw great trophy icons, and the winner's work will get into the emulator 😄 .

>Two text string with different sizes is a bit complicated though (at least for my skill level hehe). We would have two strings that each have to be drawn independently on top of each other, with each element scaling with window size from a different base size. It's a fair bit of work for minimal benefit I feel.

In this case, you need to ask for help. @DanielSvoboda Are you able to solve this problem?

bigol83:
I can confirm this works great for me too and i agree that platinum trophy should have a different shape compared to the rest of the trophies

rainmakerv3:
this is merged now so it won't be part of this PR obv, but changing the icon would be quite easy to PR, it just changing the corresponding file in the source folder. so it really does not need me, it can be done by anyone

Hermiten:
> this is merged now so it won't be part of this PR obv, but changing the icon would be quite easy to PR, it just changing the corresponding file in the source folder. so it really does not need me, it can be done by anyone

Exactly, thanks for your work !

PR #2525, opened 2025-02-24
video_core: Various small improvements and bug fixes

raphaelthegreat:
* Adds another case of missing shared memory barriers which fixes more LDS artifacts
* Improves texture cache overlap resolution to allow copying individual slices in a multi-layer mip-mapped image (credit to poly for testing and brainstorming)
* Fixes a crash when CE task finishes execution earlier than GFX task.
* Fixes crashes when emitting Quad/Rect tess shaders with default attributes


src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp
```diff
@@ -118,6 +118,10 @@ struct QuadRectListEmitter : public Sirit::Module {
 
         // Set attributes
         for (int i = 0; i < inputs.size(); i++) {
+            const auto& input = fs_info.inputs[i];
+            if (input.IsDefault()) {
```
squidbus: I'm not clear on what the purpose of this, inputs are already not added to the vector when default: https://github.com/shadps4-emu/shadPS4/blob/bc0b42ee539113fdd4a8a41aca207f905dcd61d1/src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp#L280-L284

raphaelthegreat: Yes but the loop here attempts to access them and crashes in OpAccessChain so we need to skip over them in similar fashion

squidbus: Then I think the vector should just be fixed to not have bad slots and only contain the valid inputs, no need to do this check everywhere.

Missake212:
This breaks Bloodborne 

image

squidbus:
Also breaks CUSA28193 loading into open world areas with this assert: https://github.com/shadps4-emu/shadPS4/blob/207facd7180114678846b2e21eab38d7ccede9c2/src/video_core/texture_cache/image_info.cpp#L253

raphaelthegreat:
> Also breaks CUSA28193 loading into open world areas with this assert:
> 
> https://github.com/shadps4-emu/shadPS4/blob/207facd7180114678846b2e21eab38d7ccede9c2/src/video_core/texture_cache/image_info.cpp#L253

This assert should probably be removed, as the function should be allowed to return mip 0. I can imagine this happening in cases where individual slice of first mip is accessed, so it takes right overlap path

squidbus:
> This assert should probably be removed, as the function should be allowed to return mip 0. I can imagine this happening in cases where individual slice of first mip is accessed, so it takes right overlap path

Yes seems fine when removed.

squidbus:
> This breaks Bloodborne

For this, I can see the same sort of effect as well in CUSA16404 on characters.

raphaelthegreat:
This probably has something to do with the new overlap handling commit

raphaelthegreat:
@Missake212 The bug should be fixed now

squidbus:
No longer seeing any of reported issues so far.

src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp
raphaelthegreat: Done

PR #2526, opened 2025-02-24
Allow our BMI1 emulation to work on non-macOS CPUs - also emulate TZCNT

OFFTKP:


src/core/cpu_patches.cpp
```diff
@@ -616,6 +640,15 @@ static bool FilterNoSSE4a(const ZydisDecodedOperand*) {
     return !cpu.has(Cpu::tSSE4a);
 }
 
+static bool FilterNoBMI1(const ZydisDecodedOperand*) {
+#ifdef __APPLE__
+    return FilterRosetta2Only(nullptr);
+#else
+    Cpu cpu;
+    return !cpu.has(Cpu::tBMI1);
```
squidbus, line 671: Should be fine to do this either way, Rosetta 2 does not advertise BMI1.

src/core/cpu_patches.cpp
```diff
@@ -202,21 +202,19 @@ static void RestoreStack(Xbyak::CodeGenerator& c) {
 
 /// Switches to the patch stack, saves registers, and restores the original stack.
 static void SaveRegisters(Xbyak::CodeGenerator& c, const std::initializer_list<Xbyak::Reg> regs) {
-    SaveStack(c);
+    c.lea(rsp, ptr[rsp - 128]); // red zone
```
squidbus, line 210: What if the instruction references the stack pointer? Previously we would restore the stack pointer before executing the instruction.

squidbus: Also, I don’t remember exact examples but I believe I’ve seen patched instructions themselves reference things like [rsp - (some large offset past 128)] so even with restoring the stack pointer, your saved data can get clobbered (or in the case of read, your saved data can clobber data used by the instruction).

OFFTKP: > What if the instruction references the stack pointer? Previously we would restore the stack pointer before executing the instruction.

Going to add an assert to make sure the destination register is not RSP, which is (hopefully) unlikely but possible

squidbus: I don’t think that will help with stack memory reads though.

OFFTKP: > I don’t think that will help with stack memory reads though.

Not sure what the best solution for that is. Perhaps modifying the displacement if the memory source uses RSP, to add +128?

squidbus: Yes although I don’t currently have a good solution (which is why I opted to just target Mac as it’s both trivial and the main platform that needed it). With BMI1 vs SSE4a though, BMI1 is much more common and even used by libc on boot so I’m more concerned with the robustness of that.

OFFTKP: Nevermind, I forgot those don't use gprs

OFFTKP: In that case, I could leave your functions alone and make separate ones that add 128 to the memory operand if it has RSP, to make sure rosetta has a robust solution and that x86 cpus without bmi1 have a solution that probably works

squidbus: Sure, I think you could just ifdef the contents of Save/RestoreStack with your decrement and increment rsp and that would work fine.

One other thing, as far as I know TZCNT doesn’t need emulation on Rosetta so that should be excluded from patch list for APPLE.

OFFTKP: I mentioned it in the discord but I'll paste it here too: TZCNT has the same encoding as BSF with an extra REP prefix, so on processors that don't have BMI1 (or Rosetta presumably) it will execute as a BSF so it won't throw an illegal instruction exception, and the result should be technically wrong if the source operand is zero. (This behavior is mentioned here: https://www.felixcloutier.com/x86/tzcnt)

That being said we can remove it if you want, idk if software generally relies on it but then again why would they use TZCNT over BSF, since TZCNT is one extra byte.

squidbus: I can check again when I get home but I believe I looked at the Rosetta 2 binary and it had TZCNT support.

OFFTKP: Now subtracting/adding on both save/restore and only for non-MacOS.

src/core/cpu_patches.cpp
```diff
@@ -202,21 +202,39 @@ static void RestoreStack(Xbyak::CodeGenerator& c) {
 
 /// Switches to the patch stack, saves registers, and restores the original stack.
 static void SaveRegisters(Xbyak::CodeGenerator& c, const std::initializer_list<Xbyak::Reg> regs) {
+    // Uses a more robust solution for saving registers on MacOS to avoid potential stack corruption
+    // if games decide to not follow the ABI and use the red zone.
+#ifdef __APPLE__
```
squidbus, line 207: What I meant here with the ifdefs is you can just put them inside SaveStack/RestoreStack itself around the contents, should save some duplication ifdef’ing every use like this.

src/core/cpu_patches.cpp
```diff
@@ -257,12 +275,11 @@ static void RestoreContext(Xbyak::CodeGenerator& c, const Xbyak::Operand& dst,
     RestoreStack(c);
 }
 
-#ifdef __APPLE__
-
 static void GenerateANDN(const ZydisDecodedOperand* operands, Xbyak::CodeGenerator& c) {
     const auto dst = ZydisToXbyakRegisterOperand(operands[0]);
     const auto src1 = ZydisToXbyakRegisterOperand(operands[1]);
     const auto src2 = ZydisToXbyakOperand(operands[2]);
+    ASSERT_MSG(dst.getIdx() != rsp.getIdx(), "ANDN overwriting the stack pointer");
```
squidbus, line 282: This should be fine for the Mac method, as we keep the saved registers stack elsewhere in TLS and can restore fine even when rsp is modified. Same for other instances. Perhaps could pass the dst along into the Save/Restore and let that determine if, for the current platform, the assert is needed.

PR #2530, opened 2025-02-24
OLED Theme

smiRaphi:
I did want to also make buttons completely black but then the empty slider space would just be invisible and I haven't found a way to change it :/ same with adding an outline/border to checkboxes

Screenshots:
image
image
image


Missake212:
Looks great overall but little problem I have is that the checkbox are hard to see when unchecked imo

image

smiRaphi:
I know, like I said, I'm trying to give them a border but I just can't get it to work and I don't know why

tomboylover93:
Text boxes are also pretty hard to see.
I could be wrong, but I think you should change either `QPalette::HighlightedText` or `QPalette::ToolTipBase` to Qt::white.
image

smiRaphi:
Oh I just found what I was doing wrong, I'll add borders to text boxes & checkboxes and maybe also make all buttons pitch black

smiRaphi:
I wasn't able to make all the buttons black because then I would have to include the entire style sheet for QSlider for it to look the same as the other themes, I also was only able to give checkboxes a border when they're unchecked because I would have to add custom checkmark images.

Missake212:
Too bad, looks good enough to me right now

Issue #2486, opened 2025-02-20
[GAME BUG]: Gravity Rush Remastered [CUSA01130] black screen and erratic camera glitch on Linux


birthtothunder:
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.
- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [x] I have disabled all patches and cheats and the issue is still present.
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadps4-game-compatibility?tab=readme-ov-file#informations) installed.

### Describe the Bug

Moving the camera while using the gravity ability causes it to move erratically. When you press R1 to disable the gravity ability, it causes a black screen yet the game is still running.

Demonstration:
https://streamable.com/xq8fe7

### Reproduction Steps

Not sure if this is reproducible beyond the tutorial
1. Use the gravity ability
2. Move camera 
3. Unuse the gravity ability

### Specify OS Version

Arch Linux

### CPU

Ryzen 5 5600x

### GPU

5700 Xt

### Amount of RAM in GB

16

### Amount of VRAM in GB

8

### Log File

[shad_log.txt](https://github.com/user-attachments/files/18888473/shad_log.txt)

whacka:
The fix seems to be described in the discussion here: 
https://github.com/shadps4-emu/shadps4-game-compatibility/issues/1742
If you are using a controller without gyro, turn the "Motion Sensor Sensitivity" all the way down in the settings menu.

PR #2532, opened 2025-02-24
Fix SDL gyro and acceleration sensor handling

kalaposfos13:
Fixes the blackscreen in GRR upon trying to float
The issue comes from the fact that for some reason enabling the gyro both works and throws an operation not permitted error, so the emulator assumes that the function is unavailable, so it logs the error, but it's still enabled nevertheless, and it gets read anyway, but when it's time to convert it to orientation values, the deltatime is still an uninitialized zero, which messes up the calculations and outputs garbage orientation data that's so out of expected ranges (the quaternion should be normalized to length 1, and this makes some coordinates to be in the thousands), that the camera position calculation in the game just gives up, and that's why the game turns black apart from the HUD

This has the side effect of orientation not being available to users who had this issue, but for Arch users, I found this "fix" (might work for other distros too):
```bash
sudo nano /etc/udev/rules.d/99-gamepad-motion.rules
```
Add the following text:
```ini
KERNEL=="event*", SUBSYSTEM=="input", MODE="0666"
```
Reload this config:
```bash
sudo udevadm control --reload-rules
sudo udevadm trigger
```

### DIFF
--- a/src/core/libraries/pad/pad.cpp
+++ b/src/core/libraries/pad/pad.cpp
@@ -315,6 +315,7 @@ int PS4_SYSV_ABI scePadRead(s32 handle, OrbisPadData* pData, s32 num) {
         pData[i].angularVelocity.x = states[i].angularVelocity.x;
         pData[i].angularVelocity.y = states[i].angularVelocity.y;
         pData[i].angularVelocity.z = states[i].angularVelocity.z;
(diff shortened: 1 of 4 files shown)


kalaposfos13:
Related issues:
https://github.com/shadps4-emu/shadps4-game-compatibility/issues/2680
https://github.com/shadps4-emu/shadps4-game-compatibility/issues/2511
https://github.com/shadps4-emu/shadPS4/issues/2486

PR #2540, opened 2025-02-25
Experimental: Shader cache

Fire-Cube:
First implementation of a cache for shaders. The SPIR-V translation result is now stored in a cache file. Skipping the translation can massively speed up compilation. IR should also be cached if possible. However, a way to implement this must first be found.


Xphalnos:
A question, From what I can see, the shader file takes the name of the git revision. Does this mean that the shader is obsolete with each new version of shadPS4? Knowing that shadPS4 receives several updates every day that do not necessarily affect shaders in general (like interface, translations, core, libraries...). Don't you think it is better to put a version number for the shader cache that is manually incremented with each change on the shaders part?

Fire-Cube:
> A question, From what I can see, the shader file takes the name of the git revision. Does this mean that the shader is obsolete with each new version of shadPS4? Knowing that shadPS4 receives several updates every day that do not necessarily affect shaders in general (like interface, core, libraries...). Don't you think it is better to put a version number for the shader cache that is manually incremented with each change on the shaders?

I can do this, good point

Xphalnos:
> > A question, From what I can see, the shader file takes the name of the git revision. Does this mean that the shader is obsolete with each new version of shadPS4? Knowing that shadPS4 receives several updates every day that do not necessarily affect shaders in general (like interface, core, libraries...). Don't you think it is better to put a version number for the shader cache that is manually incremented with each change on the shaders?
> 
> I can do this, good point

Don't forget clang-format. Otherwise good job!

src/emulator.cpp
```diff
@@ -248,6 +248,11 @@ void Emulator::Run(const std::filesystem::path& file, const std::vector<std::str
     }
     VideoCore::SetOutputDir(mount_captures_dir, id);
 
+    const auto shader_cache_dir = Common::FS::GetUserPath(Common::FS::PathType::ShaderDir) / "cache";
```
raphaelthegreat, line 254: This should probably be done in pipeline cache constructor

raphaelthegreat (commented):
I think a more proper cache implementation should avoid having to generate IR first as that's quite expensive step. Also having the build be part of cache is not great as any update will render it outdated. The cache should have a file format with header, identier and sizes of the spirv binaries contained.
Ideally the graphics pipelines should def be cached too as a shader is compiled most likely on pipeline creation and not with modules.
If the end goal is to reduce stutter vulkan extensions like pipeline libraries and shader objects should be considered

baggins183:
perm_idx will be inconsistent between runs so youd have to account for the specialization info itself

georgemoralis:
since that was a test that didn't go that well , i am closing it . Let's hope next try will be more complete :)

PR #2246, opened 2025-01-26
fs: handle relative paths

panzone91:
Some games (like M2 Cave portings) uses relative paths for module loading. This PR introduce a simple way to handle these type of paths.

PR #2541, opened 2025-02-25
libkernel: improve module finding in sceKernelLoadStartModule

panzone91:
This PR changes sceKernelLoadStartModule to follow its implementation in a more precise manner. It's based on the comment on a previous MR (https://github.com/shadps4-emu/shadPS4/pull/2246#issuecomment-2621075627)

This also makes Battle Garegga (https://github.com/shadps4-emu/shadps4-game-compatibility/issues/398) playable.

### DIFF
--- a/src/core/libraries/kernel/process.cpp
+++ b/src/core/libraries/kernel/process.cpp
@@ -47,23 +47,12 @@ s32 PS4_SYSV_ABI sceKernelLoadStartModule(const char* moduleFileName, size_t arg
     s32 handle;
 
     if (guest_path[0] == '/') {
-        path = mnt->GetHostPath("/system/common/lib" + guest_path);
-        handle = linker->LoadAndStartModule(path, args, argp, pRes);
-        if (handle != -1) {
-            return handle;
-        }
-
-        path = mnt->GetHostPath("/system/priv/lib" + guest_path);
-        handle = linker->LoadAndStartModule(path, args, argp, pRes);
-        if (handle != -1) {
-            return handle;
-        }
-
         path = mnt->GetHostPath(guest_path);
         handle = linker->LoadAndStartModule(path, args, argp, pRes);
         if (handle != -1) {
             return handle;
         }
+        ASSERT_MSG(0, "Trying to load system module {}", guest_path);
     } else {
         auto* process_params = linker->GetProcParam();
         if (process_params->sdk_version > 0x3ffffff) {
@@ -72,11 +61,7 @@ s32 PS4_SYSV_ABI sceKernelLoadStartModule(const char* moduleFileName, size_t arg
                 if (guest_path.contains("libScePigletv2VSH") ||
                     guest_path.contains("libSceSysCore") ||
                     guest_path.contains("libSceVideoCoreServerInterface")) {
-                    path = mnt->GetHostPath(guest_path);
-                    handle = linker->LoadAndStartModule(path, args, argp, pRes);
-                    if (handle != -1) {
-                        return handle;
-                    }
+                    ASSERT_MSG(0, "Trying to load system module {}", guest_path);
                 } else {
                     // loading prohibited
                     return ORBIS_KERNEL_ERROR_ENOENT;
@@ -93,17 +78,7 @@ s32 PS4_SYSV_ABI sceKernelLoadStartModule(const char* moduleFileName, size_t arg
                 return handle;
             }
             if ((flags & 0x10000) != 0) {
-                path = mnt->GetHostPath("/system/priv/lib/" + guest_path);
-                handle = linker->LoadAndStartModule(path, args, argp, pRes);
-                if (handle != -1) {
-                    return handle;
-                }
-
-                path = mnt->GetHostPath("/system/common/lib" + guest_path);
-                handle = linker->LoadAndStartModule(path, args, argp, pRes);
-                if (handle != -1) {
-                    return handle;
-                }
+                ASSERT_MSG(0, "Trying to load system module {}", guest_path);
             }
         } else {
             path = mnt->GetHostPath(guest_path);
--- a/src/core/libraries/kernel/process.cpp
+++ b/src/core/libraries/kernel/process.cpp
@@ -36,9 +36,7 @@ s32 PS4_SYSV_ABI sceKernelLoadStartModule(const char* moduleFileName, size_t arg
                                           u32 flags, const void* pOpt, int* pRes) {
     LOG_INFO(Lib_Kernel, "called filename = {}, args = {}", moduleFileName, args);
 
-    if (flags != 0) {
-        return ORBIS_KERNEL_ERROR_EINVAL;
-    }
+    ASSERT(flags == 0);
 
     std::string guest_path(moduleFileName);
     auto* mnt = Common::Singleton<Core::FileSys::MntPoints>::Instance();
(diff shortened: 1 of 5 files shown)


StevenMiller123:
I don't think it's a good idea to support loading modules from `system/common/lib` or `system/priv/lib` yet. Most of the libraries in those firmware locations can't be loaded LLE yet, and will cause cause crashes if (for some reason) a game chooses to try loading them.
Maybe add an assert for those cases instead?

StevenMiller123:
Did a quick test, this does seem to work fine in Unity games at least.
[shad_log.txt](https://github.com/user-attachments/files/18995305/shad_log.txt)

src/core/libraries/kernel/process.cpp
```diff
@@ -40,36 +40,55 @@ s32 PS4_SYSV_ABI sceKernelLoadStartModule(const char* moduleFileName, size_t arg
...
+    std::filesystem::path path;
+    s32 handle;
+
+    if (guest_path[0] == '/') {
+        path = mnt->GetHostPath(guest_path);
+        handle = linker->LoadAndStartModule(path, args, argp, pRes);
+        if (handle != -1) {
+            return handle;
+        }
+        ASSERT_MSG(0, "Trying to load system module {}", guest_path);
```
polybiusproxy:
```diff
- ASSERT_MSG(0, "Trying to load system module {}", guest_path);
+ UNREACHABLE();
```

src/core/libraries/kernel/process.cpp
```diff
@@ -40,36 +40,55 @@ s32 PS4_SYSV_ABI sceKernelLoadStartModule(const char* moduleFileName, size_t arg
...
+        }
+        if (!guest_path.contains('/')) {
+            path = mnt->GetHostPath("/app0/" + guest_path);
+            handle = linker->LoadAndStartModule(path, args, argp, pRes);
+            if (handle != -1) {
+                return handle;
+            }
+            if ((flags & 0x10000) != 0) {
+                ASSERT_MSG(0, "Trying to load system module {}", guest_path);
+            }
```
polybiusproxy: ~~`ASSERT(!(flags & 0x10000));`~~

src/core/libraries/kernel/process.cpp
```diff
@@ -40,36 +40,55 @@ s32 PS4_SYSV_ABI sceKernelLoadStartModule(const char* moduleFileName, size_t arg
...
+                    ASSERT_MSG(0, "Trying to load system module {}", guest_path);
+                } else {
+                    // loading prohibited
+                    return ORBIS_KERNEL_ERROR_ENOENT;
+                }
+            } else {
+                // loading prohibited
+                return ORBIS_KERNEL_ERROR_ENOENT;
+            }
+        }
```
polybiusproxy: At this moment I think it's unnecessary to check for system modules, so I'd just remove this.

polybiusproxy (changes requested):

src/core/libraries/kernel/process.cpp
polybiusproxy: Scratch the above. Just add an assert at the start of the function: `ASSERT(flags == 0);` and remove the existing check that returns EINVAL.

polybiusproxy (changes requested):
The only thing left is a bit of clean-up, duplicated code isn't good. You can have something like this:
```cpp
s32 PS4_SYSV_ABI sceKernelLoadStartModule(const char* moduleFileName, u64 args, const void* argp,
                                          u32 flags, const void* pOpt, int* pRes) {
    LOG_INFO(Lib_Kernel, "called filename = {}, args = {}", moduleFileName, args);
    ASSERT(flags == 0);

    auto* mnt = Common::Singleton<Core::FileSys::MntPoints>::Instance();
    auto* linker = Common::Singleton<Core::Linker>::Instance();
    
    std::filesystem::path path;
    const std::string guest_path(moduleFileName);

    const bool is_root = guest_path[0] == '/';
    if (is_root) { /* Root path */
        path = mnt->GetHostPath(guest_path);
    } else if (!guest_path.contains('/')) { /* Relative path */
        path = mnt->GetHostPath("/app0/" + guest_path);
    } else { /* Absolute path */
        path = mnt->GetHostPath(guest_path);
    }

    if (const s32 handle = linker->LoadAndStartModule(path, args, argp, pRes); handle != -1) {
        return handle;
    }

    if (is_root) {
        UNREACHABLE();
    }

    return ORBIS_KERNEL_ERROR_ENOENT;
}
```

I'd also appreciate if you could replace the use of `size_t` to the `u64` type.

src/core/linker.h
```diff
@@ -144,6 +144,8 @@ class Linker {
     void FreeTlsForNonPrimaryThread(void* pointer);
 
     s32 LoadModule(const std::filesystem::path& elf_name, bool is_dynamic = false);
+    s32 LoadAndStartModule(const std::filesystem::path& path, size_t args, const void* argp,
+                           int* pRes);
```
polybiusproxy: Change the type here as well, same for the implementation

polybiusproxy (changes requested):

polybiusproxy:
This looks good at the moment, can you try it out @StevenMiller123?

StevenMiller123:
Seems like Unity games don't work with the current version of this PR.
[shad_log.txt](https://github.com/user-attachments/files/19144285/shad_log.txt)

polybiusproxy:
> Seems like Unity games don't work with the current version of this PR. [shad_log.txt](https://github.com/user-attachments/files/19144285/shad_log.txt)

Not sure why is it failing, would you be able to debug it?

StevenMiller123:
@panzone91 @polybiusproxy Seems like the issue I was hitting was just because I was testing on an action. I was able to boot the game I'm testing fine after compiling this branch locally.
image
[shad_log.txt](https://github.com/user-attachments/files/19145182/shad_log.txt)

src/core/libraries/kernel/process.cpp
```diff
@@ -32,44 +32,33 @@ void* PS4_SYSV_ABI sceKernelGetProcParam() {
...
-    }
-    handle = linker->LoadModule(path, true);
-    if (handle == -1) {
-        return ORBIS_KERNEL_ERROR_ESRCH;
+
+    std::filesystem::path path;
+    std::string guest_path(moduleFileName);
+
+    const bool is_root = guest_path[0] == '/';
+    if (is_root || guest_path.contains('/')) {
```
wojek07, line 47: I wonder how that logic would behave with paths like "lib/module.sprx" (which is clearly relative and as such probably should be prefixed by "/app0/"). From what I see, given example could be processed as absolute path, which may not be desirable.

panzone91: Is that a legal path however? From what we've seen in the actual implementation it should not be possible. Do you have some examples where there is something similar?

red-prig: The search algorithm that I described in that post is based on kernel decompilation, yes, this is all a rather unstable construction and this is exactly how it works.

wojek07: Unfortunately, I do not have any evidence of this occurring in the wild, so feel free to treat it as just speculation.

red-prig: I've actually seen games that load from subfolders, but they do something like: "/app0/lib/module.sprx"

polybiusproxy (approved):

PR #2544, opened 2025-02-27
Qt: Add Initial KBM remapping GUI

rainmakerv3:
Adds a full GUI for KBM remapping. Think this should be tested a lot first, there's a lot of code/keys so high potential for error

Limitation:
Cannot map right shift / alt /ctrl / meta - QT only has one key event for some reason for each modifier, so these keys will always map to the left since that is probably the side used by the vast majority

I am not sure if this limitation can be worked around, Dolphin has the same limitation as well with a (I think) similar implementation.

Multiple inputs to an output is not supported, and multiple key inputs other than modifier + key. I believe it is possible to have these, but some more work is needed so maybe in the future this could be added.

image


ghost:
its just an idea but what about making the number row show only numbers? like i dont think any person would want to use ! as a key input, and the samke things with the keys on the right like [ ] ' ; , . /

 and caps lock can be CapsLK, since some keyboards do that to reduce the text size,

GHU7924:
I haven't tried this PR yet, but I'll make my observations:

1) @rainmakerv3 you did well))
2) The letters (labels) on the buttons should be... w -> W
3) Rename the buttons in the "Face Buttons" section to `Num*`, and since we have alternative keys, they should also be added and separated by slashes, something like this - `Num8 / C`, `Num4 / V`, `Num2 / N`, `Num6 / B`.

 Is it possible for the label on the button to change depending on the keystroke?
I mean, it was `Num2 / N`, and I changed `N -> X`, and the label became `Num2 / X`.

And I think it might be worth moving the alternative keys to the left (select `Z, X, C, V`), in my opinion, it will be more convenient.

4) The field with the "Text Editor" and "Help" buttons should probably be aligned to the level of the fields on the left and right.
5) Keyboard image... It doesn't look very good, but if you need a smaller image, you can do something like this as a joke:
keyboard
6) Delete all labels (def 0.***) in the "Mouse Movement Options" and add the "Reset" button under the sliders to change the values to the default ones.

ghost:
here is a hot take, why does it need to have a keyboard in the middle actually? like is there any reason to show the keyboard to the user aside from making it known that you are using the keyboard and mouse as the input method? if you guys think its important to have the keyboard  how about using something super simple like this ?
https://www.freepik.com/free-vector/keyboard-realistic_35202474.htm#fromView=search&page=1&position=1&uuid=43711cc7-809f-4f0b-8cf5-b655ef318c87&query=Keyboard

ghost:
and while at it i want to add a few things how about having a little space to the left side of the texts?
image
i feel rpcs3 did it really well in here.

and while at it i feel these boxes can be a little smaller 
image
again i feel rpcs3 did it great .

im really sorry if i sound like i am nit picking and being annoying and such i just want this emulator to look great hope i am not being pushy and such

rainmakerv3:
The format of the numbers right now matches the syntax needed by the config file, so if I change them (making the letters uppercase, changing kp to num, etc) I would have to create a mapping table for all keys, convert them on loading, and convert them back on saving. So unless there's a large benefit to doing that, maybe we can leave that for the future, as this PR was already a very large amount of work and I kind of want to take a break also.

I do think it's nice to have some sort of keyboard image in the middle. The interface is kind of similar to the controller one so it's best it users can immediately see if they're on the right screen. If you have an image we can use to replace the existing one, feel free to post it here and we can see how it looks with the interface.

The button width was lengthened to fit the longest possible mapping name (shift + mousewheeldown), anything narrower cuts off text for that

As mentioned in the limitations, the GUI cannot current handle alternate keys (short of hardcoding them, which I don't think is ideal), so this unfortunately cannot be used with the GUI as currently designed. After a break, maybe I can look into this some more.

PR #2550, opened 2025-02-27
Support for multiple controllers and players

kalaposfos13:
Edit: this description is outdated, don't take everything in it as the current state of this PR.

Contains:
- replacing a whole lot of things that were just a global with a 4 long array global
- implementing sceUserServiceGetEvent to handle users connecting/disconnecting (only half done, currently you need to specify the number of players from config
- Even more shit added to remapping, this time a new field denoting the controller ID for the input or output (for inputs it only concerns gamepads, as those are the only ones that you can connect more than one of)
- reverts ngoquang's engine, as it's currently unused, and I didn't want to deal with yet another system that had to be updated
- Non-remappable controls now work correctly

What's missing yet:
- Handling connecting and disconnecting controllers properly
- Compatibility with rainmaker's controller selection code

How to use: 
- If you use controllers, all of them will default to the mapping that's set in the remap file
- If you use the keyboard, you can set inputs to go to a certain controller by adding ` : <controllerNumber>` after the output. Example:
```
axis_left_x_minus = a
axis_left_x_plus = d
axis_left_y_minus = w
axis_left_y_plus = s

axis_left_x_minus : 2 = j
axis_left_x_plus : 2 = l
axis_left_y_minus : 2  = i
axis_left_y_plus : 2 = k
```


kalaposfos13:
True, I forgot to add the fact that I have written no GUI for this to the description, my bad
I will probably leave the GUI to someone else though, I'm more interested in writing the backend for this, and that isn't even finished yet, as there are still some bugs, and there are things that have to be done manually, like specifying the amount of players for it to work.

GHU7924:
>I'm more interested in writing the backend for this, and that isn't even finished yet, as there are still some bugs, and there are things that have to be done manually, like specifying the amount of players for it to work.

Unfortunately, the only game for multiple players I have right now is `Beach Buggy Racing 2: Island Adventure`. I will test it anyway and tell you about the results.

@Missake212 Can you help? Do you have any working games where you can play at least two people? Fighting games or platformers? We need games with split screen or where both players fit on one screen. Since the compatibility database was reset, it's hard to say, but maybe Cuphead or Lego Games?

To test this PR, you need at least one gamepad and a keyboard, if you have two gamepads, that would be great too.

Missake212:
@GHU7924 The only game I have that supports multiple players is Minecraft, and I believe I already tested and sent my results to @kalaposfos13 , I don't think there has been any new commits since then (but I could be mistaking) I have a lot of controllers so I'll be glad to test again if I see a new commit

GHU7924:
@kalaposfos13 Unfortunately I couldn't test the game with the assistant today, maybe in a couple of days I will.

I thought I only had one game for multiple players, but then I remembered that I also have 
`Teenage Mutant Ninja Turtles: Shredder's Revenge`, which if you remember didn't have working controls (so I decided to remove it from my playlist until better times).

I remember that this game could be played with multiple players, but running it on the "main" didn't seem to change anything, so to be honest I didn't have high hopes for the changes in your PR, but it turns out there are changes, and in the latest build from your PR this game throws an error:

>[Lib.UserService] <Info> userservice.cpp:127 sceUserServiceGetEvent: Event processed by the game: 0 1
[Tty] <Info> logger.cpp:62 log_flush: [stdout] ERROR: Couldn't assign user 1 to player slot!
[Debug] <Critical> controller.h:103 operator[]: Unreachable code!
Index out of bounds for GameControllers!

@Missake212 Okay, I understand you. Thank you for answering.

GHU7924:
I reread your first post and realized my mistakes in testing the PR, so I did a few more retests.

I continued testing on `Beach Buggy Racing 2: Island Adventure`.
For the test I had 4 devices: 2 Xbox controllers and 2 Keyboards.

I set the value `numberOfPlayers = 4`.
Since it was necessary to add other devices, I did everything through your GUI.

_Note: As it turns out, you only need to have one configuration written down for one controller to make all controllers work. I had two controllers and they worked. (I didn't read carefully at first, that's why I had problems with the gamepads working)_

Problems I found while testing PR:

1) I don't know why, but this build didn't read the settings from CUSA26789.ini, but it read the settings from default.ini perfectly well
2) At the moment, in PR, apparently, it is necessary that there be a certain order of devices.

Let's say you have 4 devices (2 gamepads and 2 keyboards).

Gamepads must come first, then keyboards. This is the sequence.

If you specify the keyboard as the second device (` : 2 `), then the second gamepad will NOT be automatically considered the third device, it will also be detected as the second device, thanks to which the second player will be able to control both from the second gamepad and from the keyboard. But when you specify the keyboard as the 3rd device, then two gamepads will work fine, and the keyboard will also work as the 3rd player.

3) If everything is fine with the gamepads and they work with one layout, then there is a problem with the keyboards.
I specified one keyboard as the 3rd player (` : 3 `), the second keyboard as the 4th player (` : 4 `),
but unlike the gamepads, with the same layout they can only work as one device.

What is written below does not work, the devices work as one, not two different ones.

Keyboard bindings (Player 3)

>triangle : 3 = i
circle : 3 = l
cross : 3 = k
square : 3 = j

Keyboard bindings (Player 4)

>triangle : 4 = i
circle : 4 = l
cross : 4 = k
square : 4 = j

@kalaposfos13 In order for you to understand this problem, you need to have two connected keyboards with the same controls and conduct tests until they start working separately with the same controls.

_Note: If you specify completely different keys for the keyboards (without repeating at all), then, it seems, you can play normally from two keyboards._

#

Next

There were some performance issues, the frame rate dropped from 60 to 30 for a while and there were also some graphical glitches (I don't know if they are related to this PR, maybe it's a port issue or something like that), for example, when driving, a dust effect appears and for one player everything may look normal, but for another, some kind of interlaced scanning effect appears:




Otherwise, everything seems to be working fine.

Jrguittar:
I would like to test PR with Saban's might morfin power rangers for 4 players on joystick

GHU7924:
@kalaposfos13 I haven't tested the latest build, but I downloaded it to make sure of one thing.
Update the help! Add a new section "Support for multiple controllers and players", where there will be a description of how it works or how to make it work.
Also, you added new keys "F4/F5", but did not include them in the list of emulator keys, and there is no description of their operation.

I understand that you first need approval for this PR, so you can make changes to the help later.
What needs to be added... 
Do I need to add a line to the config file, or will it already be there with some default value,
do you need something if you only have gamepads, how to configure the keyboard, and so on,
otherwise, users who encounter this for the first time will probably be forced to do fortune telling on coffee grounds.

src/common/config.cpp
```diff
@@ -75,6 +77,7 @@ static double trophyNotificationDuration = 6.0;
 static bool useUnifiedInputConfig = true;
 static bool overrideControllerColor = false;
 static int controllerCustomColorRGB[3] = {0, 0, 255};
+static bool separateupdatefolder = false;
```
georgemoralis: ??

src/core/libraries/pad/pad.cpp
```diff
@@ -98,25 +96,25 @@ int PS4_SYSV_ABI scePadGetControllerInformation(s32 handle, OrbisPadControllerIn
         pInfo->stickInfo.deadZoneLeft = 1;
         pInfo->stickInfo.deadZoneRight = 1;
         pInfo->connectionType = ORBIS_PAD_PORT_TYPE_STANDARD;
-        pInfo->connectedCount = 1;
+        pInfo->connectedCount = 2;
```
georgemoralis: that should be 1-4 probably by an enabled switch

src/core/libraries/pad/pad.cpp
```diff
@@ -98,25 +96,25 @@ int PS4_SYSV_ABI scePadGetControllerInformation(s32 handle, OrbisPadControllerIn
...
+        return 0;
     }
     pInfo->touchPadInfo.pixelDensity = 1;
     pInfo->touchPadInfo.resolution.x = 1920;
     pInfo->touchPadInfo.resolution.y = 950;
     pInfo->stickInfo.deadZoneLeft = 1;
     pInfo->stickInfo.deadZoneRight = 1;
     pInfo->connectionType = ORBIS_PAD_PORT_TYPE_STANDARD;
-    pInfo->connectedCount = 1;
+    pInfo->connectedCount = 2;
```
georgemoralis: similar to above

src/core/libraries/system/userservice.cpp
```diff
@@ -573,11 +586,10 @@ s32 PS4_SYSV_ABI sceUserServiceGetLoginUserIdList(OrbisUserServiceLoginUserIdLis
...
     }
     // TODO only first user, do the others as well
-    userIdList->user_id[0] = 1;
-    userIdList->user_id[1] = ORBIS_USER_SERVICE_USER_ID_INVALID;
-    userIdList->user_id[2] = ORBIS_USER_SERVICE_USER_ID_INVALID;
-    userIdList->user_id[3] = ORBIS_USER_SERVICE_USER_ID_INVALID;
-
+    for (int i = 0; i < 4; i++) {
+        auto controllers = *Common::Singleton<Input::GameControllers>::Instance();
+        userIdList->user_id[i] = controllers[i]->user_id;
```
georgemoralis: should only init enabled controllers

kalaposfos13: Not enabled controllers have a user_id of -1 aka ORBIS_USER_SERVICE_USER_ID_INVALID, so it's fine like this

src/core/libraries/pad/pad.cpp
StevenMiller123: Still a valid comment, this shouldn't be constant.

src/core/libraries/pad/pad.cpp
StevenMiller123: Same as above.

src/core/libraries/pad/pad.cpp
```diff
@@ -290,12 +290,12 @@ int PS4_SYSV_ABI scePadOutputReport() {
 
 int PS4_SYSV_ABI scePadRead(s32 handle, OrbisPadData* pData, s32 num) {
     LOG_TRACE(Lib_Pad, "called");
+    LOG_DEBUG(Lib_Pad, "handle: {}", handle);
```
StevenMiller123: This should probably be one log call.

src/core/libraries/pad/pad.cpp
```diff
@@ -290,12 +290,12 @@ int PS4_SYSV_ABI scePadOutputReport() {
...
     int connected_count = 0;
     bool connected = false;
     Input::State states[64];
-    auto* controller = Common::Singleton<GameController>::Instance();
-    const auto* engine = controller->GetEngine();
-    int ret_num = controller->ReadStates(states, num, &connected, &connected_count);
+    auto controllers = *Common::Singleton<Input::GameControllers>::Instance();
+    int ret_num = controllers[handle - 1]->ReadStates(states, num, &connected, &connected_count);
```
StevenMiller123: You should make sure handle is in the expected bounds here.

src/core/libraries/pad/pad.cpp
```diff
@@ -363,15 +357,15 @@ int PS4_SYSV_ABI scePadReadHistory() {
 
 int PS4_SYSV_ABI scePadReadState(s32 handle, OrbisPadData* pData) {
     LOG_TRACE(Lib_Pad, "called");
+    LOG_DEBUG(Lib_Pad, "handle: {}", handle);
```
StevenMiller123: Same as above

src/core/libraries/pad/pad.cpp
```diff
@@ -363,15 +357,15 @@ int PS4_SYSV_ABI scePadReadHistory() {
...
         return ORBIS_PAD_ERROR_INVALID_HANDLE;
     }
-    auto* controller = Common::Singleton<GameController>::Instance();
-    const auto* engine = controller->GetEngine();
+    auto controllers = *Common::Singleton<Input::GameControllers>::Instance();
     int connectedCount = 0;
     bool isConnected = false;
     Input::State state;
-    controller->ReadState(&state, &isConnected, &connectedCount);
+    controllers[handle - 1]->ReadState(&state, &isConnected, &connectedCount);
```
StevenMiller123: Same as above

StevenMiller123 (changes requested):
A couple minor nitpicks here. I've only looked at the library code so far.

src/core/libraries/pad/pad.cpp
kalaposfos13: Due to how handles are handled, this will never be out of bounds, barring the possibility of cosmic gamma rays flipping a bit in memory.

StevenMiller123: Handle is an input from the game. It's entirely possible that I write a homebrew that passes in 0 as a handle, thus proving your statement wrong.

GHU7924:
@kalaposfos13 I hope that you and I are still friends and that you won’t tear your hair out because of me 😄.
I already forgot that it was necessary to test the latest build of this PR. 

But first, some questions:

1) But either I don’t understand something, or you’ve made it too complicated.
If there is `F4` and `F5`, do we now need to specify the quantity `[Input] numberOfPlayers` in `config.toml`?

Or would it be correct to immediately specify the maximum number of players (4) by default, since we have the ability to add controllers or remove them.

I just realized that now the emulator distinguishes physical devices from virtual ones.
In `config.toml` I had the value = 3 players set. I used 2 physical gamepads and a keyboard. The keyboard ( : 3 ) is set in default.ini, but until I pressed `F5` and `options` (options : 3 = enter) it was not added, but the games understood that there were 2 gamepads and it was easy to connect the second player by pressing one button in the game itself.

2) I didn't quite understand about `F4` and `F5`.

Is it necessary to delete all controllers? Maybe a limitation is needed? 
When deleting, you can delete all controllers except the main one.
When adding, you can add a maximum of 3 more controllers, because there is already a main one.

Or is there a good reason why it works this way?

3) I still think that a new section "Support for multiple controllers and players" should be added to the help, where it will be stated once again (even if it is written again) that by default there will be so many players, controllers can be added or removed in such a way that this is necessary to configure the keyboard, the keyboard is a virtual device and should come after the physical ones, so that the controls are not confused.

To reduce the number of questions users have, I think such a separate help section will help.

---

Now to the tests of the latest build at the moment. Build 0a58c71

`Beach Buggy Racing 2: Island Adventure` Controllers are added and removed, everything seems to be fine in the game. I don't know if everything is controlled correctly, because I tested it alone.

In fact, I noticed that in Main I got `Teenage Mutant Ninja Turtles: Shredder's Revenge` (1.00) working, let me remind you that the controls didn't work in it. Now that the controls work, I was able to test this PR. Everything seems to be fine too. Here are 3 players:

TMNT SR_2

kalaposfos13:
F4 and F5 are not needed now unless you want to have more controllers emulated than you have connected
numberOfPlayers isn't used anymore at all, it is now all dynamic based on the actual number of controllers connected
Yes, this means that multiple controllers emulated from only a keyboard will now be a bit more manual to set up every time, but since a single keyboard and mouse isn't really the best setup for more than one player so no one will use that anyway, I simply don't care.

GHU7924:
@kalaposfos13 Okay, I get it.

You should update (add) information to your first comment in this PR so that it doesn't mislead new testers if any come along.

kyle788:
What is left to do in this PR before it gets merged? Anxious to have it in main.

dips2queue:
Everything appears to be working as intended, but the back-button binding isn’t functioning with the touchpad. It does work with the other buttons, though. Really good work.

GHU7924:
@kalaposfos13 You should update (add) information to your first comment in this PR so that it doesn't mislead new testers if any come along.

kalaposfos13:
> [...] but the back-button binding isn’t functioning with the touchpad.

This should be fixed now.

kalaposfos13:
The PR is now moved to #3832.

dips2queue:
> The PR is now moved to #3832.

Thanks, that worked perfectly! I’m really looking forward to it. I have another question—would it be possible to navigate the game list using the gamepad?

PR #2557, opened 2025-02-28
Add unexcepted depth format to PromoteFormatToDepth unreachable

Randomuser8219:
Demonstration on PC Building Simulator (CUSA14566):
image


gandalfthewhite19890404:
I should create separate issue for earch unexpected format? Or better to use aggregate?
Or report here those formats?

Randomuser8219:
Yes

Issue #2552, opened 2025-02-27
[Feature Request]: Right click on a game to directly get to the log folder of this game


Missake212:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

After  #2504 we can now have a log for each game, which is great, but I think what would be better is to have access through the emulator to each log by right clicking the game in question and arriving in a folder where your game log is. You'd have to generate a folder for each game thought which could be pretty annoying.

### Reason

This would be really useful when testing games and reporting about them, instead of testing them all then matching the CUSA with the right game all you would have to do is right click on the game in question inside the emulator, then click on open folder and open log folder to get the game logs and drop it directly into GitHub or Discord for example.

### Examples

Basically instead of having this current implementation:

![Image](https://github.com/user-attachments/assets/7e2fc2a7-8a21-4d34-b5b7-22174a71fe3f)

We'd have something like this:

![Image](https://github.com/user-attachments/assets/418af046-3d82-4933-bdb5-b2fe02de042e)

And right clicking directly in the emulator would lead you directly inside the game's folder.

![Image](https://github.com/user-attachments/assets/9dc04086-b26f-4343-9783-695ee5973b72)

imnltsa:
Each game would still only have one log, right? It seems kind of unnecessary to put them into their own folders unless maybe logs were retained renaming them based off of the timestamp when the game was played.

Could also just add a "Open in Notepad" button removing the need for separate directories, though that assumes people have Notepad installed, which is pretty likely.

Missake212:
> Each game would still only have one log, right? It seems kind of unnecessary to put them into their own folders unless maybe logs were retained renaming them based off of the timestamp when the game was played.
> 
> Could also just add a "Open in Notepad" button removing the need for separate directories, though that assumes people have Notepad installed, which is pretty likely.

There's an open PR ( #2464 ) to retain a backup log, I believe it could be possible to have in the same folder the last log and the one before that. Obviously having one folder for each game seems like unnecessary but making it an available option would be nice I believe. Currently if you don't enable the "Separate log file" you only get one log anyway so you wouldn't need that extra folder

imnltsa:
Instead of opening an isolated directory containing a single log, try "selecting" the log files.

Windows is able to natively "select" a file through the File Explorer executable with the [`/select`](https://stackoverflow.com/a/13680458) argument.

[...]

PR #2560, opened 2025-02-28
'Select' the log when opening the folder if SeparateLogFiles

DanielSvoboda:
issues: https://github.com/shadps4-emu/shadPS4/issues/2552

- When SeparateLogFiles is disabled:
It will only open the Log folder

- When SeparateLogFiles is enabled:
It will open the Log folder and 'select' the file.
image
If the file for the specific game does not exist, it will display this message:
image


GHU7924:
>However, I think that at the moment option "Open Log Folder" is not very conveniently located and it is worth, so to speak, raising it to a higher level, that is it needs to be removed from the "Open folder..." and simply placed as a new item in the context menu:
Image

Or do you think it's not worth it?

DanielSvoboda:
> Or do you think it's not worth it?

'Log Files' is an open folder button, I see no reason to remove it from 'Open Folder'. I think it's more organized the way it is

PR #2539, opened 2025-02-25
Custom Trophy images / sound | and improvements

DanielSvoboda:
issues: https://github.com/shadps4-emu/shadPS4/issues/2537 https://github.com/shadps4-emu/shadPS4/issues/2461

- It is possible to add custom images to the trophies and a sound.

Add the custom files to this folder : `\user\custom_trophy`
*There is a button in the settings to get to this folder more easily.
The files must have these names:
```
trophy.mp3

bronze.png
gold.png
platinum.png
silver.png
```

*I recommend a square resolution image, for example 200x200, 500x500, the same size as the height and width

- An animation has been added to notifications
- Side menu with filters:
-Show Earned Trophies
-Show Not Earned Trophies
-Show Hidden Trophies

- Added "Time Unlocked" column
The date you achieved the achievement will only work from that PR,
what was achieved before will unfortunately remain as 'unknown'

- Button - Delete Trophy

----

Folder with customized files:
image

Screen with new options on the left side
image

Time Unlocked
image

Screen with new settings
image

My horn audio that I added when I won a trophy, by default, doesn't even have any audio.

https://github.com/user-attachments/assets/13a5d65d-ec91-4456-962d-cf33491c139f

Left and right position, it is possible to change in the settings

https://github.com/user-attachments/assets/6cfa8802-c271-4488-b5d4-28ea6644f5c0

This was the old screen:
image

PR #2566, opened 2025-02-28
Adding Top and Bottom trophy option for pop window + Trophy improvements 

diegolix29:
I always wanted the pop up to come from top as it will never overlap anything its more like a personal preference so feel free not to merge it if its unnecessary.

seems to be working ok on my end. 

image

https://github.com/user-attachments/assets/b463a3fc-1230-402a-a9cb-2bf4a8a5ca6b


Hermiten:
Seem's good, from what I see on the video :)

diegolix29:
> Seem's good, from what I see on the video :)

Thanks i really liked the new custom trophy so with this im thrilled haha

GHU7924:
Then you should add the "bottom" option 😄 .

In my opinion, if possible, then when selecting the "top" and "bottom" options, the text should be positioned in the center of the notification.

It looks really good.

diegolix29:
> Then you should add the "bottom" option 😄 .
> 
> In my opinion, if possible, then when selecting the "top" and "bottom" options, the text should be positioned in the center of the notification.
> 
> It looks really good.

thanks, and sure haha i can add the bottom option too. Also will center the text then thanks for the insight.

diegolix29:
OK so with @DanielSvoboda finished up the requests i think it all should work ok now 

image

https://github.com/user-attachments/assets/68e14c00-ff34-41bf-8f8c-681e61bca1bf

GHU7924:
Can't the achievement text be centered for the "Top" and "Bottom" options?

And it also seems that no one knows how to separate the text "Trophy earned" and the text of the achievement itself so that they can be set to different font sizes.  The text "Trophy earned" if it was a bigger size, it would be great.

diegolix29:
i have centered the text but making the trophy earned line bigger its tricky

diegolix29:
Seems like imgui dosent support mixed font sizes on a single rendered text, so that may take a bit more at least on my end to achive it

diegolix29:
I tried to separate the text to be render separately and make one bigger but it mess up the other one and dosent properly renders after the separation. Imma do some internal testing today to see if i can achieve it

AhayriSG:
Is the implementation of custom sound files for notifications feasible at this point?

GHU7924:
@diegolix29 I wanted to look at the changes in the trophy viewer, but I have a problem:
изображение

>Seems like imgui dosent support mixed font sizes on a single rendered text, so that may take a bit more at least on my end to achive it
I tried to separate the text to be render separately and make one bigger but it mess up the other one and dosent properly renders after the separation. Imma do some internal testing today to see if i can achieve it

It's not urgent, but it would be a good improvement.

diegolix29:
> @diegolix29 I wanted to look at the changes in the trophy viewer, but I have a problem: изображение
> 
> > Seems like imgui dosent support mixed font sizes on a single rendered text, so that may take a bit more at least on my end to achive it
> > I tried to separate the text to be render separately and make one bigger but it mess up the other one and dosent properly renders after the separation. Imma do some internal testing today to see if i can achieve it
> 
> It's not urgent, but it would be a good improvement.

i think now u need ur trophy key to unlock them, if u have ur files from before the pr that introduced the trophy key u still can access ur trophy files, but on newer version u need ur trophy key there are steps to get it on discord iirc

image

diegolix29:
Yeah i will see into the fonts for imgui pop up window during these days.

GHU7924:
In the last build of "main", I have "Trophy Viewer" working, but not with this PR.

I hadn't used the trophy key before and didn't plan to do so in the future,
but because of this PR, I don't have any trophies displayed at all, I just wanted to see what the "bottom" option looks like now. I do not know, but maybe your trophy key got into PR or something... as I understand it.

diegolix29:
are u using separate folders for each installation? if so u need to copy ur game data folder from user folder to the pr folder to have ur trophies shown

GHU7924:
>are u using separate folders for each installation? if so u need to copy ur game data folder from user folder to the pr folder to have ur trophies shown

Oh, I didn't know about this, because I usually don't keep track of trophies, but for tests it was always enough for me to copy my saves or start a new game. It's working now. Thank you.

>i have centered the text

It looks good.


@DanielSvoboda didn't quite understand me correctly, so he did half the job, I didn't like that the trophy window now opens to the full screen, before it was just a window, I would like it to stay that way. He fixed only the second problem with the window.

Next
изображение
The checkboxes are also reset. For example, I selected all 3 checkboxes, but once I close and reopen Trophy Viewer, only 2 checkboxes will be selected.

Next
I also suggested this - https://github.com/shadps4-emu/shadPS4/pull/2539#issuecomment-2691540870 
The added button also works well, but maybe it will be better with a checkbox, that is, it will be possible to show the sidebar or not.

diegolix29:
Ok now i have changed the fullscreen to window, just doing the checkboxes stuff

DanielSvoboda:
@GHU7924 I had said that I fixed it, but what I mean is that it does this: it opens in full screen, but when you click to make it smaller, it becomes a decent size, and not very small as it was shown. However, always starting maximized seems better to me, as there are always a lot of items...
yes, currently the values ​​of these checkboxes are not saved anywhere, using comfig.toml might be a good idea.

GHU7924:
>I had said that I fixed it, but what I mean is that it does this: it opens in full screen, but when you click to make it smaller, it becomes a decent size, and not very small as it was shown.

That's why I said you fixed only the second problem with the window. 😄 

>However, always starting maximized seems better to me, as there are always a lot of items...

This would be true for normal resolutions, I could agree with you, but for users who have widescreen monitors with an aspect ratio of 21:9 (or 32:9), Trophy Viewer just will be too stretched. Therefore, I suggest doing this:


Add two new checkboxes (highlighted in blue):
1) Open trophy Viewer to full screen
2) Show Filters Sidebar

Make the window size as follows:
Height to fit 7 icons (highlighted in red)
When the sidebar is open, there is no slider at the bottom of the window (yellow arrow indicates)
*If this window looks big in Full HD, then you can reduce it: the width of the window may end where the "Type" column ends, and the height may be limited to 6 (or 5) icons.

>yes, currently the values ​​of these checkboxes are not saved anywhere, using comfig.toml might be a good idea.

I think that the choice should be preserved, which will give more opportunities for the user to customize the trophies.

PR #2575, opened 2025-03-01
video_core: support for multi-layered outputs

polybiusproxy:
Games can choose what slice of the render target to write to from within a shader—this is commonly used for drawing cubemaps in a single pass. This can be done from geometry, vertex, or tesselation eval shaders.
This works by outputting the selected slice to the Z component of the pos1 target:
```
s_load_dwordx16 s[40:55], s[6:7], 0x0
/*snip*/
v_mov_b32       v15, s40
/*snip*/
exp             pos1, off, off, v15, off done
```

Can also be seen used on the AMD compilers:
- https://godbolt.org/z/xT9nfGdcs <- Example of compiler using pos1 exports to output RT slice target.
- https://github.com/GPUOpen-Drivers/llpc/blob/6222f1ff6ba7f00b085003c1a05fdbdc31356235/lgc/lowering/LowerInOut.cpp#L1526

This PR also fixes a bug where depth targets weren't refreshed - fixes a bug where a depth buffer cleared from a compute shader won't clear the depth image.

### DIFF
--- a/src/qt_gui/kbm_gui.cpp
+++ b/src/qt_gui/kbm_gui.cpp
@@ -608,21 +608,28 @@ void KBMSettings::CheckMapping(QPushButton*& button) {
     MappingTimer -= 1;
     button->setText(tr("Press a key") + " [" + QString::number(MappingTimer) + "]");
 
+    if (pressedKeys.size() > 0) {
+        QStringList keyStrings;
+
+        for (const QString& buttonAction : pressedKeys) {
+            keyStrings << buttonAction;
+        }
+
+        QString combo = keyStrings.join(",");
+        SetMapping(combo);
+        MappingCompleted = true;
+        EnableMapping = false;
+
+        MappingButton->setText(combo);
+        pressedKeys.clear();
+        timer->stop();
+    }
     if (MappingCompleted) {
         EnableMapping = false;
         EnableMappingButtons();
         timer->stop();
 
-        if (mapping == "lshift" || mapping == "lalt" || mapping == "lctrl" || mapping == "lmeta" ||
-            mapping == "lwin") {
-            modifier = "";
-        }
-
-        if (modifier != "") {
-            button->setText(modifier + ", " + mapping);
-        } else {
-            button->setText(mapping);
-        }
+        button->setText(mapping);
     }
 
     if (MappingTimer <= 0) {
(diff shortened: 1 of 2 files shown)


src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -76,7 +76,11 @@ Id OutputAttrPointer(EmitContext& ctx, IR::Attribute attr, u32 element) {
     case IR::Attribute::Position0: {
         return ctx.OpAccessChain(ctx.output_f32, ctx.output_position, ctx.ConstU32(element));
     }
-    case IR::Attribute::Position1:
+    case IR::Attribute::Position1: {
+        ASSERT(ctx.stage == Stage::Geometry || ctx.stage == Stage::Vertex);
+        ASSERT_MSG(element == 2u, "Unsupported pos1 export");
```
raphaelthegreat: Hardcoding position1 as gl_Layer seems not too good idea. Does using VsOutputAttrPointer not work here like other exports? When it uses gl_Layer it should have value of VsOutput::GsMrtIndex

src/video_core/renderer_vulkan/vk_scheduler.cpp
```diff
@@ -34,18 +34,22 @@ void Scheduler::BeginRendering(const RenderState& new_state) {
...
 
-    const auto width =
+    const u32 width =
         render_state.width != std::numeric_limits<u32>::max() ? render_state.width : 1;
-    const auto height =
+    const u32 height =
         render_state.height != std::numeric_limits<u32>::max() ? render_state.height : 1;
 
+    const u32 cb_layers =
+        *std::max_element(render_state.cb_slices.begin(), render_state.cb_slices.end());
```
raphaelthegreat:
```suggestion
        *std::ranges::max_element(render_state.cb_slices);
```
Small nit

raphaelthegreat (commented):
Thinking about it, I think we should get rid of the Position1/2/3 IR attributes and lower them to actual shader attributes like RenderTargetId/ClipDistance/CullDistance directly during export, since we have ctx.runtime_info.vs_info.outputs available. This should remove the need for hard coding things like the check for info.has_layer_output and simplify code a lot in backend

Randomuser8219:
Knack crashes with this PR (Unsupported pos1 export).
[CLbAFJE.txt](https://github.com/user-attachments/files/19040421/CLbAFJE.txt)

Randomuser8219:
Knack now gives:
[Render.Recompiler] <Warning> emit_spirv_context_get_set.cpp:22 MiscOutputAttrPointer: Misc. vector output used but disabled on stage vs, attr Position1, comp 0

src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
```diff
@@ -33,49 +33,52 @@ constexpr static std::array DescriptorHeapSizes = {
...
-               ctl.use_vtx_edge_flag
-                   ? VsOutput::EdgeFlag
-                   : (ctl.use_vtx_gs_cut_flag ? VsOutput::GsCutFlag : VsOutput::None),
-               ctl.use_vtx_kill_flag
-                   ? VsOutput::KillFlag
-                   : (ctl.use_vtx_render_target_idx ? VsOutput::GsMrtIndex : VsOutput::None),
-               ctl.use_vtx_viewport_idx ? VsOutput::GsVpIndex : VsOutput::None);
+    auto& outputs = info.GetOutputs();
+    if (ctl.vs_out_misc_enable) {
+        auto& misc_vec = outputs[0];
```
raphaelthegreat: Using static index 0/1/2 here is incorrect. If VS_OUT_MISC_VEC for example is disabled, then the next one VS_OUT_CCDIST0 will occupy the first output. Define a local num_outputs and index with num_outputs++ here and other branches

raphaelthegreat (commented):
This looks okay in general, but it would be a bit better if we moved the vertex misc output selection into IR. At the moment its kinda awkwardly worked around in backend side

PR #2580, opened 2025-03-02
Specialize vertex attributes on dst_sel

baggins183:


baggins183:
I remember seeing this as an issue for PT, where some components are mapped to 0 or 1 and this changes without recompilation. I havent reprod recently but think this is good to have

squidbus:
You will have to change the condition a bit here, currently since the specialization was for number type, this path is only used when `VK_EXT_legacy_vertex_attributes` is unavailable.

raphaelthegreat:
> You will have to change the condition a bit here, currently since the specialization was for number type, this path is only used when `VK_EXT_legacy_vertex_attributes` is unavailable.

Dst sel I believe is applied on shader side always since we cannot swizzle vertex attributes from host so I don't think this extension covers it.
EDIT: I think I misread the statement, you mean this still needs to be done but change it to apply always

squidbus:
> EDIT: I think I misread the statement, you mean this still needs to be done but change it to apply always

Yes, currently this specialization only happens when legacy is unavailable since that's what it was needed for. Now that there is specialization that is needed regardless, the condition needs to be updated.

baggins183:
Ok thanks fixed that.

Im confused though because I see in spirv_emit_context.cpp DefineInputs(), GetNumberFmt() and GetDataFmt() are used in codegen which should mean they need to always be compared in specialization info

squidbus (approved):

squidbus:
> Im confused though because I see in spirv_emit_context.cpp DefineInputs(), GetNumberFmt() and GetDataFmt() are used in codegen which should mean they need to always be compared in specialization info

For `GetNumberFmt()`, this is handled with legacy attributes extension as it allows mismatching number types between shader input and vertex attribute, and then the actual shader instructions will interpret it as whatever format is needed. For `GetDataFmt()`, this seems to only be used for step rate stuff which uses a separate buffer binding that might catch those cases, plus it's only used for the number of components which is unlikely to change for a single shader. Latter may be worth adding specialization for just in case but I'm not sure.

To be honest I'm not sure if the legacy attributes part is really needed either since I doubt there will be many cases of permutations from changing vertex attribute number types, but it's also not that complicated to have in.

baggins183:
Makes sense, I see NumberFmt and DataFmt are already builtin to the pipeline cache key (Ignored when legacy attribs are supported also) as a combined VkFormat.
NumberFmt needs to be in Spec info to trigger a VS recompile when legacy attribs are unsupported, so does it make sense to get rid of these in the pipeline key and only have them as Spec info? I could do that and move DataFmt to spec info.

squidbus:
> Makes sense, I see NumberFmt and DataFmt are already builtin to the pipeline cache key (Ignored when legacy attribs are supported also) as a combined VkFormat.

These are used when `VK_EXT_vertex_input_dynamic_state` is not supported, which is different from legacy attributes, to update vertex inputs in pipeline state instead when the format changes.

> NumberFmt needs to be in Spec info to trigger a VS recompile when legacy attribs are unsupported, so does it make sense to get rid of these in the pipeline key and only have them as Spec info?

This is for the shader to have the correct number type for the input, if needed because no `VK_EXT_legacy_vertex_attributes` support.

baggins183:
Yeah i was confusing the extensions, so NumberFmt and DataFmt are needed in the pipeline key and NumberFmt in spec info for different reasons.
I'll assume DataFmt is uneccessary for spec info then unless you want me to add it

squidbus:
> Yeah i was confusing the extensions, so NumberFmt and DataFmt are needed in the pipeline key and NumberFmt in spec info for different reasons. I'll assume DataFmt is uneccessary for spec info then unless you want me to add it

You can do a similar thing to what you have here for number format: if the attribute `UsesStepRates()` is true, then specialize data format, otherwise don't as it is unnecessary.

src/shader_recompiler/specialization.h
```diff
@@ -89,12 +91,16 @@ struct StageSpecialization {
...
-        if (info_.stage == Stage::Vertex && fetch_shader_data &&
-            !profile_.support_legacy_vertex_attributes) {
+        if (info_.stage == Stage::Vertex && fetch_shader_data) {
             // Specialize shader on VS input number types to follow spec.
             ForEachSharp(vs_attribs, fetch_shader_data->attributes,
-                         [](auto& spec, const auto& desc, AmdGpu::Buffer sharp) {
-                             spec.num_class = AmdGpu::GetNumberClass(sharp.GetNumberFmt());
+                         [&profile_](auto& spec, const auto& desc, AmdGpu::Buffer sharp) {
+                             spec.data_format =
+                                 desc.UsesStepRates() ? sharp.GetDataFmt() : AmdGpu::DataFormat{};
```
squidbus: One more thing here, it occurred to me after that we can even just specialize on `NumComponents(GetDataFmt())` to narrow it down further, as that's how the data format is used.

PR #2582, opened 2025-03-02
Fix time - sceKernelClockGettime

DanielSvoboda:
sceKernelClockGettime was not correctly getting the current date

When saving a game, this function was called, but it wasn’t returning the date correctly, as seen in the image with 1969 or 1970.
And as shown, the last two saves have the correct date, as this PR was used to save.

*If you have any game that shows the wrong date, it will only be corrected after making a new save.

Game: VA-11 Hall-A
image
*Yes, 02/03/2025 is correct because I am in Brazil, so it shows my date format, dd/mm/yyyy. This varies according to your system, as shown in the screenshot below.

Game: KNACK 2
| Before | After | 
|-------------|-------------|
image | image

It's very likely that this is also used in other places besides the save. I remember seeing some logs with incorrect dates.

### DIFF
--- a/src/core/libraries/kernel/time.cpp
+++ b/src/core/libraries/kernel/time.cpp
@@ -115,14 +115,16 @@ int PS4_SYSV_ABI sceKernelClockGettime(s32 clock_id, OrbisKernelTimespec* tp) {
         break;
     }
 
-    timespec t{};
-    int result = clock_gettime(pclock_id, &t);
-    tp->tv_sec = t.tv_sec;
-    tp->tv_nsec = t.tv_nsec;
-    if (result == 0) {
-        return ORBIS_OK;
+    time_t raw_time = time(nullptr);
+
+    if (raw_time == (time_t)(-1)) {
+        return ORBIS_KERNEL_ERROR_EINVAL;
     }
-    return ORBIS_KERNEL_ERROR_EINVAL;
+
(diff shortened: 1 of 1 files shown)


derholzmann12321:
the following game spams the below line
coffee talk CUSA18112
[Lib.Ime] <Info> ime.cpp:272 sceImeKeyboardOpen: called

DanielSvoboda:
> the following game spams the below line coffee talk CUSA18112 [Lib.Ime] ime.cpp:272 sceImeKeyboardOpen: called

I believe this is not related to my PR...
This spam also happens in the game 'CUSA51046 - Neva', it stays in a loop. But it has nothing to do with 'time'

shinra-electric:
This fixes the time in Dead Cells. 

(I was hoping it might fix saves not being persistent, but they still disappear after a restart)

Screenshot 2025-03-02 at 18 33 37
Screenshot 2025-03-02 at 18 34 26

DanielSvoboda:
> This fixes the time in Dead Cells.
> 
> (I was hoping it might fix saves not being persistent, but they still disappear after a restart)
> 

yes it corrects the date, this save issue also happens in the 'main wip'

Xphalnos:
Thanks! This fixed the time display in Farming Simulator 17!

Before:
Before

After:
After

PR #2586, opened 2025-03-03
Make sceImeKeyboardOpen log debug only

Randomuser8219:
I've recently found out that all my Unity games started spamming this every frame.
Reduces log spam in Unity games.


georgemoralis:
no it is better to dig for the issue , i will do it

PR #2589, opened 2025-03-04
Add global/common user folder for Windows

### LINKED ISSUE #2549: [Feature Request]: Option to Change User Folder Path to AppData
### Checklist

### Description

Currently, the user folder in shadPS4 is linked to the portable directory of the emulator. This setup can be problematic for users who run multiple versions of shadPS4 for testing purposes.

To improve flexibility, I propose adding an option during the initial setup to choose where the user folder should be stored—either in the emulator's directory or in the system's AppData folder. This would allow users to maintain consistency across different versions and avoid conflicts.

### Reason

**Proposed Behavior**

When launching shadPS4 for the first time (or if no user folder is found), the emulator should prompt the user to select a preferred location for the user folder, with a brief explanation of the benefits of each option:

    **AppData path:** Recommended for users who want a persistent location for saves and trophies across different versions.
    **Portable path:** Useful for those who prefer keeping all data within the emulator's directory for portability.

### Examples

**Handling Existing User Folders**

There would be three possible cases:

-     If a user folder exists in AppData, shadPS4 automatically uses it.
-     If a user folder exists in the portable directory, shadPS4 defaults to that.
-     If no user folder is found in either location, shadPS4 prompts the user to choose a preferred location.

This enhancement would provide more flexibility and better support for users with multiple emulator versions.

rainmakerv3:
Closes https://github.com/shadps4-emu/shadPS4/issues/2549

Similar to other OS's, this creates a global user folder for shadps4 on Windows (Appdata/Roaming/shadps4). Same as other OS, if there is a portable user folder present, shadPS4 will use that instead, so users can still have separate settings per folder if they want.

Also adds a button for easily creating a portable user folder under Settings Dialog / paths tab:

Screenshot 2025-03-04 164751

### DIFF
--- a/src/common/path_util.cpp
+++ b/src/common/path_util.cpp
@@ -17,6 +17,8 @@
 #ifdef _WIN32
 // This is the maximum number of UTF-16 code units permissible in Windows file paths
 #define MAX_PATH 260
+#include <Shlobj.h>
+#include <windows.h>
 #else
 // This is the maximum number of UTF-8 code units permissible in all other OSes' file paths
 #define MAX_PATH 1024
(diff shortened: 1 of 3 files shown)


kalaposfos13:
If I understand this correctly, for the new folder to be used, you currently have to reopen the app, you should either add that to the description of the button, or make it so the settings are reloaded upon clicking the button

rainmakerv3:
Yeah I think it has to be restarted. Edited the description text accordingly

PR #2598, opened 2025-03-05
Adds missing tr functions for certain GUI strings that should be translatable

rainmakerv3:
Some GUI strings which are user-facing currently do not have tr functions, so they cannot be translated. There are some more other than this which are more difficult to do (some requiring maps, and some like the special keybinding section which are difficult to handle), but at least this quick PR covers the easy ones.

### DIFF
--- a/src/qt_gui/settings_dialog.cpp
+++ b/src/qt_gui/settings_dialog.cpp
@@ -235,21 +235,6 @@ SettingsDialog::SettingsDialog(std::span<const QString> physical_devices,
                                       Common::FS::GetUserPath(Common::FS::PathType::CustomTrophy));
             QDesktopServices::openUrl(QUrl::fromLocalFile(userPath));
         });
-
-        connect(ui->PortableUserButton, &QPushButton::clicked, this, []() {
-            QString userDir;
-            Common::FS::PathToQString(userDir, std::filesystem::current_path() / "user");
-            if (std::filesystem::exists(std::filesystem::current_path() / "user")) {
-                QMessageBox::information(NULL, tr("Cannot create portable user folder"),
-                                         userDir + " " + tr("already exists"));
-            } else {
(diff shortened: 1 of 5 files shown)


DanielSvoboda:
This is very good, there is just one adjustment that needs to be made. in this part because the text has more than one line it is better to use these commands before and after and leave everything on one line
// clang-format off
// clang-format on
image
image

src/qt_gui/settings_dialog.cpp
```diff
@@ -240,14 +240,14 @@ SettingsDialog::SettingsDialog(std::span<const QString> physical_devices,
             QString userDir;
             Common::FS::PathToQString(userDir, std::filesystem::current_path() / "user");
             if (std::filesystem::exists(std::filesystem::current_path() / "user")) {
-                QMessageBox::information(NULL, "Cannot create portable user folder",
-                                         userDir + " already exists");
+                QMessageBox::information(NULL, tr("Cannot create portable user folder"),
+                                         userDir + " " + tr("already exists"));
```
Osyotr: `userDir` should be passed as an argument.

DanielSvoboda: Yes, you must use arguments like this for the translation to be consistent in all languages.
```cpp
                QMessageBox::information(
                    NULL, tr("Cannot create portable user folder %1 already exists"), userDir);
```

rainmakerv3: ok done

rainmakerv3:
done settings the long strings as one line as well with clang format off

rainmakerv3:
> This is very good, there is just one adjustment that needs to be made. in this part because the text has more than one line it is better to use these commands before and after and leave everything on one line // clang-format off // clang-format on image image

Ok done

PR #2600, opened 2025-03-05
Remove Unused Semicolons, Fix Windows.h includes and Minor Warnings Fixes

Xphalnos:
I used warning level 4 for Clang to find the warnings

- Remove unused semicolons
- Changed includes from `windows.h` to `Windows.h` as this creates a warning
- Minor Warnings Fixes
- Adding spaces below files that don't have them (Another clang warning)


DanielSvoboda:
maskOffsetStr is used in mask and mask jump 32, I believe I can't put this () to just validate in mask_jump32
src/common/memory_patcher.cpp


image

DanielSvoboda:
in this code too, you should not make this change, because the logic is that when type is mask or mask_jump32 AND maskOffsetStr has some content, then mask also uses Offset
image
image

Xphalnos:
@DanielSvoboda OK, in doubt I removed the modifications on this part.

georgemoralis:
not neccesary changes atm

PR #2602, opened 2025-03-05
Fix show background image not working with the OLED theme

Missake212:
Should fix issue shadps4-emu/shadps4-qtlauncher#31 , I'll put it as a draft for the time being because I'm not sure if that's "correct", hoping someone could take a quick look and tell me if that's the right way to do it.


smiRaphi:
Input boxes and unmarked checkboxes are invisible with the things you've removed

Missake212:
> Input boxes and unmarked checkboxes are invisible with the things you've removed

You're right I didn't check that, my bad, not too familiar with QT stuff.
But if you select another theme first, click on a game an go back to the OLED theme the background will show up (although that would just bug out the image), doesn't that mean that there should be a way to have this without getting rid of the unmarked checkboxes ?


https://github.com/user-attachments/assets/a2677153-c45c-409a-8b5e-9438d0e6b69c

Missake212:
Since this doesn't seem to be a proper solution I'll just close this one, hoping someone else with more QT knowledge can figure it out.

PR #2603, opened 2025-03-06
Play Trophy sound from .wav file

smiRaphi:
Translations for IT, NB, RU, CN, SE, BR, PT, NO & FR would need to be updated to include the "or".


georgemoralis:
you don't need to add translations files ,they will be recreate from crowdin

smiRaphi:
Alright, I removed the translation files from the pull request.

PR #2606, opened 2025-03-06
Fix OLED theme background image issue

### LINKED ISSUE #31: [APP BUG]: Show background image doesn't work with the OLED theme
### Checklist

### Describe the Bug

It seems like using the OLED theme implemented by @smiRaphi of shadPS4 makes the "Show background image" option not work at all, one could argue it makes sense since you don't want a static image with an OLED display but I believe it would be for the best to let the user decide of that, and that's why the option can be toggled on and off, and is off by default. Here's a video showing the issue:

https://github.com/user-attachments/assets/7e5d8b90-51cb-4396-b7d2-7bde7ac735d3

### Reproduction Steps

Select the OLED theme with "Show background image" enabled and the image won't appear. If you do it after already being on a game with another theme (so the background image will show up) clicking on another game won't change it accordingly.

### Expected Behavior

Background images should work properly, like other themes, with this setting enabled.

### Specify OS Version

Windows 11

SharkStudiosSK:
Fixes shadps4-emu/shadps4-qtlauncher#31

Update the OLED theme to support the "Show background image" option.

* Add code to `src/qt_gui/main_window_themes.cpp` to set the background image if `Config::getShowBackgroundImage()` is true.
* Adjust the opacity and other settings for the OLED theme in `src/qt_gui/main_window_themes.cpp`.
* Update the `SetGridBackgroundImage` function in `src/qt_gui/game_grid_frame.cpp` to handle the OLED theme.
* Update the `SetListBackgroundImage` function in `src/qt_gui/game_list_frame.cpp` to handle the OLED theme.

### DIFF
--- a/src/qt_gui/game_grid_frame.cpp
+++ b/src/qt_gui/game_grid_frame.cpp
@@ -1,9 +1,9 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
-
 #include "common/path_util.h"
 #include "game_grid_frame.h"
 #include "qt_gui/compatibility_info.h"
+
 
 GameGridFrame::GameGridFrame(std::shared_ptr<GameInfoClass> game_info_get,
                              std::shared_ptr<CompatibilityInfoClass> compat_info_get,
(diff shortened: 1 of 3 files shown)


Missake212:
Seems to still be broken on my end.

georgemoralis:
closed due to inactivity

Issue #2604, opened 2025-03-06
[APP BUG]: ORBIS_VIDEODEC2_ERROR_STRUCT_SIZE does not give out a proper error


Hasster1:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

The error code ORBIS_VIDEODEC2_ERROR_STRUCT_SIZE is being triggered by Five Nights At Freddy's Security Breach (CUSA24172) in videodec2.cpp (sceVideodec2GetPictureInfo, currently at line 168) when trying to decode an mp4 video.

Emulator crashes with seemingly no errors

![Image](https://github.com/user-attachments/assets/cbc18b1f-ea48-4379-ba6c-508c9d2bbc4a)

[shad_log.txt](https://github.com/user-attachments/files/19101205/shad_log.txt)

### Reproduction Steps

Trigger that error code.

### Expected Behavior

Error should be shown in the log.

### Specify OS Version

Windows 10

thecatontheceiling:
Have you made sure the log type is set to sync? I think that can affect things like this

Hasster1:
> Have you made sure the log type is set to sync?

I've tried with both sync and not, it doesn't really help. 

On another note, i just realized that it might just be an internal ps4 code, and the actual error should just be logged beside that return function... Is that right?

PR #2611, opened 2025-03-06
 Videodec2 mismatching picture info size crash fix

Hasster1:
Returning before setting *picInfo causes the emu to crash. I think it's better to let it deal with whatever corrupted size it is, rather than forcing the emu to crash.
Also added logging.

Closes the https://github.com/shadps4-emu/shadPS4/issues/2604

First pull request here, please correct me if something's wrong.

### DIFF
--- a/src/core/libraries/videodec/videodec2.cpp
+++ b/src/core/libraries/videodec/videodec2.cpp
@@ -162,12 +162,15 @@ s32 PS4_SYSV_ABI sceVideodec2GetPictureInfo(const OrbisVideodec2OutputInfo* outp
     if (p1stPictureInfoOut) {
         OrbisVideodec2AvcPictureInfo* picInfo =
             static_cast<OrbisVideodec2AvcPictureInfo*>(p1stPictureInfoOut);
+
+        *picInfo = gPictureInfos.back();
+
         if (picInfo->thisSize != sizeof(OrbisVideodec2AvcPictureInfo)) {
-            LOG_CRITICAL(Lib_Vdec2, "Mismatching sizes, first = {}, second = {}",
+            LOG_ERROR(Lib_Vdec2, "Mismatching sizes, first = {}, second = {}",
                          picInfo->thisSize, sizeof(OrbisVideodec2AvcPictureInfo));
             return ORBIS_VIDEODEC2_ERROR_STRUCT_SIZE;
         }
-        *picInfo = gPictureInfos.back();
+        
     }
 
     if (outputInfo->pictureCount > 1) {
--- a/src/core/libraries/videodec/videodec2.cpp
+++ b/src/core/libraries/videodec/videodec2.cpp
@@ -163,6 +163,8 @@ s32 PS4_SYSV_ABI sceVideodec2GetPictureInfo(const OrbisVideodec2OutputInfo* outp
         OrbisVideodec2AvcPictureInfo* picInfo =
             static_cast<OrbisVideodec2AvcPictureInfo*>(p1stPictureInfoOut);
         if (picInfo->thisSize != sizeof(OrbisVideodec2AvcPictureInfo)) {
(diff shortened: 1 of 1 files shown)


roamic:
I don't think you intended to update the submodules here.

Hasster1:
> I don't think you intended to update the submodules here.

Yeah, it's my first pull request, sorry. I don't even know how that happened.

src/core/libraries/videodec/videodec2.cpp
```diff
@@ -162,10 +162,15 @@ s32 PS4_SYSV_ABI sceVideodec2GetPictureInfo(const OrbisVideodec2OutputInfo* outp
     if (p1stPictureInfoOut) {
         OrbisVideodec2AvcPictureInfo* picInfo =
             static_cast<OrbisVideodec2AvcPictureInfo*>(p1stPictureInfoOut);
+
+        *picInfo = gPictureInfos.back();
```
mailwl, line 166: something wrong here.
we need to check structure size BEFORE actually copy pic info (and if it lesser then needed, do not copy at all)

Hasster1: That is probably the intended way, but currently, if emu returns the error code before setting picInfo, it will crash (at least for FNAF SB). There should be another way to solve this issue, but for now i think it's better to let the emu process the picture even with a mismatching size, because it at least won't crash immediately.

Hasster1: > and if it lesser then needed, do not copy

Wait, are you sure about that? Then why is it checking for equality in the original code?

mailwl: i don't know why checking exact size, because it seems orbis os version depends (or videodec2 version?). for example in 9.0 it much bigger than 4.5
probably better copy only part of structure, something like this:
```
    if (p1stPictureInfoOut) {
        OrbisVideodec2AvcPictureInfo* picInfo =
            static_cast<OrbisVideodec2AvcPictureInfo*>(p1stPictureInfoOut);
        auto info = gPictureInfos.back();
        auto size = std::min(picInfo->thisSize, info.thisSize);
        info.thisSize = size;
        std::memcpy(picInfo, &info, size);
    }
```
not sure if it's good way

Hasster1: I think that it is a better way to deal with this issue, but it still needs to be logged when the size doesn't match (Probably as an error?)

```
    if (p1stPictureInfoOut) {
        OrbisVideodec2AvcPictureInfo* picInfo =
            static_cast<OrbisVideodec2AvcPictureInfo*>(p1stPictureInfoOut);
        auto info = gPictureInfos.back();
        if (picInfo->thisSize != info.thisSize) {
            LOG_ERROR(Lib_Vdec2, "Mismatching sizes, first = {}, second = {}", picInfo->thisSize,
                      sizeof(OrbisVideodec2AvcPictureInfo));
        }
        auto size = std::min(picInfo->thisSize, info.thisSize);
        info.thisSize = size;
        std::memcpy(picInfo, &info, size);
    }
```

How does this look?

polybiusproxy: It looks like the structs changed between releases. I have to recheck how the library handles this.

mailwl: btw, "Five Nights At Freddy's Security Breach (CUSA24172)" version 1.0 (compiled with sdk 8.5) doesn't have this error, unlike latest patch (sdk 10.5)

mailwl: > How does this look?

i saw one problem, if `picInfo` structure is bigger than `info`, the tail of `picInfo` may contains garbage (and actual shadPS4 videodec2 version implementation lower, than needed). This is not a problem on real ps4, because it doesn't let to install patch, compiled for fw 10.5 on fw 8.5

georgemoralis:
not a proper solution

PR #2614, opened 2025-03-07
Unified Memory Allocation

diegolix29:
Allow a game to expand memory when needs it implementing a basic way for a game to call for more memory in a dynamic way. 

Makes Shadow Returns CUSA28615 get ingame. 

image

image

image

### DIFF
--- a/src/core/libraries/kernel/memory.cpp
+++ b/src/core/libraries/kernel/memory.cpp
@@ -276,7 +276,7 @@ int PS4_SYSV_ABI sceKernelDirectMemoryQuery(u64 offset, int flags, OrbisQueryInf
 
 s32 PS4_SYSV_ABI sceKernelAvailableFlexibleMemorySize(size_t* out_size) {
     auto* memory = Core::Memory::Instance();
-    *out_size = memory->GetTotalUnifiedMemorySize();
+    *out_size = memory->GetAvailableUnifiedMemorySize();
     LOG_INFO(Kernel_Vmm, "called size = {:#x}", *out_size);
     return ORBIS_OK;
 }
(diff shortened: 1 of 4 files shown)


polybiusproxy:
There's no such thing as 'dynamic memory allocation' on the PS4, you're also significantly increasing the total memory up to 20 GB. This appears to be one big hack for one game and thus should be closed in my opinion.
The issue with this game should be properly debugged instead of working around it.

diegolix29:
yeah its just a test, as i was really surprised it worked, i want to have some way other people test it thats why is a draft, i will see a good way of appraching this as the game isnt acctually recibing more memory, cause the only game set here is bloodborne for the cusa, but the game per se asked for a way to allocate more memory, and with this approach it got ingame, so i guess that it  may be cause the ps4 allocate memory between ram and vram, so theres always a way a game can ask for more memory, even if it dosent uses it. that maybe why the visuals are fried. but its only my guess im not expert like yall guys

polybiusproxy:
It doesn't work like that, as I said this is not the way to do this, and the underlying cause must be debugged instead of hackily giving as much memory as the game wants.

diegolix29:
I changed the scope here to unify as the ps4 does the memory allocation, still has a extra parameter but searching i find out that the ps4 allocate 1gb more to game, but i find out putting 2, as the mutiplicator does, * 1.5, it will add 2.8gb more for the games to use, only when neomode is on

StevenMiller123:
This is still a hack, and it will break a handful of games. An assumption like this should be backed up with evidence, such as hardware tests demonstrating this behavior, a decompilation of the game that suggests something like this should occur, or a decompilation of the relevant firmware functions proving that this occurs.

In this particular case, it's highly likely that you're running into a known bug with Unity games, caused by issues with our exceptions support on Windows. Many Unity games are throwing similar memory errors to what you showed from this title on the Discord server, but these Unity games only demonstrate this behavior on Windows, and don't require any weird hacks like this to progress further on Linux.

diegolix29:
Yeah is ok im closing this as it is irrelevant now. Thanks for clarifying it out.

PR #2616, opened 2025-03-07
fix: build linux with glibc 2.39 and QT 6.4

jardon:
- build linux binaries in noble container

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -284,7 +284,7 @@ jobs:
       options: --privileged -it
     steps:
     - name: Install dependencies
-      run: sudo apt-get update && sudo apt install -y git libx11-dev libxext-dev libwayland-dev libdecor-0-dev libxkbcommon-dev libglfw3-dev libgles2-mesa-dev libfuse2 clang build-essential libasound2-dev libpulse-dev libopenal-dev libudev-dev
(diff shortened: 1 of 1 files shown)


tomboylover93:
Is there any significant difference from just running it with the Ubuntu 24.04 runner provided by GitHub?

jardon:
I'm playing around with this as a means to increase compatibility.  glibc has forward compatibility (for the most part) and therefore building with the oldest version of glibc that can be compiled against, will enable older distributions to run them.  I tried dropping back to Ubuntu 22.04, but there are some hard requirements that arent satisfied by the dependencies in jammy.

at the end of the day, linux distributions are the ones that freeze their libs and would need to provide packaging for their specific distro.  this is why we only build flatpaks and appimages as they ship the dependencies needed to allow better portability.  this is related to the glibc issue that someone reported which im trying to solve by bundling glibc, but this was a separate effort to increase compatibility

jardon:
OH i see your point.  i got my codenames mixed up

jardon:
so the correct thing to do here would be to build off of jammy, but that wont work based on the QT version thats shipped with jammy

tomboylover93:
There's an ubuntu64 artifact that gets generated with every workflow run, but it's SDL. What I think you should do is change that from a single executable to an AppImage, make a Qt variant and change those to use the Ubuntu jammy container (might need some hacking around). Or perhaps you could package the Ubuntu jammy AppImage along with the Ubuntu noble one on a separate folder inside the same zip file?

tomboylover93:
For the older Qt version I think you could circumvent this by downloading the Qt online installer for Linux (https://www.qt.io/download-dev#eval-form) and setting it up the same way as the Windows workflow runs.

PR #2617, opened 2025-03-07
Simplified chinese translation, oral habits

lmutig:


georgemoralis:
translations is now handled on https://crowdin.com/project/shadps4-emulator

You should provide your fix there

PR #2621, opened 2025-03-07
sceKernelLoadStartModule: convert relative path to absolute

mailwl:
fixes https://github.com/shadps4-emu/shadps4-game-compatibility/issues/2721


mailwl:
checked on UE4 engine games - they prodive abs path to ./prx/ folder
xna game the same

georgemoralis:
there is also #2541  did you check if that works?

mailwl:
oh, yes, it fixes, too

PR #2623, opened 2025-03-08
sceKernelReleaseDirectMemory fix

StevenMiller123:
This addresses an issue with the `ReleaseDirectMemory` functions, where some games would hit an assert by calling them with `len = 0`. I also fixed an inaccurate error return in `sceKernelMunmap` (based on FreeBSD documentation), though I don't have any games that were hitting that error return.

This should fix some cases of `[Debug] <Critical> memory.cpp:766 operator(): Assertion Failed!`

This brings TerraTech (CUSA11362) ingame.
image

### DIFF
--- a/src/core/libraries/kernel/memory.cpp
+++ b/src/core/libraries/kernel/memory.cpp
@@ -79,6 +79,9 @@ s32 PS4_SYSV_ABI sceKernelAllocateMainDirectMemory(size_t len, size_t alignment,
 }
 
 s32 PS4_SYSV_ABI sceKernelCheckedReleaseDirectMemory(u64 start, size_t len) {
+    if (len == 0) {
+        return ORBIS_OK;
+    }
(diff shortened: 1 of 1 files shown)


src/core/libraries/kernel/memory.cpp
```diff
@@ -506,8 +512,8 @@ s32 PS4_SYSV_ABI sceKernelConfiguredFlexibleMemorySize(u64* sizeOut) {
 
 int PS4_SYSV_ABI sceKernelMunmap(void* addr, size_t len) {
     LOG_INFO(Kernel_Vmm, "addr = {}, len = {:#x}", fmt::ptr(addr), len);
-    if (len == 0) {
-        return ORBIS_OK;
+    if (len <= 0) {
```
mailwl: size_t is unsigned, cannot be negative

StevenMiller123: Makes sense

Randomuser8219:
Cuphead now gets further on Windows!
image

PR #2625, opened 2025-03-08
Make button bar translatable in controller/KBM dialog

rainmakerv3:
Fixes the button bars (save/apply/cancel/restore default) not being translatable on both remapping GUIs

### DIFF
--- a/src/qt_gui/control_settings.cpp
+++ b/src/qt_gui/control_settings.cpp
@@ -28,6 +28,11 @@ ControlSettings::ControlSettings(std::shared_ptr<GameInfoClass> game_info_get, Q
         }
     });
 
+    ui->buttonBox->button(QDialogButtonBox::Save)->setText(tr("Save"));
+    ui->buttonBox->button(QDialogButtonBox::Apply)->setText(tr("Apply"));
+    ui->buttonBox->button(QDialogButtonBox::RestoreDefaults)->setText(tr("Restore Defaults"));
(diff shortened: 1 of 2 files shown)


GHU7924:
I do not know how it works in other languages, but I noticed that the list for the "Mouse to Joystick" option (none, right, left) and mouse movement options (deadzone...) has not been translated. These options may not be available for translation.

Next (This is what the KBM window looks like in the Russian version).
You can see the elements overlapping each other.
config

I also suggest making such a change:
config2

Replace the keyboard image with something similar (By default, the mouse is not used, so the keyboard image will be enough):
https://thecode.media/wp-content/uploads/2022/10/74-1920x1543.png
(this is a sample image, and it doesn't fit because it doesn't have arrow buttons and a nampad, but if the default buttons are "C, V, B, N", then you don't need to add the numpad, and if you replace the arrows with the combination "Alt + W, A, S, D", then these options are quite suitable.)

Example of what it should look like in the end:




P.S. Make these changes or not, decide for yourself, these are just suggestions.

rainmakerv3:
I can take a look at these later on (I'm taking a bit of a break right now,  maybe for some weeks only doing small PRs). I suggest you do a feature request / issue report (espeically for the spacing issues on other languages, that one is very important) on this and I'll see when I can get back to it (or maybe someone else can pick it up if I take too long)

GHU7924:
You can simply save this PR in your notifications.
When you decide to make corrections, then take a look, rest, there is no need to rush it.

PR #2626, opened 2025-03-08
Memory allocation improvements

diegolix29:
Right now memory is allocated without checking for the best available block which can lead to fragmentation during gameplay, this try to improv that by fetching the larger contiguous blocks

also avoids skipping over useful gaps that could be used for future allocations

let me know if this is a good improvement. Or if its redundant. added like 5+ fps on my end for Bloodborne.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -128,36 +128,38 @@ PAddr MemoryManager::Allocate(PAddr search_start, PAddr search_end, size_t size,
     std::scoped_lock lk{mutex};
     alignment = alignment > 0 ? alignment : 16_KB;
 
-    auto dmem_area = FindDmemArea(search_start);
+    PAddr best_addr = -1;
+    size_t best_size = 0;
 
-    const auto is_suitable = [&] {
-        if (dmem_area == dmem_map.end()) {
-            return false;
+    for (auto dmem_area = FindDmemArea(search_start);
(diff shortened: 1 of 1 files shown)


TagoSMD:
Captura de pantalla 2025-03-08 095014
Captura de pantalla 2025-03-08 094933
Nice +4 fps

diegolix29:
Nicee thanks for testing

georgemoralis:
This PR's code seems questionable, and we suspect it may have been AI generated. If this is the case, please refrain from making further contributions using such tools

diegolix29:
The same here i was working on a other approach for the unified memory allocation cause i was reading about it the. I wanted to try some things here and there cause i had mis understanding some things, my other approach was to fully unified them but then i read and poly point out how the ps4 actually does things as i only know how code wise things can be implemented but no much into the ps4 way. Then tried to implement some variations, when i hit a block i ask of how the ps4 would handle it and i try to add it the proper way. Thats the result here. And on the other pr. I starterd to read and my videos too got into more heavy stuff and im trying to figure things out even i asked if that was a good improvement or if it was redundant. But the same if u dont trust it dont worry i can stop contributing to this and just keep learning haha but true im learning to help here so i might just keep it as a side learning skill after all. Thanks again for helping me onto get to understand and learn this haha i was so in the dark last year. Also if u want i can only contribute on gui stuff thats more soft coding.

PR #2632, opened 2025-03-09
New icons for the GUI

Hermiten:
Suggested here:

https://github.com/shadps4-emu/shadps4-qtlauncher/issues/43

I did change two icons to be more accurate. Now this is like this:

Before:
image

After:
image


diegolix29:
They seem good and accurate to the others but i think the controller one should resemble more to the ps4 controller

smiRaphi:
Me too, there is also a simpler ps4 controller icon linked in the mentioned issue that can be used

Hermiten:
I did a try with the other one but this one is much better. I know this is not a PS4 controller but we can use every controller we want and not only the DS4 one :)

diegolix29:
The emulator is a ps4 emulator so it should be a ps4 controller 🎮

smiRaphi:
The controller in the remapping screen is also a ps4 controller so having one as the icon for that would just make sense

diegolix29:
Screenshot_20250309_101731
 

Ngl this really looks good

diegolix29:
Maybe just adjust the spacing for them to be centered

Hermiten:
I think it's much less in keeping with the other icons, but if I have to, I can suggest this one, yes.
Here is it: 

image

My current PR:
image

You can vote or decide or idk

smiRaphi:
I would make the ps4 version a bit bigger and maybe move it down a little so it feels more centered

Hermiten:
The problem is the thickness of the line, which can't be changed

smiRaphi:
~~Why not?~~
Edit: looked at it a bit more and yeah the parts are too close to each other, but just increasing the line thickness on the outside for the outline and on the inside for the joysticks and making the other parts just bigger in general might work

ghost:
if the thickness is the issue i feel a simple controller would be better too, since if it looks out of place it will look odd but it wont matter much if its not really a ps4 controller imo, and thank you for the PR really appreciated.

diegolix29:
can u make the frame of the controller smaller so it will pick up more image?

georgemoralis:
i think old ones looks better :D

PR #2633, opened 2025-03-09
Kernel: Changes to scekernelUsleep to improve precision

diegolix29:
Modifies sceKernelUsleep to improve precision by using CLOCK_MONOTONIC_RAW in certain cases.

Utilizing GetCurrentProcess instead of  GetCurrentThread  may also help reduce crashes in games that rely on precise memory timings by improving thread synchronization and reducing timing drift

Changes the sleep logic to a while loop that updates req with rem if nanosleep is interrupted this helps for a better synchronization with real PS4 behavior, and slighly smoother control of time sensitive operations

Changins the generic POXIS implementation to a more orbis specific behaviour with tp
Using time pointer in this cases makes it clear that the argument is a pointer to a time related structure helping avoid confusion in function calls

This integrate my pr that helped Deathlight get ingame with the last changes from the kernel gettime pr just merged 

image

Makes
Deadlight Director's Cut | CUSA04434 | 1.00 Playable (according to reports with image proof)

Mages of mystralia : works again but needs a few starts. (according to reports)

Mad max: is back ingame again. (acording to reports)

### DIFF
--- a/src/core/libraries/kernel/time.cpp
+++ b/src/core/libraries/kernel/time.cpp
@@ -72,17 +72,13 @@ int PS4_SYSV_ABI sceKernelUsleep(u32 microseconds) {
 
     return 0;
 #else
-    timespec start;
-    timespec remain;
-    start.tv_sec = microseconds / 1000000;
-    start.tv_nsec = (microseconds % 1000000) * 1000;
-    timespec* requested = &start;
-    int ret = 0;
-    do {
-        ret = nanosleep(requested, &remain);
-        requested = &remain;
-    } while (ret != 0);
-    return ret;
+    struct timespec req, rem;
+    req.tv_sec = microseconds / 1000000;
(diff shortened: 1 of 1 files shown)


georgemoralis:
This PR's code seems questionable, and we suspect it may have been AI generated. If this is the case, please refrain from making further contributions using such tools

diegolix29:
No i dont use it for coding at all i sometimes ask google to tell me something if i dont quite understand it. But nothing more. If a function i added needs debugging i do ask for the right path but i do the coding. Not ai generated. I dont pay for it at all and i don't think the free can do this kind of things. Also this was the first time i tried something more than just gui stuff. Took me awhile to understand the changes u can look at development i had a lot of asking and telling. But true if u cant trust it is ok i can just step aside from the project itself. Its yours so i take it step by step since like 4 months u told me how to properly learn this.

PR #2640, opened 2025-03-10
Log mismatching picture size in GetPictureInfo

Hasster1:
Alright, fixing that problem is a bit harder than i thought...
Still, it needs to be logged.
Currently it's set at critical, since the emu crashes if it returns before setting *picInfo


georgemoralis:
that won't probably solve any issue , the best solution to this will be to check with ghidra videodec2 module to see how it is done

Hasster1:
...Obviously it won't solve the issue, Still, at least having an error message would be good, because right now the emu crashes without any explicit error.

georgemoralis:
which game btw? i will try to take a look

Hasster1:
> which game btw? 

Five nights at Freddy's: Security Breach (CUSA24172) 01.16

PR #2647, opened 2025-03-14
fix: add missing packages to shell.nix

jardon:


georgemoralis:
is there a reason why that is still a draft?

jardon:
this is incomplete.  i wasnt able to figure it out in the time i spent on it.  i forgot about it and left it open.  ill close for now

PR #2651, opened 2025-03-16
build: Compile for Sandy Bridge CPU target.

squidbus:
* Set `-mcpu=sandybridge` to target Sandy Bridge level instruction set, which is roughly what we support for PS4 guest applications when accounting for CPU patches.
* Disable AVX2 for CryptoPP for similar reasons; this needs its own flag set as the compile options we supply don't affect its `try_compile` support check.
* Move CryptoPP to source build on Windows, so that updated flags will apply.
* Minor improvement to architecture detection to account for if `CMAKE_SYSTEM_PROCESSOR` is missing for some reason.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,10 +6,6 @@
 	path = externals/cryptopp
 	url = https://github.com/shadps4-emu/ext-cryptopp.git
 	shallow = true
-[submodule "externals/cryptoppwin"]
-	path = externals/cryptoppwin
-	url = https://github.com/shadps4-emu/ext-cryptoppwin.git
-	shallow = true
 [submodule "externals/zlib-ng"]
 	path = externals/zlib-ng
 	url = https://github.com/shadps4-emu/ext-zlib-ng.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,8 +37,10 @@ option(ENABLE_UPDATER "Enables the options to updater" ON)
 # First, determine whether to use CMAKE_OSX_ARCHITECTURES or CMAKE_SYSTEM_PROCESSOR.
 if (APPLE AND CMAKE_OSX_ARCHITECTURES)
     set(BASE_ARCHITECTURE "${CMAKE_OSX_ARCHITECTURES}")
-else()
+elseif (CMAKE_SYSTEM_PROCESSOR)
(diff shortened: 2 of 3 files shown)


squidbus:
Draft:
* Needs Windows testing
* Needs submodule PR merge: https://github.com/shadps4-emu/ext-cryptopp/pull/1

diegolix29:
what do u need testing on windows? i just booted bb on it all good

squidbus:
> what do u need testing on windows? i just booted bb on it all good

Testing package installer would be sufficient, that’s where cryptopp is primarily used.

klaxons1:
so this PR will help to run ShadPS4 without AVX?

squidbus:
> so this PR will help to run ShadPS4 without AVX?

No, just any compiled code that might be unintentionally using AVX2.

squidbus:
Updated submodule since PR there was merged, just needs verification now that package installer is still working on Windows.

diegolix29:
booted bloodborne all normal with the last push here

georgemoralis:
you have to test pkg extraction. That is what cryptoopp is using mostly

diegolix29:
I can do one quickly once sec

diegolix29:
it extracted ok

squidbus:
Thanks, should be good then.

amayra:
Isn't Ivy Bridge a better target tho ?

squidbus:
> Isn't Ivy Bridge a better target tho ?

Sandy Bridge should be fine

FeelFried:
`-mcpu=sandybridge`
Is that supposed to be a launch key? if so - crash on startup without log, win11 i7-13620H

StevenMiller123:
@FeelFried that is a compiler flag added to one of our submodules. This PR shouldn't change anything for users (aside from potentially fixing PKG extraction issues on older CPUs).

aldekotan:
> * Needs Windows testing

Shadps4 crashes when installing PKG without a log. I assume my Intel 2600k isn't compatible?

diegolix29:
it does work so it may be on ur end

> > * Needs Windows testing
> 
> Shadps4 crashes when installing PKG without a log. I assume my Intel 2600k isn't compatible?

PR #2652, opened 2025-03-17
RCAS attenuation slider to config, under graphics tab

diegolix29:
just passing the RCAS slider to config, if someone request i can add the checkboxes but i think isnt relevant atm

### DIFF
--- a/src/video_core/renderer_vulkan/host_passes/fsr_pass.cpp
+++ b/src/video_core/renderer_vulkan/host_passes/fsr_pass.cpp
@@ -273,7 +273,7 @@ vk::ImageView FsrPass::Render(vk::CommandBuffer cmdbuf, vk::ImageView input,
 
         { // rcas
             consts = {};
-            FsrRcasCon(reinterpret_cast<AU1*>(&consts.Const0), settings.rcas_attenuation);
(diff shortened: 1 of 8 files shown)


src/video_core/renderer_vulkan/host_passes/fsr_pass.cpp
```diff
@@ -273,7 +273,7 @@ vk::ImageView FsrPass::Render(vk::CommandBuffer cmdbuf, vk::ImageView input,
 
         { // rcas
             consts = {};
-            FsrRcasCon(reinterpret_cast<AU1*>(&consts.Const0), settings.rcas_attenuation);
+            FsrRcasCon(reinterpret_cast<AU1*>(&consts.Const0), Config::getRcasAttenuation());
```
viniciuslrangel: Don't change here. It should keep reading from settings

viniciuslrangel (changes requested):

src/video_core/renderer_vulkan/host_passes/fsr_pass.cpp
diegolix29: i think its done right now can u take a look?

diegolix29:
Seems like its ready now

PR #2654, opened 2025-03-19
Changing "submit report" button behavior in the GUI

Missake212:
Disabling the emulator version automatically filling up when creating a report with the GUI, also changing game-code to game-serial. Need to wait for another change to happen first before merging this. Also would be nice if someone could figure out a way to completely disable the option to report on a WIP builds and only allow it on Release ones, the deleted line is a temporary workaround at best.

### DIFF
--- a/src/qt_gui/gui_context_menus.h
+++ b/src/qt_gui/gui_context_menus.h
@@ -521,7 +521,7 @@ public:
                 "title", QString("%1 - %2").arg(QString::fromStdString(m_games[itemID].serial),
                                                 QString::fromStdString(m_games[itemID].name)));
             query.addQueryItem("game-name", QString::fromStdString(m_games[itemID].name));
-            query.addQueryItem("game-code", QString::fromStdString(m_games[itemID].serial));
+            query.addQueryItem("game-serial", QString::fromStdString(m_games[itemID].serial));
             query.addQueryItem("game-version", QString::fromStdString(m_games[itemID].version));
             query.addQueryItem("emulator-version", QString(Common::VERSION));
             url.setQuery(query);
--- a/src/qt_gui/gui_context_menus.h
+++ b/src/qt_gui/gui_context_menus.h
@@ -523,7 +523,6 @@ public:
             query.addQueryItem("game-name", QString::fromStdString(m_games[itemID].name));
             query.addQueryItem("game-serial", QString::fromStdString(m_games[itemID].serial));
             query.addQueryItem("game-version", QString::fromStdString(m_games[itemID].version));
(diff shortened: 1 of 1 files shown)


GHU7924:
Personally, I don't understand why such changes are needed...
Maybe I didn't understand something...

It would be better if we had a PR with improved reports.
Right now we have one template, but it would be better if there were two.

The first one is complete, as it is now, for games that do not have a report yet,
and the second (abbreviated) one, which is added when the version changes (checkpoints change), and it will look like this:

Used version: 0.6.0
Game version: 01.11
Status: Nothing
Log (Sync): shad_log.txt
Comment/Screenshot: No response

StevenMiller123:
The bigger issue we're running into is people creating reports on WIP builds, when that's against our reporting rules. By not filling in that box, people will be able to see one more warning about how we don't allow reports on WIP builds.

We're still in the process of discussing more fool-proof solutions to that issue.

Missake212:
Like I said in the PR this doesn't fix the issue, it just disables the line that automatically fills the template for the emulator version, because the compatibility list doesn't allow reports on WIP builds. Ideally someone with C++ and QT knowledge could figure out a way to disable the option to submit a report all together for WIP builds but I don't know how to do that.

Here's screenshots to show what this PR does.

Before:
image

After
image

GHU7924:
> The bigger issue we're running into is people creating reports on WIP builds, when that's against our reporting rules. By not filling in that box, people will be able to see one more warning about how we don't allow reports on WIP builds.
> 
> We're still in the process of discussing more fool-proof solutions to that issue.

Then you should make 3 templates.
I have already described two of them, and the third would contain information with a warning.

I do not know how this can be done in the code, but for an example:
If version 0.*.0, then the report is available (Submit a report), you can create a new report (if there is none) or make an interim report (abbreviated, between versions), otherwise, if version 0.*.1, then open template number three.

Template number three (for an example):

`Thank you, friend, for wanting to share the compatibility report, we really appreciate it and are grateful for your help, but unfortunately our rules do not allow us to accept reports from nightly builds and so on, because and so on.`

Missake212:
I've added the line back so all this PR does is change game-code to game-serial, will help when the other change I talked about happens, I think I'll just create a new issue with all the information about the features that have issue in the GUI about compatibility stuff.

GHU7924:
@StevenMiller123 #2656 this solution will even be better.

I also think that in `Important: Read First`, the first line should be the link:
https://github.com/shadps4-emu/shadps4-game-compatibility?tab=readme-ov-file#--shadps4-game-compatibility

Also, I think it's still worth adding a template for reports between versions.

PR #2659, opened 2025-03-20
clock_gettime fixes for windows

diegolix29:
This PR changes `struct timespec` to `OrbisKernelTimespec` in `orbis_clock_gettime`, to fix binary compatibility issues with some games, such as Deathlight, Mages of mystralia and Mad max on windows. 

Also for me hollow knight can load savefiles too after this. Before only could start a new game. 

Makes those games boot again and deatlight can probably be full playable. Haven ask for a new report on it.

### DIFF
--- a/src/core/libraries/kernel/time.cpp
+++ b/src/core/libraries/kernel/time.cpp
@@ -121,7 +121,7 @@ static u64 FileTimeTo100Ns(FILETIME& ft) {
     return *reinterpret_cast<u64*>(&ft);
 }
 
-static s32 clock_gettime(u32 clock_id, struct OrbisKernelTimespec* ts) {
+static s32 clock_gettime(u32 clock_id, struct timespec* ts) {
     switch (clock_id) {
     case CLOCK_REALTIME:
     case CLOCK_REALTIME_COARSE: {
@@ -273,13 +273,10 @@ int PS4_SYSV_ABI orbis_clock_gettime(s32 clock_id, struct OrbisKernelTimespec* t
 }
 
 int PS4_SYSV_ABI sceKernelClockGettime(s32 clock_id, OrbisKernelTimespec* tp) {
-    struct OrbisKernelTimespec ts;
-    const auto res = orbis_clock_gettime(clock_id, &ts);
+    const auto res = orbis_clock_gettime(clock_id, tp);
     if (res < 0) {
         return ErrnoToSceKernelError(res);
     }
-    tp->tv_sec = ts.tv_sec;
-    tp->tv_nsec = ts.tv_nsec;
     return ORBIS_OK;
 }
 
--- a/src/core/libraries/kernel/time.cpp
+++ b/src/core/libraries/kernel/time.cpp
@@ -265,7 +265,11 @@ int PS4_SYSV_ABI orbis_clock_gettime(s32 clock_id, struct OrbisKernelTimespec* t
         return EINVAL;
     }
 
-    return clock_gettime(pclock_id, ts);
+    timespec t{};
+    int result = clock_gettime(pclock_id, &t);
+    ts->tv_sec = t.tv_sec;
+    ts->tv_nsec = t.tv_nsec;
+    return result;
 }
 
 int PS4_SYSV_ABI sceKernelClockGettime(s32 clock_id, OrbisKernelTimespec* tp) {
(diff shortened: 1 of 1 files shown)


src/core/libraries/kernel/time.cpp
```diff
@@ -265,11 +265,15 @@ int PS4_SYSV_ABI orbis_clock_gettime(s32 clock_id, struct timespec* ts) {
         return EINVAL;
     }
 
-    return clock_gettime(pclock_id, ts);
+    timespec t{};
+    int result = clock_gettime(pclock_id, &t);
```
roamic, line 269: Here you're using `clock_gettime` from winpthreads instead of the one I've written rendering the function useless.

roamic: If it fixes some games and doesn't break others then it means that my implementation of `clock_gettime` for windows is faulty.

src/core/libraries/kernel/time.cpp
```diff
@@ -121,7 +121,7 @@ static u64 FileTimeTo100Ns(FILETIME& ft) {
     return *reinterpret_cast<u64*>(&ft);
 }
 
-static s32 clock_gettime(u32 clock_id, struct timespec* ts) {
+static s32 clock_gettime(u32 clock_id, struct OrbisKernelTimespec* ts) {
```
roamic: Can you remove this change and see if it still works?

src/core/libraries/kernel/time.cpp
```diff
@@ -265,11 +265,15 @@ int PS4_SYSV_ABI orbis_clock_gettime(s32 clock_id, struct timespec* ts) {
...
+    timespec t{};
+    int result = clock_gettime(pclock_id, &t);
+    ts->tv_sec = t.tv_sec;
+    ts->tv_nsec = t.tv_nsec;
+    return result;
 }
 
 int PS4_SYSV_ABI sceKernelClockGettime(s32 clock_id, OrbisKernelTimespec* tp) {
-    struct timespec ts;
+    struct OrbisKernelTimespec ts;
```
roamic: Here the temporary structure is not needed anymore you can just pass `tp` directly to the `orbis_clock_gettime`.

src/core/libraries/kernel/time.cpp
diegolix29: yes im waiting for people to test games here so i know i didnt regress anything, but the issues on windows for those games gets fixed with this, what i dont know cause i dont have linux or mac if this dosent change anything on that end regarding what ur pr and daniel one fixed so probably tomorrow i will ask some folks to try games here. Thanks for the feedback the line i changed here i had to cause without it only window compiled so practically its done so linux and mac could compile

src/core/libraries/kernel/time.cpp
diegolix29: yes give me a sec

src/core/libraries/kernel/time.cpp
diegolix29: ah nice i was doing one by one change to see what worked and didnt realize this was reduntant

diegolix29:
Let me know if you could do some testing on linux to see if this dosent regress anything? at least on my end dosent, bb effects work ok, the games boots and save files are on good time

diegolix29:
image

hollow knight seems to be able to load my saves with no issues too

roamic (approved):
Needs testing, but otherwise LGTM.

diegolix29:
> Needs testing, but otherwise LGTM.

nice thanks i will ask people to test it on general tomorrow, for the 3 systems. hopefully should be ok

georgemoralis:
important for 90% of the users : does it work with bloodborne?:D

diegolix29:
Haha the first thing on the list to see of it improves or brakes lol haha

Missake212:
Date still shows up wrong in NieR Replicant (not sure if that's related)

image

Here's the log if that helps 
[shad_log.txt](https://github.com/user-attachments/files/19362456/shad_log.txt)

diegolix29:
Is working right now with the current implementation of roamic? 

> Date still shows up wrong in NieR Replicant (not sure if that's related)
> 
> image
> 
> Here's the log if that helps [shad_log.txt](https://github.com/user-attachments/files/19362456/shad_log.txt)

Missake212:
Not sure I understand but if the current implementation you're talking about is on main then it's the same thing there 

image

diegolix29:
Yes his merge fixes some saves on games that had the wrong one. So that one didnt  fixed nier. Ok thanks at least is not regressing that one. 🙏

gandalfthewhite19890404:
image
Stardew Valley now fixed - boots fast without issues

diegolix29:
Nicee thanks for testing 🙏

PR #2660, opened 2025-03-20
Sync text editor changes to KBM GUI

rainmakerv3:
Loads the saved changes to the text editor in the GUI. Otherwise the non-updated values in the GUI will overwrite the changes made in the text editor if save or apply is pressed.

### DIFF
--- a/src/qt_gui/kbm_gui.cpp
+++ b/src/qt_gui/kbm_gui.cpp
@@ -15,7 +15,6 @@
 #include "ui_kbm_gui.h"
 
 HelpDialog* HelpWindow;
-
 KBMSettings::KBMSettings(std::shared_ptr<GameInfoClass> game_info_get, QWidget* parent)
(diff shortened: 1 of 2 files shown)


diegolix29:
what this changes suppost to do? the behaviour u discribing is how it right now, if u edit somehting and then to exit the gui u press save as normally it would, it will overwrite anything u typed on text editor. Same happens here, so i dont know what was the intend of the pr

diegolix29:
right nw u need to press cancel under gui to keep all the changes done in the text editor. and ur pr does the same

rainmakerv3:
> right nw u need to press cancel under gui to keep all the changes done in the text editor. and ur pr does the same

I don't know what you're saying, the changes are syncing correctly now with PR

diegolix29:
yeah my bad seems like i placed the exe elsewhere when doing the test, it pass it to the gui now thanks

PR #2667, opened 2025-03-22
shader_recompiler: Improve divergence handling and readlane elimintation

raphaelthegreat:
Resolves a bunch of issues I found, needs a lot of testing for regressions

### DIFF
--- a/src/shader_recompiler/recompiler.cpp
+++ b/src/shader_recompiler/recompiler.cpp
@@ -1,9 +1,6 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
-#include "common/config.h"
-#include "common/io_file.h"
-#include "common/path_util.h"
 #include "shader_recompiler/frontend/control_flow_graph.h"
 #include "shader_recompiler/frontend/decode.h"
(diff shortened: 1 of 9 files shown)


src/shader_recompiler/recompiler.cpp
```diff
@@ -63,26 +60,18 @@ IR::Program TranslateProgram(std::span<const u32> code, Pools& pools, Info& info
...
         Shader::Optimization::HullShaderTransform(program, runtime_info);
     } else if (info.l_stage == LogicalStage::TessellationEval) {
-        Shader::Optimization::ConstantPropagationPass(program.post_order_blocks);
         Shader::Optimization::TessellationPreprocess(program, runtime_info);
-        Shader::Optimization::ConstantPropagationPass(program.post_order_blocks);
         Shader::Optimization::DomainShaderTransform(program, runtime_info);
     }
-    Shader::Optimization::ConstantPropagationPass(program.post_order_blocks);
-    Shader::Optimization::RingAccessElimination(program, runtime_info, stage);
-    Shader::Optimization::ConstantPropagationPass(program.post_order_blocks);
```
squidbus, line 85: Should there still be one of these toward the end? I’ve observed usually that various passes after the first constant propagation may output code that could benefit from a final constant propagation.

PR #2674, opened 2025-03-23
Handle "-patch" as the suffix for game update folders

kalaposfos13:
Some game dumping tools name game updates `CUSAXXXXX-patch`, so the user had to rename that to `CUSAXXXXX-UPDATE` previously. With this PR, those dumps now become plug-and-play. Everything works as far as I could see from my testing, but I do admit that just like a typical programmer I also don't like to test extensively, so some things might be wrong.

### DIFF
--- a/src/core/file_format/pkg.cpp
+++ b/src/core/file_format/pkg.cpp
@@ -350,7 +350,7 @@ bool PKG::Extract(const std::filesystem::path& filepath, const std::filesystem::
                     auto title_id = GetTitleID();
 
                     if (parent_path.filename() != title_id &&
-                        !fmt::UTF(extract_path.u8string()).data.ends_with("-UPDATE")) {
+                        !fmt::UTF(extract_path.u8string()).data.ends_with("-patch")) {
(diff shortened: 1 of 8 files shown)


src/core/file_format/pkg.cpp
```diff
@@ -350,7 +350,7 @@ bool PKG::Extract(const std::filesystem::path& filepath, const std::filesystem::
                     auto title_id = GetTitleID();
 
                     if (parent_path.filename() != title_id &&
-                        !fmt::UTF(extract_path.u8string()).data.ends_with("-UPDATE")) {
+                        !fmt::UTF(extract_path.u8string()).data.ends_with("-patch")) {
```
georgemoralis, line 353: replacing this won't break the ones that ends with -UPDATE?

kalaposfos13: No, for PKG extraction, I just replaced the `-UDPATE` strings with `-patch`, as this is only for the one time extraction code
Everywhere where a previously generated `-UPDATE` folder can also be used was updated to handle both cases, though

PR #2677, opened 2025-03-23
Multykey

tlarok:
Updated GUI to handle multi-key inputs and handling right keys, wasnt tested with linux


thecatontheceiling:
code is fully AI generated

PR #2686, opened 2025-03-25
HLE discmap

georgemoralis:
This eliminates the need for discmap , based on @red-prig suggestions

According to RE on ghidra all calls seems to check if lib is init first using a global

`  if (DAT_010080e0 == 0) {
    uVar2 = ORBIS_DISC_MAP_ERROR_NO_BITMAP_INFO;
  }`

so it is rather safe to return this , since it is likely the same we do when we load this lib lle

### DIFF
--- a/src/core/libraries/disc_map/disc_map.cpp
+++ b/src/core/libraries/disc_map/disc_map.cpp
@@ -13,11 +13,11 @@ int PS4_SYSV_ABI sceDiscMapGetPackageSize(s64 fflags, int* ret1, int* ret2) {
     return ORBIS_DISC_MAP_ERROR_NO_BITMAP_INFO;
 }
 
-int PS4_SYSV_ABI sceDiscMapIsRequestOnHDD(char* path, s64 param2, s64 param3, int* ret) {
+int PS4_SYSV_ABI sceDiscMapIsRequestOnHDD(char* path, s64 offset, s64 nbytes, int* ret) {
     return ORBIS_DISC_MAP_ERROR_NO_BITMAP_INFO;
 }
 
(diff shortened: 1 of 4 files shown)


src/core/libraries/disc_map/disc_map.cpp
```diff
@@ -9,29 +9,29 @@
...
 
-int PS4_SYSV_ABI sceDiscMapGetPackageSize() {
-    LOG_WARNING(Lib_DiscMap, "(DUMMY) called");
+int PS4_SYSV_ABI sceDiscMapGetPackageSize(s64 fflags, int* ret1, int* ret2) {
     return ORBIS_DISC_MAP_ERROR_NO_BITMAP_INFO;
 }
 
-int PS4_SYSV_ABI sceDiscMapIsRequestOnHDD() {
-    LOG_WARNING(Lib_DiscMap, "(DUMMY) called");
+int PS4_SYSV_ABI sceDiscMapIsRequestOnHDD(char* path, s64 param2, s64 param3, int* ret) {
```
polybiusproxy:
```diff
-int PS4_SYSV_ABI sceDiscMapIsRequestOnHDD(char* path, s64 param2, s64 param3, int* ret) {
+int PS4_SYSV_ABI sceDiscMapIsRequestOnHDD(char* path, s64 arg1, s64 arg2, int* ret) {
```
I think it'd look better this way, same with the rest of functions with unknown arguments.

georgemoralis: fixed

squidbus (approved):

PR #2060, opened 2025-01-05
sceNet

georgemoralis:
New implementation still initial

PR #2687, opened 2025-03-26
Implement `sceHttpWaitRequest`

DelishusCake:


georgemoralis:
that actually can't call it implementation just a quick hack around it . Are you based it on a specific game?

DelishusCake:
> that actually can't call it implementation just a quick hack around it . Are you based it on a specific game?

I'm actually going to close this PR for now. This was just a stub while I worked on a more complete solution, but that requires `sceNetGetsockopt` and `sceNetResolverGetError`, both of which remain unimplemented until #2060 is complete and merged.

georgemoralis:
ok yes i already try to RE http lib , you are right about the above functions if you want to help here is my https://github.com/shadps4-emu/shadPS4/pull/2666

PR #2693, opened 2025-03-27
[Hack] P.T workarounds

kalaposfos13:
"workarounds , use it on your own risk"
blame @georgemoralis for this one, he came up with the idea for this :)


georgemoralis:
As said this is a workaround version , full of hacks . WE don't support it , don't ask support for it . Fixes are probably only affecting P.T so might have unexpected behaviour in other games.

kalaposfos13:
It breaks rendering in a lot of other games, because it has a hack to intentionally mess up textures, as due to an issue somewhere, all the textures are black, so without that, the game looks like it doesn't render anything (it does, but pitch black on pitch black doesn't really stand out), and with the hack, the textures look like memory garbage, but that's still better than not seeing anything

src/video_core/texture_cache/image.cpp
```diff
@@ -246,7 +246,7 @@ boost::container::small_vector<vk::ImageMemoryBarrier2, 32> Image::GetBarriers(
                 // NOTE: these loops may produce a lot of small barriers.
                 // If this becomes a problem, we can optimize it by merging adjacent barriers.
                 const auto subres_idx = mip * info.resources.layers + layer;
-                ASSERT(subres_idx < subresource_states.size());
+                // ASSERT(subres_idx < subresource_states.size());
                 auto& state = subresource_states[subres_idx];
```
StevenMiller123: I know these are hacks and all, but at least prevent the OOB memory access here.

polybiusproxy: That probably explains some of the rendering errors.

kalaposfos13:
Aight April Fools is almost over so no more hack builds for y'all

PR #2696, opened 2025-03-27
Feature/initial ps4 ime keyboard

w1naenator:
Initial system keyboard.
image


polybiusproxy:
Keep in mind there's the Ime library (don't confuse with ImeDialog!) which should also have a virtual keyboard, therefore this code should be moved to a common place to have implementions on both libraries.

w1naenator:
> Keep in mind there's the Ime library (don't confuse with ImeDialog!) which should also have a virtual keyboard, therefore this code should be moved to a common place to have implementions on both libraries.

Got it! I'm still a newbie, but I'll check it out and try to fix it!

w1naenator:
Did I understand correctly - the keyboard is part of the IME library and it opens as a separate module from the input dialog?

polybiusproxy:
Ime and ImeDialog are pretty much the same. Both libraries call VSH functions to open the virtual keyboard so under the hood it's the same code, if I recall correctly.

w1naenator:
> Ime and ImeDialog are pretty much the same. Both libraries call VSH functions to open the virtual keyboard so under the hood it's the same code, if I recall correctly.

So, can I just duplicate my code in both places?

polybiusproxy:
You could but it'd be better to not duplicate code, and do what I said and make a common implementation that can be used by both libraries.

src/core/libraries/ime/ime_dialog_ui.cpp
```diff
@@ -337,6 +340,93 @@ void ImeDialogUi::DrawMultiLineInputText() {
...
+    draw_row(kb_mode == KeyboardMode::Letters ? row1_letters : row1_symbols, 20.0f);
+    draw_row(kb_mode == KeyboardMode::Letters ? row2_letters : row2_symbols, 35.0f);
+    draw_row(kb_mode == KeyboardMode::Letters ? row3_letters : row3_symbols, 80.0f);
+
+    SetCursorPosX(20.0f);
+
+    // Fix: safely push/pop style only if shift was enabled before clicking
+    bool highlight = shift_enabled;
+    if (highlight)
+        PushStyleColor(ImGuiCol_Button, ImVec4(0.2f, 0.6f, 1.0f, 1.0f));
```
polybiusproxy:
```diff
-    if (highlight)
+    if (highlight) {
        PushStyleColor(ImGuiCol_Button, ImVec4(0.2f, 0.6f, 1.0f, 1.0f));
+    }
```

src/core/libraries/ime/ime_dialog_ui.cpp
```diff
@@ -337,6 +340,93 @@ void ImeDialogUi::DrawMultiLineInputText() {
...
+
+    if (Button("SHIFT", ImVec2(75, 35))) {
+        shift_enabled = !shift_enabled;
+    }
+
+    if (highlight)
+        PopStyleColor();
```
polybiusproxy:
```diff
-    if (highlight)
+    if (highlight) {
        PopStyleColor();
+    }
```

src/core/libraries/ime/ime_dialog_ui.cpp
```diff
@@ -337,6 +340,93 @@ void ImeDialogUi::DrawMultiLineInputText() {
     }
 }
 
+void ImeDialogUi::DrawKeyboard() {
+    static bool shift_enabled = false;
+    enum class KeyboardMode { Letters, Symbols };
```
polybiusproxy: Don't define enum types inside of a function, since this will be moved to a common implementation, move this to a header file.

src/core/libraries/ime/ime_dialog_ui.cpp
```diff
@@ -337,6 +340,93 @@ void ImeDialogUi::DrawMultiLineInputText() {
...
+    const char* row2_letters = "ASDFGHJKL";
+    const char* row3_letters = "ZXCVBNM";
+
+    const char* row1_symbols = "1234567890";
+    const char* row2_symbols = "!@#$%^&*()";
+    const char* row3_symbols = "-_=+[]{}";
+
+    auto draw_row = [&](const char* row, float offset_x) {
+        SetCursorPosX(offset_x);
+        for (int i = 0; row[i]; ++i) {
```
polybiusproxy:
```diff
-        for (int i = 0; row[i]; ++i) {
+        for (int i = 0; row[i] != '\0'; i++) {
```
Wouldn't add implicit checks for these things.

polybiusproxy (changes requested):
Reviewing what I've noticed so far.

w1naenator:
> You could but it'd be better to not duplicate code, and do what I said and make a common implementation that can be used by both libraries.

OK, as I see it, the Ime and ImeDialog classes already contains a lot of duplicated code.

polybiusproxy:
I'd still like Ime having the virtual keyboard as well.

cuesta4:
Lovely!

w1naenator:
Do you know booting game where Ime used? I need for test.

polybiusproxy:
I can't think of any games off the top of my head.

w1naenator:
What if I just add a quick copy of the keyboard to IME for now (though I’m not sure how to test it), since I’ve actually started working on the proper layout (closer to the original OSK look) and gamepad implementation for it?

w1naenator:
I currently only have access to PS5 Remote Play, so I can't see the full keyboard layout. Can anyone explain the meaning of the green-marked buttons? Also, what does the blue-marked button do — is it a second symbol layout? If possible, could someone share screenshots of all EN layouts? I couldn’t find them anywhere online.
IMG-20250328-WA0001

polybiusproxy:
The blue marked button is there to navigate through the keyboard's symbols. It's kind of like the "?123" button you'd have on a phone keyboard.

w1naenator:
image
Still in a very early state.
Gamepad input not implemented.
Cursor movement not implemented.
I'm using characters for labels, but the font used in ImGui doesn't support advanced characters.
For now, I've replaced the characters (icons) with abbreviations.
I also don't yet know all symbols layouts I need to implement.

w1naenator:
I need help, don't understand how to handle controller buttons pressed.
`auto current_pad_button = Common::Singleton<Input::GameController>::Instance()->GetLastState().buttonsState;` returns nothing

georgemoralis:
i think you should look at imgui buttons that emu buttons @viniciuslrangel  might can help

CMakeSettings.json
```diff
@@ -6,7 +6,7 @@
       "configurationType": "Release",
       "buildRoot": "${projectDir}\\Build\\${name}",
       "installRoot": "${projectDir}\\Install\\${name}",
-      "cmakeCommandArgs": "",
+      "cmakeCommandArgs": "-DENABLE_QT_GUI=ON -DCMAKE_PREFIX_PATH=C:\\Qt\\6.8.2\\msvc2022_64",
```
mailwl: remove this change

StevenMiller123: Given that this is a draft, I'd assume that change was just something temporary they did.

diegolix29:
> I need help, don't understand how to handle controller buttons pressed. `auto current_pad_button = Common::Singleton<Input::GameController>::Instance()->GetLastState().buttonsState;` returns nothing

iirc it should be onkeypress or onpresskey from sdl, so u can link it from it

viniciuslrangel:
> I need help, don't understand how to handle controller buttons pressed.
> `auto current_pad_button = Common::Singleton<Input::GameController>::Instance()->GetLastState().buttonsState;` returns nothing

If any imgui window has focus, it will consume all events from sdl so it won't reach the game running. Inside imgui there's function like ImGui::IsKeyPressed that should work as you intend

w1naenator:
Have no a lot time, but still in progress.

georgemoralis:
it appears that ImGuiNavInput is not available with latest imgui update , @w1naenator check if you can use something alternative

w1naenator:
It looks a bit more authentic now, though shortcut keys are still not working. I removed that functionality from the IME for now, as I’m not sure how to detect their usage, and I don’t know of any games that actually use them.
I'm still in need of screenshots showing the symbols and accented letters layouts—if anyone could provide those, it would be very helpful.
Screenshot_20250421-113925 (1)

diegolix29:
the shortcuts stuff can u help a bit with that? @kalaposfos13  i think u have a better understanding on inputs iirc.

w1naenator:
> > I need help, don't understand how to handle controller buttons pressed.
> > `auto current_pad_button = Common::Singleton<Input::GameController>::Instance()->GetLastState().buttonsState;` returns nothing
> 
> If any imgui window has focus, it will consume all events from sdl so it won't reach the game running. Inside imgui there's function like ImGui::IsKeyPressed that should work as you intend

@viniciuslrangel already suggested that I use ImGui::IsKeyPressed(), but I haven't tried it yet due to limited free time.

w1naenator:
Undertale now open ImeDialog.
Screenshot_20250427-033210 (1)

w1naenator:
@georgemoralis, can I add the functionality needed for IME to src/imgui/?
For example, I need the ability to keep the cursor always visible, set its position, and other related features.

georgemoralis:
yes sure

w1naenator:
I haven't found a way to control the built-in caret when the text input field is not focused, so I draw my own caret. I'm not sure if it's a good idea or not.

georgemoralis:
> I haven't found a way to control the built-in caret when the text input field is not focused, so I draw my own caret. I'm not sure if it's a good idea or not.

i think it's ok

LNDF:
> Do you know booting game where Ime used? I need for test.

Metaphor and nier automata I think use ime.

Also, keep in mind that there are different keyboard layouts depending on the county.

I haven't looked into ime for a long way but iirc the difference between ime and imedialog is that, while imedialog provides a dialog for entering text and after finished the text is returned to the application, ime is a full input library. Providing physical keyboard support, event based input (like when entering a character for example), and additionally provides an on screen keyboard that the game can request (the keyboard doesn't need to have a text box).

For example, patapon oses both ime and imedialog.
Imedialog is used to ask for the name of the main character and ime is used to poll a physical USB keyboard if one is connected (maybe for keyboard controls?) and, while it uses ime, it doesn't use the on screen keyboard that ime provifes.

Another thing to take into account is that games that use ime and the on screen keyboard tend to rende the text that the user is typing themselves (since they handle eventswhen you insert a character). Currently, the text is shown twice in games that use ime. Onec on the text box and another time where the game itself renders it.

StevenMiller123:
For reference, Metaphor: ReFantazio cannot be dumped until we've got an updated jailbreak.

LNDF:
> For reference, Metaphor: ReFantazio cannot be dumped until we've got an updated jailbreak.

The I might have confused it for another game...

w1naenator:
@LNDF 

As I understand, there are additional problems with hieroglyph input, such as Japanese and Chinese languages, because it is not just a layout with symbols. It requires backend support to select the correct character based on typing basic symbols. So, maybe I should start with English first.

I am a hardware programmer, not a software developer, so it is difficult for me to understand how to organize data exchange between GUI layers and the backend. Maybe someone knows how to do it and can explain me. I also have very limited free time to work on this keyboard implementation.

Also, I am still away from home and do not have access to my PS4, so I still cannot get the correct keyboard layouts.

I also found that the last two rows — the function keys — might not need to be part of the layout, because their behavior must change depending on certain conditions. For example, the Shift key has three states: unshifted, shifted, and caps locked. There is also a New Line button that appears in multiline mode. The Done button has multiple labels, such as OK, Search, and Go, and if I remember correctly, it is translated based on the system language.

w1naenator:
There are also other variations of the keyboard, such as a numeric keyboard and special keyboards for email or URL input.
At the moment, I have no idea how to organize all of this.

LNDF:
> There are also other variations of the keyboard, such as a numeric keyboard and special keyboards for email or URL input. At the moment, I have no idea how to organize all of this.

Maybe you can create a base keyboard UI class and inherit from that. You could manage events in a common way since really keyboard only changes visually but internally it is the same.

georgemoralis:
Is this still on works?

w1naenator:
I am very busy at the moment, but I occasionally work on it. Currently, I’m using Ghidra to understand the IME library’s structure and logic.

gandalfthewhite19890404:
can you rebase?

w1naenator:
> can you rebase?

@gandalfthewhite19890404  Done

georgemoralis:
Is that still being working on?

w1naenator:
I'll return to this after the IME fixes are completed. I'm finally starting to understand how to implement the GUI part of the OSK, and many things will need to be redesigned. You can close this request for now—I'll open a new one when I'm ready to continue. At that point, I plan to implement it in both IME and IMEDialog in parallel.

PR #2697, opened 2025-03-27
Fork detection: Fix PR actions only showing HEAD as the branch name

kalaposfos13:
This will probably have quite a few commits of me trying random shit until something finally works, as I don't know how to debug Github runners locally, but when it works, it should make artifacts downloaded from PR actions be more easily identifiable
Also, due to how this works I can't even just test it without opening a PR for it as actions on one's own fork work fine :(


kalaposfos13:
Turns out that the very first try actually worked, so after some cleanup, this PR is now done and can be merged

Issue #2378, opened 2025-02-08
[APP BUG]: "Directory to Install Game Path doesn't handle characters with accents


Exo61:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

When installing ShadPS4 into a directory or path with a character containing an accent (like "É"), the path has to be re-added every time ShadPS4 is being launched.

The main reason is that it replaces the character with an accent by multiple symbols, corrupting the path and therefore not listing games anymore.

Here is a screenshot: https://ibb.co/27x4Stw4

Can you fix this issue? 

Thank you

### Reproduction Steps

1. Add a Directory to Install Game Path containing a symbol with an accent, like "É"
2. Close ShadPS4
3. Re-open ShadPS4
4. Check the the Directory to Install Game Path again and see the symbols
5. Games are not listed anymore until re-adding the path with the right character again

### Expected Behavior

Directory to Install Game Path remains the same everytime ShadPS4 is open, no matter the name of the folders

### Specify OS Version

Windows 11

KnifeOnlyI:
Hello,

I am currently investigating the cause of the problem and can already conclude that it is not due to the saving of the path in the configuration file. In the config.toml file, the path is correctly stored with the accents. Therefore, it is probably the in-memory loading of the string that is causing the issue.

After reopen ShadPS4 : 

![Image](https://github.com/user-attachments/assets/0f19a64c-0656-49f9-8520-936c2299840b)

![Image](https://github.com/user-attachments/assets/71a76f43-4526-433d-90a4-a47e26ea91d1)

The string in `config.toml` file : 

![Image](https://github.com/user-attachments/assets/0040dbf4-2838-4344-adc9-adbc67388363)

If I use a debugger, this is what I get: the string loaded in memory doesn't handle the accents. It would be interesting to see if it's the same on Linux. (`src/common/config.cpp`)

![Image](https://github.com/user-attachments/assets/13d415fa-e226-4928-8aaf-34266b593c3f)

AfonsoMendoncaJacinto:
If I understand this correctly, the solution would be to issue a warning in the frontend, if the desired path has accents?

KnifeOnlyI:
[...]

PR #2699, opened 2025-03-28
Fix support for unicode paths for game install directories

k3dr1:
Addressing [Issue 2378](https://github.com/shadps4-emu/shadPS4/issues/2378), namely not supporting unicode characters in the path to game install folders. Toml already stores them correctly, so by just changing the toml::find_or<> template to std::u8string we can read them correctly (relevant https://github.com/ToruNiina/toml11/issues/104). This is fine because the install paths get turned into std::filesystem::path anyway.

Before:
shadps4_pSlEPKPCh6

After:
shadps4_3ueD4t3aTE
shadps4_N9IpHw6TrE


k3dr1:
Just noticed an identical fix by JooPT (https://github.com/JoooPT/shadPS4/commit/d8b421c74fd2f8ef4d134b5882dbbca65850a54b) in the issue thread. I didn't see it before so I apologize.

k3dr1:
There seems to be a bit more to this. The call to std::filesystem::path::string() fails on Windows when trying to read Japanese characters if locale isn't set correctly (https://stackoverflow.com/questions/78750150/issue-with-stdfilesystempath-conversion-to-stdstring-in-c). We could call std::filesystem::path::u8string() instead to circumvent this issue, but Qt doesn't provide a function to convert u8strings to QStrings (https://bugreports.qt.io/browse/QTBUG-83643). Therefore I decided to convert the path to u8string, then pass a temp constructed std::string off of it to QString::fromStdString()

PR #2707, opened 2025-03-28
externals: Remove need for cryptopp build.

squidbus:
Currently cryptopp is only used for:
* SHA1 for NID calculation.
* AES CBC for trophy decryption.

Since the uses are so small and simple, this PR replaces cryptopp with small single-header libraries for each case, eliminating a large sub-module dependency from the build which both hasn't been maintained in a while and has caused problems in the past.

Tested by:
* Verifying that NIDs for some known function names calculate correctly.
* Using the trophy viewer with a clean game data cache.


abouvier:
You forgot to remove the `find_package(cryptopp 8.9.0 MODULE)` in `CMakeLists.txt`.

PR #2713, opened 2025-03-29
Diplay PR actions as "pr-{number}-{branchname}"

kalaposfos13:
Saw an image on Discord where the branch name was `pr-main`, so I figured I'll add the PR number to the string as well, so now that'll look like `pr-2682-main`, and hopefully this is now really the last thing that was still an issue with fork/branch detection.


topsex88 (commented):
فحل زوجتك

PR #2714, opened 2025-03-29
Replacing const with constexpr

Xphalnos:
Use constexpr because it's more optimized since the value is set at compile time and not at startup.

### DIFF
--- a/src/common/decoder.cpp
+++ b/src/common/decoder.cpp
@@ -15,7 +15,7 @@ DecoderImpl::~DecoderImpl() = default;
 
 std::string DecoderImpl::disassembleInst(ZydisDecodedInstruction& inst,
                                          ZydisDecodedOperand* operands, u64 address) {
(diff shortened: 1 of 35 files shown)


georgemoralis:
that doesn't actually provide something useful so...

thecatontheceiling:
CTRL H "const" to "constexpr"
one of the PR's of all time

thecatontheceiling:
99% of the replacements you made here have negligible impact on performance, `constexpr` exists for a different reason than just performance optimization 

Even `const` variable values are set during compile time if the value is evaluatable by the compiler

Essentially `constexpr` just requires/guarantees that the latter is the case, which is unnecessary in the context of some parts of ShadPS4

This is why replacing every mention of `const` with `constexpr` is useless in this context and it can cause problems if you just spam it everywhere for no reason.

Hope you learnt something new

PR #2719, opened 2025-03-29
Getting rid of the "Separate Update Folder" option

Missake212:
Since the PKG Install is gone, we can get rid of this option in the GUI since it's only purpose was to install PKGs as two different folder instead of overwriting the original game files with the update files. Tested it and the GUI doesn't show the checkbox anymore but please take a quick look to make sure I didn't miss anything, still trying to get familiar navigating in all the files.


tomboylover93:
Did a quick search for "separate" with VS Code and all I could find is Separate Log Files so I think this should be good. The only thing that remains is the en_US.ts file (remove all instances of Separate Update Folder from it and the other .ts files should be updated on the next translation update).

Missake212:
Thank you for looking, should be good now.

PR #2725, opened 2025-03-30
vulkan: Lower list primitive restart warning to debug log.

squidbus:
Lowers the log when list primitive restart is not supported from warning to debug. In practice most games don't actually seem to use list restart and probably just left primitive restart turned on, so it's usually just log noise. Leaving it as a debug log though just in case.


georgemoralis:
Found in macOS builds because MoltenVK doesn't support list primitive restart

PR #2726, opened 2025-03-30
Emulate sceKernelInternalMemory mapping

StevenMiller123:
During game relocation, libkernel performs a call to libSceLibcInternal's _malloc_init function. This function seems to perform many behaviors, most of which only matter for LLE firmware library behaviors.
This PR emulates one behavior of this function, a large flexible memory mapping it performs to the system reserved area of memory. Without handling this, games have more available flexible memory than they may expect.

This fixes a variable underflow on The Last of Us™ Part II startup, allowing it to progress further.

### DIFF
--- a/src/core/linker.cpp
+++ b/src/core/linker.cpp
@@ -109,7 +109,7 @@ void Linker::Execute(const std::vector<std::string> args) {
     void* addr_out{reinterpret_cast<void*>(KernelAllocBase)};
 
     const s32 ret = Libraries::Kernel::sceKernelMapNamedFlexibleMemory(
-        &addr_out, InternalMemorySize, 3, 0, "SceKernelPrimaryTcbTls");
+        &addr_out, InternalMemorySize, 3, 0, "SceKernelInternalMemory");
     ASSERT_MSG(ret == 0, "Unable to perform sceKernelInternalMemory mapping");
 
     main_thread.Run([this, module, args](std::stop_token) {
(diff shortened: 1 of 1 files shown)


src/core/linker.cpp
```diff
@@ -101,6 +101,17 @@ void Linker::Execute(const std::vector<std::string> args) {
...
 
+    // Simulate sceKernelInternalMemory mapping, a mapping usually performed during libkernel init.
+    // Due to the large size of this mapping, failing to emulate it causes issues in some titles.
+    // This mapping belongs in the system reserved area, which starts at address 0x880000000.
+    static constexpr VAddr KernelAllocBase = 0x880000000ULL;
+    static constexpr s64 InternalMemorySize = 0x1000000;
+    void* addr_out{reinterpret_cast<void*>(KernelAllocBase)};
+
+    const s32 ret = Libraries::Kernel::sceKernelMapNamedFlexibleMemory(
+        &addr_out, InternalMemorySize, 3, 0, "SceKernelPrimaryTcbTls");
```
squidbus: Is this the correct name? Seems to be the same as the name used for another mapping below.

StevenMiller123: Oh, good catch. Copy pasted without fixing the name.

squidbus (approved):

squidbus:
Seems good to me code-wise, would like some confirmation no games break from change in memory available.

PR #2739, opened 2025-04-01
shader_recompiler: Use VK_AMD_shader_trinary_minmax when available.

squidbus:
Implements use of `VK_AMD_shader_trinary_minmax` instructions for trinary min/max/median when available. Specifically this PR introduces separate IR instructions for these trinary operations, then determines whether to emit a trinary instruction or a compatibility fill based on what is available. This should reduce the number of instructions needed and possibly improve performance on supported drivers.

Note that these new IR instructions are named with the `Tri` suffix for trinary to avoid awkward naming with numbers like `FPMin332` when the bit suffix is added.

Tested this with a few games on MoltenVK, which supports this extension.


src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -1166,59 +1166,63 @@ void Translator::V_MIN3_F32(const GcnInst& inst) {
...
-void Translator::V_MAX3_U32(bool is_signed, const GcnInst& inst) {
+void Translator::V_MAX3_I32(const GcnInst& inst) {
     const IR::U32 src0{GetSrc(inst.src[0])};
     const IR::U32 src1{GetSrc(inst.src[1])};
     const IR::U32 src2{GetSrc(inst.src[2])};
-    SetDst(inst.dst[0], ir.IMax(src0, ir.IMax(src1, src2, is_signed), is_signed));
+    SetDst(inst.dst[0], ir.SMaxTri(src0, src1, src2));
+}
+
+void Translator::V_MAX3_U32(const GcnInst& inst) {
```
raphaelthegreat: This doesn't have to be duplicated, you could do
```
    SetDst(inst.dst[0], is_signed ? ir.SMaxTri(src0, src1, src2) : ir.UMaxTri(src0, src1, src2));
```

raphaelthegreat: Another idea is to make IREmitter function into IMaxTri and pass is_signed there, similar to how other opcodes do that

squidbus: Do you want me to simplify the others then? I was matching it with min and mid.

raphaelthegreat: If you want to yeah sure

squidbus: Okay, done.

raphaelthegreat (approved):

PR #2745, opened 2025-04-04
build: Target same CPU architecture level as PS4.

squidbus:
Improves the build target CPU architecture by setting to `btver2` (Bobcat Version 2 - the name used by compilers for AMD Jaguar), which matches the actual PS4 CPU. SSE4a is explicitly excluded to not inadvertently use AMD-only instructions.

This enables more instruction set extensions than approximating with `sandybridge` does, increasing support for potential optimizations while still preventing the CPU requirements from exceeding those of games themselves.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,8 +54,8 @@ else()
 endif()
 
 if (ARCHITECTURE STREQUAL "x86_64")
-    # Set x86_64 target level to Sandy Bridge to generally match what is supported for PS4 guest code with CPU patches.
-    add_compile_options(-march=sandybridge)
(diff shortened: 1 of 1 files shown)


NinjaGaiden529:
wow thats amazing george nice work

andre-leonardo:
hell yeah

DiabeticCrab:
Would it make sense to add "-mtune=generic" or "-mtune=haswell" or something along these lines to make sure the generated code performs well on recent-ish CPUs? Using -march alone implies -mtune for the same architecture aswell, which could hurt performance on archs that deviate a lot from btver2.

PR #2749, opened 2025-04-05
renderer_vulkan: Use more depth-stencil dynamic state.

squidbus:
Introduces use of more dynamic depth-stencil state from `VK_EXT_extended_dynamic_state` and `VK_EXT_extended_dynamic_state2`. We already assume we have these as extensions that were promoted to Vulkan 1.3. May help reduce the number of pipelines needed in some cases.

### DIFF
--- a/src/video_core/amdgpu/liverpool.h
+++ b/src/video_core/amdgpu/liverpool.h
@@ -1423,6 +1423,10 @@ struct Liverpool {
             return num_samples;
         }
 
+        bool IsClipDisabled() const {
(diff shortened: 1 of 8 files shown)


raphaelthegreat:
Looks okay. For the future I recommend having some dirty flag system so these are only called when the guest GPU state changes instead of every draw. It's not needed quite yes but it will be, if more states are added

squidbus:
Yeah I’ve been thinking about that, will probably implement something in next PR.

PR #2751, opened 2025-04-05
renderer_vulkan: Only update dynamic state when changed.

squidbus:
Implements dynamic state tracking to only emit state update commands when necessary.

Details:
* Keeps track of all dynamic state values along with corresponding dirty flags.
* Dirty flags are only set when a value changes or the command buffer changes, to apply the state to the new command buffer.
* State is only set in the command buffer when needed; for example, if stencil test is disabled, updating any stencil test parameters is skipped. Once it is enabled, any deferred dirty state will be set.

This system will allow us to continue expanding our use of dynamic state where possible without ballooning the number of commands issued for every draw.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -946,19 +946,19 @@ void Rasterizer::UnmapMemory(VAddr addr, u64 size) {
     mapped_ranges -= boost::icl::interval<VAddr>::right_open(addr, addr + size);
 }
 
-void Rasterizer::UpdateDynamicState(const GraphicsPipeline& pipeline) {
+void Rasterizer::UpdateDynamicState(const GraphicsPipeline& pipeline) const {
     UpdateViewportScissorState();
     UpdateDepthStencilState();
 
-    const auto& regs = liverpool->regs;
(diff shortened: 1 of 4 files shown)


raphaelthegreat (approved):
This is good solution 👍🏻 Another approach I have seen would be to have dirty flags against guest registers instead of individual pieces of GPU state. For example have a big bitset for all regs and set bits when SetShReg/SetContextReg/... are processed. If some state (i.e color buffers) has a lot of registers a second bitset can be used for state "categories" that can be checked quickly and set from a lookup table that maps indices from the main bitset to that one. It's more complex but the advantage is that its backend agnostic and it covers all possible state. However this also good in that its nice and simple

squidbus:
I did consider something using guest registers but I wanted to keep it a bit simple, where we can have one part of the code focus on converting to host state as needed and another focused on making sure that host state is only changed as needed. It’s maybe something we could consider in the future if it has enough benefits.

As far as being backend-agnostic goes, I didn’t consider it a strong goal as different backends will likely have different capabilities for what can be dynamic state. Maybe something where all the state can be easily dirty tracked for the backend to pick whatever it can best make use of, but that also might be excessive as I imagine a lot of the registers won’t be needed.

PR #2752, opened 2025-04-06
build: Move versioning to CMake file.

squidbus:
Moves the version and release variables to the main CMake file, configured into the generated `scm_rev.cpp` for source access. This allows properly versioning the macOS release bundle without maintaining two different version strings.


squidbus:
(also in the future it might be a good idea to pull this from the current git tag or something, but I wanted to keep this simple for now)

abouvier:
Why not put the version in `project()` (and use `CMAKE_PROJECT_VERSION`) instead of using a new variable?

squidbus:
As far as I can tell from the docs, that does not support arbitrary version strings (e.g. how we add “WIP”) and requires a numerical version number only.

georgemoralis (approved):
Looking fine to me

PR #2758, opened 2025-04-07
Add information about firmware files (sys modules) in the README

Missake212:
Adding a simple mention about system modules in the README, it was already on the compatibility list repository but it would make more sense to also have it on the main repository README so that more people are aware about their existence. Maybe it could be placed somewhere else in the README I just decided to place it below the KMB mappings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ R3 | M |
 Keyboard and mouse inputs can be customized in the settings menu by clicking the Controller button, and further details and help on controls are  also found there. Custom bindings are saved per-game. Inputs support up to three keys per binding, mouse buttons, mouse movement mapped to joystick input, and more.
 
 
-## Firmware files:
+# Firmware files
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -122,6 +122,27 @@ R3 | M |
...
+# Firmware files
+
+shadPS4 can load some PlayStation 4 firmware files, these must be dumped from your legally owned PlayStation 4 console.\
+The following firmware modules are supported and must be placed in shadPS4's `user/sys_modules` folder.
+
+<div align="center">
+
+| Modules                 | Modules                 | Modules                 | Modules                 |  
+|-------------------------|-------------------------|-------------------------|-------------------------|  
+| libSceCesCs.sprx        | libSceDiscMap.sprx      | libSceFont.sprx         | libSceFontFt.sprx       |
```
squidbus: `libSceDiscMap.sprx` is no longer loaded LLE and can be removed from here

Missake212: Nice catch thank you.

PR #2762, opened 2025-04-08
[Libs] Http HLE (part2)

georgemoralis:
More HLE implementation of http lib


src/core/libraries/network/http.cpp
```diff
@@ -9,6 +9,26 @@
 
 namespace Libraries::Http {
 
+void NormalizeAndAppendPath(char* dest, char* src) {
+    char* lastSlash;
+    size_t length;
+
+    lastSlash = strrchr(dest, 0x2f);
```
illusionyy:
```suggestion
    lastSlash = strrchr(dest, '/');
```

georgemoralis: yah i was lazy on RE :D

georgemoralis:
more RE stuff :D

georgemoralis:
I am still working on this, but being busy with other stuff as well so if someone what to add something i don't mind :D

PR #2767, opened 2025-04-10
shader_recompiler: (WIP) Implement more accurate ReadConst support (including dynamic indexing)

LNDF:
The goal of this pull request is to be able to more accurately support ReadConst (LOAD_DWORD) in shaders.
The current implementation doesn't support dynamic offsets, and doesn't correctly handle the cases where the base address is modified between the source ReadConst/UserData and the user ReadConst. Example:
```
%20 = ReadConst %19 #0
%21 = IAdd32  %20 #8
%22 = CompositeConstructU32x2 %21 #0
%23 = ReadConst %22 #0      <--- Here we dont take the addition in %21 into account
```
This PR aims to solve this issue in the following way.

1. First all the ReadConsts in the shader being compiled are collected in a list
2. A sub IR-Program is created from the shader IR-Program that contains all the relevant instructions to compute the ReadConsts. This includes all arguments of each ReadConst and so recursively, the necessary conditionals and loops, etc.
3. Subprogram is aanalyzed to extract how many times each ReadConst can be executed (it it's inside a loop, it may be executed more than once and read different values). That way we can decide offsets for the ReadConst data in the flatbuf (extended user data)
4. Subprogram is modified to introduce instructions to save the result from each ReadConst into the flatbuf. Also, the ReadConsts in the original shader IR are modified to point to the correct location in the flatbuf.
5. Subprogram is compiled to x64 assembly and run every time the flatbuf is refreshed (before drawing)
Doing that, the flatbuf should contain all the data the shader needs.

## Current issues and status
This is a draft, because it regresses/braks games right now.
I was able to toest in 2 games.
* Gran Turismo Sport: After doing the initial brighness configuration screen, I'm able to continue further and see a washed out version fo the initial credits, press circle and hit a "Attempted to track non-GPU memory address". In current main, I cannot make out from the initial configuration screen (new save) and I see a black screen during the initial credits, to hit the same memory error if I press circle.
* Bloodborne: Currently, there is an issue in with a specific shader that I'm not able to resolve.

### Bloodborne issue
A crash occurs when executing the x64 code generated for the 0x8bc6ea32 shader. The following is the IR of the subprogram:
```
Block $0
         Prologue
%8     = GetUserData SGPR6 (uses: 4)
%9     = GetUserData SGPR7 (uses: 4)
%10    = CompositeConstructU32x2 %8, %9 (uses: 4)
%11    = ReadConst (flags=0)  %10, #0 (uses: 2)
         SetUserData #16, %11
%12    = ReadConst (flags=0)  %10, #1 (uses: 2)
         SetUserData #17, %12
%13    = ReadConst (flags=0)  %10, #2 (uses: 2)
         SetUserData #18, %13
%14    = ReadConst (flags=0)  %10, #3 (uses: 2)
         SetUserData #19, %14
%15    = CompositeConstructU32x2 %8, %9 (uses: 4)
%16    = ReadConst (flags=0)  %15, #16 (uses: 2)
         SetUserData #20, %16
%17    = ReadConst (flags=0)  %15, #17 (uses: 2)
         SetUserData #21, %17
%18    = ReadConst (flags=0)  %15, #18 (uses: 2)
         SetUserData #22, %18
%19    = ReadConst (flags=0)  %15, #19 (uses: 2)
         SetUserData #23, %19
%20    = CompositeConstructU32x4 %11, %12, %13, %14 (uses: 1)
%21    = ReadConstBuffer %20, #32 (uses: 1)

Block $1
%22    = CompositeConstructU32x2 %8, %9 (uses: 8)
%23    = ReadConst (flags=0)  %22, #4 (uses: 1)
         SetUserData #24, %23
%24    = ReadConst (flags=0)  %22, #5 (uses: 1)
         SetUserData #25, %24
%25    = ReadConst (flags=0)  %22, #6 (uses: 1)
         SetUserData #26, %25
%26    = ReadConst (flags=0)  %22, #7 (uses: 1)
         SetUserData #27, %26
%27    = ReadConst (flags=0)  %22, #8 (uses: 1)
         SetUserData #28, %27
%28    = ReadConst (flags=0)  %22, #9 (uses: 1)
         SetUserData #29, %28
%29    = ReadConst (flags=0)  %22, #10 (uses: 1)
         SetUserData #30, %29
%30    = ReadConst (flags=0)  %22, #11 (uses: 1)
         SetUserData #31, %30

Block $2
%31    = IEqual32 %21, #0 (uses: 1)
%32    = LogicalNot %31 (uses: 1)
%33    = ConditionRef %32 (uses: 0) (goes to block $5 if false)

Block $3
%34    = CompositeConstructU32x4 %16, %17, %18, %19 (uses: 2)
%35    = ReadConstBuffer %34, #18 (uses: 1)
%36    = ReadConstBuffer %34, #19 (uses: 1)

Block $4
%37    = Phi [ %35, {Block $3} ], [ %8, {Block $2} ] (uses: 1)
%38    = Phi [ %36, {Block $3} ], [ %9, {Block $2} ] (uses: 1)

Block $5
%39    = CompositeConstructU32x2 %37, %38 (uses: 4)
%40    = ReadConst (flags=0)  %39, #12 (uses: 1)
         SetUserData #32, %40
%41    = ReadConst (flags=0)  %39, #13 (uses: 1)
         SetUserData #33, %41
%42    = ReadConst (flags=0)  %39, #14 (uses: 1)
         SetUserData #34, %42
%43    = ReadConst (flags=0)  %39, #15 (uses: 1)
         SetUserData #35, %43

Block $6
         Epilogue
```
If condition %33 is true, the shader gets an address (%35 and %36) that will be used to read from later (%40-%43). 
The address obtained in %35 and %36 is not a valid address (0xC28C270A00000030). The buffer resource used to point to the buffer that is read from seems correct (base address, stride, etc.). I think the ReadConstBuffer implementation in the x64 shader backend is correct (apart from clamping)and I'm not able to find the cause of this issue.
Would be great if someone with more knowledge on how buffers work could take a look at this.

## Additional notes
Now, if shader dumping is enabled, the ASL will also be dumped. Additionaly, IR, ASL and assembly code will be dumped for the subprogram.


StevenMiller123:
image
image
image
image
image
image
image
image

Almost everything I can test regresses from this. Usually it's either completely broken rendering, a new unreachable, or some exception in my AMD drivers that wasn't happening before.
I started testing a while ago, so if anything's been fixed in the last half hour, I haven't tested that yet.

I'm not providing logs since nothing useful seems to get logged beyond the messages shown in the image. If there's any specific case you want a log for, let me know and I'll send one.

LNDF:
Please test with the latest commit again if you can. But this is a work in progress so it is expected to regress games.

Latest commit will fix the second and 3rd errors. I'm interested on the readconstbuffer. It may be helpful to fix the segfault in Bloodborne

StevenMiller123:
I'm still facing some regressions, but all regressions I showed in the first post are now fixed.

A couple games so far are now crashing on this error:
image

Persona 3 Reload's still regressed, but with a new error:
image

baggins183:
Thanks for working on this!
The readconstbuffer thing is weird, because IIRC when reading indirect user data/SRTs only S_LOAD_DWORD variants should be used, not S_BUFFER_LOAD_DWORD. Also I dont think bloodborne does anything horrible when it comes to user data indirection. But its been a while since I looked at this stuff so I might be confused. 

I can review more on the weekend, will have to refresh a lot

baggins183:
/*0000000002f8*/ s_buffer_load_dwordx8 s[0:7], s[16:19], 0xc
...
/*000000000334*/ s_branch        .L1308_0
.L824_0:
/*000000000338*/ v_mul_lo_i32    v0, v13, 3
/*000000000340*/ s_load_dwordx4  s[0:3], s[6:7], 0xc

I looked at the shader and it looks like the phi arg is coming from this s_buffer_load_dwordx8, but the s_load_dwordx4 shouldnt be reachable from there because theres an unconditional branch over it (.L1308_0). So Im guessing its a problem with CFG structurization

LNDF:
> I'm still facing some regressions, but all regressions I showed in the first post are now fixed.
> 
> A couple games so far are now crashing on this error: image
> 
> Persona 3 Reload's still regressed, but with a new error: image

Both of those issues should be fixed with latest commit. Thaks for the testing work!

> /_0000000002f8_/ s_buffer_load_dwordx8 s[0:7], s[16:19], 0xc ... /_000000000334_/ s_branch .L1308_0 .L824_0: /_000000000338_/ v_mul_lo_i32 v0, v13, 3 /_000000000340_/ s_load_dwordx4 s[0:3], s[6:7], 0xc
> 
> I looked at the shader and it looks like the phi arg is coming from this s_buffer_load_dwordx8, but the s_load_dwordx4 shouldnt be reachable from there because theres an unconditional branch over it (.L1308_0). So Im guessing its a problem with CFG structurization

Will take a look today later, thanks.

Missake212:
I know it's WIP but just letting you know what's currently broken, hopefully this can help you figure it out:

Nier Replicant
[CUSA18742.log](https://github.com/user-attachments/files/19680658/CUSA18742.log)

The Last of Us Remastered
[CUSA00556.log](https://github.com/user-attachments/files/19680669/CUSA00556.log)

Uncharted: The Nathan Drake Collection (not sure if broken or if it hits a new/different unreachable)
[CUSA02826.log](https://github.com/user-attachments/files/19680694/CUSA02826.log)

The Last Guardian (nothing shows up in the log though)

LNDF:
> I know it's WIP but just letting you know what's currently broken, hopefully this can help you figure it out:
> 
> Nier Replicant [CUSA18742.log](https://github.com/user-attachments/files/19680658/CUSA18742.log)
> 
> The Last of Us Remastered [CUSA00556.log](https://github.com/user-attachments/files/19680669/CUSA00556.log)
> 
> Uncharted: The Nathan Drake Collection (not sure if broken or if it hits a new/different unreachable) [CUSA02826.log](https://github.com/user-attachments/files/19680694/CUSA02826.log)
> 
> The Last Guardian (nothing shows up in the log though)

Could you send me the following shader dumps?
* 0x43b8ee5e for Uncharted: The Nathan Drake Collection
* 0x42cb16a5  for The Last of Us (does this work in main?)
* 0xc4fcd50f  for Nier

Will take a look at them later today

raphaelthegreat:
This look like great work!

However I do think this might be a little over-engineered solution when a much simpler one exists. The problem with non-standard use of S_LOAD_DWORD is that the shader is basically accessing a raw GPU virtual address. In vulkan this can be easily addressed by using VK_KHR_buffer_device_address in combination with a page table in following fashion:

* Allocate a 2^(40-14)* 8 = 512MB GPU buffer that stores an 8-byte virtual address for every 16KB page.
* Every time GPU memory is mapped to the address space, use host memory import extension to create a vulkan buffer from that host memory and then retrieve the BDA from that (I haven't tried this but I assume it is possible*) to fill the relevant page table entries.
* S_LOAD_DWORD instructions can now be translated into a simple page table lookup for guest VA -> host BDA and a buffer fetch.

The existing flatten pass still must exist for fetching sharps, as those must be available at compile time. It can be simplified and combined into resource tracking pass so it only works on ReadConst nodes that are sharp loads. For example the flatten pass and first part of resource tracking can be combined into 1 pass, and the second part of resource tracking can be combined with the lower buffer fmt pass into a single resource post-process/resolution pass.

Using a page table like this will allow backend almost any use case of S_LOAD_DWORD (even more weird cases like waterfall loops) without adding massive amounts of platform specific code to handle such cases

The s_buffer_load_dword issue in BB appears to be not a bug per-se, but a result of the structurization having control flow paths that appear reachable in code but are logically unreachable, thus SSA pass creates a phi node. Phi node traversal for resource tracking can be enhanced, by considering reachability on the original unstructured CFG (i.e collect all phi sources and perform a reachability test on them to see which one is the "correct" one)

(*) If deriving a BDA from host memory turns out to be impossible, the buffer cache can be modified to have a "reserve" mode where it creates full buffers for entire gpu mappings, instead of creating buffers on demand. Thus BDAs can be created from GPU memory directly. This is more complex than just using host memory, since it also requires splitting the allocation and tracking granularity of buffer and requires more conservative CPU syncing logical to ensure all pages likely to be accessed by shader have been uploaded correctly.

LNDF:
> This look like great work!
> 
> However I do think this might be a little over-engineered solution when a much simpler one exists. The problem with non-standard use of S_LOAD_DWORD is that the shader is basically accessing a raw GPU virtual address. In vulkan this can be easily addressed by using VK_KHR_buffer_device_address in combination with a page table in following fashion:
> 
>     * Allocate a 2^(40-14)* 8 = 512MB GPU buffer that stores an 8-byte virtual address for every 16KB page.
> 
>     * Every time GPU memory is mapped to the address space, use host memory import extension to create a vulkan buffer from that host memory and then retrieve the BDA from that (I haven't tried this but I assume it is possible*) to fill the relevant page table entries.
> 
>     * S_LOAD_DWORD instructions can now be translated into a simple page table lookup for guest VA -> host BDA and a buffer fetch.
> 
> 
> The existing flatten pass still must exist for fetching sharps, as those must be available at compile time. It can be combined into resource tracking pass so it only works on ReadConst nodes that are sharp loads.
> 
> The s_buffer_load_dword issue in BB appears to be not a bug per-se, but a result of the structurization having control flow paths that appear reachable in code but are logically unreachable, thus SSA pass creates a phi node. Phi node traversal for resource tracking can be enhanced, by considering reachability on the original unstructured CFG (i.e collect all phi sources and perform a reachability test on them to see which one is the "correct" one)
> 
>     * If deriving a BDA from host memory turns out to be impossible, the buffer cache can be modified to have a "reserve" mode where it creates full buffers for entire gpu mappings, instead of creating buffers on demand. Thus BDAs can be created from GPU memory directly. This might also require more conservative CPU syncing logical to ensure all pages likely to be accessed by shader have been uploaded correctly

Thanks. Didn't know about that Vulkan feature. Should this be closed then?

raphaelthegreat:
I think its fine to leave this open until the other solution can be prototyped (if you are willing to work on it that is). Having this also working (but more complex) one might be good for testing stuff

LNDF:
Latest commit fixes bloodborne CFG bug. Thanks @raphaelthegreat and @baggins183 for pointing out.

Regarding the unnecesary Phi nodes added by ssa, this PR adds a ConditionData structure to each block. That structure contains a pointer to the ASL node of the condition or loop that the block is inside of, as well as a pointer to a parent ConditionData structure that holds the parent condition/loop, and so recursively. That way it is very easy to know which conditions/loops need to be true in order for a given block to execute. For example:
```
Block $1
[...]
%10 = ConditionRef %9
Block $2
[...]
%12 = ConditionRef %11
Block $3
[some code]
```
The ConditionData for Block $3 will point to the ASL if node for condition %12, and the parent pointer to the ConditionData of Block $1. Then Block $2's ConditionData will point to the ASL node for condition %10 and the parent would be null. That makes it very easy, given a block, to know which conditions must be true in ordef for that block to be executed.

The above and adding an field to the If ASL node data structure that ponts to an else block (if it exists any) would ease the process of optimizing the SSA pass.

AboMedoz:
### InFAMOUS
image

### Edit: 
Forcing it to move Further will result in `LoadSharedU32` & `LoadSharedU64`

LNDF:
> ### InFAMOUS
> 
> image
> ### Edit:
> 
> Forcing it to move Further will result in `LoadSharedU32` & `LoadSharedU64`

Do both happen on the same shader?
Could you please enable shader dumping and upload the shader(s) in question?

Xcedf:
Uncharted shader if still needed
[43b8ee53.zip](https://github.com/user-attachments/files/19687456/43b8ee53.zip)

StevenMiller123:
@Xcedf Shader binaries are copyrighted, do not share those.

Xcedf:
No bins this time
[43b8ee53.zip](https://github.com/user-attachments/files/19687629/43b8ee53.zip)

brad0demx:
[yookalaylee shaders.zip](https://github.com/user-attachments/files/19687662/yookalaylee.shaders.zip)

CUSA16139 - "Yooka Laylee"

brad0demx:
[CUSA36848 - Lies of P](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/87)
Crashes with signal handler while compiling shaders on this PR 
[lies of P shaders.zip](https://github.com/user-attachments/files/19688534/lies.of.P.shaders.zip)

LNDF:
> This look like great work!
> 
> However I do think this might be a little over-engineered solution when a much simpler one exists. The problem with non-standard use of S_LOAD_DWORD is that the shader is basically accessing a raw GPU virtual address. In vulkan this can be easily addressed by using VK_KHR_buffer_device_address in combination with a page table in following fashion:
> 
>     * Allocate a 2^(40-14)* 8 = 512MB GPU buffer that stores an 8-byte virtual address for every 16KB page.
> 
>     * Every time GPU memory is mapped to the address space, use host memory import extension to create a vulkan buffer from that host memory and then retrieve the BDA from that (I haven't tried this but I assume it is possible*) to fill the relevant page table entries.
> 
>     * S_LOAD_DWORD instructions can now be translated into a simple page table lookup for guest VA -> host BDA and a buffer fetch.
> 
> 
> The existing flatten pass still must exist for fetching sharps, as those must be available at compile time. It can be simplified and combined into resource tracking pass so it only works on ReadConst nodes that are sharp loads. For example the flatten pass and first part of resource tracking can be combined into 1 pass, and the second part of resource tracking can be combined with the lower buffer fmt pass into a single resource post-process/resolution pass.
> 
> Using a page table like this will allow backend almost any use case of S_LOAD_DWORD (even more weird cases like waterfall loops) without adding massive amounts of platform specific code to handle such cases
> 
> The s_buffer_load_dword issue in BB appears to be not a bug per-se, but a result of the structurization having control flow paths that appear reachable in code but are logically unreachable, thus SSA pass creates a phi node. Phi node traversal for resource tracking can be enhanced, by considering reachability on the original unstructured CFG (i.e collect all phi sources and perform a reachability test on them to see which one is the "correct" one)
> 
> (*) If deriving a BDA from host memory turns out to be impossible, the buffer cache can be modified to have a "reserve" mode where it creates full buffers for entire gpu mappings, instead of creating buffers on demand. Thus BDAs can be created from GPU memory directly. This is more complex than just using host memory, since it also requires splitting the allocation and tracking granularity of buffer and requires more conservative CPU syncing logical to ensure all pages likely to be accessed by shader have been uploaded correctly.

I was just reading about Vulkan memory importing extension. Is there a reason why we cannot use a single DBA to address all the memory? Ot at least bigger chunks than 16K? 500MB page table seems like a lot.

Sorry if the answers are obvious  but I just learned about it.

raphaelthegreat:
For the first question, to use a single BDA for the entire page table you would need to import the entire address space which is impossible as not all GPU memory has been mapped.  If guest maps 2MB of GPU memory you can import that memory and take its BDA to fill the page table entries by offsetting the address for each page.

For the second question, 16K is proposed as its the smallest granularity GPU memory can be mapped. If your chunk size is 32KB for example and the guest does 2 separate 16KB mappings next to each other, the table won't support this properly. This is similar to emulating MMUs and virtual memory so if you are unfamiliar I suggest reading about those topics.

500MB is indeed a lot, but it's a simple way to cover the entire possible address space. After you get it working the size can be optimized by making the page table buffer itself sparse which would reduce its size to less than 1MB. But that can be done later

Ammar-Sadaoui:
> @Xcedf Shader binaries are copyrighted, do not share those.

can you share with me the source of bin shaders copyrighted material

LNDF:
> For the first question, to use a single BDA for the entire page table you would need to import the entire address space which is impossible as not all GPU memory has been mapped. If guest maps 2MB of GPU memory you can import that memory and take its BDA to fill the page table entries by offsetting the address for each page.
> 
> For the second question, 16K is proposed as its the smallest granularity GPU memory can be mapped. If your chunk size is 32KB for example and the guest does 2 separate 16KB mappings next to each other, the table won't support this properly. This is similar to emulating MMUs and virtual memory so if you are unfamiliar I suggest reading about those topics.
> 
> 500MB is indeed a lot, but it's a simple way to cover the entire possible address space. After you get it working the size can be optimized by making the page table buffer itself sparse which would reduce its size to less than 1MB. But that can be done later

I made an early version on a branch on my shadps4 repo. I can get a BDA for a host imported memory and it put into the pagetable.

The only problem I found is that if I try to map a memory region below 0x10000000 I get a VK_ERROR_UNKNOWN error. This is on Linux on AMD with RADV drivers. Internaly it sets errno to EPERM (permission denied). Maybe there is a way arround it, will take a look.

raphaelthegreat:
It is likely the DRM driver is rejecting your host memory import, because the memory is backed by a file
This is discussed here https://gitlab.freedesktop.org/mesa/drm/-/issues/89
The code that checks for this is here, which matches your error code https://github.com/torvalds/linux/blob/5aaaedb0/drivers/gpu/drm/amd/amdgpu/amdgpu_ttm.c#L726
If you have another GPU vendor I recommend testing with that, otherwise you can build the kernel with the check removed

LNDF:
I don't have a GPU of another vendor but the memory import doesn't fail on a kernel that doesn't have that check.

The file that is map to the memory trying to import is the DMem backing file.

LNDF:
I don't have a GPU of another vendor but the memory import doesn't fail on a kernel that doesn't have that check.

The file that is map to the memory trying to import is the DMem backing file.

squidbus:
Isn't the BDA and memory import approach going to run into issues with CPU side command buffer synchronization? Currently such synchronization is done while processing the guest command buffer, and the host Vulkan command buffer is only submitted after everything is processed. If buffer memory changes, it is handled by the staging/temp buffer copies we do in the cache, which makes sure that each version of the buffer used by a draw/dispatch is conserved. However, directly using host memory means that the whole command buffer will execute against the memory as it is at the end of the guest command buffer. And inserting Vulkan command submit-and-waits at each sync point can tank performance.

raphaelthegreat:
Yes that is true, though I suspect it will work well enough in practice, for a few cases, as data read this way is often streamed in. The end goal is indeed to use GPU buffers as the source of the BDAs, but since that requires additional modifications using host memory might be convenient for the prototyping the page table implementation.

Using GPU buffers for this requires the following:

* Making the buffer cache, cache entire mappings instead of on-demand buffers. So on mappings, instead of importing the host memory and filling page table, request a large GPU buffer from the cache, covering the entire mapping, and fill page table with BDA from that. The cache will take care of validating the entire range from the CPU when first created. Join operations performed in JoinOverlap must also update the page table with the new buffer, or have an assert that it's never called (it will only happen if a buffer that crosses mapping boundaries is used) 

* Keep track a list of all dirty pages in the buffer cache. Because reading memory in this fashion, means that it's not easy to know what addresses the shader uses (especially if the shader applies math ops to the address), every time a shader is bound that uses ReadConst outside of sharps, all dirty pages must be flushed to the GPU to ensure correctness. This can be done by having each RegionManager keep track of the number of dirty cpu pages it tracks (number of set bits in the cpu std::vector) and then perform a similar iteration as ForEachUploadRange but over all active managers

LNDF:
If we map every memory address to gpu memory I quickly go to up to like 7 GB of VRAM usage. Maybe an hybrid solution where we try to import host memory and if it fails cache it on GPU local memory is okay? How bad is that for synchronization?

LNDF:
> If we map every memory address to gpu memory I quickly go to up to like 7 GB of VRAM usage. Maybe an hybrid solution where we try to import host memory and if it fails cache it on GPU local memory is okay? How bad is that for synchronization?

Or maybe there is a way to get something like a GPU page fault so we can associate what memory is used by what shader and only cache those pages?

kalaposfos13:
[CUSA08809.log](https://github.com/user-attachments/files/19755719/CUSA08809.log)
Shadow of the Colossus crashes with the following error on this PR:
```
[Debug] <Critical> liverpool.h:1532 unhandled_exception: Unreachable code!
Unhandled exception: bad size of register
```

raphaelthegreat:
> > If we map every memory address to gpu memory I quickly go to up to like 7 GB of VRAM usage. Maybe an hybrid solution where we try to import host memory and if it fails cache it on GPU local memory is okay? How bad is that for synchronization?
> 
> Or maybe there is a way to get something like a GPU page fault so we can associate what memory is used by what shader and only cache those pages?

That would be ideal but I don't think sparse APIs allow you to handle buffer access page faults, let alone if the access is from a BDA (that can cause bsod even on windows). So reserving the entire mapping is the only way to ensure all accessed memory will be in the GPU. Is the guest actually mapping that much GPU memory to make you reach 7GB? Maybe the vulkan memory allocator is reserving more than needed.

The closest solution I can think of that doesn't involve reserving entire mappings, would be to asynchronously keep a list of "faulted" pages and cache them later. So start by filling the page table with host memory. Then increase the buffer allocation granularity to 16 or 64KB (16KB is the minimum but bigger granularities have more chance of a page being cached. 64KB will also allow for an optimization I describe bellow) and fill page table entries on demand. This means that the page table has a mix of GPU local addresses on cached pages but CPU memory elsewhere. In the page table you can take advantage of the BDA address alignment and pack the memory type inside the LSB of the address.

In the shader when accessing the page table you can look at the type by extracting it from the LSB and if the type is CPU memory you consider it "faulted" (technically its still a valid access that is why I put it in quotes) and write the page into a utility storage buffer. Then the CPU can periodically poll this buffer and instruct the buffer cache to cache these pages when they are written. This means the first frame or so will access host memory but will be mapped later.

Another neat trick you can do is take advantage that sparse resident buffers are essentially reserving virtual address space on the GPUs MMU and implement "GPU fastmem". So create a huge 2^40 sparse resident buffer (or closest size to that which is allowed) and bind pages to it on demand either from host memory or from buffer cache. A hybrid scheme can also be used here. This saves a buffer fetch in the shader as the address translation is done in hw by the MMU (the page table buffer fetch is removed), however it has limitations. You can't use the same trick as a software page table to know the page type (you can maintain a separate buffer for that however). Also GPU hw have limits on the granularity of sparse buffer blocks that can be bound. AMD sparse page size is 4KB which is great, NVIDIA is 64KB which is larger than the guest mapping granularity of 16 (it could still work fine tho by increasing caching granularity to 64KB as I mentioned above). Just mentioning this as food for thought.

In the end you could also just play with allocation granularity. Allocating buffers at 4MB chunks for example is gonna waste more VRAM but is far better than reserving mappings probably. You could see which size suits best for your testing for now.

squidbus:
FWIW MoltenVK does not support sparse so any use of sparse needs to be an optional path.

LNDF:
Thsi pull request is dead since it has been replaced by #2819. Closing it. Will reopen it if it must be revived in the future.

Also, i thik threr are useful things in this PR that might be used for other things. I may include them in some future RP.

Regarding the x64 shader backend, it might be used to solve the bindless texture/buffer problem.

PR #2694, opened 2025-03-27
Audio3d

auser1337:
WIP implementation of `libSceAudio3d`.

Audio output is disabled for now, as there are extreme issues in most cases. Don't want to get sued for hearing loss 😅

Allows `CUSA00341` (Uncharted 4: A Thief's End™) to progress slightly further.

PR #2776, opened 2025-04-13
libraries: Initial Audio3d implementation.

squidbus:
Continuing off of https://github.com/shadps4-emu/shadPS4/pull/2694 to get Audio3d to a good-enough state to start getting some games working.

This has a few changes over the PR:
* Fixed queued audio to output to a dedicated Audio3d output opened on initialize, with the same parameters as the real library uses.
* Fixed `sceAudio3dAudioOut*` to use separate audio out handles from queued audio, so that audio properly plays at the same time. This fixes issues in Project DIVA X where the audio would be slow and full of gaps, since the audio was being interleaved instead of overlapped.
* Fixed issues with channel counts and formats. For `sceAudio3dBedWrite2`, both the format and channel count are respected, and for `sceAudio3dObjectSetAttributes`, format is respected and number of channels is always 1. In both cases, the audio is properly converted down to the signed 16-bit stereo that Audio3d uses for its primary audio output.
* Fix audio pointers being passed on directly instead of retaining audio data in the queue.
* Fix up advance vs push to properly advance to the next item in the queue on advance and then send the current item to audio out on push.
* Implement some other functions like `sceAudio3dAudioOutClose` and `sceAudio3dGetDefaultOpenParameters`.
* Misc clean up and formatting.

With this, Project DIVA X has working audio from my testing. Other testing indicates some more complex games at least have fewer audio issues up to the point they can currently be tested, for example loud garbage audio should be gone in many cases.

There are still many missing features (e.g. handling actual positional audio attributes, some port modes) so I don't expect this to be perfect for many games yet.


znssky0107:
CUSA04518 & CUSA02908
Crash when loading opening animation.

Different logs compared to version 0.7.0
[Kernel.Vmm] <Warning> memory.cpp:22 sceKernelGetDirectMemorySize: called
[Kernel.Vmm] <Info> memory.cpp:70 sceKernelAllocateDirectMemory: searchStart = 0x0, searchEnd = 0x135c00000, len = 0x1000000, alignment = 0x4000, memoryType = 0x0, physAddrOut = 0x12f200000
[Kernel.Vmm] <Info> memory.cpp:207 sceKernelMapDirectMemory: called, redirected to sceKernelMapNamedDirectMemory
[Kernel.Vmm] <Info> memory.cpp:175 sceKernelMapNamedDirectMemory: in_addr = 0x0, len = 0x1000000, prot = 0x33, flags = 0x0, directMemoryStart = 0x12f200000, alignment = 0x4000, name = ''
[Kernel.Vmm] <Info> memory.cpp:201 sceKernelMapNamedDirectMemory: out_addr = 0x2b2800000

The problem should be due to other reasons.
I can't test it.

georgemoralis:
Can you test with main if you have the same issue?

StevenMiller123:
Hatsune Miku Project Diva X is extremely prone to race conditions and random crashes on startup. Depending on your setup, it may take multiple attempts to get ingame.
I'll test the US release of the game in a moment.

StevenMiller123:
Looks like Diva X was affected by that ffmpeg issue that just got fixed on main. Latest main works, but all actions from before the submodule update crash while loading the opening.

squidbus:
Yes it was, I had a temporary fix in the test branch before this when I didn't know the cause, and after didn't realize it would affect actions builds too. I've rebased so it should be fixed here now.

StevenMiller123:
image
image
Everything seems to be working as expected on my end now.
[shad_log.txt](https://github.com/user-attachments/files/19726666/shad_log.txt)

Randomuser8219:
With this PR compared to Auser1337's PR, Zaccaria Pinball has less choppy audio, such as UI sounds and flipper sounds.

PR #2778, opened 2025-04-13
multikey for kbm_gui

tlarok:
decided to make code which adds basic multi-key support for kbm_gui.

Changes
EventFilter blocks extra key bindings if isAutoRepeat is true or if 3 keys are already registered.
Replaced all SetMapping() calls in EventFilter() with pressedKeys.insert() to track active keys.
Changed regular checks in CheckMapping() to handle multy key mapping.
Ctrl, Alt, and Shift now detect both left and right key variants.


src/qt_gui/kbm_gui.h
```diff
@@ -33,6 +49,7 @@ private Q_SLOTS:
     void EnableMappingButtons();
     void SetMapping(QString input);
 
+    QSet<QString> pressedNonInt;
```
kalaposfos13: I don't get why the name has to be pressedNonInt here, as far as I can see, this handles numbers too, did its usage change but not its name?

src/qt_gui/kbm_gui.cpp
kalaposfos13, line 1: Here are the things that are not working/inconsistent to me (on Linux):
- left/right/middle click
- the Windows key brings up the system menu, which is fine I guess, but it is also being saved as "1,lwin" for me

The following inputs are missing:
- home/end
- pgup/pgdown
- Caps Lock

src/qt_gui/kbm_gui.cpp
```diff
@@ -696,322 +703,331 @@ bool KBMSettings::eventFilter(QObject* obj, QEvent* event) {
...
+                pressedKeys.insert("backspace");
                 break;
             case Qt::Key_Return:
-                SetMapping("enter");
+                pressedKeys.insert("enter");
                 break;
             case Qt::Key_Enter:
-                SetMapping("kpenter");
+                pressedKeys.insert("kpenter");
                 break;
```
kalaposfos13:
```suggestion
                pressedKeys.insert("kpenter");
                break;
            case Qt::Key_Home:
                pressedKeys.insert("home");
                break;
            case Qt::Key_End:
                pressedKeys.insert("end");
                break;
            case Qt::Key_PageDown:
                pressedKeys.insert("pgdown");
                break;
            case Qt::Key_PageUp:
                pressedKeys.insert("pgup");
                break;
            case Qt::Key_CapsLock:
                pressedKeys.insert("capslock");
                break;
```

kalaposfos13 (changes requested):

kalaposfos13 (approved):

PR #2780, opened 2025-04-13
misc: Few small fixes.

squidbus:
* Use PNG for logo in Qt UI, as it seems SVG is not supported by default. Worked on my local but does not work from CI builds.
* Fix some CMake issues:
  * `CMAKE_OSX_DEPLOYMENT_TARGET` is recommended to be set as a cache variable.
  * Avoid adding to `CMAKE_MODULE_PATH` as it can cause conflicts with submodules; for example, there is an error currently because of a conflict between the shadPS4 and SDL3 scripts for finding libusb.

Tested the PNG change in CI to make sure the logo is working there this time.


abouvier:
What was the error with SDL3? Clearing `CMAKE_MODULE_PATH` is a ugly hack :p
At worst, clear it only in `externals/CMakeLists.txt`.

squidbus:
> What was the error with SDL3? Clearing `CMAKE_MODULE_PATH` is a ugly hack :p

Not sure what you mean by clearing as an ugly hack, I'm not clearing it. I'm just popping off the path we added to it.

The issue is that we have some module find scripts by the same name but different capitalization as SDL3 does, which is causing CMake to pick ours instead of theirs and to throw up errors when the package name casing doesn't match.

abouvier:
> The issue is that we have some module find scripts by the same name but different capitalization as SDL3 does, which is causing CMake to pick ours instead of theirs and to throw up errors when the package name casing doesn't match.

You could have just renamed `cmake/Findlibusb.cmake` then, like e.g. `Findshadps4-libusb.cmake` and then use `find_package(shadps4-libusb 1.0.27 MODULE)`.

squidbus:
> You could have just renamed `cmake/Findlibusb.cmake` then, like e.g. `Findshadps4-libusb.cmake` and then use `find_package(shadps4-libusb 1.0.27 MODULE)`.

I don't see any need when we can just add to the path temporarily instead.

abouvier:
> I don't see any need when we can just add to the path temporarily instead.

Well, you need to hardcode the path of every custom cmake module now.

Anyway, I can't reproduce your SDL3 error. Are you building on a case-insensitive filesystem?

squidbus:
> Well, you need to hardcode the path of every custom cmake module now.

But the alternative was also hardcoding the same paths, being added to the module path.

> Anyway, I can't reproduce your SDL3 error. Are you building on a case-insensitive filesystem?

Yes

abouvier:
> But the alternative was also hardcoding the same paths, being added to the module path.

Only twice, for `cmake/` and `externals/cmake-modules/`.
Every module will be in one of those directories. You're just lucky there is only two custom modules right now :p

On second thought I think that `cmake/Findlibusb.cmake` should match the case of the SDL3 module, because you **want** to mask it, and use the same `libusb` library for both shadPS4 and SDL3. Currently it is possible that the two modules will find two different `libusb` libraries, one for shadPS4 and one for SDL3.

You can also prepend `externals/sdl3/cmake/` to `CMAKE_MODULE_PATH` **before** shadPS4's own `cmake/` and thus use the SDL3 modules for both projects (but don't do it).

PR #2788, opened 2025-04-15
vk_rasterizer: Control mapped_ranges access with shared lock.

squidbus:
The GPU mapped memory ranges set may be modified by one thread at the same time as another thread checks it. Reading the set mid-write like this can produce an incorrect result, making it think that valid memory is not mapped.

To address this, add a shared lock to control access to `mapped_ranges`. Any number of readers can access the map at the same time, but when it is being written by a map/unmap operation, that is given exclusive access.

Also made the intervals and set contains check a bit more correct.

Fixes instances of incorrect `Attempted to track non-GPU memory` assert in games that map and unmap memory a lot beyond boot.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -930,12 +930,17 @@ bool Rasterizer::IsMapped(VAddr addr, u64 size) {
         // There is no memory, so not mapped.
         return false;
     }
-    return mapped_ranges.find(boost::icl::interval<VAddr>::right_open(addr, addr + size)) !=
-           mapped_ranges.end();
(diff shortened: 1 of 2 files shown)


StevenMiller123:
In my tests, this appears to fix the `Attempted to track non-GPU memory` asserts in more recent Unreal Engine titles. I'm no longer hitting this assert in DRAGON BALL Z: KAKAROT, KINGDOM HEARTS Ⅲ, or Persona 3 Reload.

kalaposfos13:
This fixes access violations and significantly cuts down on `Attempted to track non-GPU memory` asserts in CUSA08809.

LNDF:
This also fixes non-GPU memory assert on Gran Turismo Sport

PR #2793, opened 2025-04-16
shader_recompiler: Implement S_FLBIT_I32_B32 and V_MUL_HI_I32.

squidbus:
* Implement `S_FLBIT_I32_B32` same as `V_FFBH_U32`, since they seem to do the same thing but for scalar vs vector.
* Implement `V_MUL_HI_I32`. Added casting to the SPIR-V implementation to satisfy type rules for `OpSMulExtended`, which relies on the fact that our other integer instructions always produce `U32` to be valid. Refactored the IR instructions to just return the `hi` component to make this casting easier.


StevenMiller123:
Monster Hunter World: Iceborne is now bootable with this PR.
image
[MHWI log.zip](https://github.com/user-attachments/files/19778637/MHWI.log.zip)

PR #2796, opened 2025-04-16
shader_recompiler: Implement S_SUBB_U32 instruction

Fire-Cube:
This instruction is needed in CUSA02694


squidbus:
You need to also call the function you added from the opcodes switch case in the same file. Please place the case according to the relative order of instructions in the enum.

Would also appreciate if someone could test although the instruction implementation looks okay to me.

Fire-Cube:
> You need to also call the function you added from the opcodes switch case in the same file. Would also appreciate if someone could test although the instruction implementation looks okay to me.

ah yes
forgot to add it back when switching branch

PR #2797, opened 2025-04-16
Implementing DS_SUB_U32, DS_INC_U32, DS_DEC_U32.

diegolix29:
Based on AMD GCN ISA documentaion this pr add some of the missing opcode cases for read and write memory(DS), this is my first implementation of this cases so feel free to correct me if i missed something. Most of the implementations are based on the already done ones. This will only benefit games that use shared memory but for the long run i guess is better to add them first instead of being needed.

### DIFF
--- a/src/shader_recompiler/frontend/translate/data_share.cpp
+++ b/src/shader_recompiler/frontend/translate/data_share.cpp
@@ -19,8 +19,6 @@ void Translator::EmitDataShare(const GcnInst& inst) {
         return DS_INC_U32(inst, false);
     case Opcode::DS_DEC_U32:
         return DS_DEC_U32(inst, false);
-    case Opcode::DS_SUB_RTN_U32:
-        return DS_SUB_U32(inst, true);
     case Opcode::DS_MIN_I32:
         return DS_MIN_U32(inst, true, false);
     case Opcode::DS_MAX_I32:
@@ -43,6 +41,8 @@ void Translator::EmitDataShare(const GcnInst& inst) {
         return DS_WRITE(32, false, true, true, inst);
     case Opcode::DS_ADD_RTN_U32:
         return DS_ADD_U32(inst, true);
+    case Opcode::DS_SUB_RTN_U32:
+        return DS_SUB_U32(inst, true);
     case Opcode::DS_MIN_RTN_U32:
         return DS_MIN_U32(inst, false, true);
     case Opcode::DS_MAX_RTN_U32:
--- a/src/shader_recompiler/frontend/translate/data_share.cpp
+++ b/src/shader_recompiler/frontend/translate/data_share.cpp
@@ -19,10 +19,6 @@ void Translator::EmitDataShare(const GcnInst& inst) {
         return DS_INC_U32(inst, false);
     case Opcode::DS_DEC_U32:
         return DS_DEC_U32(inst, false);
-    case Opcode::DS_WRITE_SRC2_B32:
-        return DS_WRITE_SRC2_B32(inst, true);
-    case Opcode::DS_WRITE_SRC2_B64:
-        return DS_WRITE_SRC2_B64(inst, true);
     case Opcode::DS_SUB_RTN_U32:
         return DS_SUB_U32(inst, true);
     case Opcode::DS_MIN_I32:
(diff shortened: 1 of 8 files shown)


StevenMiller123:
Are there any games these are known to improve?
I believe our general stance is to implement shader opcodes when there's a game we can test that uses them.

squidbus:
Some of these like INC/DEC/SUB look like trivial enough changes that I’d be fine with it, the WRITE ones I think I’d like to have a test game before we do that.

However for the implementations of INC/DEC/SUB, SPIR-V has equivalent OpAtomicIIncrement/OpAtomicIDecrement/OpAtomicISub, so I would prefer to see you add IR instructions for those instead of using the atomic add op.

diegolix29:
In my testing with this 

> Are there any games these are known to improve? I believe our general stance is to implement shader opcodes when there's a game we can test that uses them.


From local testing only got small improvementes into loading areas thats why i wanted to push a pr to see if people could test it. 

yeah i wanted to add mostly the trivial ones.

diegolix29:
> Some of these like INC/DEC/SUB look like trivial enough changes that I’d be fine with it, the WRITE ones I think I’d like to have a test game before we do that.
> 
> However for the implementations of INC/DEC/SUB, SPIR-V has equivalent OpAtomicIIncrement/OpAtomicIDecrement/OpAtomicISub, so I would prefer to see you add IR instructions for those instead of using the atomic add op.

Ah ok sure np i will then remove the write ones, and add the ir instructions for the other 3 thanks for the feedback.

StevenMiller123:
If the game you're testing already works, then this wouldn't improve anything for that game.
If a game uses a missing shader opcode, it would crash as soon as that shader tries to compile.

diegolix29:
Ok i think the instructions seems ok and working, i had to add a new SharedAtomicU32_IncDec to handle 4 arg instead of 5 for the increment and decrement hope its ok now for u @squidbus

src/shader_recompiler/frontend/translate/scalar_alu.cpp
```diff
@@ -508,6 +508,21 @@ void Translator::S_BFE(const GcnInst& inst, bool is_signed) {
     ir.SetScc(ir.INotEqual(result, ir.Imm32(0)));
 }
 
+void Translator::S_BFE_I32(const GcnInst& inst) {
```
squidbus: This seems to be extra, if you could remove.

src/shader_recompiler/frontend/translate/translate.h
```diff
@@ -274,6 +274,9 @@ class Translator {
     void DS_READ(int bit_size, bool is_signed, bool is_pair, bool stride64, const GcnInst& inst);
     void DS_APPEND(const GcnInst& inst);
     void DS_CONSUME(const GcnInst& inst);
+    void DS_SUB_U32(const GcnInst& inst, bool);
```
squidbus:
```suggestion
    void DS_SUB_U32(const GcnInst& inst, bool rtn);
```

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -13,6 +13,14 @@ void Translator::EmitDataShare(const GcnInst& inst) {
...
     case Opcode::DS_ADD_U32:
         return DS_ADD_U32(inst, false);
+    case Opcode::DS_SUB_U32:
+        return DS_SUB_U32(inst, false);
+    case Opcode::DS_INC_U32:
+        return DS_INC_U32(inst, false);
+    case Opcode::DS_DEC_U32:
+        return DS_DEC_U32(inst, false);
+    case Opcode::DS_SUB_RTN_U32:
+        return DS_SUB_U32(inst, true);
```
squidbus: Can you move this case for `DS_SUB_RTN_U32` to be below the case for `DS_ADD_RTN_U32`? We try to maintain the same order of opcodes as the enum.

squidbus (commented):
Looks mostly good to me, just a few minor comments.

src/shader_recompiler/frontend/translate/scalar_alu.cpp
diegolix29: yes np the whole maping right?

squidbus: Yes this whole function

src/shader_recompiler/frontend/translate/data_share.cpp
diegolix29: Yes just did i will push the new action then. Thanks for the feedback

squidbus (approved):

PR #2803, opened 2025-04-18
libkernel: Fix sceKernelAllocateDirectMemory and sceKernelAvailableDirectMemorySize

StevenMiller123:
Both sceKernelAllocateDirectMemory and sceKernelAvailableDirectMemorySize handled the search area parameters improperly. This PR adjusts these functions to behave closer to observed hardware behavior.

Real hardware:
image

Main:
image

This PR:
image

I've also reduced the log severity of sceKernelGetDirectMemorySize, since it's behavior appears to be hardware accurate.

So far I've tested a few games and it seems like it doesn't break anything. Wouldn't hurt to have more people test though.


georgemoralis:
looking good

PR #2806, opened 2025-04-18
fix translation

bigQY:
The original translated text is not displayed properly.


georgemoralis:
i am not even sure what is changed here , changelog looks like you moved one line to another place

bigQY:
> i am not even sure what is changed here , changelog looks like you moved one line to another place

Yes, the original function may have been called too early, and the corresponding translations were not yet loaded, so you can't get the translated text. 

image

PR #2808, opened 2025-04-18
NpAuth library stub

kalaposfos13:
Adds the NpAuth library with a dummy implementation of sceNpAuthCreateRequest and sceNpAuthDeleteRequest.


kalaposfos13:
Added a dummy implementation of sceNpAuthCreateRequest and sceNpAuthDeleteRequest, which allows NBA 2K17 to boot further, and now crashes on a device lost.

PR #2812, opened 2025-04-19
[QoL] Audio Volume & Mute Audio

Xphalnos:
Will need implementation in Qt GUI because I'm sucks at Qt

### DIFF
--- a/src/core/libraries/audio/sdl_audio.cpp
+++ b/src/core/libraries/audio/sdl_audio.cpp
@@ -5,6 +5,7 @@
 #include <SDL3/SDL_audio.h>
 #include <SDL3/SDL_hints.h>
 
+#include "common/config.h"
 #include "common/logging/log.h"
 #include "core/libraries/audio/audioout.h"
 #include "core/libraries/audio/audioout_backend.h"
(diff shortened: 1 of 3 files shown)


diegolix29:
I could do the qt, will take a look at this after lunch today

src/core/libraries/audio/sdl_audio.cpp
```diff
@@ -76,8 +77,18 @@ class SDLPortBackend : public PortBackend {
             return;
         }
         // SDL does not have per-channel volumes, for now just take the maximum of the channels.
-        const auto vol = *std::ranges::max_element(ch_volumes);
-        if (!SDL_SetAudioStreamGain(stream, static_cast<float>(vol) / SCE_AUDIO_OUT_VOLUME_0DB)) {
+        // const auto vol = *std::ranges::max_element(ch_volumes);
```
squidbus, line 80: You should not be ignoring the input volume, multiply them or something instead.

Also you need to apply your config volume when audio is initialized, it’s not guaranteed a game will actually call this function.

georgemoralis:
partially invalid

PR #2813, opened 2025-04-19
Add missing SurfaceFormat mapping for FormatBc6 Float

StevenMiller123:
Encountered by STAR OCEAN: Integrity and Faithlessness (CUSA03219)


squidbus:
Have you checked the result, it seems that this is also mapped for Unorm so presumably one of these is wrong. Also for whether this should be the signed or unsigned version. Checking Mesa might provide answers from how the Vulkan formats are mapped.

StevenMiller123:
I took a look at AMD's PAL code, they treat data format FormatBc6 + number format Float as invalid and return their undefined format.
I've been trying to understand Mesa's code, but I still haven't found anything relevant there.

I'll close this for now, and let someone more experienced take a proper look at what's going on. I don't have the game myself, so I can't test it locally.

PR #2815, opened 2025-04-20
[Libs] sceNet

georgemoralis:
A restart with RE in mind . Not much yet . Added it because @kalaposfos13 is sharing them secretely


kalaposfos13:
It's not secret if I'm doing it in the public dev channel :)

georgemoralis:
some more RE work :D

georgemoralis:
some net calls added maybe now we can have some initial results :D

georgemoralis:
seems we are back on booting ...

image

georgemoralis:
several implementations , go test ppl

georgemoralis:
The pr will probably close here . There are plenty of stuff to be done to sceNet yet but let's not make it bigger for now. Test for regressions and we gonna to be merged (it gonna follow a new pr with more implementations after)

Missake212:
No regression on my end.

brad0demx:
No regressions on my end

georgemoralis:
i like to live dangerously

PR #2820, opened 2025-04-20
Implement IMAGE_ATOMIC_FMIN and IMAGE_ATOMIC_FMAX for 32bit floats

mikusp:
I've tried to do what other atomic image operations do, let me know if there is anything that needs to be changed

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv.cpp
@@ -270,6 +270,10 @@ void SetupCapabilities(const Info& info, const Profile& profile, EmitContext& ct
     if (info.has_image_query) {
         ctx.AddCapability(spv::Capability::ImageQuery);
     }
+    if (info.uses_atomic_float_min_max) {
+        ctx.AddExtension("SPV_EXT_shader_atomic_float_min_max");
+        ctx.AddCapability(spv::Capability::AtomicFloat32MinMaxEXT);
+    }
     if (info.uses_lane_id) {
         ctx.AddCapability(spv::Capability::GroupNonUniform);
     }
--- a/src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
@@ -196,11 +196,37 @@ Id EmitImageAtomicUMax32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords
 }
 
 Id EmitImageAtomicFMax32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords, Id value) {
-    return ImageAtomicF32(ctx, inst, handle, coords, value, &Sirit::Module::OpAtomicFMax);
+    if (ctx.profile.supports_image_fp32_atomic_min_max) {
+        return ImageAtomicF32(ctx, inst, handle, coords, value, &Sirit::Module::OpAtomicFMax);
+    }
+
+    const auto u32_value = ctx.OpBitcast(ctx.U32[1], value);
+    const auto sign_bit_set =
+        ctx.OpBitFieldUExtract(ctx.U32[1], u32_value, ctx.ConstU32(31u), ctx.ConstU32(1u));
+
+    const auto result = ctx.OpSelect(
+        ctx.F32[1], sign_bit_set,
+        EmitBitCastF32U32(ctx, EmitImageAtomicUMin32(ctx, inst, handle, coords, u32_value)),
+        EmitBitCastF32U32(ctx, EmitImageAtomicSMax32(ctx, inst, handle, coords, u32_value)));
+
+    return result;
 }
 
 Id EmitImageAtomicFMin32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords, Id value) {
-    return ImageAtomicF32(ctx, inst, handle, coords, value, &Sirit::Module::OpAtomicFMin);
+    if (ctx.profile.supports_image_fp32_atomic_min_max) {
+        return ImageAtomicF32(ctx, inst, handle, coords, value, &Sirit::Module::OpAtomicFMin);
+    }
+
+    const auto u32_value = ctx.OpBitcast(ctx.U32[1], value);
+    const auto sign_bit_set =
+        ctx.OpBitFieldUExtract(ctx.U32[1], u32_value, ctx.ConstU32(31u), ctx.ConstU32(1u));
+
+    const auto result = ctx.OpSelect(
+        ctx.F32[1], sign_bit_set,
+        EmitBitCastF32U32(ctx, EmitImageAtomicUMax32(ctx, inst, handle, coords, u32_value)),
+        EmitBitCastF32U32(ctx, EmitImageAtomicSMin32(ctx, inst, handle, coords, u32_value)));
+
+    return result;
 }
 
 Id EmitImageAtomicInc32(EmitContext&, IR::Inst*, u32, Id, Id) {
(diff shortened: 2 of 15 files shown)


squidbus:
At a glance this looks good, I think you need to enable VK_EXT_shader_atomic_float2 though. Probably would be good to add a support flag for it to the shader Profile too, so we can warn if it’s not available when the instruction is used for now and maybe try to find an alternative later.

src/shader_recompiler/frontend/translate/vector_memory.cpp
```diff
@@ -117,6 +117,8 @@ void Translator::EmitVectorMemory(const GcnInst& inst) {
         return IMAGE_ATOMIC(AtomicOp::Umin, inst);
     case Opcode::IMAGE_ATOMIC_SMAX:
         return IMAGE_ATOMIC(AtomicOp::Smax, inst);
+    case Opcode::IMAGE_ATOMIC_FMAX:
+        return IMAGE_ATOMIC(AtomicOp::Fmax, inst);
```
StevenMiller123, line 123: Is there a reason Opcode::IMAGE_ATOMIC_FMIN isn't added to the switch case here?

georgemoralis:
do you have any game that uses these?

mikusp:
Ghost of Tsushima [CUSA13323] uses atomic fmax

mikusp:
Thanks @squidbus, turns out that extensions is not widely implemented (only Radeons) so I've added a check and a fallback implementation using only integer operations, based on https://stackoverflow.com/a/51549250

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
```diff
@@ -186,6 +194,44 @@ Id EmitImageAtomicUMax32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords
...
+    const auto u32_value = ctx.OpBitcast(ctx.U32[1], value);
+    const auto s32_value = ctx.OpBitcast(ctx.S32[1], value);
+    const auto sign_bit_set = ctx.OpUGreaterThan(
+        ctx.U1[1], ctx.OpShiftRightLogical(ctx.U32[1], u32_value, ctx.ConstU32(31u)),
+        ctx.ConstU32(0U));
+
+    const auto result = ctx.OpSelect(
+        ctx.F32[1], sign_bit_set,
+        EmitBitCastF32U32(ctx, EmitImageAtomicUMin32(ctx, inst, handle, coords, u32_value)),
+        EmitBitCastF32U32(ctx, EmitImageAtomicSMax32(ctx, inst, handle, coords, s32_value)));
```
squidbus: `EmitImageAtomicSMax32` actually can still take in a `U32` type and will do the correct operation, so no need for two different casts.

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
```diff
@@ -186,6 +194,44 @@ Id EmitImageAtomicUMax32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords
...
+Id EmitImageAtomicFMax32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords, Id value) {
+    if (ctx.profile.supports_image_fp32_atomic_min_max) {
+        return ImageAtomicF32(ctx, inst, handle, coords, value, &Sirit::Module::OpAtomicFMax);
+    }
+
+    const auto u32_value = ctx.OpBitcast(ctx.U32[1], value);
+    const auto s32_value = ctx.OpBitcast(ctx.S32[1], value);
+    const auto sign_bit_set = ctx.OpUGreaterThan(
+        ctx.U1[1], ctx.OpShiftRightLogical(ctx.U32[1], u32_value, ctx.ConstU32(31u)),
+        ctx.ConstU32(0U));
```
squidbus: You can just use ~~`OpSignBitSet`~~ bitfield extract.

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
```diff
@@ -186,6 +194,44 @@ Id EmitImageAtomicUMax32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords
...
+
+    const auto result = ctx.OpSelect(
+        ctx.F32[1], sign_bit_set,
+        EmitBitCastF32U32(ctx, EmitImageAtomicUMax32(ctx, inst, handle, coords, u32_value)),
+        EmitBitCastF32U32(ctx, EmitImageAtomicSMin32(ctx, inst, handle, coords, s32_value)));
```
squidbus: Same as above.

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
```diff
@@ -186,6 +194,44 @@ Id EmitImageAtomicUMax32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords
...
+Id EmitImageAtomicFMin32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id coords, Id value) {
+    if (ctx.profile.supports_image_fp32_atomic_min_max) {
+        return ImageAtomicF32(ctx, inst, handle, coords, value, &Sirit::Module::OpAtomicFMin);
+    }
+
+    const auto u32_value = ctx.OpBitcast(ctx.U32[1], value);
+    const auto s32_value = ctx.OpBitcast(ctx.S32[1], value);
+    const auto sign_bit_set = ctx.OpUGreaterThan(
+        ctx.U1[1], ctx.OpShiftRightLogical(ctx.U32[1], u32_value, ctx.ConstU32(31u)),
+        ctx.ConstU32(0U));
```
squidbus: Same as above.

squidbus (commented):
Mostly looks good, just a few small suggestions for the fallbacks.

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
mikusp: Fixed

src/shader_recompiler/backend/spirv/emit_spirv.cpp
```diff
@@ -270,6 +270,10 @@ void SetupCapabilities(const Info& info, const Profile& profile, EmitContext& ct
     if (info.has_image_query) {
         ctx.AddCapability(spv::Capability::ImageQuery);
     }
+    if (info.uses_atomic_float_min_max) {
```
squidbus, line 273: There’s actually one more small thing I missed: we should not add this capability here if the extension is not supported (and thus we are using the fallbacks). Otherwise the PR looks good.

mikusp: Won't that be only set if AtomicFM{in,ax} instructions are actually used and these would only be used if it is supported by the profile?

mikusp: Checked here https://github.com/shadps4-emu/shadPS4/pull/2820/files#diff-5cc4a5538fabc4f316107d4f70cd19235a8b32e07d99ac63ebc819a6b9fb2ecdR199

squidbus: The checks there are for the SPIR-V emit but the check here for whether to emit the capability is based on the IR instructions which are there regardless.

georgemoralis:
@squidbus are we ok now with this?

squidbus:
Once last comment is addressed it’ll be good to go

squidbus:
Actually I can probably just quickly patch in the last change.

PR #2823, opened 2025-04-21
Implement SET_PC_B64 instruction

Fire-Cube:
Implement this instruction by resolving the target addresses of the setpcs and replacing them with branches.  
Fixes unreachable in CUSA02694
image

### DIFF
--- a/src/shader_recompiler/frontend/control_flow_graph.cpp
+++ b/src/shader_recompiler/frontend/control_flow_graph.cpp
@@ -70,8 +70,9 @@ static bool IgnoresExecMask(const GcnInst& inst) {
 
 static std::optional<u32> ResolveSetPcTarget(std::span<const GcnInst> list, u32 setpc_index,
                                              std::span<const u32> pc_map) {
-    if (setpc_index < 3)
+    if (setpc_index < 3) {
         return std::nullopt;
+    }
 
     const auto& getpc = list[setpc_index - 3];
     const auto& arith = list[setpc_index - 2];
@@ -96,7 +97,7 @@ static std::optional<u32> ResolveSetPcTarget(std::span<const GcnInst> list, u32 
     const u32 base_pc = pc_map[setpc_index - 3] + getpc.length;
 
     const u32 result_pc = static_cast<u32>(static_cast<s32>(base_pc) + signed_offset);
-    LOG_INFO(Render_Recompiler, "SetPC target: {} + {} = {}", base_pc, signed_offset, result_pc);
+    LOG_DEBUG(Render_Recompiler, "SetPC target: {} + {} = {}", base_pc, signed_offset, result_pc);
     return result_pc & ~0x3u;
 }
 
@@ -124,10 +125,18 @@ void CFG::EmitLabels() {
         if (inst.IsUnconditionalBranch()) {
             u32 target = inst.BranchTarget(pc);
             if (inst.opcode == Opcode::S_SETPC_B64) {
-                if (auto t = ResolveSetPcTarget(inst_list, i, index_to_pc))
+                if (auto t = ResolveSetPcTarget(inst_list, i, index_to_pc)) {
                     target = *t;
+                } else {
+                    ASSERT_MSG(
+                        false,
+                        "S_SETPC_B64 without a resolvable offset at PC {:#x} (Index {}): Involved "
+                        "instructions not recognized or invalid pattern",
+                        pc, i);
+                }
             }
             AddLabel(target);
+            // Emit this label so that the block ends with the branching instruction
             AddLabel(pc + inst.length);
         } else if (inst.IsConditionalBranch()) {
             const u32 true_label = inst.BranchTarget(pc);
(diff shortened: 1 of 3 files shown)


src/shader_recompiler/frontend/control_flow_graph.cpp
```diff
@@ -67,6 +68,38 @@ static bool IgnoresExecMask(const GcnInst& inst) {
     return false;
 }
 
+static std::optional<u32> ResolveSetPcTarget(std::span<const GcnInst> list, u32 setpc_index,
+                                             std::span<const u32> pc_map) {
+    if (setpc_index < 3)
+        return std::nullopt;
```
squidbus, line 74: Braces around the if body are preferred. Same for the other if blocks in this PR.

src/shader_recompiler/frontend/control_flow_graph.cpp
```diff
@@ -67,6 +68,38 @@ static bool IgnoresExecMask(const GcnInst& inst) {
...
+
+    const u32 imm = arith.src[1].code;
+
+    const s32 signed_offset =
+        (arith.opcode == Opcode::S_ADD_U32) ? static_cast<s32>(imm) : -static_cast<s32>(imm);
+
+    const u32 base_pc = pc_map[setpc_index - 3] + getpc.length;
+
+    const u32 result_pc = static_cast<u32>(static_cast<s32>(base_pc) + signed_offset);
+    LOG_INFO(Render_Recompiler, "SetPC target: {} + {} = {}", base_pc, signed_offset, result_pc);
```
squidbus: Should probably remove this log, or at least lower to debug/trace.

src/shader_recompiler/frontend/control_flow_graph.cpp
```diff
@@ -89,19 +122,22 @@ void CFG::EmitLabels() {
         index_to_pc[i] = pc;
         const GcnInst inst = inst_list[i];
         if (inst.IsUnconditionalBranch()) {
-            const u32 target = inst.BranchTarget(pc);
+            u32 target = inst.BranchTarget(pc);
+            if (inst.opcode == Opcode::S_SETPC_B64) {
+                if (auto t = ResolveSetPcTarget(inst_list, i, index_to_pc))
+                    target = *t;
```
squidbus, line 129: We should probably have some assert instead if it fails to resolve, as we can't compile the shader properly.

src/shader_recompiler/frontend/control_flow_graph.cpp
```diff
@@ -89,19 +122,22 @@ void CFG::EmitLabels() {
...
+            }
             AddLabel(target);
-            // Emit this label so that the block ends with s_branch instruction
```
squidbus, line 94: Instead of removing this you should probably update the comment:
```
// Emit this label so that the block ends with the branching instruction
```

src/shader_recompiler/frontend/control_flow_graph.cpp
```diff
@@ -89,19 +122,22 @@ void CFG::EmitLabels() {
...
             AddLabel(pc + inst.length);
         } else if (inst.IsConditionalBranch()) {
             const u32 true_label = inst.BranchTarget(pc);
             const u32 false_label = pc + inst.length;
             AddLabel(true_label);
             AddLabel(false_label);
         } else if (inst.opcode == Opcode::S_ENDPGM) {
-            const u32 next_label = pc + inst.length;
-            AddLabel(next_label);
+            AddLabel(pc + inst.length);
```
squidbus: Unneeded change

src/shader_recompiler/frontend/instruction.cpp
```diff
@@ -18,7 +18,7 @@ bool GcnInst::IsTerminateInstruction() const {
 }
 
 bool GcnInst::IsUnconditionalBranch() const {
-    return opcode == Opcode::S_BRANCH;
+    return opcode == Opcode::S_BRANCH or opcode == Opcode::S_SETPC_B64;
```
squidbus: Prefer using `||` to `or` syntax.

src/shader_recompiler/frontend/translate/scalar_flow.cpp
```diff
@@ -18,6 +18,8 @@ void Translator::EmitFlowControl(u32 pc, const GcnInst& inst) {
         return;
     case Opcode::S_GETPC_B64:
         return S_GETPC_B64(pc, inst);
+    case Opcode::S_SETPC_B64:
+        return;
```
squidbus: Don't need a separate return here, can just let it fallthrough into the existing `return;` case below.

squidbus (commented):
Looks good, just some more minor comments from me.

src/shader_recompiler/frontend/translate/scalar_flow.cpp
Fire-Cube: We process the setpc before here so it will throw an unreachable here

squidbus: I’m not sure what you mean, what I’m saying is there’s already a switch case that just returns that you can add this to, there’s no need to put another return here.

Fire-Cube:
I have now implemented the suggestions.

squidbus (approved):

squidbus:
LGTM, have a few more code style nits but I can probably smooth them out later. I’ll merge in a bit when CI is done.

PR #2825, opened 2025-04-21
making init func work after load modules 

diegolix29:
right now it seems init func is not working properly, this helps init func to be able to pass any custom sycalls or anything similar and ensure any customization gets invoked after loading the module adding a check and simplifying it a bit.

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -333,7 +333,7 @@ void Emulator::LoadSystemModules(const std::string& game_serial) {
             LOG_INFO(Loader, "Loading {} from game serial file {}", entry.path().string(),
                      game_serial);
             if (linker->LoadModule(entry.path()) == -1) {
-                entry.path().string();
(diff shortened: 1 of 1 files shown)


squidbus:
I'm not clear on what this is for, if we load the LLE module we should not call the HLE init.

diegolix29:
yes sorry my idea is to make a check to ensure its loaded properly, and to make it a bit more easy to check. But if its irrelevant its not a problem i have been doing some files checkups on VS so im trying some of its suggestions

squidbus:
Yeah I don’t think this will change much, the continue statement that you replaced would make sure the following code you put in an if block does not execute anyway.

squidbus:
Going to close this as I don't think its a necessary change, the outcome should be the same

diegolix29:
yeah forgot to close it myself, thanks.

PR #2826, opened 2025-04-21
Sysmodules

diegolix29:
Some sysmodules inconsistencies fixed. Based on Visual studio flags if they are irrelevant lmk

### DIFF
--- a/src/core/libraries/system/sysmodule.cpp
+++ b/src/core/libraries/system/sysmodule.cpp
@@ -20,12 +20,40 @@ int PS4_SYSV_ABI sceSysmoduleGetModuleHandleInternal() {
 }
 
 s32 PS4_SYSV_ABI sceSysmoduleGetModuleInfoForUnwind(VAddr addr, s32 flags, void* info) {
-    LOG_ERROR(Lib_SysModule, "(STUBBED) called");
-    Kernel::OrbisModuleInfoForUnwind module_info;
-    module_info.st_size = 0x130;
-    s32 res = Kernel::sceKernelGetModuleInfoForUnwind(addr, flags, &module_info);
-    if (res == 0) {
-        std::memcpy(info, &module_info, sizeof(module_info));
+    LOG_TRACE(Lib_SysModule, "sceSysmoduleGetModuleInfoForUnwind(addr=0x{:X}, flags=0x{:X})", addr,
+              flags);
+
+    auto* module_info = static_cast<Kernel::OrbisModuleInfoForUnwind*>(info);
+    module_info->st_size = sizeof(Kernel::OrbisModuleInfoForUnwind);
+
+    s32 res = Kernel::sceKernelGetModuleInfoForUnwind(addr, flags, module_info);
+    if (res != 0) {
+        return res;
+    }
+
+    static constexpr const char* modules_to_hide[] = {"libc.prx",
+                                                      "libc.sprx",
+                                                      "libSceAudioLatencyEstimation.prx",
+                                                      "libSceFace.prx",
+                                                      "libSceFaceTracker.prx",
+                                                      "libSceFios2.prx",
+                                                      "libSceFios2.sprx",
+                                                      "libSceFontGsm.prx",
+                                                      "libSceHand.prx",
+                                                      "libSceHandTracker.prx",
+                                                      "libSceHeadTracker.prx",
+                                                      "libSceJobManager.prx",
+                                                      "libSceNpCppWebApi.prx",
+                                                      "libSceNpToolkit.prx",
+                                                      "libSceNpToolkit2.prx",
+                                                      "libSceS3DConversion.prx",
+                                                      "libSceSmart.prx"};
+
+    for (const char* hidden_name : modules_to_hide) {
+        if (std::strcmp(module_info->name.data(), hidden_name) == 0) {
+            std::memset(module_info->name.data(), 0, module_info->name.size());
+            break;
+        }
     }
     return res;
 }
--- a/src/core/libraries/system/sysmodule.cpp
+++ b/src/core/libraries/system/sysmodule.cpp
@@ -24,6 +24,9 @@ s32 PS4_SYSV_ABI sceSysmoduleGetModuleInfoForUnwind(VAddr addr, s32 flags, void*
     Kernel::OrbisModuleInfoForUnwind module_info;
     module_info.st_size = 0x130;
     s32 res = Kernel::sceKernelGetModuleInfoForUnwind(addr, flags, &module_info);
+    if (res == 0) {
+        std::memcpy(info, &module_info, sizeof(module_info));
+    }
     return res;
 }
 
@@ -57,7 +60,7 @@ int PS4_SYSV_ABI sceSysmoduleIsLoadedInternal(OrbisSysModuleInternal id) {
 
 int PS4_SYSV_ABI sceSysmoduleLoadModule(OrbisSysModule id) {
     auto color_name = magic_enum::enum_name(id);
-    LOG_ERROR(Lib_SysModule, "(DUMMY) called module = {}", magic_enum::enum_name(id));
+    LOG_ERROR(Lib_SysModule, "(DUMMY) called module = {}", color_name);
     return ORBIS_OK;
 }
 
--- a/src/core/libraries/system/sysmodule.cpp
+++ b/src/core/libraries/system/sysmodule.cpp
@@ -19,41 +19,39 @@ int PS4_SYSV_ABI sceSysmoduleGetModuleHandleInternal() {
     return ORBIS_OK;
 }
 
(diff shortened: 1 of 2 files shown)


src/core/libraries/system/sysmodule.cpp
```diff
@@ -24,6 +24,9 @@ s32 PS4_SYSV_ABI sceSysmoduleGetModuleInfoForUnwind(VAddr addr, s32 flags, void*
     Kernel::OrbisModuleInfoForUnwind module_info;
     module_info.st_size = 0x130;
     s32 res = Kernel::sceKernelGetModuleInfoForUnwind(addr, flags, &module_info);
+    if (res == 0) {
+        std::memcpy(info, &module_info, sizeof(module_info));
+    }
```
StevenMiller123: This may work, but from what little I decompiled, it seemed like libSceSysmodule uses a different info struct from libkernel.
Make sure you've tested that this doesn't cause any weird issues, especially when testing Unity games on release shadPS4 builds for Linux.

diegolix29: yeah i have asked people to test the pr on linux and all modules load ok it seems, just need to make sure a game that gets ingame with modules dependant does not present any weird inconsistencies with this.

diegolix29:
@StevenMiller123 I think its good cause the people that test it didnt have any regressions. I could ask squidbus opinion if u want.

src/core/libraries/system/sysmodule.cpp
squidbus: I would actually suggest you change the type of `void* info` to `Kernel::OrbisModuleInfoForUnwind* info` and pass `info` directly into `Kernel::sceKernelGetModuleInfoForUnwind`. That seems to be what the real implementation is doing, followed by some condition that may clear the module name inside to zero after.

squidbus: More specifically, if after calling `Kernel::sceKernelGetModuleInfoForUnwind`, the `info->name` is equal to any of the following, it should clear the `info->name` to all zero:
```
libc.prx
libc.sprx
libSceAudioLatencyEstimation.prx
libSceFace.prx
libSceFaceTracker.prx
libSceFios2.prx
libSceFios2.sprx
libSceFontGsm.prx
libSceHand.prx
libSceHandTracker.prx
libSceHeadTracker.prx
libSceJobManager.prx
libSceNpCppWebApi.prx
libSceNpToolkit.prx
libSceNpToolkit2.prx
libSceS3DConversion.prx
libSceSmart.prx
```

diegolix29: pushed an actions with a try of what u say tested and no regressions on my end at least on hollow knight

src/core/libraries/system/sysmodule.cpp
```diff
@@ -20,10 +20,41 @@ int PS4_SYSV_ABI sceSysmoduleGetModuleHandleInternal() {
...
 
 s32 PS4_SYSV_ABI sceSysmoduleGetModuleInfoForUnwind(VAddr addr, s32 flags, void* info) {
-    LOG_ERROR(Lib_SysModule, "(STUBBED) called");
-    Kernel::OrbisModuleInfoForUnwind module_info;
-    module_info.st_size = 0x130;
-    s32 res = Kernel::sceKernelGetModuleInfoForUnwind(addr, flags, &module_info);
+    LOG_TRACE(Lib_SysModule, "sceSysmoduleGetModuleInfoForUnwind(addr=0x{:X}, flags=0x{:X})", addr,
+              flags);
+
+    auto* module_info = static_cast<Kernel::OrbisModuleInfoForUnwind*>(info);
```
squidbus: Suggest just changing the function parameter to this type instead of casting, in this file and in the header file if needed.

src/core/libraries/system/sysmodule.cpp
```diff
@@ -20,10 +20,41 @@ int PS4_SYSV_ABI sceSysmoduleGetModuleHandleInternal() {
...
+    module_info->st_size = sizeof(Kernel::OrbisModuleInfoForUnwind);
```
squidbus: This line shouldn't be needed anymore.

src/core/libraries/system/sysmodule.cpp
```diff
@@ -20,10 +20,41 @@ int PS4_SYSV_ABI sceSysmoduleGetModuleHandleInternal() {
...
+                                                      "libSceNpToolkit2.prx",
+                                                      "libSceS3DConversion.prx",
+                                                      "libSceSmart.prx"};
+
+    for (const char* hidden_name : modules_to_hide) {
+        if (std::strcmp(module_info->name.data(), hidden_name) == 0) {
+            std::memset(module_info->name.data(), 0, module_info->name.size());
+            break;
+        }
+    }
```
squidbus, line 55: For the list of modules down to here, suggest something like this:
```c++
    static constexpr std::array<std::string_view, 17> modules_to_hide = {
        "libc.prx",
        "libc.sprx",
        "libSceAudioLatencyEstimation.prx",
        "libSceFace.prx",
        "libSceFaceTracker.prx",
        "libSceFios2.prx",
        "libSceFios2.sprx",
        "libSceFontGsm.prx",
        "libSceHand.prx",
        "libSceHandTracker.prx",
        "libSceHeadTracker.prx",
        "libSceJobManager.prx",
        "libSceNpCppWebApi.prx",
        "libSceNpToolkit.prx",
        "libSceNpToolkit2.prx",
        "libSceS3DConversion.prx",
        "libSceSmart.prx",
    };
    if (const std::string_view module_name = info->name.data();
        std::ranges::find(modules_to_hide, module_name) != modules_to_hide.end()) {
        std::ranges::fill(info->name, '\0');
    }
```

diegolix29: is this what u meant?

image

squidbus: Basically yes, just remove the line setting `info->st_size` and put a comma at the end of `"libSceSmart.prx"` so that formatting puts the closing brace on a new line.

diegolix29: nice sure and also bringed back the magic enum that i replaced for color_name suggested by VS cause it was making the pause menu font on hollow knight to be invisible but then i think the color_name value stored to it never gets read, could be better of to take it out? i tested without it and it bahaves the same as with it. 
image

squidbus: For the code below this you should just remove `auto color_name = magic_enum::enum_name(id);` and keep `LOG_ERROR(Lib_SysModule, "(DUMMY) called module = {}", magic_enum::enum_name(id));`

diegolix29: yeah i tested and yeah it works like that. perfect, then i will push the last suggestion altho i think i have to rebase but will do it force pushed here. Thank you.

src/core/libraries/system/sysmodule.cpp
```diff
@@ -19,11 +19,40 @@ int PS4_SYSV_ABI sceSysmoduleGetModuleHandleInternal() {
     return ORBIS_OK;
 }
 
-s32 PS4_SYSV_ABI sceSysmoduleGetModuleInfoForUnwind(VAddr addr, s32 flags, void* info) {
-    LOG_ERROR(Lib_SysModule, "(STUBBED) called");
-    Kernel::OrbisModuleInfoForUnwind module_info;
-    module_info.st_size = 0x130;
-    s32 res = Kernel::sceKernelGetModuleInfoForUnwind(addr, flags, &module_info);
+static s32 PS4_SYSV_ABI OrbisModuleInfoForUnwind(VAddr addr, s32 flags,
```
squidbus: Why is this renamed and made static, and why is the `LIB_FUNCTION` mapping changed down below?

This should be:
```
s32 PS4_SYSV_ABI sceSysmoduleGetModuleInfoForUnwind(VAddr addr, s32 flags,
                                                    Kernel::OrbisModuleInfoForUnwind* info) {
```

And the below `LIB_FUNCTION` part should be unchanged.

If you're running into a compile error from changing the function signature, you just need to change it in `sysmodule.h` as well.

diegolix29: yeah thats why i send the screenshot if what i sent was what u meant, ok i will change it to this and see if i have any compiling errors, and yeah i changed lib_funtion cause of the compile error.

squidbus: Yeah sorry I missed that part in the screenshot.

diegolix29: np i think it should be ok now

squidbus:
Looks good now, I'll check CI and then if it's all good I'll merge it in.

diegolix29:
Nice thank you

PR #2827, opened 2025-04-22
renderer_vulkan: Restore Vulkan version to 1.3

squidbus:
Restores Vulkan version to 1.3, now that MoltenVK support is finished. Along with this, drops the extension enables and EXT/KHR suffixes for functionality that is part of core Vulkan 1.3.


squidbus:
Upstream merged the 1.3 PR so this is ready now, will merge in after next release is cut.

PR #2829, opened 2025-04-22
Revert "Adding missmatch info to Texture cache assert"

georgemoralis:
Reverts shadps4-emu/shadPS4#2828


squidbus:
I'll close this as I pushed a fixed version of the PR code.

PR #2830, opened 2025-04-22
texture_cache: Relax mismatched image type from assert to cache miss.

squidbus:
Relaxes the image type check to a cache miss. In cases where a dimension is 1 or conditions are relaxed I think it should be a valid miss. Worst case we just have some games that were crashing before that are now just cache missing instead.


diegolix29:
makes DMC pass the  [Debug] <Critical> texture_cache.cpp:330 operator(): Assertion when loading the city 
image

Missake212:
Dead Island - Definitive Edition gets in game with this PR, although with some really broken visuals.

image

[CUSA03291.log](https://github.com/user-attachments/files/19857204/CUSA03291.log)

PR #2838, opened 2025-04-24
sceKernelAllocateDirectMemory hotfixes

StevenMiller123:
This should (hopefully) be the last time I'll need to PR fixes to this thing.

The main issue this addresses is that the end of an allocation doesn't need to be aligned to our alignment value. Fixing this should address the regression seen in https://github.com/shadps4-emu/shadps4-game-compatibility/issues/451

I've also cleaned up the logic slightly, as we should never need to check against search_start in the while loop. That code will never execute on a dmem area containing the search_start address, since the dmem area returned by FindDmemArea will contain search_start, and the while loop only triggers when we need to iterate past that mapping.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -151,9 +151,7 @@ PAddr MemoryManager::Allocate(PAddr search_start, PAddr search_end, size_t size,
         dmem_area++;
 
         // Update local variables based on the new dmem_area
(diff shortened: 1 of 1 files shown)


shinra-electric:
This fixes the regression.

Screenshot 2025-04-24 at 19 58 14

PR #2839, opened 2025-04-24
Implement sceImeDialogGetPanelSize

kalaposfos13:
Implements the above function using sceImeGetPanelSize as a template. Fixes a popup in CUSA05486 - Skyrim.
image

### DIFF
--- a/src/core/libraries/ime/ime_dialog.cpp
+++ b/src/core/libraries/ime/ime_dialog.cpp
@@ -83,9 +83,31 @@ int PS4_SYSV_ABI sceImeDialogGetPanelPositionAndForm() {
     return ORBIS_OK;
 }
 
-int PS4_SYSV_ABI sceImeDialogGetPanelSize() {
-    LOG_ERROR(Lib_ImeDialog, "(STUBBED) called");
-    return ORBIS_OK;
+Error PS4_SYSV_ABI sceImeDialogGetPanelSize(const OrbisImeDialogParam* param, u32* width, u32* height) {
+    LOG_INFO(Lib_ImeDialog, "called");
+
+    if (!width || !height) {
+        return Error::INVALID_ADDRESS;
+    }
+
+    switch (param->type) {
+    case OrbisImeType::Default:
+    case OrbisImeType::BasicLatin:
+    case OrbisImeType::Url:
+    case OrbisImeType::Mail:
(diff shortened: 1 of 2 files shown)


georgemoralis:
noice :)

src/core/libraries/ime/ime_dialog.cpp
```diff
@@ -83,9 +83,32 @@ int PS4_SYSV_ABI sceImeDialogGetPanelPositionAndForm() {
...
+    }
+
+    switch (param->type) {
+    case OrbisImeType::Default:
+    case OrbisImeType::BasicLatin:
+    case OrbisImeType::Url:
+    case OrbisImeType::Mail:
+        // We set our custom sizes, commented sizes are the original ones
+        *width = 500;  // 793
+        *height = 100; // 408
```
squidbus: This needs to be adjusted for IME dialog.

IME UI:
```
    ImVec2 window_size = {500.0f, 100.0f};
```

Versus IME Dialog UI:
```
    if (state->is_multi_line) {
        window_size = {500.0f, 300.0f};
    } else {
        window_size = {500.0f, 150.0f};
    }
```

Also in terms of the commented real values, I'm not sure they apply to IME Dialog.

squidbus: `state->is_multi_line` here being `True(param->option & OrbisImeDialogOption::Multiline)`

squidbus:
One more thing: I got the following real values from the library if you could comment them next to the customized case similar to IME, and update the number case to match:

For non-number dialog: width = 793, height = 576 for multiline or 476 for not
For number dialog: width = 370, height = 470

src/core/libraries/ime/ime_dialog.cpp
```diff
@@ -83,9 +83,35 @@ int PS4_SYSV_ABI sceImeDialogGetPanelPositionAndForm() {
...
+        *width = 500; // original: 793
+        if (True(param->option & OrbisImeDialogOption::Multiline)) {
+            *height = 300; // original: 576
+        } else {
+            *height = 150; // original: 476
+        }
+        break;
+    case OrbisImeType::Number:
+        *width = 370;  // original: 370
+        *height = 402; // original: 470
```
squidbus: You can directly use the original values here without comment

squidbus:
Looks good to me now, thanks

PR #2841, opened 2025-04-24
Unpack SnormNz 2x16

AboMedoz:
My poor attempt to implement this case but i think there is a Problem with it but it should solve the error in InFAMOUS
```
[Debug] <Critical> ir_emitter.cpp:866 Unpack2x16: Unreachable code!
Unsupported 2x16 number format: SnormNz
```
this case is responsible for face textures as i noticed 

### PR :
image

### SnormNz stubbed with Snorm :
image


squidbus:
I would suggest that you handle this with a broader format conversion similar to Uscaled, Sscaled, and Ubnorm is handled. This should cover all cases for render targets, images, buffers, etc where you might encounter this format by translating it into a more supported one.

You can define a number conversion here: https://github.com/shadps4-emu/shadPS4/blob/d370ea32f4d0e7a00e87ef104d2bde22eaa9c343/src/video_core/amdgpu/types.h#L195-L200

Remap the format to your target here:
https://github.com/shadps4-emu/shadPS4/blob/d370ea32f4d0e7a00e87ef104d2bde22eaa9c343/src/video_core/amdgpu/types.h#L285-L304

And then set up the conversion to be returned here:
https://github.com/shadps4-emu/shadPS4/blob/d370ea32f4d0e7a00e87ef104d2bde22eaa9c343/src/video_core/amdgpu/types.h#L339-L350

Then you can implement your conversion in each direction in these two parts:
https://github.com/shadps4-emu/shadPS4/blob/d370ea32f4d0e7a00e87ef104d2bde22eaa9c343/src/shader_recompiler/ir/reinterpret.h#L24-L40
https://github.com/shadps4-emu/shadPS4/blob/d370ea32f4d0e7a00e87ef104d2bde22eaa9c343/src/shader_recompiler/ir/reinterpret.h#L54-L72

AboMedoz:
looks like i made a mistake

src/shader_recompiler/ir/reinterpret.h
```diff
@@ -34,6 +34,12 @@ inline F32 ApplyReadNumberConversion(IREmitter& ir, const F32& value,
     case AmdGpu::NumberConversion::UnormToUbnorm:
         // Convert 0...1 to -1...1
         return ir.FPSub(ir.FPMul(value, ir.Imm32(2.f)), ir.Imm32(1.f));
+    case AmdGpu::NumberConversion::SintToSnormNz: {
+        const IR::U32 additon = ir.IAdd(ir.IMul(ir.BitCast<U32>(value), ir.Imm32(2)), ir.Imm32(1));
+        const IR::F32 left = ir.ConvertSToF(32, 32, additon);
+        const IR::F32 max = ir.Imm32(float(std::numeric_limits<u32>::max()));
```
squidbus: You need to use the appropriate maximum based on the `DataFormat` here, for example:
* `Format8`, `Format8_8`, `Format8_8_8_8` => `std::numeric_limits<u8>::max()`
* `Format16`, `Format16_16`, `Format16_16_16_16` => `std::numeric_limits<u16>::max()`

You can accomplish this by passing `DataFormat` to `MapNumberConversion` wherever it is called, and then based on the data format emit variants of your conversion like `Sint8ToSnormNz` or `Sint16ToSnormNz`. Then here and for write, you can have both go into the same switch case and select the correct max value based on which one it is.

I think for now you can just worry about 8-bit and 16-bit, and leave other formats to the future if they are actually encountered.

AboMedoz: rendering is all broken 

### PR : 
image

### Main :
image

squidbus: Was this part working before adding any of this PR? Seems odd if this broke it

AboMedoz: No from the previous one, SintToSnormNz it was broken from here

squidbus: I mean what does this look like on main

AboMedoz: The pic i sent above titled main if i get your question correctly, tho I'm sure i didn't

squidbus: Ah right okay, I thought it was before and after within this PR for some reason. That seems weird that it would break, if it wasn't crashing before it shouldn't be SnormNz and thus shouldn't be affected

AboMedoz: It was crashing with snormNz but i used to stub it with snorm

squidbus: I see

src/shader_recompiler/ir/reinterpret.h
```diff
@@ -66,6 +78,18 @@ inline F32 ApplyWriteNumberConversion(IREmitter& ir, const F32& value,
     case AmdGpu::NumberConversion::UnormToUbnorm:
         // Convert -1...1 to 0...1
         return ir.FPDiv(ir.FPAdd(value, ir.Imm32(1.f)), ir.Imm32(2.f));
+    case AmdGpu::NumberConversion::Sint8ToSnormNz: {
+        const IR::F32 max = ir.Imm32(float(std::numeric_limits<u8>::max()));
+        const IR::F32 mul = ir.FPMul(ir.FPClamp(value, ir.Imm32(-1.f), ir.Imm32(1.f)), max);
+        const IR::F32 left = ir.FPSub(mul, ir.Imm32(1.f));
+        return ir.BitCast<F32>(U32{ir.FPDiv(left, ir.Imm32(2.f))});
```
squidbus:
```suggestion
        const IR::U32 raw = ir.ConvertFToS(32, ir.FPDiv(left, ir.Imm32(2.f)));
        return ir.BitCast<F32>(raw);
```

Same for below

AboMedoz: done but the render is still broken

GHU7924:
inFAMOUS™ First Light - Main - 81fa9b7

>[Debug] <Critical> ir_emitter.cpp:872 Unpack2x16: Unreachable code!
>Unsupported 2x16 number format: SnormNz

inFAMOUS™ First Light - PR - b9fdf86

>[Debug] <Critical> liverpool_to_vk.h:109 PromoteFormatToDepth: Unreachable code!
>Unexpected depth format R8G8B8A8Unorm

But `The Last Guardian` is also broken, now its colors and background in the menu look like this:

Снимок экрана 2025-04-28 150522
Снимок экрана 2025-04-28 150630
(These may be known issues, but the character can't jump on the monster normally and always falls off of it, and the spears that are stuck in the monster are also not visible)

AboMedoz:
image
thx to @squidbus last main fix SnormNz now works  good 
@GHU7924 can you also test TLG ?

GHU7924:
@AboMedoz TLG - Now everything is fine, it works as it did before the color broke.

InFAMOUS - PromoteFormatToDepth, but I will talk about this in PR 2818.

PR #2845, opened 2025-04-26
Use GetSrc in VALU insts instead of assuming vector reg (was vcc_lo)

baggins183:


baggins183:
fixes worst souls 3 perf :)

baggins183:
I saw some other VALU insts use GetVectorReg (eg V_AND_B32). Im assuming the one i hit was using VOP3. I forget how the emu handles different encodings. It should be fine to change GetVectorReg to GetSrc for most cases right?

squidbus (approved):
Thanks, I remember correcting a bunch of these some months back but I guess there were more hiding around.

hgh32:
Seems to be fixed for sekiro at the same time
Now reaches this assertion
`[Debug] <Critical> ring_access_elimination.cpp:130 operator(): Assertion Failed!`
[shad_log.txt](https://github.com/user-attachments/files/19932255/shad_log.txt)

PR #2847, opened 2025-04-26
ci: Use mold linker for Linux builds

MajorP93:
* The default linker which happens to be BFD in Ubuntu 24.04 does not support Clang's ThinLTO which CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE=ON tries to enable.
* Using mold linker fixes this and reduces build time a bit.
* For consistency reasons we enable mold linker for GCC builds aswell.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -287,7 +287,7 @@ jobs:
         sudo add-apt-repository 'deb http://apt.llvm.org/noble/ llvm-toolchain-noble-19 main'
 
     - name: Install dependencies
-      run: sudo apt-get update && sudo apt install -y libx11-dev libxext-dev libwayland-dev libdecor-0-dev libxkbcommon-dev libglfw3-dev libgles2-mesa-dev libfuse2 clang-19 build-essential libasound2-dev libpulse-dev libopenal-dev libudev-dev
+      run: sudo apt-get update && sudo apt install -y libx11-dev libxext-dev libwayland-dev libdecor-0-dev libxkbcommon-dev libglfw3-dev libgles2-mesa-dev libfuse2 clang-19 mold build-essential libasound2-dev libpulse-dev libopenal-dev libudev-dev
 
     - name: Cache CMake Configuration
       uses: actions/cache@v4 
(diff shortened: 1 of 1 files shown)


Osyotr:
To use `CMAKE_LINKER_TYPE` you also have to bump `cmake_minimum_required` to at least 3.29.
https://github.com/shadps4-emu/shadPS4/blob/e816bc4b99ec1859467f7e69fce68b46009d821b/CMakeLists.txt#L5

https://cmake.org/cmake/help/latest/variable/CMAKE_LINKER_TYPE.html

MajorP93:
> To use `CMAKE_LINKER_TYPE` you also have to bump `cmake_minimum_required` to at least 3.29.
> 
> https://github.com/shadps4-emu/shadPS4/blob/e816bc4b99ec1859467f7e69fce68b46009d821b/CMakeLists.txt#L5
> 
> https://cmake.org/cmake/help/latest/variable/CMAKE_LINKER_TYPE.html

Thanks for the hint. While it is true that we need at least CMake version 3.29 in the Ubuntu runner in order to use CMAKE_LINKER_TYPE I do not think we should bump the required CMake version in CMakeLists.txt.
The project can still be built with older CMake versions and this PR only changes the flag in CI and not globally.

MajorP93:
Maybe somebody can convert this PR to a draft. I’d like to do some more testing and changes.
I can not convert it myself - either I am blind or missing the permission.

tomboylover93:
The button to convert a PR to a draft PR should be here in the Reviewers section:

image

.github/workflows/build.yml
```diff
@@ -347,8 +353,14 @@ jobs:
         wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
         sudo add-apt-repository 'deb http://apt.llvm.org/noble/ llvm-toolchain-noble-19 main'
 
+    - name: Add Kitware repository
+      run: |
+        test -f /usr/share/doc/kitware-archive-keyring/copyright ||
+        wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null
```
Margen67: Why are you suppressing stderr from wget?

MajorP93: I copied those from the Kitware homepage in order to test with a recent version of CMake.
Now I removed those parts though as I used a different method to set the linker to mold.

MajorP93:
This PR now uses `CMAKE_EXE_LINKER_FLAGS="-fuse-ld=mold"` and `CMAKE_SHARED_LINKER_FLAGS="-fuse-ld=mold"` instead of `CMAKE_LINKER_TYPE=MOLD`.
This method also works with older CMake versions so we do not have to bump CMake in CI.

.github/workflows/build.yml
Osyotr: You could have used https://github.com/marketplace/actions/get-cmake

Margen67 (approved):

PR #2848, opened 2025-04-26
usbd cpp suggestions 

diegolix29:
fixes a bunch of errors on usbd.cpp that wasnt fetching the actual libusb files.

### DIFF
--- a/cmake/Findlibusb.cmake
+++ b/cmake/Findlibusb.cmake
@@ -1,15 +1,13 @@
-# SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
+# SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-find_package(PkgConfig QUIET)
-pkg_search_module(LIBUSB QUIET IMPORTED_TARGET libusb-1.0)
+find_path(LIBUSB_INCLUDE_DIR NAMES libusb.h)
+find_library(LIBUSB_LIBRARY NAMES usb-1.0.27)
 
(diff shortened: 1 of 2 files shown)


diegolix29:
With the last commits it boots Soul Calibur 6

diegolix29:
image

cmake/Findlibusb.cmake
```diff
@@ -1,8 +1,7 @@
 # SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-find_package(PkgConfig QUIET)
-pkg_search_module(LIBUSB QUIET IMPORTED_TARGET libusb-1.0)
+find_library(LIBUSB_LIBRARY NAMES libusb-1.0.27)
```
abouvier: What the hell is this? You won't find any library with that name.

diegolix29:
Im learning cmake so wrote a basic cmake for this to test why the old one had so many flags on vs and compared to the old one when i pass it it fetches the files so ig something is doing but at least on my side all VS flags that had went away. Maybe if u know a better the proper way would be great. I just push it as it seems to solve that issue on my end. Or it may not fetch anything but then the old one isnt doing it either cause of the many flags the usbd.cpp has on vs.

cmake/Findlibusb.cmake
thecatontheceiling: AI moment

diegolix29: No it a simple cmake write. But i remove it cause it seems the error i had wasnt there anymore. Dunno why when i did the pr i had most of the usbd file filled with flags

georgemoralis:
no

PR #2854, opened 2025-04-27
Code Review: SuspiciousPriority

ljhcage:
Priority of the '&&' operation is higher than that of the '||' operation.It's possible that parentheses should be used in the expression.


squidbus (approved):

squidbus:
Good to go if you could fix the clang-format error.

PR #2856, opened 2025-04-27
Implement MUBUF instructions for shorts/bytes

baggins183:


Missake212:
With this PR (and a patch made to bypass the hang), Ratchet and Clank now goes a little bit further and crashes on:

``[Debug] <Critical> vector_memory.cpp:250 operator(): Assertion Failed!``

[CUSA01073.log](https://github.com/user-attachments/files/19932029/CUSA01073.log)

baggins183:
Ok i was hoping this wouldnt get hit. I pushed something that will let it past this to see what will happen but need to figure out this case

Randomuser8219:
Now the error is:
[Debug] <Critical> page_manager.cpp:207 operator(): Assertion Failed!
Attempted to track non-GPU memory at address 0x16a1c86000, size 0x30000.

Missake212:
Got a similar error as Randomuser, here's the new log

[CUSA01073.log](https://github.com/user-attachments/files/19932099/CUSA01073.log)

georgemoralis:
this pr needs manual rebase

baggins183:
ready

Randomuser8219:
[shad_log.txt](https://github.com/user-attachments/files/21189543/shad_log.txt)
Ratchet & Clank silently crashes with the rebase.

Randomuser8219:
[shad_log.txt](https://github.com/user-attachments/files/21189600/shad_log.txt)
Log with validation enabled and debugger

Missake212:
Ratchet and Clank asserts here on the latest commit 
image

baggins183:
whats your GPU? That needs a TODO for when you dont have i8/i16 support but im suprised you hit it

Missake212:
4090 🤔

[CUSA01073.log](https://github.com/user-attachments/files/21190043/CUSA01073.log)

baggins183:
try this

Missake212:
Crash in game code :/ (I did try a few times and the address is consistent at least)

image

[CUSA01073.log](https://github.com/user-attachments/files/21190503/CUSA01073.log)

Missake212:
Tried Linux out of curiosity and I get this instead:

``[Debug] <Critical> page_manager.cpp:252 operator(): Assertion Failed!``
``Attempted to track non-GPU memory at address 0x16a1b60000, size 0x156000.``

[CUSA01073.log](https://github.com/user-attachments/files/21190667/CUSA01073.log)

src/shader_recompiler/backend/spirv/emit_spirv_convert.cpp
```diff
@@ -271,4 +271,18 @@ Id EmitConvertU32U8(EmitContext& ctx, Id value) {
     return ctx.OpUConvert(ctx.U32[1], value);
 }
 
+Id EmitConvertS32S8(EmitContext& ctx, Id value) {
+    ASSERT(ctx.profile.support_int8); // TODO
```
squidbus: What are the TODOs here and below for exactly?

FWIW I'm pretty sure we've been just assuming support for these as part of our minimum requirements, hence the profile flags not bothered being set before, maybe we should just unconditionally request it from the driver.

baggins183: I saw some other checks for these in the profile and assumed it needed a similar workaround. Is it pretty safe to assume its available?

Anyways stuff like this:
`
Id ExtractS8(EmitContext& ctx, Id value) {
    if (ctx.profile.support_int8) {
        return ctx.OpSConvert(ctx.S8, value);
    } else {
        return ctx.OpBitFieldSExtract(ctx.U32[1], value, ctx.u32_zero_value, ctx.ConstU32(8u));
    }
}
`

would always take the fallback currently because those never got set in the profile so ill keep those lines in the PipelineCache constructor for now.

baggins183:
> Tried Linux out of curiosity and I get this instead:
> 
> `[Debug] <Critical> page_manager.cpp:252 operator(): Assertion Failed!` `Attempted to track non-GPU memory at address 0x16a1b60000, size 0x156000.`
> 
> [CUSA01073.log](https://github.com/user-attachments/files/21190667/CUSA01073.log)

that might be separate problems
Can you run with vkValidation?
A few commits ago there was
> Id 2934 is defined more than once.

caused by this PR which might not be fixed yet

Randomuser8219:
I'm also getting this error on Windows now:
[Debug] <Critical> page_manager.cpp:252 operator(): Assertion Failed!
Attempted to track non-GPU memory at address 0x16a1b60000, size 0x156000.

src/shader_recompiler/backend/spirv/emit_spirv_convert.cpp
squidbus: I think those checks (and the profile flags) are leftovers, there are definitely other places we just assume we have it. But I'm not strongly in favor of either way here.

squidbus: I think my ideal approach would be just to assume we have it and add a lowering pass if needed, similar to how we have an fp64 to fp32 lowering pass for when that is unsupported. But that might be too much for this PR, at minimum I think you can just assume we have it (and remove the other checks if you want) for now.

Missake212:
Sorry for being late, here you go

[CUSA01073.log](https://github.com/user-attachments/files/21194396/CUSA01073.log)

src/shader_recompiler/backend/spirv/emit_spirv_convert.cpp
baggins183: I deleted the other checks for int8/16.
Also I removed the bitcasts i was doing before/after the SConvert's. I think its legal to sign-extend an unsigned type to a wider unsigned type in spv from the spec. Not completely sure so would be good to see another run with vkValidation.

Randomuser8219:
[shad_log.txt](https://github.com/user-attachments/files/21202574/shad_log.txt)
Ratchet & Clank gets further, then hangs after shader compilation

Missake212:
[CUSA01073.log](https://github.com/user-attachments/files/21204221/CUSA01073.log)

squidbus (approved):

georgemoralis:
https://github.com/georgemoralis/shadPS4/actions/runs/16366689055

PR #2859, opened 2025-04-28
vk_rasterizer: Fix updating wrong color attachment when skipped by mask.

squidbus:
Fix a bug where if a color attachment is skipped and bound to `VK_NULL_HANDLE` due to being masked out, the updates applied later based on `cb_descs` would not also skip those attachments and thus updates after the skipped index would be off.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -696,14 +696,19 @@ void Rasterizer::BindTextures(const Shader::Info& stage, Shader::Backend::Bindin
 
 void Rasterizer::BeginRendering(const GraphicsPipeline& pipeline, RenderState& state) {
     int cb_index = 0;
-    for (auto& [image_id, desc] : cb_descs) {
+    for (auto attach_idx = 0u; attach_idx < state.num_color_attachments; ++attach_idx) {
+        if (state.color_attachments[attach_idx].imageView == VK_NULL_HANDLE) {
+            continue;
+        }
+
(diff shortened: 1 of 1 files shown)


Randomuser8219:
Could this fix this validation error seen in Knack?
```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-vkCmdDraw-dynamicRenderingUnusedAttachments-08910: vkCmdDraw(): VkRenderingInfo::pColorAttachments[0].imageView format (VK_FORMAT_R8G8B8A8_UNORM) must match the corresponding format in VkPipelineRenderingCreateInfo::pColorAttachmentFormats[0] (VK_FORMAT_B8G8R8A8_UNORM).
The Vulkan spec states: If the dynamicRenderingUnusedAttachments feature is not enabled, and the current render pass instance was begun with vkCmdBeginRendering and VkRenderingInfo::colorAttachmentCount greater than 0, then each element of the VkRenderingInfo::pColorAttachments array with an imageView not equal to VK_NULL_HANDLE must have been created with a VkFormat equal to the corresponding element of VkPipelineRenderingCreateInfo::pColorAttachmentFormats used to create the bound graphics pipeline (https://vulkan.lunarg.com/doc/view/1.4.309.0/windows/antora/spec/latest/chapters/drawing.html#VUID-vkCmdDraw-dynamicRenderingUnusedAttachments-08910)
```

squidbus:
I don't think so, it should fix validation errors complaining that an attachment layout is `VK_IMAGE_LAYOUT_UNDEFINED`.

PR #2860, opened 2025-04-28
pp_pass: Use correct surface format.

squidbus:
Use the actual format of the frame instead of fixing to a specific format.


squidbus:
@Randomuser8219 Can you check if this makes that validation error you have go away?

Randomuser8219:
Once I turn on my ROG Ally

Randomuser8219:
[shad_log.txt](https://github.com/user-attachments/files/19933783/shad_log.txt)
Seems like it

PR #2863, opened 2025-04-29
Fix for intel 12 gen up Crashes on Bloodborne

diegolix29:
Seems like this debug trace was the cause of the intel crash on windows for 12 gen and up. 

tested several times and with several sfx, also it was taken from av improvements pr that had this and i backtrace the fix to find it was this. so the fix came from @roamic thanks

### DIFF
--- a/src/common/logging/log.h
+++ b/src/common/logging/log.h
@@ -39,14 +39,10 @@ void FmtLogMessage(Class log_class, Level log_level, const char* filename, unsig
     Common::Log::FmtLogMessage(log_class, log_level, Common::Log::TrimSourcePath(__FILE__),        \
                                __LINE__, __func__, __VA_ARGS__)
 
-#ifdef _DEBUG
 #define LOG_TRACE(log_class, ...)                                                                  \
     Common::Log::FmtLogMessage(Common::Log::Class::log_class, Common::Log::Level::Trace,           \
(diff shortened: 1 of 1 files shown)


StevenMiller123:
No improvements on my end.
[shad_log.txt](https://github.com/user-attachments/files/19965144/shad_log.txt)

diegolix29:
image


working on mine unmodded game lets see if others can test it and give feedback thanks

squidbus:
This doesn’t really make sense, all you are doing is enabling trace logs on release builds when normally the macro does nothing in releases. If anything it suggests that you’re at best seeing a timing related issue and this isn’t an actual fix.

diegolix29:
Yes as i stated just works now but it seems like is another thing that is the source cause of it maybe coming up. With this i can do small plays without any crashes but it will randomly crash. So at least it fixes the constant crash and may help narrow down the actual issue. U can check the development channel in discord and i will later on after work send the stack trace this does when randomly crash now.

TagoSMD:
Still crash but its random sometimes it takes 2-3m.

squidbus:
To be clear, all this PR would do is slow down release builds, which could impact timing related issues like this.

georgemoralis:
no sorry :P

diegolix29:
No problem it was more of an hype but at least it gives some sort of path to trace. No worries here. Thanks 🙏

PR #2866, opened 2025-04-30
Export eboot address

anna12831920:


georgemoralis:
okk but what is the purpose of this change? can you be more detail?

anna12831920:
So tools like cheat engine, x64dbg can have a defined symbol for game module base address. It's only set for eboot in module load currently.
Like `MemoryPatcher::g_eboot_address+0x1234` instead of `0x8ffffc000+0x1234`

Issue #2650, opened 2025-03-15
[GAME BUG]: Color issue with B8G8R8A8Srgb


Xphalnos:
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.
- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [x] I have disabled all patches and cheats and the issue is still present.
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadps4-game-compatibility?tab=readme-ov-file#informations) installed.

### Describe the Bug

On Farming Simulator 17 (and probably other games...), a color bug inverts the colors. After investigation, it's the `B8G8R8A8Srgb` format that's causing the problem. I tried to fix it by inverting the colors of `B8G8R8A8Srgb` in the `8_8_8_8` format, except that it also inverts the other format colors. [**On my fork**](https://github.com/Xphalnos/shadPS4/commit/0896259d10365c51b83a8870a03e98c146e4f6c3) you can see the changes made. Feel free to use them as inspiration for a fix.

Master:
![Image](https://github.com/user-attachments/assets/0c983396-1f70-4c86-aed4-4f8cd1320a92)
![Image](https://github.com/user-attachments/assets/48cea731-b8b7-47a1-a59a-ab3918f6ab45)

With my fix:
![Image](https://github.com/user-attachments/assets/316182fd-af9f-47d4-b99a-acc3582ce9c9)
![Image](https://github.com/user-attachments/assets/6dd9810a-cf11-4512-9d34-ac8cef971b7a)

### Reproduction Steps

 .

### Specify OS Version

Windows 11 23H2

### CPU

Intel Core Pentium Gold 8505

### GPU

Intel UHD Graphics

### Amount of RAM in GB

16GB

### Amount of VRAM in GB

2GB

### Log File

Nothing in the log is related to this color bug.

DanielSvoboda:
Other games have this same 'problem' where the Format8_8_8_8 colors are inverted, such as CUSA08253 and CUSA03455.
Inverting them is not correct because in other games they are correct.
Maybe some shader is causing this

squidbus:
It's most likely related to output swizzle, but there are complications relating to how swizzle interacts with blending that are standing in the way of properly fixing this.

[...]

PR #2871, opened 2025-05-02
vk_presenter: Use correct format and swizzle for output frame image view.

squidbus:
Video out frames can have some versions of formats that are not present for other images, for example BGRA Srgb instead of RGBA Srgb. In cases like this the texture cache will likely contain the RGBA version, especially if exact format was requested by a resolve operation. To handle this, use Vulkan formats for video out frames consistent with other supported GPU formats, and perform any required swizzle in the frame image view. This basically reverses the swizzle that games do when rendering to BGRA for the final frame.

Should fix inverted colors in several games: https://github.com/shadps4-emu/shadPS4/issues/2650

### DIFF
--- a/src/video_core/renderer_vulkan/vk_presenter.cpp
+++ b/src/video_core/renderer_vulkan/vk_presenter.cpp
@@ -270,7 +270,25 @@ Frame* Presenter::PrepareLastFrame() {
     return frame;
 }
 
-Frame* Presenter::PrepareFrameInternal(VideoCore::ImageId image_id, bool is_eop) {
+static vk::Format GetFrameViewFormat(const Libraries::VideoOut::PixelFormat format) {
+    switch (format) {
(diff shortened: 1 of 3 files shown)


squidbus:
To get ahead of the inevitable comment: no this does not seem to fix Knack colors, at least on my end.

StevenMiller123:
This fixes colors in Digimon World: Next Order (CUSA05469)
image

squidbus:
Nice, I also made a slight last minute change to just use a swizzle in presenting instead of an alternate format which should make the formats more consistent and also reduce the number of formats we use.

StevenMiller123:
image
image
With the new push, the colors in Digimon World: Next Order appear to be more washed out than normal.

squidbus:
Interesting, I reverted it back to formatted views instead of swizzle if you could check again. Maybe drivers don't like it being swizzled for some reason.

StevenMiller123:
image
image
Looks good now.

PR #2873, opened 2025-05-02
Bypassing usb count error

diegolix29:
This helps some UE games to bypass the USB device count but also sending a log to the numbers of devices conected when its called by the game. Returning orbis ok to avoid booting errors. 

Helps UE game Souls Calibur 6 to boot. 

Its the only game i have that encounter issue with it as it dosent show any logging that this was the cause i had to go back to 0.7.0 to figure it out.

### DIFF
--- a/src/core/libraries/usbd/usbd.cpp
+++ b/src/core/libraries/usbd/usbd.cpp
@@ -44,7 +44,18 @@ void PS4_SYSV_ABI sceUsbdExit() {
 s64 PS4_SYSV_ABI sceUsbdGetDeviceList(SceUsbdDevice*** list) {
     LOG_DEBUG(Lib_Usbd, "called");
 
-    return libusb_to_orbis_error(libusb_get_device_list(g_libusb_context, list));
+    static ssize_t last_count = -1;
+
+    ssize_t count = libusb_get_device_list(g_libusb_context, list);
+    if (count <= 0)
+        return libusb_to_orbis_error((int)count);
+
+    if (count != last_count) {
+        LOG_INFO(Lib_Usbd, "Found {} USB devices", count);
+        last_count = count;
+    }
+
+    return ORBIS_OK;
 }
 
 void PS4_SYSV_ABI sceUsbdFreeDeviceList(SceUsbdDevice** list, s32 unref_devices) {
--- a/src/core/libraries/usbd/usbd.cpp
+++ b/src/core/libraries/usbd/usbd.cpp
@@ -49,9 +49,9 @@ s64 PS4_SYSV_ABI sceUsbdGetDeviceList(SceUsbdDevice*** list) {
         return libusb_to_orbis_error((int)count);
 
     if (count > 2)
-        LOG_WARNING(Lib_Usbd, "Too Many USB Devices Connected");
+        LOG_WARNING(Lib_Usbd, "Too many USB devices connected to list: {}", count);
     libusb_unref_device((*list)[2]);
-    count = 0;
+    count = -1;
 
     return libusb_to_orbis_error((int)count);
 }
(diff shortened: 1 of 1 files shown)


squidbus:
Probably need confirmation on what this function should return if this is an issue, it seems wrong to me that we never tell the caller how many devices there are with this change.

diegolix29:
The return orbis ok is cause the games that need a usb free for whatever will soft lock if there are more, as the error in the ps4 that says, "Too Many USB Devices Connected" so this is just to bypass any error send by the game, cause it may be interpreted as hardware failure. I cant test with only 3 usbs cause i need at least 4 to use my pc and it soft locks. Shadow told me on development that we could just go back to stubb it if it was a problem. But cant think of another workaround, ps4 can handle 7 usb only if they are self powered.

squidbus:
If games have some kind of limit on how many USB devices for some reason we should just limit the count then, not return 0.

diegolix29:
Ah ok so i can do that, to clmp the count to 1 or 2 and then return     return libusb_to_orbis_error((int)count) right? that may be more sophisticated bypass

squidbus:
Yeah I think if we just clamp (and would be nice to confirm what the expected max is from game) it should be fine.

diegolix29:
thanks while testing i can only get the game to boot when hardcoding the count to 0, 1, 2. but if i go pass that it dosent boot.

diegolix29:
So i think i have to add a fallback and nullprr the list cause im having memleaks with a simple clamp and return count

squidbus:
I think this will probably require some more thought to come up with a working solution because:
* As you say, just clamping the size may result in memory leaks from not enough being freed, we need to actually free off the excess data before giving it to the app.
* How will a game find the USB device it needs if it is past the list size cut-off?

I guess for now we can check if size is greater and if so, free the list and return empty. Please log an error that too many USB devices are connected if this happens so it’s noticeable. But this will need to be revisited later for proper USB listing.

diegolix29:
So after adding the logging is seems the game calls the list all the time thats why the leak, so i pass the count to 0 and its loging the too many usb devices, when there are more than 2 connected, and it seems no leaks happens with this aproach. Will send a push and let me know how u see it.

src/core/libraries/usbd/usbd.cpp
```diff
@@ -44,7 +44,16 @@ void PS4_SYSV_ABI sceUsbdExit() {
...
 
-    return libusb_to_orbis_error(libusb_get_device_list(g_libusb_context, list));
+    ssize_t count = libusb_get_device_list(g_libusb_context, list);
+    if (count < 0)
+        return libusb_to_orbis_error((int)count);
+
+    if (count > 2)
+        LOG_WARNING(Lib_Usbd, "Too Many USB Devices Connected");
+    libusb_unref_device((*list)[2]);
+    count = 0;
```
squidbus:
```suggestion
    if (count > 2) {
        LOG_WARNING(Lib_Usbd, "Too many USB devices connected to list: {}", count);
        for (auto i = 2; i < count; i++) {
            libusb_unref_device((*list)[i]);
        }
        count = 2;
    }
```

Seems to me that if we're just unreferencing devices past the end, we should a) dereference all of them, not just the one at index 2, and b) set the count to 2 to let the game see the remaining devices we did not unreference. If this doesn't end up being stable we can probably just do `libusb_free_device_list` and return an error instead.

diegolix29: So after several testing, i had to return -1 as an error so it will not have any memleaks, i spend last night 40 minutes on main menu no issues, but passing count as 1-2 creates a leak, i tried to do freedevice too and it didnt pass booting into main menu, stucked there. I can do more testing later on this week i need  to dump some games. feel free to propuse any other change i can test it and see how it behaves.

diegolix29: with the last division i can confirm no memory leaks, and also is set to 2 instead of -1

georgemoralis:
checking RE of the lib there isn't a place it assumed it will fail so i mark it as invalid (the case that works on a game is not make it valid)

PR #2876, opened 2025-05-03
PM4CmdEventWrite implementation.

diegolix29:
Implementing PM4CmdEventWrite as they where commented out + some VS suggestions too. 

If they where commented out for some specific reason then ignore this pr.

### DIFF
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -563,7 +563,7 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
                 break;
             }
             case PM4ItOpcode::EventWrite: {
-                // const auto* event = reinterpret_cast<const PM4CmdEventWrite*>(header);
+                const auto* event = reinterpret_cast<const PM4CmdEventWrite*>(header);
                 break;
             }
             case PM4ItOpcode::EventWriteEos: {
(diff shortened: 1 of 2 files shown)


squidbus:
They’re commented because they aren’t used for anything there, just uncommenting the casts does not change anything.

diegolix29:
I  didnt just uncomment them. I added the struct too based on all the calls. Also changed to steady_clock cause Chrono have it like that, instead of high_resolution_clock. But as i say if this arent relevant it can be closed. It was cause i noticed they were commented out and lacked the struct.

squidbus:
> I didnt just uncomment them. I added the struct too based on all the calls.

Yes but there’s still no actual logic happening in the switch cases, just a cast.

> Also changed to strady_clock cause Chrono have it like that, instead of high_resolution_clock.

I’m not clear on why you felt the need to do this, high resolution clock is an alias that points to the clock provided with the highest resolution, which is what we want, and may point to a different clock depending on the standard library implementation and OS.

diegolix29:
Vs studio flag high_resolution_clock and when follow it, just send me to the conversion chrono uses i send it on development channel so with the change vs dosent flag anymore. Probably just a cleaner to the code. Cause then i dunno why we convert to steady_clock chrono
image(124)

diegolix29:
Ok so its there so others OS uses it. Ok np then. Thanks for the answer. 

Qq does then the event write will never be used? Only the oep one?

PR #2877, opened 2025-05-03
Clear stack before executing guest code

viniciuslrangel:
Not really sure if 13Kb is the magical number

Fixes Catherine Full Body. Finally, we don't need patches anymore

### DIFF
--- a/src/core/tls.h
+++ b/src/core/tls.h
@@ -44,9 +44,9 @@ Tcb* GetTcbBase();
 void EnsureThreadInitialized();
 
 template <size_t size>
-void ClearStack() {
-    void* buf = alloca(size);
-    memset(buf, 0, size);
+__attribute__((optnone)) void ClearStack() {
+    volatile void* buf = alloca(size);
+    memset(const_cast<void*>(buf), 0, size);
     buf = nullptr;
 }
 
--- a/src/core/tls.h
+++ b/src/core/tls.h
@@ -5,6 +5,8 @@
 
 #include "common/types.h"
 
+void* memset(void* ptr, int value, size_t num);
+
 namespace Xbyak {
 class CodeGenerator;
 }
@@ -41,9 +43,18 @@ Tcb* GetTcbBase();
 /// Makes sure TLS is initialized for the thread before entering guest.
 void EnsureThreadInitialized();
 
+template <size_t size>
+void ClearStack() {
+    void* buf = alloca(size);
+    memset(buf, 0, size);
+    buf = nullptr;
+}
+
 template <class ReturnType, class... FuncArgs, class... CallArgs>
 ReturnType ExecuteGuest(PS4_SYSV_ABI ReturnType (*func)(FuncArgs...), CallArgs&&... args) {
     EnsureThreadInitialized();
+    // clear stack to avoid trash from EnsureThreadInitialized
+    ClearStack<13_KB>();
     return func(std::forward<CallArgs>(args)...);
 }
 


viniciuslrangel:
Anyone, please, test this in your machine in any game. I'm not 100% sure about the side-effects

ConnorG512:
Hi I have tried this build with Catherine and am getting 

[Debug] <Critical> signals.cpp:85 SignalHandler: Unreachable code!
Unhandled access violation in thread 'FileSeqThread' at code address 0x9c001dc44: Write to address 0x0

[shad_log.txt](https://github.com/user-attachments/files/20024507/shad_log.txt)

squidbus:
It might be more robust to just switch stacks for the initialization process or something

viniciuslrangel:
> It might be more robust to just switch stacks for the initialization process or something

I agree. It's just a workaround to the current implementation (if we dont see regressions)

Missake212:
Catherine Full Body now goes further thanks to this PR, I also saw no regressions when testing my games on it.

src/core/libraries/fiber/fiber_context.s
polybiusproxy, line 1: The original library does not do this. Why clear registers, twice?

viniciuslrangel: Oh, didn't see that it's already clearing it right below :p

polybiusproxy: I could care less about everything else, but I am not letting anyone touch my precious fibers.

squidbus: Just to be clear, was this addressed as I don’t see any commit following this to change anything.

viniciuslrangel: Yes, I "duplicated" an existing logic

squidbus: Okay, so... does that mean the duplicate will be removed? Bit confused on how this is resolved

viniciuslrangel: Yes, I already solved removing the duplicated. This file `fiber_context.s` ended having no changes

StevenMiller123: I don't think you pushed that then?

viniciuslrangel: Wut, did github rollback my changes? LOL
Very unusual

squidbus (approved):
As mentioned would be good to try and optimize this with stack swapping or something, but should be okay to start with.

abouvier:
Why are you forward declaring `memset`? Aren't you supposed to include `<cstring>` and use `std::memset`?

viniciuslrangel:
> Why are you forward declaring `memset`? Aren't you supposed to include `<cstring>` and use `std::memset`?

Just because this is a header and I don't want to include cstring everywhere. (Also, this is temporary code, I hope)

abouvier:
What do you think of that?

```
[ 47%] Building CXX object CMakeFiles/shadps4.dir/src/video_core/texture_cache/tile_manager.cpp.o
[ 47%] Linking CXX executable shadps4
/usr/bin/ld: CMakeFiles/shadps4.dir/src/core/libraries/network/net_ctl_obj.cpp.o: in function `void Core::ClearStack<13312ul>()':
net_ctl_obj.cpp:(.text._ZN4Core10ClearStackILm13312EEEvv[_ZN4Core10ClearStackILm13312EEEvv]+0x5d): undefined reference to `memset(void*, int, unsigned long)'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/shadps4.dir/build.make:5214: shadps4] Error 1
make[1]: *** [CMakeFiles/Makefile2:332: CMakeFiles/shadps4.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

viniciuslrangel:
Is it being defined as a macro? Didn't expect that. Just out of curiosity, what platform/compiler are you building it?

abouvier:
> Is it being defined as a macro? Didn't expect that. Just out of curiosity, what platform/compiler are you building it?

It's not really my issue, but I will say gcc on archlinux: https://aur.archlinux.org/packages/shadps4-git#comment-1023984

viniciuslrangel:
@abouvier Pushed the fix

PR #2879, opened 2025-05-04
Core: Fix module load addresses

StevenMiller123:
On a PS4, all modules are mapped starting at address `0x800000000`, and with no gaps in-between. Since all platforms are able to map to these addresses, we can hard code this address as a constant and use that as the "base" address we map to.

The information this PR is based on comes from observations visible in this klog:
[Memory Map Query klog.txt](https://github.com/user-attachments/files/20025668/Memory.Map.Query.klog.txt)

These changes fix `[Debug] <Critical> memory.cpp:331 operator(): Assertion Failed!` when booting some Call of Duty games.
Specifically from my testing, this fixes updated versions of Call of Duty®: Advanced Warfare (CUSA00803) crashing on startup (note that these titles still need to be loaded from `default.elf` to bypass `sceSystemServiceLoadExec`)
[CoD AW main log.txt](https://github.com/user-attachments/files/20025613/CoD.AW.main.log.txt)
[CoD AW PR log.txt](https://github.com/user-attachments/files/20025617/CoD.AW.PR.log.txt)

Screenshot from 2025-05-03 20-29-05

This doesn't appear to regress any games on my end, wouldn't hurt to have testers though.

### DIFF
--- a/src/core/module.cpp
+++ b/src/core/module.cpp
@@ -19,8 +19,7 @@ namespace Core {
 
 using EntryFunc = PS4_SYSV_ABI int (*)(size_t args, const void* argp, void* param);
 
-static u64 LoadOffset = CODE_BASE_OFFSET;
-static constexpr u64 CODE_BASE_INCR = 0x010000000u;
+static u64 LoadOffset = 0;
 
 static u64 GetAlignedSize(const elf_program_header& phdr) {
     return (phdr.p_align != 0 ? (phdr.p_memsz + (phdr.p_align - 1)) & ~(phdr.p_align - 1)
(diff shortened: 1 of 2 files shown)


Missake212:
No regressions on my end.

src/core/module.cpp
```diff
@@ -84,7 +83,7 @@ static std::string StringToNid(std::string_view symbol) {
 }
 
 Module::Module(Core::MemoryManager* memory_, const std::filesystem::path& file_, u32& max_tls_index)
-    : memory{memory_}, file{file_}, name{file.stem().string()} {
+    : memory{memory_}, file{file_}, name{file.filename().string()} {
```
squidbus, line 86: You need to update the check `if (name == "eboot") {` later down for memory patching.

Xcedf:
Battlefield 4 same error, but unfortunately still not affected
[shad_log.log](https://github.com/user-attachments/files/20026654/shad_log.log)

StevenMiller123:
As a note for future reference, this PR still isn't fully accurate.
Specifically (according to red-prig), modules loaded by games (through sceKernelLoadStartModule) should be loaded at a base address of `0x80000000`.
Additionally, the executable sections of the game's `eboot.bin` should be mapped at `0x400000`.

Due to our current address space implementation, handling these accurately isn't possible on Windows. Both of these addresses can be before the `system_managed_base` set during emulator init.

rafael-57:
> As a note for future reference, this PR still isn't fully accurate. Specifically (according to red-prig), modules loaded by games (through sceKernelLoadStartModule) should be loaded at a base address of `0x80000000`. Additionally, the executable sections of the game's `eboot.bin` should be mapped at `0x400000`.
> 
> Due to our current address space implementation, handling these accurately isn't possible on Windows. Both of these addresses can be before the `system_managed_base` set during emulator init.

image

Wouldn't it be better to add a comment explaining this here or nah?

StevenMiller123:
@rafael-57
I'll probably end up making a feature request issue detailing all these address-space issues I've come across in my recent PRs.
We'll need to rework it on all platforms if we want to accurately handle the PS4's memory map.

PR #2880, opened 2025-05-04
vulkan: load VK_EXT_extended_dynamic_state

oltolm:
fixes Bloodborne crashing on RX 580

Bloodborne crashes in https://github.com/shadps4-emu/shadPS4/blob/17b6343f18a39bbd6436a94956fb6cb90f8f554c/src/video_core/buffer_cache/buffer_cache.cpp#L180


squidbus:
Can you try just changing `bindVertexBuffers2EXT` to `bindVertexBuffers2`? We should be using that version anyway now that we're fully on Vulkan 1.3, I guess I missed that one.

oltolm:
> Can you try just changing `bindVertexBuffers2EXT` to `bindVertexBuffers2`? We should be using that version anyway now that we're fully on Vulkan 1.3, I guess I missed that one.

Yes that works too.

squidbus:
Thanks, if you could update this to do that I can pull it in, or if you'd rather I can just make a commit

oltolm:
done

PR #2885, opened 2025-05-06
add scePthreadSetaffinity and emulate affinity

Fire-Cube:
Fixes spam in UE games and also applies affinity settings to the host cpu if it has at least 8 cores
Someone should test this on Linux because i only have Windows

### DIFF
--- a/src/core/libraries/kernel/threads/pthread.cpp
+++ b/src/core/libraries/kernel/threads/pthread.cpp
@@ -291,7 +291,8 @@ int PS4_SYSV_ABI posix_pthread_create_name_np(PthreadT* thread, const PthreadAtt
     int ret = new_thread->native_thr.Create(RunThread, new_thread, &new_thread->attr);
 
     ASSERT_MSG(ret == 0, "Failed to create thread with error {}", ret);
-    if (ret == 0 && attr != nullptr && *attr != nullptr && (*attr)->cpuset != nullptr) {
+
+    if (attr != nullptr && *attr != nullptr && (*attr)->cpuset != nullptr) {
         new_thread->SetAffinity((*attr)->cpuset);
     }
     if (ret) {
@@ -523,6 +524,7 @@ int PS4_SYSV_ABI posix_pthread_setcancelstate(PthreadCancelState state,
         *oldstate = oldval ? PthreadCancelState::Enable : PthreadCancelState::Disable;
     }
 }
+
 int Pthread::SetAffinity(const Cpuset* cpuset) {
     const auto processor_count = std::thread::hardware_concurrency();
     if (processor_count < 8) {
@@ -532,22 +534,14 @@ int Pthread::SetAffinity(const Cpuset* cpuset) {
         return POSIX_EINVAL;
     }
 
-    std::vector<int> cpus;
     u64 mask = cpuset->bits;
-    for (int cpu = 0; cpu < 64; ++cpu) {
-        if (mask & (1ULL << cpu))
-            cpus.push_back(cpu);
-    }
-
-    if (cpus.empty()) {
-        return POSIX_EINVAL;
-    }
 
     uintptr_t handle = native_thr.GetHandle();
     if (handle == 0) {
         return POSIX_ESRCH;
     }
 
+/* We don't use this currently because some games gets performance problems when applying affinity even on strong hardware
 #ifdef _WIN64
     DWORD_PTR affinity_mask = static_cast<DWORD_PTR>(mask);
     if (!SetThreadAffinityMask(reinterpret_cast<HANDLE>(handle), affinity_mask)) {
@@ -558,8 +552,9 @@ int Pthread::SetAffinity(const Cpuset* cpuset) {
     cpu_set_t cpu_set;
     CPU_ZERO(&cpu_set);
 
-    for (int cpu : cpus) {
-        if (cpu < CPU_SETSIZE) {
+    u64 mask = cpuset->bits;
+    for (int cpu = 0; cpu < std::min(64, CPU_SETSIZE); ++cpu) {
+        if (mask & (1ULL << cpu)) {
             CPU_SET(cpu, &cpu_set);
         }
     }
@@ -570,7 +565,7 @@ int Pthread::SetAffinity(const Cpuset* cpuset) {
         return POSIX_EINVAL;
     }
 #endif
-
+*/
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


StevenMiller123:
Seems like it's not regressing anything, no improvement either (aside from removing the stub spam)
[shad_log.txt](https://github.com/user-attachments/files/20068088/shad_log.txt)
[shad_log.txt](https://github.com/user-attachments/files/20068141/shad_log.txt)

src/core/libraries/kernel/threads/pthread.cpp
```diff
@@ -289,7 +289,11 @@ int PS4_SYSV_ABI posix_pthread_create_name_np(PthreadT* thread, const PthreadAtt
     /* Create thread */
     new_thread->native_thr = Core::NativeThread();
     int ret = new_thread->native_thr.Create(RunThread, new_thread, &new_thread->attr);
+
     ASSERT_MSG(ret == 0, "Failed to create thread with error {}", ret);
+    if (ret == 0 && attr != nullptr && *attr != nullptr && (*attr)->cpuset != nullptr) {
```
squidbus: No need to check ret here as it's already asserted on

src/core/libraries/kernel/threads/pthread.cpp
```diff
@@ -518,9 +522,71 @@ int PS4_SYSV_ABI posix_pthread_setcancelstate(PthreadCancelState state,
     if (oldstate) {
         *oldstate = oldval ? PthreadCancelState::Enable : PthreadCancelState::Disable;
     }
+}
+int Pthread::SetAffinity(const Cpuset* cpuset) {
```
squidbus, line 529: Missing blank line between end of function and start of new one.

Fire-Cube:
Couldn't observe any direct improvements either, but it might help in specific cases.

src/core/libraries/kernel/threads/pthread.cpp
```diff
@@ -518,9 +522,71 @@ int PS4_SYSV_ABI posix_pthread_setcancelstate(PthreadCancelState state,
...
+                                              const Cpuset* cpusetp) {
+    if (thread == nullptr || cpusetp == nullptr) {
+        return POSIX_EINVAL;
+    }
+    thread->attr.cpusetsize = cpusetsize;
+    return thread->SetAffinity(cpusetp);
+}
+
+int PS4_SYSV_ABI scePthreadSetaffinity(PthreadT thread, const Cpuset mask) {
+    return posix_pthread_setaffinity_np(thread, 0x10, &mask);
```
squidbus: I'm guessing you added a separate `scePthreadSetaffinity` function instead of using the `ORBIS(...)` wrapper because of the extra parameter. However:
* This needs the extra translation of error code from POSIX to ORBIS, see: https://github.com/shadps4-emu/shadPS4/blob/fed064931ad599f2de628cd9ad72c640da3f061b/src/core/libraries/kernel/kernel.h#L26-L37
* Just to check, `0x10` has been verified to be correct?

src/core/libraries/kernel/threads/pthread.h
```diff
@@ -332,6 +332,7 @@ struct Pthread {
             return true;
         }
     }
+    int SetAffinity(const Cpuset* cpuset);
```
squidbus, line 336: Ditto here, blank line between functions.

squidbus (commented):
Mostly just minor comments from me

Missake212:
Noticed performance takes a big hit on this PR in Bloodborne, but also Tekken 7 in menus. In Tekken's case the spam in the log is indeed gone.

image

[CUSA03173.log](https://github.com/user-attachments/files/20068153/CUSA03173.log)

[CUSA06014.log](https://github.com/user-attachments/files/20068174/CUSA06014.log)

src/core/libraries/kernel/threads/pthread.cpp
squidbus: I checked myself and `0x10` is right so you can disregard the last point.

src/core/libraries/kernel/threads/pthread.cpp
```diff
@@ -518,9 +522,71 @@ int PS4_SYSV_ABI posix_pthread_setcancelstate(PthreadCancelState state,
     if (oldstate) {
         *oldstate = oldval ? PthreadCancelState::Enable : PthreadCancelState::Disable;
     }
+}
```
squidbus, line 527: Missed on first past but also the return value here got lost.

src/core/libraries/kernel/threads/pthread.cpp
```diff
@@ -518,9 +522,71 @@ int PS4_SYSV_ABI posix_pthread_setcancelstate(PthreadCancelState state,
...
+int Pthread::SetAffinity(const Cpuset* cpuset) {
+    const auto processor_count = std::thread::hardware_concurrency();
+    if (processor_count < 8) {
+        return 0;
+    }
+    if (cpuset == nullptr) {
+        return POSIX_EINVAL;
+    }
+
+    std::vector<int> cpus;
```
squidbus: Vector seems unnecessary as it's unused for Windows path, and for Linux path you can just populate the cpuset directly.

gandalfthewhite19890404:
Metal Gear Solid Collection 2 and 3 does not boot with that PR

gandalfthewhite19890404:
Same - Final Fantasy Crystal Chronicles does not boots

StevenMiller123:
@gandalfthewhite19890404 Could you share logs of these crashes?

gandalfthewhite19890404:
There are none any critical, just silently crashes, and none common is seen
[shad_log-smootssummer.txt](https://github.com/user-attachments/files/20075645/shad_log-smootssummer.txt)
[shad_log-ffcc.txt](https://github.com/user-attachments/files/20075649/shad_log-ffcc.txt)
[shad_log-mgs2.txt](https://github.com/user-attachments/files/20075650/shad_log-mgs2.txt)
[shad_log-mgs3.txt](https://github.com/user-attachments/files/20075651/shad_log-mgs3.txt)

gandalfthewhite19890404:
Diablo 3 has very low framerate with PR
image

main
image

squidbus:
I think to start at least we can keep all the wiring around this (with comments addressed) and leave out the actual affinity setting, then separate that out as it is causing performance issues currently.

Fire-Cube:
I have fixed all things i think and commented the affinity code out with a additional comment

StevenMiller123:
@Fire-Cube It doesn't seem like you fixed the deleted return in `posix_pthread_setcancelstate`.

Fire-Cube:
> @Fire-Cube It doesn't seem like you fixed the deleted return in `posix_pthread_setcancelstate`.

It doesn't have anything to with my code but I can do it

Fire-Cube:
fixed, i deleted it but github diff hided it from me

squidbus:
Code seems okay if testing indicates everything is fine now.

Missake212:
No regressions on my end.

PR #2886, opened 2025-05-06
shader_recompiler: Widen num_conversion bitfield

MajorP93:
We do this in order to be able to actually fit in all possible values from AmdGpu::NumberConversion.

Fixes gcc compiler warnings:
warning: ‘Shader::PsColorBuffer::num_conversion’ is too small to hold all values of ‘enum class AmdGpu::NumberConversion’


MajorP93:
I tested this on Linux, no regressions so far (at least in Bloodborne).

squidbus:
Thanks, missed this when we added more recently. Although this doesn't get used by Bloodborne with regards to your testing, it looks fine to me.

PR #2888, opened 2025-05-08
Optimize games that hit unpatchable EXTRQ/INSERTQ

OFFTKP:
Some extrq/insertq instructions are less than 5 bytes required for a trampoline, so we would handle them in a signal handler. However, this would happen after TryPatchJit returned false, which was after a global mutex lock and a std::ranges::find, making it more expensive than it needs to be

This PR first checks if the instruction is going to be too small for a trampoline and if so handles it immediately instead of trying to patch it first.

---

Should improve performance in games that use extrq/insertq like Infamous First Light
PS for Bloodborne fans: No it doesn't impact Bloodborne performance

Needs testing

---

Slowness was profiled by @AboMedoz

### DIFF
--- a/src/core/cpu_patches.cpp
+++ b/src/core/cpu_patches.cpp
@@ -523,12 +523,12 @@ static bool Is4ByteExtrqOrInsertq(void* code_address) {
 }
 
 static bool TryExecuteIllegalInstruction(void* ctx, void* code_address) {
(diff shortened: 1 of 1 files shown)


AboMedoz:
image
smooth as butter great work pari

Missake212:
No regressions on my end.

georgemoralis:
well infamous is a good reason for merge :D

PR #2889, opened 2025-05-09
gnm: Implement sceGnmDrawIndexIndirectMulti

squidbus:
Implements `sceGnmDrawIndexIndirectMulti`. It seems like it would be implemented using the same packet as `sceGnmDrawIndexIndirectCountMulti` with the count address disabled, but it actually seems to be a different packet ID and structure. I'm guessing this is due to it being an early version of the packet or something in the PS4 GPU.

Also fixed the validation condition for `sceGnmDrawIndexIndirectCountMulti` a bit.

Used by Final Fantasy XV.

### DIFF
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -505,9 +505,10 @@ s32 PS4_SYSV_ABI sceGnmDrawIndexIndirectCountMulti(u32* cmdbuf, u32 size, u32 da
                                                    u32 flags) {
     LOG_TRACE(Lib_GnmDriver, "called");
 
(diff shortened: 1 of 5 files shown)


MajorP93:
@squidbus You renamed draw_initiator to drawInitiator in [src/video_core/amdgpu/pm4_cmds.h](https://github.com/shadps4-emu/shadPS4/pull/2889/files#diff-f7fd11b105cb2f076c7c2a3069da036927f16dd5c02660976e244dfd739b66f6) but draw_initiator seems to be still getting used in other files.
Shouldn't all occurrences of draw_initiator be renamed?

squidbus:
It doesn't really matter since this one is unused but I'll try to make it consistent.

Exhigh:
Seems like PR #2892's commit got added in as well.

PR #2890, opened 2025-05-09
[Libs] Stubs sceSigninDialog

mailwl:
This PR stubs sceSigninDialog.
First step to make SineMora EX (CUSA08076) works


src/core/libraries/signin_dialog/signindialog.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+s32 PS4_SYSV_ABI sceSigninDialogGetStatus() {
+    LOG_ERROR(Lib_SigninDialog, "(STUBBED) called");
+    return ORBIS_OK;
+}
+
+s32 PS4_SYSV_ABI sceSigninDialogUpdateStatus() {
+    LOG_ERROR(Lib_SigninDialog, "(STUBBED) called");
+    return 3;
```
georgemoralis: I guess it is something like dialog finished?

georgemoralis:
does it proccess further?

src/core/libraries/signin_dialog/signindialog.cpp
mailwl: yes, some magic value, after this, game closes library, and stop updating status. i will add enum with one value

georgemoralis: If so i guess you can work around it like common dialog returns. I guess it is the same mechanism

mailwl:
oh, this library doesn't even call, if we return OK in this line:
https://github.com/shadps4-emu/shadPS4/blob/58df609ba00e09435c79d6a6649bce6176f06f78/src/core/libraries/np_manager/np_manager.cpp#L954

now i am not sure, if this signin library needs to be stubbed, because it's possible to avoid it

georgemoralis:
stubbing is not an issue just at least update sceSigninDialogUpdateStatus log message to say that you returned 3 so if it is used somewhere to be known that it is not return 0 (stubbed calls mostly return 0 )

mailwl:
done, now logs looks like this (stucks at sceNpGetAccountIdA, which still return SignedOut):

```
[Lib.SigninDialog] <Error> signindialog.cpp:13 sceSigninDialogInitialize: (STUBBED) called
[Lib.SigninDialog] <Error> signindialog.cpp:18 sceSigninDialogOpen: (STUBBED) called
[Lib.SigninDialog] <Error> signindialog.cpp:28 sceSigninDialogUpdateStatus: (STUBBED) called, return 'finished' status
[Lib.SigninDialog] <Error> signindialog.cpp:43 sceSigninDialogTerminate: (STUBBED) called
[Lib.NpManager] <Debug> np_manager.cpp:949 sceNpGetAccountIdA: user_id 1
[Lib.NpManager] <Error> np_manager.cpp:925 sceNpGetAccountCountryA: (STUBBED) called
[Loader] <Info> emulator.cpp:417 UpdatePlayTime: Playing time for CUSA08076: 0:25:32
```

georgemoralis:
CUSA of the game? A full log here might be good for further reference

mailwl:
added to first post (CUSA08076)

PR #2905, opened 2025-05-11
core: Implement Linux TCB access using guest<->host swapping.

squidbus:
Reviving an old experiment I had to implement TCB access without code patches on Linux. Essentially, whenever a call to guest code happens, we swap GS and FS so the guest TCB is in FS, and when a call to host code happens we do the reverse.

I've done some basic testing with asserts to check that the TCB should be in the correct state when swaps happen during execution of a few games. Needs actual testing from Linux users though to make sure it's working correctly.

### DIFF
--- a/src/core/cpu_patches.cpp
+++ b/src/core/cpu_patches.cpp
@@ -402,10 +402,7 @@ static const std::unordered_map<ZydisMnemonic, PatchInfo> Patches = {
     {ZYDIS_MNEMONIC_INSERTQ, {FilterNoSSE4a, GenerateINSERTQ, true}},
 
 #if defined(_WIN32)
-    // Windows needs a trampoline.
     {ZYDIS_MNEMONIC_MOV, {FilterTcbAccess, GenerateTcbAccess, true}},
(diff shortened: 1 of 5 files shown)


squidbus:
Split this into two commits: one to introduce the host call wrapper, and one to introduce the TCB swap. Would be interested to know if there are any issues with the first commit, or if it's definitely the actual TCB swap part causing issues.

StevenMiller123:
Seems like the tcb swap is the problematic part. Locally reverting that is enough to get games running again.

squidbus:
I've updated it to try explicitly avoiding touching the stack at all for the TCB swap process, maybe try again and if it still doesn't work I guess it'll need a deeper dive into what is going on.

StevenMiller123:
image
image
image
image

The regressions I reported on Discord are still present.

squidbus:
Updating on this: I’ve pulled as much out into main as possible to minimize the change set down to just removing the TLS patches for Linux and implementing the swap. I don’t have a setup for debugging on Linux and fixing this currently; if anyone wants to try picking it up feel free.

PR #2906, opened 2025-05-11
sceKernelAddTimerEvent implementation

Fire-Cube:
sceKernelAddTimerEvent was implemented based on RE and the existing functions.
It’s used in CUSA05786, but I can’t confirm that it changes anything.
However, based on the RE of how sceKernelAddTimerEvent is used in the game and the output in the emulator, it appears to behave correctly.

### DIFF
--- a/src/core/libraries/kernel/equeue.cpp
+++ b/src/core/libraries/kernel/equeue.cpp
@@ -300,24 +300,62 @@ int PS4_SYSV_ABI sceKernelDeleteHRTimerEvent(SceKernelEqueue eq, int id) {
     }
 }
 
+extern boost::asio::io_context io_context;
+
+static bool EventExists(SceKernelEqueue eq, u64 id, s16 filter) {
+    std::scoped_lock lock{eq->m_mutex}; 
+
+    const auto& it = std::ranges::find_if(eq->m_events, [id, filter](auto& ev) {
+        return ev.event.ident == id && ev.event.filter == filter;
+    });
+
+    return it != eq->m_events.cend();
+}
+
+static void TimerCallback(const boost::system::error_code& error, SceKernelEqueue eq,
+                          SceKernelEvent kevent, SceKernelUseconds interval_ms) {
+    if (EventExists(eq, kevent.ident, kevent.filter)) {
+        eq->TriggerEvent(kevent.ident, SceKernelEvent::Filter::Timer, kevent.udata);
+
+        if (!(kevent.flags & SceKernelEvent::Flags::OneShot)) {
+            auto timer = std::make_unique<boost::asio::steady_timer>(
+                io_context, std::chrono::milliseconds(interval_ms));
+
+            timer->async_wait([eq, kevent, interval_ms](const boost::system::error_code& ec) {
+                TimerCallback(ec, eq, kevent, interval_ms);
+            });
+        }
+    }
+}
+
 int PS4_SYSV_ABI sceKernelAddTimerEvent(SceKernelEqueue eq, int id, SceKernelUseconds usec,
                                         void* udata) {
     if (eq == nullptr) {
         return ORBIS_KERNEL_ERROR_EBADF;
     }
 
+    const u64 interval_ms = static_cast<u64>(usec & 0xFFFFFFFF) / 1000;
+
     EqueueEvent event{};
     event.event.ident = static_cast<u64>(id);
     event.event.filter = SceKernelEvent::Filter::Timer;
     event.event.flags = SceKernelEvent::Flags::Add;
     event.event.fflags = 0;
-    event.event.data = static_cast<u64>(usec & 0xFFFFFFFF) / 1000;
+    event.event.data = interval_ms;
     event.event.udata = udata;
     event.time_added = std::chrono::steady_clock::now();
     LOG_INFO(Kernel_Event,
              "Added timing event: queue name={}, queue id={}, ms-intevall={}, func pointer={:x}",
-             eq->GetName(), event.event.ident, event.event.data,
+             eq->GetName(), event.event.ident, interval_ms,
              reinterpret_cast<uintptr_t>(udata));
+
+    event.timer = std::make_unique<boost::asio::steady_timer>(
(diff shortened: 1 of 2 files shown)


StevenMiller123:
With the way you've implemented this, I'm not sure the event will ever trigger?

kalaposfos13:
It does not work for LBP3, inputs still don't get read
```
[Kernel.Event] <Info> equeue.cpp:186 sceKernelCreateEqueue: name = Pad
[Kernel.Event] <Info> equeue.cpp:320 sceKernelAddTimerEvent: Added timing event: queue name=Pad, queue id=0, ms-intevall=12, func pointer=0
```

red-prig:
a backend is needed for the timer system, within the scope of this PR the games will most likely hang in an endless wait for the timer, maybe someone will try to add a backend to this PR?

Fire-Cube:
Thought it was already handled but didn't see HRTimer is oneshot

StevenMiller123:
Our high-res timers rely on a callback function that does the actual triggering, see https://github.com/shadps4-emu/shadPS4/blob/ff1339b0b663fe45c3376083bb3377b3deb99dbb/src/core/libraries/kernel/equeue.cpp#L160

Fire-Cube:
Not sure if this is the correct way to do it, but at least now the game is spamming the log?
No idea if that's expected behavior.

StevenMiller123:
Seems to help slightly with the irregular performance issues in KINGDOM HEARTS HD 2.8 FINAL CHAPTER PROLOGUE (CUSA05795).
I`ll need to spend some time decompiling the game to confirm if this is related or not though. It would explain the spam though, the game would be adding a new timer event for every frame if it's using these for flip timing (though there are VideoOut events designed for this purpose), and it would also explain why the timer is a 16 ms timer (which means it would trigger slightly more often than once per frame).
Main:
image

PR:
image

[shad_log.zip](https://github.com/user-attachments/files/20152312/shad_log.zip)

ghost:
LittleBigPlanet 3 can go further but gets stuck in a loop. (sceNetEpollCreate ---> sceNetEpollDestroy)
(This is already known to occur but I'm going to include it regardless)
LBP3

src/core/libraries/kernel/equeue.h
```diff
@@ -152,10 +152,12 @@ class EqueueInternal {
 
     int WaitForSmallTimer(SceKernelEvent* ev, int num, u32 micros);
 
-private:
-    std::string m_name;
     std::mutex m_mutex;
     std::vector<EqueueEvent> m_events;
```
squidbus: If you need to use these for your implementation, you should make a function inside the class instead of just exposing them.

Fire-Cube:
should be ready now

StevenMiller123:
Fixes audio issues in Kingdom Hearts, Kingdom Hearts Re: Chain of Memories, and Kingdom Hearts 2, and both Kingdom Hearts HD Remix launchers.

Note: These logs all have the filter "Core:Critical Lib.PlayGoDialog:Critical" applied to stop log spam.

Kingdom Hearts HD 1.5 + 2.5 Remix Launcher: [shad_log.txt](https://github.com/user-attachments/files/20175980/shad_log.txt)
Kingdom Hearts 1: [shad_log.txt](https://github.com/user-attachments/files/20175940/shad_log.txt)
Kingdom Hearts Re:Chain of Memories: [shad_log.txt](https://github.com/user-attachments/files/20175936/shad_log.txt)
Kingdom Hearts 2: [shad_log.txt](https://github.com/user-attachments/files/20175817/shad_log.txt)
Kingdom Hearts HD 2.8 Final Chapter Prologue Launcher: [shad_log.txt](https://github.com/user-attachments/files/20175991/shad_log.txt)

src/core/libraries/kernel/equeue.cpp
```diff
@@ -300,6 +310,83 @@ int PS4_SYSV_ABI sceKernelDeleteHRTimerEvent(SceKernelEqueue eq, int id) {
...
+
+    KernelSignalRequest();
+    std::thread([]() { io_context.run(); }).detach();
+
+    timer->async_wait(
+        [eq, event_data = event.event, interval_ms, timer](const boost::system::error_code& ec) {
+            TimerCallback(ec, eq, event_data, interval_ms);
+        });
+
+    if (!eq->AddEvent(event)) {
```
squidbus: Question: Should we add the event before creating the async timer and thread, or is there a reason not to? Thinking about the (probably very unlikely) edge case where the timer callback fires before we add the event.

src/core/libraries/kernel/equeue.cpp
```diff
@@ -300,6 +310,83 @@ int PS4_SYSV_ABI sceKernelDeleteHRTimerEvent(SceKernelEqueue eq, int id) {
...
+
+    LOG_DEBUG(Kernel_Event,
+              "Added timing event: queue name={}, queue id={}, ms-intevall={}, pointer={:x}",
+              eq->GetName(), event.event.ident, interval_ms, reinterpret_cast<uintptr_t>(udata));
+
+    auto timer = std::make_shared<boost::asio::steady_timer>(
+        io_context, std::chrono::milliseconds(interval_ms));
+
+    KernelSignalRequest();
+    std::thread([]() { io_context.run(); }).detach();
```
squidbus: Why do we do this? KernelSignalRequest already triggers an io_context run in the Kernel Service Thread.

src/core/libraries/kernel/equeue.cpp
Fire-Cube: You're right. Additionally, this way the timer only runs if the AddEvent was actually executed correctly.

Fire-Cube:
Pushed a commit
Please retest the games to see if it works
I now don't see timerevent spam

StevenMiller123:
New push breaks audio in KH games. Looks like the threads relying on timer events are stuck waiting on the timer event to trigger (infinitely waiting in sceKernelWaitEqueue).

StevenMiller123:
New push re-fixes the audio in KH games.

PR #2907, opened 2025-05-11
kernel: Clean up and fix some mistakes.

squidbus:
Isolating some misc changes from https://github.com/shadps4-emu/shadPS4/pull/2905 to:
* Make sure they get in regardless of whether that works
* Make sure that these aren't causing some of the issues with that PR.

Summary:
* Remove extra var-args definition from `VA_ARGS` macro to make the functions wrappable. As far as I can tell this wasn't needed.
* Add missing `PS4_SYSV_ABI` to `sceKernelAioInitializeImpl`.
* Clean up some commented debugging stuff from the `ORBIS` wrapper to make backtrace symbols cleaner.
* Fix size of `Cpuset` in pthread affinity functions.
* Simplify `scePthreadSetaffinity` using `ORBIS` similar to how we do `scePthreadAttrGetaffinity`/`scePthreadAttrSetaffinity`: create a separate function that just handles the argument difference, then wrap it to handle the error code translation.
* Make clear that the parameter to `LIB_OBJ` is an object pointer, not a function.
* Fix some missing uses of `ExecuteGuest` in the linker code.

Regression testing is appreciated.

### DIFF
--- a/src/common/va_ctx.h
+++ b/src/common/va_ctx.h
@@ -8,7 +8,7 @@
 #define VA_ARGS                                                                                    \
     uint64_t rdi, uint64_t rsi, uint64_t rdx, uint64_t rcx, uint64_t r8, uint64_t r9,              \
         uint64_t overflow_arg_area, __m128 xmm0, __m128 xmm1, __m128 xmm2, __m128 xmm3,            \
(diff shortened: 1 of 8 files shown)


StevenMiller123:
No regressions on Windows, at least in my tests.
Will check on Linux in a moment.

StevenMiller123:
No regressions for me on Linux either.

PR #2911, opened 2025-05-12
fix rough mouse movement due to incorrect check

ringolol:
Hey! I was having problems playing BB on a mouse, the movement was too rough, especially when i was trying to move mouse slowly.

Found a strange check in the mouse handler. When the mouse is moved perfectly horizontally or vertically, the controller values are not updated. 

Perhaps its better to use OR here? That solved the problems for me, now the mouse movements are pretty smooth

### DIFF
--- a/src/input/input_mouse.cpp
+++ b/src/input/input_mouse.cpp
@@ -60,7 +60,7 @@ Uint32 MousePolling(void* param, Uint32 id, Uint32 interval) {
     float angle = atan2(d_y, d_x);
     float a_x = cos(angle) * output_speed, a_y = sin(angle) * output_speed;
 
-    if (d_x != 0 && d_y != 0) {
+    if (d_x != 0 || d_y != 0) {
         controller->Axis(0, axis_x, GetAxis(-0x80, 0x7f, a_x));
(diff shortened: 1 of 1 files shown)


ringolol:
mb it's better to remove this check completely

kalaposfos13:
You're right, these branches can just be merged into one, do that then
It's a leftover from a version where not moving the mouse had to be handled differently, but now that is not needed anymore, although the incorrect boolean expression was just a skill issue from my part.

ringolol:
@kalaposfos13 merged the branches, could you check pls?

kalaposfos13:
I just remembered why I did it this way, d_x = 0 and d_y = 0 will result in a_x = {deadzone offset} and a_y = 0 due to the trigonometry involved, so without this not moving the mouse will result in the stick drifting to the side constantly, therefore the separate if branches shouldn't be removed here, my bad
Just revert it to how it was in the original commit, and then I think it'll be okay for merging.

ringolol:
got it, i tested with a small deadzone offset, so i didn't notice this

rechecked with default parameters, looks ok, but a bit twitchy for my taste

For me no offsets and a high multiplier feels to be the best:
- deadzone offset = 0
- speed multiplier = 5
- speed offset = 0

PR #2915, opened 2025-05-12
buffer_cache: Split updateBuffer calls into 65536 byte chunks.

squidbus:
`vkCmdUpdateBuffer` can only be used for copies up to 65536 bytes, but games may hit the `InlineData` path with slightly larger amounts of data. To solve this, split into multiple updates.

Fixes constant validation errors in CUSA01047

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -238,7 +238,15 @@ void BufferCache::InlineData(VAddr address, const void* value, u32 num_bytes, bo
         .bufferMemoryBarrierCount = 1,
         .pBufferMemoryBarriers = &pre_barrier,
     });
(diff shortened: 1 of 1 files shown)


raphaelthegreat:
This is likely because InlineData is being used for all cases of DmaData, including Memory -> Memory which should be a copy operation on buffer cache. The proper fix to this would be https://github.com/shadps4-emu/shadPS4/pull/1775

squidbus:
It might be better but also the specific case I was seeing here was only something like 70k bytes instead of 65k so it didn’t seem worth it to drastically change.

raphaelthegreat:
Thought to mention it as it indicates the game might be broken anyway if the source memory area was gpu modified

squidbus:
Makes sense, yeah it will probably ultimately need the other DMA fix as well. At least in the interim would like to make sure anything taking the current path behaves in a valid way though.

PR #2917, opened 2025-05-12
Implement sceKernelIsStack

mikusp:
Called by Resident Evil 2 [CUSA09171]

### DIFF
--- a/src/core/libraries/kernel/memory.cpp
+++ b/src/core/libraries/kernel/memory.cpp
@@ -284,6 +284,13 @@ int PS4_SYSV_ABI sceKernelGetDirectMemoryType(u64 addr, int* directMemoryTypeOut
                                        directMemoryEndOut);
 }
 
+int PS4_SYSV_ABI sceKernelIsStack(void* addr, void** start, void** end) {
+    LOG_DEBUG(Kernel_Vmm, "called, addr = {:#x}, start = {:#x}, end = {:#x}", fmt::ptr(addr),
+              fmt::ptr(start), fmt::ptr(end));
+    auto* memory = Core::Memory::Instance();
+    return memory->IsStack(std::bit_cast<VAddr>(addr), start, end);
+}
+
 s32 PS4_SYSV_ABI sceKernelBatchMap(OrbisKernelBatchMapEntry* entries, int numEntries,
                                    int* numEntriesOut) {
     return sceKernelBatchMap2(entries, numEntries, numEntriesOut,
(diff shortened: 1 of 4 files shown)


aakh1361:
It is also call in the game Resident Evil 4 Remake. [CUSA33388]

aakh1361:
It is also call in the game Resident Evil 4 Remake. [CUSA33388]

[Core] <Error> stubs.cpp:42 CommonStub: Stub: sceKernelIsStack (nid: yDBwVAolDgg) called, returning zero to 0x80652c615

src/core/libraries/kernel/memory.cpp
```diff
@@ -284,6 +284,13 @@ int PS4_SYSV_ABI sceKernelGetDirectMemoryType(u64 addr, int* directMemoryTypeOut
                                        directMemoryEndOut);
 }
 
+int PS4_SYSV_ABI sceKernelIsStack(void* addr, void** start, void** end) {
+    LOG_DEBUG(Kernel_Vmm, "called, addr = {:#x}, start = {:#x}, end = {:#x}",
+        std::bit_cast<VAddr>(addr), std::bit_cast<VAddr>(start), std::bit_cast<VAddr>(end));
```
squidbus: Instead of bit casting here, just use fmt::ptr

src/core/memory.cpp
```diff
@@ -816,4 +816,29 @@ int MemoryManager::GetDirectMemoryType(PAddr addr, int* directMemoryTypeOut,
...
+    auto stack_end = 0ul;
+    if (vma.type == VMAType::Stack) {
+        stack_start = vma.base;
+        stack_end = vma.base + vma.size;
+    }
+
+    if (start != nullptr)
+        *start = std::bit_cast<void*>(stack_start);
+
+    if (end != nullptr)
```
squidbus: Also for here and other instances, generally prefer braces even for single line if body.

src/core/memory.cpp
```diff
@@ -816,4 +816,29 @@ int MemoryManager::GetDirectMemoryType(PAddr addr, int* directMemoryTypeOut,
     return ORBIS_OK;
 }
 
+int MemoryManager::IsStack(VAddr addr, void** start, void** end) {
+    auto vma_handle = FindVMA(addr);
+    if (vma_handle == vma_map.end())
+        return ORBIS_KERNEL_ERROR_EINVAL;
```
StevenMiller123: Due to the way FindVMA is implemented, this code path will not always trigger when it should (we return the previous of the upper bound for a given key, so addresses outside the vma_map usually return the last valid vma instead of the end).
I'd recommend using our VMA Contains method instead, or comparing address to the returned vma's base. 
https://github.com/shadps4-emu/shadPS4/blob/b3abb83fc5f7888fab4191447fcf97f0a9adac3d/src/core/memory.cpp#L582

mikusp:
Fixed all of the issues

PR #2918, opened 2025-05-12
Mprotect only over whole pages

mikusp:
Resident Evil 2 calls mprotect with address not aligned to the page boundary, causing mprotect(2) to fail. On PS4 it seems mprotect silently aligns the parameters to 16kB pages

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -554,12 +554,13 @@ s32 MemoryManager::Protect(VAddr addr, size_t size, MemoryProt prot) {
     s64 protected_bytes = 0;
 
     auto aligned_addr = Common::AlignDown(addr, 16_KB);
-    auto aligned_size = Common::AlignUp(size, 16_KB);
+    auto aligned_size = Common::AlignUp(size + addr - aligned_addr, 16_KB);
     do {
         auto it = FindVMA(aligned_addr + protected_bytes);
         auto& vma_base = it->second;
         auto result = 0;
-        result = ProtectBytes(aligned_addr + protected_bytes, vma_base, aligned_size - protected_bytes, prot);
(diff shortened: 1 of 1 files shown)


squidbus:
Seems fine to me if you pass clang-format. Checking in with @StevenMiller123 as well who’s been more in depth on the memory system.

StevenMiller123:
Seems reasonable considering the FreeBSD source code, though I haven't tested Mprotect on real hardware yet.

src/core/memory.cpp
```diff
@@ -552,17 +552,20 @@ s64 MemoryManager::ProtectBytes(VAddr addr, VirtualMemoryArea vma_base, size_t s
 s32 MemoryManager::Protect(VAddr addr, size_t size, MemoryProt prot) {
     std::scoped_lock lk{mutex};
     s64 protected_bytes = 0;
+
+    auto aligned_addr = Common::AlignDown(addr, 16_KB);
+    auto aligned_size = Common::AlignUp(size, 16_KB);
```
squidbus: Actually I think I do see one problem here. Suppose we have `addr = 0x3FFF, size = 0x3FFF`. Then `aligned_addr = 0x0, aligned_size = 0x4000`, and we protect from `0x0 - 0x3FFF`. But it seems to me that we should do alignment on the end address instead of the size, giving us `end = 0x3FFF + 0x3FFF = 0x7FFE, aligned_end = 0x8000`, and then we protect from `0x0 - 0x7FFF`, which fully overlaps the original requested area.

squidbus: For reference, FreeBSD seems to just do the equivalent of this which would also work:
```
auto aligned_size = Common::AlignUp(size + addr - aligned_addr, 16_KB);
```

PR #2919, opened 2025-05-12
Handle -1 as V_CMP_NE_U64 argument

mikusp:
Used by Resident Evil 2

### DIFF
--- a/src/shader_recompiler/frontend/translate/vector_alu.cpp
+++ b/src/shader_recompiler/frontend/translate/vector_alu.cpp
@@ -989,13 +989,22 @@ void Translator::V_CMP_NE_U64(const GcnInst& inst) {
         }
     };
     const IR::U1 src0{get_src(inst.src[0])};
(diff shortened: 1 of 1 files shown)


georgemoralis:
omg another RE2 fan :)

src/shader_recompiler/frontend/opcodes.h
```diff
@@ -2382,6 +2382,7 @@ enum class OperandField : u32 {
     ExecHi,
     ConstZero,
     SignedConstIntPos = 129,
+    ConstMinusOne = 193,
```
squidbus: This is the same as `SignedConstIntNeg` and doesn't need to be re-defined.

mikusp: Yes, but SignedConstIntNeg is a whole class of operands, ConstMinusOne like ConstZero is clearer when reading

squidbus: I’d prefer to not have the repeated enum value regardless.

PR #2928, opened 2025-05-13
Avoid post-increment of SGPR in S_*_LOAD_DWORD

mikusp:
Hit by Killzone Shadow Fall - when translating an instruction like `s_load_dwordx8  s[96:103], ...`, the post-increment would hit an overflow error on the last iteration of the loop


squidbus:
Can you define overflow error here? I'm not clear what the difference in the outcome is between these versions.

mikusp:
With `dst_reg++` on the last iteration this assert is hit https://github.com/shadps4-emu/shadPS4/blob/073f93172985480961ba09fece0048d94d763d6c/src/shader_recompiler/ir/reg.h#L438

As it uses dst_reg == 103 and then post-increments

squidbus:
Okay, I figured it was an assert on something like that but I was having trouble finding it, thanks.

PR #2930, opened 2025-05-13
liverpool: Lower SetQueueReg to warning log.

squidbus:
I have not been able to find info yet on where the register range for this falls to implement it, but I believe it's mainly stuff related to queue priorities that we can ignore in many cases, so lowering from assert to warning to let games pass this.


StevenMiller123:
Now Final Fantasy XV crashes on `[Debug] <Critical> tile_manager.cpp:273 operator(): Assertion Failed!`
[shad_log.txt](https://github.com/user-attachments/files/20196455/shad_log.txt)

PR #2819, opened 2025-04-20
video_core: Implement DMA.

LNDF:
This implements arbitrary memory access from the GPU to CPU memory.

When mapping memory, tries to import host memory with VK_KHR_external_memory. When a buffer is created in the same address range of an improted region, device local memory is used.

Buffer Device Addresses are used to get pointers to the data in shaders. A BDA pagetable buffer is created on the GPU that holds the BDA for every 16K page.

Additionaly a "fault" buffer is also created which holds a bitfield that tells the CPU if a non GPU local address has been accessed from a shader. This is done by embedding in the LSB of the BDA if a page is cached by the buffer cache. If a page has been accessed by the GPU and is not cached by the buffer cache it will be cached before the next frame.

It may happen that the host memory import fails. (on AMD GPUs on Linux sometimes), in those cases, a fallback value is used and the page bit is set in the "fault" buffer.

Problems:
* I am not able to make the "fault" buffer work. Either the shader is not writing or the CPU side is not reading.
* Seems wrong to do `scheduler.Finish();` every time I download the "fault" buffer. in `BufferCache::CreateFaultBuffers`, If there is a better way to sync I can implement it.
* @raphaelthegreat, you told me in the other PR that I need 64K pages to do the BDA LSB optimization. I can't use 64K pages because we could accidetly try to track non-GPU memory. Hpe 16K is fine.
* Can't be merged until https://github.com/shadps4-emu/sirit/pull/11 is merged. Currently Sirit is pointing to my fork.
* I am not able to do a lot of testing right now, so this needs testing.
* Probably something I forgot to mention.

This is currently used for ReadConst, and work like this:
1. First, we try to access the memory if it exists in device local memory
2. If it fails (will only fail the first frame), we try to access host memory.
3. If it fails, we fallback to the old flatbuf method.
4. If it fails (the offset is synamic), we return zero.

This can also potentialy be used in GetBufferSize.

PR #2934, opened 2025-05-14
build: Target x86-64-v3 CPU architecture

MajorP93:
As discussed in #2819 the usage of AVX2 ops could improve performance quite significantly in a lot of cases which is why this PR sets `-march=x86-64-v3` as the target CPU architecture. This strikes a good balance between enabling performance-critical instructions (e.g., AVX2) and maintaining compatibility with relatively older systems.

The x86-64-v3 target corresponds to Intel's Haswell era architectures. Haswell was introduced in 2014. The vast majority of users today have CPUs that are significantly newer. Systems with pre-Haswell CPUs are unlikely to meet the performance requirements for PS4 emulation, making support for them impractical.

As `-march=x86-64-v3` also implies `-mtune=x86-64-v3`, the mtune part was removed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,9 +54,9 @@ else()
 endif()
 
 if (ARCHITECTURE STREQUAL "x86_64")
-    # Target the same CPU architecture as the PS4, to maintain the same level of compatibility.
-    # Exclude SSE4a as it is only available on AMD CPUs.
-    add_compile_options(-march=btver2 -mtune=generic -mno-sse4a)
+    # Target x86-64-v3 CPU architecture as this is a good balance between supporting performance critical
+    # instructions like AVX2 and maintaining support for older CPUs.
+    add_compile_options(-march=x86-64-v3)
 endif()
 
 if (APPLE AND ARCHITECTURE STREQUAL "x86_64" AND CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "arm64")
--- a/documents/Quickstart/Quickstart.md
+++ b/documents/Quickstart/Quickstart.md
@@ -21,9 +21,9 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 - A processor with at least 4 cores and 6 threads
 - Above 2.5 GHz frequency
-- A CPU supporting the following instruction sets: MMX, SSE, SSE2, SSE3, SSSE3, SSE4.1, SSE4.2, AVX, F16C, CLMUL, AES, BMI1, MOVBE, XSAVE, ABM
+- A CPU supporting the x86-64-v3 baseline.
   - **Intel**: Haswell generation or newer
-  - **AMD**: Jaguar generation or newer
+  - **AMD**: Excavator generation or newer
   - **Apple**: Rosetta 2 on macOS 15.4 or newer
 
 ### GPU


MajorP93:
I went with x86-64-v3 instead of core-avx2 as mtune=core-avx2 would result in Intel specific scheduler optimizations being applied.
x86-64-v3 corresponds to exactly the same feature / instruction set while being a generic CPU target that is not specific to one single manufacturer.
The resulting tuning applied by the compiler makes more sense for this use case.

GHU7924:
This is not necessary, but then you might want to update Wiki (Quick Start), specify the minimum processor families there, and at the same time add minimum video cards (with support for all instructions and sufficient video memory), if needed.

StevenMiller123:
If I recall, our general stance is that we don't want our system requirements to exceed the PS4's hardware capabilities.
Pinging @squidbus for an opinion since they're the one who set this compiler flag.

georgemoralis:
in the other hand if AVX2 can provide worth speedup , it might be a shame not to use it. Advanced games will surely require better cpu than ps4's

GHU7924:
>If I recall, our general stance is that we don't want our system requirements to exceed the PS4's hardware capabilities.
Pinging @squidbus for an opinion since they're the one who set this compiler flag.

>in the other hand if AVX2 can provide worth speedup , it might be a shame not to use it. Advanced games will surely require better cpu than ps4's

Here you need to look at what you as developers are focusing on.

Because for example, the PS4 Pro also came out, I won't go into technical details, but it had an overclocked processor, the frequency on the GPU became a little higher, and in general the video card became 2.5 times more productive, if I'm not mistaken.

You have unchecked the PS4 Pro box for now, but in theory it might benefit from more power.

MajorP93:
> If I recall, our general stance is that we don't want our system requirements to exceed the PS4's hardware capabilities. Pinging @squidbus for an opinion since they're the one who set this compiler flag.

I understand where you are coming from but this might be a bit impractical considering what people out there are actually using (in regards of hardware).

Also: this level of compatibility is currently not given.
PS4's GPU (Liverpool) can only support Vulkan 1.1 while shadPS4 currently requires Vulkan 1.3.

This results in the GPU requirement being at least Maxwell (nvidia) or GCN4 (AMD).

Considering that the bar of hardware support was already raised and that AVX2 is supported by CPUs being older than 10 years I think the proposed change is reasonable.

squidbus:
I'm not opposed to bumping it as long as we have a clear and consistent minimum requirement, which is what putting in an explicit arch was intended to do.

Which leads me to the main comment I have here - you need to update the minimum CPU requirements in the Quickstart.md file:
```
- A CPU supporting the x86-64-v3 baseline.
  - **Intel**: Haswell generation or newer
  - **AMD**: Excavator generation or newer
  - **Apple**: Rosetta 2 on macOS 15.4 or newer
```

MajorP93:
> - A CPU supporting the x86-64-v3 baseline.
>   - **Intel**: Haswell generation or newer
>   - **AMD**: Excavator generation or newer
>   - **Apple**: Rosetta 2 on macOS 15.4 or newer

Thanks for the hint.
I included the documentation change accordingly.

abouvier:
Because this flag is appended, it currently cannot be overridden from the environment with a better arch like `x86-64-v4` or `native` (on a brand new CPU):
```bash
export CXXFLAGS="-march=native"
cmake -B build (...)
cmake --build build
(...)
[57/443] /usr/lib/ccache/bin/c++ (...) -march=native (...) -march=x86-64-v3 (...)
```

This flag, like every non-required optimization flags, should be prepended instead of appended.

squidbus:
I don’t think it’s relevant to bring that up here and now because these flags were already being set for a while now and are only being modified by this PR. If you have a preferred alternative feel free to PR it.

NinjaGaiden529:
onii chan thank you this means a lot

PR #2938, opened 2025-05-15
Proper error handling for MapMemory errors

StevenMiller123:
On real hardware, most functions we've attached to MapMemory will return ENOMEM if they've got flag `MAP_FIXED` and try mapping to already mapped addresses.

Handling this error case properly is necessary for Assassin's Creed® Unity (CUSA00663) and Need for Speed™ Heat (CUSA15081) to progress further.

Once this is merged, I'll probably start an aggregate issue for tracking known memory issues, as this will make it harder to identify them.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -386,7 +386,7 @@ int MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, size_t size, M
         vma = FindVMA(mapped_addr)->second;
         remaining_size = vma.base + vma.size - mapped_addr;
         if (vma.IsMapped() || remaining_size < size) {
-            LOG_ERROR(Kernel_Vmm, "Not enough memory to map {:#x} bytes at address {:#x}", size,
+            LOG_ERROR(Kernel_Vmm, "Unable to map {:#x} bytes at address {:#x}", size,
                       mapped_addr);
             return ORBIS_KERNEL_ERROR_ENOMEM;
         }
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -383,12 +383,13 @@ int MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, size_t size, M
             vma = FindVMA(unmap_addr)->second;
         }
 
-        // This should return SCE_KERNEL_ERROR_ENOMEM but rarely happens.
         vma = FindVMA(mapped_addr)->second;
         remaining_size = vma.base + vma.size - mapped_addr;
-        ASSERT_MSG(!vma.IsMapped() && remaining_size >= size,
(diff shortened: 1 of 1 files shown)


StevenMiller123:
Draft while I research some edge cases.

StevenMiller123:
I've attempted to add support for overwriting VMAs when MemoryMapFlags::NoOverwrite isn't present. This may fix some additional memory issues, though I won't be able to test any of this for a few hours.

src/core/memory.cpp
```diff
@@ -376,19 +376,25 @@ int MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, size_t size, M
...
+        // When it isn't provided, VMAs can be overwritten regardless of if they're mapped.
+        auto should_overwrite = False(flags & MemoryMapFlags::NoOverwrite) || !vma.IsMapped();
+        while (should_overwrite && unmap_addr < mapped_addr + size && remaining_size < size) {
             auto unmapped = UnmapBytesFromEntry(unmap_addr, vma, unmap_size);
             unmap_addr += unmapped;
             unmap_size -= unmapped;
+            // Unmap addr should be the address of the next VMA
+            // if the previous VMA is fully unmapped.
             vma = FindVMA(unmap_addr)->second;
+            should_overwrite = False(flags & MemoryMapFlags::NoOverwrite) || !vma.IsMapped();
```
squidbus: Do we need to duplicate this condition? It may be a bit lengthy but I think having it in the while condition instead of a variable set to the same condition in two places would be preferable.

StevenMiller123: I just did that to avoid dealing with Clang, I can move the condition properly now that I have some time.

mailwl:
doesn't help with CUSA00625 Title: Battlefield™ Hardline:
```
[Kernel.Vmm] <Info> memory.cpp:70 sceKernelAllocateDirectMemory: searchStart = 0x0, searchEnd = 0x120000000, len = 0x1010000, alignment = 0x4000, memoryType = 0x0, physAddrOut = 0x52c00000
[Kernel.Vmm] <Info> memory.cpp:168 sceKernelMapNamedDirectMemory: in_addr = 0x106c700000, len = 0x1010000, prot = 0x3, flags = 0x90, directMemoryStart = 0x52c00000, alignment = 0x0, name = 'Client'
[Kernel.Vmm] <Error> memory.cpp:397 MapMemory: Unable to map 0x1010000 bytes at address 0x106c700000
```
why emu can't map on already allocated memory?

StevenMiller123:
Doesn't seem like my changes regressed anything I've got.
Swapping back to ready for review.

georgemoralis (approved):

PR #2941, opened 2025-05-16
Fix reading not existing savedata

mailwl:
This PR fix crashing CUSA02472 - Penarium

### DIFF
--- a/src/core/libraries/save_data/save_memory.cpp
+++ b/src/core/libraries/save_data/save_memory.cpp
@@ -10,6 +10,7 @@
 #include <utility>
 #include <fmt/format.h>
 
+#include "boost/icl/concept/interval.hpp"
 #include "common/elf_info.h"
 #include "common/logging/log.h"
 #include "common/path_util.h"
@@ -43,6 +44,7 @@ struct SlotData {
 
 static std::mutex g_slot_mtx;
 static std::unordered_map<u32, SlotData> g_attached_slots;
+static size_t g_memory_size = 0;
 
 void PersistMemory(u32 slot_id, bool lock) {
     std::unique_lock lck{g_slot_mtx, std::defer_lock};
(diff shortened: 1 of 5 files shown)


georgemoralis:
@viniciuslrangel  save data fixes IXX

StevenMiller123:
Fixes God of War savedata crash.
[shad_log.txt](https://github.com/user-attachments/files/20243640/shad_log.txt)

viniciuslrangel:
Hi, tyvm. Before we merge, can you please check if any other error also fixes your side? IIRC, this func doesn't return NOT_FOUND. Maybe what we are missing is initializing memory_cache with zeroes in SetupSaveMemory, so when the game wants to read, we returning a big blob of empty zeroes instead of being truncated like it's right now as if never had written anything (because in this situation the file is never created).

mailwl:
i remove error, and allocate save memory with given size. Please check if it still works on other games with empty save (Penarium still works, save works, too)

src/core/libraries/save_data/save_memory.cpp
```diff
@@ -44,6 +44,7 @@ struct SlotData {
 
 static std::mutex g_slot_mtx;
 static std::unordered_map<u32, SlotData> g_attached_slots;
+static size_t g_memory_size = 0;
```
viniciuslrangel: Store this in SlotData struct instead of  global, games can have multiple save at the same time. You can store this during `SetupSaveMemory` call

viniciuslrangel (changes requested):

src/core/libraries/save_data/save_memory.cpp
```diff
@@ -196,9 +199,10 @@ void ReadMemory(u32 slot_id, void* buf, size_t buf_size, int64_t offset) {
     auto& data = g_attached_slots[slot_id];
     auto& memory = data.memory_cache;
     if (memory.empty()) { // Load file
+        memory.resize(data.memory_cache_size);
+        memset(memory.data(), 0, data.memory_cache_size);
```
viniciuslrangel: When resizing, the vector already sets the elements as default-initialization, so don't need to memset again

viniciuslrangel (changes requested):

StevenMiller123:
God of War still works properly.
[shad_log.txt](https://github.com/user-attachments/files/20256190/shad_log.txt)

viniciuslrangel (approved):

viniciuslrangel:
Thank you!

PR #2942, opened 2025-05-16
[Libs] Companion httpd

georgemoralis:
Added companion httpd and return disconnected device in GetEvents .  Helps Tearaway® Unfolded  spamming less connecting calls (but it still checks for it periodicaly so i added the function to Debug log)


src/core/libraries/companion/companion_httpd.cpp
```diff
@@ -0,0 +1,141 @@
...
+s32 PS4_SYSV_ABI
+sceCompanionHttpdGet2ndScreenStatus(Libraries::UserService::OrbisUserServiceUserId) {
+    LOG_ERROR(Lib_CompanionHttpd, "(STUBBED) called");
+    return ORBIS_OK;
+}
+
+s32 PS4_SYSV_ABI sceCompanionHttpdGetEvent(OrbisCompanionHttpdEvent* pEvent) {
+    pEvent->event = 0x10000002; // disconnected
+    LOG_DEBUG(Lib_CompanionHttpd, "device disconnected");
+    return 0x80E40008; // No events to obtain
```
squidbus: Probably should start some error constants for this like the other libraries.

georgemoralis: ok!

georgemoralis: done

squidbus (approved):

PR #2944, opened 2025-05-16
Network Play: set user signed in

mailwl:
This PR may leads to great incompatibility. Need heavy testing on games uses Np.
It makes "CUSA08076 - Sine Mora EX" playable
removes "Error 0x80550006" dialog on WATCH DOGS and some other games (WD still stucks on loading screen)


squidbus:
I don’t think this is a good idea since it will lead to games expecting the user to be signed in to PSN and try many other things, instead we should make sure other things function as expected when signed out.

georgemoralis:
if at least that should be added as an option , for further testing

georgemoralis:
ok with the config it should be ok

PR #2948, opened 2025-05-17
Fix passing user data in user-triggered equeue events

mikusp:
RE2 calls sceKernelTriggerUserEvent and then expects the .udata field to contain the provided data


georgemoralis (approved):

squidbus:
I believe there are at least a few places assuming this would write `data` and not `udata` that need to be updated, thinking mainly of some stuff with Gnm events.

squidbus:
Or alternatively, make it so `data` and `udata` can be updated as needed, and from this user event code path update `udata`. Maybe a `TriggerUser` path if the filter matches similar to how we have `TriggerDisplay`.

georgemoralis:
@mikusp any update on this?

mikusp:
@georgemoralis created TriggerUser as requested

squidbus:
I don’t see a change in the diff

PR #2949, opened 2025-05-17
Interpret (PthreadT)-1 as current thread in posix_pthread_setaffinity_np

mikusp:
I don't remember which game triggered that

### DIFF
--- a/src/core/libraries/kernel/threads/pthread.cpp
+++ b/src/core/libraries/kernel/threads/pthread.cpp
@@ -585,6 +585,9 @@ int PS4_SYSV_ABI posix_pthread_setaffinity_np(PthreadT thread, size_t cpusetsize
     if (thread == nullptr || cpusetp == nullptr) {
         return POSIX_EINVAL;
     }
+    if (thread == reinterpret_cast<PthreadT>(-1)) {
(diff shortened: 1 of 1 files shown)


squidbus:
I don't see any indication the real `libkernel` would treat it this way, there's:
* A check if it matches the current thread pointer in the thread-local data, which is not `-1`.
* A check if it is null to return an error.

And otherwise it is used like a normal pointer.

mikusp:
It's in the kernel itself, that function ends up calling cpuset_which and here the id is compared to -1 https://github.com/freebsd/freebsd-src/blob/release/9.0.0/sys/kern/kern_cpuset.c#L432

squidbus:
> It's in the kernel itself, that function ends up calling cpuset_which and here the id is compared to -1 

Yes but that is only directly called when the input parameter matches the pointer in TLS, otherwise it goes through a code path that dereferences the pointer.

squidbus:
excerpt

You can see from this excerpt of `pthread_setaffinity_np` that `cpuset_setaffinity` is used with -1 when the input thread pointer is self, but the actual parameter to `pthread_setaffinity_np` is not -1.

PR #2951, opened 2025-05-17
Savefixes X

viniciuslrangel:
respect install dir in param.sfo to select the game save folder

### DIFF
--- a/src/core/libraries/save_data/savedata.cpp
+++ b/src/core/libraries/save_data/savedata.cpp
@@ -316,7 +316,9 @@ static std::array<std::optional<SaveInstance>, 16> g_mount_slots;
 
 static void initialize() {
     g_initialized = true;
-    g_game_serial = ElfInfo::Instance().GameSerial();
+    g_game_serial = Common::Singleton<PSF>::Instance()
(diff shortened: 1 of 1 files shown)


MajorP93:
With this my Bloodborne savedata is not picked up anymore.
Previous commit was working fine.
I am using the AppImage on Arch Linux.

StevenMiller123:
Some versions of Bloodborne are supposed to save to a different folder on real hardware.
For example, CUSA03173 is supposed to save in the CUSA00207 folder, instead of CUSA03173.
This PR fixes inaccuracies related to this, as some games were broken due to our inaccurate implementation.

MajorP93:
> Some versions of Bloodborne are supposed to save to a different folder on real hardware. For example, CUSA03173 is supposed to save in the CUSA00207 folder, instead of CUSA03173. This PR fixes inaccuracies related to this, as some games were broken due to our inaccurate implementation.

I see. Thanks for the explanation.

MajorP93:
Right clicking a game --> Open Folder --> Open Save Data Folder opens the old folder which is not being picked up during emulation by the way.
Feels a bit misleading.
I feel like the change that got implemented in this PR is not yet reflected by the before-mentioned UI option.

PR #2963, opened 2025-05-20
Libs: CompanionUtil

Fire-Cube:
In this PR, I tried to reproduce the behavior of sceCompanionUtilGetEvent as minimally as possible. The game I used (Star Wars Battlefront) never calls Initialize.
Maybe I misinterpreted something in the RE, so a thorough review would be great.


src/core/libraries/companion/companion_util.cpp
```diff
@@ -0,0 +1,83 @@
...
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "common/logging/log.h"
+#include "core/libraries/error_codes.h"
+#include "core/libraries/libs.h"
+#include "core/libraries/companion/companion_util.h"
+
+namespace Libraries::CompanionUtil {
+
+uint32_t PS4_SYSV_ABI getEvent(sceCompanionUtilContext* ctx, sceCompanionUtilEvent* outEvent,
```
kalaposfos13:
```suggestion
u32 PS4_SYSV_ABI getEvent(sceCompanionUtilContext* ctx, sceCompanionUtilEvent* outEvent,
```
You can use the typedefs of the project for these, and same for all the other occurrences

src/core/libraries/companion/companion_util.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Core::Loader {
+class SymbolsResolver;
+}
+
+namespace Libraries::CompanionUtil {
+
+static constexpr uint32_t ORBIS_COMPANION_UTIL_OK = 0;
+
+static constexpr uint32_t ORBIS_COMPANION_UTIL_INVALID_ARGUMENT = 0x80000004;
```
kalaposfos13:
```suggestion
static constexpr uint32_t ORBIS_COMPANION_UTIL_INVALID_ARGUMENT = 0x80AD0004;
```
This way you can drop the bitwise AND with `0xad0000` in the return codes, since those are only used to achieve this, just in a way where the library ID is easily swappable

georgemoralis: error codes should go to [companion_error.h](https://github.com/shadps4-emu/shadPS4/blob/main/src/core/libraries/companion/companion_error.h)

src/core/libraries/companion/companion_util.h
```diff
@@ -0,0 +1,37 @@
...
+static constexpr uint32_t ORBIS_COMPANION_UTIL_INVALID_POINTER = 0x80000006;
+static constexpr uint32_t ORBIS_COMPANION_UTIL_NO_EVENT = 0x80000008;
+
+struct sceCompanionUtilEvent {
```
georgemoralis, line 16: the struct should be similar to the one i added in companion_httpd i will check

PR #2969, opened 2025-05-22
Texture cache: Relaxing constraint on image type info, adding loggin and separate expandimage to improve overlap handling.

diegolix29:
this pr tries to fit a better way of handling the overlaps, adds loggin to know whats expanding or reusing

help with this assert, on ps1 game i had converted to boot on ps4 console (Chrono Cross) and now tried on the emulator

image

if i just skip the assert also i dont see this first image when booting the game, it only show a white screen. 

image

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -223,11 +223,28 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
         }
 
         ImageId new_image_id{};
-        if (image_info.type == tex_cache_image.info.type) {
-            ASSERT(image_info.resources > tex_cache_image.info.resources);
+        if (image_info.resources.levels > tex_cache_image.info.resources.levels ||
+            image_info.resources.layers > tex_cache_image.info.resources.layers) {
+            LOG_INFO(Render,
+                     "Expanding image at guest address 0x{:X}: resources ({}x{}) -> ({}x{})",
+                     image_info.guest_address, tex_cache_image.info.resources.levels,
+                     tex_cache_image.info.resources.layers, image_info.resources.levels,
+                     image_info.resources.layers);
             new_image_id = ExpandImage(image_info, cache_image_id);
+        } else if (image_info.resources.levels == tex_cache_image.info.resources.levels &&
+                   image_info.resources.layers == tex_cache_image.info.resources.layers) {
+            LOG_INFO(Render, "Reusing exact match at guest address 0x{:X}: resources ({}x{})",
+                     image_info.guest_address, image_info.resources.levels,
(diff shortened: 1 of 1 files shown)


src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -223,11 +223,28 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
         }
 
         ImageId new_image_id{};
-        if (image_info.type == tex_cache_image.info.type) {
-            ASSERT(image_info.resources > tex_cache_image.info.resources);
```
squidbus, line 227: The reason this assert is here is because given the previous conditions (address is equal, dimensions are equal, format is equal, type is equal, byte size is greater than or equal), having fewer layers/levels should not be possible.

Could you share a log-out of the parameters of `image_info` and `tex_cache_image.info` here when this happens?

src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -223,11 +223,28 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
...
+        if (image_info.resources.levels > tex_cache_image.info.resources.levels ||
+            image_info.resources.layers > tex_cache_image.info.resources.layers) {
+            LOG_INFO(Render,
+                     "Expanding image at guest address 0x{:X}: resources ({}x{}) -> ({}x{})",
+                     image_info.guest_address, tex_cache_image.info.resources.levels,
+                     tex_cache_image.info.resources.layers, image_info.resources.levels,
+                     image_info.resources.layers);
```
squidbus, line 232: These should probably be debug or even trace if you want to add logs, the frequency would be too much.

diegolix29: yes in this case i added like this cause the game spams a lot of stubs so i could kind catch one during without having to digest the log, but i know now that it works as it pass some, expand some and reuses some, the log is working ok, i will change them to debug then. should the idetifier be still just Render, or u want render_vulkan?

src/video_core/texture_cache/texture_cache.cpp
diegolix29: i added more logging info from the screenshot i send on discord, this is the final veredict. so if u want i can make it simpler now, to reuse when matching and then expand when needed, 
image

squidbus: I think I see the case here, would ask that you try https://github.com/shadps4-emu/shadPS4/pull/2978 which is a more targeted carve-out for this case with details.

squidbus:
Going to close these as I'm pretty sure now this is going to be a size difference due to different tiling modes. Will figure out which way to handle that case and open something for it later.

PR #2973, opened 2025-05-22
shader_recompiler: Fix buffer type reading from step rate attribute.

squidbus:
`EmitReadConstBuffer` is reading as `U32` when we need `F32` for this case, causing a crash since the buffer is not defined with the `U32` alias. Fix by separating out a template function which can vary the alias.


georgemoralis:
fixed a sirit error in CUSA10416 - Unravel TWO  (Compiling vs shader 0xc61b571e)

PR #2974, opened 2025-05-22
amdgpu: Handle 32-bit Unorm formats.

squidbus:
Map 32-bit Unorm formats to Uint, and apply Uint <-> Unorm number conversion.


georgemoralis:
Unravel TWO seems to render stuff now 

εικόνα

Hermiten:
Fast debug :O

PR #2976, opened 2025-05-22
Devtools: Add Module Viewer

Fire-Cube:
This PR adds a new menu which shows what LLE modules was loaded and if it is from game modules dir or sys_modules from user dir

### DIFF
--- a/src/core/devtools/widget/module_list.cpp
+++ b/src/core/devtools/widget/module_list.cpp
@@ -16,10 +16,6 @@ namespace Core::Devtools::Widget {
 std::filesystem::path ModuleList::game_folder;
 
 void ModuleList::Draw() {
-    {
-        std::scoped_lock lock(modules_mutex);
-    }
-
     SetNextWindowSize({550.0f, 600.0f}, ImGuiCond_FirstUseEver);
     if (!Begin("Module List", &open)) {
         End();
@@ -32,6 +28,7 @@ void ModuleList::Draw() {
         TableSetupColumn("Modulname", ImGuiTableColumnFlags_WidthStretch);
         TableHeadersRow();
 
+        std::scoped_lock lock(modules_mutex);
         for (const auto& module : modules) {
             TableNextRow();
 
--- a/src/core/devtools/widget/module_list.cpp
+++ b/src/core/devtools/widget/module_list.cpp
@@ -20,17 +20,18 @@ void ModuleList::Draw() {
         for (const auto& entry : s_modules) {
             ModuleInfo info;
             info.name = entry.name;
+            info.is_sys_module = IsSystemModule(entry.path);
             modules.push_back(info);
         }
     }
 
-    SetNextWindowSize({500.0f, 600.0f}, ImGuiCond_FirstUseEver);
+    SetNextWindowSize({550.0f, 600.0f}, ImGuiCond_FirstUseEver);
     if (!Begin("LLE Module List", &open)) {
         End();
         return;
     }
 
-    if (BeginTable("ModuleTable", 1,
+    if (BeginTable("ModuleTable", 2,
                    ImGuiTableFlags_Borders | ImGuiTableFlags_Resizable | ImGuiTableFlags_Sortable |
                        ImGuiTableFlags_RowBg)) {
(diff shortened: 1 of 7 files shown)


georgemoralis:
Can you provide a screenshot?:D

Fire-Cube:
> Can you provide a screenshot?:D

Looks like this
image

georgemoralis:
might be interesting if possible to add all the modules that games load and have a column that specifies if it is LLE or HLE (just a thought)

Fire-Cube:
> might be interesting if possible to add all the modules that games load and have a column that specifies if it is LLE or HLE (just a thought)

I thought also about this but I couldn't find out where the HLE gets loaded

Fire-Cube:
I have an idea

mailwl:
this PR includes indirect loaded libs, like:
`[Lib.SysModule] <Error> sysmodule.cpp:88 sceSysmoduleLoadModule: (DUMMY) called module = ORBIS_SYSMODULE_GAME_LIVE_STREAMING` ?

Fire-Cube:
Now it is like this
image

src/core/devtools/widget/module_list.cpp
```diff
@@ -0,0 +1,61 @@
...
+using namespace ImGui;
+
+namespace Core::Devtools::Widget {
+
+std::filesystem::path ModuleList::game_folder;
+
+void ModuleList::Draw() {
+    {
+        std::scoped_lock lock(modules_mutex);
+    }
```
viniciuslrangel: This lock is not doing anything, move this to right before the for-loop

src/core/devtools/widget/module_list.h
```diff
@@ -0,0 +1,89 @@
...
+        bool is_lle = false;
+        auto it = std::find_if(modules.begin(), modules.end(),
+                               [&name, is_sys_module, is_lle](const ModuleInfo& entry) {
+                                   return entry.name == name && entry.is_lle == false;
+                               });
+
+        if (it == modules.end()) {
+            modules.push_back({name, is_sys_module, is_lle});
+        }
+    }
```
viniciuslrangel: You could move these implementations to the cpp file

src/emulator.cpp
```diff
@@ -75,6 +76,8 @@ void Emulator::Run(const std::filesystem::path& file, const std::vector<std::str
         }
     }
 
+    Core::Devtools::Widget::ModuleList::SetGameFolder(game_folder);
+
```
viniciuslrangel: You could store this in `ElfInfo`, it may be useful to future code

src/core/devtools/widget/module_list.h
```diff
@@ -0,0 +1,89 @@
...
+    static void AddModule(const std::string& name, std::filesystem::path path) {
+        if (name == "eboot.bin") {
+            return;
+        }
+        std::scoped_lock lock(modules_mutex);
+        modules.push_back({name, IsSystemModule(path), true});
+    }
+
+    static void AddModule(std::string& name) {
+        name = name + ".prx";
```
viniciuslrangel: Don't edit reference parameters 

```suggestion
    static void AddModule(std::string name) {
        name = name + ".prx";
```

src/core/devtools/widget/module_list.h
```diff
@@ -0,0 +1,89 @@
...
+        std::scoped_lock lock(modules_mutex);
+
+        bool is_sys_module = IsSystemModule(name);
+        bool is_lle = false;
+        auto it = std::find_if(modules.begin(), modules.end(),
+                               [&name, is_sys_module, is_lle](const ModuleInfo& entry) {
+                                   return entry.name == name && entry.is_lle == false;
```
viniciuslrangel:
```suggestion
                                   return entry.name == name && !entry.is_lle;
```

viniciuslrangel (changes requested):

src/core/devtools/widget/module_list.h
Fire-Cube: I’d rather leave it there so that GUI and logic remain separate

viniciuslrangel: This is not a good usage in general, but for this debug, I guess it's ok

viniciuslrangel (approved):

PR #2978, opened 2025-05-22
texture_cache: Handle greater guest_size with equal resources.

squidbus:
Handle case in texture cache where two resolved images are essentially equal, likely indicating a leftover image that needs to be cleaned up.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -222,14 +222,15 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
                 -1, -1};
         }
 
-        ImageId new_image_id{};
-        if (image_info.type == tex_cache_image.info.type) {
-            ASSERT(image_info.resources > tex_cache_image.info.resources);
-            new_image_id = ExpandImage(image_info, cache_image_id);
-        } else {
-            UNREACHABLE();
+        ASSERT(image_info.type == tex_cache_image.info.type);
+        if (image_info.resources == tex_cache_image.info.resources) {
+            // Could happen due to texture and render target guest_size mismatch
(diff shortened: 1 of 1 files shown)


rafael-57:
What should we expect from this When testing? Less memory usage/performance improvements?

squidbus:
Expectation is to avoid the ASSERT that was there previously.

diegolix29:
it throwed the assert on merged image now
image

squidbus:
> it throwed the assert on merged image now

Can you share the `guest_size` values from the data you sent previously, I don't see how it's possible to get here without any merge image ID.

squidbus:
Please check again.

diegolix29:
yes seems like its working no assert. But now when im ingame i have this on some textures sometimes and it corrects itself after a few 5 seconds or 8 
image

like this

image

squidbus:
Okay I made another tweak, can you see if that still happens?

diegolix29:
yes that big one aint happening anymore 

even show other lines that wasnt shownig before, but it does create this still and that one was also on the previous version

image

image

image

image
and they correct them selves sometimes
image


it just updates every few seconds like this 

https://github.com/user-attachments/assets/9be4e0c1-6733-4c58-9974-fd249b88f3b9

squidbus:
If you have time maybe try changing the case added to return just ‘merged_image_id’ unconditionally (I’m away from PC now), if it still has those glitches we may need to abandon the assert change and look deeper into why the size is different. Perhaps printing the image tiling modes in your previous logging would help identify as well.

Xcedf:
This assert fixed in Battlefield 4, game compiles few more shaders but still gets stuck.

squidbus:
Going to close these as I'm pretty sure now this is going to be a size difference due to different tiling modes. Will figure out which way to handle that case and open something for it later.

PR #2980, opened 2025-05-23
Added back the "Attempted to track non-GPU memory" assert.

LNDF:
Bring back the assert that I removed when rewritting part of the page_manager.

Please test this before in case the assert is too strct.


gandalfthewhite19890404:
Yes, Diablo III (updated versions, not base 1.00) hits it. So better with assert to understand what is going on instead of silent crash.
```
[Debug] <Critical> page_manager.cpp:220 operator(): Assertion Failed!
Attempted to track non-GPU memory at address 0x1052140000, size 0x40000.
```

LNDF:
I think this is ok now

PR #2981, opened 2025-05-23
texture_cache: Handle overlap with equal address and different tiling mode

squidbus:
Handles case where we reached the end of equal guest address overlap, the `guest_size` is greater, the `resources` count is not, and the tiling mode is different. In this case the address is likely being reused for a different image, so free the cached image and return the current merged image ID.


squidbus:
@diegolix29 Please check the game you had the texture cache assert with

diegolix29:
Seems like its working ok with this, still see some weird graphics but it might be an issue with the method i used for the original convertion, i will test with another ps1 converted game and see if i see irregularities. Thanks this is def better than yesterday

PR #2985, opened 2025-05-25
shader_recompiler: Fix incorrect float type on FPRecip64

LNDF:
Currently erroneously a float is emitted insteed of a double, cusing a SPIR-V validation error.

For example, in InFamous: First Light (shader 0xa71733ca):
```
error: line 256: Invalid word count: OpConstant starting at word 1043 says it has 4 words, but found 5 words instead.
```


AboMedoz:
tested fixes the error

PR #2990, opened 2025-05-25
Wip implementation of isdevkit and neo mode to experimental tab.

diegolix29:
i will have to adjust the box later on please test locally it worked ok

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -473,6 +473,10 @@ void setNeoMode(bool enable) {
     isNeo = enable;
 }
 
+void setDevKitMode(bool enable) {
+    isDevKit = enable;
+}
+
 void setLogType(const std::string& type) {
     logType = type;
 }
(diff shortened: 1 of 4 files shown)


StevenMiller123:
These features are intentionally excluded from the GUI, as we don't want inexperienced users enabling them.

diegolix29:
i saw somewhere that putting them on a experimental tab with descriptions could be a good choice as some people dont know how to enable some things on the config toml. and devkit and neo mode are needed for some specific stuff some games need it and some patches now too.

GHU7924:
I think that it is possible to do it in such a way that it does not interfere with users who do not need it, and those who do need it, so that they do it at their own risk.

EF

@diegolix29 This is what my tab looks like at the moment.

изображение

GamerzHell9137:
Is there a chance to add a message on the lower bar on the main window of shad that experimental options are enabled? 
Something like:
image

georgemoralis:
No

PR #2991, opened 2025-05-26
shader_recompiler: Patch SRT walker on segfault

LNDF:
This PR introduces patching the SRT walker when an access violation happens. Access violation can happen because SRT wlker is not 100% accurate to the base shader (for example, arithmetic operations between base ReadConst/UserData are omited). This can lead to a situiation where the SRT waslker reads junk data or, when indirection happens, can even cause access violation. Additionaly, we omit ReadConst with non.immediate offset from the SRT walker, increesing the cnances of an access violation.

As far as I know, flatbuf is used in 3 instances.
1. To implement S_LOAD_DEORD (only as a fallback since #2819 ),
1. To get buffer size in shaders,
1. In the resource tracking pass to track sharps

The goal of this PR is to let the emulation continue if an access violation occurs on SRT walker, providing a fallback value (0). The idea is that most games that use this level of indirection that could lead to an access violation fall into the first case, meaning that we would only (maybe) get an artifact for a frame.

This lets Until Dawn go further.

Note that this doesn't aim to fix not being able to track resources flatbuf inaccuracies (case 3). This happens on Persona 3 (which is a UE game).

Note that this is my first time dynamic-patching instructions, so I might have done something horribly wrong :)

Draft because I would like to be tested first.


Missake212:
Until Dawn (CUSA02636) now goes further thanks to this PR. Still crashes before menus though.

Screenshot 2025-05-26 122749

Screenshot 2025-05-26 122752

Screenshot 2025-05-26 122812

[CUSA02636.log](https://github.com/user-attachments/files/20440376/CUSA02636.log)

I also tested all my games (on Windows) and none of them regressed.

LNDF:
This should be ok I think. No regressions found and makes until dawn go further.

georgemoralis:
@squidbus  any thoughts?

LNDF:
Maybe I should add a comment in the code explaining why this is done, for anyone reading the code. It seems a bit random without having the full context.

raphaelthegreat:
This is quite hackish. As a temporary solution its probably okay, but should not be done in the long term imo. It can be solved better by having case 1 only handled by DMA, which requires all the other changes discussed in the DMA PR

LNDF:
> This is quite hackish. As a temporary solution its probably okay, but should not be done in the long term imo. It can be solved better by having case 1 only handled by DMA, which requires all the other changes discussed in the DMA PR

Currently, you would only get erroneous data for 1 frame in case 1 {after implementing the option to cache all mapped ranges, you would never get erroneous data). However the SRT walker would still run every time a drawcall (for sharp tracking) so this signal handler would be still needed (or as an alternative, all non sharp readconsts could be skipped if the option to cache all mapped regions is enabled).

Case 2 can be viewed as case 1. It is really accessing memory through a shader and can be implemented with DMA, getting erroneous values for 1 frame (or 0 after the PR mentioned above).

Case 3 is the problem as we need the sharps at shader compile time on the CPU side. Maybe a more accurate SRT walker can be built for that, or an entirely new system could be used.

raphaelthegreat:
> all non sharp readconsts could be skipped if the option to cache all mapped regions is enabled

Yes this would prevent the majority of these segfaults and if they still didnt happen, it could be tweaked to handle them. 

The [first part](https://github.com/shadps4-emu/shadPS4/blob/99ccf569389240b0261c716ae389dedc43f6bf95/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp#L835-L846) of resource_tracking and flatten_extended_userdata could be combined into a single pass, so that ReadConst coming from resources are only included in it. That pass will also eliminate those ReadConst instructions leaving only ReadConst that DMA will handle. And then the second part of resource_tracking could be combined with lower_buffer_format_to_raw as a sort of resolve pass.

So with this case 1 and 2 will go through DMA only and flatbuf only case 3, which is good because its scope is limited as much as possible, to the data that is needed at compile time and fetched every draw.

diogoaduartereis:
[CUSA02636.log](https://github.com/user-attachments/files/20446105/CUSA02636.log)
I was able to see Sam's black silhouette in the main menu, but it crashed immediately: 
image

LNDF:
> The [first part](https://github.com/shadps4-emu/shadPS4/blob/99ccf569389240b0261c716ae389dedc43f6bf95/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp#L835-L846) of resource_tracking and flatten_extended_userdata could be combined into a single pass, so that ReadConst coming from resources are only included in it.

Will take a look at it.

> That pass will also eliminate those ReadConst instructions leaving only ReadConst that DMA will handle. And then the second part of resource_tracking could be combined with lower_buffer_format_to_raw as a sort of resolve pass.

The dead code elimination already takes care of those readconsts.

> So with this case 1 and 2 will go through DMA only and flatbuf only case 3, which is good because its scope is limited as much as possible, to the data that is needed at compile time and fetched every draw.

I would also like to put a (disabled by default) option to not cache all mapped regions (for lower end systems with not a lot of VRAM).


Additionaly, current srt walker is not accurate enouth for some games. Persona 3 for example uses incorrect data from flatbuf as V# descriptor, using a random memory location as a buffer (and triggering the attempted to track non-GPU memory assert). 
 For those cases, a more complex walker generator must be implemented (maybe parts from my x64 ir backend could be used?).

raphaelthegreat:
> Additionaly, current srt walker is not accurate enouth for some games
> For those cases, a more complex walker generator must be implemented

Yes the srt walker generation code can be updated to handle specific cases of sharp fetching (e.g bindless) as needed. I don't think an x64 ir backend will be needed in vast majority of cases with some smart pattern detection

PR #2994, opened 2025-05-26
Equeue: fix WaitEqueue assert on nullptr

Fire-Cube:
On real hardware nullptr is not a problem and based on RE WaitEqueue simply passes 0 to kqueue if timeout is nullptr

Pls test again before merge

### DIFF
--- a/src/core/libraries/kernel/equeue.cpp
+++ b/src/core/libraries/kernel/equeue.cpp
@@ -266,23 +266,16 @@ int PS4_SYSV_ABI sceKernelWaitEqueue(SceKernelEqueue eq, SceKernelEvent* ev, int
         return ORBIS_KERNEL_ERROR_EINVAL;
     }
 
+    // When the timeout is nullptr, we wait indefinitely
     if (eq->HasSmallTimer()) {
-        ASSERT(timo && *timo);
-        *out = eq->WaitForSmallTimer(ev, num, *timo);
+        *out = eq->WaitForSmallTimer(ev, num, (timo != nullptr) ? *timo : 0);
     } else {
-        if (timo == nullptr) { // wait until an event arrives without timing out
-            *out = eq->WaitForEvents(ev, num, 0);
-        }
-
-        if (timo != nullptr) {
+        if (*timo == 0) {
             // Only events that have already arrived at the time of this function call can be
             // received
-            if (*timo == 0) {
-                *out = eq->GetTriggeredEvents(ev, num);
-            } else {
-                // Wait until an event arrives with timing out
-                *out = eq->WaitForEvents(ev, num, *timo);
-            }
+            *out = eq->GetTriggeredEvents(ev, num);
+        } else {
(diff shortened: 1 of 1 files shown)


georgemoralis:
give us some screenies :D

Missake212:
UFC 4 now boots but gets stuck spamming net stuff.

image

[CUSA14209.zip](https://github.com/user-attachments/files/20447422/CUSA14209.zip)

The Sims 4 (with the update) now behaves the same way as version 1.00, meaning it crashes almost immediately with the last line in log being about SceVideoOutGetResolutionStatus

image

[CUSA09216.log](https://github.com/user-attachments/files/20447430/CUSA09216.log)

src/core/libraries/kernel/equeue.cpp
```diff
@@ -266,23 +267,23 @@ int PS4_SYSV_ABI sceKernelWaitEqueue(SceKernelEqueue eq, SceKernelEvent* ev, int
...
+            *out = eq->WaitForSmallTimer(ev, num, *timo);
+        }
     } else {
-        if (timo == nullptr) { // wait until an event arrives without timing out
+        if (timo == nullptr) {
             *out = eq->WaitForEvents(ev, num, 0);
-        }
-
-        if (timo != nullptr) {
+        } else if (*timo == 0) {
```
squidbus, line 284: It seems we have a discrepancy now where small, precise timers will wait indefinitely on 0 (small timer path) and larger timers will return immediately on 0 (here). I assume that the small timer path should do the same as here, if you could confirm the correct behavior and make them consistent.

Fire-Cube: I can confirm that (*timo == 0) is just a simple wait

Fire-Cube:
I think this is ready now
Based on hardware tests this should be correct

src/core/libraries/kernel/equeue.cpp
```diff
@@ -266,23 +267,27 @@ int PS4_SYSV_ABI sceKernelWaitEqueue(SceKernelEqueue eq, SceKernelEvent* ev, int
...
-        ASSERT(timo && *timo);
-        *out = eq->WaitForSmallTimer(ev, num, *timo);
+        if (timo == nullptr) {
+            *out = eq->WaitForSmallTimer(ev, num, 0);
+        } else if (*timo == 0) {
+            // Only events that have already arrived at the time of this function call can be
+            // received
+            *out = eq->GetTriggeredEvents(ev, num);
+        } else {
+            ASSERT(*timo);
```
squidbus: Should remove assert here since previous case already checks this.

squidbus:
I'm not sure I like the duplicatation of the `*timo == 0` block in both paths but I suppose it's fine. Mainly just the one comment about the assert then I'm fine with this.

src/core/libraries/kernel/equeue.cpp
Fire-Cube: forgot about it, is removed now

PR #2997, opened 2025-05-27
Core: More Memory Cleanup & Fixes

StevenMiller123:
The main change of this PR is completely removing our `PoolReserve` and `Reserve` functions in `Core::Memory`. Both of these functions were effectively re-implementations of `MapMemory` with minor changes, since all three functions are different uses of the `mmap` syscall.
Instead of constantly updating all three separately, it's easier to group all behavior in one function with additional checks where we need to behave slightly differently.

I've also added proper error handling in `posix_mmap`, added proper conditional rasterizer memory mapping logic based on protection values, and fixed a few other edge cases I've noticed in testing.

I'm opening this as a draft for now, as I need to make some more minor cleanup/style commits, and these changes need proper testing before any merges.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -314,7 +314,7 @@ s32 MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, u64 size, Memo
         // When MemoryMapFlags::Fixed is not specified, and mapped_addr is 0,
         // search from address 0x200000000 instead.
         alignment = alignment > 0 ? alignment : 16_KB;
-        mapped_addr = mapped_addr == 0 ? 0x200000000 : mapped_addr;
+        mapped_addr = virtual_addr == 0 ? 0x200000000 : mapped_addr;
         mapped_addr = SearchFree(mapped_addr, size, alignment);
         if (mapped_addr == -1) {
             // No suitable memory areas to map to
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -400,10 +400,7 @@ s32 MemoryManager::MapFile(void** out_addr, VAddr virtual_addr, u64 size, Memory
     impl.MapFile(mapped_addr, size_aligned, phys_addr, std::bit_cast<u32>(prot), handle);
 
     if (prot >= MemoryProt::GpuRead) {
-        // PS4s only map to GPU memory when the protection includes GPU access.
-        // If the address to map to is too high, PS4s throw a page fault and crash.
-        ASSERT_MSG(IsValidGpuMapping(mapped_addr, size_aligned), "Invalid address for GPU mapping");
-        rasterizer->MapMemory(mapped_addr, size_aligned);
+        ASSERT_MSG(false, "Files cannot be mapped to GPU memory");
     }
 
     // Add virtual memory area
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -396,7 +396,7 @@ s32 MemoryManager::MapFile(void** out_addr, VAddr virtual_addr, u64 size, Memory
 
     const auto handle = file->f.GetFileMapping();
 
-    impl.MapFile(mapped_addr, size_aligned, phys_addr, std::bit_cast<u32>(prot), fd);
+    impl.MapFile(mapped_addr, size_aligned, phys_addr, std::bit_cast<u32>(prot), handle);
 
     if (prot >= MemoryProt::GpuRead) {
         // PS4s only map to GPU memory when the protection includes GPU access.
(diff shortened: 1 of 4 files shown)


StevenMiller123:
Pushed up a fix for a regression seen in https://github.com/shadps4-emu/shadps4-game-compatibility/issues/341
@PirkyP If you can test this, please let me know if that regression is fixed.

src/core/memory.cpp
```diff
@@ -449,8 +389,22 @@ int MemoryManager::MapFile(void** out_addr, VAddr virtual_addr, size_t size, Mem
...
+    }
+
+    const auto handle = file->f.GetFileMapping();
+
+    impl.MapFile(mapped_addr, size_aligned, phys_addr, std::bit_cast<u32>(prot), handle);
+
+    if (prot >= MemoryProt::GpuRead) {
+        // PS4s only map to GPU memory when the protection includes GPU access.
+        // If the address to map to is too high, PS4s throw a page fault and crash.
+        ASSERT_MSG(IsValidGpuMapping(mapped_addr, size_aligned), "Invalid address for GPU mapping");
```
red-prig: As far as I know, regular files cannot be made visible on the GPU, so I think there should be an unconditional Assert here.

src/core/memory.cpp
```diff
@@ -563,8 +563,9 @@ s64 MemoryManager::ProtectBytes(VAddr addr, VirtualMemoryArea vma_base, size_t s
         vma_base.size - start_in_vma < size ? vma_base.size - start_in_vma : size;
 
     if (vma_base.type == VMAType::Free) {
-        LOG_ERROR(Kernel_Vmm, "Cannot change protection on free memory region");
-        return ORBIS_KERNEL_ERROR_EINVAL;
+        // On PS4, protecting freed memory does nothing.
+        LOG_WARNING(Kernel_Vmm, "Attempting to protect free memory");
```
red-prig: Free regions can simply be skipped in mprotect, this is expected behavior

StevenMiller123: That's what this change does. By returning size here, the loop we run will skip the free VMA and move to the next one. https://github.com/StevenMiller123/shadPS4/blob/e2215f0f4f47d5b5b1dfe3a414702ea19630d3f4/src/core/memory.cpp#L625
Will remove the log though, since it's probably not important to log this case.

src/core/memory.cpp
```diff
@@ -449,8 +389,19 @@ int MemoryManager::MapFile(void** out_addr, VAddr virtual_addr, size_t size, Mem
...
-    impl.MapFile(mapped_addr, size_aligned, offset, std::bit_cast<u32>(prot), fd);
+    // Get the file to map
+    auto file = h->GetFile(fd);
+    if (file == nullptr) {
+        return ORBIS_KERNEL_ERROR_EBADF;
+    }
+
+    const auto handle = file->f.GetFileMapping();
+
+    impl.MapFile(mapped_addr, size_aligned, phys_addr, std::bit_cast<u32>(prot), handle);
```
red-prig, line 400: In case of file mappings there is a complicated case when the file size is not a multiple of the page size, in this case the upper part of the mapped page (the virtual address is larger than the file) is always equally available for reading/writing and is filled with zeros by default, the problem arises firstly because Windows requires specifying an unaligned file size otherwise an error (Does Linux allow an aligned size larger than the file? Correct me) and also there is a problem with the fact that the host page multiple is 4KB and the guest page multiple is 16KB, so additional code will be needed to add private pages for alignment to the entire guest page (for example, the file size is 3KB, it turns out one virtual page of 4KB for the host, so at the end you will need to add another 3 4KB private pages)

StevenMiller123: I'm not too familiar with our address space code yet, which is why my changes to our file mapping code are very minimal in this PR. All I did was move around a couple lines (and made that GPU mapping change I've already reverted).

From what I can tell, our implementation on both Windows and Linux allow for the aligned up size we've been using here. On Windows we're using MapViewOfFile3 for most cases, and VirtualAlloc2 + ReadFile for read-only file mappings. On Linux we use the host's mmap. So far, I don't think I've seen any games crop up that fail on these calls, though we've got asserts in place to identify any errors.
As for reading the upper part of the page, apparently our current Linux implementation will throw a SIGBUS exception looking at their mmap docs. I'm not sure how our Windows implementation behaves here.

I'll try taking a look at this issue at some point, but I'm not sure if I'll fix that in this PR or not, since I'll need to learn how all these functions work first.

PirkyP: Sorry for the delay. I ran the test and it's a different error now. The boot screen displayed for a few seconds before it crashed. Let me know if I need to retest it.

`[Debug] <Critical> vector_memory.cpp:182 operator(): Assertion Failed!`
`Non immediate offset not supported`
[shad_log.txt](https://github.com/user-attachments/files/20474790/shad_log.txt)

StevenMiller123:
It doesn't seem like I've caused regressions, so I'm opening this up for review.

PR #2998, opened 2025-05-27
Stub PM4 0x8E opcode

mikusp:


kalaposfos13:
Shadow of the Colossus goes to menus with nullGPU, and has sound as well, but crashes upon the second input with an access violation in game code, which didn't happen pre-0.9.0 with hacks, so it might be a regression. The game is also prone to randomly hang at times.

[CUSA08809.log](https://github.com/user-attachments/files/20463968/CUSA08809.log)

gandalfthewhite19890404:
image

PR #2999, opened 2025-05-27
return nullptr from inet_ntop on error

mikusp:
Allows Beyond: Two Souls to go ingame

### DIFF
--- a/src/core/libraries/network/net.cpp
+++ b/src/core/libraries/network/net.cpp
@@ -962,7 +962,7 @@ const char* PS4_SYSV_ABI sceNetInetNtop(int af, const void* src, char* dst, u32 
     const char* res = inet_ntop(af, src, dst, size);
 #endif
     if (res == nullptr) {
-        UNREACHABLE();
+        return nullptr;
     }
     return dst;
 }


src/core/libraries/network/net.cpp
```diff
@@ -962,7 +962,7 @@ const char* PS4_SYSV_ABI sceNetInetNtop(int af, const void* src, char* dst, u32
     const char* res = inet_ntop(af, src, dst, size);
 #endif
     if (res == nullptr) {
-        UNREACHABLE();
+        return nullptr;
```
squidbus, line 965: Error case should also set the errno value in `*sceNetErrnoLoc()` with one of the `ORBIS_NET_E*` values.

georgemoralis: *sceNetErrnoLoc() = ORBIS_NET_EAFNOSUPPORT;

georgemoralis: a LOG_ERROR would be nice too as well

squidbus: > *sceNetErrnoLoc() = ORBIS_NET_EAFNOSUPPORT;

In this case it should be based on the inet_ntop error code shouldn’t it?

georgemoralis: based on RE that is the only valid sceNetErrno value it can take in case of failure

squidbus: I see `ORBIS_NET_ENOSPC` as well

squidbus: Which matches up with the documentation for `inet_ntop` which is why I think we should just convert the error code: https://www.man7.org/linux/man-pages/man3/inet_ntop.3.html

```
       EAFNOSUPPORT
              af was not a valid address family.

       ENOSPC The converted address string would exceed the size given by
              size.
```

georgemoralis: i think that it is weird that inet_ntop returns null anyway , should be debugged further

squidbus: Returning null on error is correct here, cleaned up rough flow from the function disassembly:
```
  if (param_1 == 0x1c) {
    return ...call some func...;
  }
  if (param_1 == 2) {
    if (...attempt to print out address...) {
      ...some additional copy stuff...
      return ...addr...;
    }
    *sceNetErrnoLoc() = ORBIS_NET_ENOSPC;
    return nullptr;
  }
  *sceNetErrnoLoc() = ORBIS_NET_EAFNOSUPPORT;
  return nullptr;
```

squidbus: And with regards to `inet_ntop`:
```
On success, inet_ntop() returns a non-null pointer to dst.
NULL is returned if there was an error, with errno set to indicate the error.
```

mikusp: @georgemoralis the game calls the function with some bogus values, like af = 223 and af = 249

squidbus: If you want to see source it actually matches up to this (except not setting the errno): https://github.com/freebsd/freebsd-src/blob/main/sys/libkern/inet_ntop.c

georgemoralis: > @georgemoralis the game calls the function with some bogus values, like af = 223 and af = 249

in that case ORBIS_NET_EAFNOSUPPORT should be fine for it

georgemoralis: well i think i should rewrite it to match more the RE . Atm we we want to accept this i think at least ORBIS_NET_EAFNOSUPPORT be added somethink like

if(AF_NET ! = 2 || other valid AF_NET ){
    *sceNetErrnoLoc() = ORBIS_NET_EAFNOSUPPORT;
  return nullptr;
}

georgemoralis: and unreachable should left for the other cases

PR #3003, opened 2025-05-28
Implemented sceNetInetNtop in RE

georgemoralis:
fully (mostly) RE of sceNetInetNtop . This is added as replacement for #2999

### DIFF
--- a/src/core/libraries/network/net.cpp
+++ b/src/core/libraries/network/net.cpp
@@ -1078,29 +1078,22 @@ const char* freebsd_inet_ntop6(const char* src, char* dst, u64 size) {
     return (dst);
 }
 const char* PS4_SYSV_ABI sceNetInetNtop(int af, const void* src, char* dst, u32 size) {
-    char temp[16];
-    u32 len;
-
-    if (af == 0x1C) { // AF_INET6
-        if (src && dst) {
-            return freebsd_inet_ntop6((const char*)src, dst, size);
-        } else {
-            *sceNetErrnoLoc() = ORBIS_NET_ENOSPC;
-            LOG_ERROR(Lib_Net, "returned ORBIS_NET_ENOSPC");
-        }
+    if (!(src && dst)) {
+        *sceNetErrnoLoc() = ORBIS_NET_ENOSPC;
+        LOG_ERROR(Lib_Net, "returned ORBIS_NET_ENOSPC");
+        return nullptr;
     }
-    if (af == 2) { // AF_INET
-        if (src && dst) {
-            return freebsd_inet_ntop4((const char*)src, dst, size);
-        } else {
-            *sceNetErrnoLoc() = ORBIS_NET_ENOSPC;
-            LOG_ERROR(Lib_Net, "returned ORBIS_NET_ENOSPC");
-        }
-    } else {
+
+    switch (af) {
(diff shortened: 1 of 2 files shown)


squidbus:
I think you should take a look at the FreeBSD source linked in the other PR https://github.com/freebsd/freebsd-src/blob/main/sys/libkern/inet_ntop.c as it is, as far as I can tell, the same logic that is used in libSceNet, but presented a lot cleaner and includes IPv6 implementation.

georgemoralis:
hmm yah looks cleaner but generally the idea is the same :D . i will modified it

georgemoralis:
Added AF_INET with modified freebsd source to support error codes.

Need to look the code for IPv6 a bit more

georgemoralis:
Added freebsd_inet_ntop6 too

georgemoralis:
should be mergable now

src/core/libraries/network/net.cpp
```diff
@@ -955,16 +955,152 @@ u16 PS4_SYSV_ABI sceNetHtons(u16 host16) {
...
+        *tp++ = ':';
+    *tp++ = '\0';
+
+    /*
+     * Check for overflow, copy, and we're done.
+     */
+    if ((u64)(tp - tmp) > size) {
+        *sceNetErrnoLoc() = ORBIS_NET_ENOSPC;
+        LOG_ERROR(Lib_Net, "returned ORBIS_NET_ENOSPC");
+        return nullptr;
```
squidbus, line 1071: I think we can simplify these ENOSPC return cases by just putting into sceNetInetNtop directly that, if either of the v4/v6 functions return null, report ENOSPC. Since that’s the only error they can return.

src/core/libraries/network/net.cpp
```diff
@@ -955,16 +955,152 @@ u16 PS4_SYSV_ABI sceNetHtons(u16 host16) {
...
+    }
+    strcpy(dst, tmp);
+    return (dst);
+}
+const char* PS4_SYSV_ABI sceNetInetNtop(int af, const void* src, char* dst, u32 size) {
+    char temp[16];
+    u32 len;
+
+    if (af == 0x1C) { // AF_INET6
```
squidbus: I think we should make this function simpler by:
* Using switch case similar to FreeBSD version.
* Define constants for these address families (if not already defined for something else here) and use by name.
* Both paths are doing src and dst null check and returning ENOSPC error, instead just do that one time before this logic.

src/core/libraries/network/net.cpp
georgemoralis: since the if is a bit different in case i suggest we leave it there

src/core/libraries/network/net.cpp
georgemoralis: done

src/core/libraries/network/net.cpp
squidbus: Yeah but it returns nullptr in any ENOSPC error case, so we can just say

```if (returnvalue == nullptr) set errno to ENOSPC```

in the parent function.

georgemoralis: done

squidbus (approved):

PR #3006, opened 2025-05-29
Fix touchpad handling and change gyro calculation

WujekFoliarz:
- Both touch ids are now being tracked correctly
- Acceleration values are now more accurate to original libScePad
- Orientation calculation should be more accurate now


src/RCa10748
georgemoralis, line 1: that shouldn't be there for sure :D

CMakeSettings.json
georgemoralis, line 1: that probably shouldn't be here either

georgemoralis:
looked your library btw very interesting ;) (duaLib)

georgemoralis:
Have anyone checked this if there is any more issues? Code looks OK need play confirmation

WujekFoliarz:
@kalaposfos13 said it works fine last time he checked it, I'll double check later today

WujekFoliarz:
Ok I just checked the orientation in Gravity Rush 2 and it's pretty stable actually

WujekFoliarz:
https://github.com/user-attachments/assets/56bd46b4-6d49-4b46-b3cb-f0f00a591bfc

PR #3007, opened 2025-05-29
Handle DS_READ_U16, DS_WRITE_B16, DS_ADD_U64

mikusp:
Hit by God of War


src/shader_recompiler/backend/spirv/emit_spirv_shared_memory.cpp
```diff
@@ -6,6 +6,13 @@
 
 namespace Shader::Backend::SPIRV {
 
+Id EmitLoadSharedU16(EmitContext& ctx, Id offset) {
+    const Id shift_id{ctx.ConstU32(2U)};
```
squidbus: Are you sure shift by 2 (divide by 4) is still correct for a 16-bit value?

src/shader_recompiler/backend/spirv/emit_spirv_shared_memory.cpp
```diff
@@ -23,6 +30,13 @@ Id EmitLoadSharedU64(EmitContext& ctx, Id offset) {
                                     ctx.OpLoad(ctx.U32[1], rhs_pointer));
 }
 
+void EmitWriteSharedU16(EmitContext& ctx, Id offset, Id value) {
+    const Id shift{ctx.ConstU32(2U)};
```
squidbus: Same as read case.

src/shader_recompiler/backend/spirv/emit_spirv_shared_memory.cpp
```diff
@@ -6,6 +6,13 @@
 
 namespace Shader::Backend::SPIRV {
 
+Id EmitLoadSharedU16(EmitContext& ctx, Id offset) {
+    const Id shift_id{ctx.ConstU32(2U)};
+    const Id index{ctx.OpShiftRightArithmetic(ctx.U32[1], offset, shift_id)};
+    const Id pointer = ctx.OpAccessChain(ctx.shared_u16, ctx.shared_memory_u32, index);
```
squidbus: Pretty sure this will need a separate alias for u16 access instead of shared_memory_u32.

src/shader_recompiler/backend/spirv/emit_spirv_shared_memory.cpp
```diff
@@ -23,6 +30,13 @@ Id EmitLoadSharedU64(EmitContext& ctx, Id offset) {
                                     ctx.OpLoad(ctx.U32[1], rhs_pointer));
 }
 
+void EmitWriteSharedU16(EmitContext& ctx, Id offset, Id value) {
+    const Id shift{ctx.ConstU32(2U)};
+    const Id word_offset{ctx.OpShiftRightArithmetic(ctx.U16, offset, shift)};
+    const Id pointer = ctx.OpAccessChain(ctx.shared_u16, ctx.shared_memory_u32, word_offset);
```
squidbus: Same as read case.

src/shader_recompiler/backend/spirv/emit_spirv_shared_memory.cpp
mikusp: No idea, documentation doesn't really say that, they use `addr` for all accesses 16/32/64bit

squidbus: Well I suppose regardless of guest requirements you do need to have the right index for whatever we are using on the SPIR-V side, for example if we have a shared memory u16 array you need to shift 1 over to index by u16.

red-prig: Since you are using an array of 32 bit values, so dividing by 4 will be the base index, and then you need to take either the upper or lower part depending on the first bit:
`result = ((offset & 1)=0) ? buf[offset/4] & 0xFFFF : buf[offset/4] & 0xFFFF0000 `

red-prig: Oh, the second bit is not the first, lol
`result = ((offset & 2)=0) ? buf[offset/4] & 0xFFFF : buf[offset/4] & 0xFFFF0000 `

squidbus: The approach mentioned by @red-prig here is probably correct if we don't have a way in SPIR-V to alias the same shared memory array as a different type (I don't remember if this is the case). The code as of current revision creates two separate arrays which could be problematic if a shader mixes access sizes.

red-prig: Creating aliases usually works fine, the problem is more likely in addressing a 16-bit value, not all GPUs support this, and those that do may work slower

squidbus: I think we are fine in that regard given we already may do so for buffers.

mikusp: I'm working on a better implementation with aliasing definitions for u16/32/64, as the partial access mentioned by red_prig won't work for atomic operations

src/shader_recompiler/backend/spirv/emit_spirv_bounds.h
```diff
@@ -0,0 +1,46 @@
...
+        const Id in_bounds = ctx.OpULessThan(ctx.U1[1], index, buffer_size);
+        const Id ib_label = ctx.OpLabel();
+        const Id oob_label = ctx.OpLabel();
+        const Id end_label = ctx.OpLabel();
+        ctx.OpSelectionMerge(end_label, spv::SelectionControlMask::MaskNone);
+        ctx.OpBranchConditional(in_bounds, ib_label, oob_label);
+        ctx.AddLabel(ib_label);
+        const auto ib_result = emit_func();
+        ctx.OpBranch(end_label);
+        ctx.AddLabel(oob_label);
```
LNDF: You don't need to add an empty else branxh. You can use the merge label as the else branch in the OpBranchConditional.

Then, for the OpPhi, you can save ctx.last_label into a variable before any AddLabel and use that as the predecesor block for the zero value.

src/shader_recompiler/backend/spirv/emit_spirv_shared_memory.cpp
```diff
@@ -1,43 +1,86 @@
...
+        return ctx.OpLoad(ctx.U64, pointer);
+    });
+}
+
+void EmitWriteSharedU16(EmitContext& ctx, Id offset, Id value) {
+    const Id shift{ctx.ConstU32(1U)};
+    const Id index{ctx.OpShiftRightLogical(ctx.U32[1], offset, shift)};
+    const u32 num_elements{Common::DivCeil(ctx.runtime_info.cs_info.shared_memory_size, 2u)};
+
+    AccessBoundsCheck<16>(ctx, index, ctx.ConstU32(num_elements), [&] {
```
LNDF, line 52: You could return an std::optional (or just an invalid Id). Then in AccessBoundsCheck check if you returned a valid value and only then emit the OpPhi. This also applies to other instances when you return an Undef.

Notice also that you have different behaviour if the store is inbounds (you return undef) or it is out of bounds (you return zero).

src/shader_recompiler/backend/spirv/emit_spirv_bounds.h
mikusp: Fixed now

src/shader_recompiler/backend/spirv/emit_spirv_shared_memory.cpp
mikusp: The different return value should be fine as writes ignore the returned value. For the optional OpPhi, I've tried a couple of solutions but cannot figure out how to make it compile 🤔

LNDF: Use an std::optional or return an invalid Id. Then check for it in AccessBoundsCheck and only add the phi if valid. That will save an invalid phi

georgemoralis:
Review issues seems to be fixed

LNDF (approved):

squidbus (approved):

PR #3008, opened 2025-05-29
Handle V_CVT_F64_U32

mikusp:
Hit by God of War


squidbus:
~~Need to add this to the fp64 to fp32 lowering pass as well.~~

Never mind got mixed up, this doesn’t add any new IR.

PR #3009, opened 2025-05-29
Misc opcodes fixes

mikusp:


src/shader_recompiler/frontend/translate/translate.cpp
```diff
@@ -380,7 +380,7 @@ T Translator::GetSrc64(const InstOperand& operand) {
         break;
     case OperandField::VccLo:
         if constexpr (is_float) {
-            UNREACHABLE();
+            value = ir.PackDouble2x32(ir.CompositeConstruct(ir.GetVccLo(), ir.GetVccHi()));
```
squidbus, line 383: Out of curiosity what case do you have that’s reading a floating point from Vcc registers?

PR #3010, opened 2025-05-29
Handle R128 bit in MIMG instructions

mikusp:
A lot of shaders in God of War use that


squidbus:
Has this been tested and confirmed working? Mainly just concerned whether we handle missing the second half of the image struct correctly.

mikusp:
It doesn't go much further to allow testing if it behaves correctly unfortunately

squidbus:
Guess we’ll find out later

PR #3013, opened 2025-05-30
amdgpu: Implement 0x5u for Depth_MicroTiled.

Randomuser8219:
Used by KNACK (CUSA00068).
Spotted by Niko on the shadPS4 Discord development channel.


squidbus:
Why? As you can see from GetTilingMode this is already handled.

PR #3014, opened 2025-05-30
Provide custom border color to samplers

mikusp:
Found the issue on Discord, affects KNACK ~1.10~

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -716,7 +716,7 @@ void Rasterizer::BindTextures(const Shader::Info& stage, Shader::Backend::Bindin
                 ssharp.max_aniso.Assign(AmdGpu::AnisoRatio::One);
             }
         }
-        const auto vk_sampler = texture_cache.GetSampler(ssharp);
+        const auto vk_sampler = texture_cache.GetSampler(ssharp, liverpool->regs.ta_bc_base);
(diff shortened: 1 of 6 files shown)


src/video_core/texture_cache/sampler.cpp
```diff
@@ -20,7 +21,29 @@ Sampler::Sampler(const Vulkan::Instance& instance, const AmdGpu::Sampler& sample
...
+                .format = vk::Format::eR32G32B32A32Sfloat,
+            };
+            return ret;
+        } else {
+            return std::nullopt;
+        }
+    }();
+
     const vk::SamplerCreateInfo sampler_ci = {
+        .pNext = customColor ? &*customColor : nullptr,
```
squidbus, line 50: Should be conditional on the extension for this being supported. Same for the value of borderColor below if it is set to custom.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -716,7 +716,7 @@ void Rasterizer::BindTextures(const Shader::Info& stage, Shader::Backend::Bindin
                 ssharp.max_aniso.Assign(AmdGpu::AnisoRatio::One);
             }
         }
-        const auto vk_sampler = texture_cache.GetSampler(ssharp);
+        const auto vk_sampler = texture_cache.GetSampler(ssharp, liverpool->regs);
```
squidbus: At least for now I think I would prefer we just pass on the border color part of the regs.

squidbus (approved):

PR #3015, opened 2025-05-30
Handle immediate inline samplers

mikusp:
Previously inline samplers were initialized from a single u32, leaving the other 96 bits zeroed. This PR aims to recover the whole S#. I have no idea what I'm doing in the shader recompiler but it seems to work and allows Dishonored 2 to get past intro.

### DIFF
--- a/src/shader_recompiler/frontend/translate/vector_memory.cpp
+++ b/src/shader_recompiler/frontend/translate/vector_memory.cpp
@@ -531,8 +531,7 @@ IR::Value EmitImageSample(IR::IREmitter& ir, const GcnInst& inst, const IR::Scal
     // Load first dword of T# and S#. We will use them as the handle that will guide resource
     // tracking pass where to read the sharps. This will later also get patched to the SPIRV texture
     // binding index.
-    const IR::Value handle =
-        ir.CompositeConstruct(ir.GetScalarReg(tsharp_reg), ir.GetScalarReg(sampler_reg));
+    const IR::Value handle = ir.GetScalarReg(tsharp_reg);
     const IR::Value inline_sampler =
         ir.CompositeConstruct(ir.GetScalarReg(sampler_reg), ir.GetScalarReg(sampler_reg + 1),
                               ir.GetScalarReg(sampler_reg + 2), ir.GetScalarReg(sampler_reg + 3));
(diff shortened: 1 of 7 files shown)


src/shader_recompiler/frontend/translate/vector_memory.cpp
```diff
@@ -527,6 +527,9 @@ IR::Value EmitImageSample(IR::IREmitter& ir, const GcnInst& inst, const IR::Scal
     // binding index.
     const IR::Value handle =
         ir.CompositeConstruct(ir.GetScalarReg(tsharp_reg), ir.GetScalarReg(sampler_reg));
```
squidbus: There's no need to pass the base sampler reg in twice, if we're passing in a separate arg for sampler regs we should probably change this to just the `tsharp_reg` value.

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -418,29 +418,45 @@ void PatchImageSharp(IR::Block& block, IR::Inst& inst, Info& info, Descriptors&
...
             // Inline sampler resource.
             if (handle.IsImmediate()) {
                 LOG_WARNING(Render_Vulkan, "Inline sampler detected");
-                const auto inline_sampler = AmdGpu::Sampler{.raw0 = handle.U32()};
-                const auto binding = descriptors.Add(SamplerResource{
-                    .sharp_idx = std::numeric_limits<u32>::max(),
-                    .inline_sampler = inline_sampler,
-                });
-                return {binding, inline_sampler};
+                const auto inline_sampler_arg = inst.Arg(5).InstRecursive();
```
squidbus: A lot of complexity in this PR seems to come from doing tracking on the inline sampler, but have you checked to see if all four sampler regs are immediate? Then you can just do exactly as before but with all 128 bits, and remove all the `std::variant` stuff.

mikusp: I've simplified the resource tracking pass, but I think keeping the std::variant gives stronger type safety guarantees vs using u32::max()

mikusp:
There's no std::variant on windows? 😱

squidbus:
There is, we already use it. Are you missing an include?

kalaposfos13:
This PR regresses at least two games, Gravity Rush Remastered and STRANGER OF PARADISE FINAL FANTASY ORIGIN
[CUSA01113.log](https://github.com/user-attachments/files/20847208/CUSA01113.log)
[Discord link](https://discord.com/channels/1080089157554155590/1281178723214688279/1384448868707471411) to STRANGER OF PARADISE FINAL FANTASY ORIGIN regression

diegolix29:
was doing some testing on my ps1 and psp games,  and this pr regressed some of them to the point only the ps1 logo shows and the rest of the games are black, one of them the lost world of jurassic park for ps1 if i rever this pr i see graphics again

on main is just black screen after ps1 logos

image

image

PR #3019, opened 2025-05-31
Core: Revert new rasterizer mapping logic

StevenMiller123:
Based on the recent influx of non-GPU memory asserts in recent regressions, it's likely there's either an issue with my new logic, or an issue somewhere in our code prompting games to try using CPU mappings for GPU code.
Until we're able to resolve this bug, this PR reverts the changes to our rasterizer memory logic made in #2997 

Pinging @gandalfthewhite19890404 and @panzone91 for tests, both have reported this type of regression on the official Discord server.


panzone91:
Tested this branch and I can confirm this seems to revert the regression I've reported.

PR #3021, opened 2025-06-01
Core: Pthread affinity fixups

StevenMiller123:
In the pthread affinity functions, when `thread != g_curthread`, we should be using `ThrState::FindThread` to get and lock the appropriate thread. Adding this logic fixes a potential crash in `scePthreadSetAffinity`, helping Need for Speed™ Heat (CUSA15081) get slightly further.
I've also added an export for `posix_pthread_attr_getschedparam`, as it's used by Kingdom Come: Deliverance Royal Edition (CUSA15436).


StevenMiller123:
Apparently there's a regression?
Swapping to draft while I figure out what I messed up.

StevenMiller123:
Further discussion suggests that it wasn't actually a regression, so I've marked this as ready for review.
Would be nice to have some others testing The Last Guardian, as that's the game I might've regressed apparently.

PR #3025, opened 2025-06-02
Fork detection fixes VII

kalaposfos13:
The sole reason for this PR is so @StevenMiller123 and the ~3 other people who use SSH remotes can have their remote names formatted correctly in the window titles. As this is only a formatting change, it shouldn't affect anything important.

Image of the issue:
image


StevenMiller123:
image
Seems like these changes work on my end.

PR #3031, opened 2025-06-03
Check if region is registered after value is written to a buffer

dima-xd:
By looking at readbacks PR, I found an interesting thing which actually fixes updated Driveclub when it's stuck on playgo


Xcedf:
Yea, that's it, it also allows to Whitch and The Hundred Knight to go ingame(with hacks) and not to hang on loading forever

squidbus:
You can’t just remove the check there. The reason we check if the region is registered is because if it is, we have to consider the GPU copy of the buffer memory. Just doing a memcpy regardless could discard GPU-side data.

PR #3032, opened 2025-06-03
Stub PM4 COPY_DATA opcode

mikusp:
Used by Dragon Age: Inquisition to copy GPU clock somewhere, with the stub goes almost ingame.

### DIFF
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -394,7 +394,7 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
                 break;
             }
             case PM4ItOpcode::SetPredication: {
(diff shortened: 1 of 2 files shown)


src/video_core/amdgpu/liverpool.cpp
```diff
@@ -673,6 +673,16 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
                 }
                 break;
             }
+            case PM4ItOpcode::CopyData: {
+                const auto* copy_data = reinterpret_cast<const PM4CmdCopyData*>(header);
+                LOG_WARNING(Render_Vulkan,
```
squidbus: nit: Use Render log class here as this is not Vulkan specific.

squidbus (approved):
Thanks for fixing up the other log classes as well. Looks good if clang-format is fixed.

PR #3033, opened 2025-06-03
s_flbit_i32_b64

mikusp:


squidbus:
This isn’t going to work as-is on the SPIR-V emit side. From the OpFindUMsb documentation:

```
This instruction is currently limited to 32-bit width components.
```

You will probably need to emit instructions to do two ops, one on each half, and then resolve the final result from those. You can keep this logic in the SPIR-V emit as I do prefer having it be a simple op on the IR side.

PR #2984, opened 2025-05-24
qt: save gui settings to separate file

georgemoralis:
Initial work on saving gui settings to different file than emulator settings . This makes emu saving file less complex and also helps external launcher to save un-needed stuff

Not yet finished place it here for general review

Also

Added --set-addon-folder  for the sdl version
Added catalan language to config

PR #3036, opened 2025-06-04
qt: Save button label setting

rainmakerv3:
Corrects the config functions/calls related to the button label GUI setting (View->show labels under icons) and add the corresponding reads/writes to the config file to allow saving the setting across sessions


georgemoralis:
qt settings are being rewritten to save separately in #2984 , i suggest you make your fixes there

rainmakerv3:
actually everything here would seem to be covered in that PR, closing this instead

PR #3039, opened 2025-06-04
Add option to ignore game patch

Fire-Cube:
Would be nice if someone can try a big game with the new option to be sure it doesnt break something

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -63,8 +63,8 @@ Emulator::~Emulator() {
     Config::saveMainWindow(config_dir / "config.toml");
 }
 
-void Emulator::Run(std::filesystem::path file, const std::vector<std::string> args,
-                   bool ignore_game_patch) {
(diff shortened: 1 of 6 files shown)


src/emulator.cpp
```diff
@@ -63,7 +63,8 @@ Emulator::~Emulator() {
     Config::saveMainWindow(config_dir / "config.toml");
 }
 
-void Emulator::Run(std::filesystem::path file, const std::vector<std::string> args) {
+void Emulator::Run(std::filesystem::path file, const std::vector<std::string> args,
+                   bool ignore_game_patch) {
```
squidbus: Think I would prefer if you set the `ignore_game_patches` var from the main function instead of adding it as another thing to pass through here, similar to how `MemoryPatcher::patchFile` gets set from a command line in at least one of them.

Fire-Cube: done

squidbus (approved):

PR #3041, opened 2025-06-05
libSceZlib: Fix Queues

StevenMiller123:
Queues are a FIFO data structure, so pop() removes the front, not the end. Therefore, we should be getting from the front of a queue before calling pop(), not the back.

Fixing this issue fixes a strange memory issue observed in Rise of the Tomb Raider (CUSA09193), though the game doesn't get much further yet.

### DIFF
--- a/src/core/libraries/zlib/zlib.cpp
+++ b/src/core/libraries/zlib/zlib.cpp
@@ -51,7 +51,7 @@ void ZlibTaskThread(const std::stop_token& stop) {
             if (!task_queue_cv.wait(lock, stop, [&] { return !task_queue.empty(); })) {
                 break;
             }
-            task = task_queue.back();
+            task = task_queue.front();
             task_queue.pop();
         }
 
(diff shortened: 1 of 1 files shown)


StevenMiller123:
I'll need someone to test this with Catherine: Full Body, as apparently that uses libSceZlib too, based on the description of https://github.com/shadps4-emu/shadPS4/pull/2256

Osyotr:
Catherine doesn't really use libSceZlib. The only function it calls is `sceZlibInitialize`.

georgemoralis:
seems valid

StevenMiller123:
@Osyotr For reference, the only libSceZlib function we log currently is sceZlibInitialize, all the others are debug logs and require a log filter to check for.
Though you could technically still be right, I don't have that game to check.

squidbus:
It does use more than initialize for reference.

Osyotr:
It looks like Catherine calls `libSceZlib` functions only when you start a new game.
I did load one of my saves and didn't hit any of `libSceZlib` functions except, of course, `sceZlibInitialize`.

PR #3043, opened 2025-06-05
Select

georgemoralis:
Currently only simulated select for windows

Select is used for sockets,pipes and file operation.

Issue with linux is that ndfs is used for sockets,pipes and files and we have separate maps for these , so it can't be done so easy

A windows win :D


georgemoralis:
Thread discussion [https://shadps4.net/forums/showthread.php?tid=36](https://shadps4.net/forums/showthread.php?tid=36)
or post here

mailwl:
when this PR will be merged?

georgemoralis:
redone in #3353

PR #3047, opened 2025-06-06
PM4 type 2 in acb

mailwl:
this PR passes **CUSA00625 - Battlefield™ Hardline** through main menu

### DIFF
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -229,12 +229,12 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
 
         switch (type) {
         default:
-            UNREACHABLE_MSG("Unsupported PM4 type {}", type);
+            UNREACHABLE_MSG("Wrong PM4 type {}", type);
             break;
         case 0:
-            LOG_ERROR(Render, "Unimplemented PM4 type 0, base reg: {}, count: {}",
-                      header->type0.base.Value(), header->type0.NumWords());
-            // TODO fill registers with given values
+            UNREACHABLE_MSG("Unimplemented PM4 type 0, base reg: {}, size: {}",
+                            header->type0.base.Value(), header->type0.NumWords());
+            // comment "UNREACHABLE_MSG()" above ^^^ to skip "PM4 type 0" command
             dcb = NextPacket(dcb, header->type0.NumWords() + 1);
             continue;
         case 2:
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -228,10 +228,14 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
         const u32 type = header->type;
 
         switch (type) {
-        case 0:
-        case 1:
+        default:
             UNREACHABLE_MSG("Unsupported PM4 type {}", type);
             break;
+        case 0:
+            LOG_ERROR(Render, "Unimplemented PM4 type 2, base reg: {}, count: {}",
+                      header->type0.base.Value(), header->type0.NumWords());
+            dcb = NextPacket(dcb, header->type3.NumWords() + 1);
+            continue;
         case 2:
             // Type-2 packet are used for padding purposes
             dcb = NextPacket(dcb, 1);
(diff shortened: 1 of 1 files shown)


StevenMiller123:
This PR progresses updated versions of Call of Duty®: Advanced Warfare (CUSA00803) further, though they still crash before reaching ingame.
[CUSA00803 v1.24 PR.zip](https://github.com/user-attachments/files/20631268/CUSA00803.v1.24.PR.zip)
[CUSA00803 v1.24 Main.txt](https://github.com/user-attachments/files/20631269/CUSA00803.v1.24.Main.txt)

Randomuser8219:
[shad_log.txt](https://github.com/user-attachments/files/20631430/shad_log.txt)
Knack now gives another garbage PM4 without debug filter.

squidbus:
Last I remember, seeing errors for PM4 like this are usually other issues with processing the command buffer and not actually unimplemented packets.

StevenMiller123:
Type 0 case definitely feels like a hack, though I'm pretty sure type 2 is valid.
Call of Duty®: Advanced Warfare hits the type 2 error, and I think if it were a bad packet, then implementing type 2 properly wouldn't fix it.

mailwl:
Wrong command size, return as been in first commit, now error:
```
[Debug] <Critical> vk_presenter.cpp:646 operator(): Assertion Failed!
```
in knack, and all screen blink black. Something really broken.

Battlefield still softlock on game start

mailwl:
@StevenMiller123, do you think, something wrong with offsets, and PM4t2 - is part (data) of other command?

Randomuser8219:
> Wrong command size, return as been in first commit, now error:
> 
> ```
> [Debug] <Critical> vk_presenter.cpp:646 operator(): Assertion Failed!
> Device lost during waiting for a frame
> ```
> 
> in knack, and all screen blink black. Something really broken.
> 
> Battlefield still softlock on game start

Knack has a pretty bad race condition going on still. This PR won't fix it.

StevenMiller123:
There are a lot of reasons why games might be triggering the PM4 type 0 unreachable, and it will need to be looked into at some point. It's pretty clear in most games that the PM4 type 0 error is caused by some race condition. No amount of properly handling type 0 packets will fix these crashes in games like Knack or The Last of Us Remastered.
The type 2 packet handling is fine, if there was an issue there then Call of Duty wouldn't work on my end.

georgemoralis:
ok maybe we can keep type 2 for this pr?

mailwl:
i leave the possibility to skip Type0, for developers

src/video_core/amdgpu/liverpool.cpp
```diff
@@ -228,10 +228,15 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
...
-        case 1:
-            UNREACHABLE_MSG("Unsupported PM4 type {}", type);
+        default:
+            UNREACHABLE_MSG("Wrong PM4 type {}", type);
             break;
+        case 0:
+            UNREACHABLE_MSG("Unimplemented PM4 type 0, base reg: {}, size: {}",
+                            header->type0.base.Value(), header->type0.NumWords());
+            // comment "UNREACHABLE_MSG()" above ^^^ to skip "PM4 type 0" command
+            dcb = NextPacket(dcb, header->type0.NumWords() + 1);
```
squidbus: I don't think we should have this kind of thing committed into the main branch source

mailwl: why? this code not really executed without editing code
and source code full of stubs, which cannot be reversed from fw ATM

squidbus: Because it’s clutter, and this isn’t even a stub, it’s an invalid packet. This is just hiding another issue. Hacks don’t belong in mainline.

mailwl: this code is NOT executing in mainline at all, it crash early on UNREACHABLE_MSG()
it's just for devs, to see, how long a game goes on w/o this crash.
After fixing reasons, why games go to this case, the code probably never reach this point (but, as it depricated, may still be used)

squidbus: Yes, and I think even commented out we should not be committing incorrect hacks into mainline. There’s nothing stopping someone from making this change for debugging or having a branch but I won’t approve on this as-is.

PR #3049, opened 2025-06-07
Patch movntss and movntsd

OFFTKP:


src/core/cpu_patches.cpp
```diff
@@ -383,8 +386,60 @@ static void GenerateINSERTQ(const ZydisDecodedOperand* operands, Xbyak::CodeGene
...
+    }
+
+    // Some sanity checks
+    ASSERT(found);
+    ASSERT(index >= 2);
+    ASSERT(ptr[index - 2] == 0xF2);
+    ASSERT(ptr[index - 1] == 0x0F);
+
+    // This turns the MOVNTSD to a MOVSD m64, xmm
+    ptr[index] = 0x11;
```
squidbus: This seems basically identical to the other function except a slight change to one assert, can we de-dupe these?

OFFTKP: True, done

PR #3051, opened 2025-06-07
Dc swizzle

AboMedoz:
This PR fixes Swizzle Regression that was introduced from #2226 and a Random Error that happens that make the screen full black in the game, thx for Turtle for helping me debug this

### DIFF
--- a/src/shader_recompiler/frontend/translate/export.cpp
+++ b/src/shader_recompiler/frontend/translate/export.cpp
@@ -33,8 +33,6 @@ void Translator::ExportMrtCompressed(IR::Attribute attribute, u32 idx, const IR:
     AmdGpu::NumberFormat num_format;
     switch (color_buffer.export_format) {
     case AmdGpu::Liverpool::ShaderExportFormat::Zero:
-        // No export
-        return;
     case AmdGpu::Liverpool::ShaderExportFormat::ABGR_FP16:
         num_format = AmdGpu::NumberFormat::Float;
         break;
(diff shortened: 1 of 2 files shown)


src/shader_recompiler/frontend/translate/export.cpp
```diff
@@ -33,8 +33,6 @@ void Translator::ExportMrtCompressed(IR::Attribute attribute, u32 idx, const IR:
     AmdGpu::NumberFormat num_format;
     switch (color_buffer.export_format) {
     case AmdGpu::Liverpool::ShaderExportFormat::Zero:
-        // No export
-        return;
```
squidbus: AMD docs:
```
00 - SPI_SHADER_ZERO: No exports done
```

Can you explain how you arrived at this change?

AboMedoz: It does fix the game, the hypothesis is zero format maybe doesn't mean no write, but it writes zero instead that's confirmed from the game setting zero swizzle

squidbus: I'd like some clearer confirmation on this as:
* that seems to contradict what the setting is supposed to do according to the docs
* that's not really what you're doing here either, you're just treating it as a normal `ABGR_FP16` export 
* it could easily be the case that the current code is correct and this is working around some other issue elsewhere

squidbus: Also looking at Mesa, `SPI_SHADER_ZERO` is used for attachments that should not be written to at all.

raphaelthegreat: Hi, so after a little bit of looking turns out the game is using dual source blending for a draw, with src1 color as dest blend factor. So when mrt1 export is skipped, this factor is zero and the previous contents are wiped from framebuffer (it looks like mrt1 can be used as index = 1 of the first attachment instead of another mrt)

AboMedoz: okay i reverted the first commit, the second one is still valid right and it solves an issue

PR #3055, opened 2025-06-08
shader_recompiler: Implement linear interpolation support

raphaelthegreat:
This is used by Driveclub (CUSA00093) for some light passes. Fixes the warp effect in the edges of the screen during nighttime races and a "haze" effect inside the car

Note: Technically this isn't fully accurate as it doesn't handle Sample and Centroid. This can be remedied by using VK_AMD_shader_explicit_vertex_parameter but that would add complexity that isnt needed here so it wasn't done


squidbus (approved):

Missake212:
This PR regressed both ハコニワカンパニワークス (CUSA08248) and Tekken 7 (CUSA06014). Logs don't tell much.

[CUSA08248.log](https://github.com/user-attachments/files/20655394/CUSA08248.log)
[CUSA06014.log](https://github.com/user-attachments/files/20655397/CUSA06014.log)

Tekken 7 call stack:
image

ハコニワカンパニワークス call stack:
image

PR #3056, opened 2025-06-08
texture_cache: Handle compressed views of uncompressed images

raphaelthegreat:
Fixes the missing car colors in Driveclub (CUSA00093). The game renders the car paint textures as R32G32B32A32 and then uses a compute shader to write to another texture of the same format. It turns out this is a BC3 compression shader, as the game will later request the that storage image as BC3. The texture cache considered this a bad match and threw out the generated one, in favor of an empty black image.

This is fixed by relaxing the mismatch check to use dimention in blocks instead of texels and checking bits per block as a rough estimate of compatibility. Then an image copy is issued in the uncompressed -> compressed conversion. On AMD GPUs having a compressed view of an uncompressed image works just fine, but on NVIDIA it doesn't, so the copy is required.

The vulkan compatibility rule table has also been expanded to handle size compatibility, when VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT is enabled which is used now. This ensures that if the image is requested in uncompressed format again, the cache can use an image view on the compressed image instead of creating a new image again.


kalaposfos13:
You forgot to push CMakeLists.txt

kalaposfos13:
If you revert a change in #2997 that broke P.T, it shows graphics with this now
image
image
image
image

kalaposfos13:
@raphaelthegreat Can you rebase this? With #3059 merged, that would mean that this PR will be able to display graphics on P.T. without modification to the code

kalaposfos13:
Also, here's a video of textures starting off loaded, but then slowly deloading over time: https://youtu.be/zVU-NSEpEKM

src/video_core/texture_cache/image_info.cpp
```diff
@@ -152,6 +152,80 @@ ImageInfo::ImageInfo(const AmdGpu::Image& image, const Shader::ImageResource& de
...
+}
+
+bool ImageInfo::IsTilingCompatible(u32 lhs, u32 rhs) const {
+    if (lhs == rhs) {
+        return true;
+    }
+    if (lhs == 0x0e && rhs == 0x0d) {
+        return true;
+    }
+    if (lhs == 0x0d && rhs == 0x0e) {
```
squidbus, line 223: Can we define some constants for these or at least some comments to indicate what they are?

raphaelthegreat: I didn't write this and don't know what the constants mean sadly

squidbus: Ah I see this was just moved. Well from `TilingMode`:
```
    Texture_MicroTiled = 0xDu,
    Texture_MacroTiled = 0xEu,
```
So I assume we can just use these existing definitions.

raphaelthegreat: Can this be done on another PR? So this one isn't cluttered more

squidbus: Sure

PR #3059, opened 2025-06-08
Core: Do not coalesce dmem pages

StevenMiller123:
Looks like this change is what broke P.T. (CUSA01127).
Until I'm able to debug the issue, this reverts exactly one change from #2997 

Pinging @kalaposfos13 to verify.


kalaposfos13:
Not going to bother checking this PR out as locally I did the exact same thing except I commented the line out instead, but I can verify it fixes the issue

PR #3062, opened 2025-06-08
Hack ReadLane to prevent device loss

mikusp:
Allows Death Stranding 1.13 to compile a shader previously resulting in a device loss. Other games using a `v_readlane` instruction might benefit from it too

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
@@ -498,7 +498,7 @@ Id EmitLaneId(EmitContext& ctx);
 Id EmitWarpId(EmitContext& ctx);
 Id EmitQuadShuffle(EmitContext& ctx, Id value, Id index);
 Id EmitReadFirstLane(EmitContext& ctx, Id value);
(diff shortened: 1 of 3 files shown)


squidbus:
I understand this results in a device loss for that game but it seems to me that ignoring the lane completely would be worse overall.

raphaelthegreat:
There is an entire IR pass dedicated to eliminating ReadLane instructions. You first need to check if the ReadLane is compiler generated or handwritten. If its the former then the pass needs to be debugged to see why its failing, otherwise it needs to be emulated on NVIDIA with shared memory

mikusp:
I've seen that pass, but it seems it checks whether readlane can be eliminated. Is it expected to eliminate _all_ readlane instances?

raphaelthegreat:
> I've seen that pass, but it seems it checks whether readlane can be eliminated. Is it expected to eliminate _all_ readlane instances?

Its goal is to eliminate all _compiler generated_ readlanes

mikusp:
I see. This one is in the original shader code

squidbus:
Compiler generated as in by the AMD shader compiler, not our shader recompiler.

PR #3064, opened 2025-06-08
shader_recompiler: Better handling of geometry shader scenario G

raphaelthegreat:
The sea islands register reference mentions 2 different behaviors for VGT_GS_MAX_VERT_OUT. For Scenario C it represents the actual amount of output vertices a GS will make. For Scenario G on the other hand it represents the maximum, so it can't always be used for mapping offset to attribute, as the actual component stride might be smaller.

This fixes a crash when compiling a geometry shader in Driveclub (CUSA00093) which "lies" about the number of vertices it outputs. While the actual number is 4, the register reports a much larger number (44). Instead attempt to compute the output vertices from the copy shader and if they mismatch do some validation checks and use that instead. I've also added a warning in case this regresses any other geoshaders, so it can be seen easily

~~Oops wrong branch~~

### DIFF
--- a/src/shader_recompiler/runtime_info.h
+++ b/src/shader_recompiler/runtime_info.h
@@ -149,6 +149,7 @@ struct GeometryRuntimeInfo {
     u32 out_vertex_data_size{};
     AmdGpu::PrimitiveType in_primitive;
     GsOutputPrimTypes out_primitive;
+    AmdGpu::Liverpool::GsMode::Mode mode;
     std::span<const u32> vs_copy;
     u64 vs_copy_hash;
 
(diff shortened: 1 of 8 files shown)


src/shader_recompiler/runtime_info.h
```diff
@@ -149,6 +149,7 @@ struct GeometryRuntimeInfo {
     u32 out_vertex_data_size{};
     AmdGpu::PrimitiveType in_primitive;
     GsOutputPrimTypes out_primitive;
+    AmdGpu::Liverpool::GsMode::Mode mode;
```
squidbus, line 152: Does the `operator==` need updating here?

raphaelthegreat: No it was left out intentionally because it isn't used in compilation but only for validation purposes

squidbus (approved):

raphaelthegreat:
After this is merged I have one final PR that will allow DC to boot without any hacks

PR #3066, opened 2025-06-09
texture_cache: Basic handling of partially resident images

raphaelthegreat:
This allows Driveclub (CUSA00093) to go in-game without any hacks.

The game setups a PRT area, maps only some pages inside it and creates a render target covering all of it causing a gpu tracking assert to kick in. To avoid this the memory manager will register active PRT areas and map those areas to gpu, so the page manager can track any image inside the region. Meanwhile access to cpu memory will be handled via a new function CopySparseMemory which behaves the same as a memcpy, except when source address is a PRT area, where it will copy only mapped regions, while zeroing unmapped areas.

Using host sparse feature would be ideal, but its very inconvenient as host sparse page size and alignment might differ a lot from the guest. So its a lot easier to eat the extra memory cost of creating the non sparse version of the full image. This can also be expanded in the future with subresource uploads, on every gpu map we can invalidate the texture cache and reupload parts of PRT that were mapped. This isn't done here as that would trash the entire thing (due to unimplemented subres upload) and DC uses it as a render target

### DIFF
--- a/src/video_core/texture_cache/tile_manager.cpp
+++ b/src/video_core/texture_cache/tile_manager.cpp
@@ -269,7 +269,10 @@ std::pair<vk::Buffer, u32> TileManager::TryDetile(vk::Buffer in_buffer, u32 in_o
     params.height = info.size.height;
     if (info.tiling_mode == AmdGpu::TilingMode::Texture_Volume ||
         info.tiling_mode == AmdGpu::TilingMode::Display_MicroTiled) {
-        ASSERT(info.resources.levels == 1);
+        if (info.resources.levels != 1) {
+            LOG_ERROR(Render_Vulkan, "Unexpected mipmaps for volume and display tilings {}",
+                      info.resources.levels);
+        }
         const auto tiles_per_row = info.pitch / 8u;
         const auto tiles_per_slice = tiles_per_row * ((info.size.height + 7u) / 8u);
         params.sizes[0] = tiles_per_row;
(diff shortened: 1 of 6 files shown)


StevenMiller123:
This PR, more specifically the choice to downgrade that tile_manager assert to an error, brings Final Fantasy XV to menus on Linux devices.
image

[shad_log.txt](https://github.com/user-attachments/files/20647805/shad_log.txt)

src/video_core/texture_cache/tile_manager.cpp
```diff
@@ -269,7 +269,10 @@ std::pair<vk::Buffer, u32> TileManager::TryDetile(vk::Buffer in_buffer, u32 in_o
     params.height = info.size.height;
     if (info.tiling_mode == AmdGpu::TilingMode::Texture_Volume ||
         info.tiling_mode == AmdGpu::TilingMode::Display_MicroTiled) {
-        ASSERT(info.resources.levels == 1);
+        if (info.resources.levels != 1) {
+            LOG_ERROR(Render_Vulkan, "Unexpected mipmaps for volume and display tilings {}",
+                      info.resources.levels);
```
squidbus, line 274: Shouldn't we actually implement passing through all the sizes instead of just lowering this?

raphaelthegreat: I don't have much experience on this code, otherwise I would implement it.
It was the last assertion preventing the game from booting and I didn't think 1 texture was worth crashing the entire application

squidbus: Maybe we could at least separate it to a different PR? It seems to not be directly related to PRTs.

raphaelthegreat: The change was very small so I didnt consider it worthy of separate PR, but yeah it's reasonable to separate it

PR #3068, opened 2025-06-09
better handling of tiling size, dividing check and assert more specifc

diegolix29:
this pr helps handle all tiling sizes under texture, divide the check so display due to shader limitations dosent get overloaded. 
added psucien suggestions,


diegolix29:
@psucien can u check it? may still need correctness altho for linux worked really nice for me.

PR #3070, opened 2025-06-09
video_core: Page manager/region manager optimization

LNDF:
Rewrite parts of the page manager and region manager to be faster

Use a bit array that uses AVX to accelerate.
Also tries to batch mprotects in the same region (another pr is needed to take advantage of this)

I have done zero testing right now with this and i rewrote parts of the memory tracker so there is a very high chance I broke something.

This is part of a series of PR with the hope of improving performance of games that use DMA, since it is currently quite slow (mainly because the ammount of memory protections and synced buffers).

### DIFF
--- a/src/video_core/buffer_cache/region_manager.h
+++ b/src/video_core/buffer_cache/region_manager.h
@@ -188,7 +188,7 @@ private:
         if (mask.None()) {
             return; // No changes to the CPU tracking state
         }
-        
+
         writeable = cpu;
         tracker->UpdatePageWatchersForRegion<add_to_tracker>(cpu_addr, mask);
     }
--- a/src/video_core/buffer_cache/region_manager.h
+++ b/src/video_core/buffer_cache/region_manager.h
@@ -184,8 +184,13 @@ private:
     void UpdateProtection() {
         RENDERER_TRACE;
         RegionBits mask = cpu ^ writeable;
+
+        if (mask.None()) {
+            return; // No changes to the CPU tracking state
+        }
+        
         writeable = cpu;
-        tracker->UpdatePageWatchersMasked<add_to_tracker>(cpu_addr, mask);
+        tracker->UpdatePageWatchersForRegion<add_to_tracker>(cpu_addr, mask);
     }
 
 #ifdef PTHREAD_ADAPTIVE_MUTEX_INITIALIZER_NP
--- a/src/video_core/buffer_cache/region_manager.h
+++ b/src/video_core/buffer_cache/region_manager.h
@@ -29,7 +29,7 @@ public:
         : tracker{tracker_}, cpu_addr{cpu_addr_} {
         cpu.Fill();
         gpu.Clear();
-        untracked.Fill();
+        writeable.Fill();
     }
     explicit RegionManager() = default;
 
@@ -47,13 +47,13 @@ public:
 
     template <Type type>
     RegionBits& GetRegionBits() noexcept {
-        static_assert(type != Type::Untracked);
+        static_assert(type != Type::Writeable);
         if constexpr (type == Type::CPU) {
             return cpu;
         } else if constexpr (type == Type::GPU) {
             return gpu;
-        } else if constexpr (type == Type::Untracked) {
-            return untracked;
+        } else if constexpr (type == Type::Writeable) {
+            return writeable;
         } else {
             static_assert(false, "Invalid type");
         }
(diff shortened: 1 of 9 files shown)


LNDF:
@raphaelthegreat pinging you since you were the one who suggested to optimize the memory tracker in the fns pr.

Keep in mind that there are still at least a regression that I need to find the cause of.

src/video_core/buffer_cache/region_manager.h
```diff
@@ -0,0 +1,207 @@
...
+#ifdef PTHREAD_ADAPTIVE_MUTEX_INITIALIZER_NP
+    Common::AdaptiveMutex lock;
+#else
+    Common::SpinLock lock;
+#endif
+    PageManager* tracker;
+    VAddr cpu_addr = 0;
+    RegionBits cpu;
+    RegionBits gpu;
+    RegionBits untracked;
```
raphaelthegreat: I recommend renaming untracked to write like in readbacks commit. It makes it a lot more clear what it does (set bits mean that page is writable, unset means write protected)

src/video_core/buffer_cache/region_manager.h
```diff
@@ -0,0 +1,207 @@
...
+        std::unique_lock lk{lock};
+        static_assert(type != Type::Untracked);
+
+        RegionBits& bits = GetRegionBits<type>();
+        RegionBits mask{};
+        mask.SetRange(start_page, end_page);
+        if constexpr (type == Type::GPU) {
+            mask &= ~untracked;
+        }
+        mask &= bits;
```
raphaelthegreat: The if condition is actually not necessary and I had removed it in readbacks commit. It essentially keeps only the pages that are write protected, but if you mark a region as gpu modified all pages are write protected so shouldn't matter (assuming you handle gpu modified ranges properly, i.e unmarking them on write faults which isn't actually done on main, due to lack of readbacks)

Aside from that, you could possibly optimize this by fusing the SetRange and &= bits operations into a single one (could add a constructor to BitArray that receives another BitArray object and a range of pages to copy)

src/video_core/buffer_cache/region_manager.h
```diff
@@ -0,0 +1,207 @@
...
+     *
+     * @tparam add_to_tracker True when the tracker should start tracking the new pages
+     */
+    template <bool add_to_tracker, typename Lock>
+    void UpdateProtection(Lock&& lk) {
+        RENDERER_TRACE;
+        RegionBits mask = cpu ^ untracked;
+        untracked = cpu;
+        lk.unlock();
+        tracker->UpdatePageWatchersMasked<add_to_tracker>(cpu_addr, mask);
```
raphaelthegreat: Does this make a difference between iterating the pages here and using normal update function? Assuming non contiguous mask, the gaps aren't gonna be affected so you will still do N = num contiguous regions, mprotect calls

LNDF: Yes.
For example, in the following situation:
I want to untrack from page 0 to 15. But, since page 1 to 14 are already untracked, I end up only needing to untrack page 0 and 15.
With the previous implementation, UpdatePageWatchers would be called twoiceonce for page 0 and another time for page 15. 2 mprotects.
With the implementation on this PR if we want to untrack (unprotect) from page 0 to 15 and every page inbetween is not write protected, it will be handled as 1 single mprotect from page 0 to 15. You can see how that is handled in the UpdatePageWatchersMasked function.

raphaelthegreat: I see that makes sense

LNDF:
Currently made page manager protects locked behind the mutex. Apparently, this was causing some issues on some games (infamous for example). I think they were race conditions on virtualprotect acting on the same range of memory..

Missake212:
No regressions on my end.

raphaelthegreat:
Has performance been compared to the previous implementation before merging? Those unaligned AVX2 loads might not help performance as much.

Missake212:
I personally didn't notice any difference in performance but all my games were already hitting the frame rate limit so I'm probably not the best person to ask about this.

LNDF:
> Has performance been compared to the previous implementation before merging? Those unaligned AVX2 loads might not help performance as much.

I can make the bitarray 32 bytes aligned. 
But I think the main improvement of this PR is the hability to batchh protects with the Masked version of the updateWatchers function. I want to make a PR that batches buffer syncing. For what I've seen, games that use DMA and need to sync all buffers end up syncing a lot of small contiguous buffers. If we can batch sync (and batch untrack) the buffers, we could reduce the number of protects (the main source of slowness).

raphaelthegreat:
The size of the bit array doesn't really matter, since you load the AVX2 register from the start word of the range. I think it may be okay but it needs to be tested. Also it might be a good idea to increase the region manager size from 4MB to 16 or 32MB (so from 16 qwords to 64 or 128) that will allow more mprotects to be batched (as mprotects are split in region boundaries) and take better advantage of AVX2 (each region will fit 16 or 32 AVX2 registers instead of only 4). The disadvantage is that locking will cover more area and might stall more writes but ig that can be solved if encountered

PR #3072, opened 2025-06-09
Automatically Update shadPS4 on Windows Startup

davidantunes23:
[Feature Request]: Automatically Update shadPS4 on OS Startup shadps4-emu/shadps4-qtlauncher#45
Created a separate .exe for updating. Added a checkbox in the settings to enable the option to update on Windows startup. When the option is enabled, a PowerShell script adds the new updater exe to the Windows Task Scheduler and sets it to run at startup. If there are updates available at startup, the update windows pops up. Disabling the option runs another script to remove the updater from the Task Scheduler.
Currently the feature only works for Windows but can be extended to support the other operative systems.


squidbus:
This looks to be set up incredibly incorrectly and I question the utility in adding a whole new updater executable to update the emulator on every boot instead of just updating when you open the emulator.

gandalfthewhite19890404:
Why not set shadps4 as startup application and disabling all other Windows functions except shadps4's?
Shortly - I'm against that PR (and whole its idea) completely.

georgemoralis:
yah that doesn't seem like something that we can add , there are plenty of reasons someone will not want to update to latest version , or maybe he has multiple versions of the emulator and that can confuse things . Thanks for the efford anyway :)

PR #3073, opened 2025-06-10
mac: Fix building on macOS 26.

squidbus:
* Fix linker flags to work with Xcode 26.
* Update MoltenVK with a compile fix for Xcode 26.
* Fix missing `pthread.h` header include.
* Fix possible CMake dependency loop on non-Qt macOS build due to bundled MoltenVK directory create.
* Remove duplicate SDL3 link.


squidbus:
Note: Qt build is still failing for me with Xcode 26 due to some issue with Qt and the OpenGL frameworks it links in, but SDL builds successfully.

PR #3075, opened 2025-06-10
shader_recompiler: Various fixes to shared memory and atomics.

squidbus:
Fixes a number of issues from https://github.com/shadps4-emu/shadPS4/pull/3007 and other changes:
* Add a bunch of missing instructions to shared instruction types condition, side-effects condition, etc.
* Add missing shared instruction replacements to shared memory to storage pass.
* Fix incorrect addressing when lowering shared memory to buffer.
* Fix load/write shared U16/U64 types to be consistent with buffer types.
* Implement `BufferAtomicInc` and `BufferAtomicDec`.
* Fix `BufferAtomicIAdd` not emitting 64-bit variant when needed.


squidbus:
Changed course on the instruction types, now the shared and buffer load/stores use the proper U8/U16/U32/U64/F32 type consistently. Also unified the bounds check stuff more.

raphaelthegreat:
Not using shared memory at all when VK_KHR_workgroup_memory_explicit_layout is not available might significantly slow down GPU performance on macos, maybe it should be limited if shader actually uses U16 shared memory types and use the old shared memory creation code for u32

squidbus:
> Not using shared memory at all when VK_KHR_workgroup_memory_explicit_layout is not available might significantly slow down GPU performance on macos, maybe it should be limited if shader actually uses U16 shared memory types and use the old shared memory creation code for u32

That's not something introduced by this PR, but it is something I'm working on.

squidbus:
Would appreciate regression testing if anyone is available to, by the way.

Missake212:
No regressions on my end.

PR #3076, opened 2025-06-10
shader_recompiler: Cleanup fragment attribute handling

raphaelthegreat:
Main purpose is to fix regression reported in https://github.com/shadps4-emu/shadPS4/pull/3055#issuecomment-2955788516 but I also placed some minor fixes I found in the process as well

The attribute indexing code was using the attribute semantic (i.e which vertex shader paramN output it uses) for indexing into arrays while the added interpolation code was using the fragment attribute index. This was confusing so I changed the code to always just index using the attribute number and only use the semantic for the location specifier in SPIRV


Missake212:
That fixed both games, thank you for the quick PR.

Screenshot 2025-06-10 154503

PR #3077, opened 2025-06-10
libSceVideodec2: Update structs to match newer firmwares

### LINKED ISSUE #2604: [APP BUG]: ORBIS_VIDEODEC2_ERROR_STRUCT_SIZE does not give out a proper error

StevenMiller123:
Based on some decompilation I did, some libSceVideodec2 structs were adjusted somewhere around firmware 6.50. Our previous code was built around older versions of the library, and as a result, newer versions of libSceAvPlayer.sprx and newer games that use libSceVideodec2 would hit the struct size error return.
This PR updates the structs, and updates the struct size checks to match how newer versions of libSceVideodec2 behave.

This should properly fix #2604, though the author of that issue probably can't test this. 

I've also thrown in a function export needed for Assassin's Creed Unity, since I don't see a point in making a separate one line PR for that.

### DIFF
--- a/src/core/libraries/videodec/videodec2.h
+++ b/src/core/libraries/videodec/videodec2.h
@@ -73,8 +73,10 @@ struct OrbisVideodec2OutputInfo {
     u32 frameHeight;
     void* frameBuffer;
     u64 frameBufferSize;
+    u32 frameFormat;
+    u32 framePitchInBytes;
 };
-static_assert(sizeof(OrbisVideodec2OutputInfo) == 0x30);
+static_assert(sizeof(OrbisVideodec2OutputInfo) == 0x38);
 
 struct OrbisVideodec2FrameBuffer {
     u64 thisSize;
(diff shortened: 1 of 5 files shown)


src/core/libraries/videodec/videodec2.h
```diff
@@ -73,8 +73,10 @@ struct OrbisVideodec2OutputInfo {
     u32 frameHeight;
     void* frameBuffer;
     u64 frameBufferSize;
+    u32 frameFormat;
+    u32 framePitchInBytes;
```
squidbus, line 77: Shouldn’t we also populate these with something if they’re included in the struct?

StevenMiller123: Should be fixed now, though after looking into this, I'm not sure we're setting `framePitch` properly.
I'll run some hardware tests to verify how this works once I get a chance.

PR #3079, opened 2025-06-10
 texture_cache: Implement color<->depth copies

raphaelthegreat:
This fixes the skybox clouds covering everything in the screen in Driveclub (CUSA00093).

The game uses a compute shader with the depth buffer as input and generates a R32 image based on it. Later when rendering the clouds texture it binds that image as a D32 target with writes disabled to control which ones are rendered. The code for handling depth and color overlaps already existed, but the actual copy was unimplemented so it just discarded the old color image and resulted in empty depth buffer, allowing all clouds to render and overlap the screen. The copy is implemented here by using an intermediate buffer, as vulkan doesn't allow image copies with different aspects.

The compute shader that generated the R32 texture also had an issue with divergence on NVIDIA which I also fixed here while at it. The main shader is covered by an if condition, so all blocks had branch_depth = 1, which prevented barrier insertion. To address this, the branch_depth variable is replaced with a divergence_depth one which is the same but only increments when the branch has a divergent condition.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -8,7 +8,6 @@
 #include "common/debug.h"
 #include "video_core/buffer_cache/buffer_cache.h"
 #include "video_core/page_manager.h"
-#include "video_core/renderer_vulkan/liverpool_to_vk.h"
 #include "video_core/renderer_vulkan/vk_instance.h"
 #include "video_core/renderer_vulkan/vk_scheduler.h"
 #include "video_core/texture_cache/host_compatibility.h"
@@ -126,7 +125,7 @@ void TextureCache::UnmapMemory(VAddr cpu_addr, size_t size) {
 
 ImageId TextureCache::ResolveDepthOverlap(const ImageInfo& requested_info, BindingType binding,
                                           ImageId cache_image_id) {
-    const auto& cache_image = slot_images[cache_image_id];
+    auto& cache_image = slot_images[cache_image_id];
 
     if (!cache_image.info.IsDepthStencil() && !requested_info.IsDepthStencil()) {
         return {};
@@ -169,18 +168,21 @@ ImageId TextureCache::ResolveDepthOverlap(const ImageInfo& requested_info, Bindi
     }
 
     if (recreate) {
-        auto new_info{requested_info};
-        new_info.resources = std::max(requested_info.resources, cache_image.info.resources);
-        new_info.UpdateSize();
+        auto new_info = requested_info;
+        new_info.resources = std::min(requested_info.resources, cache_image.info.resources);
         const auto new_image_id = slot_images.insert(instance, scheduler, new_info);
         RegisterImage(new_image_id);
 
         // Inherit image usage
-        auto& new_image = GetImage(new_image_id);
+        auto& new_image = slot_images[new_image_id];
         new_image.usage = cache_image.usage;
+        new_image.flags &= ~ImageFlagBits::Dirty;
 
-        // TODO: perform a depth copy here
+        // Perform depth<->color copy using the intermediate copy buffer.
+        const auto& copy_buffer = buffer_cache.GetUtilityBuffer(MemoryUsage::DeviceLocal);
+        new_image.CopyImageWithBuffer(cache_image, copy_buffer.Handle(), 0);
 
+        // Free the cache image.
         FreeImage(cache_image_id);
         return new_image_id;
     }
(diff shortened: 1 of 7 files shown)


src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -580,16 +582,14 @@ void TextureCache::RefreshImage(Image& image, Vulkan::Scheduler* custom_schedule
     }
 
     auto* sched_ptr = custom_scheduler ? custom_scheduler : &scheduler;
+    const auto cmdbuf = sched_ptr->CommandBuffer();
```
squidbus: If the command buffer is moved up here, it can become invalidated by a flush in `ObtainBufferForImage` before being used.

squidbus (approved):
It’s worth noting that with VK_KHR_maintenance8, copying between color and depth/stencil images is now possible. Could be a future optimization to use that when available.

ElBread3:
Since this PR was merged Kyoei Toshi and World of Final Fantasy have regressed to encountering device lost during their loading screens.
Reposted from Discord, from Kyoei Toshi:
```[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-vkCmdCopyImageToBuffer-pRegions-00183: Validation Error: [ VUID-vkCmdCopyImageToBuffer-pRegions-00183 ] Object 0: handle = 0x24276150, name = CommandPool: Command Buffer 3, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xba7514000000002a, name = StreamBuffer(DeviceLocal):0x1000000, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x710595bf | vkCmdCopyImageToBuffer(): pRegions[0] is trying to copy 67108864 bytes plus 0 offset to/from the VkBuffer (VkBuffer 0xba7514000000002a[StreamBuffer(DeviceLocal):0x1000000]) which exceeds the VkBuffer total size of 16777216 bytes. The Vulkan spec states: dstBuffer must be large enough to contain all buffer locations that are accessed according to Buffer and Image Addressing, for each element of pRegions (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdCopyImageToBuffer-pRegions-00183)```
```[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-vkCmdCopyBufferToImage-pRegions-00171: Validation Error: [ VUID-vkCmdCopyBufferToImage-pRegions-00171 ] Object 0: handle = 0x24276150, name = CommandPool: Command Buffer 3, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xba7514000000002a, name = StreamBuffer(DeviceLocal):0x1000000, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0x6f4d3c00 | vkCmdCopyBufferToImage(): pRegions[0] is trying to copy 67108864 bytes plus 0 offset to/from the VkBuffer (VkBuffer 0xba7514000000002a[StreamBuffer(DeviceLocal):0x1000000]) which exceeds the VkBuffer total size of 16777216 bytes. The Vulkan spec states: srcBuffer must be large enough to contain all buffer locations that are accessed according to Buffer and Image Addressing, for each element of pRegions (https://vulkan.lunarg.com/doc/view/1.3.296.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdCopyBufferToImage-pRegions-00171)```

File: src/video_core/texture_cache/texture_cache.cpp [hunk 2]
```diff
@@ -169,18 +168,21 @@ ImageId TextureCache::ResolveDepthOverlap(const ImageInfo& requested_info, Bindi
     }
 
     if (recreate) {
-        auto new_info{requested_info};
-        new_info.resources = std::max(requested_info.resources, cache_image.info.resources);
-        new_info.UpdateSize();
+        auto new_info = requested_info;
+        new_info.resources = std::min(requested_info.resources, cache_image.info.resources);
```
squidbus, line 172: FYI: Since this change from max to min, CUSA16404 now spams:
```
[Render.Vulkan] <Warning> texture_cache.cpp:405 FindImage: Image overlap resolve failed
```

raphaelthegreat: Was it the removal of UpdateSize or the new resource change?
The UpdateSize I removed because the requested ImageInfo was constructed from a depth buffer and running UpdateSize resulted in a slightly different size, so the new depth image would not be an exact match with the info and it would keep trying to execute the resolve depth function creating new images.

squidbus: Just the min/max change, changing that one part back removes the spam.

PR #3082, opened 2025-06-11
shader_recompiler: Reduce cases where shared memory to buffer pass is needed.

squidbus:
Reduce instances where the shared memory to buffer pass is needed by:
* Modifying the pass to ensure that workgroup explicit layout support is only a factor if multiple shared memory types are needed.
* Modifying the shared memory definition logic to only define used shared memory types, and avoid decorating for aliasing if only one is used.
* Adding a new pass to simplify 16-bit and 64-bit shared memory accesses to 32-bit if atomics of the respective type are not used. This pass only operates if workgroup explicit layout is not supported, as otherwise we can trivially alias shared memory.

As usual additional regression testing is appreciated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -870,6 +870,7 @@ set(SHADER_RECOMPILER src/shader_recompiler/exception.h
                       src/shader_recompiler/ir/passes/ring_access_elimination.cpp
                       src/shader_recompiler/ir/passes/shader_info_collection_pass.cpp
                       src/shader_recompiler/ir/passes/shared_memory_barrier_pass.cpp
+                      src/shader_recompiler/ir/passes/shared_memory_simplify_pass.cpp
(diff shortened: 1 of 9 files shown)


Missake212:
Seems to regress Driveclub (CUSA00003), crashes before getting in game.
[CUSA00003.log](https://github.com/user-attachments/files/20690656/CUSA00003.log)
image

squidbus:
Should be fixed

Missake212:
That fixed it, but now both Uncharted 4 and The Last of US part 2 crash here instead of throwing the usual

``[Debug] address_space.cpp:operator():574: Assertion Failed!``
``Partial unmapping of flexible allocations is not supported``
image

Not sure if that's a regression or an improvement.

squidbus:
I don’t see how memory mapping errors would be related to these changes.

Missake212:
Forgive me, I guess the debugger didn't want to throw the assert in the log, with the action it throws it. No other regressions on my end.

PR #3090, opened 2025-06-12
There was no translation i ts file.

zemzema:


georgemoralis:
you need to update the translation into crowdin not here

PR #3091, opened 2025-06-12
User management

georgemoralis:
PS4 unlikely previous consoles is user oriented , that means that we have to have user management . 

This is still WIP , i will update it once it goes on


georgemoralis:
Just a pic

image

GHU7924:
The idea is great, but @georgemoralis, how far have you thought through this PR?

If we have several users, then each one needs its own settings. For example, linking Discord. Maybe different people play the emulator on the same device.

1) We have a user tab where you can set the user name, maybe there should be a replacement for the user's choice or a button to open the User Manager.
2) Create User = Create New User (how many users will there be? Are there any restrictions?)
3) As I understand it, there is no user choice yet, something like switching users (logging out of the account and logging in as another).
4) The name of the user who was selected should be written somewhere.
5) The User folder should now probably contain folders with usernames and each user's saves and patches in them, but then it's worth moving the modules folder somewhere else (if you leave it in the folder with the other user folders, you might accidentally delete it... unless you do it manually, that should count too).

The PR is great, but a lot needs to be taken into account and made convenient.

georgemoralis:
> The idea is great, but @georgemoralis, how far have you thought through this PR?
> 
> If we have several users, then each one needs its own settings. For example, linking Discord. Maybe different people play the emulator on the same device.
> 
>     1. We have a user tab where you can set the user name, maybe there should be a replacement for the user's choice or a button to open the User Manager.
> 
>     2. Create User = Create New User (how many users will there be? Are there any restrictions?)
> 
>     3. As I understand it, there is no user choice yet, something like switching users (logging out of the account and logging in as another).
> 
>     4. The name of the user who was selected should be written somewhere.
> 
>     5. The User folder should now probably contain folders with usernames and each user's saves and patches in them, but then it's worth moving the modules folder somewhere else (if you leave it in the folder with the other user folders, you might accidentally delete it... unless you do it manually, that should count too).
> 
> 
> The PR is great, but a lot needs to be taken into account and made convenient.

ok most of the things you ask , i have already solutions but let me go further with this pr and we can discuss it :D

georgemoralis:
we can now populate table with active user

image

GHU7924:
>User ID - 00000001

👀 When I asked about user limits, I thought it would be 16 (or less), because if I remember correctly, on the original hardware you could only have 16 simultaneous accounts on one device. Then you could do something like number 1 out of 16 (001 of 016).

georgemoralis:
> > User ID - 00000001
> 
> 👀 When I asked about user limits, I thought it would be 16 (or less), because if I remember correctly, on the original hardware you could only have 16 simultaneous accounts on one device. Then you could do something like number 1 out of 16 (001 of 016).

ps4 can have 32bit unsigned user ids , and it is not even integers , it is random hex numbers . At least i tried to keep the 32 bit part :D

w1naenator:
> > > User ID - 00000001
> > 
> > 
> > 👀 When I asked about user limits, I thought it would be 16 (or less), because if I remember correctly, on the original hardware you could only have 16 simultaneous accounts on one device. Then you could do something like number 1 out of 16 (001 of 016).
> 
> ps4 can have 32bit unsigned user ids , and it is not even integers , it is random hex numbers . At least i tried to keep the 32 bit part :D

How I understand values 0, 254, 255, 4294967295 are reserved.

red-prig:
The user ID is a 32-bit number with a hexadecimal mask of 1XXXXXXX, where X contains a random number determined when the user is created.

red-prig:
> > > > User ID - 00000001
> > > 
> > > 
> > > 👀 When I asked about user limits, I thought it would be 16 (or less), because if I remember correctly, on the original hardware you could only have 16 simultaneous accounts on one device. Then you could do something like number 1 out of 16 (001 of 016).
> > 
> > 
> > ps4 can have 32bit unsigned user ids , and it is not even integers , it is random hex numbers . At least i tried to keep the 32 bit part :D
> 
> How I understand values 0, 254, 255, 4294967295 are reserved.

4294967295 is -1 for the sign representation used to provide users by default, 0, 254, 255 are not used at all but cannot be said to be reserved

georgemoralis:
reimplemented on this : https://github.com/shadps4-emu/shadps4-qtlauncher/pull/181

PR #3095, opened 2025-06-13
KBM Input Bug Fixes / Added Binds

nickci2002:
Fixed input issues where some inputs would not bind when pressing (side mouse buttons, some symbols, etc). Also, fixed up code formatting in altered files (removed C-style casts and replaced with C++ <static_casts>, added a few macros and one member functions).

### DIFF
--- a/input_handler.h
+++ b/input_handler.h
@@ -1,0 +1,439 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include <array>
+#include <map>
+#include <string>
+#include <unordered_set>
+
+#include "SDL3/SDL_events.h"
+#include "SDL3/SDL_timer.h"
+
+#include "common/logging/log.h"
+#include "common/types.h"
+#include "core/libraries/pad/pad.h"
+#include "fmt/format.h"
+#include "input/controller.h"
+
+// +1 and +2 is taken
+#define SDL_MOUSE_WHEEL_UP SDL_EVENT_MOUSE_WHEEL + 3
+#define SDL_MOUSE_WHEEL_DOWN SDL_EVENT_MOUSE_WHEEL + 4
+#define SDL_MOUSE_WHEEL_LEFT SDL_EVENT_MOUSE_WHEEL + 5
+#define SDL_MOUSE_WHEEL_RIGHT SDL_EVENT_MOUSE_WHEEL + 7
+
+// idk who already used what where so I just chose a big number
+#define SDL_EVENT_MOUSE_WHEEL_OFF SDL_EVENT_USER + 10
+
+#define LEFTJOYSTICK_HALFMODE 0x00010000
+#define RIGHTJOYSTICK_HALFMODE 0x00020000
+#define BACK_BUTTON 0x00040000
+
+#define KEY_TOGGLE 0x00200000
+
+namespace Input {
+using Input::Axis;
+using Libraries::Pad::OrbisPadButtonDataOffset;
+
+struct AxisMapping {
+    u32 axis;
+    s16 value;
+    AxisMapping(SDL_GamepadAxis a, s16 v) : axis(a), value(v) {}
+};
+
+enum class InputType { Axis, KeyboardMouse, Controller, Count };
+const std::array<std::string, 4> input_type_names = {"Axis", "KBM", "Controller", "Unknown"};
+
+class InputID {
+public:
+    InputType type;
+    u32 sdl_id;
+    InputID(InputType d = InputType::Count, u32 i = UINT32_MAX) : type(d), sdl_id(i) {}
+    bool operator==(const InputID& o) const {
+        return type == o.type && sdl_id == o.sdl_id;
+    }
+    bool operator!=(const InputID& o) const {
+        return type != o.type || sdl_id != o.sdl_id;
+    }
+    bool operator<=(const InputID& o) const {
+        return type <= o.type && sdl_id <= o.sdl_id;
+    }
+    bool IsValid() const {
+        return *this != InputID();
+    }
+    std::string ToString() {
+        return fmt::format("({}: {:x})", input_type_names[static_cast<u8>(type)], sdl_id);
+    }
+};
+
+class InputEvent {
+public:
+    InputID input;
+    bool active;
+    s8 axis_value;
+
+    InputEvent(InputID i = InputID(), bool a = false, s8 v = 0)
+        : input(i), active(a), axis_value(v) {}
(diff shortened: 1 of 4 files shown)


squidbus:
Why is this just adding 4 source files in the root of the project? Your commit should be modifying the files where they actually are.

input_handler.h
```diff
@@ -0,0 +1,439 @@
...
+    {"kp8", SDLK_KP_8},
+    {"kp9", SDLK_KP_9},
+    {"kp .", SDLK_KP_PERIOD},
+    {"kp ,", SDLK_KP_COMMA},
+    {"kp /", SDLK_KP_DIVIDE},
+    {"kp *", SDLK_KP_MULTIPLY},
+    {"kp -", SDLK_KP_MINUS},
+    {"kp +", SDLK_KP_PLUS},
+    {"kp =", SDLK_KP_EQUALS},
+    {"kp enter", SDLK_KP_ENTER},
```
kalaposfos13, line 247: These don't work, due to having a space in them, and the parser automatically strips whitespace, meaning no config will ever match to these
Example:
```
pad_right = kp -
pad_left = kp +
```
```
[Input] <Warning> input_handler.cpp:379 ParseInputConfig: Invalid format at line: 18, data: "pad_right=kp-", skipping line.
[Input] <Warning> input_handler.cpp:379 ParseInputConfig: Invalid format at line: 19, data: "pad_left=kp+", skipping line.
```

input_handler.h
```diff
@@ -0,0 +1,439 @@
...
+    {"7", SDLK_7},
+    {"8", SDLK_8},
+    {"9", SDLK_9},
+    
+    // symbols
+    {"`", SDLK_GRAVE},
+    {"~", SDLK_TILDE},
+    {"!", SDLK_EXCLAIM},
+    {"@", SDLK_AT},
+    {"#", SDLK_HASH},
```
kalaposfos13, line 170: Why did you switch to using symbols instead of names like `grave`, `tilde` etc? They do work, I'm just curious. Also, you'll have to update the text in kbm_help_dialog.h to reflect both this change and the new keys added in general.

input_handler.h
```diff
@@ -0,0 +1,439 @@
...
+};
+
+enum class InputType { Axis, KeyboardMouse, Controller, Count };
+const std::array<std::string, 4> input_type_names = {"Axis", "KBM", "Controller", "Unknown"};
+
+class InputID {
+public:
+    InputType type;
+    u32 sdl_id;
+    InputID(InputType d = InputType::Count, u32 i = UINT32_MAX) : type(d), sdl_id(i) {}
```
kalaposfos13, line 52: If you're already cleaning up these unnamed magic values, you should define an enum for this, like SDL_INVALID_ID

kbm_gui.cpp
```diff
@@ -0,0 +1,1067 @@
...
+}
+
+// // Lambda version of GetModifiedButton
+// auto GetModifiedButton = [](Qt::KeyboardModifiers modifier, const std::string& m_button, const std::string& n_button) -> QString {
+//     if (QApplication::keyboardModifiers() & modifier) {
+//         return QString::fromStdString(m_button);
+//     } else {
+//         return QString::fromStdString(n_button);
+//     }
+// };
```
kalaposfos13, line 1065: Since this is used in only one place, you're free to use lambdas, no one is stopping you from it, however you should decide on one and the other should be removed.

kbm_gui.cpp
```diff
@@ -0,0 +1,1067 @@
...
+                break;
+            case Qt::Key_AsciiCircum:
+                pressedKeys.insert("^");
+                break;
+            case Qt::Key_Ampersand:
+                pressedKeys.insert("&");
+                break;
+            case Qt::Key_Asterisk:
+                pressedKeys.insert(GetModifiedButton(Qt::KeypadModifier, "kp *", "*"));
+                break;
```
kalaposfos13, line 802: Same issue as before, it has a space in it

kbm_gui.h
kalaposfos13, line 1: Apart from everything else I said, as squidbus also mentioned, the files are in the wrong place, please fix that too.

kalaposfos13 (changes requested):

input_handler.h
```diff
@@ -0,0 +1,439 @@
...
+    {"<", SDLK_LESS},
+    {".", SDLK_PERIOD},
+    {">", SDLK_GREATER},
+    {"/", SDLK_SLASH},
+    {"?", SDLK_QUESTION},
+
+    // special keys
+    {"printscreen", SDLK_PRINTSCREEN},
+    {"scrolllock", SDLK_SCROLLLOCK},
+    {"pausebreak", SDLK_PAUSE},
```
kalaposfos13, line 202: You also removed escape from this list, add it back

nickci2002: Pressing escape when binding keys in the UI causes the value to be "unmapped". I didn't add it, it was already in the codebase.

input_handler.h
nickci2002: It's more intuitive to show the symbol than the word for a symbol. It's like putting "letter a" for a bind to the "a" button.

kbm_gui.h
nickci2002: I uploaded the files directly in GitHub. I thought it would replace the files automatically, not add them to root. My bad :/

input_handler.h
kalaposfos13: While I don't fault you for this misunderstanding, you're working on an incorrect assumption here. When input remapping was first added, I only made a very simple text editor GUI for it (that you can still access to this day with the Text editor button in the Keyboard menu), which is still more powerful than the new GUI by rainmakerv3, somewhat akin to how a GUI wrapper for a CLI app might be easier to use, at the expense of not having all the features the original program has. No offense to rainmaker with this, the remapping backend simply has too much things in it for it to have a feature-complete graphical interface that isn't a nightmare to either develop or use. This new menu also introduced remapping by pressing the corresponding key on your keyboard/mouse, which, while very cool, also arbitrarily chose to assign escape to unmap the current selection, which I didn't entirely agree with, but since I almost never use the graphical interface of the emulator, and prefer to edit config files directly instead, I simply didn't care much about it. However, I have escape mapped to options in every game, and I'd think many other people out there also do the same, as that is a very common keybind, so you shouldn't remove it from being able to map it at all.
TL;DR: This is only a limitation in the GUI, you can still bind escape in the text editor and therefore it shouldn't be removed.

kalaposfos13:
Why did you close this? Apart from the files being in the wrong place, and the few easily correctable nitpicks I mentioned, there was nothing wrong with it

nickci2002:
I was experimenting with git and deleted the branch. I'm learning GitKraken and I ended up butchering the branch so I just made a new one.

PR #3101, opened 2025-06-16
Changed the order of defining the current code

ivan999:
GcnDecodeContext private methods structure has been changed.
getEncodingLength, getOpMapOffset have been deleted.

GetInstructionEncoding function has been deleted.
Logic from this function has been put to a set of new private methods.

The main idea was to reduce repeating of the switch cases.

Small corrections have been added too:
32 bits encodings methods parameter has been changed in the header file to u32, to be the same as in the cpp file.
InstEncoding::ILLEGAL is not needed anymore, so it has been removed.


ivan999:
I have changed the formatting to pass the clang check.

ivan999:
I have changed the order of checking the codes in the purpose to clarify the most frequently used codes faster. 

output

VOP1, VOPC, and VOP2 can be checked first because they're the only instruction types with the top bit (bit 31) set to 0.

Among them, VOP1 and VOPC should be checked before VOP2, because their encodings are identified using the top 7 bits (MASK_7bit), which provides more specificity than the single top bit used by VOP2 (MASK_1bit).

This way, VOP1 and VOPC won’t accidentally be misclassified as VOP2.

ivan999:
I haven't changed anything in the last version not be able to build the project on MacOS.

ivan999:
I merged with main, maybe I had been loosing some fixes before...

PR #3105, opened 2025-06-16
Log improper image format uses

mikusp:
New logs for situations where images are used in ways violating Vulkan spec


georgemoralis:
this seems quite interesting let's see what @squidbus has to say

squidbus:
Is this extra log needed? We already log from the start what format capabilities are missing compared to what the docs say the AMD GPU may use.

mikusp:
Not needed strictly speaking but it might be more informative when people report issues without Vulkan validation turned on - some device losts might originate in such improper format usage

PR #3109, opened 2025-06-17
KBM Input Bug Fixes / Added Binds v2

nickci2002:
Fixed input issues where some inputs would not bind when pressing (side mouse buttons, some symbols, etc). Also, fixed up code formatting in altered files (removed C-style casts and replaced with C++ <static_casts>, added a few macros and one member functions).
This is v2 of my commit, addressing all issues brought up by @kalaposfos13


rainmakerv3:
This reverts https://github.com/shadps4-emu/shadPS4/pull/3098, but otherwise looks fine to me

src/input/input_handler.cpp
```diff
@@ -223,8 +223,10 @@ void ParseInputConfig(const std::string game_id = "") {
     while (std::getline(file, line)) {
         lineCount++;
 
+        auto line_kpmask = (line.substr(0, 2) == "kp ") ? line.begin() + 2 : line.begin();
```
kalaposfos13: Even ignoring the build failure (don't worry, we've all been been there, either Windows or Mac always have some strange bullshit going on that makes building fail), this will not work. You're trying to check whether or not the first 3 characters in the line equal to "kp ", but you're wrong in multiple ways here:
- line.substr(0, 2) only gets the first two characters, not three
- This also assumes no leading whitespace, which, while reasonable, is technically a regression from the previous version as that one handled that
- Keypad inputs will only ever be on the right side of the equal sign, as they are inputs that you bind to controller outputs that are on the left, not the other way around

I recommend going back to no space in these or switching it to an underscore instead, and reverting the parsing changes, although I do get that you did this to avoid strings like "kp+" or "kp_*", for example. In my opinion, even those aren't that bad, but I'll leave this up to you.
P. S: The rest of my prior comments have all been addressed, to which I'm thankful for, so nice work, although I do recommend that you do an at least cursory testing of your new features next time.

kalaposfos13 (changes requested):

src/qt_gui/kbm_help_dialog.h
```diff
@@ -102,47 +104,61 @@ Whitespace doesn't matter, <output>=<input>; is just as valid as <output> = <inp
...
-        'axis_left_x_plus', 'axis_left_x_minus', 'axis_left_y_plus', 'axis_left_y_minus',
-        'axis_right_x_plus', ..., 'axis_right_y_minus',
-        'l2'
-    Axes if you bind them to another axis input:
-        'axis_left_x' 'axis_left_y' 'axis_right_x' 'axis_right_y',
-        'l2'
-)";
+        QString result;
+        result += """The following names should be interpreted without the '' around them, and for inputs that have left and right versions, only the left one is shown, but the right can be inferred from that.\n";
+        result += "Example: 'lshift', 'rshift'\n\n";
```
kalaposfos13: Congrats, you found a bug in Qt's MOC parser. If you move the function body (which is just a `return R"(...)";` here) to the source file, the issue will be fixed, as it will no longer be parsed by MOC, only the declaration. Please note that if you do that, you should move all of these, not just this one, to remain consistent.

src/input/input_handler.cpp
```diff
@@ -254,8 +250,14 @@ void ParseInputConfig(const std::string game_id = "") {
 
         std::string output_string = line.substr(0, equal_pos);
         std::string input_string = line.substr(equal_pos + 1);
-        std::size_t comma_pos = input_string.find(',');
+        // Remove trailing semicolon from input_string
+        if (!input_string.empty() && input_string[input_string.length() - 1] == ';') {
+            if (!(input_string == ";" || input_string == "kp;")) {
+                line = line.substr(0, line.length() - 1);
+            }
+        }
```
kalaposfos13: While I get the reasoning behind this, what I don't see is why check for "kp;"? That isn't a valid key
Also, you naming the commit `Fixed formatting from clang-format` made me skip over it at first, you shouldn't name commits in a misleading way

nickci2002 (commented):
Removed kp;

kalaposfos13 (commented):
If you fix the merge conflict then I think the PR is good to go.

src/qt_gui/kbm_help_dialog.cpp
```diff
@@ -110,3 +110,142 @@ HelpDialog::HelpDialog(bool* open_flag, QWidget* parent) : QDialog(parent) {
...
+
+QString HelpDialog::faq() {
+    return R"(Q: What are the emulator-wide keybinds?
+A: -F12: Triggers Renderdoc capture
+-F11: Toggles fullscreen
+-F10: Toggles FPS counter
+-Ctrl F10: Open the debug menu
+-F9: Pauses emultor, if the debug menu is open
+-F8: Reparses the config file while in-game
+-F7: Toggles mouse capture and mouse input
```
kalaposfos13, line 134: You left out the newly added F6

src/qt_gui/kbm_help_dialog.cpp
```diff
@@ -110,3 +110,142 @@ HelpDialog::HelpDialog(bool* open_flag, QWidget* parent) : QDialog(parent) {
...
+
+Controller:
+    The touchpad currently can't be rebound to anything else, but you can bind buttons to it.
+    If you have a controller that has different names for buttons, it will still work, just look up what are the equivalent names for that controller
+    The same left-right rule still applies here.
+    Buttons:
+            'triangle', 'circle', 'cross', 'square', 'l1', 'l3',
+            'options', touchpad', 'up', 'down', 'left', 'right'
+        Input-only:
+             'lpaddle_low', 'lpaddle_high'
```
kalaposfos13, line 219: There's no touchpad here either

src/input/input_handler.h
```diff
@@ -56,7 +56,7 @@ class InputID {
 public:
     InputType type;
     u32 sdl_id;
-    InputID(InputType d = InputType::Count, u32 i = (u32)-1) : type(d), sdl_id(i) {}
+    InputID(InputType d = InputType::Count, u32 i = SDL_UNMAPPED) : type(d), sdl_id(i) {}
```
kalaposfos13, line 59: I defined SDL_UNMAPPED as UINT32_MAX - 1 as to not conflict with SDL_INVALID_ID in this PR, but you're right, the two can be merged into one value without issue

kalaposfos13 (changes requested):

src/qt_gui/kbm_help_dialog.cpp
```diff
@@ -110,3 +110,142 @@ HelpDialog::HelpDialog(bool* open_flag, QWidget* parent) : QDialog(parent) {
...
+
+'key_toggle' = <key>, <key_to_toggle>; 
+    This assigns a key to another key, and if pressed, toggles that key's virtual value. If it's on, then it doesn't matter if the key is pressed or not, the input handler will treat it as if it's pressed.
+    You can make an input toggleable with this, for example: Let's say we want to be able to toggle l1 with t. You can then bind l1 to a key you won't use, like kpenter, then bind t to toggle that, so you will end up with this:
+        l1 = kpenter;
+        key_toggle = t, kpenter;
+'analog_deadzone' = <device>, <value>, <value>;
+    Values go from 1 to 127 (no deadzone to max deadzone), first is the inner, second is the outer deadzone
+    If you only want inner or outer deadzone, set the other to 1 or 127, respectively
+    Devices: leftjoystick, rightjoystick, l2, r2)";
```
kalaposfos13: `mouse_gyro_roll_mode` is similarly missing from here (I thought I added this to my review but I guess not)

kalaposfos13 (approved):

PR #3117, opened 2025-06-18
Potential MacOS Build Fix

nickci2002:
Changed std::string to std::string_view after new update broke the MacOS build.


kalaposfos13:
I think it'd be better to add `#include <string>` to the top of the file instead

kalaposfos13:
This PR only fixes SDL builds, Qt builds still fail with a linker error.

Issue #2649, opened 2025-03-15
[Feature Request]: Favorites in the game list


GHU7924:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

Add a "Favorites" column for the list of games. Add a checkbox to each line for this column.
If a check mark is set for the game, then the game gets to the top of the list.
The same sorting that you have selected was also applied to the games from your favorites (for example, by name).

### Reason

If you have a large library of games, but you only play two or three games (for example, because you want to complete them or only they are currently working), then it would be convenient to place these games at the top of the list, so as not to search for them every time in the game list (in fact, your library). 

### Examples

![Image](https://github.com/user-attachments/assets/b4e51a12-6aff-425e-a7bc-3230f7249cf7)

GHU7924:
Done in #3071. Thanks to @davidantunes23 and @JoooPT for the implementation.

PR #3119, opened 2025-06-19
QT save fixes II

georgemoralis:
This can also be an example of how to fix the saves of favourites for #2649 

1. Added recent files saves to qt save file
2. gui language selection
3. FIX: savedata was using gui language and not console language


georgemoralis:
fixed language issue with savedata (it was saving based on gui language and not on console language)

PR #3122, opened 2025-06-19
Emulate motion controls with a mouse

kalaposfos13:
Adds the ability to emulate motion control inputs via a mouse. Since a gyroscope is a three-dimensional sensor, while a mouse is only two-dimensional, the full range of controls is achieved by having a (rebindable) toggle key to switch the behaviour of moving the mouse sideways between panning and rolling. While this is active, acceleration is spoofed to a constant downwards direction.

To enable it, press F6. You can also seamlessly switch between joystick and gyro emulation, if you select one while the other is active, it will cleanly switch modes, and if you press the toggle for the currently active mode, it will turn off.

The name for the keybind that switches between panning and rolling is `mouse_gyro_roll_mode`.

This is only the backend code and since this seems to be true for almost every PR I make, I think it will stay that way and I will pass off integrating it into the frontend to someone else.

Since it's using the same setup as joystick emulation, this is also conflicting with gyro data coming from a connected controller the same way joystick data does.

I only tested it with Gravity Rush 2, so further testing is very welcome.

### DIFF
--- a/src/core/libraries/pad/pad.cpp
+++ b/src/core/libraries/pad/pad.cpp
@@ -447,21 +447,18 @@ int PS4_SYSV_ABI scePadReadState(s32 handle, OrbisPadData* pData) {
 
     // Only do this on handle 1 for now
     if (engine && handle == 1) {
-        const auto gyro_poll_rate = engine->GetAccelPollRate();
-        if (gyro_poll_rate != 0.0f) {
             auto now = std::chrono::steady_clock::now();
-            float deltaTime = std::chrono::duration_cast<std::chrono::microseconds>(
-                                  now - controller->GetLastUpdate())
+        float deltaTime =
+            std::chrono::duration_cast<std::chrono::microseconds>(now - controller->GetLastUpdate())
                                   .count() /
                               1000000.0f;
             controller->SetLastUpdate(now);
             Libraries::Pad::OrbisFQuaternion lastOrientation = controller->GetLastOrientation();
             Libraries::Pad::OrbisFQuaternion outputOrientation = {0.0f, 0.0f, 0.0f, 1.0f};
-            GameController::CalculateOrientation(pData->acceleration, pData->angularVelocity,
-                                                 deltaTime, lastOrientation, outputOrientation);
+        GameController::CalculateOrientation(pData->acceleration, pData->angularVelocity, deltaTime,
+                                             lastOrientation, outputOrientation);
             pData->orientation = outputOrientation;
             controller->SetLastOrientation(outputOrientation);
-        }
     }
     pData->touchData.touchNum =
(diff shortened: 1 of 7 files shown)


GHU7924:
I don't quite understand it yet, so I'd like to hear an explanation:

1) Here I have a mouse on the table, how does it relate to space?

2) As far as I understand, only rotations along the axes are implemented. I did not understand how tilts and other things happen.
Or, on the contrary, tilts work, but there is no rotation. I can't figure it out for sure. There is no visualizer on the screen.

inFAMOUS crashes for me with an error, so I only managed to do the first action once. It suggested tilting the gamepad as if you were holding a can of paint. I somehow did it. The next action was to shake the gamepad, which is how you shake the can, but I understand that this action is no longer available.

Then my game crashed over and over again, so I had to look for something else. And then I remembered Bloodborne, where I can use gestures, but I didn't figure it out either. The actions worked, but I didn't understand how I did it.

The PR seems to be working, but for me personally nothing is clear yet.

GHU7924:
You can test the PR in Discord with Steam Deck and Bloodborne owners.
Steam Deck has a gyroscope, in the settings there is an option to make the gyroscope work as a mouse.
Previously, it just rotated the camera, but with your new setting, everything may work differently now.

kalaposfos13:
> I don't quite understand it yet, so I'd like to hear an explanation

6DOF_en

If you take "forward" as the direction that the PS4 controller's lightbar is pointing, and "up" as the direction the controller buttons are facing, then moving the mouse up and down emulates a change in Pitch, and moving sideways corresponds to a change in Yaw by default or a change in Roll if you're holding the modifier key. If a game is relying on the accelerometer more than the gyro, then this approach will not work very well as acceleration is currently a constant thing, but from what you said it looks like this works well enough in inFAMOUS to pass the intro.

georgemoralis (approved):

GHU7924:
I know about the gyroscope itself, but your implementation is still complicated for me, I need visualization or something, otherwise I get confused where I need to move the mouse.

If we had the ability to select an action, it would be great. Maybe I'm overdoing it, but for example, I pressed F6, activated the gyroscope, it's at 0, then selected the desired axis, for example, vertical (up-down), for this, let's say, I have to hold down the 1 key and, moving the mouse forward, I get the up action, and if I move the mouse back, I get the down action. Key 2, let's say, left or right, that is, I also have to move the mouse left or right. The other axes also have their own key.

As for "Bad Reputation," I managed to film the beginning where I somehow manage to do the first act, but after that I don't know if I can do it or not.

https://github.com/user-attachments/assets/4380402a-51d5-481f-b2e9-96c975025474

As you can see, there are complex control schemes here, that is, I have to first change the position of the controller from horizontal to vertical and while maintaining this position I have to shake it a little bit to make everything go further.

PR #3125, opened 2025-06-20
Update gui with new touchpad inputs

rainmakerv3:
New touchpad commit breaks touchpad click when saved using the GUI. This updates the GUI to work with the three new touchpad inputs

### DIFF
--- a/src/qt_gui/control_settings.h
+++ b/src/qt_gui/control_settings.h
@@ -39,14 +39,28 @@ private:
         "pad_left",     "pad_right", "axis_left_x", "axis_left_y", "axis_right_x",
         "axis_right_y", "back"};
 
-    const QStringList ButtonOutputs = {"cross",    "circle",    "square",   "triangle", "l1",
-                                       "r1",       "l2",        "r2",       "l3",
+    const QStringList ButtonOutputs = {"cross",
+                                       "circle",
+                                       "square",
+                                       "triangle",
+                                       "l1",
+                                       "r1",
+                                       "l2",
+                                       "r2",
+                                       "l3",
 
-                                       "r3",       "options",   "pad_up",
+                                       "r3",
+                                       "options",
+                                       "pad_up",
 
                                        "pad_down",
 
-                                       "pad_left", "pad_right", "touchpad", "unmapped"};
+                                       "pad_left",
+                                       "pad_right",
+                                       "touchpad_left",
+                                       "touchpad_center",
+                                       "touchpad_right",
+                                       "unmapped"};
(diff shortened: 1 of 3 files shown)


src/qt_gui/control_settings.h
```diff
@@ -39,14 +39,28 @@ private Q_SLOTS:
...
 
                                        "pad_down",
 
-                                       "pad_left", "pad_right", "touchpad", "unmapped"};
+                                       "pad_left",
+                                       "pad_right",
+                                       "touchpad_left",
+                                       "touchpad_center",
+                                       "touchpad_right",
+                                       "unmapped"};
```
kalaposfos13, line 63: These are indeed button outputs, however you are using them as inputs, which is a bug that wasn't introduced with this PR, and was in the code since quite a while, but in any case you can't bind touchpad presses to other outputs

kalaposfos13 (changes requested):
The Controller GUI should probably also be updated to include touchpad remapping which it currently doesn't have in any form, and also the default config also needs to be updated

src/qt_gui/control_settings.h
rainmakerv3: These are outputs, they're saved as output strings. You can test the PR to confirm. All that was really needed was to add the output strings to the drop-down menu and the controller GUI is properly updated

kalaposfos13: In this case the functionalites of the keyboard and the controller GUI are inverted, in the keyboard GUI the selection is about what is the input, while in the controller one the selection is about what is the output, which is just inconsistent design

rainmakerv3: It is inconsistent yes, but both work and seemingly hasn't been an issue for users (they probably use only one or the other in the vast majority of cases).

I was/am still planning to address that inconsistency with a rewrite for the controller GUI (to listen to controller inputs similar to the KBM GUI) but I'm just not sure when I can find a time to do that.

Up to you guys though what you want to do, I think it's good to have this quick fix so that the GUI can work with the new touchpad inputs

georgemoralis: let's make them work and the rest can be short out with a new pr

rainmakerv3: ok, I'll try and at least switch inputs and outputs for controller soon-ish (i hope lol)

GHU7924:
I have already suggested moving the button settings to the controller window and reorganizing the elements a bit.

Image

I don't know how it will be with the new changes, but it might still be useful.

PR #3126, opened 2025-06-20
Common_Ntapi: Error handling via logs

Nerveustech:
Added error handling for the NTAPI functions to manage cases where they may fail and return NULL.

### DIFF
--- a/src/common/logging/filter.cpp
+++ b/src/common/logging/filter.cpp
@@ -67,7 +67,7 @@ bool ParseFilterRule(Filter& instance, Iterator begin, Iterator end) {
     CLS(Common)                                                                                    \
     SUB(Common, Filesystem)                                                                        \
     SUB(Common, Memory)                                                                            \
-	SUB(Common, NtApi)                                                                             \
+    SUB(Common, NtApi)                                                                             \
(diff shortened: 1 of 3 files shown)


red-prig:
This doesn't make much sense, this kind of error can happen in very rare cases, NtQueueApcThreadEx can be missing on systems below win10 which is already below the minimum system requirements, "ntdll.dll"/NtClose/NtSetInformationFile/NtCreateThread/NtTerminateThread can't fail at all on a normal system (even on Wine this won't happen as far as I know)

Nerveustech:
Ah, okay, thanks for the feedback! Honestly, I didn't know such specific details. Thanks again.

polybiusproxy:
Indeed, these calls should never fail. If they do, I'd consider reinstalling the OS. This just adds quite the amount of unnecessary code in my opinion. I would, at most, simply add asserts if I really wanted to check if they failed that much.

Nerveustech:
Thanks, should I implement ASSERT_MSG or close the pull request?

polybiusproxy:
Up to you. I'd personally close it though

Nerveustech:
Okay, thanks again guys for your feedback.

PR #3130, opened 2025-06-20
buffer_atomic_imax_x2

mikusp:
Used in KNACK 2 as reported on Discord


kalaposfos13:
Here are all the unimplemented opcodes in Knack 2 that I know of:
- V_CMP_GT_U64
- BUFFER_ATOMIC_UMIN_X2
- BUFFER_ATOMIC_UMAX_X2
- BUFFER_ATOMIC_FCMPSWAP

nickci2002:
There's only one 64-bit V_CMP opcode implemented out of the, like, 50.

kalaposfos13:
@nickci2002 our general rule of thumb is that we only implement things that are in use by at least one game, even if doing related opcodes would be trivial

mikusp:
@nickci2002 V_CMP_X_64 are not trivial to implement in the general case as they can be used to compare 64bit values as well as manipulate exec masks, and that needs special processing to track where the value is used

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
```diff
@@ -226,10 +226,18 @@ Id EmitBufferAtomicSMax32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id addre
...
+Id EmitBufferAtomicSMax64(EmitContext& ctx, IR::Inst* inst, u32 handle, Id address, Id value) {
+    return BufferAtomicU64(ctx, inst, handle, address, value, &Sirit::Module::OpAtomicSMax);
+}
+
 Id EmitBufferAtomicUMax32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id address, Id value) {
     return BufferAtomicU32(ctx, inst, handle, address, value, &Sirit::Module::OpAtomicUMax);
 }
 
+Id EmitBufferAtomicUMax64(EmitContext& ctx, IR::Inst* inst, u32 handle, Id address, Id value) {
+    return BufferAtomicU64(ctx, inst, handle, address, value, &Sirit::Module::OpAtomicUMax);
```
squidbus, line 238: These require feature ‘shaderBufferInt64Atomics’ and the SPIR-V capability ‘Int64Atomics’

nickci2002:
@mikusp I thought it was just like the 32-bit codes but with 64-bit sources. I copied the code from the 32-bit area but used getsrc64 instead of getsrc. 
I'll make a pr with V_CMP_EQ_U64 and V_CMP_GT_U64 for knack 2 and dead rising 4 by tonight. If you can test resident evil for V_CMP_NE_U64, that would be great.

squidbus:
> I thought it was just like the 32-bit codes but with 64-bit sources.

No, it is not that simple. They are often used with registers that have per-thread bits that require different implementation to handle.

nickci2002:
Is that the case for all 64-bit versions?

squidbus:
Yes, for compare.

squidbus:
You could probably get away with a simple implementation for some registers but have to make sure to include an assert on operands like EXEC.

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
mikusp: I've added that and also included SharedAtomicIAdd64 to be handled. However macOS doesn't support 64bit atomic adds - how to handle that case?

squidbus: Tbh I am not sure on a satisfactory fallback for this, we might either need to make a best attempt at 64-bit atomic fallback (even if it doesn’t fully properly work) or assert on missing support.

mikusp: I've left an assert, should only be hit by MoltenVK on games using those opcodes

squidbus (approved):

kalaposfos13:
For some reason I forgot to actually test this with Knack 2, and looks like it doesn't work correctly:
[CUSA08014.log](https://github.com/user-attachments/files/21020857/CUSA08014.log)
```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:516 CompileModule: Compiling cs shader 0x26f60de6 
shadps4: /media/kalaposfos/Shared/shadps4/repos/shadps4-testing/externals/sirit/src/stream.h:36: Sirit::OpId::OpId(spv::Op, Id): Assertion `result_type.value != 0' failed.
```
[callstack.txt](https://github.com/user-attachments/files/21020879/callstack.txt)

Randomuser8219:
> For some reason I forgot to actually test this with Knack 2, and looks like it doesn't work correctly: [CUSA08014.log](https://github.com/user-attachments/files/21020857/CUSA08014.log)
> 
> ```
> [Render.Vulkan] <Info> vk_pipeline_cache.cpp:516 CompileModule: Compiling cs shader 0x26f60de6 
> shadps4: /media/kalaposfos/Shared/shadps4/repos/shadps4-testing/externals/sirit/src/stream.h:36: Sirit::OpId::OpId(spv::Op, Id): Assertion `result_type.value != 0' failed.
> ```
> 
> [callstack.txt](https://github.com/user-attachments/files/21020879/callstack.txt)

And I was wondering why it just crashed without an error on Windows...

PR #3135, opened 2025-06-21
Stub out sceKernelSetGPO and sceKernelGetGPI

kalaposfos13:
Stubs out these two functions that are similarly stubbed on real hardware


squidbus:
For the return codes we usually use `s32`, not `u32`.

kalaposfos13:
Oops made a typo while search and replacing, should be correct now

PR #3138, opened 2025-06-21
Fix deprecated localtime warning in savedata dialog

dalathegreat:
### What
This PR replaces fmt::localtime(std::chrono::system_clock::to_time_t(...)) with direct std::chrono formatting

### Why
It should be possible to compile without deprecation warnings

image

### How
The change is purely technical (no functionality change)

- Output format remains identical (%d %b, %Y %R)
- Uses more modern C++ best practices
- Requires C++17 (but project already uses it)

Further improvements; This fixes Windows and Linux builds, MacOS is still built the old way.


kalaposfos13:
The same deprecation warning is also present in `src/core/devtools/widget/memory_map.cpp`.

dalathegreat:
Let's fix that in another PR! :+1:

squidbus:
I’m not sure I see much of a point if it isn’t supported by all of our platforms, we still rely on the deprecated function and now there’s more complexity and platform-specific differences.

PR #3139, opened 2025-06-21
Revert "Handle immediate inline samplers (#3015)"

kalaposfos13:
This PR caused a regression in Gravity Rush Remastered downgrading it from playable to menus and STRANGER OF PARADISE FINAL FANTASY ORIGIN, and since no solution has been found by @mikusp or anyone else so far, a proposal to revert it until a fix exists has been made.


kalaposfos13:
Due to #3143, this PR isn't needed anymore.

PR #3141, opened 2025-06-21
Add CMake Presets for Qt build and add auto-detection for Qt in Windows

Fire-Cube:
This is a continuation of the PR #3131 
It solves the problem of having to enable Qt by directly editing the file. I’ve added presets for Qt builds, and to make that work, an auto-detection for Qt on Windows.


CMakePresets.json
```diff
@@ -0,0 +1,115 @@
...
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/Install/x64-Clang-Debug"
+      },
+      "vendor": {
+        "microsoft.com/VisualStudioSettings/CMake/1.0": {
+          "intelliSenseMode": "windows-clang-x64"
+        }
+      }
+    },
+    {
+      "name": "x64-Clang-Debug-Qt",
```
Osyotr, line 34: Use inheritance to remove boilerplate.

CMakeLists.txt
```diff
@@ -214,6 +214,8 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/common/scm_rev.cpp.in" "${CMAKE_
 
 message("end git things, remote: ${GIT_REMOTE_NAME}, branch: ${GIT_BRANCH}")
 
+include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/findQt.cmake")
```
kalaposfos13: This should probably be wrapped in a `if(ENABLE_QT_GUI)` block, as SDL builds do not need this and will just result in an unnecessary warning if the path is not found

Osyotr: Doesn't Qt installer add Qt to PATH? I.e. CMake should be able to find Qt without additional setup.
Also, this script should be renamed, otherwise it's confusing because it's named as a CMake find module but it's not.

Fire-Cube: On my install it's not detected

Fire-Cube:
I’ve adjusted everything as requested, and it passed (my) checks.

Fire-Cube:
i need to check a thing

CMakeLists.txt
diegolix29: Yeah visual studio need to be manually edited to have the qt found also on my end. So this will be very useful

Fire-Cube:
Should be good now, tested on two different windows devices with CLion and Visual Studio

PR #3143, opened 2025-06-21
shader_recompiler: Fix handling unbound depth image.

squidbus:
When an unbound image is detected, make sure to use the null depth image for depth instructions instead of the regular null image. Should avoid some confusing messaging where the unbound null image is detected as an improper depth image.

### DIFF
--- a/src/shader_recompiler/info.h
+++ b/src/shader_recompiler/info.h
@@ -308,17 +308,17 @@ constexpr AmdGpu::Image ImageResource::GetSharp(const Info& info) const noexcept
     if (!is_r128) {
         image = info.ReadUdSharp<AmdGpu::Image>(sharp_idx);
     } else {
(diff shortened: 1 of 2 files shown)


squidbus:
@kalaposfos13 Don't expect this to fully fix but can you re-test the games from https://github.com/shadps4-emu/shadPS4/pull/3139?

squidbus:
Actually I think I realized how it leads to a crash. Again not sure if it will fully fix because I don't think the image was supposed to be detected as unbound in the first place, but please re-test as I think the depth assert should be gone.

kalaposfos13:
I can confirm that Gravity Rush Remastered does indeed gets fixed by this PR, and I don't own the other game nor was I the one who reported that regression, but gandalfthewhite also confirmed that this fixes STRANGER OF PARADISE FINAL FANTASY ORIGIN as well over on [Discord](https://discord.com/channels/1080089157554155590/1281178723214688279/1386192923376357507).

PR #3144, opened 2025-06-22
Controller Remapping GUI v2

rainmakerv3:
Re-did the controller remapping GUI to be similar to the KBM GUI, it now listens for controller inputs instead of choosing items from a drop-down box.

Also to make it more consistent with the KBM GUI, the selections now represent inputs instead of outputs.

For now, controller only be remapped while a game in not running (so as to simplify SDL event loop handling).

Have not been able to test a lot, would be good to get a couple testers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1068,6 +1068,8 @@ set(QT_GUI src/qt_gui/about_dialog.cpp
            src/qt_gui/gui_settings.h
            src/qt_gui/settings.cpp
            src/qt_gui/settings.h
+           src/qt_gui/sdl_event_wrapper.cpp
+           src/qt_gui/sdl_event_wrapper.h
            ${EMULATOR}
            ${RESOURCE_FILES}
            ${TRANSLATIONS}
--- a/src/qt_gui/control_settings.cpp
+++ b/src/qt_gui/control_settings.cpp
@@ -752,6 +752,8 @@ void ControlSettings::CheckMapping(QPushButton*& button) {
         button->setText(mapping);
         EnableButtonMapping = false;
         EnableAxisMapping = false;
+        L2Pressed = false;
+        R2Pressed = false;
         EnableMappingButtons();
         timer->stop();
     }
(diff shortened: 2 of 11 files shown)


kalaposfos13:
Currently it's not possible to map bindings with more than one input key, and I think you can just copy the logic for it from the Keyboard GUI to do that

rainmakerv3:
> Currently it's not possible to map bindings with more than one input key, and I think you can just copy the logic for it from the Keyboard GUI to do that

Yup, I put that in here too

kalaposfos13:
> Yup, I put that in here too

Hmm in this case it doesn't work for me, whenever I push a button on my controller it gets set immediately, maybe you switched up something and set the closing condition to pressing a key instead of releasing it?

rainmakerv3:
> > Yup, I put that in here too
> 
> Hmm in this case it doesn't work for me, whenever I push a button on my controller it gets set immediately, maybe you switched up something and set the closing condition to pressing a key instead of releasing it?

You can set it, but you shouldn't be able to save it, seem to work for me

kalaposfos13:
No I mean setting combinations like `cross = l3, cross`

rainmakerv3:
> No I mean setting combinations like `cross = l3, cross`

Ah I din't know you could do that. Sure, maybe tomorrow or the day after I'll add that in. Like you said, should just be the same logic

rainmakerv3:
Combo inputs now added. this is pretty cool to use with the multiple touchpad outputs,

rainmakerv3:
May have an idea how to make this work with the game's event loop, setting to draft while I try stuff

rainmakerv3:
Ok that seemed to have worked, ready for review

GHU7924:
I haven't tested it, but I did look at the controller window. There is one thing I personally think you should do, which is make sure the touchpad on the left (unmapped) is set to act as a back button, as it was before.

UPD: Hmmm... All 3 touchpads were `unmapped`, then I pressed restore default and the left touchpad was detected as `back`. It was weird.

https://github.com/user-attachments/assets/b8683ffd-f38d-4856-a69d-1c911e2216d9

rainmakerv3:
> I haven't tested it, but I did look at the controller window. There is one thing I personally think you should do, which is make sure the touchpad on the left (unmapped) is set to act as a back button, as it was before.
> 
> UPD: Hmmm... All 3 touchpads were `unmapped`, then I pressed restore default and the left touchpad was detected as `back`. It was weird.
> 
>  2025-06-24.14-53-43.mp4

This is because of the recent change to make separate inputs for touchpad left, touchpad center, and touchpad right. The old config files had none of these values so the three touchpad inputs will be detected as unmapped. Once a mapping for touchapd is saved,, this will no longer show as unmapped in the future

georgemoralis:
@kalaposfos13 !

kalaposfos13 (commented):
Update to main and it's good to go

rainmakerv3:
ok, rebased

PR #3146, opened 2025-06-22
Kbm GUI - minor fixes

rainmakerv3:
1. Put wheel and mouse events under the if (EnableMapping) condition to ensure no extra inputs are captured.
2.  Make GUI more responsive by emitting a signal to complete the mapping on key/mouse releases or unmodified wheel movements, instead of waiting for the per-second check
3. Bit of cleanup for things no longer used / unneeded

### DIFF
--- a/src/qt_gui/kbm_gui.cpp
+++ b/src/qt_gui/kbm_gui.cpp
@@ -874,7 +874,6 @@ bool KBMSettings::eventFilter(QObject* obj, QEvent* event) {
                 }
                 break;
             case Qt::Key_Meta:
-                activateWindow();
 #ifdef _WIN32
                 pressedKeys.insert("lwin");
 #else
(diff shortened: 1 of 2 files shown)


kalaposfos13:
I finally found out why mouse inputs didn't work for me most of the time, clicking on the button makes the button "steal" the click event, but if I click to the side of it setting the input works fine
This doesn't have much to do with the PR I just found this out while testing it

rainmakerv3:
> I finally found out why mouse inputs didn't work for me most of the time, clicking on the button makes the button "steal" the click event, but if I click to the side of it setting the input works fine This doesn't have much to do with the PR I just found this out while testing it

Thanks for letting me know, somehow wasn't aware of this! I don't think this behavior can be overriden, but probably could be fixed if I used SDL events for mouse instead. Dunno if it that's worth it yet though

georgemoralis:
@kalaposfos13 is that good to merge?

kalaposfos13 (commented):
Update to main and it's good to go

rainmakerv3:
Ok, rebased both

PR #3147, opened 2025-06-22
libkernel: Check returned module in sceKernelGetModuleInfoFromAddr

StevenMiller123:
Fixes another issue @kalaposfos13 hit while modding Driveclub.


StevenMiller123:
Swapping to draft so I can validate the error we're returning.

StevenMiller123:
Some quick decompilation shows I got the error return wrong, and that our GetModuleInfoForUnwind also had the same wrong error return.
I've fixed that, so I've swapped this back to ready for review.

src/core/libraries/kernel/process.cpp
```diff
@@ -118,11 +118,23 @@ s32 PS4_SYSV_ABI sceKernelGetModuleInfoForUnwind(VAddr addr, s32 flags,
...
 }
 
-int PS4_SYSV_ABI sceKernelGetModuleInfoFromAddr(VAddr addr, int flags,
+s32 PS4_SYSV_ABI sceKernelGetModuleInfoFromAddr(VAddr addr, s32 flags,
                                                 Core::OrbisKernelModuleInfoEx* info) {
+    if (flags >= 3) {
+        std::memset(info, 0, sizeof(Core::OrbisKernelModuleInfoEx));
+        return ORBIS_KERNEL_ERROR_EINVAL;
+    }
+    if (info == nullptr) {
```
squidbus, line 127: Having a case that uses memset on the info before a null check seems wrong.

StevenMiller123: Unless I'm misunderstanding something, this appears to be how the library behaves.
You should be able to validate in your own libkernel decompilation if you need to.

squidbus: I’ll take a look later, I’m a bit torn if this is how the real library orders it because this doesn’t look safe if the info is nullptr and we take the flags case before that check.

squidbus: Looks like it does indeed do this... oh well

PR #3153, opened 2025-06-23
V_CMP_EQ_U64 support

nickci2002:
Added V_CMP_EQ_U64 opcode support for Dead Rising 4 (https://github.com/shadps4-emu/shadPS4/issues/496#issuecomment-2933971873) and added 64-bit relational operators (<,>,<=,>=).


src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -996,39 +998,64 @@ void Translator::V_CMP_U32(ConditionOp op, bool is_signed, bool set_exec, const
...
+//     }
+//     switch (inst.dst[1].field) {
+//     case OperandField::VccLo:
+//         return ir.SetVcc(result);
+//     case OperandField::ScalarGPR:
+//         return ir.SetThreadBitScalarReg(IR::ScalarReg(inst.dst[1].code), result);
+//     default:
+//         UNREACHABLE();
+//     }
+// }
```
squidbus: Would rather not leave commented code here, add it when it's ready.

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -996,39 +998,64 @@ void Translator::V_CMP_U32(ConditionOp op, bool is_signed, bool set_exec, const
...
+
+void Translator::V_CMP_EQ_U64(ConditionOp op, const GcnInst& inst) {
```
squidbus: Even if this is only EQ so far we should just name it `V_CMP_U64` since presumably other cases would be added here later.

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -996,39 +998,64 @@ void Translator::V_CMP_U32(ConditionOp op, bool is_signed, bool set_exec, const
...
+    const IR::U64 src1{GetSrc64(inst.src[1])};
+    const IR::U1 result = [&] {
+        switch (op) {
+        case ConditionOp::EQ:
+            return ir.IEqual(src0, src1);
+        case ConditionOp::LG: // NE
+            return ir.INotEqual(src0, src1);
         default:
-            UNREACHABLE_MSG("unhandled V_CMP_NE_U64 source argument {}", u32(inst.src[1].field));
+            UNREACHABLE_MSG("Unsupported V_CMP_EQ_U64 condition operation: {}", u32(op));
```
squidbus: Same here for the name

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -996,39 +998,64 @@ void Translator::V_CMP_U32(ConditionOp op, bool is_signed, bool set_exec, const
...
         }
-    };
+    }();
+
     switch (inst.dst[1].field) {
     case OperandField::VccLo:
-        ir.SetVcc(op(src0));
-        break;
+        return ir.SetVcc(result);
```
squidbus, line 1024: Prefer to just leave as set then break, no need to return void.

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -996,39 +998,64 @@ void Translator::V_CMP_U32(ConditionOp op, bool is_signed, bool set_exec, const
...
     case OperandField::ScalarGPR:
-        ir.SetThreadBitScalarReg(IR::ScalarReg(inst.dst[1].code), op(src0));
-        break;
+        return ir.SetThreadBitScalarReg(IR::ScalarReg(inst.dst[1].code), result);
```
squidbus, line 1026: Same here as other case.

squidbus:
Just some small comments

squidbus (approved):

PR #3154, opened 2025-06-24
libSceVideodec2: Fix sceVideodec2GetPictureInfo regressions

StevenMiller123:
After #3077 and #3087, it became clear that something was wrong with the outputted AvcPictureInfo in some games. After increasing the struct size, they were no longer able to access the picture data properly.

This PR takes a simple approach to resolving these regressions, keep the old AvcPictureInfo struct, and create a separate vector for storing picture infos using the legacy struct. The necessary struct can be determined using received OrbisVideodec2OutputInfo's thisSize parameter.

This fixes the recent regression in Uncharted: The Nathan Drake Collection™ (CUSA02320)

I'm open to cleaner solutions if anyone has better ideas for solving this, as this is a sort of brute-force solution.

### DIFF
--- a/src/core/libraries/videodec/videodec2.cpp
+++ b/src/core/libraries/videodec/videodec2.cpp
@@ -171,19 +171,40 @@ s32 PS4_SYSV_ABI sceVideodec2GetPictureInfo(const OrbisVideodec2OutputInfo* outp
         LOG_ERROR(Lib_Vdec2, "Invalid struct size");
         return ORBIS_VIDEODEC2_ERROR_STRUCT_SIZE;
     }
-    if (outputInfo->pictureCount == 0 || gPictureInfos.empty()) {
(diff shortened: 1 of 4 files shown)


squidbus:
I see this got merged already but I strongly disagree with this approach, this is not how different struct version should be handled at all.

StevenMiller123:
I'm open to alternatives, I'm aware this is a very messy solution.
My main goal was getting a fix out there, since we've got several PRs these regressed titles are affected by and a release coming in a few weeks. I tried some simpler solutions, but couldn't find a clean way to get these games to cooperate.

squidbus:
Well personally I would have just reverted the other PRs until it’s fixed.

georgemoralis:
surely it can be consider a workaround until a better solution found

PR #3155, opened 2025-06-24
video_core: Page manager and memory tracker improvenets

LNDF:
This adds some improvements to the page manager and memory tracker:
* Better calculation of what memory pages to (un)protect
* Increese region manager from 4 MB to 16 MB
* Regional locking in page manager. Currently in main, a single mutex is used to lock all watcher updates, creating a bottleneck every time memory is tracked or untracked. This creates a lock for every 1024 pages (16 MB). This is currently the same as memory tracker, I don't think it makes sense for it to be more granular.

### DIFF
--- a/src/video_core/page_manager.cpp
+++ b/src/video_core/page_manager.cpp
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include <boost/container/small_vector.hpp>
+#include <boost/thread/lock_algorithms.hpp>
 #include "common/assert.h"
 #include "common/debug.h"
 #include "common/signal_context.h"
@@ -194,11 +195,14 @@ struct PageManager::Impl {
         size_t page = addr >> PAGE_BITS;
         const u64 page_end = Common::DivCeil(addr + size, PAGE_SIZE);
 
-        const size_t lock_start = page / PAGES_PER_LOCK;
+        // Acquire locks for the range of pages
+        size_t lock_start = page / PAGES_PER_LOCK;
         const size_t lock_end = Common::DivCeil(page_end, PAGES_PER_LOCK);
-        for (size_t i = lock_start; i < lock_end; ++i) {
-            locks[i].lock();
+        boost::container::small_vector<std::unique_lock<LockType>, 8> unique_locks;
+        for (; lock_start < lock_end; ++lock_start) {
+            unique_locks.emplace_back(locks[lock_start], std::defer_lock);
         }
+        boost::lock(unique_locks.begin(), unique_locks.end());
 
         auto perms = cached_pages[page].Perm();
         u64 range_begin = 0;
@@ -251,11 +255,6 @@ struct PageManager::Impl {
 
         // Add pending (un)protect action
         release_pending();
-
-        // Unlock all locks
-        for (size_t i = lock_start; i < lock_end; ++i) {
-            locks[i].unlock();
-        }
     }
 
     template <bool track>
--- a/src/video_core/page_manager.cpp
+++ b/src/video_core/page_manager.cpp
@@ -59,6 +59,7 @@ struct PageManager::Impl {
 
     static constexpr size_t ADDRESS_BITS = 40;
     static constexpr size_t NUM_ADDRESS_PAGES = 1ULL << (40 - PAGE_BITS);
+    static constexpr size_t NUM_ADDRESS_LOCKS = NUM_ADDRESS_PAGES / PAGES_PER_LOCK;
     inline static Vulkan::Rasterizer* rasterizer;
 #ifdef ENABLE_USERFAULTFD
     Impl(Vulkan::Rasterizer* rasterizer_) {
@@ -189,11 +190,20 @@ struct PageManager::Impl {
     template <bool track>
     void UpdatePageWatchers(VAddr addr, u64 size) {
         RENDERER_TRACE;
-
+        
         size_t page = addr >> PAGE_BITS;
+        const u64 page_end = Common::DivCeil(addr + size, PAGE_SIZE);
+        
+        const size_t lock_start = page / PAGES_PER_LOCK;
+        const size_t lock_end = Common::DivCeil(page_end, PAGES_PER_LOCK);
+        for (size_t i = lock_start; i < lock_end; ++i) {
+            locks[i].lock();
+        }
+
         auto perms = cached_pages[page].Perm();
         u64 range_begin = 0;
         u64 range_bytes = 0;
+        u64 potential_range_bytes = 0;
 
         const auto release_pending = [&] {
             if (range_bytes > 0) {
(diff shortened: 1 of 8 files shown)


Missake212:
No regressions on my end, but also no changes at all, didn't notice better or worse performance and all my games behave the same.

src/video_core/page_manager.cpp
```diff
@@ -191,23 +192,30 @@ struct PageManager::Impl {
         RENDERER_TRACE;
 
         size_t page = addr >> PAGE_BITS;
+        const u64 page_end = Common::DivCeil(addr + size, PAGE_SIZE);
+
+        const size_t lock_start = page / PAGES_PER_LOCK;
+        const size_t lock_end = Common::DivCeil(page_end, PAGES_PER_LOCK);
+        for (size_t i = lock_start; i < lock_end; ++i) {
```
raphaelthegreat: ~~Memory protect ranges are split on region manager boundaries, so is it necessary to lock more than 1 mutex?~~
EDIT: I think it is as texture cache uses this API directly

Locking in a loop like that might cause deadlocks, consider using boost::lock which accepts an iterator range as that will avoid deadlocks https://www.boost.org/doc/libs/latest/doc/html/thread/synchronization.html#thread.synchronization.lock_functions.lock_range

src/video_core/page_manager.cpp
```diff
@@ -284,19 +304,24 @@ struct PageManager::Impl {
                 release_pending();
                 perms = new_perms;
             } else if (range_bytes != 0) {
-                // If the protection did not change, extend the current range
-                range_bytes += PAGE_SIZE;
+                // If the protection did not change, extend the potential range
+                potential_range_bytes += PAGE_SIZE;
```
raphaelthegreat, line 303: I think the logic in this functions can be simplified quite a bit by just using the page permission instead of the counter, something like the following (untested but looks like it could work). I believe this also accomplishes the goal of the new potential_range_bytes variable

```cpp
auto perms = cached_pages[base_page + start_range.first].Perm();
size_t page = start_range.first;
u64 range_begin = page;
u64 range_bytes = 0;
bool update_range = false;

const auto release_pending = [&] {
    if (range_bytes > 0 && update_range) {
        RENDERER_TRACE;
        Protect(range_begin << PAGE_BITS, range_bytes, perms);
    }
    range_begin = page;
    range_bytes = 0;
    update_range = false;
};

// Iterate pages
for (; page < end_range.second; ++page) {
    PageState& state = cached_pages[base_page + page];
    const bool update = mask.Get(page);
    
    // Apply the change to the page state
    if (update) {
        state.AddDelta<track ? 1 : -1>();
    }

    // If the protection changed add pending (un)protect action
    if (auto new_perms = state.Perm(); new_perms != perms) [[unlikely]] {
        release_pending();
        perms = new_perms;
    }

    // Extend the current range
    range_bytes += PAGE_SIZE;
    update_range |= update;
}

// Add pending (un)protect action
release_pending();
```

src/video_core/page_manager.cpp
```diff
@@ -256,21 +274,23 @@ struct PageManager::Impl {
...
                 Protect((range_begin << PAGE_BITS), range_bytes, perms);
                 range_bytes = 0;
+                potential_range_bytes = 0;
             }
         };
 
-        std::scoped_lock lk(lock);
-
         // Iterate pages
         for (size_t page = start_range.first; page < end_range.second; ++page) {
```
raphaelthegreat, line 290: start_range  and end_range appear to be used as a general start and end boundary of the update and are computed twice (one time in region manager and one here). I think in region manager you could use iterators on the mask to get start and end, and then pass the start and end pages in this function to avoid recomputing them twice

src/video_core/page_manager.cpp
LNDF: Yes, the reason that more than one lock is needed is because calls from texture cache could span more than one  lock region.

src/video_core/page_manager.cpp
LNDF: The reason for computing the first and last range is the following:

First range returns the start and end of the first sequence of coniguous 1s in the bitarray starting from the beginning.
Last range returns the start and end of the first sequence of contiuous 1s in the bitarray starting from the end (so the last range).
Then we check if both ranges are the same. It means that only exists one contiuous range, so we can use the normal update function on that range. Maybe this optimization is not necesary or has more negatives than positives.

If we pass the start and end of the requested range, we don't know if the update mask (untracked ^ cpu) is one contiguous region or not.

LNDF:
Oh. It pulled boost from my system (that's why it compiled for me localy) install. Will need to add threading boost lib or find another way...

src/video_core/page_manager.cpp
LNDF: I will explain what the logic behind the current implementation is:

This is how it works on current main:
We start with no range defined (range_bytes = 0). Then, when we find a page who's state has changed (the conditional at the end where we check new delta's value).
If we have a range, we keep extending it until we find a page with different permissions (this will happen if a page doesn't require changing state), at which point to perform the (un)protect and reset range state.

However, this has a problem that the potential range bytes try to solve:
Let's put a situation where we want to unprotect from page 0 to 15, but only page 3 is protected, but only have page 0, 3 and 15 on the mask; and page 0 and 15 have 2 watchers so they aren't unprotected.. When the algorithm reaches page 3 it will start a new range. Then, the range will be extended until page 14 because those pages have the same permission as page 3. When reaching page 15 it will have a different permission so the pending range will be released and the range reset. That would result in an mprotect from page 3 to 14 (including).

The potential range fixes that. Insteed of extending the actual range, we extend a potential range and only updating the range when a page requires changing the state.

Continuing with the previous example the potential range would be extended starting at page 3. Then in page 15, when releasing the pending range, only page 3 would be unprotected because the actual range has not been updated with the potential range.

Regarding your implementation, I think you are incorrectly initializing the range at the first page (range_begin = page, which may not be true on every circumstance). Also, you are not considering the case where a page needs to be updated (protected) but there is already a watcher (so it is not necessary to protect it again.

Maybe I'm understanding something wrong, if it is the case please pont me out.

raphaelthegreat: You mention "only page 3 is protected" and also "page 0 and 15 have 2 watchers so they aren't unprotected" which implies page 0 and 15 are also protected, so I don't quite understand the initial state of the problem

raphaelthegreat: So if we assume only page 3 is protected, the problem is that the mprotect will go from page 3 to 14 instead of just mprotecting page 3, is that correct?

raphaelthegreat: Also can you specify a case where range_begin = page is incorrect?

LNDF: Yes. That's is the thing I try to fix on this PR.

The example was:
Requested pages to update: 0, 3 and 15 (so from 0 to 15)
Page 0 is protected and has 2 watchers
Page 3 is protected and has 1 watcher
Page 15 is protected and has 2 watchers.
All other pages are unprotected.

Current main mprotects from 3 to 14, this PR only mprotects 3.

LNDF: If you set range_begin to page you are assuming that the first page needs changing protection. And that may not be always true (like in the example I gave, where page 0 has 2 watchers).

raphaelthegreat: I see, thanks for the explanation

PR #3158, opened 2025-06-25
Changed symbol bindings to their names

nickci2002:
Replaced ',' with 'comma', '.' with 'period', etc. due to binding errors. Also updated kbm_help_dialog.cpp to show symbols with their names.


src/qt_gui/kbm_help_dialog.cpp
```diff
@@ -228,7 +198,41 @@ The following names should be interpreted without the '' around them. For inputs
...
+    ( 'lparen'
+    ) 'rparen'
+    - 'minus'
+    _ 'underscore'
+    = 'equals'
+    + 'plus'
+    [ 'lbracket'
+    ] 'rbracket'
+    { 'lbrace'
+    } 'rbrace'
```
kalaposfos13: For all the different parentheses it's enough to only show the left version, the others can be inferred from that, as per line 152

src/qt_gui/kbm_help_dialog.cpp
```diff
@@ -267,3 +271,35 @@ You can find these here, with detailed comments, examples, and suggestions for m
 'mouse_gyro_roll_mode':
     Controls whether moving the mouse sideways causes a panning or a rolling motion while mouse-to-gyro emulation is active.)";
 }
+
+QString HelpDialog::faq() {
```
kalaposfos13, line 272:
```suggestion
constexpr QString HelpDialog::faq() {
```
Same for the other functions too

kalaposfos13 (changes requested):
Duplicate of #3157 but honestly your documentation is better and I also forgot the GUI bindings so I guess we roll with this one

src/qt_gui/kbm_help_dialog.cpp
nickci2002: I can't add constexpr because QString isn't a literal data type.

kalaposfos13: No, you couldn't add constexpr because you forgot to update the header

kalaposfos13 (approved):

PR #3160, opened 2025-06-26
WIP: Shader Cache

Fire-Cube:
This is currently still unfinished and there may be crashes and graphical errors. The caching system is currently kept as simple as possible but when everything works I will make sure that we only have one or two files for saving


LNDF:
From the discord: 
>  just as an idea, shouldn't shader caches be separated by game (so they can be easily deleted or transferred?
> Also, maybe it is a good idea to combine shader of a game into a single file for sharing proposes and dump individual cached shaders if shader dumping is enabled (idk if sharing caches would be ok. After all, it's just spirv with metadata and afaik sharing spirv is ok)

georgemoralis:
Just a quote here : sharing cache shaders is also copyright material and it is not allowed

src/common/binary_helper.h
```diff
@@ -0,0 +1,21 @@
+#pragma once 
+
+#include <iostream>
+#include <vector>  
+#include "common/logging/log.h"
+
+using u32 = uint32_t;
+
+template <typename T>  
+void writeBin(std::ostream& os, const T& v) {
```
LNDF: You are not following the name convenient here. It should be WriteFile

src/common/binary_helper.h
```diff
@@ -0,0 +1,21 @@
...
+    LOG_INFO(Render_Recompiler, "BinaryHelper: Pos: {}", static_cast<int64_t>(os.tellp()));
+    os.write(reinterpret_cast<const char*>(&v), sizeof(T));  
+}  
+
+template <typename T>  
+void readBin(std::istream& is, T& v) {
```
LNDF: Same

src/common/hash.h
```diff
@@ -5,10 +5,7 @@
...
 #include "common/types.h"
 
-[[nodiscard]] inline u64 HashCombine(const u64 seed, const u64 hash) {
-    return seed ^ (hash + 0x9e3779b9 + (seed << 12) + (seed >> 4));
-}
-
-[[nodiscard]] inline u32 HashCombine(const u32 seed, const u32 hash) {
-    return seed ^ (hash + 0x9e3779b9 + (seed << 6) + (seed >> 2));
+template <typename T, typename U>
+T HashCombine(const T& seed, const U& value) {
```
LNDF: Caution. You can see how the 32 but and 64 bit versions of the hash are different (see the bit shifts).

I don't know if this affects the quality of the hash. You should check this.

src/video_core/renderer_vulkan/shader_cache.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
```
LNDF, line 4: You forgot the SPDX notice here (shadPS4 2025 copyright)

src/common/binary_helper.h
```diff
@@ -0,0 +1,21 @@
...
+    if (is.eof()) {
+        LOG_WARNING(Render_Recompiler, "BinaryHelper: EOF");
```
LNDF: Consider changing the log class if this is a common utility

src/video_core/renderer_vulkan/shader_cache.h
```diff
@@ -0,0 +1,21 @@
...
+u64 CalculateSpecializationHash(const Shader::StageSpecialization& spec);
+void SerializeInfo(
+    std::ostream& info_serialized, Shader::Info info);
+void DeserializeInfo(std::istream& info_serialized, Shader::Info& info);
+
+bool CheckShaderCache(std::string shader_id);
+void GetShader(std::string shader_id, Shader::Info& info, std::vector<u32>& spv);
+void AddShader(std::string shader_id, std::vector<u32> spv, std::ostream& info_serialized);
+
+} // namespace ShaderCache
```
LNDF: Maybe  instead of managing this in a namespace, use a class. Pass the id of the name in the constructor so it knows where to save the cache.

I would create an instance in PipelineCache (since is there where it's used.

src/video_core/renderer_vulkan/shader_cache.cpp
```diff
@@ -0,0 +1,798 @@
+#include <iostream>
```
LNDF: SPDX here

src/video_core/renderer_vulkan/shader_cache.cpp
```diff
@@ -0,0 +1,798 @@
...
+
+using u64 = uint64_t;
+using u32 = uint32_t;
+
+namespace ShaderCache {
+
+const auto shader_cache_dir = Common::FS::GetUserPath(Common::FS::PathType::ShaderDir) / "cache";
+std::unordered_map<std::string, std::vector<u32>> g_ud_storage;
+
+u64 CalculateSpecializationHash(const Shader::StageSpecialization& spec) {
```
LNDF, line 24: Maybe specialize std::hash instead of calculating here. Then you could store the specialization info itself as a key in the map (which would handle collisions)

src/video_core/renderer_vulkan/shader_cache.cpp
```diff
@@ -0,0 +1,798 @@
...
+
+        hash = HashCombine(hash, spec.info->ud_mask.mask);
+
+        hash = HashCombine(hash, spec.info->uses_patches);
+    }
+
+    return hash;
+}
+
+void SerializeInfo(std::ostream& info_serialized, Shader::Info info) {
```
LNDF: I suggest using a serialization library. That will make things a lot easier.

You can manage versioning, serialize maps automatically and manage file layout automatically.

src/video_core/renderer_vulkan/shader_cache.cpp
```diff
@@ -0,0 +1,798 @@
...
+    readBin(info_serialized, info.pgm_base);
+
+
+    // Check if there are any remaining bytes in the stream
+    if (info_serialized.peek() != EOF) {
+        LOG_WARNING(Render_Vulkan, "There are remaining bytes in the cache file");
+    }
+}
+
+bool CheckShaderCache(std::string shader_id) {
```
LNDF: I think it is better to store the entire shader cache into a single file.

I would use a map with a key consisting of the shader id and specialization, and value being the info and the spirv blob.

Then load the entire cache into memory when the game starts. This is easily achievable with a serialization library.

Maybe in the future, we could cache entire pipeline states, as well as use a separate cache for the vulkan compiled pipelines themselves.

StevenMiller123:
Assassin's Creed® Unity crashes on a non-GPU memory assert when running with cached shaders.
```
[Debug] <Critical> page_manager.cpp:215 operator(): Assertion Failed!
Attempted to track non-GPU memory at address 0x103fa50000, size 0x7f000.
```
[CUSA00663.log](https://github.com/user-attachments/files/20952895/CUSA00663.log)

BLEACH Rebirth of Souls crashes on an assert when running with cached shaders.
```
[Debug] <Critical> vk_graphics_pipeline.cpp:299 operator(): Assertion Failed!
Failed to create graphics pipeline: ErrorUnknown
```
[CUSA27765.log](https://github.com/user-attachments/files/20952918/CUSA27765.log)

Bloodborne™ crashes on `[Debug] <Critical> info.h:291 operator(): Assertion Failed!` with no cached shaders when loading ingame.
[CUSA00900.log](https://github.com/user-attachments/files/20952942/CUSA00900.log)

Call of Duty®: Advanced Warfare crashes on an assert when running with cached shaders.
```
[Debug] <Critical> vk_graphics_pipeline.cpp:299 operator(): Assertion Failed!
Failed to create graphics pipeline: ErrorUnknown
```
[CUSA00803.log](https://github.com/user-attachments/files/20952955/CUSA00803.log)

For the sake of time, these are all I'll test for right now, seems like a lot of games are suffering from that graphics pipeline crash.

src/video_core/renderer_vulkan/shader_cache.cpp
Fire-Cube: I don't quite understand what you mean

src/video_core/renderer_vulkan/shader_cache.cpp
```diff
@@ -0,0 +1,337 @@
...
+#include "video_core/renderer_vulkan/shader_cache_serialization.h"
+
+#include "shader_cache.h"
+
+using u64 = uint64_t;
+using u32 = uint32_t;
+
+namespace ShaderCache {
+
+const auto shader_cache_dir = Common::FS::GetUserPath(Common::FS::PathType::ShaderDir) / "cache";
```
LNDF: I would save into `shader/cache/portable`. In the future we might want to cache the vulkan pipelines (which are driver/platform specific).

Portable because the cache you are saving can be transfered to another computer/OS/driver version.

src/video_core/renderer_vulkan/shader_cache_serialization.h
```diff
@@ -0,0 +1,92 @@
...
+#include <cereal/types/variant.hpp>
+#include <cereal/types/utility.hpp>
+#include "shader_recompiler/info.h"
+
+namespace cereal {
+
+// boost::small_vector
+template<class Archive, class T, std::size_t N, class Alloc>
+void save(Archive& ar, boost::container::small_vector<T, N, Alloc> const& smallVector) {
+    ar(static_cast<std::uint32_t>(smallVector.size()));
```
LNDF: Maybe it's better to use `cereal::make_size_tag`. I don't know if it makes any difference, but cereal itself uses it.
https://github.com/USCiLab/cereal/blob/a56bad8bbb770ee266e930c95d37fff2a5be7fea/include/cereal/types/vector.hpp#L67

Also, `small_vector` is a common data structure used in the emulator. It may be a good idea to put those under something like `common/serialization`

src/video_core/renderer_vulkan/shader_cache_serialization.h
```diff
@@ -0,0 +1,92 @@
...
+    ar(elementCount);
+    smallVector.resize(elementCount);
+    for (auto& element : smallVector)
+        ar(element);
+}
+
+// Shader::Info::UserDataMask
+template<class Archive>
+void serialize(Archive& ar, Shader::Info::UserDataMask& mask) {
+    ar(mask.mask);
```
LNDF, line 20: This is opinionated, and I don't either know what the best design is.

But you can declare the member function serialize/save/load inside of the class directly and it will automaticly be serializable (I'm not 100% into it either, but just in case you didn't know).

https://uscilab.github.io/cereal/serialization_functions.html

src/video_core/renderer_vulkan/shader_cache.cpp
LNDF: If I understand correctly, this function takes an `StageSpecialization` and outputs the hash. I think it is better to specialize `std::hash<Shader::StageSpecialization>` insteed.

src/common/hash.h
Fire-Cube: I’ve now adjusted it so that it always shifts like the 64-bit version that shouldn’t really cause any issues.

src/video_core/renderer_vulkan/shader_cache.h
```diff
@@ -0,0 +1,39 @@
...
+#include "shader_recompiler/info.h"
+#include "shader_recompiler/specialization.h"
+
+namespace ShaderCache {
+
+#define SHADER_CACHE_DIR                                                                           \
+    (Common::FS::GetUserPath(Common::FS::PathType::ShaderDir) / "cache" / "portable")
+
+#define SHADER_CACHE_BLOB_PATH                                                                     \
+    (SHADER_CACHE_DIR / (std::string{Common::ElfInfo::Instance().GameSerial()} + "_shaders.bin"))
```
LNDF, line 20: Don't use macros like that. It seems confusing while reading the code and it looks like shaders are not game separated.

src/video_core/renderer_vulkan/shader_cache.cpp
```diff
@@ -0,0 +1,398 @@
...
+#include "common/io_file.h"
+#include "common/logging/log.h"
+#include "common/path_util.h"
+#include "shader_cache.h"
+#include "shader_recompiler/info.h"
+#include "shader_recompiler/ir/type.h"
+#include "shader_recompiler/specialization.h"
+#include "video_core/renderer_vulkan/shader_cache_serialization.h"
+
+namespace ShaderCache {
```
LNDF, line 22: It is a lot better now. But I think there are some key issues. I will first describe them and explain a possible way to solve them later.
* You are managing offsets to shader data manualy. That should be the job of the serialization library.
* Shader cache being in a namespace doesn't follow the style of ofther components of this emulator, like the PipelineCache.

I can see that the shader cache is a dependency of PipelineCache. So I would do the following:

This will be all pseudocode.
```c++
struct ShaderCache {
    struct ShaderCacheKey {
        u32 pgm_hash;
        StageSpecialization spec; // You specialize std::hash<StageSpecialization>() for convinience (you already have this as CalculateSpecializationHash, you just convert it to the std::hash<> specialization.
    // You also specialize std::hash<ShaderCacheKey>() so you can use in std::unordered_map as the key. You just CombineHash() both.
    };

    struct ShaderCacheEntry {
        Info info; // Also, define serialization for this
        std::vector<u32> spv;

        // Define serialize for this (archive both).
    };

    std::unordered_map<ShaderCacheKey, ShaderCacheEntry> cache_entries;
    // Maybe it is interesting to save some more info about the game in order to invalidate cache?
    std::string game_id; // IDK if the type is correct (in general, IDK if types are correct in this example)

    std::filesystem::path cache_path; // You don't need to serialize this.

    // Define serialization/deserialization for this class. Archive all members besides shader path (dynamicly generated).

    ShaderCache(const std::string& game_id) {
        // Here, you calculate the cache path and deserialize itself. like ar(*this, ...);
        // After deserializing, validate whether the game id is equal. Also, cereal includes a versioning system for serialization
        // If you use the versioning here, you could invalidate the cache if the version is different.
    }

    // Define the AddShader/GetShader/IsShaderCached functions passing the shader key as argument.
    // It should be as simple as interacting with the std::unordered_map

    // You also want a strategy to know when to save the cache to disk (I/O is slow and serialzing also is slow). Maybe every N shaders? Or when emu closes?
};
```

As I mentioned before, this is a dependency of PipelineCache. So I would pass instantiate the ShaderCache as a member of the PipelineCache.

If you have any problem implementing something like this, reach out on Discord.

src/common/hash.h
```diff
@@ -5,10 +5,10 @@
 
 #include "common/types.h"
 
-[[nodiscard]] inline u64 HashCombine(const u64 seed, const u64 hash) {
-    return seed ^ (hash + 0x9e3779b9 + (seed << 12) + (seed >> 4));
-}
+template <typename T1, typename T2>
+[[nodiscard]] constexpr u64 HashCombine(T1 seed, T2 hash) noexcept {
```
LNDF, line 9: Still not sure about this...
You can see how the output hash is also different with.  Maybe it is a better way to do this (I susptect you are wanting to combine 32 and 64 bit types?)

src/video_core/renderer_vulkan/shader_cache.cpp
```diff
@@ -0,0 +1,398 @@
...
+    std::string info_blob = info_serialized.str();
+    blob_ar(spv, info_blob);
+
+    shader_registry[shader_id] = offset;
+    if (Config::getShaderCachePreloadEnabled()) {
+        shader_cache[shader_id] = std::make_pair(spv, info_blob);
+    }
+}
+
+void SerializeInfo(std::ostream& info_serialized, Shader::Info& info) {
```
LNDF, line 312: Is there a reason to serialize like this insteed of using the serializer?

src/video_core/renderer_vulkan/shader_cache.cpp
Fire-Cube: I understand that the separate offsets aren’t ideal, but in my opinion, we have to rely on them.
Cereal doesn’t support partial loading, which wouldn’t be practical for on-demand mode.
I believe we’d run into a similar issue when saving.

LNDF: Gow big is a shader cache? I think the advantage of a shader cache is to pre load all of them before starting the game. That way you could even compile the pipelines so there are no stutters (yes, that would take longer to load but afaik other emus do that)

In the future we could even cache the compiled pipelines so the games could load quick even with shader cache.

Fire-Cube: Dragon Quest uses around 17 MB just from a single boot. I’m not sure how much bigger it gets during active gameplay, since the game is affected by the UE rendering issue

LNDF: Imo that is not a big size. But maybe it is better to have someone else's opinion on this. Maybe it is good idea to ask on discord.

georgemoralis:
outdated and not properly working anyway , if something better appears repopen it :D

PR #3162, opened 2025-06-27
video_core: Calculate anisotrophic ratio

StevenMiller123:
Some games use higher max_aniso values than what our enum provides, this PR swaps the enum for a simple formula to calculate the appropriate anisotrophic filtering ratio.

This fixes cases of `[Debug] <Critical> resource.h:519 MaxAniso: Unreachable code!`, and improves NEO : The World Ends with You (CUSA26376).

### DIFF
--- a/src/video_core/amdgpu/resource.h
+++ b/src/video_core/amdgpu/resource.h
@@ -389,14 +389,6 @@ enum class ClampMode : u64 {
     MirrorOnceBorder = 7,
 };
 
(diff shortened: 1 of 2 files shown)


squidbus:
I don't think this is actually correct, as far as I know the AMD GPU only supports up to 16x. If you're seeing invalid values and its not some other issue, we need to understand how real hardware treats them.

PR #3167, opened 2025-06-28
Avoid clearing depth on partial HTILE writes

raphaelthegreat:
This fixes the shadows on Driveclub (CUSA00093)

| main  | PR |
| ------------- | ------------- |
| bad_shadows3 | proper_shadows4 |

There is a shader that performs some bit operations on HTILE, but not a full clear which got mis-recognized as a clear. According to [mesa](https://github.com/chaotic-cx/mesa-mirror/blob/646977348ba49630515f0da1542b77ef4919d68e/src/amd/vulkan/meta/radv_meta_clear.c#L661) this fetches the upper bit of SR1 and clears/sets zmask depending on that bit. I'm not sure what it is, but doesn't appear to be a depth or stencil clear as it leaves zmin/zmax and sr0/sr1 untouched.

```glsl
htile | ((1u & (~bitfieldExtract(htile, 7, 1))) * 15))
```

### DIFF
--- a/src/shader_recompiler/info.h
+++ b/src/shader_recompiler/info.h
@@ -58,6 +58,7 @@ struct BufferResource {
     BufferType buffer_type;
     u8 instance_attrib{};
     bool is_written{};
+    bool is_read{};
     bool is_formatted{};
 
     bool IsSpecial() const noexcept {
--- a/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
+++ b/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
@@ -413,11 +413,12 @@ void PatchImageSharp(IR::Block& block, IR::Inst& inst, Info& info, Descriptors& 
     // Read image sharp.
     const auto tsharp = TrackSharp(tsharp_handle, info);
     const auto inst_info = inst.Flags<IR::TextureInstInfo>();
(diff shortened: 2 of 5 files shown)


raphaelthegreat:
Note, shadows probably only work on AMD for now because NVIDIA doesn't support R8Srgb as color attachment (which spams in the log)

GHU7924:
>Note, shadows probably only work on AMD for now because NVIDIA doesn't support R8Srgb as color attachment (which spams in the log)

I have Nvidia. I noticed this problem in Bloodborne after this PR change. 

Before

BB_Old_Hunters

After

https://github.com/user-attachments/assets/f02d2a28-0c35-4222-90aa-d89023e5e400

Can anyone else check this? This is The Old Hunters.

Missake212:
Can confirm I'm having the same issue, tested the build before this one got merged and it looks fine.

MajorP93:
@georgemoralis why did this get merged before testing was done? Been wondering this for a few merges in the last few weeks…

georgemoralis:
> @georgemoralis why did this get merged before testing was done? Been wondering this for a few merges in the last few weeks…

it is normal to have regressions during WIP versions . Only a small amount of testers , do test prs , so merging and testing in main is actually more interactive

PR #3168, opened 2025-06-29
Qt: Directly install update after downloading

Xphalnos:
This allows you to install the update directly after downloading it.
This dialog box is useless since there is nothing else you can do except OK.

image


georgemoralis:
don't see a big harm to have a notify window ... sorry but closing

PR #3170, opened 2025-06-29
removing "Install" from game directory code

diegolix29:
removes all install words from game directory cause we arent installing games anymore


src/qt_gui/game_install_dialog.h
kalaposfos13, line 1: I think you meant to delete this file and its source file, they're not included in CMakeLists anymore and are just copies of the new game_install_dialog files

diegolix29: yes it should be fixed now, as i push that one blindly it seems.

diegolix29:
this is how it will look no more "directory to install games" 
image

PR #3171, opened 2025-06-29
Misc fixes

mikusp:
Explanation inline with code changes

### DIFF
--- a/src/shader_recompiler/frontend/copy_shader.cpp
+++ b/src/shader_recompiler/frontend/copy_shader.cpp
@@ -14,7 +14,7 @@ CopyShaderData ParseCopyShader(std::span<const u32> code) {
     constexpr u32 token_mov_vcchi = 0xBEEB03FF;
     ASSERT_MSG(code[0] == token_mov_vcchi, "First instruction is not s_mov_b32 vcc_hi, #imm");
 
-    std::array<s32, 32> offsets{};
+    std::array<s32, 64> offsets{};
     offsets.fill(-1);
 
     std::array<s32, 256> sources{};
@@ -52,6 +52,8 @@ CopyShaderData ParseCopyShader(std::span<const u32> code) {
             break;
         }
         case Gcn::Opcode::BUFFER_LOAD_DWORD: {
+            ASSERT_MSG(inst.src[1].code < offsets.size(),
+                       "offsets array for geometry shaders is too short");
             offsets[inst.src[1].code] = inst.control.mubuf.offset;
             if (inst.src[3].field != Gcn::OperandField::ConstZero) {
                 const u32 index = inst.src[3].code;
@@ -65,7 +67,7 @@ CopyShaderData ParseCopyShader(std::span<const u32> code) {
         }
     }
 
-    if (last_attr != IR::Attribute::Position0) {
+    if (!IsPosition(last_attr)) {
         data.num_attrs = static_cast<u32>(last_attr) - static_cast<u32>(IR::Attribute::Param0) + 1;
         const auto it = data.attr_map.begin();
         const u32 comp_stride = std::next(it)->first - it->first;
--- a/src/video_core/amdgpu/pm4_cmds.h
+++ b/src/video_core/amdgpu/pm4_cmds.h
@@ -699,7 +699,7 @@ struct PM4CmdWaitRegMem {
 struct PM4CmdWriteData {
     PM4Type3Header header;
     union {
-        BitField<8, 11, u32> dst_sel;
+        BitField<8, 4, u32> dst_sel;
         BitField<16, 1, u32> wr_one_addr;
         BitField<20, 1, u32> wr_confirm;
         BitField<30, 1, u32> engine_sel;


src/shader_recompiler/frontend/copy_shader.cpp
```diff
@@ -52,6 +52,8 @@ CopyShaderData ParseCopyShader(std::span<const u32> code) {
             break;
         }
         case Gcn::Opcode::BUFFER_LOAD_DWORD: {
+            ASSERT_MSG(inst.src[1].code < offsets.size(),
+                       "offsets array for geometry shaders is too short");
```
mikusp, line 56: Prevents a libcxx assert for accessing std::array out of bounds

src/shader_recompiler/frontend/copy_shader.cpp
```diff
@@ -66,7 +68,10 @@ CopyShaderData ParseCopyShader(std::span<const u32> code) {
     }
 
     if (last_attr != IR::Attribute::Position0) {
-        data.num_attrs = static_cast<u32>(last_attr) - static_cast<u32>(IR::Attribute::Param0) + 1;
+        data.num_attrs =
```
mikusp: When a geometry shader outputs Position1 as the last attritbute, data.num_attrs would underflow and become a very large value

src/shader_recompiler/ir/passes/ring_access_elimination.cpp
```diff
@@ -142,14 +142,18 @@ void RingAccessElimination(const IR::Program& program, const RuntimeInfo& runtim
...
-                ASSERT(it != info.gs_copy_data.attr_map.cend());
-                const auto& [attr, comp] = it->second;
-
-                inst.ReplaceOpcode(IR::Opcode::SetAttribute);
-                inst.ClearArgs();
-                inst.SetArg(0, IR::Value{attr});
-                inst.SetArg(1, data);
-                inst.SetArg(2, ir.Imm32(comp));
+                if (it == info.gs_copy_data.attr_map.cend()) {
+                    LOG_ERROR(Render_Recompiler, "attr_map missing entry: {}", vc_read_ofs);
```
mikusp: Killzone seems to access a value that has not been set in the previous stage, triggering that assert - looks like a game error but I'm not 100% sure

src/video_core/amdgpu/pm4_cmds.h
```diff
@@ -699,7 +699,7 @@ struct PM4CmdWaitRegMem {
 struct PM4CmdWriteData {
     PM4Type3Header header;
     union {
-        BitField<8, 11, u32> dst_sel;
+        BitField<8, 4, u32> dst_sel;
```
mikusp, line 702: Probably a typo, documentation says dst_sel occupies bits 8:11

Missake212:
Matterfall (CUSA02475) now goes further (used to crash when compiling a geometry shader) performance is horrendous (unreal engine game), gets to the main menu while not showing the first intro video (black screen) and main menu not really displaying everything it should, crashes while trying to get in game on what seems to be a memory error.

``[Kernel.Vmm] <Error> memory.cpp:943 SearchFree: Couldn't find a free mapping for address 0x1000000000, size 0xfffffffc66f5c000``

image
image
Screenshot 2025-06-29 232529


[CUSA02475.log](https://github.com/user-attachments/files/20970421/CUSA02475.log)

src/shader_recompiler/frontend/copy_shader.cpp
squidbus: Shouldn't this just be a change to the `if` condition above?

src/shader_recompiler/ir/passes/ring_access_elimination.cpp
squidbus: I'm not so sure about this one, it seems like the assert is purposeful here and this would need a deeper investigation.

src/video_core/amdgpu/pm4_cmds.h
red-prig: sure

brad0demx:
Using this PR, [Lies of P](https://github.com/shadps4-emu/shadps4-game-compatibility/issues/87) progresses further than main, where it crashes after compiling gs shaders and printing warnings regarding pos1 exports. 
It progresses through the menus to the loading screen, where it crashes before reaching 1% with 
```
[Debug] <Critical> texture_cache.cpp:259 ResolveOverlap: Unreachable code!
```

image
[CUSA36848.log](https://github.com/user-attachments/files/20991713/CUSA36848.log)

georgemoralis:
@mikusp can you check squidbus reviews?

src/shader_recompiler/frontend/copy_shader.cpp
mikusp: I don't think so, seems like num_attrs is the number of params exported and those lines below should be executed whenever the copy shader exports more than one position attribute

squidbus: It looks to me like the code is for when it exports param attributes. Otherwise why subtract from Param0

squidbus: @raphaelthegreat Can you check this

raphaelthegreat: num_attrs variable is for defining input attributes to geoshader. Position0/1 are system values so yeah they should not be included. From that sense, this looks correct to me

squidbus: If we don’t care about Position1 here then I think my point stands then that we can just change the if condition then, no?

raphaelthegreat: Yes it should be changed to !IsPosition(last_attr)

src/shader_recompiler/ir/passes/ring_access_elimination.cpp
mikusp: That is the only comment left, is that a deal-breaker for merging?

squidbus: I think we should probably keep this as it was, maybe do a separate PR if you want. I’m not familiar enough with the geometry shader stuff to know for sure but just removing the instruction seems like a bit of a hack.

squidbus (approved):

georgemoralis:
https://github.com/georgemoralis/shadPS4/actions/runs/16402630340

PR #3173, opened 2025-06-29
Fix CMake presets for Linux

Fire-Cube:


LNDF:
We are not using ninja in ci currently. I see you have added it as the generator. I would add the dependency in the build documentation or switch to make.

EDIT: Just tested, I can build on Linux.

PR #3174, opened 2025-06-29
Update template hyperlink to the wiki instead of the compatibility list

Missake212:
Updating the GAME BUG template hyperlink to the wiki instead of the compatibility list, could also redirect to the README in the emulator's repo instead but I thought redirecting to the Wiki would give users more info.


GHU7924:
The template already has line `Before submitting an issue please check [Game Compatibility Repository](https://github.com/shadps4-emu/shadps4-game-compatibility) for the information about the status of the game.`, so I think changing the modules link to one that points to the Wiki seems like a logical step. 

@Missake212 I fully support this change 👍 .

PR #3176, opened 2025-06-29
fix the fix for the presets

Fire-Cube:
Shame at me and LNDF


georgemoralis:
same on you both . No carrots today

PR #2668, opened 2025-03-22
video_core: Readbacks proof of concept

raphaelthegreat:
Readbacks are complex to get right, even more so to do that fast. At the moment they are implemented in relatively slowish way to test for correctness/data-validity. Before merging performance will be considered

PR #3178, opened 2025-06-30
Readbacks proof of concept rebased

raphaelthegreat:
The same as https://github.com/shadps4-emu/shadPS4/pull/2668 but rebased on new region manager and page manager implementations. Mostly opening so it can be compared with the readbacks-lite approach on correctness. There are also some optimizations I thought of to improve performance here, but will take time

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -51,6 +51,7 @@ static bool isShowSplash = false;
 static std::string isSideTrophy = "right";
 static bool isNullGpu = false;
 static bool shouldCopyGPUBuffers = false;
+static bool readbacksEnabled = false;
 static bool shouldDumpShaders = false;
 static bool shouldPatchShaders = true;
 static u32 vblankDivider = 1;
@@ -240,6 +241,10 @@ bool copyGPUCmdBuffers() {
     return shouldCopyGPUBuffers;
 }
 
+bool readbacks() {
+    return readbacksEnabled;
+}
+
 bool dumpShaders() {
     return shouldDumpShaders;
 }
@@ -342,6 +347,10 @@ void setAllowHDR(bool enable) {
 
 void setCopyGPUCmdBuffers(bool enable) {
     shouldCopyGPUBuffers = enable;
+}
+
+void setReadbacks(bool enable) {
+    readbacksEnabled = enable;
 }
 
 void setDumpShaders(bool enable) {
@@ -586,6 +595,7 @@ void load(const std::filesystem::path& path) {
         screenHeight = toml::find_or<int>(gpu, "screenHeight", screenHeight);
         isNullGpu = toml::find_or<bool>(gpu, "nullGpu", false);
         shouldCopyGPUBuffers = toml::find_or<bool>(gpu, "copyGPUBuffers", false);
+        readbacksEnabled = toml::find_or<bool>(gpu, "readbacks", false);
         shouldDumpShaders = toml::find_or<bool>(gpu, "dumpShaders", false);
         shouldPatchShaders = toml::find_or<bool>(gpu, "patchShaders", true);
         vblankDivider = toml::find_or<int>(gpu, "vblankDivider", 1);
@@ -735,6 +745,7 @@ void save(const std::filesystem::path& path) {
     data["GPU"]["screenHeight"] = screenHeight;
     data["GPU"]["nullGpu"] = isNullGpu;
     data["GPU"]["copyGPUBuffers"] = shouldCopyGPUBuffers;
+    data["GPU"]["readbacks"] = readbacksEnabled;
     data["GPU"]["dumpShaders"] = shouldDumpShaders;
     data["GPU"]["patchShaders"] = shouldPatchShaders;
     data["GPU"]["vblankDivider"] = vblankDivider;
--- a/src/common/config.h
+++ b/src/common/config.h
@@ -45,6 +45,8 @@ bool nullGpu();
 void setNullGpu(bool enable);
 bool copyGPUCmdBuffers();
 void setCopyGPUCmdBuffers(bool enable);
+bool readbacks();
+void setReadbacks(bool enable);
 bool dumpShaders();
 void setDumpShaders(bool enable);
 u32 vblankDiv();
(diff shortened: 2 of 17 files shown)


raphaelthegreat:
@ctrlinsert Can you test if Driveclub behaves better here?

kalaposfos13:
SotC with hacks now renders some 3D
image
[CUSA08809.log](https://github.com/user-attachments/files/20978101/CUSA08809.log)

Missake212:
Tested Bloodborne for a bit, no vertex but performance takes a hit like on the old poc readbacks, I also had the game completely freeze on me but I haven't tested extensively.
[CUSA03173.log](https://github.com/user-attachments/files/20979410/CUSA03173.log)

Driveclub also has those freezes, but in this case the game works after waiting for a bit, also the daytime is fixed on that PR too but hard to finish the first race with all the freezes.
image
[CUSA00003.log](https://github.com/user-attachments/files/20979549/CUSA00003.log)

Can't seem to get to the main menu in GT Sport compared to that readbacks-lite PR (so same thing as main).
EDIT: After testing some more here is the conclusion I came to: on main the game always crashes on game code, on the Readbacks-lite PR it gets past that crash in game code and it crashes on Vector memory while showing the main menu a bit, and for this PR it can either hit the game code crash or the Vector memory crash but it doesn't show the main menu.
[CUSA02168.log](https://github.com/user-attachments/files/20979588/CUSA02168.log)

Star Ocean The Second Story R now hits this critical, which doesn't happen in main but they behave the same anyway (both games hang after displaying boot screen while spamming ime like most unity games).
``[Common.Memory] <Critical> address_space.cpp:313 Protect: Unsupported protection flag combination for address 0x212237000, size 4096, read=false, write=true, execute=false``
[CUSA40652.log](https://github.com/user-attachments/files/20979814/CUSA40652.log)

Uncharted the Nathan Drake Collection now goes to menus but crashes silently after a bit of time, Uncharted 2 launched via the elf file boots but hits a condvar unreachable, and as for 3 it crashes in game code with the main menu screen flickering to black a bit.
image
image
[CUSA02826.log](https://github.com/user-attachments/files/20979866/CUSA02826.log)

Seems like The Last Guardian hits the same address_space critical as Star Ocean but gets in game fine and spear renders. Didn't notice visual issues like on main but could be luck I guess. Climbing doesn't work but it is indeed better than on main, instead of straight up falling you climb for a bit then get pushed up in the air (had to upload the vid on YouTube it was too big).
image
image


https://www.youtube.com/watch?v=VJzrWEZOCYc


[CUSA03745.log](https://github.com/user-attachments/files/20980290/CUSA03745.log)

StevenMiller123:
I can't thoroughly compare everything since the rewind-flush branch is rather outdated, but here are a few comparisons I can give:
Assassin's Creed® Unity (Same graphics improvement seen in rewind-flush is present here):
| Main | rewind-flush | This PR |
-------|--------------|---------|
image | image | image

PowerWash Simulator (Powerwashing works on rewind-flush and here, not on main):
| Main | rewind-flush | This PR |
|------|--------------|---------|
| image | image | image |

Like Missake showed earlier, I'm also able to reproduce Uncharted: The Nathan Drake Collection reaching menus on Windows. The game still crashes at nothing for Windows devices on main and rewind-flush. Menu rendering is broken though.
image

diogoaduartereis:
Was going to post the same as you steven just now. For me when comparing this PR to main and other PRs. It's the only way that Uncharted Boots for me. It goes to the menus and then crashes.
image
image

ctrlinsert:
Driveclub renders car selection properly now without using validation layers on Nvidia and the auto-exposure issues seem to be fixed. The crash when loading the Cayoosh Point, Canada track is fixed as well.

The game still has some graphical issues, namely that during the car start animation in the car select screen the window on the door does not render properly when compared to real hardware. See [here](https://cdn.discordapp.com/attachments/1387506662541103196/1389215145363898519/DRIVECLUBTM_20250630080203.webm?ex=6863ceed&is=68627d6d&hm=d88466ad496eeecc4ac14a2a1ba999093615cde70091f4c389b696f53e046577&) for expected behavior on hardware and [here](https://cdn.discordapp.com/attachments/1387506662541103196/1389215271595675648/2025-06-30_08-04-24.mkv?ex=6863cf0b&is=68627d8b&hm=9077c3014ed30ca2d44ded2f56ab7c9fbf69a36993d3d35162e5796203b74413&) for behavior on emulator. Additionally, particle rendering is still incorrect and tire smoke and water droplets do not render properly (at least on Nvidia). Additionally, shader 0xee1ee51b causes hangs and device losts regularly and consistently.

raphaelthegreat:
I have added a config option to enable/disable readbacks behavior. This is an attempt to allow the PR to get merged in its current state and avoid having it bit-rot again but not affect performance for everyone

georgemoralis:
i think it is ok for merge , what others think?

src/common/config.cpp
```diff
@@ -240,6 +241,10 @@ bool copyGPUCmdBuffers() {
     return shouldCopyGPUBuffers;
 }
 
+bool readbacks() {
+    return readbacksEnabled;
```
StevenMiller123, line 245: If the plan is to treat readbacks as an optional experimental feature, we should be logging whether it is enabled.

StevenMiller123 (changes requested):

src/common/config.cpp
raphaelthegreat: I have added a log

diegolix29:
Cayoosh Point, Canada track hangs if u pick other cars that arent the first one 

windows 10/nvidia 3080

image

diegolix29:
it showed the particles but does have still vertex explosions 
image
 the fps are ok tho also sometimes im crashing when warping but its rare crash

StevenMiller123:
I'm not able to reproduce vertex explosions on my end, though I'm not particularly far in the game.
@diegolix29 please provide a log to confirm you've enabled readbacks proeprly.

diegolix29:
> I'm not able to reproduce vertex explosions on my end, though I'm not particularly far in the game. @diegolix29 please provide a log to confirm you've enabled readbacks proeprly.

deleted as my configuration file wasnt regenerating ok until fresh install of the whole build

StevenMiller123:
This PR regressed KINGDOM HEARTS Ⅲ with an assert that was added.
`[Debug] <Critical> buffer_cache.cpp:314 operator(): Assertion Failed!`
[CUSA12031.log](https://github.com/user-attachments/files/21124431/CUSA12031.log)

PR #3179, opened 2025-07-01
video_core: use shared first mutex (reader priority shared mutex) to resolve deadlock

LNDF:
From the discord:

> Just checked and the lock that I thought was exclusive was already a shared lock. 
> What is currently happening is:
> thread a (gpucommand) locks the mapped_range mutex (shared lock)
> thread b (the mapmemory thread)  waits to lock the mapped_range mutex (exclusive lock)
> thread c (the one invalidating memory) waits to lock the mapped_range mutex (shared lock)
>
> The logic says to me that, if thread a and c have shared locks, they could both acquire the lock. But apparently, if thread b tries to acquire exclusive lock before thread c tries to acquire a shared lock, exclusive lock takes preference and all other shared locks need to wait untile thread b releases the lock (causing a deadlock in this instance). To solve that I will have to implement custom mutex that doesn't have this behaviour.

I can change the name of the mutex if you think is not descriptive enough

@StevenMiller123

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -689,6 +689,7 @@ set(COMMON src/common/logging/backend.cpp
            src/common/recursive_lock.cpp
            src/common/recursive_lock.h
            src/common/sha1.h
+           src/common/shared_first_mutex.h
            src/common/signal_context.h
            src/common/signal_context.cpp
            src/common/singleton.h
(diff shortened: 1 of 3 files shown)


StevenMiller123:
This does appear to stop the deadlocks I reported over on Discord.

georgemoralis:
fixed my regressions good to go

Issue #397, opened 2024-08-10
Configuration file persistence issue


gandalfthewhite19890404:
Configuration file remains same after updates. So you need to remove it regularly to recreate and get all features.

StevenMiller123:
Sould be resolved now that #3181 is merged. If the update doesn't work on your end, let me know and I can reopen this.

PR #3181, opened 2025-07-02
Core: Update config files on startup

StevenMiller123:
This PR cleans up the config code, unifies our default settings, and updates config files on each emulator boot to fill in missing fields.

This fixes a handful of strange default and backup settings, and addresses #397

Opening as a draft since I'm not setup to test GUI-side changes, like the additional defaults added to `setDefaultValues()`


gandalfthewhite19890404:
`If the config doesn't have the right number of entries, then regenerate it.`
Whole cfg gonna be regenerated or only new entries gonna be added? I'm against whole regeneration

StevenMiller123:
@gandalfthewhite19890404 I apologize for the poor phrasing.
From the user's perspective, all that changes is that new settings appear. I've made sure that none of your already modified settings will change when the config updates.
While the config is technically "regenerated", all previous settings are loaded first, and the loaded settings are used when regenerating.
If you're not confident, you can always copy your config before testing my changes.

StevenMiller123:
Rebased with the new DMA toggle.

Silver-Skyy:
sorry to ask it like this here, i wanted to know if this will make a new config file or just put the new lines of config in the existing one?

StevenMiller123:
@Silver-Skyy 
The new config lines will appear in the existing config, all previously changed settings will remain unchanged.
If this doesn't work properly, please let me know.

Missake212:
Tried it out by downloading the first build from your fork (on this branch), changing my settings both in the GUI and the config then downloading the latest build and replacing only the .exe, it seems like it behaves as expected for me, would like to have at least one other person testing to make sure still.

One little nitpick however that has nothing to do with your PR since it happens on main (I can open a new issue for this one if needed) but there's some issues with the way the emulator's language setting works, hard to put it into words so here's a quick video:

https://github.com/user-attachments/assets/d0e0cf40-6aa3-486e-98e2-c489efa54d49

StevenMiller123:
@Missake212 
I'm not sure where the GUI language is stored, could you open an issue so someone familiar with the Qt code can take a look?

Since this works as expected, I'll swap this to ready for review.

Randomuser8219:
This PR always says:
Outdated config detected, updating config file.
on startup.

StevenMiller123:
I'm aware of the issue, my PR failed to account for configs with options that don't exist on main.
For now, the issue is harmless since the config update is non-destructive, and the time it takes to update the config is negligible.

PR #3184, opened 2025-07-02
vector_alu: Improve handling of mbcnt append/consume patterns

raphaelthegreat:
This fixes missing grass in Driveclub (CUSA00093)
The existing implementation was written to handle a single pattern of mbcnt before the DS_APPEND instruction

```
v_mbcnt_hi_u32_b32 vX, exec_hi, 0
v_mbcnt_lo_u32_b32 vX, exec_lo, vX
ds_append       vY offset:4 gds
v_add_i32       vX, vcc, vY, vX
```

In this case however the DS_APPEND is before the mbcnt pattern (but is same functionality wise as above)
```
ds_append       vX gds
v_mbcnt_hi_u32_b32 vY, exec_hi, vX
v_mbcnt_lo_u32_b32 vZ, exec_lo, vY
```

The mbcnt instructions are always in pairs of hi/lo and in general are quite flexible. But they assume the subgroup size is 64 so they are not recompiled literally. Together with DS_APPEND they are used to derive a unique per thread index in a buffer (different from using thread_id as order could be random). DS_APPEND instruction works on per subgroup level, by adding number of active threads of subgroup to the GDS counter, essentially giving a multiple-of-64 base index to all threads. Then each thread executes the mbcnt pair which returns the number of active threads with id less than itself and adds it with the base.

The recompiler translates DS_APPEND into an atomic increment of a storage buffer counter, which already gives the desired unique index, so this pattern is a no-op. On main it was set to zero as per the first pattern to avoid altering the DS_APPEND result. The new handling passes through the initial value of the pattern instead, which has the same effect but works on either case.

### DIFF
--- a/src/shader_recompiler/frontend/translate/vector_alu.cpp
+++ b/src/shader_recompiler/frontend/translate/vector_alu.cpp
@@ -558,27 +558,31 @@ void Translator::V_BCNT_U32_B32(const GcnInst& inst) {
 
 void Translator::V_MBCNT_U32_B32(bool is_low, const GcnInst& inst) {
     if (!is_low) {
-        // v_mbcnt_hi_u32_b32 v2, -1, 0
+        // v_mbcnt_hi_u32_b32 vX, -1, 0
         if (inst.src[0].field == OperandField::SignedConstIntNeg && inst.src[0].code == 193 &&
             inst.src[1].field == OperandField::ConstZero) {
             return;
         }
-        // v_mbcnt_hi_u32_b32 vX, exec_hi, 0
-        if (inst.src[0].field == OperandField::ExecHi &&
-            inst.src[1].field == OperandField::ConstZero) {
-            return;
+        // v_mbcnt_hi_u32_b32 vX, exec_hi, 0/vZ
+        if ((inst.src[0].field == OperandField::ExecHi ||
+             inst.src[0].field == OperandField::VccHi) &&
+            (inst.src[1].field == OperandField::ConstZero ||
+             inst.src[1].field == OperandField::VectorGPR)) {
+            return SetDst(inst.dst[0], GetSrc(inst.src[1]));
         }
+        UNREACHABLE();
     } else {
-        // v_mbcnt_lo_u32_b32 v2, -1, vX
+        // v_mbcnt_lo_u32_b32 vY, -1, vX
         // used combined with above to fetch lane id in non-compute stages
         if (inst.src[0].field == OperandField::SignedConstIntNeg && inst.src[0].code == 193) {
-            SetDst(inst.dst[0], ir.LaneId());
+            return SetDst(inst.dst[0], ir.LaneId());
         }
-        // v_mbcnt_lo_u32_b32 v20, exec_lo, vX
-        // used combined in above for append buffer indexing.
-        if (inst.src[0].field == OperandField::ExecLo) {
-            SetDst(inst.dst[0], ir.Imm32(0));
+        // v_mbcnt_lo_u32_b32 vY, exec_lo, vX
+        // used combined with above for append buffer indexing.
+        if (inst.src[0].field == OperandField::ExecLo || inst.src[0].field == OperandField::VccLo) {
+            return SetDst(inst.dst[0], GetSrc(inst.src[1]));
         }
+        UNREACHABLE();
     }
 }
 
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -471,7 +471,7 @@ bool Rasterizer::BindResources(const Pipeline* pipeline) {
         uses_dma |= stage->uses_dma;
     }
 
-    if (uses_dma && !fault_process_pending) {
+    if (uses_dma) {
         // We only use fault buffer for DMA right now.
         {
             Common::RecursiveSharedLock lock{mapped_ranges_mutex};


raphaelthegreat:
Thanks to @Nenkai for helping me find the bugged shader

Randomuser8219:
Can confirm that this PR exposed a deeper issue in Driveclub, it hangs during both the loading screen and intro.
[shad_log.txt](https://github.com/user-attachments/files/21026771/shad_log.txt)

raphaelthegreat:
The latest commit seems to help freezes for me

raphaelthegreat:
This PR appears to fix climbing Trico in The Last Guardian for me can someone confirm
Screenshot_20250703_041318

squidbus (approved):
Assuming any issues that were encountered are resolved now, the code looks good to me.

raphaelthegreat:
> Assuming any issues that were encountered are resolved now, the code looks good to me.

The freeze mentioned always existed in Driveclub, even before this PR. But seems it made it more pronounced by fixing other parts. Before merging the new DMA hotfix added here would have to be tested and @LNDF should comment on it, because I think it might affect performance.

Side note, but I think adding a toggle for DMA like readbacks would be beneficial and easy to do

Randomuser8219:
All the Driveclub freezes, including the one every multiple seconds before this PR is fixed with latest commit.

Missake212:
Can confirm that climbing also works for me, even managed to remove the spear.

image

Ruahh:
Wouldn't say climbing in TLG is _fixed_, but it's definitely less broken! There's still no "clinging", but simply standing on Trico seems to work! No more being teleported above his head!
Is behavior different on 1.03?

Also of note that I was having instability and crashing upon trying to create a new save. Enabling vk validation let me continue that.

image

Edit: Have confirmed that this is a 1.00 issue. Climbing seems to be perfect on 1.03!

kalaposfos13:
SotC hits the following unreachable after compiling shaders for a full 5 minutes beforehand:
```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:516 CompileModule: Compiling cs shader 0x41d379bc 
[Debug] <Critical> vector_alu.cpp:573 V_MBCNT_U32_B32: Unreachable code!
```
[CUSA08809.log](https://github.com/user-attachments/files/21032056/CUSA08809.log)

Edit: the relevant shader does not get loaded if `+drawJustOccluders=true` is passed to the game in CommandLineArguments.txt.

diogoaduartereis:
For me the climbing in the last guardian is still broken even on version 1.03. I just slide off the creature. Not sure what else I can do.
My specs:
- i5-12400F
- RTX4060

And the log
[shad_log.txt](https://github.com/user-attachments/files/21034098/shad_log.txt)

raphaelthegreat:
> For me the climbing in the last guardian is still broken even on version 1.03. I just slide off the creature. Not sure what else I can do. My specs:
> 
>     * i5-12400F
> 
>     * RTX4060
> 
> 
> And the log [shad_log.txt](https://github.com/user-attachments/files/21034098/shad_log.txt)

You have readbacks disabled

diogoaduartereis:
My Bad. Will try again.

LNDF:
@raphaelthegreat enabling syncing all buffers for each drawcall that uses ReadConst is slow. Mainly because every buffer is synced individually.

I don't know about driverclib or how the performance of the game is. But reading your other comments I suspect that it is running above the game's target performance level (be it 30 or 60 fps) and it slows down only in certain situations.

The above is speculation. But if it's true, it could explain why syncing in every drawcall has less stutters than syncing only the first time: if performance is good enough, you may not notice the overhead of buffer syncing and could be only parent when another thing causes a la spike, making the overhead noticeable.
When syncing every drawcall, you may not be syncing as many buffers at once, since they are spreaded across other drawcalls. If buffers are synced only on the first drawcall, you have more dirty buffers to sync and that could make the lag spikes more noticeable.

I will try to fix this problem batching buffer syncing. This will allow, with the mprotect batching in page manager to reduce the number of mprotects ok nearby buffers, hopefully reducing overhead.

I can also add a toggle to more easily find regressions.

raphaelthegreat:
@LNDF From what I saw buffers with DMA are only synced once per command lists due to the condition I removed. However the ccb can write to memory (mostly sharps) while a command list is running and thus will miss syncs. Could that be the reason?

diogoaduartereis:
It worked! The first spear can be removed. It crashed twice for me on the second spear though.

LNDF:
> @LNDF From what I saw buffers with DMA are only synced once per command lists due to the condition I removed. However the ccb can write to memory (mostly sharps) while a command list is running and thus will miss syncs. Could that be the reason?

Now that you have removed the condition, buffers are synced before each draw that uses ReadConst.

mikusp:
RE2 1.05 hits one of the unreachables on:
```
v_mbcnt_hi_u32_b32 v2, v2, 0
v_mbcnt_lo_u32_b32 v1, v1, v2
v_cmpx_eq_u32   vcc, 0, v1
```

raphaelthegreat:
> Now that you have removed the condition, buffers are synced before each draw that uses ReadConst.

Yes this is my theory in trying to explain why syncing before every such draw fixes some freezes/device losses in certain cases

raphaelthegreat:
> RE2 1.05 hits one of the unreachables on:
> 
> ```
> v_mbcnt_hi_u32_b32 v2, v2, 0
> v_mbcnt_lo_u32_b32 v1, v1, v2
> v_cmpx_eq_u32   vcc, 0, v1
> ```

What value do v2 and v1 have before the pattern

mikusp:
vcc_lo is moved into v1 and vcc_hi into v2

raphaelthegreat:
If vcc has exec then this is probably gl_SubgroupID. This will need a small refactor to mbcnt handling to compile it more literally

hgh32:
elden ring network test hit this Unreachable
`[Debug] <Critical> vector_alu.cpp:573 V_MBCNT_U32_B32: Unreachable code!`
[shad_log.txt](https://github.com/user-attachments/files/21102314/shad_log.txt)

PR #3185, opened 2025-07-03
config: Add toggle for DMA

raphaelthegreat:
Opening my take on a DMA toggle as LNDF seemed positive to the idea https://github.com/shadps4-emu/shadPS4/pull/3184#issuecomment-3031815279

It can be a useful tool to find regressions and generally compare DMA vs without until it matures more

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -52,6 +52,7 @@ static std::string isSideTrophy = "right";
 static bool isNullGpu = false;
 static bool shouldCopyGPUBuffers = false;
 static bool readbacksEnabled = false;
+static bool directMemoryAccessEnabled = false;
 static bool shouldDumpShaders = false;
 static bool shouldPatchShaders = true;
 static u32 vblankDivider = 1;
@@ -245,6 +246,10 @@ bool readbacks() {
     return readbacksEnabled;
 }
 
+bool directMemoryAccess() {
+    return directMemoryAccessEnabled;
+}
+
 bool dumpShaders() {
     return shouldDumpShaders;
 }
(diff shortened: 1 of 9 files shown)


Exhigh:
shouldn't there be a log of whether DMA has been enabled or not.

raphaelthegreat:
Damn I forgot to add it _again_

StevenMiller123:
Disabling DMA behaves as expected, bringing back the old device lost Uncharted: The Nathan Drake Collection would face when skipping asserts.
[CUSA02320.log](https://github.com/user-attachments/files/21041067/CUSA02320.log)

Disabling DMA also brings Uncharted 3's rendering back to how it behaved before the DMA PR.
image

Ruahh:
No idea where's best to say this on github, but TLG's stability is GREATLY improved with DMA disabled - to the point where I'd say it's playable with readbacks enabled and DMA disabled. Just played for 20 minutes before a crash - and the crash was much more graceful (didn't lock up my whole PC like a Device Loss crash)

So, having this option is very useful for testing!

LNDF:
Do you think it is a good idea to generate different spirv depending of a config value? There is a shader cache being worked on and changing the config value would invalidate the cache.

diogoaduartereis:
> No idea where's best to say this on github, but TLG's stability is GREATLY improved with DMA disabled - to the point where I'd say it's playable with readbacks enabled and DMA disabled. Just played for 20 minutes before a crash - and the crash was much more graceful (didn't lock up my whole PC like a Device Loss crash)
> 
> So, having this option is very useful for testing!

Thanks for the heads up and it's the same for me! I managed to play 25 minutes without a single crash. Although at the end when outside, the light gets too blinding.

PR #3186, opened 2025-07-03
buffer_cache: Fix various thread races on data upload and invalidation

raphaelthegreat:
With this PR the address space critical log spam with readbacks enabled should be gone.

On x64 in any operating system having pages marked with write-only protection is not allowed and it depends on the OS how its handled. On Linux it appears to silently make the page noaccess, while on windows it throws an error like mentioned.

To avoid this the backend always first drops the write access by syncing CPU data and then marks the area as GPU modified which drops the read access. On page invalidation the opposite order is used. However because these actions are separate, when multiple threads attempt to access the same page together with GPU thread, those protect actions can become interleaved and result in invalid protection state

To resolve this, whenever 2 protect actions are necessary they are fused under a single function call and mutex lock, so page state is always in a known state before each one. I've also fixed another race that could occur during CPU data upload. It could be possible while the memcpy to staging buffer performed in SynchronizeBuffer, that another thread invalidates that page and writes new data. Unless that page is invalidated again that data would not get synced. Now the memcpy is performed inside the ForEachUploadRange function with a mutex lock and protection is moved to be before the memcpy again, so that if another page attempts to write there it will have to wait first

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include <algorithm>
+#include <mutex>
 #include "common/alignment.h"
-#include "common/config.h"
 #include "common/debug.h"
 #include "common/scope_exit.h"
 #include "common/types.h"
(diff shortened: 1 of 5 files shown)


georgemoralis:
updated branch for testing with dma option off

StevenMiller123:
Readbacks-related address space log spam does appear to be fixed by this. Persona 5 Royal is still unstable with readbacks enabled though.
[CUSA17419 readbacks.log](https://github.com/user-attachments/files/21043025/CUSA17419.readbacks.log)

raphaelthegreat:
By unstable you mean crashing during hiding sequence or just random crashes that are caused by readbacks

StevenMiller123:
Random crashes, in that test run I crashed during the first battle in the tutorial.

LNDF (approved):

hgh32:
Seems to cause grey screens in Dark Souls 3 Network Test and Bloodborne Alpha Test
ds3
bb
[ds3.txt](https://github.com/user-attachments/files/21102047/ds3.txt)
[bb.txt](https://github.com/user-attachments/files/21102048/bb.txt)

StevenMiller123:
Breaks text rendering in Uncharted: The Nathan Drake Collection.
Before:
image

After:
image

Both tests were done with directMemoryAccess enabled, since this title requires it.
[CUSA02320 before.log](https://github.com/user-attachments/files/21111117/CUSA02320.before.log)
[CUSA02320 after.log](https://github.com/user-attachments/files/21111125/CUSA02320.after.log)

StevenMiller123:
This PR also broke The Last of Us Remastered rendering

Before:
image

After:
image

[CUSA00552 before.log](https://github.com/user-attachments/files/21111321/CUSA00552.before.log)
[CUSA00552 after.log](https://github.com/user-attachments/files/21111322/CUSA00552.after.log)

PR #3190, opened 2025-07-04
Revert "Avoid clearing depth on partial HTILE writes"

georgemoralis:
Reverts shadps4-emu/shadPS4#3167


georgemoralis:
turtle asked me to revert because it causes regressions . He will rework it

PR #3194, opened 2025-07-05
video_out: Internal Resolution Support

Fire-Cube:
This PR separates the resolution of the emulator window from the one used by the shadPS4's graphics driver.
There’s a game (The Last Guardian) that uses sceVideoOutGetResolutionStatus to retrieve the resolution, which currently reflects the value stored in the config.

If you set it to 4K using a patch but the emulator window is smaller, it gets ignored due to that check.
To handle this case, I’m introducing an internal resolution that can be adjusted as needed and only affects game behavior.

I don't know if i should add a Qt config but i can do if it is the case


polybiusproxy:
I'm against exposing it on the Qt UI.
Some games use the resolution as a heuristic to determine when to use NEO-exclusive features.
User sets 4K resolution -> Game thinks it's running on a NEO and tries to use an unsupported GPU feature -> Bad things happen.

StevenMiller123:
This will need a rebase now that #3181 is merged.

Fire-Cube:
rebased

Fire-Cube:
another rebase (pls merge i don't want to do this a 3rd time)

georgemoralis:
yes sir!

src/common/config.cpp
```diff
@@ -60,8 +60,10 @@ static bool overrideControllerColor = false;
 static int controllerCustomColorRGB[3] = {0, 0, 255};
 
 // GPU
-static u32 screenWidth = 1280;
-static u32 screenHeight = 720;
+static u32 windowWidth = 1280;
+static u32 windowHeight = 720;
+static u32 internalScreenWidth = 1280;
+static u32 internalScreenHeight = 720;
```
ghost, line 66: Shouldn't this be defaulted to 1920x1080 instead of 1280x720?

StevenMiller123: 1280x720 was the default resolution before this PR too, I don't see a reason why that needs to be changed.

ghost: I was asking about the internalScreen value, since to my knowledge PS4 games are usually 1080p that's why I was curious.

StevenMiller123: While the PS4 supported up to 1080p, it has full support for specific resolutions below that, like 720p.
Telling games they're running on a 720p screen won't cause any bugs, and if you want a different resolution, you can change the config option on your end. Just note that we keep this setting out of the GUI to prevent unaware individuals from setting this to resolutions above 1080p, which some games use to trigger PS4 Pro support (since the base PS4 didn't support anything above 1080p).

ghost: Yes I get all that, I was just wondering if defaulting to 1080p wouldn't make more sense in the case a game supported 720p and would then render at 720p when it could do 1080p, but I guess this is an extremely rare case.

PR #3195, opened 2025-07-05
Ime library fixes

w1naenator:
- Restructured the code for better clarity.
- Added missing enums and structs.
- Updated OrbisImeTextAreaProperty.
- Fixed incorrect / incomplete OrbisImeEventParam.
- Replaced OrbisImeDialogOption with OrbisImeOption.
- Removed the incorrect OrbisImeDialogOption.
- Fixed OrbisImeDialogParam.
- Updated the sceImeOpen declaration to use extended parameters, but this is not yet implemented.


src/core/libraries/ime/ime_common.h
```diff
@@ -3,9 +3,108 @@
...
+    SWEDISH = 0x0000000000001000,
+    JAPANESE = 0x0000000000002000,
+    KOREAN = 0x0000000000004000,
+    SIMPLIFIED_CHINESE = 0x0000000000008000,
+    TRADITIONAL_CHINESE = 0x0000000000010000,
+    PORTUGUESE_BR = 0x0000000000020000,
+    ENGLISH_GB = 0x0000000000040000,
+    TURKISH = 0x0000000000080000,
+    SPANISH_LA = 0x0000000000100000,
+    ARABIC = 0x0000000001000000ULL,
```
kalaposfos13: Why are only Dutch and Arabic unsigned long long here?

gandalfthewhite19890404:
```
[Debug] <Critical> assert.cpp:30 assert_fail_debug_msg: Assertion Failed!
buf_len + 1 <= buf_size && "Is your input buffer properly zero-terminated?" at D:\a\shadPS4\shadPS4\externals\dear_imgui\imgui_widgets.cpp:4601
```

Still not fixed in Diablo III, game does not accept non-Latin names since https://github.com/shadps4-emu/shadPS4/commit/6ebed7ce69a54264454c122a1f6f0d84f2fb800a
But this PR was not for that issue?

src/core/libraries/ime/ime_common.h
w1naenator: Only Dutch and Arabic are marked with ULL due to a small formatting inconsistency. It doesn’t affect the result, because the enum is declared with : u64, so all values are treated as uint64_t regardless of the suffix.

I fixed all members to same format.

georgemoralis:
Does your pr applies to a specific game? Or it is generally refactoring?

thecatontheceiling:
Why are these "moved from" comments there? Why are most of the comments here as a matter of fact? What's the need for some of these changes? "Refactoring" is not a valid explanation when the purpose of some changes here are not clear. Is this PR partially AI-written?

w1naenator:
> Why are these "moved from" comments there? Why are most of the comments here as a matter of fact? What's the need for some of these changes? "Refactoring" is not a valid explanation when the purpose of some changes here are not clear. Is this PR partially AI-written?

It is not ai written, comments many structures moved to ime_common, because needed for use in ime.h and ime_dialog.h i just fixed one struct was incorrect, one updated and added some well known structs.

w1naenator:
> Does your pr applies to a specific game? Or it is generally refactoring?

Just some fixes and refactor for easer use later.

polybiusproxy:
Regardless, I don't think this should be merged (or at least part of it). Were these snippets from psOff taken with permission?

w1naenator:
> Regardless, I don't think this should be merged (or at least part of it). Were these snippets from psOff taken with permission?

But this information is available in multiple public sources, and the variable names are also known. I don't think it's a big problem.

StevenMiller123:
@w1naenator While this is true, we'd prefer to play it safe with content related to psOff, as their team has expressed discontent with us before over code like this.

w1naenator:
I can provide many other sources with exact information; I can simply choose not to show the source, and no one would be able to prove where the information comes from

StevenMiller123:
Since you're unwilling to properly address the concerns we've discussed, I'm closing this PR.

w1naenator:
@StevenMiller123 But that's not true. I just don't understand how to avoid copyright issues in this situation. In this case, the entire IME library in shadPS looks like a copy of PSOff, because the names, structures and data types are already well-known. Could you please explain what I should do?

w1naenator:
You can also find this information in public manuals for 3ds Max 2019, for example. It seems that 3ds Max itself takes this information from official PS4 SDK documentation. So I still don't understand what the problem is.

w1naenator:
https://help.autodesk.com/view/3DSMAX/2019/ENU/?guid=__lua_ref_ns_stingray_PS4ImeDialog_html

w1naenator:
@georgemoralis Could you please explain what I did wrong? My intention is not to cause any issues. I simply want to understand what steps I need to take to avoid potential problems.

georgemoralis:
reopening the pr using above as references should be fine

PR #3197, opened 2025-07-06
video_core: Batch syncing buffers for DMA (WIP)

LNDF:
This is a draft, not ready.

It aims to batch syncing buffers for DMA.

I have not tested this. Expect not to work.


StevenMiller123:
Doesn't cause any additional problems in Uncharted: The Nathan Drake Collection, though menu rendering is still broken. This PR should probably get rebased though, since a variety of rendering regressions were fixed recently.
image
[CUSA02320.log](https://github.com/user-attachments/files/21124787/CUSA02320.log)

This does seem to improve DMA performance in The Last of Us™ Remastered, though I need to use a modified version of the game to bypass an internal assert regarding performance.
Before:
image
[CUSA00552 main.log](https://github.com/user-attachments/files/21124915/CUSA00552.main.log)

After:
image
[CUSA00552 PR.log](https://github.com/user-attachments/files/21124916/CUSA00552.PR.log)

LNDF:
> Doesn't cause any additional problems in Uncharted: The Nathan Drake Collection, though menu rendering is still broken. This PR should probably get rebased though, since a variety of rendering regressions were fixed recently.
> image
> [CUSA02320.log](https://github.com/user-attachments/files/21124787/CUSA02320.log)
> 
> This does seem to improve DMA performance in The Last of Us™ Remastered, though I need to use a modified version of the game to bypass an internal assert regarding performance.
> Before:
> image
> [CUSA00552 main.log](https://github.com/user-attachments/files/21124915/CUSA00552.main.log)
> 
> After:
> image
> [CUSA00552 PR.log](https://github.com/user-attachments/files/21124916/CUSA00552.PR.log)
> 

I don't think the current state of the PR is the final one. I consider this draft a place where I put the tests I do for testing improving performance when syncing all buffers.

I will also look into other emus to see how they are doing (if they are doing anything similar to this).

LNDF:
@raphaelthegreat 

In an earlier PR (I think readbscks) you made so that buffers are synced every time ReadConst is used. While this PR I proves performance doing that, it is still not enough for games that heavily use it. (Like uncharted 3 which uses DMA in almost every draw).

You mentioned that this is necessary since it is possible that memory is modified during the execution of the command list and it would have to be synced to GPU memory. What are the cases where this can happen?

Also I have changed FuncTraits, which you originally wrote. Is it ok?

raphaelthegreat:
If you check liverpool.cpp you will see aside from dcb buffer there is ccb command list which is used for writing constants. The dcb and ccb have packets to synchronize with each other so dcb ensures data has been written before issuing a draw/dispatch. It is possible those data could be used in shader by ReadConst though I have not confirmed it. Check the DumpConstRam packet for that. I think a compromise might be to replace the plain memcpy with InlineData so its always uploaded to buffer cache and do DMA sync only 1 time per command list instead of per frame

LNDF:
> If you check liverpool.cpp you will see aside from dcb buffer there is ccb command list which is used for writing constants. The dcb and ccb have packets to synchronize with each other so dcb ensures data has been written before issuing a draw/dispatch. It is possible those data could be used in shader by ReadConst though I have not confirmed it. Check the DumpConstRam packet for that.

Will take a look into it.

> I think a compromise might be to replace the plain memcpy with InlineData so its always uploaded to buffer cache and do DMA sync only 1 time per command list instead of per frame.

Yes. That was my idea when I wrote the comment.

raphaelthegreat:
The previous implement would sync only on submit done so only 1 time per frame. Had very high chance of missing data iirc

LNDF:
> If you check liverpool.cpp you will see aside from dcb buffer there is ccb command list which is used for writing constants. The dcb and ccb have packets to synchronize with each other so dcb ensures data has been written before issuing a draw/dispatch. It is possible those data could be used in shader by ReadConst though I have not confirmed it. Check the DumpConstRam packet for that. I think a compromise might be to replace the plain memcpy with InlineData so its always uploaded to buffer cache and do DMA sync only 1 time per command list instead of per frame

I had a little time to look into Liverpool. Do I need to worry only about DumpConstRam or do I need to worry about Ops like WriteData/CopyData or the event ops that write to memory?

raphaelthegreat:
You mean the DmaData packet? I don't believe so, as those already go through the buffer cache and will make sure to have the data copied/uploaded on the GPU. Event ops writing to memory are almost always for GPU to GPU sync or  GPU to CPU sync. Labels written by EventWriteEop/Eos are typically waited on by WaitRegMem as an example of first case, don't think you should worry about those now. Replacing the memcpy in DumpConstRam with InlineData should be enough

LNDF:
@StevenMiller123 It is ready to test now. To find bugs and fix them. Uncharted 3 might be a good game since it uses DMA very heavily.

@raphaelthegreat So do you think it is better to add the post barrier insteed of a full memory barrier? Also, I moved the fault buffer processing to the command list end while keeping the image download only on submit.

raphaelthegreat:
What do you mean exactly by a post barrier? Though now that DMA is done once per command list, having the memory barrier is fine cause its not gonna be executed every draw

LNDF:
> What do you mean exactly by a post barrier? Though now that DMA is done once per command list, having the memory barrier is fine cause its not gonna be executed every draw

I was talking about the other PR, where you removed memory barrier because post sync barriers where enough. In this PR I removed post barriers entirely in fabour of memory barrier. I was asking which one to use here.

StevenMiller123:
| Main | PR |
| --- | --- |
| image | image |

[CUSA02320.log](https://github.com/user-attachments/files/21323588/CUSA02320.log)

This PR noticably improves performance of Uncharted 3 with only DMA enabled. It does make Uncharted 3 more unstable though, I've had more access violations and hangs on here than I do on main. I'm not currently setup to debug these issues.

LNDF:
> Main 	PR
> image 	image
> 
> [CUSA02320.log](https://github.com/user-attachments/files/21323588/CUSA02320.log)
> 
> This PR noticably improves performance of Uncharted 3 with only DMA enabled. It does make Uncharted 3 more unstable though, I've had more access violations and hangs on here than I do on main. I'm not currently setup to debug these issues.

I can't see anything wrong in the logs. I will check tomorrow more tomorrow.

georgemoralis:
this can't be updated to latest , if it is not useful considering closing it

LNDF:
> this can't be updated to latest , if it is not useful considering closing it

It has been a long time since I last updated it but it should help with the large number of uploads caused by DMA.

Will be closing it now. Maybe I will reopen it in the future if I take it the future (now i'm with gc2, currently stuck)

PR #3199, opened 2025-07-06
libkernel: Add missing mprotect export

StevenMiller123:
@mikusp ran into this when testing some homebrew.


squidbus:
Why draft?

StevenMiller123:
I marked this as draft so I could look into some potential issues with protect logic.
@mikusp's use case involved setting execute permissions, which we don't handle properly yet.

squidbus:
Okay, although I think this change alone is pretty straightforward.

kalaposfos13:
@squidbus I can confirm that the specific use-case this PR aims to fix (which is here: https://github.com/kalaposfos13/eboot-hooks-prx/blob/main/src/common/LightHook.h#L286) isn't fixed with this yet.

squidbus:
I don’t think that really matters for just adding a missing symbol export for an existing function

StevenMiller123:
I'll let this get merged as-is for now, and do a separate PR to properly fix mprotect calls with execute permissions.

PR #3200, opened 2025-07-06
Fix V_ADDC_U32 carry-out edge cases

OFFTKP:
Carry calculations in unsigned numbers work like this:
```
carry = result < src
```
It doesn't matter which of the two sources you check, because in unsigned addition when there's an overflow the result will be smaller than both sources, never just one.

---

In the case of add with carry, you need to check whether src1 + src2 overflows and whether (src1+src) + cf overflows separately.

The current implementation does `result < src1 || result < src2`. This doesn't check separately for the intermediate result of src1+src2 and only checks if the final result overflowed. This fails in cases where the intermediate result overflows but the final result doesn't.

---

Example where old implementation can fail:

src0 = 0xFFFF'FFFF
src1 = 0xFFFF'FFFF
carry_in = 1

result = 0xFFFF'FFFF + 0xFFFF'FFFF + 1 = 0xFFFF'FFFF

This previous implementation would check `result < src0 (false)` || `result < src1 (false)` and carry_out would be false

### DIFF
--- a/src/shader_recompiler/frontend/translate/vector_alu.cpp
+++ b/src/shader_recompiler/frontend/translate/vector_alu.cpp
@@ -623,12 +623,13 @@ void Translator::V_ADDC_U32(const GcnInst& inst) {
     const IR::U32 src0{GetSrc(inst.src[0])};
     const IR::U32 src1{GetSrc(inst.src[1])};
     const IR::U32 carry{GetCarryIn(inst)};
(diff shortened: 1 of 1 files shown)


OFFTKP:
@squidbus can you take a look at this?

squidbus (approved):

squidbus (approved):

PR #3205, opened 2025-07-07
texture_cache: Change depth resolve new image back to max of resources.

squidbus:
https://github.com/shadps4-emu/shadPS4/pull/3079 changed the resource count for the re-created depth overlap image to the minimum of requested and cached resources instead of the max. This resulted in spam of the following in at least CUSA16404:

```
[Render.Vulkan] <Warning> texture_cache.cpp:405 FindImage: Image overlap resolve failed
```

It's not clear to me why this was changed to `std::min` as usually resource overlaps would resolve to the greatest number of sub-resources rather than the least. When the number of resources in the resolved image is less than requested the code is set up to always fail the resolve.

Proposing changing this back as a PR as the short comment thread on this in the previous PR did not gain traction. With this change the game in question still works and no longer spams overlap failures.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -169,7 +169,7 @@ ImageId TextureCache::ResolveDepthOverlap(const ImageInfo& requested_info, Bindi
 
     if (recreate) {
         auto new_info = requested_info;
-        new_info.resources = std::min(requested_info.resources, cache_image.info.resources);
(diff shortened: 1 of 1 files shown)


squidbus:
Tagging @raphaelthegreat as I can't request for review

georgemoralis:
of course sakura wars :D

raphaelthegreat:
This looks okay to me. I think only the UpdateSize call removal was enough for DC in hindsight

kalaposfos13:
This also fixed the same warning spam in SotC

PR #3211, opened 2025-07-08
buffer_cache: Bring back upload batching and temporary buffer

raphaelthegreat:
This should fix regressions reported from https://github.com/shadps4-emu/shadPS4/pull/3186

Because that PR fused the write and read protections under a single function call, it was a requirement to move the actual memory copy part inside the lambda to perform it before the read protection kicks in. However on certain large data transfers it had potential for data corruption. If, for example, an upload had two copies, a 400MB and a 300MB one, the first one would fit in the staging buffer, very likely with an induced stall. However the second one wouldn't have space to fit alongside the other data, but it's also small enough for the buffer to fit it, so the staging buffer would cause a flush and wait to copy it, overwriting the previous transfer.

To address this the upload function has been reworked to allow for batching like previously but with the new locking behavior. Also the condition to use temporary buffers has been expanded to also include cases when staging buffer will stall, which should increase performance a little in some cases.

### DIFF
--- a/src/video_core/buffer_cache/buffer.cpp
+++ b/src/video_core/buffer_cache/buffer.cpp
@@ -137,12 +137,15 @@ StreamBuffer::StreamBuffer(const Vulkan::Instance& instance, Vulkan::Scheduler& 
                           size_bytes);
 }
 
-std::pair<u8*, u64> StreamBuffer::Map(u64 size, u64 alignment) {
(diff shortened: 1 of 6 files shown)


StevenMiller123:
This fixes the reported regression in The Last of Us Remastered, will check Uncharted soon.
image

StevenMiller123:
The reported regression in Uncharted: The Nathan Drake Collection is also fixed by this PR.
image

PR #3213, opened 2025-07-08
Log sceKernelAllocateDirectMemory call on error

mailwl:
If memory allocation failed, there is no logging sceKernelAllocateDirectMemory() call

### DIFF
--- a/src/core/libraries/kernel/memory.cpp
+++ b/src/core/libraries/kernel/memory.cpp
@@ -57,17 +57,17 @@ s32 PS4_SYSV_ABI sceKernelAllocateDirectMemory(s64 searchStart, s64 searchEnd, u
 
     auto* memory = Core::Memory::Instance();
     PAddr phys_addr = memory->Allocate(searchStart, searchEnd, len, alignment, memoryType);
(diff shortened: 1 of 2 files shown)


src/core/libraries/kernel/memory.cpp
```diff
@@ -57,17 +57,17 @@ s32 PS4_SYSV_ABI sceKernelAllocateDirectMemory(s64 searchStart, s64 searchEnd, u
 
     auto* memory = Core::Memory::Instance();
     PAddr phys_addr = memory->Allocate(searchStart, searchEnd, len, alignment, memoryType);
-    if (phys_addr == -1) {
-        return ORBIS_KERNEL_ERROR_EAGAIN;
-    }
-
-    *physAddrOut = static_cast<s64>(phys_addr);
 
     LOG_INFO(Kernel_Vmm,
```
StevenMiller123, line 61: Errors are already logged through the `memory->Allocate` call above, I'd suggest adding more parameters to the error log instead of moving the error return here.

StevenMiller123 (changes requested):

PR #3217, opened 2025-07-09
Zero top bits in INSERTQ/EXTRQ

OFFTKP:
AMD manual says the value of these bits is undefined but AMD CPUs seem to zero them


georgemoralis:
does it pass stress test?

OFFTKP:
> does it pass stress test?

now it does

kalaposfos13:
No regressions in Gravity Rush 2.

PR #3223, opened 2025-07-10
Define S_TRAP as InstCategory::FlowControl and stub it

kalaposfos13:
One of the shaders in SotC uses this, and it being undefined lead to an unreachable at translate.cpp:608 previously.


mikusp (approved):

georgemoralis:
i don't see any harm on this let's see what @squidbus has to say

squidbus (approved):

PR #3224, opened 2025-07-10
ir_passes: Fold readlane with ff1 pattern

raphaelthegreat:
I've recently spotted the following shader pattern, used in some waterfall loops, which is essentially a we-have-v_readfirstlane_b32-at-home

```
/*00000000000c*/ s_mov_b64       s[16:17], exec
...
.LOOP
/*000000000028*/ s_ff1_i32_b64   s18, s[16:17]
/*00000000002c*/ v_readlane_b32  s25, v3, s18
```

Which on main gets translated to
```glsl
subgroupBroadcast(uint(gl_InstanceIndex), 3305621016u);
```

There are 2 issues here. First EmitReadLane assumes the lane id is a constant u32, which is not always the case (the large number here is because that's a pointer stored in the Opaque IR::Value). Second is that s_ff1_i32_b64 is implemented with GetSrc64 instead of thread bits. This instruction should probably be swapped to thread bits but because I don't remember the original usecase, which could have been 64-bit, I won't change it in this PR.

To solve this issue, the readlane elimination pass will detect when readlane has an non immediate lane coming from FindILSB64 and substitute it with ReadFirstLane

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
@@ -529,7 +529,7 @@ Id EmitLaneId(EmitContext& ctx);
 Id EmitWarpId(EmitContext& ctx);
 Id EmitQuadShuffle(EmitContext& ctx, Id value, Id index);
 Id EmitReadFirstLane(EmitContext& ctx, Id value);
-Id EmitReadLane(EmitContext& ctx, Id value, u32 lane);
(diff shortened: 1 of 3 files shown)


raphaelthegreat:
Alternatively s_ff1_i32_b64 could be implemented as subgroupBallotFindLSB(subgroupBallot(src0)) and leave readlane untouched, that might work in more cases if this pattern it used in more weird way, though that will transition the opcode to a thread bit one. What the maintainers think would be best

squidbus (approved):

PR #3225, opened 2025-07-10
shader_recompiler: Implement ff1 with subgroup ops

raphaelthegreat:
My other PR was snipe merged before I had a chance to present the alternate implementation :/ https://github.com/shadps4-emu/shadPS4/pull/3224#issuecomment-3056958483

This approach is more faithful to original source code and might work better if ff1 is used with conditions other than current exec. I don't know if it's better than the other one though


georgemoralis:
seems to be a cleaner version of previous pr

squidbus (approved):

PR #3227, opened 2025-07-10
vk_rasterizer: Set render area to window scissor when no framebuffers are bound

raphaelthegreat:
It is possible to have a draw call without any framebuffers attached, but still rely on all invocations inside scissor area being invoked (for example writing to storage image with gl_FragCoord coodinates). On main render area would be set to 1 in that case, replace that with the maximum render area, so it covers all of scissor rect

I vaguely remember Knack doing this but not sure

### DIFF
--- a/src/video_core/renderer_vulkan/vk_instance.h
+++ b/src/video_core/renderer_vulkan/vk_instance.h
@@ -324,9 +324,19 @@ public:
         return properties.limits.maxViewportDimensions[0];
     }
 
-    ///  Returns the maximum viewport height.
+    /// Returns the maximum viewport height.
     u32 GetMaxViewportHeight() const {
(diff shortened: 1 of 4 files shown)


StevenMiller123:
I'm not noticing any graphical improvements in Knack so far.
@Randomuser8219 might be more observant though.

[CUSA00068.log](https://github.com/user-attachments/files/21167371/CUSA00068.log)

Randomuser8219:
image
I'm not seeing any skybox smearing anymore. It's 2 FPS since I'm using the debug log filter to get past the PM4 race condition.

Randomuser8219:
image
Update: I still see skybox smearing

squidbus (approved):

Randomuser8219:
I got randomly ghost mentioned in here?

PR #3233, opened 2025-07-11
Change companion util stub to return NO_EVENT

mikusp:
Some games, like Battlefield 1, really don't like getting an unexpected error here. Simply returning no event allows them to go further.

### DIFF
--- a/src/core/libraries/companion/companion_util.cpp
+++ b/src/core/libraries/companion/companion_util.cpp
@@ -29,10 +29,9 @@ u32 PS4_SYSV_ABI getEvent(sceCompanionUtilContext* ctx, sceCompanionUtilEvent* o
 }
 
 s32 PS4_SYSV_ABI sceCompanionUtilGetEvent(sceCompanionUtilEvent* outEvent) {
-    sceCompanionUtilContext* ctx = nullptr;
-    u32 ret = getEvent(ctx, outEvent, 1);
+    u32 ret = ORBIS_COMPANION_UTIL_NO_EVENT;
 
-    LOG_DEBUG(Lib_CompanionUtil, "(STUBBED) called ret: {}", ret);
+    LOG_DEBUG(Lib_CompanionUtil, "(STUBBED) called ret: {:#x}", ret);
     return ret;
 }
 


src/core/libraries/companion/companion_util.cpp
```diff
@@ -29,10 +29,9 @@ u32 PS4_SYSV_ABI getEvent(sceCompanionUtilContext* ctx, sceCompanionUtilEvent* o
 }
 
 s32 PS4_SYSV_ABI sceCompanionUtilGetEvent(sceCompanionUtilEvent* outEvent) {
-    sceCompanionUtilContext* ctx = nullptr;
-    u32 ret = getEvent(ctx, outEvent, 1);
+    u32 ret = ORBIS_COMPANION_UTIL_NO_EVENT;
```
StevenMiller123, line 32: It seems like the actual issue is that getEvent will always return `ORBIS_COMPANION_UTIL_INVALID_POINTER` since ctx is initialized as a nullptr.
I think it would be preferable to fix that bug instead of getting rid of the getEvent code entirely.

StevenMiller123 (commented):
Pinging @Fire-Cube for their opinion on this, since their code is what's bugged in the first place.

src/core/libraries/companion/companion_util.cpp
Fire-Cube: Please check whether the game calls sceCompanionUtilInitialize, because my test game, Star Wars: Battlefront, doesn’t.
That’s why I built the full check logic for but without Initialize, it just returns an error.
I think we simply need to handle the initialization properly in that case.

mikusp: With this change and epolls SW:B seems to reach menus https://github.com/shadps4-emu/shadPS4/pull/3476#issuecomment-3239429061

StevenMiller123: As discussed before, the core issue with libSceCompanionUtil comes from initializing that `ctx` variable to a nullptr.
This triggers an error return in `getEvent` which causes the issues seen in games.

This PR hasn't seen any updates since my review here, but I'd still rather see this bug properly fixed over adding in a hack.

mikusp: The game does not seem to call any other CompanionUtil functions, so where should it be initialized?

StevenMiller123: I haven't decompiled this library, so I don't know.
All I know is that, if games are unhappy with this current behavior, then it probably needs fixing.

georgemoralis:
hackish

PR #3235, opened 2025-07-12
Implement libusb callbacks with correct ABI

connorh315:
The libusb library executes the emulated callback, but does not respect the PS4's ABI, as such passing the libusb_transfer argument into the RAX register rather than the RDI register.

This pull request introduces a "trampoline" that registers a host-machine callback that then executes the emulated callback with the correct ABI, fixing the issue.

Only tested on Windows.

This now makes LEGO Dimensions enter the game and seems to be fully playable.

To be considered:

- Should this be applied to all types of transfer (i.e. bulk, control, isochronous)
- Is there a better alternative to creating and maintaining a dictionary of transfers?

### DIFF
--- a/src/core/libraries/usbd/usbd.cpp
+++ b/src/core/libraries/usbd/usbd.cpp
@@ -10,6 +10,33 @@
 #include <fmt/format.h>
 #include <libusb.h>
 
+#include <unordered_map>
+#include <mutex>
+
+namespace {
+    std::unordered_map<Libraries::Usbd::SceUsbdTransfer*, Libraries::Usbd::SceUsbdTransferCallback> g_transfer_callbacks;
+    std::mutex g_callback_mutex;
+
+    void transfer_abi_callback(libusb_transfer* transfer) {
+        using SysVCallback = void (PS4_SYSV_ABI *)(Libraries::Usbd::SceUsbdTransfer*);
+        SysVCallback callback = nullptr;
+
+        {
+            std::lock_guard<std::mutex> lock(g_callback_mutex);
+            auto it = g_transfer_callbacks.find(transfer);
+            if (it != g_transfer_callbacks.end()) {
+                callback = (SysVCallback)it->second;
+            }
+        }
+
+        if (callback) {
+            callback(transfer);
+        } else {
+            LOG_WARNING(Lib_Usbd, "No registered callback for USB transfer {:p}");
+        }
+    }
+}
+
 namespace Libraries::Usbd {
 
 s32 libusb_to_orbis_error(int retVal) {
@@ -276,6 +303,11 @@ s32 PS4_SYSV_ABI sceUsbdCancelTransfer(SceUsbdTransfer* transfer) {
 void PS4_SYSV_ABI sceUsbdFreeTransfer(SceUsbdTransfer* transfer) {
     LOG_DEBUG(Lib_Usbd, "called");
 
+    {
+        std::lock_guard<std::mutex> lock(g_callback_mutex);
+        g_transfer_callbacks.erase(transfer);
+    }
+
     libusb_free_transfer(transfer);
 }
 
@@ -305,7 +337,12 @@ void PS4_SYSV_ABI sceUsbdFillInterruptTransfer(SceUsbdTransfer* transfer,
                                                u32 timeout) {
     LOG_DEBUG(Lib_Usbd, "called");
 
-    libusb_fill_interrupt_transfer(transfer, dev_handle, endpoint, buffer, length, callback,
+    {
+        std::lock_guard<std::mutex> lock(g_callback_mutex);
+        g_transfer_callbacks[transfer] = callback;
+    }
+
+    libusb_fill_interrupt_transfer(transfer, dev_handle, endpoint, buffer, length, transfer_abi_callback,
                                    user_data, timeout);
 }
 
--- a/src/core/libraries/usbd/usbd.cpp
+++ b/src/core/libraries/usbd/usbd.cpp
@@ -32,7 +32,7 @@ namespace {
         if (callback) {
             callback(transfer);
         } else {
-            LOG_WARNING(Lib_Usbd, "No registered callback for USB transfer {:p}");
+            LOG_WARNING(Lib_Usbd, "No registered callback for USB transfer");
         }
     }
 }
--- a/src/core/libraries/usbd/usbd.cpp
+++ b/src/core/libraries/usbd/usbd.cpp
@@ -10,32 +10,33 @@
 #include <fmt/format.h>
 #include <libusb.h>
 
-#include <unordered_map>
 #include <mutex>
+#include <unordered_map>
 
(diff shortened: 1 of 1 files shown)


Graubek:
I can confirm that the game works (game spams a lot of this in console)
```
[Lib.Remoteplay] <Error> remoteplay.cpp:168 sceRemoteplayProhibit: (STUBBED) called
[Lib.Pad] <Info> pad.cpp:526 scePadResetLightBar: (DUMMY) called
[Core] <Error> stubs.cpp:42 CommonStub: Stub: sceVideoOutSetWindowModeMargins (nid: MTxxrOCeSig) called, returning zero to 0x80071509a
```
obraz

georgemoralis (approved):

georgemoralis:
that seems valid , i think it is okay for it's purpose ;)

connorh315:
> that seems valid , i think it is okay for it's purpose ;)

Perfect - I've opened the pull out of a draft!

src/core/libraries/usbd/usbd.cpp
```diff
@@ -10,6 +10,34 @@
 #include <fmt/format.h>
 #include <libusb.h>
 
+#include <mutex>
+#include <unordered_map>
+
+namespace {
+std::unordered_map<Libraries::Usbd::SceUsbdTransfer*, Libraries::Usbd::SceUsbdTransferCallback>
+    g_transfer_callbacks;
```
squidbus, line 18: Do we actually need to use a map and lock for this? Think we could just do something like wrap the user data instead. So something like:
```
// Some struct type you define
auto host_user_data = {
    .callback = callback,
    .user_data = user_data,
};
libusb_fill_interrupt_transfer(transfer, dev_handle, endpoint, buffer, length,
                               transfer_abi_callback, host_user_data, timeout);
```

And then the host `transfer_abi_callback` just pulls the callback and real user data from the host user data and makes the right call.

connorh315: This was my initial idea that got the game running. However my concern is that, whilst it's unlikely, the game could change the value of the callback, leading to a likely crash. Whilst LEGO Dimensions did work with this method, the issue is that the game setups a long-running transfer buffer, and technically the game could change the callback at any time...

squidbus: What exactly is the concern if the callback changes here? I don’t see why it can’t be replaced like this.

connorh315: Because if the game replaces the callback then it won't fire the transfer_abi_callback function anymore it will fire the callback it defines which will not be triggered with the correct ABI.

connorh315: Come to think of it even with the dictionary method that's the same issue. I was thinking of the user_data part - If the game changes the user_data after the user_data has been set to the host_user_data struct then the correct callback is lost. I guess it would just be easier... Since it's screwed either way I might as well just go for the more performant option of switching the user_data structure back and forth. I can't think of a more eloquent method unfortunately.

squidbus: If your concern is that the caller might manually set these struct fields then yes this approach with a map will not help, you would need to instead go to the libusb calls where the populated structure is passed in to actually be used and swap out the callback and user data with wrappers.

squidbus: Or alternatively we fork libusb to change the callback ABIs, although it would make it harder to keep up to date and prevent using system packages of libusb on e.g. Linux distros.

connorh315: > Or alternatively we fork libusb to change the callback ABIs, although it would make it harder to keep up to date and prevent using system packages of libusb on e.g. Linux distros.

I think this is the only true fix to the issue, granted I agree with you in that I don't think it's a very good one.

> If your concern is that the caller might manually set these struct fields then yes this approach with a map will not help, you would need to instead go to the libusb calls where the populated structure is passed in to actually be used and swap out the callback and user data with wrappers.

Yes this was the reason why I went to the dictionary, thinking it would stop this from happening, whilst it stops the user_data being changed from being an issue it doesn't change the fact that the caller could just change the callback pointer.


Bottom line is I don't think there is going to be a solution that does not implement some issue here. Even if we used a copy of the transfer structure, it would require synchronization between libusb and the caller, I guess we can determine when the structure is freed as there is a function for it.

What do you think would be the best method here? I'll re-commit with that suggestion.

georgemoralis: libusb doesn't have that regular updates so i think we can handle it..

squidbus: Yeah modifying libusb would probably be the best option here.

georgemoralis: https://github.com/shadps4-emu/ext-libusb

connorh315: Okay perfect, do I need to just pull down that git repo you've linked, make the changes to that and submit a pull request to that repo. Then close this pull request down?

georgemoralis: yes , it will require a change to cmake on this project so it will link ext-libusb

connorh315: Sure - What would that look like then? Not particularly good at cmake...

georgemoralis: hmm actually only submodule will need to be updated to link to the new repo
https://github.com/shadps4-emu/shadPS4/blob/main/.gitmodules

squidbus: You also need to remove logic to use system libusb if present since it won’t be compatible.

georgemoralis: @connorh315 if you want do the changes to the ext-libusd and i will do the rest here. just ping me when you are done

connorh315: > @connorh315 if you want do the changes to the ext-libusd and i will do the rest here. just ping me when you are done

Yep that sounds much more doable. Thank you!

connorh315: > You also need to remove logic to use system libusb if present since it won’t be compatible.

Lol I had no chance at figuring that one out...

marianotella (approved):

src/core/libraries/usbd/usbd.cpp
connorh315: @georgemoralis I've made the necessary change in ext-libusb

georgemoralis: ok thanks i will check it tomorrow , i am away from my dev pc

connorh315: > ok thanks i will check it tomorrow , i am away from my dev pc

Perfect!

georgemoralis:
replaced by #3284

PR #3238, opened 2025-07-13
shader_recompiler: Replace buffer pulling with attribute divisor for instance step rates

raphaelthegreat:
This is first part of trying to cleanup the attribute code in recompiler before I can implement what I want, as it has suffered from a bit of scope creep.

In this PR the code for manually pulling buffers for emulating instance step rates is removed in favor of  using VK_EXT_vertex_attribute_divisor. That extension is quite old and ubiquitously supported on all platforms so the benefit of code simplification is worth it. Also added divisor into specialization for checking both input rate and step rate, on main the former wasn't checked (is it even possible for an attribute to go from vertex to instance rate with same shader?) and removed code in backend for handling InstanceId0-1 as it seemed wrong; if those vgprs are used for as index for buffer loads shouldn't they be InstanceId / step_rate_0-1 instead of just step rate? Can be checked now when their assertion hits

### DIFF
--- a/src/shader_recompiler/specialization.h
+++ b/src/shader_recompiler/specialization.h
@@ -13,7 +13,6 @@
 namespace Shader {
 
 struct VsAttribSpecialization {
-    s32 num_components{};
     u32 divisor{};
     AmdGpu::NumberClass num_class{};
     AmdGpu::CompMapping dst_select{};
@@ -96,7 +95,6 @@ struct StageSpecialization {
             // Specialize shader on VS input number types to follow spec.
             ForEachSharp(vs_attribs, fetch_shader_data->attributes,
                          [&profile_, this](auto& spec, const auto& desc, AmdGpu::Buffer sharp) {
-                             spec.num_components = AmdGpu::NumComponents(sharp.GetDataFmt());
                              using InstanceIdType = Shader::Gcn::VertexAttribute::InstanceIdType;
                              if (const auto step_rate = desc.GetStepRate();
                                  step_rate != InstanceIdType::None) {
--- a/src/shader_recompiler/specialization.h
+++ b/src/shader_recompiler/specialization.h
@@ -14,6 +14,7 @@ namespace Shader {
 
 struct VsAttribSpecialization {
     s32 num_components{};
+    u32 divisor{};
     AmdGpu::NumberClass num_class{};
     AmdGpu::CompMapping dst_select{};
 
@@ -74,13 +75,13 @@ struct SamplerSpecialization {
  * after the first compilation of a module.
  */
 struct StageSpecialization {
-    static constexpr size_t MaxStageResources = 64;
+    static constexpr size_t MaxStageResources = 128;
 
     const Shader::Info* info;
     RuntimeInfo runtime_info;
+    std::bitset<MaxStageResources> bitset{};
     std::optional<Gcn::FetchShaderData> fetch_shader_data{};
     boost::container::small_vector<VsAttribSpecialization, 32> vs_attribs;
-    std::bitset<MaxStageResources> bitset{};
     boost::container::small_vector<BufferSpecialization, 16> buffers;
     boost::container::small_vector<ImageSpecialization, 16> images;
     boost::container::small_vector<FMaskSpecialization, 8> fmasks;
@@ -94,10 +95,17 @@ struct StageSpecialization {
         if (info_.stage == Stage::Vertex && fetch_shader_data) {
             // Specialize shader on VS input number types to follow spec.
             ForEachSharp(vs_attribs, fetch_shader_data->attributes,
-                         [&profile_](auto& spec, const auto& desc, AmdGpu::Buffer sharp) {
-                             spec.num_components = desc.UsesStepRates()
-                                                       ? AmdGpu::NumComponents(sharp.GetDataFmt())
-                                                       : 0;
+                         [&profile_, this](auto& spec, const auto& desc, AmdGpu::Buffer sharp) {
+                             spec.num_components = AmdGpu::NumComponents(sharp.GetDataFmt());
+                             using InstanceIdType = Shader::Gcn::VertexAttribute::InstanceIdType;
+                             if (const auto step_rate = desc.GetStepRate();
+                                 step_rate != InstanceIdType::None) {
+                                 spec.divisor = step_rate == InstanceIdType::OverStepRate0
+                                                    ? runtime_info.vs_info.step_rate_0
+                                                    : (step_rate == InstanceIdType::OverStepRate1
+                                                           ? runtime_info.vs_info.step_rate_1
+                                                           : 1);
+                             }
                              spec.num_class = profile_.support_legacy_vertex_attributes
                                                   ? AmdGpu::NumberClass{}
                                                   : AmdGpu::GetNumberClass(sharp.GetNumberFmt());
--- a/src/video_core/renderer_vulkan/vk_instance.cpp
+++ b/src/video_core/renderer_vulkan/vk_instance.cpp
@@ -277,6 +277,7 @@ bool Instance::CreateDevice() {
     image_load_store_lod = add_extension(VK_AMD_SHADER_IMAGE_LOAD_STORE_LOD_EXTENSION_NAME);
     amd_gcn_shader = add_extension(VK_AMD_GCN_SHADER_EXTENSION_NAME);
     amd_shader_trinary_minmax = add_extension(VK_AMD_SHADER_TRINARY_MINMAX_EXTENSION_NAME);
+    vertex_attribute_divisor = add_extension(VK_EXT_VERTEX_ATTRIBUTE_DIVISOR_EXTENSION_NAME);
     shader_atomic_float2 = add_extension(VK_EXT_SHADER_ATOMIC_FLOAT_2_EXTENSION_NAME);
     if (shader_atomic_float2) {
         shader_atomic_float2_features =
@@ -436,6 +437,9 @@ bool Instance::CreateDevice() {
         vk::PhysicalDeviceLegacyVertexAttributesFeaturesEXT{
             .legacyVertexAttributes = true,
         },
+        vk::PhysicalDeviceVertexAttributeDivisorFeatures{
+            .vertexAttributeInstanceRateDivisor = true,
+        },
         vk::PhysicalDeviceShaderAtomicFloat2FeaturesEXT{
             .shaderBufferFloat32AtomicMinMax =
                 shader_atomic_float2_features.shaderBufferFloat32AtomicMinMax,
@@ -497,6 +501,9 @@ bool Instance::CreateDevice() {
     }
     if (!legacy_vertex_attributes) {
         device_chain.unlink<vk::PhysicalDeviceLegacyVertexAttributesFeaturesEXT>();
+    }
+    if (!vertex_attribute_divisor) {
+        device_chain.unlink<vk::PhysicalDeviceVertexAttributeDivisorFeatures>();
     }
     if (!shader_atomic_float2) {
         device_chain.unlink<vk::PhysicalDeviceShaderAtomicFloat2FeaturesEXT>();
--- a/src/video_core/renderer_vulkan/vk_instance.cpp
+++ b/src/video_core/renderer_vulkan/vk_instance.cpp
@@ -248,6 +248,7 @@ bool Instance::CreateDevice() {
     // Required
     ASSERT(add_extension(VK_KHR_SWAPCHAIN_EXTENSION_NAME));
     ASSERT(add_extension(VK_KHR_PUSH_DESCRIPTOR_EXTENSION_NAME));
+    ASSERT(add_extension(VK_EXT_VERTEX_ATTRIBUTE_DIVISOR_EXTENSION_NAME));
 
     // Optional
     depth_range_unrestricted = add_extension(VK_EXT_DEPTH_RANGE_UNRESTRICTED_EXTENSION_NAME);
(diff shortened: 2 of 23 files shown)


src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -277,6 +277,7 @@ bool Instance::CreateDevice() {
     image_load_store_lod = add_extension(VK_AMD_SHADER_IMAGE_LOAD_STORE_LOD_EXTENSION_NAME);
     amd_gcn_shader = add_extension(VK_AMD_GCN_SHADER_EXTENSION_NAME);
     amd_shader_trinary_minmax = add_extension(VK_AMD_SHADER_TRINARY_MINMAX_EXTENSION_NAME);
+    vertex_attribute_divisor = add_extension(VK_EXT_VERTEX_ATTRIBUTE_DIVISOR_EXTENSION_NAME);
```
squidbus: Maybe should just add this to the required extensions section (with assert), since we assume it is supported and never check or fallback.

src/shader_recompiler/specialization.h
```diff
@@ -94,10 +95,17 @@ struct StageSpecialization {
         if (info_.stage == Stage::Vertex && fetch_shader_data) {
             // Specialize shader on VS input number types to follow spec.
             ForEachSharp(vs_attribs, fetch_shader_data->attributes,
-                         [&profile_](auto& spec, const auto& desc, AmdGpu::Buffer sharp) {
-                             spec.num_components = desc.UsesStepRates()
-                                                       ? AmdGpu::NumComponents(sharp.GetDataFmt())
-                                                       : 0;
+                         [&profile_, this](auto& spec, const auto& desc, AmdGpu::Buffer sharp) {
+                             spec.num_components = AmdGpu::NumComponents(sharp.GetDataFmt());
```
squidbus: This was specialized on before specifically because the input param initialization code for step rate inputs used it, but that code is gone now. Do we still need to specialize on it?

raphaelthegreat: Indeed looks like this is unused now

squidbus (approved):
LGTM, should probably have some regression testing to be sure.

Missake212:
Breaks Bloodborne:

``[Debug] <Critical> emit_spirv_context_get_set.cpp:288 EmitGetAttributeU32: Unreachable code!``
``Read U32 attribute <reserved attribute 79>``

[CUSA03173.log](https://github.com/user-attachments/files/21204844/CUSA03173.log)

raphaelthegreat:
> Breaks Bloodborne:
> 
> `[Debug] <Critical> emit_spirv_context_get_set.cpp:288 EmitGetAttributeU32: Unreachable code!` `Read U32 attribute <reserved attribute 79>`
> 
> [CUSA03173.log](https://github.com/user-attachments/files/21204844/CUSA03173.log)

Looks like those are read due to phi node stuff and not because shader actually needs it :/

Missake212:
> > Breaks Bloodborne:
> > `[Debug] <Critical> emit_spirv_context_get_set.cpp:288 EmitGetAttributeU32: Unreachable code!` `Read U32 attribute <reserved attribute 79>`
> > [CUSA03173.log](https://github.com/user-attachments/files/21204844/CUSA03173.log)
> 
> Looks like those are read due to phi node stuff and not because shader actually needs it :/

I didn't finish testing my library but Dead Island - Definitive Edition hit the same UC (if that helps)

raphaelthegreat:
I've re-added these, now step rates are considered as part of shader key and no longer have a special attribute, translator just takes normal instance id and divides it over step rate

raphaelthegreat:
Hmm looks like there are cases where step rate register is not initialized and it's 0, how should that be handled? (Though it's still not actually used by the shader, rather not have a division by zero be executed)

EDIT: For now just skipping setting the attribute as it's super unlikely its actually needed when step rate is invalid (0)

squidbus:
Is step rate 0 invalid? Doesn’t that mean repeat the attribute for all instances?

raphaelthegreat:
That is how the extension defines it yes, but I'm not sure if it applies to hardware. Neither radv or radeonsi use these registers to implement divisors, but do it in software probably to bypass the limitation of only 2 available registers

squidbus:
> That is how the extension defines it yes, but I'm not sure if it applies to hardware.

Well step rate is the rate at which to step forward in the vertex attributes, so presumably 0 would result in never stepping forward.

squidbus:
Basically assuming that's true, the Vulkan side handling should be fine and I think in IR we should set the respective instance ID VGPRs to 0 since it would never advance.

raphaelthegreat:
Still doesn't make much sense to me what "increment index every 0 instances" means but yeah setting it to zero is probably safest bet

red-prig:
0 literally means that it will always be the same value equal to the first element

squidbus:
Gave this a quick check for validation errors btw and got the following:
* `vertexBindingDivisorCount` in the pipeline create info can't be 0, I think the whole divisor state struct needs to be omitted in this case:
```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-VkPipelineVertexInputDivisorStateCreateInfoKHR-vertexBindingDivisorCount-arraylength: Validation Error: [ VUID-VkPipelineVertexInputDivisorStateCreateInfoKHR-vertexBindingDivisorCount-arraylength ] | MessageID = 0x2163c15d | vkCreateGraphicsPipelines(): pCreateInfos[0].pVertexInputState->pNext<VkPipelineVertexInputDivisorStateCreateInfoKHR>.vertexBindingDivisorCount must be greater than 0.
The Vulkan spec states: vertexBindingDivisorCount must be greater than 0 (https://vulkan.lunarg.com/doc/view/1.3.296.0/mac/1.3-extensions/vkspec.html#VUID-VkPipelineVertexInputDivisorStateCreateInfoKHR-vertexBindingDivisorCount-arraylength)
```
* Seems to be invalid to be pushing a `vk::VertexInputBindingDivisorDescriptionEXT` for the attributes that are not instanced:
```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-VkVertexInputBindingDivisorDescriptionKHR-inputRate-01871: Validation Error: [ VUID-VkVertexInputBindingDivisorDescriptionKHR-inputRate-01871 ] | MessageID = 0x3451c35 | vkCreateGraphicsPipelines(): pCreateInfos[0].pVertexInputState->pNext<VkVertexInputBindingDivisorDescriptionKHR>.pVertexBindingDivisors[3].binding is 3, but inputRate is not VK_VERTEX_INPUT_RATE_INSTANCE.
The Vulkan spec states: VkVertexInputBindingDescription::inputRate must be of type VK_VERTEX_INPUT_RATE_INSTANCE for this binding (https://vulkan.lunarg.com/doc/view/1.3.296.0/mac/1.3-extensions/vkspec.html#VUID-VkVertexInputBindingDivisorDescriptionKHR-inputRate-01871)
```


```diff
diff --git a/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp b/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
index e971a00f..49f55fa9 100644
--- a/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
+++ b/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
@@ -86,7 +86,7 @@ GraphicsPipeline::GraphicsPipeline(
     };
 
     const vk::PipelineVertexInputStateCreateInfo vertex_input_info = {
-        .pNext = &divisor_state,
+        .pNext = divisor_state.vertexBindingDivisorCount > 0 ? &divisor_state : nullptr,
         .vertexBindingDescriptionCount = static_cast<u32>(vertex_bindings.size()),
         .pVertexBindingDescriptions = vertex_bindings.data(),
         .vertexAttributeDescriptionCount = static_cast<u32>(vertex_attributes.size()),
@@ -343,7 +343,7 @@ void GraphicsPipeline::GetVertexInputs(
                 : (attrib.GetStepRate() == InstanceIdType::OverStepRate1 ? step_rate_1 : 1);
         if constexpr (std::is_same_v<Binding, vk::VertexInputBindingDescription2EXT>) {
             bindings.back().divisor = divisor;
-        } else {
+        } else if (attrib.GetStepRate() != InstanceIdType::None) {
             divisors.push_back(vk::VertexInputBindingDivisorDescriptionEXT{
                 .binding = attrib.semantic,
                 .divisor = divisor,
```

raphaelthegreat:
Should be resolved now

squidbus:
Yeah looks good in some basic testing on my end.

Randomuser8219:
image
Knack has no regressions

Missake212:
Bloodborne is still broken on my end but nothing shows up in the log.

image

red-prig:
I'll clarify what I was talking about the Vulkan behavior for step_rate = 0 inside the extension, I have doubts that for AMD VGT_INSTANCE_STEP_RATE_0 it works the same way, need to test it

raphaelthegreat:
Regression should be fixed now. This PR exposed a bug in local shaders, where it was receiving instance_id in v1 instead of vs_rel_patch_id (see [here](https://github.com/chaotic-cx/mesa-mirror/blob/ee3d6aa87de02d03ff328c2a8cc961233d6824a8/src/amd/vulkan/radv_shader_args.c#L149) mesa has a separate path for vertex shader running on HW LS i.e info->vs.as_ls = true)

This didn't result in any actual rendering bugs, but due to that, DS_WRITE instruction offsets could not be constant propagated. Now where instance id was set to a constant, it was causing a crash when trying to retrieve producer instruction. Added an assert to make sure it's constant now 

Also cleaned up a few tessellation related stuff while at it

Missake212:
I don't have time to test my other games but Bloodborne is fixed.

image

raphaelthegreat:
@Missake212 Can you also test the snow in cainhurst when you get a chance, as that is a geometry shader effect iirc

raphaelthegreat:
Actually no tess effects should be tested like the fog portals iirc

Missake212:
Cainhurst works but it seems like a lot of stuff just doesn't show up anymore 

image

image

image

[CUSA03173.log](https://github.com/user-attachments/files/21205671/CUSA03173.log)

GHU7924:
I'm still in the process of testing, but I noticed this problem:

https://github.com/user-attachments/assets/7d740338-aab5-4e8c-8d46-0ea60ae1698f

[CUSA03173.log](https://github.com/user-attachments/files/21205740/CUSA03173.log)

Randomuser8219:
image
Knack is very flickery now.
[shad_log.txt](https://github.com/user-attachments/files/21205733/shad_log.txt)

Randomuser8219:
[shad_log.txt](https://github.com/user-attachments/files/21205742/shad_log.txt)
Here's another log with 33 validation errors.

raphaelthegreat:
Missing objects should be fixed now, was a small typo

Randomuser8219:
image
Looks better

raphaelthegreat:
> Looks better

Is is same as main now or still not the same?

Randomuser8219:
Same as main.

GHU7924:
>Missing objects should be fixed now, was a small typo

Yes, everything seems to be fine now. I didn't notice any more regressions.

Missake212:
Quickly tested a few of my games (the one that do go in game) and couldn't find any regressions.

StevenMiller123:
Specifically when readbacks is enabled, Bloodborne regresses from this PR.
From the sounds of it, this specifically affects AMD GPUs, though I haven't tested that part for myself.

Before:
image

[CUSA00900 Before.log](https://github.com/user-attachments/files/21277891/CUSA00900.Before.log)

After:
image

[CUSA00900 After.log](https://github.com/user-attachments/files/21277892/CUSA00900.After.log)

raphaelthegreat:
Please always enable validation when reporting these regressions, helps when checking the log

raphaelthegreat:
I cannot reproduce this on my Vega 8 with either RADV or AMD proprietary drivers on windows

StevenMiller123:
Here's a log with validation layers enabled. Doesn't seem like any validation errors come up ingame though.
[CUSA00900.log](https://github.com/user-attachments/files/21297982/CUSA00900.log)

raphaelthegreat:
Could it be possible this is due to a new driver update? That looks like a bugged buffer->image copy which this PR does not touch

StevenMiller123:
All the tests I performed for this regression were on the same driver version, so I'm not sure that's the case.
I'll test on Linux to see if I can reproduce this issue with radv.

StevenMiller123:
The issue isn't present on radv, so this is probably another bug related to AMD proprietary drivers.
Though the fact remains that I can only reproduce the bug on shadPS4 builds with this PR merged, before this PR the bug never showed.

raphaelthegreat:
Someone with affected hardware will have to examine a capture and see where it goes wrong then

UltraDaCat:
This is my RX 550 on Windows with Readbacks enabled and newest official polaris drivers(25.5.1 in this case), cannot reproduce the issue
image

Missake212:
Letting you know here so you're aware of it, seems like this PR regresses Bloodborne, some of the trees in Forbidden Woods don't render anymore (I'm not sure if they're tessellated trees or not).

Commit before merge:

469562388-8a00a1ff-9660-4300-9c0b-f7fb7406fba5

Commit where it was merged:

469563175-b9701d0f-1515-4a94-bd88-781cc97c11ce

Log with validation but it doesn't seem like it shows anything relevant to me:

[CUSA03173withValidation.log](https://github.com/user-attachments/files/21391965/CUSA03173.log)

raphaelthegreat:
I'll probably need a save file for this

Missake212:
> I'll probably need a save file for this

[SPRJ0005.zip](https://github.com/user-attachments/files/21392653/SPRJ0005.zip)

PR #3240, opened 2025-07-13
Volume slider that adjusts how loud games are on a global level

### LINKED ISSUE #3052: [Feature Request]: Audio Controls
### Checklist

### Description

There should be some sort of settings to adjust or enhance audio.

### Reason

It will improve the audibility in some games where the audio level is way too low for some reason.

### Examples

_No response_

UltraDaCat:
This adds an option to adjust volume for games on a global level, range goes from 0-500% with the default being 100%. This allows you to lower the volume of games that are loud by default(for example, Sonic Mania), increase it in games that are quiet(such as Persona 3: Dancing in Moonlight) and mute the game altogether by setting it to 0%. There's a volume slider in QT under the General tab, but you can also adjust the volume by editing the value of volumeSlider in the config.toml file. From my testing everything works properly, but I'd appreciate it if more people tested this just to make sure.

I wanna thank @kalaposfos13 @tomboylover93 @ConnorG512 and @rainmakerv3 for helping me with this!


rainmakerv3:
This is great, would be even better if it worked ingame.

Something like this is working https://github.com/rainmakerv3/shadPS4/actions/runs/16254276312, but probably not the best way to handle calling those member functions, maybe you can work with other people to make something better, but the general idea is there.

UltraDaCat:
I was thinking of having an option to adjust volume while ingame through imgui(similar to how you can adjust brightness as shown in the image below), but I did not understand it all too well. Thought I should at least PR this in its current state while I'm still trying to figure out the ins and outs of imgui.
image

rainmakerv3:
I think that menu isn't raelly meant for the general user, it's more a dev/power-user sort of thing. Most users still probably only look at the Qt GUI.

But if you want non-Qt users to be able to adjust volume ingame too, then it's also possible to implement in-game volume adjustment in both interfaces. I do think the Qt side is more important though

rainmakerv3:
I did something for the adjustable volume that may be good enough for this PR, I PR'd it to your branch in case you wanna use it

src/qt_gui/settings_dialog.ui
```diff
@@ -86,135 +86,77 @@
...
-               <property name="topMargin">
-                <number>9</number>
-               </property>
-               <property name="rightMargin">
-                <number>9</number>
+            <layout class="QVBoxLayout" name="volumeLayout">
+             <item>
+              <widget class="QScrollArea" name="scrollArea">
+               <property name="toolTip">
+                <string>Test 1</string>
```
kalaposfos13: I don't think you meant to leave this in

kalaposfos13 (changes requested):

kalaposfos13 (changes requested):
The slider is slightly too long and goes outside of the panel
Image

UltraDaCat:
will take a look at it right now

UltraDaCat:
@kalaposfos13 could you try it now?

kalaposfos13 (approved):

georgemoralis:
will merged it once actions get fixed

georgemoralis:
https://github.com/georgemoralis/shadPS4/actions/runs/16365816097

PR #3243, opened 2025-07-14
Implement sceKernelGetSystemSwVersion

kalaposfos13:
This function is used by LLE HTTP. From comparing the behaviour on console and on the emulator with Driveclub on mikusp's network PR + some hacks + hardcoded values to get it to connect to my local server, it appears to be working fine.
The LLE library still works without this function, it would simply use whatever memory garbage was in the return struct pointer's address instead.

### DIFF
--- a/src/core/libraries/kernel/kernel.cpp
+++ b/src/core/libraries/kernel/kernel.cpp
@@ -243,6 +243,19 @@ s32 PS4_SYSV_ABI sceKernelSetGPO() {
     return ORBIS_OK;
 }
 
+s32 PS4_SYSV_ABI sceKernelGetSystemSwVersion(SwVersionStruct* ret) {
+    if (ret == nullptr) {
+        return ORBIS_OK; // but why?
+    }
+    ASSERT(ret->struct_size == 40);
+    u32 fake_fw = 0x09990000; // chosen arbitrarily, it doesn't matter much
+    ret->hex_representation = fake_fw;
+    std::snprintf(ret->text_representation, 28, "%2x.%03x.%03x", fake_fw >> 0x18,
+                  fake_fw >> 0xc & 0xfff, fake_fw & 0xfff); // why %2x?
+    LOG_INFO(Lib_Kernel, "called, returned sw version: {}", ret->text_representation);
+    return ORBIS_OK;
+}
+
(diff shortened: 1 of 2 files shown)


StevenMiller123:
For proper compatibility, I think the firmware version should be modifiable.
At the least, it should be a real firmware version, and preferrably a higher firmware like 11.00 or 12.02.

src/core/libraries/kernel/kernel.cpp
```diff
@@ -243,6 +243,19 @@ s32 PS4_SYSV_ABI sceKernelSetGPO() {
     return ORBIS_OK;
 }
 
+s32 sceKernelGetSystemSwVersion(SwVersionStruct* ret) {
```
StevenMiller123: You forgot to declare `PS4_SYSV_ABI`

StevenMiller123 (changes requested):

mikusp:
It should rather be derived from the provided sprxs, some libraries might use a different ABI depending on the reported FW version

kalaposfos13:
It isn't that complex, it's only used in one single place in LLE HTTP and nowhere else so far, which is in the headers in HTTP requests like this (example from Driveclub):
```
POST /egs/binary/indexService/getMessageDescriptors HTTP/1.1
X-Api-Version: 2
X-Request-Signature: /XKVKkZ7V8g6Q7QSVmVvI1RKXfI=
Host: 192.168.1.66:10060
X-Client-Role: game
X-Client-Version: 2
X-Client-Title: EVO11
User-Agent: OrbisHttpClient/1.0 libhttp/<ONLY USED HERE AS A STRING> (PlayStation 4)
Connection: Keep-Alive
Content-Length: 0
```

StevenMiller123:
The problem is that other libraries or games might use this function, and emulating it properly will prevent potentially hard to find issues in these cases.

StevenMiller123:
And I agree with acetone, if it is reasonable to do, then we should probably derive this from the user-provided firmware libraries instead of hardcoding it or making a config option.

StevenMiller123 (approved):
Good enough for now, but using the firmware version from user-provided firmware libraries would still be preferred (when they're provided).

PR #3020, opened 2025-05-31
Handle pipelines with mixed samples counts

mikusp:
Fixes a Device lost error encountered in WIPEOUT: Omega Collection and HEAVY RAIN:
```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-vkCmdDrawIndexed-multisampledRenderToSingleSampled-07285: Validation Error: [ VUID-vkCmdDrawIndexed-multisampledRenderToSingleSampled-07285 ] Object 0: handle = 0xfa17400, name = CommandPool: Command Buffer 0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xe9b2ee0000000094, name = Graphics Pipeline fs_0x000000004f6cc41b,vs_0x000000002012d5ea, type = VK_OBJECT_TYPE_PIPELINE; | MessageID = 0xf4b969ec | vkCmdDrawIndexed(): Color attachment (0) sample count (VK_SAMPLE_COUNT_1_BIT) must match corresponding VkPipelineMultisampleStateCreateInfo sample count (VK_SAMPLE_COUNT_4_BIT).
The Vulkan spec states: If the bound pipeline was created without a VkAttachmentSampleCountInfoAMD or VkAttachmentSampleCountInfoNV structure, and the multisampledRenderToSingleSampled feature is not enabled, and the current render pass instance was begun with vkCmdBeginRendering with a VkRenderingInfo::colorAttachmentCount parameter greater than 0, then each element of the VkRenderingInfo::pColorAttachments array with a imageView not equal to VK_NULL_HANDLE must have been created with a sample count equal to the value of rasterizationSamples for the bound graphics pipeline (https://vulkan.lunarg.com/doc/view/1.4.304.1/windows/antora/spec/latest/chapters/drawing.html#VUID-vkCmdDrawIndexed-multisampledRenderToSingleSampled-07285)
```

According to gpuinfo, `multisampledRenderToSingleSampled` is supported by only a handful of mobile GPUs, while the other two are supported by most modern cards from the respective company.

PR #3245, opened 2025-07-15
shader_recompiler: Implement guest barycentrics

raphaelthegreat:
This is second part of attribute restructuring and aims to do the following:

* Further cleanup and de-duplicate attribute handling code
* Move as much logic into IR level (such as manual interpolation)
* Implement manual interpolation on the guest shader

For the first goal the following changes are included:
* De-duplicated access to TessellationEvaluationPointU/V
* ~~Removed frag_outputs in favor of output_params as they are the same thing~~
* Make usage of GetAttributeInfo  consistent everywhere
* Implement remaining attribute qualifiers such as centroid and sample
* Implement reading of param array attributes to merge geometry shader special path

The last change is important as it allows for the second goal, moving the manual interpolation NVIDIA path into IR. GCN ISA has special instructions V_INTERP_P1_F32/V_INTERP_P2_F32 for interpolating attributes and those can be used to implement that path.

Manual interpolation also requires the IR to become aware of barycentric coordinate variables. In case of an AMD host GPU, the VK_AMD_shader_explicit_vertex_parameter extension has been added, which maps 1-1 to guest functionality. For NVIDIA and other platforms the fragment barycentric extension can also provide most of the functionality necessary (though needs a small remap of barycentric coord). If barycentrics are not supported at all, frontend will request a flat qualifier instead, which is a fallback to previous behavior.

This fixes the menu screen of The Order 1886 for both AMD and NVIDIA GPUs (though reaching the title screen technically requires https://github.com/shadps4-emu/shadPS4/pull/3020 which only works on AMD properly. Regardless I have verified it on NVIDIA as well)


raphaelthegreat:
Requires https://github.com/shadps4-emu/sirit/pull/15

src/shader_recompiler/backend/spirv/emit_spirv.cpp
```diff
@@ -293,9 +293,17 @@ void SetupCapabilities(const Info& info, const Profile& profile, EmitContext& ct
...
     }
-    if (info.stage == Stage::Fragment && profile.needs_manual_interpolation) {
-        ctx.AddExtension("SPV_KHR_fragment_shader_barycentric");
-        ctx.AddCapability(spv::Capability::FragmentBarycentricKHR);
+    if (info.stage == Stage::Fragment) {
+        if (profile.supports_fragment_shader_barycentric) {
+            ctx.AddExtension("SPV_KHR_fragment_shader_barycentric");
+            ctx.AddCapability(spv::Capability::FragmentBarycentricKHR);
+        } else if (profile.supports_amd_shader_explicit_vertex_parameter) {
+            ctx.AddExtension("SPV_AMD_shader_explicit_vertex_parameter");
```
squidbus: Going by the PR description I would have expected this to be the first if case here to prefer it?

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -159,81 +159,67 @@ Id EmitReadConstBuffer(EmitContext& ctx, u32 handle, Id index) {
...
+                          ctx.OpAccessChain(ctx.input_f32, ctx.frag_coord, ctx.ConstU32(comp)));
     case IR::Attribute::TessellationEvaluationPointU:
         return ctx.OpLoad(ctx.F32[1],
                           ctx.OpAccessChain(ctx.input_f32, ctx.tess_coord, ctx.u32_zero_value));
     case IR::Attribute::TessellationEvaluationPointV:
         return ctx.OpLoad(ctx.F32[1],
                           ctx.OpAccessChain(ctx.input_f32, ctx.tess_coord, ctx.ConstU32(1U)));
+    case IR::Attribute::BaryCoordSmooth:
+        if (ctx.profile.supports_fragment_shader_barycentric) {
+            ++comp;
```
squidbus: Maybe I'm missing something, but why are these being skipped if the extension *is* supported?

And shouldn't these also be checking the AMD extension support?

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
```diff
@@ -398,49 +374,71 @@ void EmitContext::DefineInputs() {
...
             Id attr_id{};
-            if (profile.needs_manual_interpolation && !input.is_flat) {
-                attr_id = DefineInput(TypeArray(type, ConstU32(3U)), input.param_index);
-                Decorate(attr_id, spv::Decoration::PerVertexKHR);
+            // Declare primary qualifier (pervertex, smooth, noperspective, flat)
+            if (is_array) {
+                if (profile.supports_fragment_shader_barycentric) {
+                    attr_id = DefineInput(TypeArray(type, ConstU32(3U)), input.param_index);
+                    Decorate(attr_id, spv::Decoration::PerVertexKHR);
+                } else {
```
squidbus: Same here with regards to if-else condition order. Although here it's also the case that if neither is supported, you're still applying the AMD extension decoration.

src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -274,7 +274,12 @@ bool Instance::CreateDevice() {
     depth_clamp_control = add_extension(VK_EXT_DEPTH_CLAMP_CONTROL_EXTENSION_NAME);
     vertex_input_dynamic_state = add_extension(VK_EXT_VERTEX_INPUT_DYNAMIC_STATE_EXTENSION_NAME);
     list_restart = add_extension(VK_EXT_PRIMITIVE_TOPOLOGY_LIST_RESTART_EXTENSION_NAME);
-    fragment_shader_barycentric = add_extension(VK_KHR_FRAGMENT_SHADER_BARYCENTRIC_EXTENSION_NAME);
+    amd_shader_explicit_vertex_parameter =
+        add_extension(VK_AMD_SHADER_EXPLICIT_VERTEX_PARAMETER_EXTENSION_NAME);
+    if (!amd_shader_explicit_vertex_parameter) {
+        fragment_shader_barycentric =
+            add_extension(VK_KHR_FRAGMENT_SHADER_BARYCENTRIC_EXTENSION_NAME);
```
squidbus, line 281: This needs to be force disabled for MoltenVK driver ID, because they advertise the extension but specifically `PerVertexKHR` is not supported.

squidbus: Actually I guess it might need some kind of secondary support variable to the shader recompiler as stuff like `BaryCoordKHR` does work and could be useful if the shader wants it.

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
```diff
@@ -398,49 +374,71 @@ void EmitContext::DefineInputs() {
...
+                bary_coord_smooth = DefineVariable(F32[2], spv::BuiltIn::BaryCoordSmoothAMD,
+                                                   spv::StorageClass::Input);
+            } else {
+                bary_coord_smooth =
+                    DefineVariable(F32[3], spv::BuiltIn::BaryCoordKHR, spv::StorageClass::Input);
+            }
+        }
+        if (info.loads.GetAny(IR::Attribute::BaryCoordSmoothSample)) {
+            if (profile.supports_amd_shader_explicit_vertex_parameter) {
+                bary_coord_smooth = DefineVariable(F32[2], spv::BuiltIn::BaryCoordSmoothSampleAMD,
```
squidbus:
```suggestion
                bary_coord_smooth_sample = DefineVariable(F32[2], spv::BuiltIn::BaryCoordSmoothSampleAMD,
```

src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
raphaelthegreat: The comp is incremented because gl_BaryCoordSmoothAMD.xy maps to gl_BaryCoordKHR.yz. I should probably add a check for other extension though yes

src/shader_recompiler/backend/spirv/emit_spirv.cpp
raphaelthegreat: The instance will never load both extensions so checking for one automatically excludes the other. But I can change the order regardless if you want

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
raphaelthegreat: The frontend will not request pervertex qualifier if at least one extension is not supported, however adding the explicit won't hurt either, I will add it

src/shader_recompiler/backend/spirv/emit_spirv.cpp
squidbus: I usually would explicitly just have it in order of preferred extension in case but don't really mind much if that's the case. Feel free to switch around if you want or just resolve the related comments.

src/video_core/renderer_vulkan/vk_instance.cpp
raphaelthegreat: Added has_incomplete_fragment_shader_barycentric member to profile which blocks frontend from adding pervertex inputs for MoltenVK, but barycentrics are still properly defined

squidbus:
Only done some light testing so far but looks good to me, glad to see all this attribute stuff getting cleaned up and completed.

Randomuser8219:
Namco Museum Archives crashes with:
[Debug] <Critical> vector_interpolation.cpp:64 operator(): Assertion Failed!

zezoman:
This PR breaks Bloodborne for me, crash to desktop when loading a safe.
[shad_log_last_working_shadps4-win64-qt-2025-07-15-87f6cce.txt](https://github.com/user-attachments/files/21250591/shad_log_last_working_shadps4-win64-qt-2025-07-15-87f6cce.txt)
[shad_log_breaking_PR_shadps4-win64-qt-2025-07-15-4407ebd.txt](https://github.com/user-attachments/files/21250593/shad_log_breaking_PR_shadps4-win64-qt-2025-07-15-4407ebd.txt)
Every build in main after this PR is broken the same way.

squidbus:
Noticed some of these validation errors today opening up Gravity Rush 2:
```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-VkShaderModuleCreateInfo-pCode-08737: Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08737 ] | MessageID = 0xa5625282 | vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
Operand 2 of Decorate requires one of these capabilities: SampleRateShading 
  OpDecorate %fs_in_attr0 Sample
The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://vulkan.lunarg.com/doc/view/1.3.296.0/mac/1.3-extensions/vkspec.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
```

Seems that some inputs are getting decorated with `Sample` without the `SampleRateShading` capability applying. Maybe the attribute loads that are used to set the capability are optimized out?

brad0demx:
Since this PR was merged, Rocket League now crashes with 
`[Debug] <Critical> vector_interpolation.cpp:64 operator(): Assertion Failed!`
[CUSA01433.log](https://github.com/user-attachments/files/21322667/CUSA01433.log)

StevenMiller123:
This PR appears to break a variety of games when testing with Intel GPUs.
They all crash on some validation error that looks like
```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-VkShaderModuleCreateInfo-pCode-08737: vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
The Base <id> '45[%45]' in OpAccessChain instruction must be a pointer.
  %69 = OpAccessChain %input_f32 %45 %u32_id_0
The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://vulkan.lunarg.com/doc/view/1.4.313.2/windows/antora/spec/latest/chapters/shaders.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
```

[CUSA00803-before.log](https://github.com/user-attachments/files/21379699/CUSA00803-before.log)
[CUSA00803-after.log](https://github.com/user-attachments/files/21379698/CUSA00803-after.log)

@Missake212 is better equipped to test this regression currently, and was the one who initially reported it on the shadPS4 Discord server.

bigol83:
This PR seems to have caused a regression with both Yakuza 0 and Burnout Paradise Remastered for me

Both crash with 

`[Debug] <Critical> vector_interpolation.cpp:64 operator(): Assertion Failed!`

Yakuza 0 crashes while trying to reach ingame, while Burnout Paradise Remastered crashes soon after boot

Here is Yakuza 0 log

[shad_log.txt](https://github.com/user-attachments/files/21418883/shad_log.txt)

PR #3246, opened 2025-07-15
Core: Fix read-only file unmaps on Windows

StevenMiller123:
By using the created `post_merge_vma` to check protections, `readonly_file` would always be false since the created VMA has prot `MemoryProt::NoAccess`. In games that try unmapping a read-only file, this would cause a crash for Windows users since we don't have a proper file mapping at that address.

Fixes Genshin Impact (CUSA23681) crashing on a memory error on boot. It doesn't get much further though, since it needs alertable event flag waits.


squidbus (approved):

palmeras:
dead or alive 5 last round se rompio despues del video ninja_vi.mp4 cuando iva a iniciarse el renderizado 3d

PR #3247, opened 2025-07-15
Libraries: Implement sceAudio3dTerminate

StevenMiller123:
My First Gran Turismo® (CUSA49696) uses this while initializing it's audio system. Not handling this results in sceAudio3dInitialize error spam, and would probably cause other issues if the game got further.


squidbus (commented):
If you reference the initialize function, there’s some other things that need to be de-initialized as well like the state’s AudioOut port.

squidbus (approved):

PR #3248, opened 2025-07-15
vk_instance: Remove usage of depth clamp control

raphaelthegreat:
AMD proprietary driver doesn't seem to properly implement the extension, not sure what the issue is as I can only test RADV which is fine

@StevenMiller123 Check if this fixes the regression, otherwise the extension itself might need to be excluded

### DIFF
--- a/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
+++ b/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
@@ -140,30 +140,13 @@ GraphicsPipeline::GraphicsPipeline(
         .sampleShadingEnable = false,
     };
 
-    const vk::DepthClampRangeEXT depth_clamp_range = {
-        .minDepthClamp = key.min_depth_clamp,
-        .maxDepthClamp = key.max_depth_clamp,
+    const vk::PipelineViewportDepthClipControlCreateInfoEXT clip_control = {
+        .negativeOneToOne = key.clip_space == Liverpool::ClipSpace::MinusWToW,
     };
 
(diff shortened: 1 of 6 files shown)


red-prig:
shouldn't `vk::DepthClampModeEXT::eViewportRange` work as if the `PipelineViewportDepthClampControlCreateInfoEXT` structure was missing? Then maybe we should test excluding this structure from the chain, instead of setting `vk::DepthClampModeEXT::eViewportRange`?

raphaelthegreat:
It should yes and this PR does test excluding the structure from the chain, since the driver doesn't seem to like it

StevenMiller123:
image
That definitely changed how it renders, but it's still regressed.

[CUSA47038.log](https://github.com/user-attachments/files/21236186/CUSA47038.log)

georgemoralis:
updated your AMD drivers?:P

raphaelthegreat:
I've changed it to blacklist the extension entirely. But also yes you should update the driver, they might have done some bugfixes on it

StevenMiller123:
For the record, I'm on the latest AMD drivers for this card
image

StevenMiller123:
image
Metaphor: ReFantazio works as expected on my setup now.

[CUSA47038.log](https://github.com/user-attachments/files/21236428/CUSA47038.log)

raphaelthegreat:
I am tempted to just remove the extension entirely given it's so new and no valid usecases of it are found.

raphaelthegreat:
Removed the extension entirely for now. Depth clip enable is fine though because it's used by DXVK

PR #3251, opened 2025-07-15
Fix ff1_i32_b64 not accepting vcc as its argument

mikusp:
Hit by DOOM CUSA02092


squidbus:
@raphaelthegreat Can you confirm this is still correct using vcc with the ballot stuff?

raphaelthegreat:
This is correct yes

squidbus (approved):

PR #3252, opened 2025-07-15
Avoid clearing HTILE when shader contains address calculation

raphaelthegreat:
This is the second attempt to implement https://github.com/shadps4-emu/shadPS4/pull/3167
I'm open to other ideas but that was the best heuristic I could come up with that should work in most cases

### DIFF
--- a/src/shader_recompiler/frontend/instruction.h
+++ b/src/shader_recompiler/frontend/instruction.h
@@ -3,8 +3,6 @@
 
 #pragma once
 
-#include <limits>
-#include "common/bit_field.h"
 #include "shader_recompiler/frontend/opcodes.h"
 
 namespace Shader::Gcn {
--- a/src/shader_recompiler/info.h
+++ b/src/shader_recompiler/info.h
@@ -222,6 +222,7 @@ struct Info {
     VAddr pgm_base;
     bool has_storage_images{};
     bool has_discard{};
+    bool has_bitwise_xor{};
     bool has_image_gather{};
     bool has_image_query{};
     bool uses_buffer_atomic_float_min_max{};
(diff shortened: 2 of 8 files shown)


diegolix29:
shadows in fromsoft games are ok with this, it dosent create weird shadows on any of them, but shadows and reflections on driveclub are wrong 
image
image

image

Nvidia GPU

raphaelthegreat:
This looks like a regression from barycentrics PR, but need to confirm

diegolix29:
yeah reflections seems to be from barycentrics 
image
this is before this pr

raphaelthegreat:
Will rebase this PR when the fix is merged https://github.com/shadps4-emu/shadPS4/pull/3253

diegolix29:
i tested locally and the shadows and reflections are ok 
image

image
image


only other regression found is that overlaped back layer on top of the car but its a regression from old prs

image

raphaelthegreat:
What PR regressed it?

diegolix29:
so i went locally in ur pr back to when readbacks where introduced and still got it, it may be a nvidia driver issue then, i will be testing old drivers

raphaelthegreat:
I can't reproduce this on NVIDIA linux 575 driver

diegolix29:
weird i tested on lastest and on december ones. but windows, i can jump on my linux bootable to see if its only windows related issue

diegolix29:
so i restarted my pc and with the lastest drivers nvidia on windows, its ok now. i think i was on the lastest but studio drivers, due to work. So thats not a issue then, thanks the pr is ok then.

squidbus (approved):

PR #3257, opened 2025-07-16
Add an unreachable on hitting ud2 instead of getting stuck in an infinite loop

kalaposfos13:


palmeras:
DOA5LR was broken after PR 985da0c, which is the last build it worked on. Since that build, it crashes after the ninja_vi.mp4 video, right when the DOA5LR main menu screen should be displayed. The error message is: [Debug] <Critical> emit_spirv_context_get_set.cpp:218 EmitGetAttribute: Unreachable code!
Read attribute BaryCoordSmoothCentroid

PR #3258, opened 2025-07-16
If CONTENT_ID is empty in param.sfo, try using TITLE_ID as fallback

kalaposfos13:
Honestly I have no idea why CONTENT_ID is used here in the first place, TITLE_ID is literally just the exact same thing except there's no need to use substring
Fixes a crash when trying to open firmware apps whose CONTENT_IDs are empty

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -100,8 +100,12 @@ void Emulator::Run(std::filesystem::path file, const std::vector<std::string> ar
 
         const auto content_id = param_sfo->GetString("CONTENT_ID");
         ASSERT_MSG(content_id.has_value(), "Failed to get CONTENT_ID");
-
-        id = std::string(*content_id, 7, 9);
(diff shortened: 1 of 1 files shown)


src/emulator.cpp
```diff
@@ -100,8 +100,12 @@ void Emulator::Run(std::filesystem::path file, const std::vector<std::string> ar
 
         const auto content_id = param_sfo->GetString("CONTENT_ID");
         ASSERT_MSG(content_id.has_value(), "Failed to get CONTENT_ID");
-
-        id = std::string(*content_id, 7, 9);
+        const auto title_id = param_sfo->GetString("TITLE_ID");
+        if (!content_id->empty()) {
+            id = std::string(*content_id, 7, 9);
+        } else if (content_id.has_value()) {
```
georgemoralis: maybe title_id.has_value() ?

PR #3259, opened 2025-07-16
Initialize the heap API and add new memory functions in HLE LibcInternal

kalaposfos13:
Applications that do not use their own libc.prx and instead rely on libSceLibcInternal will not initialize this, and will crash on the first call of any of these functions. This PR initializes the heap API so there's at least something for these apps to work with. I tested it with NPXS20001, and it seems to work, by comparing its behaviour to when I put a libc.prx from one of my games to its sce_module folder. However, since I'm not too knowledgeable on this topic, I'll open it as a draft for now.

### DIFF
--- a/src/core/libraries/libc_internal/libc_internal_memory.cpp
+++ b/src/core/libraries/libc_internal/libc_internal_memory.cpp
@@ -30,6 +30,43 @@ s32 PS4_SYSV_ABI internal_memcmp(const void* s1, const void* s2, size_t n) {
     return std::memcmp(s1, s2, n);
 }
 
+void* PS4_SYSV_ABI internal_malloc(size_t size) {
+    return std::malloc(size);
+}
+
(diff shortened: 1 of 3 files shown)


src/core/libraries/libc_internal/libc_internal_memory.cpp
```diff
@@ -30,6 +30,51 @@ s32 PS4_SYSV_ABI internal_memcmp(const void* s1, const void* s2, size_t n) {
     return std::memcmp(s1, s2, n);
 }
 
+void* PS4_SYSV_ABI internal_malloc(size_t size) {
+    return std::malloc(size);
+}
```
StevenMiller123, line 35: It would be more ideal to properly reimplement these instead of using host functions.
Most of these boil down to kernel memory functions if I remember right.

StevenMiller123 (changes requested):

PR #3261, opened 2025-07-16
texture_cache: Clamp buffer image height to microtile height

raphaelthegreat:
When 2D array textures are used that have mipmaps with dimentions less than a tile (< 8) the size calculation will pad both the pitch and the height of the slice to microtile dimentions. The texture cache upload code used the correct pitch, but not the aligned height. This caused issues as some inbetween slices would receive the padding image data and not the intended, which at best will look the same and at worst would poison the image with NaNs.

Fix this by clamping the mip height to microtile height. This should work for macrotile images as well, since the lower mipmaps are downgraded to microtiled tiling


raphaelthegreat:
Might affect https://github.com/shadps4-emu/shadps4-game-compatibility/issues/1184

StevenMiller123:
This fixes the weird metallic glow in Metaphor: ReFantazio.

Before:
image

After:
image

Note: Since main repo actions are disabled at the moment, I've used the latest main compiled through my fork instead. This build is identical to current main, with no additional changes.
[CUSA47038 Main.log](https://github.com/user-attachments/files/21267313/CUSA47038.Main.log)
[CUSA47038 PR.log](https://github.com/user-attachments/files/21267314/CUSA47038.PR.log)

squidbus (approved):

GHU7924:
It seems this has affected The Last Guardian.

TGL

P.S. If there is a correction, I won’t be able to check it again at the moment.

raphaelthegreat:
Hmm so aligning the mip height is not correct in all cases? In the initial version of the PR I implemented it as max(mip_height, 8) like the title says, but then changed it as it felt more correct to align up like size calculation does. Will investigate why this breaks and revert the change

raphaelthegreat:
The issue was with volume textures, aligning the height would cause slices to be offset downwards. I believe this is due to tile thickness, a tile can contain data for multiple slices (detile shader iirc hardcodes 4 for thickness) ~~which seems to eliminate the need for padding in image, making the final image data tightly packed after detiling, even though the dimentions are 36x36 which are not tile aligned~~ EDIT: They do have padding but it's not included in detiled buffer, see below comment

StevenMiller123:
Metaphor: ReFantazio is still fixed with this PR.
image

I'm not seeing any regressions on my end, though I haven't thoroughly tested my library yet.

georgemoralis:
will merged it , when we got actions fixed

raphaelthegreat:
Thinking it more I think the difference might be with how micro and macro detilers work. The macro detiler uses the current thread_id to index the [output](https://github.com/shadps4-emu/shadPS4/blob/fddded8d204374febcba210b8e2e325973f973b2/src/video_core/host_shaders/detilers/macro_32bpp.comp#L91) buffer, so it is not going to produce any padding, while the micro detiler uses it for [input](https://github.com/shadps4-emu/shadPS4/blob/fddded8d204374febcba210b8e2e325973f973b2/src/video_core/host_shaders/detilers/micro_32bpp.comp#L40) buffer so it's probably going to add the padding from the input

georgemoralis:
https://github.com/georgemoralis/shadPS4/actions/runs/16368065658

src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -605,36 +605,35 @@ void TextureCache::RefreshImage(Image& image, Vulkan::Scheduler* custom_schedule
...
+        const u32 extent_height = mip_height ? std::min(mip_height, height) : height;
+        const u32 height_aligned =
+            mip_height && image.info.IsTiled() ? std::max(mip_height, 8U) : mip_height;
 
         image_copy.push_back({
-            .bufferOffset = mip.offset,
+            .bufferOffset = mip_offset,
             .bufferRowLength = mip_pitch,
-            .bufferImageHeight = mip_height ? std::max(mip_height, 8U) : mip_height,
+            .bufferImageHeight = height_aligned,
```
squidbus, line 628: The change here in https://github.com/shadps4-emu/shadPS4/pull/3264, as well as the change here, seems to regress Sakura Wars (see https://github.com/shadps4-emu/shadPS4/pull/3264#issuecomment-3120385040)

raphaelthegreat: Can you check if the game uses a volume tiling for the bugged texture

squidbus: Yeah I addressed it in https://github.com/shadps4-emu/shadPS4/pull/3326 for now, it's a volume texture.

raphaelthegreat: Nice thanks, ideally in the future detilers should be made consistent in regards to padding

PR #3264, opened 2025-07-17
vk_pipeline_cache: Add fallbacks for R8Srgb and B5G6R5

raphaelthegreat:
Should resolve remaining daytime issues in Driveclub on NVIDIA

### DIFF
--- a/src/video_core/amdgpu/pixel_format.h
+++ b/src/video_core/amdgpu/pixel_format.h
@@ -248,6 +248,15 @@ constexpr CompMapping RemapSwizzle(const DataFormat format, const CompMapping sw
         result.a = swizzle.r;
         return result;
     }
+    case DataFormat::Format5_6_5: {
+        // Remap to a more supported component order.
+        CompMapping result;
+        result.r = swizzle.b;
+        result.g = swizzle.g;
+        result.b = swizzle.r;
+        result.a = swizzle.a;
+        return result;
+    }
     default:
         return swizzle;
     }
--- a/src/video_core/renderer_vulkan/vk_instance.cpp
+++ b/src/video_core/renderer_vulkan/vk_instance.cpp
@@ -70,7 +70,6 @@ std::unordered_map<vk::Format, vk::FormatProperties3> GetFormatProperties(
         vk::Format::eB8G8R8A8Unorm,
         vk::Format::eB8G8R8A8Srgb,
         vk::Format::eD24UnormS8Uint,
-        vk::Format::eR5G6B5UnormPack16,
     };
     for (const auto& format : misc_formats) {
         if (!format_properties.contains(format)) {
@@ -674,11 +673,6 @@ vk::Format Instance::GetSupportedFormat(const vk::Format format,
         case vk::Format::eR8Srgb:
             if (IsFormatSupported(vk::Format::eR8Unorm, flags)) {
                 return vk::Format::eR8Unorm;
-            }
-            break;
-        case vk::Format::eB5G6R5UnormPack16:
-            if (IsFormatSupported(vk::Format::eR5G6B5UnormPack16, flags)) {
-                return vk::Format::eR5G6B5UnormPack16;
             }
             break;
         default:
--- a/src/video_core/renderer_vulkan/vk_instance.cpp
+++ b/src/video_core/renderer_vulkan/vk_instance.cpp
@@ -70,6 +70,7 @@ std::unordered_map<vk::Format, vk::FormatProperties3> GetFormatProperties(
         vk::Format::eB8G8R8A8Unorm,
         vk::Format::eB8G8R8A8Srgb,
         vk::Format::eD24UnormS8Uint,
+        vk::Format::eR5G6B5UnormPack16,
     };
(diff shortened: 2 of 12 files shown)


diogoaduartereis:
Can confirm it is looking much better!
image

image

raphaelthegreat:
It appears for R8Srgb the game later requests an R8Unorm view of the image, so it ends up with mismatching bit width. For now I've changed the fallback formats to equivalents with same bpp, though for R8 one it lacks gamma correction and for the B5G6R5 it has different component order.

raphaelthegreat:
Also I realize now that the B5G6R5 change could regress cases where the format is used only as a texture and not render target. I see 3 options to solve this:
* Detect swapped format in RefreshImage and transform input data to correct layout
* Return alongside the fixed format, a swizzle so the image is read with correct channel order (I don't think this will work though when image is render target and is bound as texture later?)
* Add usage information to image info and check that cache image is superset of requested in exact match. So in case of just being a texture we can use the exact format and only for render targets it will use the swapped one

src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -669,6 +670,17 @@ vk::Format Instance::GetSupportedFormat(const vk::Format format,
             if (IsFormatSupported(vk::Format::eD32SfloatS8Uint, flags)) {
                 return vk::Format::eD32SfloatS8Uint;
             }
+            break;
+        case vk::Format::eR8Srgb:
+            if (IsFormatSupported(vk::Format::eR8Unorm, flags)) {
+                return vk::Format::eR8Unorm;
+            }
+            break;
+        case vk::Format::eB5G6R5UnormPack16:
```
squidbus: Don’t we need to re-add swizzle for this? I assume with the way the usage flags are configured it will always have color attachment and thus always fail support check, and so we will always be falling back. So then a game using a B5G6R5 texture would have backwards component order.

If so, I feel like it would be simpler to add this to the format/swizzle remapping layer to turn into a more commonly supported format and handle the swizzles.

squidbus: Okay I see while I was writing this you added a comment about it. I think you could handle this pretty easily by adding it to RemapSwizzle to turn it into R5G6B5 order for host graphics and using that format unconditionally.

raphaelthegreat: Implemented your suggestion, hopefully I didn't miss anything

squidbus:
It looks right to me, I can give it some testing in a few hours if you haven’t already.

raphaelthegreat:
Appears to work okay on my end, though it also worked before due to it being written only as render target. Would need to test something that uses it as a texture.

src/video_core/amdgpu/pixel_format.h
```diff
@@ -248,6 +248,14 @@ constexpr CompMapping RemapSwizzle(const DataFormat format, const CompMapping sw
         result.a = swizzle.r;
         return result;
     }
+    case DataFormat::Format5_6_5: {
+        CompMapping result;
```
squidbus, line 253: One minor nit here: add a comment with something like that we did this to remap to a component order with more usage flags supported by more GPUs, so anyone looking at this can understand why it’s remapped instead of using B5G6R5 directly.

squidbus (approved):
LGTM

georgemoralis:
https://github.com/georgemoralis/shadPS4/actions/runs/16367093505

StevenMiller123:
Seems like this PR regressed Sakura Wars? Based on my and @Missake212's tests, this affects all GPU vendors.

Before:
image

After:
image
(Due to the actions outage at the time, these are both tested from actions run through @georgemoralis's fork. Based on merge times, this PR should be the only thing merged between the two builds I tested.)

[CUSA16429 Before.log](https://github.com/user-attachments/files/21437635/CUSA16429.Before.log)
[CUSA16429 After.log](https://github.com/user-attachments/files/21437634/CUSA16429.After.log)

squidbus:
Are you sure it's this change? I reverted it locally and it still happens.

StevenMiller123:
It definitely seemed like it when testing precompiled builds.
I'll do some local testing to see if I come up with a different result then.

squidbus:
It does seem that it started here if I go back in history, but reverting just this on main does not seem to resolve it.

src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -626,7 +626,7 @@ void TextureCache::RefreshImage(Image& image, Vulkan::Scheduler* custom_schedule
         image_copy.push_back({
             .bufferOffset = mip.offset,
             .bufferRowLength = mip_pitch,
-            .bufferImageHeight = mip_height,
+            .bufferImageHeight = mip_height ? std::max(mip_height, 8U) : mip_height,
```
squidbus, line 629: The issue seems to be this, followed by the changes in https://github.com/shadps4-emu/shadPS4/pull/3261. Reverting that PR and this one line fixes the issue in Sakura Wars.

PR #3271, opened 2025-07-19
Add default trophy sound

rainmakerv3:
Adds a default sound upon unlocking a trophy if no custom sound file exists.

Got the sound from here: https://pixabay.com/sound-effects/search/achievement/ so should be ok to use.

At first I thought this would need some kind of volume normalization because some games are much louder than others, but it seems to play at reasonable levels for both the loud game and soft game I tried so normalization seems to be unneeded. 

The wav file is embedded in the binary similar to the trophy icons. For its source, i just put it in src/images with the other trophy files which may be kinda weird but I can't see a better location for it and probably doesn't justify a new folder.

EDIT: instead using a sound that was made by @tlarok . Thanks very much!


Silver-Skyy:
it might just be my system but i feel its a little to quiet ? i feel with the music and background sounds from the game it will not be heard at all

Edit, typo

RyuSeisuke1992:
Are the Pop up Sound trophies from Sony legally protected so that they can be implemented directly?

rainmakerv3:
@Silver-Skyy , I'm open to increasing the volume if other people feel it's too soft, maybe I'll wait for additional feedback first.

@RyuSeisuke1992 . I think it's better to be safe than sorry. Anyhow, there's already a custom sound feature that people can also use.

georgemoralis:
is the pr ready for merging? from reviewing side it seems ok

rainmakerv3:
I've played on the branch for a few hours and popped a few trophies with it, seems ok on my end

georgemoralis:
Mirror action until fixed : 
https://github.com/georgemoralis/shadPS4/actions/runs/16428994742

PR #3279, opened 2025-07-20
Implement sceAudioOutGetLastOutputTime

kalaposfos13:
Implements the titular function seen in https://github.com/shadps4-compatibility/shadps4-game-compatibility/issues/1322 (CUSA04408 - Middle-earth™: Shadow of War™). This is not the source of the hang in the game, and the spam is only a side effect of the fact that the game hangs, while the audio thread doesn't, but since it does look to be working fine I thought I'd PR it anyway.


squidbus (approved):

georgemoralis:
https://github.com/georgemoralis/shadPS4/actions/runs/16402823137

PR #3284, opened 2025-07-20
Using custom usb lib

georgemoralis:
This is a replacement of #3235 , Instead of hacking usblib calls we use custom libusb lib with fixes for sysv_abi callbacks

Use my fork actions in 
https://github.com/georgemoralis/shadPS4/actions/runs/16404431633

For testing lego dimensions that used to work on #3235 
Big thanks to @connorh315 for his work on this


valjbp:
LEGO Dimensions still works with these changes, still detects the reader.

image

squidbus (approved):
I think it would've been simpler in the fork to just re-type the callback instead of casting when called, but I suppose it's fine.

georgemoralis:
mirror actions until fixed: 
https://github.com/georgemoralis/shadPS4/actions/runs/16413220089

Missake212:
This PR regressed Tekken 7, went from being able to get to the menus/going in game to crashing on boot.

image

[CUSA06014.log](https://github.com/user-attachments/files/22198420/CUSA06014.log)

PR #3288, opened 2025-07-21
Ime fixes

w1naenator:


w1naenator:
@georgemoralis pease check if it is ok now.

src/core/libraries/ime/ime.cpp
```diff
@@ -226,71 +238,81 @@ Error PS4_SYSV_ABI sceImeGetPanelSize(const OrbisImeParam* param, u32* width, u3
...
 
 Error PS4_SYSV_ABI sceImeKeyboardClose(Libraries::UserService::OrbisUserServiceUserId userId) {
     LOG_INFO(Lib_Ime, "called");
 
     if (!g_keyboard_handler) {
+        LOG_ERROR(Lib_Ime, "No keyboard handler is open");
         return Error::NOT_OPENED;
     }
 
+    if (userId < 0 || userId > 4) { // Todo: check valid user IDs
```
georgemoralis: userId can be any 32bit unsigned number , so until we have proper user management i suggest comment this check

src/core/libraries/ime/ime.cpp
```diff
@@ -307,24 +329,57 @@ int PS4_SYSV_ABI sceImeKeyboardGetResourceId() {
...
+    }
+    if (static_cast<u32>(param->option) & ~kValidOrbisImeKeyboardOptionMask) {
+        LOG_ERROR(Lib_Ime,
+                  "Invalid param->option\n"
+                  "option:    {:032b}\n"
+                  "validMask: {:032b}",
+                  static_cast<u32>(param->option), kValidOrbisImeKeyboardOptionMask);
+        return Error::INVALID_OPTION;
+    }
+    if (userId < 0 || userId > 4) { // Todo: check valid user IDs
```
georgemoralis: same

src/core/libraries/ime/ime.cpp
w1naenator: In my understanding, on a real console, up to 16 users can be registered, and up to 4 users can be logged in at the same time.
Does OrbisUserServiceUserId refer to a registered "console" user ID (1–16), or to a currently logged-in user (IDs 1–4)?
Or am I mistaken?

georgemoralis: you can register 16 users and up to 4 users logged but their ids are not sequence. It can be any u32 number.

w1naenator: Thank you, this is clear now!

georgemoralis:
Mirror actions until fixed : 
https://github.com/georgemoralis/shadPS4/actions/runs/16438924933

kalaposfos13:
This PR makes SotC spam the following line:
`[Lib.Ime] <Error> ime.cpp:374 sceImeKeyboardOpen: Keyboard handler is already open`
[CUSA08809.log](https://github.com/user-attachments/files/21372950/CUSA08809.log)

w1naenator:
Ok,will check.

w1naenator:
@kalaposfos13 is there some kind of trick to launch SotC? 

[Debug] <Critical> address_space.cpp:330 Protect: Unreachable code!
Failed to change virtual memory protection for address 0x29b000000, size 16777216

w1naenator:
The game doesn't fully start unless the startup logo is skipped via a patch. In some cases, partial activity appears, but it never progresses past a certain point. I also confirmed that this issue exists in the release version, where IME-related fixes were not merged. The game repeatedly calls `sceImeKeyboardOpen`, even when it returns 0. The only difference is that I added more detailed logging, which results in additional spam. Further research is needed—possibly the game expects specific data or a response from the IME handler.

kalaposfos13:
For the record, you can get further in SotC with some log filters to prevent logging based race conditions, and on Linux, I've never seen that specific unreachable, so it is most likely a Windows specific thing. Also, with nullGPU, or a single assert skip that only works on Linux and Mac, you can get ingame as well if you try long enough until the game doesn't hang on boot.

PR #3293, opened 2025-07-22
Wiring misc functions

mikusp:
Extracted from Network WIP

### DIFF
--- a/src/core/libraries/kernel/memory.cpp
+++ b/src/core/libraries/kernel/memory.cpp
@@ -693,7 +693,8 @@ void RegisterMemory(Core::Loader::SymbolsResolver* sym) {
     LIB_FUNCTION("BC+OG5m9+bw", "libkernel", 1, "libkernel", 1, 1, sceKernelGetDirectMemoryType);
     LIB_FUNCTION("pO96TwzOm5E", "libkernel", 1, "libkernel", 1, 1, sceKernelGetDirectMemorySize);
     LIB_FUNCTION("yDBwVAolDgg", "libkernel", 1, "libkernel", 1, 1, sceKernelIsStack);
-    LIB_FUNCTION("jh+8XiK4LeE", "libkernel", 1, "libkernel", 1, 1, sceKernelIsAddressSanitizerEnabled);
+    LIB_FUNCTION("jh+8XiK4LeE", "libkernel", 1, "libkernel", 1, 1,
+                 sceKernelIsAddressSanitizerEnabled);
     LIB_FUNCTION("NcaWUxfMNIQ", "libkernel", 1, "libkernel", 1, 1, sceKernelMapNamedDirectMemory);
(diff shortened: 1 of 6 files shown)


src/core/libraries/network/net.cpp
```diff
@@ -2097,12 +2097,14 @@ void RegisterLib(Core::Loader::SymbolsResolver* sym) {
...
     LIB_FUNCTION("+S-2-jlpaBo", "libSceNet", 1, "libSceNet", 1, 1, sceNetGetNameToIndex);
     LIB_FUNCTION("TCkRD0DWNLg", "libSceNet", 1, "libSceNet", 1, 1, sceNetGetpeername);
+    LIB_FUNCTION("TXFFFiNldU8", "libScePosix", 1, "libkernel", 1, 1, sys_getpeername);
     LIB_FUNCTION("G3O2j9f5z00", "libSceNet", 1, "libSceNet", 1, 1, sceNetGetRandom);
     LIB_FUNCTION("6Nx1hIQL9h8", "libSceNet", 1, "libSceNet", 1, 1, sceNetGetRouteInfo);
     LIB_FUNCTION("hLuXdjHnhiI", "libSceNet", 1, "libSceNet", 1, 1, sceNetGetSockInfo);
     LIB_FUNCTION("Cidi9Y65mP8", "libSceNet", 1, "libSceNet", 1, 1, sceNetGetSockInfo6);
     LIB_FUNCTION("hoOAofhhRvE", "libSceNet", 1, "libSceNet", 1, 1, sceNetGetsockname);
     LIB_FUNCTION("xphrZusl78E", "libSceNet", 1, "libSceNet", 1, 1, sceNetGetsockopt);
+    LIB_FUNCTION("6O8EwYOgH9Y", "libScePosix", 1, "libkernel", 1, 1, sys_getsockopt);
```
StevenMiller123: Following the convention set by shadow's previous PRs, these libScePosix exports should probably be placed in kernel.cpp instead.

StevenMiller123 (changes requested):

src/core/libraries/kernel/memory.cpp
```diff
@@ -318,6 +318,10 @@ s32 PS4_SYSV_ABI sceKernelIsStack(void* addr, void** start, void** end) {
     return memory->IsStack(std::bit_cast<VAddr>(addr), start, end);
 }
 
+u32 PS4_SYSV_ABI sceKernelIsAddressSanitizerEnabled() {
+    return ORBIS_OK;
```
georgemoralis: why that should be always be true? i know it was a default behavior but still doesn't feels right

mikusp: It's 0 so false

georgemoralis: either way i think at least we need to add a log here , to at least know that it gets called

mikusp: added 👍

PR #3294, opened 2025-07-22
Extract netmask and default gateway from host system

mikusp:
I'd be grateful for pointers on how to retrieve default gateways on Windows and Mac

### DIFF
--- a/src/core/libraries/network/net_util.cpp
+++ b/src/core/libraries/network/net_util.cpp
@@ -11,6 +11,7 @@ typedef int socklen_t;
 #else
 #include <cerrno>
 #include <arpa/inet.h>
+#include <ifaddrs.h>
 #include <net/if.h>
 #include <netdb.h>
 #include <netinet/in.h>
@@ -21,7 +22,6 @@ typedef int socklen_t;
 typedef int net_socket;
 #endif
 #if defined(__APPLE__)
-#include <ifaddrs.h>
 #include <net/if_dl.h>
 #endif
 
@@ -100,6 +100,8 @@ bool NetUtilInternal::RetrieveEthernetAddr() {
         }
     }
 
+    close(sock);
+
     if (success) {
         memcpy(ether_address.data(), ifr.ifr_hwaddr.sa_data, 6);
         return true;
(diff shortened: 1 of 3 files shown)


src/core/libraries/network/net_util.cpp
```diff
@@ -100,11 +105,113 @@ bool NetUtilInternal::RetrieveEthernetAddr() {
...
 #endif
     return false;
 }
+
+const std::string& NetUtilInternal::GetDefaultGateway() const {
+    return default_gateway;
+}
+
+bool NetUtilInternal::RetrieveDefaultGateway() {
+    std::scoped_lock lock{m_mutex};
```
georgemoralis, line 124: First of all i think that the implementation should go to 
https://github.com/shadps4-emu/shadPS4/blob/main/src/core/libraries/network/net_util.cpp

src/core/libraries/network/net_util.cpp
```diff
@@ -100,11 +105,113 @@ bool NetUtilInternal::RetrieveEthernetAddr() {
...
+
+        if (destination == "00000000") {
+            u64 default_gateway{};
+            std::stringstream ss;
+            ss << std::hex << gateway;
+            ss >> default_gateway;
+
+            in_addr addr;
+            addr.s_addr = default_gateway;
+            this->default_gateway = inet_ntoa(addr);
```
georgemoralis: inet_ntoa is not thread-safe. Consider using inet_ntop

georgemoralis: i haven't tested it but something like that should work
char str[INET_ADDRSTRLEN];
in_addr addr;
addr.s_addr = default_gateway;
inet_ntop(AF_INET, &addr, str, sizeof(str));
this->default_gateway = str;

georgemoralis (approved):

src/core/libraries/network/net_util.cpp
mikusp: It is? 😄

georgemoralis:
gateway code for windows added

PR #3303, opened 2025-07-23
shader_recompiler: Fix incorrect bary coord loads when unsupported

raphaelthegreat:
Should fix https://github.com/shadps4-emu/shadPS4/pull/3245#issuecomment-3086406855 and https://github.com/shadps4-emu/shadPS4/pull/3245#issuecomment-3105738985

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv.cpp
@@ -241,7 +241,8 @@ spv::ExecutionMode ExecutionMode(AmdGpu::TessellationPartitioning spacing) {
     UNREACHABLE_MSG("Tessellation spacing {}", spacing);
 }
 
-void SetupCapabilities(const Info& info, const Profile& profile, EmitContext& ctx) {
+void SetupCapabilities(const Info& info, const Profile& profile, const RuntimeInfo& runtime_info,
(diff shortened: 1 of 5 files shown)


Missake212:
Can confirm that all three of my games that were previously regressed on Intel now are back to going in game.

StevenMiller123:
Can confirm that the game I reported with is fixed as well.
[CUSA00803.log](https://github.com/user-attachments/files/21388728/CUSA00803.log)

GHU7924:
Sorry for writing here, but it looks like the error is related to this PR. I updated to the latest Main and in game `[CUSA00562] Tearaway™ Unfolded` and I got this error (as far as I remember, this did not happen before the broken actions):

`[Debug] <Critical> emit_spirv_context_get_set.cpp:218 EmitGetAttribute: Unreachable code!`
`Read attribute BaryCoordSmoothCentroid`

The commit from this PR has the same error.

[CUSA00562.log](https://github.com/user-attachments/files/21388770/CUSA00562.log)

GHU7924:
With the new commit the critical error has been fixed.

bigol83:
According to my test the graphics regression i have with Yakuza 0 started with this PR

You can see the smoke and all the blur effect at the right side is blocky and there are some black squares 

immagine

PR #3304, opened 2025-07-23
ime-fixes-hotfix

w1naenator:
- Fixed incorrect validation of internal errors causing the keyboard GUI to not close on call to sceImeClose.
- Removed nonstop spam when ImeDialog is opened.
- Fixed text and data decoding for logs.
- Fixed incorrect validation of internal errors in sceImeKeyboardClose.
- Added partial implementation of sceImeKeyboardGetResourceId (always returns that the USB keyboard is disconnected); used in CUSA33782.


src/core/libraries/ime/ime_dialog.cpp
```diff
@@ -266,6 +262,7 @@ Error PS4_SYSV_ABI sceImeDialogInit(OrbisImeDialogParam* param, OrbisImeParamExt
 
     g_ime_dlg_result = {};
     g_ime_dlg_state = ImeDialogState(param, extended);
+    LOG_ERROR(Lib_ImeDialog, "test");
```
georgemoralis: is that needed?

w1naenator: Sorry, forgot to remove.

src/core/libraries/ime/ime.cpp
```diff
@@ -22,6 +22,7 @@ class ImeHandler {
         Init(param, false);
     }
     ImeHandler(const OrbisImeParam* param) {
+        LOG_INFO(Lib_Ime, "Creating ImeHandler for IME");
```
georgemoralis: repeat line

src/core/libraries/ime/ime.cpp
```diff
@@ -321,10 +323,42 @@ int PS4_SYSV_ABI sceImeKeyboardGetInfo() {
...
+        LOG_ERROR(Lib_Ime, "Keyboard handler not opened");
+        resourceIdArray->user_id = userId;
+        for (u32& id : resourceIdArray->resource_id)
+            id = 0;
+        return Error::NOT_OPENED;
+    }
+
+    // Simulate "no USB keyboard connected", needed for some Unity engine games
+    resourceIdArray->user_id = userId;
+    for (u32& id : resourceIdArray->resource_id)
```
georgemoralis: you should use brackets for better coding style

src/core/libraries/ime/ime_dialog_ui.cpp
```diff
@@ -264,13 +267,13 @@ void ImeDialogUi::Draw() {
         }
 
         if (state->is_multi_line) {
-            LOG_INFO(Lib_ImeDialog, "   Drawing multi-line widget…");
+            // LOG_INFO(Lib_ImeDialog, "   Drawing multi-line widget…");
```
georgemoralis: if not needed comment lines should be removed

src/core/libraries/ime/ime.cpp
w1naenator: Ok, I will consider!

src/core/libraries/ime/ime_dialog_ui.cpp
w1naenator: Got it!

PR #3305, opened 2025-07-23
qt: Add toggle in the settings window for readbacks

tomboylover93:
See PR title.
Also adds a toggle for "Readback Linear Images".
Since the current readbacks implementation is slow and will most likely be rewritten in the future and adding a toggle in the Settings window for it has been a point of discussion, I thought to lock it behind another "Advanced" section to avoid people enabling it and coming across issues in games that don't need it/don't work well with it as a result, but I felt like it was beyond the scope of this PR (perhaps a warning about not enabling it unless being told to would suffice?). Let me know if you have any ideas.

image


Missake212:
Looks good to me and works as expected, we've discussed it on the Discord already and I think the only thing that could be better is to lock it behind another tab that you'd have to enable somewhere else on the emu, and also have a pop up when turning on the setting with a warning or something (but I believe in the case of Readbacks it's not that important since it won't need a toggle in the future, just repeating what we said for future toggle able features).

GHU7924:
There was already some PR that was cancelled. I suggested doing it this way:

EF

Or

FD

BUT I won't interfere, I'll just mention it, since this function is more relevant to developers, not regular users.

P.S. I don't know what the problem was, the idea itself or the implementation, so I'll just leave it here.

rainmakerv3:
I think it would be good to implement what missake/GHU are suggesting (pop a qmessage after the checkbox as shown by GHU is clicked with a warning, then setvisible a certain tab).

Alternatively, maybe we can put at least put the readbacks stuff in a different groupbox and have a qlabel on it with a very visible warning below the settings themselves. I think the description text in the bottom may not be enough IMO

tomboylover93:
I thought about doing the latter, like in the second image GHU posted, but I can't figure out how to put another groupbox to the right of the "Advanced" groupbox.

rainmakerv3:
I can maybe just PR to your PR maybe later or tomorrow to add a new groupbox

rainmakerv3:
PR submitted to your branch

I didn't finish typing out the warning in the groupbox btw, kinda had to do this quickly

tomboylover93:
Revised the warning, though I'd like some input on if this is ideal (wording, additional details etc).

image

Missake212:
Looks perfect to me, pinging @StevenMiller123 for a proper review (hoping that speeds up the process 😁).

GHU7924:
From an interface point of view, I agree, the current version looks perfect. Great job 👍 

P.S. Actually, I originally voted for the second option because the first option was a bit redundant, due to the fact that we don't have enough features to make a whole tab for it.

GHU7924:
I have a question - Are texts `Experimental Features` and `WARNING: ...` available for translation? Or have they been forgotten?

tomboylover93:
I don't know how to set up QLabel elements (WARNING: ...) for translation so if someone could do it I'd appreciate it. I'll do the "Experimental Features" text.

PR #3306, opened 2025-07-23
shader_recompiler: Use VM bit for conditional discard

raphaelthegreat:
Shader export instructions have 2 special bits, the DONE bit is set for the last export instruction in the shader and can be safely ignored. The VM bit is also set on a single export and makes that export responsible for informing the color buffer which pixels are valid and which have been discarded. So use that to insert the conditional discard instead of guessing from the number of predecessors.

This fixes the blocky shadows in Driveclub, the fragment shader used for rendering tree shadows read the alpha value from the texture and did a conditional discard based on it. But because there was no control flow, the predecessor check failed, resulting in the whole shader being eliminated as dead code.

### DIFF
--- a/src/shader_recompiler/frontend/control_flow_graph.cpp
+++ b/src/shader_recompiler/frontend/control_flow_graph.cpp
@@ -319,8 +319,6 @@ void CFG::LinkBlocks() {
         // need to link with the next block.
         if (!end_inst.IsTerminateInstruction()) {
             auto* next_block = get_block(block.end);
-            ++next_block->num_predecessors;
-
             block.branch_true = next_block;
             block.end_class = EndClass::Branch;
             continue;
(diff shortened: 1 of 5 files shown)


diogoaduartereis:
image
image
It looks much better!

diogoaduartereis:
image
There seems to have been a regression here though

raphaelthegreat:
Which track, what settings, and where is this?

raphaelthegreat:
Also a comparison with real hardware might be needed, maybe that also happens there

Randomuser8219:
image
It looks like this also fixed the blocky shadows in Tour De France 2016. There used to be blocky shadows on the bike wheels and on trees.
image

Randomuser8219:
This also fixes the blocky shadows in Knack. Especially chapter 2-1.
Before this PR:
Screenshot 2025-07-27 123810
After this PR:
Screenshot 2025-07-27 123932

PR #3310, opened 2025-07-23
Net callbacks

mikusp:
Run the callbacks set by the game when \*check\* functions are called


StevenMiller123:
https://github.com/shadps4-emu/shadPS4/pull/2161
These callbacks were disabled intentionally a while back, due to instability in games using libSceNpToolkit2. I suggest reviewing the comments of that PR.

PR #3316, opened 2025-07-24
V_ALIGNBYTE_B32 and V_ALIGNBIT_B32

baggins183:


baggins183:
@auser1337 I remembered this one from months ago

src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -390,6 +390,10 @@ void Translator::EmitVectorAlu(const GcnInst& inst) {
         return V_CVT_PK_U8_F32(inst);
     case Opcode::V_LSHL_B64:
         return V_LSHL_B64(inst);
+    // case Opcode::V_ALIGNBIT_B32: // Not tested in games yet
+    // return V_ALIGNBIT_B32(inst);
```
squidbus: Feel like we should just uncomment this otherwise it's not going to get tested.

baggins183: I dont know a game that actually uses ALIGNBIT so I thought it was better to assert for now. Your call

baggins183: Dreams uses ALIGNBYTE but I dont have it so someone will need to test

squidbus: > I dont know a game that actually uses ALIGNBIT so I thought it was better to assert for now. Your call

Think we should just uncomment it, it looks fine to me anyway.

squidbus (approved):

PR #3317, opened 2025-07-24
inet_ntop and inet_pton

mikusp:
Extracted from Network WIP


georgemoralis:
it might be wise to use freebsd implementation for sceNetInetPton but ok it can be done later

PR #3318, opened 2025-07-24
Implement simple DNS name resolution

mikusp:
Extracted from Network WIP


georgemoralis:
i don't understand why windows qt fails ;/

PR #3319, opened 2025-07-24
Sockets are now files

mikusp:
Extracted from Network WIP

### DIFF
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -219,6 +219,18 @@ File* HandleTable::GetFile(int d) {
     return m_files.at(d);
 }
 
+File* HandleTable::GetSocket(int d) {
+    std::scoped_lock lock{m_mutex};
+    if (d < 0 || d >= m_files.size()) {
+        return nullptr;
+    }
+    auto file = m_files.at(d);
+    if (file->type != Core::FileSys::FileType::Socket) {
+        return nullptr;
+    }
+    return file;
+}
+
 File* HandleTable::GetFile(const std::filesystem::path& host_name) {
     for (auto* file : m_files) {
         if (file != nullptr && file->m_host_name == host_name) {
--- a/src/core/libraries/network/sys_net.cpp
+++ b/src/core/libraries/network/sys_net.cpp
@@ -4,22 +4,25 @@
 #include <common/assert.h>
 #include <common/logging/log.h>
 #include <core/libraries/kernel/kernel.h>
+#include "common/error.h"
 #include "common/singleton.h"
+#include "core/file_sys/fs.h"
 #include "net_error.h"
 #include "sockets.h"
 #include "sys_net.h"
 
 namespace Libraries::Net {
 
+using FDTable = Common::Singleton<Core::FileSys::HandleTable>;
+
 int PS4_SYSV_ABI sys_connect(OrbisNetId s, const OrbisNetSockaddr* addr, u32 addrlen) {
-    auto* netcall = Common::Singleton<NetInternal>::Instance();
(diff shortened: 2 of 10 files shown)


StevenMiller123 (changes requested):
As it stands, this is all unused code. I think it would be better to bring over the required changes to connect these functions to the filesystem fstat, read, and write functions too.

mikusp:
I can include the changes but it would require registering sockets in the file descriptor table, bringing over half of the remaining code in the previous PR

src/core/libraries/network/sys_net.h
```diff
@@ -5,6 +5,10 @@
 
 #include "common/types.h"
 #include "net.h"
+#ifdef _WIN32
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
+#endif
```
StevenMiller123: Rather than working around the ssize_t type, you could just use our s64 type instead.

src/core/file_sys/fs.cpp
```diff
@@ -219,6 +219,18 @@ File* HandleTable::GetFile(int d) {
...
+    if (d < 0 || d >= m_files.size()) {
+        return nullptr;
+    }
+    auto file = m_files.at(d);
+    if (file->type != Core::FileSys::FileType::Socket) {
+        return nullptr;
+    }
+    return file;
+}
+
```
StevenMiller123, line 233: Rather than using this GetSocket function and checking if the return value is null to trigger the bad file returns, I think it would be simpler to just use the above GetFile function and check if the returned file is null or not a socket.

src/core/libraries/network/sys_net.cpp
```diff
@@ -141,24 +142,32 @@ int PS4_SYSV_ABI sys_socketex(const char* name, int family, int type, int protoc
...
         UNREACHABLE_MSG("Unknown type {}", type);
     }
-    auto* netcall = Common::Singleton<NetInternal>::Instance();
-    auto id = ++netcall->next_sock_id;
-    netcall->socks.emplace(id, sock);
-    return id;
+    if (!socket->IsValid()) {
+        *Libraries::Kernel::__Error() = ORBIS_NET_EPROTONOSUPPORT;
+        return -1;
+    }
```
StevenMiller123, line 162: I'm confused by which type of error should be used here. I see most functions in this file use the `ORBIS_NET_ERROR_*` error codes, where you're using `ORBIS_NET_*` here.
Not sure which one is correct, but it might be good to properly unify these to follow whatever is correct.

src/core/file_sys/fs.cpp
mikusp: Net library does not seem to differentiate between socket fd does not exist and socket fd is not a socket cases, and GetSocket places less burden on the calling site to check whether we really get a socket. In cases where any file can be used, like close() GetFile is still used

src/core/libraries/network/sys_net.cpp
mikusp: Errno is set to a positive value describing the error, which ORBIS_NET_E* values are, so I believe that is correct. sceNet* functions get that value and convert to ORBIS_NET_ERROR_* return values

StevenMiller123: I suspected as much, and that does make more sense when considering other places where errno is used.
Not exactly a requirement, but it would be nice if you could do a quick find-and-replace to address the use of `ORBIS_NET_ERROR_*` values as errnos.

StevenMiller123 (approved):
Changes here look good to me, though I haven't had a chance to regression test it yet.

StevenMiller123:
Did some quick tests, I'm not seeing any regressions in my titles.

PR #3324, opened 2025-07-25
Return the number of samples enqueued in AudioOut

mikusp:
Another one for Agents of Mayhem CUSA03067. The game calls sceAudioOutOutput and checks if the result is larger than 0, meaning that the audio data has been properly enqueued. If it is zero (as it was before), the game deliberately crashes 🐸  RE suggests that it returns the number of samples provided.


squidbus (approved):
Thanks, I remember noticing this back when I worked on AudioOut but I think I forgot to add it in because it didn't solve whatever issue I was having.

PR #3325, opened 2025-07-25
Core: Return error if DirectMemoryQuery address is too high

StevenMiller123:
This should fix an error in The Sims™ 4 noticed by @Missake212, where the game infinitely looped sceKernelDirectMemoryQuery calls.


Missake212:
Back to spamming epolls.

[CUSA09216.log](https://github.com/user-attachments/files/21435534/CUSA09216.log)

PR #3326, opened 2025-07-25
texture_cache: Do not modify mip height for copy in volume texture.

squidbus:
As noted with TLG in https://github.com/shadps4-emu/shadPS4/pull/3261, the change to `mip_height` when refreshing textures may not be aligned correctly for volume textures. It seems this was fixed for TLG by using `max(mip_height, 8)` instead of aligning, but this is still an issue for at least Sakura Wars.

For now at least, excluding volume textures from this alignment.


squidbus:
@StevenMiller123 Can you check if Metaphor is still fine with this?

StevenMiller123:
Seems like Metaphor: ReFantazio still works fine with these changes.
Main:
image
PR:
image

Issue #2192, opened 2025-01-19
[GAME BUG]: Knack crashes on killing an enemy (AMD)


Randomuser8219:
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.
- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [x] I have disabled all patches and cheats and the issue is still present.
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadps4-game-compatibility?tab=readme-ov-file#informations) installed.

### Describe the Bug

Knack crashes on an value.cpp unreachable code error when killing an enemy.

### Reproduction Steps

1. Open Knack on shadPS4
2. Go ingame

### Specify OS Version

Windows 11 24H2

### CPU

AMD Ryzen Z1 Extreme

### GPU

AMD Radeon Graphics

### Amount of RAM in GB

16GB

### Amount of VRAM in GB

8GB

### Log File

[shad_log.txt](https://github.com/user-attachments/files/18470554/shad_log.txt)

Randomuser8219:
The crash when killing an enemy in 1-1 is gone, now when killing the first robot in 1-2, the game crashes with a device lost error instead of a value.cpp error. Log will be sent soon.

Randomuser8219:
https://files.catbox.moe/rwu9ib.txt
Here's my log. Uploaded to somewhere else because of size.

Randomuser8219:
[shad_log.txt](https://github.com/user-attachments/files/19861445/shad_log.txt)
Log with Vulkan validation.

Randomuser8219:
The new clue for the crashing on AMD is some missing formats that are otherwise present on NVIDIA.
New validation log:
[shad_log (4).txt](https://github.com/user-attachments/files/21187404/shad_log.4.txt)

Randomuser8219:
#3327 fixed the AMD enemy crash. When it gets pushed to main I’ll close this.

PR #3327, opened 2025-07-26
shader_recompiler: Rework sharp tracking for robustness

raphaelthegreat:
On main image+sampler sharp tracking was kinda broken in a silly way. PatchImageSharp would call BreadthFirstSearch on the image instruction itself. Before https://github.com/shadps4-emu/shadPS4/pull/3015 that was fine because there was a single handle argument and image address components are very unlikely to come from user data or read const instruction.

However that PR added a sampler handle argument at the end of the instruction, which caused the image tracking to search in the sampler instruction tree first (because BreadthFirstSearch iterates instruction arguments in reverse order). Thankfully the sampler argument is a composite so it wouldn't get expanded at the first level of the tree. If the image handle was a read const/user data instruction then it would get picked. But if it was a phi, then next iteration the sampler tree would be searched next, leading to the tracking sometimes picking sampler sharps instead of image sharps.

Fix this by reworking sharp tracking for robustness. Instead of searching a single source, traverse the phi tree (only the image handle this time) and find all possible sources. Then attempt to prune these sources, based on dominance analysis. If a source is dominated by another one then the former can be removed as control flow is guaranteed to pass the dominator. This also implicitly performs reachability analysis as the function will also return true if no path from source to resource access block exists in general.

This fixes the rear mirror in Driveclub without reverting the PR mentioned above.

| main  | PR |
| ------------- | ------------- |
| before | after |

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -893,6 +893,7 @@ set(SHADER_RECOMPILER src/shader_recompiler/exception.h
                       src/shader_recompiler/ir/attribute.h
                       src/shader_recompiler/ir/basic_block.cpp
                       src/shader_recompiler/ir/basic_block.h
+                      src/shader_recompiler/ir/breadth_first_search.h
                       src/shader_recompiler/ir/condition.h
                       src/shader_recompiler/ir/ir_emitter.cpp
                       src/shader_recompiler/ir/ir_emitter.h
--- a/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
+++ b/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
@@ -945,11 +945,11 @@ void EmitContext::DefineImagesAndSamplers() {
         const Id id{AddGlobalVariable(sampler_pointer_type, spv::StorageClass::UniformConstant)};
         Decorate(id, spv::Decoration::Binding, binding.unified++);
         Decorate(id, spv::Decoration::DescriptorSet, 0U);
-        auto sharp_desc = std::holds_alternative<u32>(samp_desc.sampler)
-                              ? fmt::format("sgpr:{}", std::get<u32>(samp_desc.sampler))
-                              : fmt::format("inline:{:#x}:{:#x}",
-                                            std::get<AmdGpu::Sampler>(samp_desc.sampler).raw0,
-                                            std::get<AmdGpu::Sampler>(samp_desc.sampler).raw1);
+        const auto sharp_desc =
+            samp_desc.is_inline_sampler
+                ? fmt::format("inline:{:#x}:{:#x}", samp_desc.inline_sampler.raw0,
+                              samp_desc.inline_sampler.raw1)
+                : fmt::format("sgpr:{}", samp_desc.sharp_idx);
         Name(id, fmt::format("{}_{}{}", stage, "samp", sharp_desc));
         samplers.push_back(id);
         interfaces.push_back(id);
--- a/src/shader_recompiler/frontend/control_flow_graph.cpp
+++ b/src/shader_recompiler/frontend/control_flow_graph.cpp
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include <algorithm>
-#include <unordered_map>
 #include "common/assert.h"
 #include "common/logging/log.h"
 #include "shader_recompiler/frontend/control_flow_graph.h"
@@ -350,19 +349,7 @@ void CFG::LinkBlocks() {
             block.branch_false = end_block;
             block.end_class = EndClass::Branch;
         } else if (end_inst.opcode == Opcode::S_ENDPGM) {
-            const auto& prev_inst = inst_list[block.end_index - 1];
-            if (prev_inst.opcode == Opcode::EXP && prev_inst.control.exp.en == 0) {
-                if (prev_inst.control.exp.target != 9) {
-                    block.end_class = EndClass::Kill;
-                } else if (const auto& exec_mask = inst_list[block.end_index - 2];
-                           exec_mask.src[0].field == OperandField::ConstZero) {
-                    block.end_class = EndClass::Kill;
-                } else {
-                    block.end_class = EndClass::Exit;
-                }
-            } else {
-                block.end_class = EndClass::Exit;
-            }
+            block.end_class = EndClass::Exit;
         } else {
             UNREACHABLE();
         }
@@ -401,12 +388,6 @@ std::string CFG::Dot() const {
             dot += fmt::format("\t\t{}->N{};\n", name, node_uid);
             dot +=
                 fmt::format("\t\tN{} [label=\"Exit\"][shape=square][style=stripped];\n", node_uid);
-            ++node_uid;
-            break;
-        case EndClass::Kill:
-            dot += fmt::format("\t\t{}->N{};\n", name, node_uid);
-            dot +=
-                fmt::format("\t\tN{} [label=\"Kill\"][shape=square][style=stripped];\n", node_uid);
             ++node_uid;
             break;
         }
--- a/src/shader_recompiler/frontend/control_flow_graph.h
+++ b/src/shader_recompiler/frontend/control_flow_graph.h
@@ -23,7 +23,6 @@ using Hook =
 enum class EndClass {
     Branch, ///< Block ends with a (un)conditional branch.
     Exit,   ///< Block ends with an exit instruction.
-    Kill,   ///< Block ends with a discard instruction.
 };
 
 /// A block represents a linear range of instructions.
--- a/src/shader_recompiler/frontend/structured_control_flow.cpp
+++ b/src/shader_recompiler/frontend/structured_control_flow.cpp
@@ -330,9 +330,9 @@ private:
             }
         }
         // Expensive operation:
-        if (!AreSiblings(goto_stmt, label_stmt)) {
-            UNREACHABLE_MSG("Goto is not a sibling with the label");
-        }
+        // if (!AreSiblings(goto_stmt, label_stmt)) {
+        //    UNREACHABLE_MSG("Goto is not a sibling with the label");
+        //}
         // goto_stmt and label_stmt are guaranteed to be siblings, eliminate
         if (std::next(goto_stmt) == label_stmt) {
             // Simply eliminate the goto if the label is next to it
@@ -629,6 +629,7 @@ private:
                 if (!stmt.block->is_dummy) {
                     const u32 start = stmt.block->begin_index;
                     const u32 size = stmt.block->end_index - start + 1;
+                    current_block->cfg_block = stmt.block;
                     translator.Translate(current_block, stmt.block->begin,
                                          inst_list.subspan(start, size));
                 }
@@ -769,7 +770,7 @@ private:
                 break;
             }
             default:
-                throw NotImplementedException("Statement type {}", u32(stmt.type));
+                UNREACHABLE_MSG("Statement type {}", u32(stmt.type));
             }
         }
         if (current_block) {
--- a/src/shader_recompiler/frontend/structured_control_flow.cpp
+++ b/src/shader_recompiler/frontend/structured_control_flow.cpp
@@ -39,7 +39,6 @@ enum class StatementType {
     Loop,
     Break,
     Return,
-    Kill,
     Unreachable,
     Function,
     Identity,
@@ -88,7 +87,6 @@ struct Statement : ListBaseHook {
     Statement(Break, Statement* cond_, Statement* up_)
         : cond{cond_}, up{up_}, type{StatementType::Break} {}
     Statement(Return, Statement* up_) : up{up_}, type{StatementType::Return} {}
-    Statement(Kill, Statement* up_) : up{up_}, type{StatementType::Kill} {}
     Statement(Unreachable, Statement* up_) : up{up_}, type{StatementType::Unreachable} {}
     Statement(FunctionTag) : children{}, type{StatementType::Function} {}
     Statement(Identity, IR::Condition cond_, Statement* up_)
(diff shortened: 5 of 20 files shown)


squidbus:
Sakura Wars (CUSA16404) hits:
```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:501 CompileModule: Compiling fs shader 0xc70eb6c0 
[Debug] <Critical> resource_tracking_pass.cpp:460 operator(): Assertion Failed!
Unable to deduce sharp source
```

raphaelthegreat:
Can you check which part of the sharp tracking fails (does it find no sources or prune all of them?)

squidbus:
I added the type and source count to the assert and got this:
```
[Debug] <Critical> resource_tracking_pass.cpp:465 operator(): Assertion Failed!
Unable to deduce sharp source for AmdGpu::Image: Found 2 sources
```

raphaelthegreat:
Hm could this actually be a case of 2 possible sources (could be with some scalar condition) like setting a descriptor image array in vulkan and indexing it with a uniform variable

StevenMiller123:
This PR brings Hatsune Miku: Project DIVA Future Tone back ingame.
image

On main, it crashes with
```
Unknown data_format=10 and num_format=8
```

squidbus:
> Hm could this actually be a case of 2 possible sources (could be with some scalar condition)

I believe its the same shader as some time ago where a series of branches can lead to an unconditional branch to the end.

raphaelthegreat:
Then not all should be reachable, should they 🤔

squidbus:
I need to double check the IR for this instance of the failure but I believe this is the relevant segment of assembly with annotations:

[excerpt.txt](https://github.com/user-attachments/files/21440657/excerpt.txt)

raphaelthegreat:
In both of these samples it appears the s_load_dword loading the image sharp directly precedes the image instruction without any control flow, so there shouldn't be any problems

squidbus:
Hmm yes you're right, let me figure out exactly where the failure is here

squidbus:
Actually you can see this is how the sharp is loaded before it is sampled:
```
s_load_dwordx8  s[12:19], s[8:9], 0x0
```

...and `s[8:9]` is affected by this in the branches:
```
s_load_dwordx8  s[8:15], s[8:9], 0x10
```

squidbus:
Here's the CFG for the shader as well: [cfg.txt](https://github.com/user-attachments/files/21440778/cfg.txt)

squidbus:
Okay I did a bit more thorough debugging and I was wrong on that, it's a different issue this time with divergence I think:
```
[Render.Vulkan] <Critical> resource_tracking_pass.cpp:441 TrackSharp: "0x53c" is reachable from "0x9c"
[Render.Vulkan] <Critical> resource_tracking_pass.cpp:441 TrackSharp: "0x53c" is reachable from "0x4e8"
```

In block `0x9c`:
```
/*0000000000ac*/ s_load_dwordx8  s[12:19], s[8:9], 0x0
```

In block `0x4e8`:
```
/*0000000004f8*/ s_load_dwordx8  s[12:19], s[8:9], 0x0
```

In block `0x53c`:
```
/*000000000550*/ image_sample    v1, v[3:6], s[12:19], s[20:23] dmask:8
```

In this case I think the sources are identical so it reads both as valid?

Relevant disassembly: [relevant.txt](https://github.com/user-attachments/files/21440810/relevant.txt)

EDIT: Even past that accounting for duplicates, it seems then you hit an issue with the sampler in the same instruction, which is actually different:

In block `0x9c`:
```
/*0000000000b8*/ s_load_dwordx8  s[20:27], s[8:9], 0x10
```

In block `0x4e8`:
```
/*0000000004f4*/ s_load_dwordx4  s[20:23], s[10:11], 0x0
```

This heuristic for `Valid()` seems to work to discard the wrong sampler: `return unused0 == 0 && unused1 == 0;`, but I'm not sure how reliable it is overall.

Even past that though there are even more issues with `Image` resources and divergence, sometimes with 4 sources.

raphaelthegreat:
I feel this is exposing existing issues with tracking that happened to work before by chance, or maybe didn't. But I will take a look

squidbus:
> I feel this is exposing existing issues with tracking that happened to work before by chance, or maybe didn't. But I will take a look

You're probably right that it's likely issues with divergence that were being papered over before, however if we don't have full handling for sharps in divergent conditions I feel we shouldn't be asserting on it yet.

raphaelthegreat:
I can downgrade this to a warning and pick the first source (alongside adding duplicate handling)

raphaelthegreat:
Btw I think another check we can do is if a source dominates another source in the original CFG, that should solve cases like the one you describe

GHU7924:
[CUSA00076] The Order: 1886

`[Debug] <Critical> resource_tracking_pass.cpp:383 operator(): Assertion Failed!`

[CUSA00076.log](https://github.com/user-attachments/files/21442786/CUSA00076.log)

UPD.: Removed information not related to PR.

raphaelthegreat:
I'm not sure why you have readbacks disabled and reporting vertex explosions in TLG with this PR, this is a known issue for a long time that readbacks fixed

raphaelthegreat:
> Sorry for writing this here, but I do not know what PR affected the sound in the game

Writing in random PRs about unrelated issues doesn't help yes, if you don't know what caused it open an issue

raphaelthegreat:
Newest commit I've replaced validity tracking with dominance analysis, since former can be inconsistent.
Also added the program counter to image instruction flags, helps a lot with debugging
@squidbus Can you check if this fixes your issue

raphaelthegreat:
I realized dominance analysis can also be used to implicitly do reachability analysis (the function returns true if a path from block to dest_block without maybe_dominator doesn't exist, so if a path doesn't exist in general it will also return true) so merged the passes into a single one

squidbus:
Looks like it's working now, yes (on both recent commits, before and after merging the passes).

GHU7924:
[CUSA00076] The Order: 1886

`[Debug] <Critical> resource_tracking_pass.cpp:390 FindSharpSources: Unreachable code!`
`Unable to find sharp sources`

[CUSA00076.log](https://github.com/user-attachments/files/21443270/CUSA00076.log)

raphaelthegreat:
> [CUSA00076] The Order: 1886
> 
> `[Debug] <Critical> resource_tracking_pass.cpp:390 FindSharpSources: Unreachable code!` `Unable to find sharp sources`
> 
> [CUSA00076.log](https://github.com/user-attachments/files/21443270/CUSA00076.log)

This should be fixed

GHU7924:
>This should be fixed

Yes, everything is fine now.

StevenMiller123:
Seems like this PR regresses KNACK™
```
[Debug] <Critical> resource_tracking_pass.cpp:390 FindSharpSources: Unreachable code!
Unable to find sharp sources pc=0x538
```
[CUSA00068.log](https://github.com/user-attachments/files/21445471/CUSA00068.log)

StevenMiller123:
This PR also regresses updated versions of The Last of Us™ Remastered, the game is crashing on an internal assert again.

[CUSA00552 PR.log](https://github.com/user-attachments/files/21445632/CUSA00552.PR.log)
[CUSA00552 Main.log](https://github.com/user-attachments/files/21445633/CUSA00552.Main.log)

StevenMiller123:
Call of Duty® now crashes on
```
[Debug] <Critical> resource_tracking_pass.cpp:388 FindSharpSources: Unreachable code!
Bindless sharp access detected pc=0x1b8
```
On main, it crashes on
```
[Debug] <Critical> resource_tracking_pass.cpp:452 operator(): Assertion Failed!
Unable to find image sharp source
```
[CUSA23827 PR.log](https://github.com/user-attachments/files/21445653/CUSA23827.PR.log)
[CUSA23827 Main.log](https://github.com/user-attachments/files/21445654/CUSA23827.Main.log)

StevenMiller123:
DOOM Eternal regressed (slightly). It now crashes on:
```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:501 CompileModule: Compiling fs shader 0xb99d8d13 
[Debug] <Critical> resource_tracking_pass.cpp:388 FindSharpSources: Unreachable code!
Bindless sharp access detected pc=0x58
```

On main, that shader compiles and it crashes on the next shader
```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:501 CompileModule: Compiling fs shader 0xb99d8d13 
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:501 CompileModule: Compiling vs shader 0xa97c6b18 
[Debug] <Critical> info.h:280 operator(): Assertion Failed!
```
[CUSA13338 PR.log](https://github.com/user-attachments/files/21445690/CUSA13338.PR.log)
[CUSA13338 Main.log](https://github.com/user-attachments/files/21445692/CUSA13338.Main.log)

mikusp:
Detroit: Become Human regresses with this PR:
```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:501 CompileModule: Compiling fs shader 0x130b2f21 
[Debug] <Critical> resource_tracking_pass.cpp:388 FindSharpSources: Unreachable code!
Bindless sharp access detected pc=0x798
```
[CUSA08308_PR.txt](https://github.com/user-attachments/files/21446118/CUSA08308_PR.txt)
[CUSA08308_main.txt](https://github.com/user-attachments/files/21446119/CUSA08308_main.txt)

raphaelthegreat:
Bindless sharp access assertion is not a regression, it is when a sharp is being loaded from a buffer instead of constant cache, before it was not handled at all and it would pick random stuff

raphaelthegreat:
> Seems like this PR regresses KNACK™
> 
> ```
> [Debug] <Critical> resource_tracking_pass.cpp:390 FindSharpSources: Unreachable code!
> Unable to find sharp sources pc=0x538
> ```
> 
> [CUSA00068.log](https://github.com/user-attachments/files/21445471/CUSA00068.log)

Latest commit is fix for this

StevenMiller123:
Can confirm, KNACK™ regression is fixed.
[CUSA00068.log](https://github.com/user-attachments/files/21446605/CUSA00068.log)

raphaelthegreat:
> This PR also regresses updated versions of The Last of Us™ Remastered, the game is crashing on an internal assert again.
> 
> [CUSA00552 PR.log](https://github.com/user-attachments/files/21445632/CUSA00552.PR.log) [CUSA00552 Main.log](https://github.com/user-attachments/files/21445633/CUSA00552.Main.log)

I'm not sure what to do with this, since it's not shader related

raphaelthegreat:
Knack probably needs more thorough testing, I've fixed like 4 sharp tracking crashes on it and all related to the aniso disable pattern (seems it was just broken before and the BFS would just skip over it and find random things)

Randomuser8219:
On the older compiles of this PR Knack always crashed no matter if I had Debug log filter on or not. It's fixed now
image
The game looks exactly as it does on main, with some broken colors, weird spots on ocean, smearing effect on skybox and LSD.
image

StevenMiller123:
Seems like KNACK™ is improved over main now, at least on my setup. Main crashes on a device lost at the start of 1-2 (see https://github.com/shadps4-emu/shadPS4/issues/2192#issuecomment-2813703402 for more details), while this PR gets slightly further and crashes silently during shader compilation.

I can't get the game running with validation layers enabled, as it seems to worsen the PM4 type 0 race condition the game has on start up.
[CUSA00068 main.log](https://github.com/user-attachments/files/21448084/CUSA00068.main.log)
[CUSA00068 PR.log](https://github.com/user-attachments/files/21448085/CUSA00068.PR.log)

raphaelthegreat:
Try turning [this](https://github.com/shadps4-emu/shadPS4/blob/c863b350b22a8f1df1780338011edb88dbd0dcb7/src/core/libraries/gnmdriver/gnmdriver.cpp#L2161) log into Info/Warning/Error it helps the race condition by slowing down submitter thread

Randomuser8219:
> Try turning [this](https://github.com/shadps4-emu/shadPS4/blob/c863b350b22a8f1df1780338011edb88dbd0dcb7/src/core/libraries/gnmdriver/gnmdriver.cpp#L2161) log into Info/Warning/Error it helps the race condition by slowing down submitter thread

I use the Lib.GnmDriver:Debug log filter to get past the race condition. It works somewhat the same way.

StevenMiller123:
Went the log filter route, here's tests with validation enabled. I've got this build compiling locally now, should be able to check where KNACK™ crashes in a little while.

[CUSA00068 main.log](https://github.com/user-attachments/files/21448144/CUSA00068.main.log)
[CUSA00068 PR.log](https://github.com/user-attachments/files/21448142/CUSA00068.PR.log)

Randomuser8219:
The crash could be because AMD doesn't support BC6H_UFLOAT_BLOCK and Bc5UnormBlock. The Bc5UnormBlock is the last error it gives before crashing from your log.
It doesn't crash there on NVIDIA because it supports those formats.

StevenMiller123:
Seems like the crash is relevant to the aniso disable stuff.
`[Debug] <Critical> value.h:302 operator(): Assertion Failed!`
image

Randomuser8219:
[shad_log.txt](https://github.com/user-attachments/files/21448239/shad_log.txt)
This is my NVIDIA log with extra validation logging enabled in Vulkan SDK Configurator (enabled Warning severity logging, along with GPU-assisted validation). The game is having a lot of storage buffer shenanigans.

raphaelthegreat:
Aniso crash should be fixed now

StevenMiller123:
Can confirm, the crash is fixed. I'm not encountering any new crashes so far.
[CUSA00068.log](https://github.com/user-attachments/files/21448391/CUSA00068.log)

Randomuser8219:
This brings the 1.10 patch of Knack in-game.
[shad_log.txt](https://github.com/user-attachments/files/21454770/shad_log.txt)

rafael-57:
is this good to go now?

squidbus:
Seems the issues are resolved so I assume so

PR #3328, opened 2025-07-26
Ds4 speaker audio base

nickci2002:
Basic functionality that separates the pad speaker channel from the rest of the audio. This does not include gui components, which need to be implemented.


rainmakerv3:
PR submitted to your branch for GUI

rainmakerv3:
```
        std::string port_name = port.type != OrbisAudioOutPort::PadSpk
                                    ? Config::getPadSpkOutputDevice()
                                    : Config::getMainOutputDevice();
        u32 devId;

        if (port_name == "None") {
            stream = nullptr;
            return;
        } else if (port_name == "Default Device") {
            devId = SDL_AUDIO_DEVICE_DEFAULT_RECORDING;
        } else {
            try {
                devId = static_cast<u32>(std::stoul(port_name));
            } catch (const std::exception& e) {
                LOG_ERROR(Lib_AudioOut, "Invalid audio output device: {}", port_name);
                stream = nullptr;
                return;
            }
        }
```

1) first check looks like it's meant to be port.type ==

2) default devID looks like it's meant to be SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK

3) I think there should be a check that if the chosen device isn't connected, it falls back to default playback instead of a null stream

src/core/libraries/audio/sdl_audio.cpp
```diff
@@ -21,8 +21,30 @@ class SDLPortBackend : public PortBackend {
...
+        std::string port_name = port.type != OrbisAudioOutPort::PadSpk
+                                    ? Config::getPadSpkOutputDevice()
+                                    : Config::getMainOutputDevice();
+        u32 devId;
+
+        if (port_name == "None") {
+            stream = nullptr;
+            return;
+        } else if (port_name == "Default Device") {
+            devId = SDL_AUDIO_DEVICE_DEFAULT_RECORDING;
```
squidbus:
```suggestion
            devId = SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;
```

nickci2002:
Whoops, my bad

rainmakerv3:
thanks, the only other thing I'm not sure of is if this how to get the SDL device ID from the chosen device name

devId = static_cast<u32>(std::stoul(port_name));

nickci2002:
Oh, that code was from shadow's mic PR. If it doesn't work, just lmk.

src/common/config.cpp
```diff
@@ -844,6 +872,8 @@ void save(const std::filesystem::path& path) {
     data["Input"]["isMotionControlsEnabled"] = isMotionControlsEnabled;
     data["Input"]["useUnifiedInputConfig"] = useUnifiedInputConfig;
     data["Input"]["micDevice"] = micDevice;
+    data["Output"]["mainOutputDevice"] = mainOutputDevice;
+    data["Output"]["padSpkOutputDevice"] = padSpkOutputDevice;
```
squidbus: Would prefer `Audio` as the category here (and above).

src/core/libraries/audio/sdl_audio.cpp
```diff
@@ -21,8 +21,29 @@ class SDLPortBackend : public PortBackend {
...
+        u32 devId;
+
+        if (port_name == "None") {
+            stream = nullptr;
+            return;
+        } else if (port_name == "Default Device") {
+            devId = SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;
+        } else {
+            try {
+                devId = static_cast<u32>(std::stoul(port_name));
```
squidbus, line 38: Would make more sense to be able to specify your audio device by name rather than the device number, which may not be stable.

squidbus (commented):
Sorry for the delay, two comments (also needs a rebase).

src/core/libraries/audio/sdl_audio.cpp
nickci2002: There's no function that immediately gets the device id by name in SDL. I added a check to assert the devid has the name specified in the config file. This should make it stable.

nickci2002 (commented):
See updates in sdl_audio.cpp

squidbus:
Your PR doesn't seem to be rebased right, it includes a bunch of commits from main.

nickci2002:
How do I fix the rebase issue?

nickci2002:
I'm moving this to a new PR. This one is too far gone...

nickci2002:
See #3607 for new pr

PR #3330, opened 2025-07-26
Make libSceRtc fully HLE

kalaposfos13:
Unless I'm missing something, this library had an almost fully complete HLE reimplementation already, so I'm not sure why this is still in the list of LLE loadable libraries. I implemented the rest of the stubbed functions (well moved the stubbed parts to the kernel technically) as well, and removed it from the list of LLE loadable libraries. If there is a preexisting reason as to why it wasn't HLE'd before, please do let me know.


squidbus:
Seems okay to me, if anyone can do some regression testing would be good to know this is working as expected.

bigol83:
> Seems okay to me, if anyone can do some regression testing would be good to know this is working as expected.

i checked with my games (16 games) and it doesn't seem to have any regression

PR #3336, opened 2025-07-26
Apply cheat/patch only if AppVer matches game version

marecl:
Current behaviour:  
Multiple patches for the same game, sharing the same metadata except for AppVer are enabled/disabled at once.  
In effect, multiple potentially exclusive patches are enabled.

Patched behaviour:  
Only patch with AppVer matching game version is enabled. "Mask" still applies.  

Addition:  
Removed redundancies in saving routine

### DIFF
--- a/src/qt_gui/cheats_patches.cpp
+++ b/src/qt_gui/cheats_patches.cpp
@@ -389,61 +389,41 @@ void CheatsPatches::onSaveButtonClicked() {
                 xmlWriter.writeStartElement(xmlReader.name().toString());
 
                 QString name = xmlReader.attributes().value("Name").toString();
+                QString version = xmlReader.attributes().value("AppVer").toString();
+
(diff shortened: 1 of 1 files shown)


georgemoralis:
@DanielSvoboda can you check this?

DanielSvoboda:
> @DanielSvoboda can you check this?

I'll take a look tomorrow 👍

DanielSvoboda:
This really improved the redundancies. However, I'm a little confused about what is being fixed. Does your PR fix the problem that patches from other AppVers that are active are applied?

marecl:
@DanielSvoboda yes. if AppVer is the only difference between cheats, all of them are set as enabled

DanielSvoboda:
Now I understand, I managed to replicate it, and yes, it solves the problem 👍

PR #3338, opened 2025-07-27
fixed RetrieveNetmask for windows

georgemoralis:
This is probably fixes netmask retrival for windows . 

@Missake212 i think you spotted the issue if you can test


Missake212:
Looks correct now.

``[Lib.NetCtl] <Debug> netctl.cpp:204 sceNetCtlGetInfo: netmask: 255.255.255.0``

[CUSA00787.log](https://github.com/user-attachments/files/21453954/CUSA00787.log)

PR #3350, opened 2025-07-30
video_core: garbage collector (part 1)

LNDF:
The intent is to implement a garbage collector for GPU resources.

This garbage collector is based on the yuzu implementation.

For anyone testing, you are noit suposed to see any inmediate effects upon launching the game. This is a garbage collector, and the goal is to delete OLD_DATA (😉) from GPU in order to keep memory usage under control.

A note regarding DMA: Due to how DMA is currently implemented, garbage collection will not work on buffers on games that make use of DMA. This is because current DMA implementation syncs all buffers, touching them all. Referencing my other DMA PR (#3197) since it could be solved there.

Checklist:
- [x] Memory usage collecting
- [x] Garbage collector for buffer cache
- [x] Garbage collector for texture cache*

\* The texture cache will not garbage collect non-linear GPU modified images. This will be fixed in a future PR.


src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -1154,6 +1182,34 @@ void BufferCache::WriteDataBuffer(Buffer& buffer, VAddr address, const void* val
...
+    const bool aggressive = total_used_memory >= critical_gc_memory;
+    const u64 ticks_to_destroy = std::min<u64>(aggressive ? 80 : 160, gc_tick);
+    int max_deletions = aggressive ? 64 : 32;
+    const auto clean_up = [&](BufferId buffer_id) {
+        if (max_deletions == 0) {
+            return;
+        }
+        --max_deletions;
+        Buffer& buffer = slot_buffers[buffer_id];
+        InvalidateMemory(buffer.CpuAddr(), buffer.SizeBytes());
```
raphaelthegreat: This should probably asynchronously download the buffer data back to CPU in case it was GPU modified and it's used again in the future

LNDF: I think I got the async buffer downloading right.
Now, if I try to do the same with images, bloodborne crashes. We want to get back image data since games (bb, driverclub...) modify it and it gets lots, leading to visual artifacts. Do I need add handling for non-linear images?

LNDF:
We are currently working around the issue by just skipping GC on non linear gpu modified images. I don't consider this pr ready yet.

src/video_core/buffer_cache/buffer_cache.cpp
raphaelthegreat: Downloading non linear images requires a tiler shader which does not exist currently. I think skipping GPU modified images from GC is fine solution

src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -852,6 +881,77 @@ void TextureCache::UntrackImageTail(ImageId image_id) {
...
+        ticks_to_destroy = std::min(ticks_to_destroy, gc_tick);
+        num_deletions = aggresive ? 40 : pressured ? 20 : 10;
+    };
+    const auto clean_up = [&](ImageId image_id) {
+        if (num_deletions == 0) {
+            return true;
+        }
+        --num_deletions;
+        auto& image = slot_images[image_id];
+        const bool download = True(image.flags & ImageFlagBits::GpuModified);
```
raphaelthegreat: You should also check that the image does not have the Dirty flag (so it's neither gpu or cpu dirty) otherwise you risk trashing CPU data with stale image data

src/video_core/buffer_cache/buffer_cache.cpp
LNDF: Good. Currently, all non linear GPU modified images are skipped. Linear ones are downloaded

src/video_core/texture_cache/texture_cache.cpp
LNDF: I think this is fixed now.

src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -641,6 +689,20 @@ void Instance::CollectToolingInfo() const {
     }
 }
 
+u64 Instance::GetDeviceMemoryUsage() const {
```
raphaelthegreat, line 692: This might need to use vma function like vmaGetHeapBudgets instead, as the allocator could batch allocate memory blocks for future use causing this function to overreport memory and GC try to free it, when not needed

src/video_core/texture_cache/image.h
```diff
@@ -111,6 +111,11 @@ struct Image {
         return track_addr != 0 && track_addr_end != 0;
     }
 
+    bool SafeToDownload() const {
+        return True(flags & ImageFlagBits::GpuModified) &&
+               False(flags & (ImageFlagBits::GpuDirty | ImageFlagBits::CpuDirty));
```
raphaelthegreat, line 116: There is already a Dirty flag that is bitwise or for all possible dirty flags

src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -641,6 +689,20 @@ void Instance::CollectToolingInfo() const {
...
+    vk::PhysicalDeviceMemoryBudgetPropertiesEXT memory_budget_props{};
+    vk::PhysicalDeviceMemoryProperties2 props = {
+        .pNext = &memory_budget_props,
+    };
+    physical_device.getMemoryProperties2(&props);
+
+    u64 total_usage = 0;
+    for (const size_t heap : valid_heaps) {
```
raphaelthegreat, line 700: Doesn't this include host memory usage from i.e staging buffer. This should probably only account of device local allocations

LNDF: total_heap is populated on CollectPhysicalMemoryInfo in line 649. There is a if statement where if it is not integrated gpu and it is not local memory will skip that heap.

src/video_core/renderer_vulkan/vk_instance.cpp
LNDF: Will take a look at it

PR #3351, opened 2025-07-30
Implement BufferAtomicFCmpSwap

kalaposfos13:
Used by Knack 2.
https://github.com/shadps4-emu/shadPS4/issues/496#issuecomment-3065989966

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
@@ -111,6 +111,21 @@ Id BufferAtomicU32CmpSwap(EmitContext& ctx, IR::Inst* inst, u32 handle, Id addre
     });
 }
 
+Id BufferAtomicF32CmpSwap(EmitContext& ctx, IR::Inst* inst, u32 handle, Id address, Id value,
+                          Id cmp_value,
+                          Id (Sirit::Module::*atomic_func)(Id, Id, Id, Id, Id, Id, Id)) {
+    const auto& buffer = ctx.buffers[handle];
+    if (const Id offset = buffer.Offset(PointerSize::B32); Sirit::ValidId(offset)) {
+        address = ctx.OpIAdd(ctx.F32[1], address, offset);
+    }
+    const auto [id, pointer_type] = buffer.Alias(PointerType::U32);
+    const Id ptr = ctx.OpAccessChain(pointer_type, id, ctx.f32_zero_value, address);
+    const auto [scope, semantics]{AtomicArgs(ctx)};
+    return AccessBoundsCheck<32>(ctx, address, buffer.Size(PointerSize::B32), [&] {
+        return (ctx.*atomic_func)(ctx.F32[1], ptr, scope, semantics, semantics, value, cmp_value);
+    });
+}
+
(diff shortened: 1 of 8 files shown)


src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
```diff
@@ -111,6 +111,21 @@ Id BufferAtomicU32CmpSwap(EmitContext& ctx, IR::Inst* inst, u32 handle, Id addre
     });
 }
 
+Id BufferAtomicF32CmpSwap(EmitContext& ctx, IR::Inst* inst, u32 handle, Id address, Id value,
+                          Id cmp_value,
+                          Id (Sirit::Module::*atomic_func)(Id, Id, Id, Id, Id, Id, Id)) {
+    const auto& buffer = ctx.buffers[handle];
+    if (const Id offset = buffer.Offset(PointerSize::B32); Sirit::ValidId(offset)) {
+        address = ctx.OpIAdd(ctx.F32[1], address, offset);
```
squidbus, line 119: Address is not F32

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
```diff
@@ -111,6 +111,21 @@ Id BufferAtomicU32CmpSwap(EmitContext& ctx, IR::Inst* inst, u32 handle, Id addre
...
+    }
+    const auto [id, pointer_type] = buffer.Alias(PointerType::U32);
```
squidbus, line 121: This is where it should be F32

squidbus:
According to the spec, `OpAtomicCompareExchange` only accepts integers so I don't think this will work.

src/shader_recompiler/backend/spirv/emit_spirv_atomic.cpp
kalaposfos13: With that I'd get a crash with the following error:
```
shadps4: /media/kalaposfos/Shared/shadps4/repos/shadps4-dev/externals/sirit/src/stream.h:36: Sirit::OpId::OpId(spv::Op, Id): Assertion `result_type.value != 0' failed.
```
which is probably related to your comment below

squidbus: Hmm it seems the other float atomics are fine using the U32 pointer so I guess this part isn't needed?

(for reference that error is just because nothing is telling the generator to set up a F32 alias for the buffer)

mikusp: This opcode should probably be added to BufferDataType in resource_tracking_pass.cpp to create the alias

Niram7777:
> According to the spec, `OpAtomicCompareExchange` only accepts integers so I don't think this will work.

I think we can just Compare And Swap by casting the pointer from f32* to u32* ?
this is what is done for example in some app in CUDA apparently

red-prig:
In Vulkan, type casting for pointers is only available for BDA buffers.

raphaelthegreat:
Descriptor aliasing can achieve the same thing, and a system of buffer type aliases already exists

PR #3352, opened 2025-07-30
Implement V_CMP_GT_U64

kalaposfos13:
Used by Knack 2.

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
@@ -535,6 +535,7 @@ Id EmitReadLane(EmitContext& ctx, Id value, Id lane);
 Id EmitWriteLane(EmitContext& ctx, Id value, Id write_value, u32 lane);
 Id EmitBallot(EmitContext& ctx, Id bit);
 Id EmitBallotFindLsb(EmitContext& ctx, Id mask);
+Id EmitGroupAny(EmitContext& ctx, Id bit);
 Id EmitDataAppend(EmitContext& ctx, u32 gds_addr, u32 binding);
 Id EmitDataConsume(EmitContext& ctx, u32 gds_addr, u32 binding);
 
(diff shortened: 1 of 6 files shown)


squidbus:
@raphaelthegreat Can you check this as I'm still not sure on the 64-bit compare ops

raphaelthegreat:
64-bit compares seem relatively safe as they are probably not used by thread masks, but this needs to be confirmed by looking the shader snippet if it uses 64-bit integers. If it does, it also needs to checked if it also uses bitwise opcodes, cause that needs additional tracking in translator to know if an SGPR is thread mask or not

kalaposfos13:
@raphaelthegreat where can I check that?

kalaposfos13:
Well I found this in the .irprogram.txt file, not sure if it's the relevant part:
```
Block $67
[000055d7b73144e8] %915   = CompositeConstructU32x2 %782, %784 (uses: 1)
[000055d7b7314588] %916   = PackUint2x32 %915 (uses: 1)
[000055d7b7314628] %917   = UGreaterThan64 %916, #0 (uses: 1)
[000055d7b7314bc8] %918   = SelectF32 %917, #-0, #0 (uses: 1)
[000055d7b7314c68] %919   = BitCastU32F32 %918 (uses: 1)
[000055d7b7314ee8] %920   = BitwiseOr32 %777, %919 (uses: 1)
[000055d7b73150c8] %921   = BitwiseOr32 #536870912, %920 (uses: 2)
[000055d7b7315348] %922   = CompositeConstructU32x2 %102, %103 (uses: 1)
[000055d7b7315848] %923   = ReadConst (flags=0x12)  %922, #2 (uses: 2)
[000055d7b7315ca8] %924   = ShiftLeftLogical32 %912, #1 (uses: 4)
[000055d7b73775e8]          Void
[000055d7b7377688]          Void
[000055d7b7377728]          Void
[000055d7b73777c8] %925   = ShiftRightLogical32 %443, #2 (uses: 1)
[000055d7b7316248] %926   = BufferAtomicSwap32 #1, %925, %912 (uses: 1)
[000055d7b7316428] %927   = BitwiseAnd32 #2147483648, %921 (uses: 1)
[000055d7b7316608] %928   = IAdd32 %924, #1 (uses: 1)
[000055d7b73167e8] %929   = INotEqual32 #0, %927 (uses: 1)
[000055d7b7316ba8] %930   = UGreaterThan32 %923, %924 (uses: 1)
[000055d7b73170a8] %931   = ConditionRef %930 (uses: 0)
```

raphaelthegreat:
You need to disassemble the original shader with clrxdisasm (so the bin file of the same name).

raphaelthegreat:
The current version of the PR is okay to me, however I feel it should be stripped down to only this case, as I doubt the other opcodes were verified to be true 64bit and this turned out to be a thread mask op

PR #3353, opened 2025-07-30
select

mikusp:
Select for sockets (and regular files on Linux)

### DIFF
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -1075,6 +1075,7 @@ s32 PS4_SYSV_ABI sceKernelUnlink(const char* path) {
     }
     return result;
 }
+#ifdef _WIN32
 
 static HANDLE fileToHandle(std::FILE* file) {
     if (!file)
@@ -1087,7 +1088,7 @@ static HANDLE fileToHandle(std::FILE* file) {
     HANDLE h = reinterpret_cast<HANDLE>(_get_osfhandle(fd)); // Convert to HANDLE
     return h;
 }
-
+#endif
 s32 PS4_SYSV_ABI posix_select(int nfds, fd_set* readfds, fd_set* writefds, fd_set* exceptfds,
                               OrbisKernelTimeval* timeout) {
     LOG_INFO(Kernel_Fs, "nfds = {}, readfds = {}, writefds = {}, exceptfds = {}, timeout = {}",
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -1075,36 +1075,6 @@ s32 PS4_SYSV_ABI sceKernelUnlink(const char* path) {
     }
     return result;
 }
-#ifdef _WIN32
-
-static HANDLE fileToHandle(std::FILE* file) {
-    if (!file)
-        return INVALID_HANDLE_VALUE;
-
-    int fd = _fileno(file); // Get the file descriptor
-    if (fd == -1)
-        return INVALID_HANDLE_VALUE;
-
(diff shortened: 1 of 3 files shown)


src/core/libraries/kernel/file_system.cpp
```diff
@@ -1068,6 +1076,119 @@ s32 PS4_SYSV_ABI sceKernelUnlink(const char* path) {
...
+
+    if (ret < 0) {
+#ifndef _WIN32
+        auto error = errno;
+#else
+        auto error = WSAGetLastError();
+#endif
+        LOG_ERROR(Kernel_Fs, "native select call failed with {} ({})", error,
+                  Common::NativeErrorToString(error));
+        SetPosixErrno(error);
```
StevenMiller123: This still won't work properly on Windows, since winsock uses their net error codes here, and SetPosixErrno was designed around platform-specific errno values instead.

StevenMiller123 (changes requested):

georgemoralis:
Updated select for windows

 Sockets via WSAPoll on Windows
Files via WaitForMultipleObjects (read) and assumed writable (write)

georgemoralis:
file handling on windows is a bit hackish , need to get it proper with the new abstract OS level filesystem that handles HANDLE on windows instead of FILE object

src/core/libraries/kernel/file_system.cpp
```diff
@@ -1068,6 +1076,236 @@ s32 PS4_SYSV_ABI sceKernelUnlink(const char* path) {
...
+    if (!file)
+        return INVALID_HANDLE_VALUE;
+
+    int fd = _fileno(file); // Get the file descriptor
+    if (fd == -1)
+        return INVALID_HANDLE_VALUE;
+
+    HANDLE h = reinterpret_cast<HANDLE>(_get_osfhandle(fd)); // Convert to HANDLE
+    return h;
+}
```
StevenMiller123: This needs to be encased in an ifdef, since HANDLE is a Windows type.

StevenMiller123 (changes requested):

src/core/libraries/kernel/file_system.cpp
```diff
@@ -1067,6 +1075,324 @@ s32 PS4_SYSV_ABI sceKernelUnlink(const char* path) {
...
+    if (read_host.fd_count == 0 && write_host.fd_count == 0 && except_host.fd_count == 0) {
+        LOG_WARNING(Kernel_Fs, "No sockets in fd_sets, select() will return immediately");
+    }
+
+    if (readfds)
+        FD_ZERO_POSIX(readfds);
+    if (writefds)
+        FD_ZERO_POSIX(writefds);
+    if (exceptfds)
+        FD_ZERO_POSIX(exceptfds);
```
StevenMiller123: In games that only supply device fds, zeroing out the fd_set_posix parameters here will result in the function returning with empty output fds despite returning 0. Probably could be solved by adding the device fds to the `*_ready` sets above, then adding them back in alongside the regular file fds.
Addressing this problem will fix a regression in Cyberpunk 2077.

StevenMiller123 (changes requested):
I only ran into the one regression noted here with Cyberpunk 2077, everything else seems to behave fine.

StevenMiller123 (approved):

Randomuser8219:
Not sure if this PR caused it, but LBP3 gives a different recvfrom error:
[Lib.Net] <Error> sys_net.cpp:244 sys_recvfrom: error code returned: 35

StevenMiller123:
Most of LBP3's errors are tied to our lack of a P2P socket implementation.
The error you're seeing right now is a stub error return I added to the P2P recvfrom implementation so our previous error stub wouldn't cause strange issues in games.

PR #3360, opened 2025-08-01
Core: Simulate write-only file access with read-write access

StevenMiller123:
The standard library functions we're using for opening files will erase files when opening with write access. This differs from real hardware behavior, where the file opens without getting erased.

To properly emulate this behavior, I've made write-only access identical to read-write access, and added checks in `read`, `posix_preadv`, and `readv` to ensure that reading from a file with write-only access fails in a hardware-accurate way. I've also created a separate create mode that uses write-access to create files to simplify the process of creating a file in `open` since read-write mode doesn't create files.

Handling this behavior properly fixes saving in Catherine: Full Body (CUSA15036)
image


StevenMiller123:
This PR should probably receive further testing before any merge, to ensure there aren't any other issues I've caused by messing with FileAccessMode::Write.

rafael-57:
image

image

Tested loading, updating older saves and also making a new one. No issues found in Bloodborne, even with multiple sessions.

rafael-57:
Anything blocking this other than need of more testing?

StevenMiller123:
I think the main holdup here comes from PR #3368. Once that's ready for merge, this PR becomes an unnecessary hack.

rafael-57:
> I think the main holdup here comes from PR #3368. Once that's ready for merge, this PR becomes an unnecessary hack.

ah nice

StevenMiller123:
Rebased with a slightly outdated main, for anyone who wants to test it.

georgemoralis:
since #3368 will take a while maybe we shall merge that as workaround?

StevenMiller123:
I kept this PR maintained for a reason, should be clear to merge if desired.
As far as I'm aware, the current version of this doesn't break anything.

georgemoralis:
since #3771 is highly workaround still , merging that and we can fix it later :D

PR #3365, opened 2025-08-01
Implement ORBIS_NET_SO_ERROR_EX in GetSocketOptions

kalaposfos13:
Used by LLE HTTP. Pinging @mikusp as networking is their expertise after all.


src/core/libraries/network/posix_sockets.cpp
```diff
@@ -405,6 +405,18 @@ int PosixSocket::GetSocketOptions(int level, int optname, void* optval, u32* opt
...
+        case ORBIS_NET_SO_ERROR_EX: {
+            socklen_t optlen_temp = *optlen;
+            auto retval = ConvertReturnErrorCode(
+                getsockopt(sock, level, SO_ERROR, (char*)optval, &optlen_temp));
+            *optlen = optlen_temp;
+            if (retval < 0) {
+                s32 r = *Libraries::Kernel::__Error();
+                *Libraries::Kernel::__Error() = 0;
+                return r;
+            }
```
mikusp, line 417: That if statement can be removed entirely and then it will be correct

PR #3368, opened 2025-08-02
[Draft] Use native file handlers

marecl:
This PR changes how IOFile is handled on OS-level.  
Improves compatibility and granularity with Orbis file-related kernel calls.  
Existing (in-app call) behaviour and usage is not affected (I think).  

Currently implemented for UNIX (POSIX calls).  
Windows implementation pending.  
Further tests are needed to confirm correct operation.  

Subject to changes.


src/core/libraries/kernel/file_system.cpp
```diff
@@ -108,51 +107,77 @@ s32 PS4_SYSV_ABI open(const char* raw_path, s32 flags, u16 mode) {
...
-    if (directory) {
-        if (!std::filesystem::is_directory(file->m_host_name)) {
+    if (truncate && read) {
+        // This combination is actually legal, but its behaviour is described as "undefined"
+        // Preventing potential data loss
+        LOG_WARNING(
+            Kernel_Fs,
+            "Read and Truncate flags specified. This makes no sense, neutering Truncate flag");
+        truncate = false;
+    }
```
StevenMiller123: We're emulating the PS4, not the POSIX spec.
Undefined behavior or not, we should follow how a PS4 behaves.
Since the truncate flag causes an error return in a read-only file system, and we emulate that properly, any data loss visible in emulator would be visible on real hardware too.

src/core/libraries/kernel/file_system.cpp
```diff
@@ -108,51 +107,77 @@ s32 PS4_SYSV_ABI open(const char* raw_path, s32 flags, u16 mode) {
...
             // If the opened file is not a directory, return ENOTDIR.
             // This will trigger when create & directory is specified, this is expected.
             h->DeleteHandle(handle);
             *__Error() = POSIX_ENOTDIR;
             return -1;
         }
+    }
 
-        file->type = Core::FileSys::FileType::Directory;
+    if (directory && (write || rdwr || truncate)) {
```
StevenMiller123, line 170: More of a nitpick than anything, but modifying this check to include truncate is unnecessary since truncate with any form of write access hits an earlier error return.

src/core/libraries/kernel/file_system.cpp
```diff
@@ -162,63 +187,17 @@ s32 PS4_SYSV_ABI open(const char* raw_path, s32 flags, u16 mode) {
...
-            // Write only
-            e = file->f.Open(file->m_host_name, Common::FS::FileAccessMode::Write);
-        } else if (rdwr) {
-            // Read and write
-            e = file->f.Open(file->m_host_name, Common::FS::FileAccessMode::ReadWrite);
-        }
+        e = file->f.Open(file->m_host_name, Common::FS::AccessModeOrbisToPOSIX(flags), mode);
     }
 
     if (e != 0) {
```
StevenMiller123, line 207: From what I can see, your `f.Open` call returns the native result (-1 on fail).
If you want to keep that as-is, this check here needs to pass through and check an std::error_code for the error instead.
Alternatively, switch `f.Open` to return the `std::error_code` value like it does on main, and this code will function properly.

src/core/libraries/kernel/file_system.cpp
```diff
@@ -1041,15 +1015,16 @@ s32 PS4_SYSV_ABI posix_unlink(const char* path) {
...
         *__Error() = POSIX_EPERM;
         return -1;
     }
 
     auto* file = h->GetFile(host_path);
     if (file == nullptr) {
         // File to unlink hasn't been opened, manually open and unlink it.
-        Common::FS::IOFile file(host_path, Common::FS::FileAccessMode::ReadWrite);
+        // marecl: *may* fail. IDK if it's supposed to exist (yet)
+        Common::FS::IOFile file(host_path, Common::FS::FileAccessMode::ReadExtended);
```
StevenMiller123: If the file to unlink doesn't exist, we should be setting errno to ENOENT and returning -1.
I believe that case should be caught by the `mnt->GetHostPath` call above though.

src/core/libraries/kernel/file_system.cpp
```diff
@@ -108,51 +107,77 @@ s32 PS4_SYSV_ABI open(const char* raw_path, s32 flags, u16 mode) {
...
+        return -1;
+    }
 
-        file->type = Core::FileSys::FileType::Directory;
+    if (directory && !read) {
+        // Cannot open directory without read
+        h->DeleteHandle(handle);
+        *__Error() = POSIX_EINVAL;
+        return -1;
+    }
```
StevenMiller123: Also more of a nitpick, but this will never be hit.
read will always be true by this point, since if neither read, write, or rdwr is specified, then the call will hit the EINVAL error case at the start of this function, and if either write or rdwr is specified, the call hits the EISDIR error case instead.

StevenMiller123 (changes requested):
I've looked primarily at the changes to file_system.cpp in this review, since I wrote much of the code currently in there.
I haven't dug too deeply into your changes to flag handling in `open`, but I know there's some incorrect changes there too.

src/core/libraries/kernel/file_system.cpp
marecl: Change: File is truncated in this check, flag is removed, opened normally later in the function

src/core/libraries/kernel/file_system.cpp
marecl: Change: removed `truncate`

src/core/libraries/kernel/file_system.cpp
marecl: No change

src/core/libraries/kernel/file_system.cpp
marecl: Change: removed comment

src/core/libraries/kernel/file_system.cpp
marecl: Change: Removed block

georgemoralis:
is that still on works? if so it must be rebased

marecl:
> is that still on works? if so it must be rebased

This particular branch no. Keeping it open to indicate WIP, but main course will be delivered via different branch.

PR #3377, opened 2025-08-03
Fix Play Time - Wasn't saving every minute

DanielSvoboda:
`QTimer::timeout` stopped working for some reason I couldn't identify 🙃
So replaced it with a `std::thread` and `sleep_for` to update play time every 60 seconds.


DanielSvoboda:
I also removed the ASSERT MSG, I don't think it makes sense, and at the beginning of UpdatePlayTime there is already a LOG_INFO with the same validation

PR #3382, opened 2025-08-04
SystemGesture

DanielSvoboda:
Just a stubbed version


src/core/libraries/system_gesture/SystemGesture.cpp
kalaposfos13, line 1: These files don't follow the project's file naming convention and are the only two PascalCase files in the repo

DanielSvoboda: I made a PR to fix this

PR #3391, opened 2025-08-05
Hotkey config changes

kalaposfos13:
I rewrote the entire backend for this, as the original solution was a bit hacky, and not very scalable, then @rainmakerv3 rewrote the frontend to work with it
The new code works like this:
- I removed the hardcoded hotkeys, and added output bindings for each of them (most of the hotkeys in the Ctrl F10 dev menu still remain, except for F10 bringing up the small FPS counter)
- I added the F keys as rebindable keys, plus added the PS4 "share" button as an input too, although that was already available, just under a different name (the Xbox controllers' "back" button)
- There's a new input config file called global.ini, which is a completely standard input config file, except for the fact that it will always get loaded alongside the main config, which by default will contain the list of the original keyboard hotkey bindings, and those are the main intended use case for it, but theoretically you can put any binding there


rainmakerv3:
I'm not sure if this is just on my end, but I seem to having a couple issues:

1) For example, setting a hotkey to L2+R2+L3 for fullscreen, the first time it works as expected, but then after that pressing just L3 will toggle fullscreen for me (tried on two controllers).

2) For the same combo input, it seems to be triggering the event multiple times per press (I was holding the triggers and pressing the button)

It may be something odd on my end because as I mentioned, the controller quit hotkey isn't working for me either (the other ones do), but maybe we could get another tester to be sure

rainmakerv3:
seems the non-working hotkey I mentioned is due to the mapping. mapping controller quit to l2+r2+l3 works just fine (other than the event triggering multiple times per press) but mapping the same to l2+r2+triangle doesn't seem to work for me

kalaposfos13:
@rainmakerv3 I pushed a commit that should fix the repeated inputs

rainmakerv3:
Can confirm, repeated inputs completely gone now!

Also this issue I mentioned is fixed as well:

For example, setting a hotkey to L2+R2+L3 for fullscreen, the first time it works as expected, but then after that pressing just L3 will toggle fullscreen for me (tried on two controllers).

Still can't seem to trigger L2+R2+triangle as a hotkey

kalaposfos13:
> Still can't seem to trigger L2+R2+triangle as a hotkey

This appears to be an input bug that's unrelated to hotkeys, and is an issue with how axis inputs are handled if they're bound to a binary output. I'll try to fix this as well, although I think it's outside of the scope of this PR so depending on the size of the fix I might open a new PR for it

rainmakerv3:
> Also this issue I mentioned is fixed as well:
> 
> For example, setting a hotkey to L2+R2+L3 for fullscreen, the first time it works as expected, but then after that pressing just L3 will toggle fullscreen for me (tried on two controllers).

It seems this issue is still present with certain combinations as well. I can reproduce it with R2 + triangle. I suppose it's the same core issue with the triggers.

Since these are pretty signficant issues IMO, you may want to consider just disabling the triggers as allowed inputs (just commenting out the axis motion event block should do that) for now

PR #3396, opened 2025-08-06
video_core: garbage collector (part 2)

LNDF:
This is a contiuation of #3350.

This will implement non-linear GPU modified image collection without losing the GPU modifications. This will reduce memory leaks even further (for example, in Bloodborne).

Currently, this depends on #3374 which implements tilers, required for non-linear image download.

PR #3404, opened 2025-08-08
video_core: Readback optimizations

raphaelthegreat:
Note there is still some cleanup to do, this is not final code. It can also cause freezes/bugs (hope not though)

### General idea

If you were to write a vulkan program that generates some data on the GPU and later want to access that data on the host, you need a sync operation to ensure the GPU has finished its work. Said sync operation is called a _fence_ because it makes the CPU wait for the GPU.

In a similar fashion the guest also has to use fences before reading GPU data on the host. Because of its unified memory, it doesn't have to copy said data to a host visible memory, but it still must sync with a fence operation before accessing it. The emulator can rely on that promise, that the guest will not overwrite nor read any GPU generated data before a fence operation has given it opportunity to sync with the GPU.

So the main idea of the PR is to attempt to detect these fence operations in the PM4 command stream and _defer_ read-protecting GPU modified pages until right before them. If a page read/write then happens before a fence, it will pass through without a flush, because the emulator can be sure the guest cannot access the data yet.

The aforementioned detection is not trivial though, because there is little indication to the emulator about what sync operations are used for. AMD GCN uses labels, 4 or 8 byte memory addresses, where "signal" packets write to and "wait" packets can wait on, or the host can poll in case of fence. All that means its close to impossible to detect actual wait of a fence. Instead, this PR implements a prepass which scans input command lists and tries to "guess" which packets act as fences and which not.

Possible packets that can write labels are EventWriteEos, EventWriteEop, WriteData (GFX) and ReleaseMem (ACB). There is a simple heuristic where if the label of a signal packet is waited by the GPU with WaitRegMem packet it is considered a GPU->GPU sync (something akin to pipeline barriers). It is in fact possible for a label to act both as fence and pipeline barrier so the heuristic can fail but its very unlikely.

Deferring read protections allows for some powerful optimizations, 2 of which are implemented here.

### Rewind indirect patch

The rewind packet has a misleading name, as it implies execution going back somewhere, but what it does is tell CP to drop all prefetched packets and reload them from memory. It is almost exclusively used for command list self modification (from a compute shader for example), which driveclub does for a dozen dispatches at the start of the frame. It uses a compute shader to patch the dimentions of the DispatchDirect PM4 packet before executing it. Why it didnt use an indirect dispatch I'm not sure.

Before readbacks this lead to launching a dispatch with garbage (often huge) dimensions, freezing the GPU. Readbacks, on the other hand, fixed it by read protecting the memory and flushing the modified data. That works but is very expensive; around a dozen flushes, one per patched dispatch.

Defering read protections allows emulator to reach rewind packet before a flush. Then the emulator can scan the pending GPU ranges inside the current command list, check they are dispatch dimention patches and convert the direct dispatch into an indirect dispatch. The latter reads dimentions from GPU buffers avoiding need for flushing memory.

### Preemptive buffer downloads

This optimization is a lot more general then above one and should affect all games that rely on readbacks. It is possible to implement without CPU fence detection, but it makes the implementation more efficient because the emulator can batch preemptive download copies upon reaching the fence.

The idea is to track how many times a page has been flushed and if that number exceeds a threshold, any future GPU data inside it will be copied to host asynchronously. If a flush is triggered, the GPU thread simply has to wait for the GPU to finish and copy data to guest memory. The advatange here is the reduction or (in certain cases) elimination of the wait time, as GPU likely has had time to catch up to host. In addition, once the wait has been done, the rest of preemptive downloads become "free" and don't need further stalls.

- [x] Add config option to control aggressiveness of fence detection
- [ ] Revoke preempt status from pages that stop being flushed


bigol83:
Quickly checked Bloodborne, for me there is a 10-15 fps improvement in this PR compared to master

squidbus:
For your build error, need to use std::bit_cast

GHU7924:
@raphaelthegreat , tell me how to test this PR correctly and what options should be activated, because I don’t quite understand how to do this and give you the necessary information.

1) Should all games be tested with options `Enable Readbacks` and `Enable Readbacks Linear Images` enabled?
2) How can I even tell if a particular game requires this option? (I don't know for sure, but there may be games that don't care about this option)

I don't provide logs yet because I might be doing something wrong, but I will report bugs. I noticed some unusual behavior in 4 of my games:

1) Bloodborne

The problem that exists on Intel processors has worsened (Without readbacks enabled).


If you enable the two options that I mentioned above, the picture becomes like this (in Main the game also looks like this for me in this place).


2) TLG 

Without readbacks enabled lines `Continue` and `Options` were not displayed in the menu, but with the options enabled, these lines were. 


I also get a crash when I try to pull the second spear out of Trico (the very beginning of the game), this might be worth testing further.

3) inFAMOUS™ First Light

An error has returned (Without readbacks enabled):

`[Debug] <Critical> vk_presenter.cpp:646 operator(): Assertion Failed!`
`Device lost during waiting for a frame`

With Readbacks options I can get into the game, but I get this picture:




4) The Order: 1886

Regardless of whether reverse reads are enabled or not, the game returns this error:

`[Debug] <Critical> vk_scheduler.cpp:166 operator(): Assertion Failed!`
`Device lost during submit`

raphaelthegreat:
@GHU7924 Testing of this PR should be with readbacks enabled (a lot of code atm assumes readbacks is on, so with readbacks off some stuff might break a little until I address it). Linear image readback option is not affected here and should only be enabled if its needed. Make sure to _only_ compare to main build, don't just report bugs that you see, that is not helpful, you need to report bugs that this PR causes on its own.

StevenMiller123:
No improvements or regressions in my titles, with a small performance improvement across the board in titles I tested.
This was tested before the most recent update with the swap to bitcasts, not sure if that impacted performance in any meaningful way.

Main:
image
image

PR:
image
image

[CUSA02320 PR.log](https://github.com/user-attachments/files/21693040/CUSA02320.PR.log)
[CUSA02320.log](https://github.com/user-attachments/files/21693041/CUSA02320.log)
[CUSA00663.zip](https://github.com/user-attachments/files/21693043/CUSA00663.zip)
[CUSA00663 PR.zip](https://github.com/user-attachments/files/21693044/CUSA00663.PR.zip)

Randomuser8219:
image
image
Driveclub now works without readbacks.

coolllman:
Main
Screenshot_20250809_135240_Moonlight
Screenshot_20250809_141036_Moonlight

Pr
Screenshot_20250809_141913_Moonlight
Screenshot_20250809_140844_Moonlight

Driveclub , TLG fps boost with readbacks)
Spec: i7 11700, rtx 3060

raphaelthegreat:
By how much?

rafael-57:
@raphaelthegreat from the screenshots he posted it looks like 30FPS in main, 42FPS with your PR, in driveclub. Not sure if that's valid because one is during the day and another during the night

Anyway, here's my experience. At least on my PC, readbacks still seem CPU bottlenecked, even with a 5800X3D. RAM is at 3200Mhz.

I see CPU usage ranging from 80 to 98% on cores and jumping between cores.
GPU usage is never close to full usage even with Bloodborne at 4K. Unlike main without readbacks which easily stresses 98% of the GPU.

image
image
image

(ignore the dirty tag, I just *2 memory to run 4K)

Performance seems very very slightly improved (I get 14-16fps in this area on main, 15-18 with this PR).

If you have time, I am available to run tracy on this with your guidance :-)

raphaelthegreat:
Running the game at 4K significantly increases the cost of readbacks so its not too surprising. Do you get a larger boost at 1080p?

rafael-57:
It's definetely more significant!

1080p main:
image

1080p PR:
image

image

Not strictly relevant to this PR but looking back at the empty bridge I even get 37-41fps, 4k just stayed the same instead. I didn't realize the resolution would increase the redback cost

coolllman:
Uncharted 3
Main

https://github.com/user-attachments/assets/ccaf378a-1ce0-4a24-bed3-594c0655e84a

Pr

https://github.com/user-attachments/assets/f1ad6b57-7212-47f8-ac97-47d42e0fc7f8

Gray character models in last pr, before 34892a2 same as main.

raphaelthegreat:
I dont see the gray character models
EDIT: Actually yes I can see them

coolllman:
My settings: Dma=false, readbacks=true
Screenshot_20250809_155120_Samsung Internet

raphaelthegreat:
With Dma=true, readbacks=false how does it look? Naughty Dog games technically dont abide by the fence promise because the flush is for SRT buffer which should work with DMA

coolllman:
With dma=true i don't go in the menu

raphaelthegreat:
I suspect this game might be incompatible with fence detection because it lies about the sizes of storage buffers (which causes the clamp size spam), causing large areas to be marked as GPU modified and emulator thinking it can't write yet. I might be able to think of something but strict readbacks are the only sure way to ensure proper sync

yaya54840:
Hi
With this pr, Drive Club works for me at 15 fps without any bugs.
I'm attaching my log if it helps.
[shad_log.zip](https://github.com/user-attachments/files/21697822/shad_log.zip)

raphaelthegreat:
@coolllman If you add ```fenceDetection = 0``` under [GPU] section in config.toml with newest commit does it fix the issue in Uncharted

raphaelthegreat:
PS: I'm open to suggestions about what the setting should be called, that name might be confusing to users but I couldn't think of anything else. Basically a setting about how strict readbacks are (0 is most strict like on main, 1 is with the fence detection so less strict)

coolllman:
No, fenceDetection = 0 or 1 not fix issue

raphaelthegreat:
Can you check which change in https://github.com/shadps4-emu/shadPS4/pull/3404/commits/34892a2212e29a74ae11007eea11ec90709dc891 broke the game by compiling it locally

coolllman:
I trying compiling, but no luck(

StevenMiller123:
I can take a look

StevenMiller123:
Seems like this is tied to the fenceDetection setting. With fenceDetection = 0 I cannot reproduce the grey characters issue.

raphaelthegreat:
So it does fix the game after all? Because the reply made me think game was still universally broken
> No, fenceDetection = 0 or 1 not fix issue

StevenMiller123:
fenceDetection = 0 does fix the game on my end. With it set to 1 I get the same results shown by coolllman above, while setting it to 0 fixes the issue.
image
[CUSA02320.log](https://github.com/user-attachments/files/21699869/CUSA02320.log)

Side note, it might be a good idea to log this setting.

raphaelthegreat:
The setting is not final, still thinking on how implement it in a way that is more clear in what it does

diogoaduartereis:
PT crashes for me with critical error 
[Debug] <Critical> fence_detector.h:42 DetectFences: Unreachable code!
Unimplemented PM4 type 0, base reg: 0, size: 1

Driveclub I noticed a minimum increase of 20% and max of 35% in performance. Depending on the race. Tried time trial in the first course of each map. Which totals to 6 races.

Bloodborne increased from 21FPS average to 25FPS this at 1080p when running the same course through yharnam.

In TLG I noticed a very nice increase in performance except for when you finally reach outside which for some moments it seems the performance is worse by 1 or 2 FPS.

coolllman:
Confirm, fenceDetection = 0 fix the issue, renew config file and its work, my mistake(

raphaelthegreat:
A suggestion for refactoring the readback settings, to make the difference between settings more clear, could be:

readbacks -> readbackBuffers
readbackLinearImages -> readbackImages

As far as this PR is concerned maybe its better to incorporate fenceDetection setting into a more general readbackBufferAccuracy setting that has 3 values, "extreme", "high", "low" as its easier to understand that way

- "extreme" makes buffer readbacks work like on main, ensure accuracy in all cases at the cost of performance. It can be useful for testing purposes or cases where readbacks are relied upon in unconventional ways (i.e Naughty Dog games that actually need DMA but happen to be fixed by readbacks too)

- "high" is readbacks with fence detection and preemptive downloads this PR implements. Is applicable in cases where guest wants to access GPU data from host, like climbing in TLG, exposure in Driveclub, washing in Powerwash Simulator. Performance should be higher than "extreme" as shown with testing here

- "low" would be same as high, but fence detection is more relaxed and it only triggers on page writes, so no read protection. This is useful when guest doesn't want to access GPU data ever but readbacks are needed to fix data corruption (i.e vertex explosions). Performance is expected to be higher than "high" (possible, but haven't implemented it)

gandalfthewhite19890404:
> A suggestion for refactoring the readback settings, to make the difference between settings more clear, could be:
> 
> readbacks -> readbackBuffers readbackLinearImages -> readbackImages
> 
> As far as this PR is concerned maybe its better to incorporate fenceDetection setting into a more general readbackBufferAccuracy setting that has 3 values, "extreme", "high", "low" as its easier to understand that way
> 
> * "extreme" makes buffer readbacks work like on main, ensure accuracy in all cases at the cost of performance. It can be useful for testing purposes or cases where readbacks are relied upon in unconventional ways (i.e Naughty Dog games that actually need DMA but happen to be fixed by readbacks too)
> * "high" is readbacks with fence detection and preemptive downloads this PR implements. Is applicable in cases where guest wants to access GPU data from host, like climbing in TLG, exposure in Driveclub, washing in Powerwash Simulator. Performance should be higher than "extreme" as shown with testing here
> * "low" would be same as high, but fence detection is more relaxed and it only triggers on page writes, so no read protection. This is useful when guest doesn't want to access GPU data ever but readbacks are needed to fix data corruption (i.e vertex explosions). Performance is expected to be higher than "high" (possible, but haven't implemented it)

can you post that info in configuration file as commented lines

kalaposfos13:
@raphaelthegreat it sounds okay to me, however due to way the settings are handled currently, it'd mean that people who update from an older version would get both the old names and the new names in their config which I assume would lead to some confusion

raphaelthegreat:
> can you post that info in configuration file as commented lines

It can be added yes

> people who update from an older version would get both the old names and the new names in their config

There already exist a config file version check, couldn't the old entries be removed automatically?

StevenMiller123:
I could probably add that, but from what I've seen, some people are against removing old settings.
I think it'd be better as some sort of button in the GUI, that people can choose to press if they want to clean up their config file.

raphaelthegreat:
Maybe it would be better to not rename anything at all and just name new setting as readbackAccuracy

raphaelthegreat:
Another proposal is to change the readback setting but incorporate the accuracy into the main one i.e instead of having enable + accuracy setting have a single setting with "none", "low", "high",  "extreme"

GHU7924:
>I think it'd be better as some sort of button in the GUI, that people can choose to press if they want to clean up their config file.

Maybe we should introduce special profiles for games? Something like right-clicking on a game - game profile. A small window opens where you need to check the box - use the game profile, and below there would be all sorts of settings that can be activated and deactivated. (You can even somehow combine this with patches and cheats).

That is, to separate global settings and optional ones.

I believe that in order for the emulator to develop further, it is necessary to update the settings, because new settings will be added, and some may be removed because the implementation has changed (For example, kalaposfos13 first made the setting in PR 2550 via config.toml, but then found another way. The setting remained written in the file, although it is no longer used).

kalaposfos13:
@GHU7924 #3376 is what you're looking for, except the GUI part

GHU7924:
@kalaposfos13 😮 Oh my God, I somehow missed this PR...

It's even better this way, I think it's worth trying to somehow combine them so that the settings from this PR can be written into your PR. That is, there is a global config.toml file, and then there will be an optional Default.toml file with the rest of the settings, but which can be replaced with a special file from your PR.

You can even partially use your code from your GUI, and also add help where the options will be described and what they do.

4tl4sh:
> Maybe it would be better to not rename anything at all and just name new setting as readbackAccuracy

From a user experience point of view, I think that if it's going to make something easier to understand it may worth a shot. Specially in the context of an evergrowing emulator. A year from now there probably will be more settings and many selections for those settings. And if settings labels are not as clear as they can be, and they accumulate, things get worse over time.

I don't understand the reason for not to remove old and not currently used settings. If they have no use, then I think it's better they get removed. We can use a UI checkbox that debug users can disable manually, active by default (Remove deprecated settings from config file).

rainmakerv3:
> > Maybe it would be better to not rename anything at all and just name new setting as readbackAccuracy
> 
> From a user experience point of view, I think that if it's going to make something easier to understand it may worth a shot. Specially in the context of an evergrowing emulator. A year from now there probably will be more settings and many selections for those settings. And if settings labels are not as clear as they can be, and they accumulate, things get worse over time.
> 
> I don't understand the reason for not to remove old and not currently used settings. If they have no use, then I think it's better they get removed. We can use a UI checkbox that debug users can disable manually, active by default (Remove deprecated settings from config file).

For testing forks/PRs (particularly where new config entries are added) or going back to older versions, there are definitely some reasons why retaining currently unused settings would be useful. For similar reasons RPCS3 only deletes old config entries with user confirmation as well.

I also would assume most users don't really look at the config file itself and just interact with the settings GUI, in which case there isn't really that much benefit to cleaning up the config file itself.

I think it should just be a button to cleanup old entries instead of it being automatically enabled and always active.

More on topic, I personally like this:

Another proposal is to change the readback setting but incorporate the accuracy into the main one i.e instead of having enable + accuracy setting have a single setting with "none", "low", "high", "extreme"

rainmakerv3:
From very brief testing with Config::ReadbackAccuracy::Low (0). it seems it does give back the normal performance for Bloodborne while still eliminating vertex explosions, but with a lot of additional stutter during moments when a lot of data loading is presumably occuring (initial load-in, warps, traversing certain specific in the map that I assume trigger loading certain assets).

image

rafael-57:
> From very brief testing with Config::ReadbackAccuracy::Low (0). it seems it does give back the normal performance for Bloodborne while still eliminating vertex explosions, but with a lot of additional stutter during moments when a lot of data loading is presumably occuring (initial load-in, warps, traversing certain specific in the map that I assume trigger loading certain assets).
> 
>

image

image


Accuracy 0 still has vertex explosions sadly.

rafael-57:
[shad_log.txt](https://github.com/user-attachments/files/21710458/shad_log.txt)
I've also had a crash, not sure if it's caused by this PR as I haven't been playing much for a while

[Lib.Http] <Error> http.cpp:847 sceHttpWaitRequest: (STUBBED) called
[Debug] <Critical> exception.cpp:156 sceKernelDebugRaiseException: Unreachable code!
error -0x5ffdfff6

raphaelthegreat:
It's possible vertex explosions in BB rely on memory being read too hmm. Are there any cases where low setting could be useful then, cause I can't think of any

rainmakerv3:
> > From very brief testing with Config::ReadbackAccuracy::Low (0). it seems it does give back the normal performance for Bloodborne while still eliminating vertex explosions, but with a lot of additional stutter during moments when a lot of data loading is presumably occuring (initial load-in, warps, traversing certain specific in the map that I assume trigger loading certain assets).
> 

> Accuracy 0 still has vertex explosions sadly.

Just to make sure, are you sure you're on the right commit? For reference here's my video showing the working facegen and no vertex explosions with low accuracy. On main, dying in this location with 3 NPCs produces vertex explosions more than 90% of the time:

https://www.youtube.com/watch?v=73SF0vqZ_bU&ab_channel=rainmakerv2

Also, from my own testing (and how I have it set-up locally), simply removing that call to UpdateProtection after if Config::readbacks without any other changes does indeed fix vertex explosions without performance impact (apart from all the additional stutter).

image

So I think unless there's anything else here that breaks readbacks for bloodborne, disabling that particular call with a setting should produce this result.

rafael-57:
## PR Testing 4K
Yeah accuracy 0 definetely is not enough sadly. It aso gives me these weird artifacts after trying for a while in the character creator. Never seen it before personally:
https://github.com/user-attachments/assets/bf37fbc4-3774-4ffc-991b-ddc16da6dd33

Not sure if it's videly known, but vertex explosions in BB can already happen even before loading into the game. The character portrait is not a simple PNG loaded from savedata. Just by going into the load screen the characters are already rendered, and they can bug out.

These portraits can still bug out even with accuracy 1:
image

Trying to load them also caused multiple crashes with device lost, not sure if that's the case with everyone else. This happens to me in main too with readbacks on. This instance in particular was with accuracy set to 1
`[Kernel.Fs] <Info> file_system.cpp:273 close: Closing /app0/dvdroot_ps4/parts/hr_a_0109.partsbnd.dcx
[Kernel.Fs] <Info> file_system.cpp:273 close: Closing /app0/dvdroot_ps4/parts/bd_f_8000.partsbnd.dcx
[Lib.PlayGo] <Info> playgo.cpp:317 scePlayGoSetInstallSpeed: called
[Debug] <Critical> vk_scheduler.cpp:166 operator(): Assertion Failed!
Device lost during submit`

## Main Testing 4K
Crashes while loading portraits with readbacks on also happens in main, I just double-checked. When they do load they can have some weird artifacts during loading (doesn't happen on PS4 ofc) but at the end they load just fine.

They can also just bug out and stay bugged even in main with readbacks on. 
image

image

Readbacks linear images ON or OFF doesn't seem to make any difference for this.

In gameplay though, I've never seen any vertex explosion with accuracy 1 or main readbacks, and that still is the case.

rafael-57:
> > > From very brief testing with Config::ReadbackAccuracy::Low (0). it seems it does give back the normal performance for Bloodborne while still eliminating vertex explosions, but with a lot of additional stutter during moments when a lot of data loading is presumably occuring (initial load-in, warps, traversing certain specific in the map that I assume trigger loading certain assets).
> > 
> > 
> > image image
> > Accuracy 0 still has vertex explosions sadly.
> 
> Just to make sure, are you sure you're on the right commit? For reference here's my video showing the working facegen and no vertex explosions with low accuracy. On main, dying in this location with 3 NPCs produces vertex explosions more than 90% of the time:
> 
> https://www.youtube.com/watch?v=73SF0vqZ_bU&ab_channel=rainmakerv2
> 
> Also, from my own testing (and how I have it set-up locally), simply removing that call to UpdateProtection after if Config::readbacks without any other changes does indeed fix vertex explosions without performance impact (apart from all the additional stutter).
> 
> image
> So I think unless there's anything else that breaks readbacks for bloodborne, disabling that particular call with a setting should produce this result.

Yeah I had just pulled and built latest from turtle's branch when testing:
image

I'm also 100% sure I was on latest commit when making that comment because the config file update on its own and added the readbacksAccuracy setting, which I set to 0. I also screenshotted the config file just to be safe.

To add another proof, I was playing 4k and got 34fps so it definetely was with accuracy 0. Accuracy 1 gives me 15-17fps in that area. I should say all of this testing was with 4k btw

Loading cathedral ward just crashes the emu instantly now. I'm suspecting GPU memory/RAM may not get cleared completely on closing the emu. I'm going to try a PC restart

rainmakerv3:
Maybe try stock res first. All I can verify that it is working on my set-up with only the 60fps patch (which I enabled only to prove no performance loss)

Missake212:
Gave it a try with readbackAccuracy set to 0 and like @rainmakerv3 I don't get any vertex explosion while also having good performance.

image


[CUSA03173.log](https://github.com/user-attachments/files/21711312/CUSA03173.log)

rafael-57:
> Maybe try stock res first. All I can verify that it is working on my set-up with only the 60fps patch (which I enabled only to prove no performance loss)

Indeed for me the explosions happened only once while loading cainhurst castle, 4K, PR, readbacks accuracy 0. Haven't managed to replicate that. I did also get artifacts in the character creation screen though, which don't happen with accuracy 1.

I have now tried 1080p readbacks PR accuracy 0 in Cathedral Ward. Like you said, I haven't had any vertex explosions as well for a while now. 

Switching to main with no readbacks in that same area and vertex explosions are very easy to replicate indeed.

Character portraits in the load. screen can still fail to load but I guess that's not 100% fixed by readbacks so not in scope of this PR anyway. 

I'd say it looks like accuracy 0 removes explosions 99% of the time without hindering performance much so it's definetely worth keeping! Looks like a huge improvement. Makes the game a lot more playable

As for feedback for the setting, I would also make it all in one setting. Like a dropdown menu titled Readbacks from which you can select:
None, Low Accuracy, High Accuracy, Extreme Accuracy

image

Not sure how to handle readback linear images though. Maybe just keep it a separate checkmark

4tl4sh:
I would like to add something that may or not may be useful.

In The Last Guardian, readbacks seem to put more pressure on memory causing not only worse performance in general but big drops while loading new areas. This PR improved general performance but did not alleviate those dips. This is in line with some previous comments regarding traversal stutter on BB.

In addition, this PR with fence detection ON along with vblank = 4 helps minimize lightning problems. I don't know if this is causation or just correlation. I don't know if a higher vblank gives readbacks more cycles to do whatever it does. I tried several combinations many times and these results were somewhat consistent:

- Readbacks OFF Linear Images OFF Vblank 4, GLITCH
- Readbacks OFF Linear Images ON, Vblank 4, GLITCH
- Readbacks ON Linear Images OFF Vblank 4, GLITCH
- Readbacks ON Linear Images ON VBlank 4, NO glitch
- Any combination from above with VBlank <4, GLITCH (there's a certain amount of randomness here as sometimes it loads properly NO glitch)

**This is GLITCH**

IMG20250810125704

**This is NO glitch**

IMG20250810133535

I tested yesterday's version so there was no accuracy setting yet.

raphaelthegreat:
@4tl4sh Linear images is necessary for proper lighting so testing with it off doesnt have much meaning. I dont think I've heard of these lighting bugs before, vblank divider only really makes game present faster, find it strange the game would have totally broken lighting except under one very specific and rarely used configuration

4tl4sh:
@raphaelthegreat I was trying to isolate the problem, to see if it was related to readbacks and to what setting specifically. Finding that Vblank thing was a coincidence. As I don't know the deep technical specifics I just threw a batch of combinations. As I'm not very knowledgeable on how things should work I'm not very biased and my results are as naive as possible. This fresh view sometimes can be helpful as a hint hence I try not to comment about anything if I can't find a repeatable pattern. I hope this isn't much trouble.

bigol83:
I can confirm that with this PR there are heavy stutters with readbacks enabled. I am testing with Bloodborne just by using Hunter's Mark, soon after loading you can see heavy stutters (very noticeable with 60 fps patch, framerate reaches as low as 23 fps during these stutters). The stutter is not there even with this PR when reabacks are disabled and it's not there on main build, even if performance is much worse in general with readbacks enabled.

I also tested readback accuracy at 0 and i can't see any vertex explosion or graphics issues, tried teleporting many times, that is what easily triggers vertex explosions for me. 

I am on nvidia rtx 3060 gpu.

raphaelthegreat:
Does the stutter exist with readbackAccuracy = 2 ?

coolllman:
With last pr changes (readbackAccuracy)  Infamous Second son don't go in game with readbacks=false, in main go in game
[shad_log.txt](https://github.com/user-attachments/files/21713143/shad_log.txt)

bigol83:
> Does the stutter exist with readbackAccuracy = 2 ?

With readback accuracy to 2 it behaves just like on main, doesn't seem to be a noticeable stutter, but the game freezes during loading, maybe because performance is too low. 

With readback accuracy set to 1 there is stutter too.

raphaelthegreat:
@rafael-57 (or anyone else familiar with tracy) can you profile the stutters that occur with the newer accuracy?

raphaelthegreat:
@coolllman Please enable validation in config and send new log

bigol83:
I played Bloodborne a bit more with readback accuracy at 0 and unfortunately vertex explosion is still there even if it's much less frequent.

Missake212:
After testing for a while I managed to find a big problem, not only did I get a vertex explosion but the colors were all weird (testing methodology was simply going to the hunter's dream and to cathedral ward back and forth hoping to see a vertex explosion, on one of those warps the colors looked like this).

image

image


[CUSA03173.log](https://github.com/user-attachments/files/21715465/CUSA03173.log)

raphaelthegreat:
Does this happen only with low accuracy or also with high?

Missake212:
I will try with high and get back to you.

Missake212:
Looks like it does also happen with high.

image

[CUSA03173.log](https://github.com/user-attachments/files/21716258/CUSA03173.log)

StevenMiller123:
People were reporting a similar color bug with LNDF's #3396 as well, maybe this is actually a regression in main?

Missake212:
> People were reporting a similar color bug with LNDF's #3396 as well, maybe this is actually a regression in main?

I tried to replicate it on main with Readbacks enabled but I nearly always hang in loading screen when going back and forth between the dream and cathedral wards (I also got the softlock a few times with high accuracy on this PR, didn't get any with low). Also for LNDF's PR it seems like things go black, here colors are just wrong so I'd say this is a bit different. 

I also got vertex explosions (not sure if you can call them that since they didn't happen on NPCs this time) with High accuracy so not sure what's wrong there.

Screenshot 2025-08-11 153443

Screenshot 2025-08-11 153457

But also next warp after that vertex explosion that wasn't created by an NPC colors were bugged out so perhaps it really is a regression from the GC PR I can't tell.

Screenshot 2025-08-11 153649


Screenshot 2025-08-11 153634

Screenshot 2025-08-11 153641

raphaelthegreat:
The GC PR didn't use the Dirty flag in SafeToDownload, so small (32x32 BC7 images or less) images that are marked with MaybeCpuDirty could be downloaded incorrectly and cause corruption, but I'm not sure

Missake212:
I've been testing main for the past 15 minutes without readbacks and colors never changed, so that leaves us with a few possibilities:

-Regression from this PR.

-~~Regression with Readbacks in general (less likely since they've been merged for a while and someone probably would've reported it, but since i'm getting so many hangs in loading screens perhaps no one has tested).~~

-Regression from the GC PR but only when Readbacks are enabled.

Conclusion seems to be that it only happens with readbacks enabled. It would be nice to have more people testing this (and ideally without changing the emulator's code or using game patches)

Useless log:

[CUSA03173.log](https://github.com/user-attachments/files/21717338/CUSA03173.log)

raphaelthegreat:
Probably best to try a build right before GC got merged or build locally and comment out the RunGarbageCollector calls [here](https://github.com/shadps4-emu/shadPS4/blob/1b621e4b1d6b473d3fdc9f02e96912e483a399d6/src/video_core/renderer_vulkan/vk_rasterizer.cpp#L455)

Missake212:
Tried a build prior to GC merge and didn't have the color issue, I don't have VS set up so if someone else could test commenting out what Turtle said above to see if GC is indeed the issue it would help me out.

[CUSA03173.log](https://github.com/user-attachments/files/21717749/CUSA03173.log)

GHU7924:
I tested the latest Main build.

Test without Readbacks.



With Readbacks.




These are the problems that I personally saw. Apparently there are regressions in Main itself.

coolllman:
Infamous
[shad_log.txt](https://github.com/user-attachments/files/21718374/shad_log.txt)

Update: fix last pr, thank you

rafael-57:
> @rafael-57 (or anyone else familiar with tracy) can you profile the stutters that occur with the newer accuracy?

yes! But do I need to log something in particular or do I just run tracy with everything set to default? Also do I pull fix readbacks off or nah?

rafael-57:
image

image

image

https://github.com/user-attachments/assets/5cb7bcfa-794c-4376-8122-93bccec78e6c

After warping at least 100 times with readbacks on and garbage collector ON/OFF, I can say I'm seeing some weird stuff with readbacks on even with accuracy 1 and the garbage collector commented.

Not just character faces but models in the level like tombstones and even weapons (see video) are exploding now, even with readbacks 1 which I don't remember seeing before

This is without 4k patches and stock memory. I should say, this branch is very unstable and very prone to crashing when opening the loading screen/warping to an area 

EDIT: I'm pretty sure it's not the garbage collector now but something is just off with this PR now

I did not manage to replicate the extreme color issues the others got, all I got were small corruptions even with the GC off

image

rafael-57:
image
Readbacks-POC branch
Readbacks accuracy 1
latest commit reverted
Garbace Collector commented

Still getting garbage data. Not sure what is going on anymore...

In any case, it looks like every time a single particular asset has a corrupt texture/explodes. Sometimes it's the a particular wood plank texture, sometimes it's the sword exploding, sometimes it's every single column in the area. I'm guessing assets just get corrupted and by chance it happens only on 1 asset at a time per my testing

raphaelthegreat:
Can you keep reverting commits and test a bit to see if something broke it

Missake212:
I think I got an answer to this, this is an old screenshot I took when testing readbacks a while ago, I think the issue has always been there but it barely happened, and now changing the readbackAccuracy actually exposes it more, I could be wrong but I think this is it.

image

rafael-57:
> Can you keep reverting commits and test a bit to see if something broke it

Well before this I just tested this morning right before "fix readbacks off" and everything was fine, and then I've also reverted it.

I've just got this in main too, with readbacks on.:
image

image

I've done a lot of testing with main and readbacks off and I'm not finding any corrupt textures after many warps. 

I'm beginning to believe we just tested a lot today and started exposing some inherent issues that were already present with readbacks.

Is readbacks 2 more accurate than current main readbacks? I could test that too

rafael-57:
My main takeways after losing my sanity testing for 2 hours:
1. Garbace colector 1 isn't broken
2. Readbacks aren't perfect, even in main (prone to hang/crash the emu, can still have some corrupt textures and random vertex explosions, but they're rare)
3. Readbacks to 0 is still a great improvement for BB, greatly reducing the occurence of vertex explosions and with a much lesser performance hit

raphaelthegreat:
Readbacks 2 is the same as on main. The garbage textures seems like it could happen because texture cache relies on a hashing workaround for that, but readbacks on main having vertex explosions sounds weird to me, does it actually happen?

rafael-57:
> Readbacks 2 is the same as on main. The garbage textures seems like it could happen because texture cache relies on a hashing workaround for that, but readbacks on main having vertex explosions sounds weird to me, does it actually happen?

Ok, I was testing with readback accuracy 1 instead of 2. So that could be the cause for vertex explosions on objects (not faces unlike main with readbacks OFF).

I do remember making this comment in the original readbacks PR though:
https://github.com/shadps4-emu/shadPS4/pull/2668#issuecomment-2746249849

After finnicking a lot I just got this even in main, whatever it is:
https://github.com/user-attachments/assets/8abced79-ef88-493c-9316-e6296144faef

Anyway I wouldn't go too out of scope with this PR. Existing readbacks are already in main with all their pros and cons, and this improves performance a bit + adds more granular settings.

Would be nice if readbacks were more stable in general though. Would it help if I collect a stack with visualstudio when crashing?

What about tracy? Do I just run it with stock code + readbacks accuracy set to 0?

raphaelthegreat:
> Would be nice if readbacks were more stable in general though. Would it help if I collect a stack with visualstudio when crashing?

Yes

> What about tracy? Do I just run it with stock code + readbacks accuracy set to 0?

Yes

I suspect the reason of the stutter is the game overwriting some large GPU modified region from CPU which would cause page-by-page flush. I don't like that this PR has these bugs though so I was thinking of shelving it until its fixed

rafael-57:
> > Would be nice if readbacks were more stable in general though. Would it help if I collect a stack with visualstudio when crashing?
> 
> Yes
> 
> > What about tracy? Do I just run it with stock code + readbacks accuracy set to 0?
> 
> Yes
> 
> I suspect the reason of the stutter is the game overwriting some large GPU modified region from CPU which would cause page-by-page flush. I don't like that this PR has these bugs though so I was thinking of shelving it until its fixed

Tracy, started right during warp:
https://drive.google.com/file/d/1Ss5ZT_cMSdsfidoXC0eOThcePE4sh29_/view?usp=sharing

bigol83:
So testing more with Bloodborne, readback accuracy 0 is the only one that doesn't freeze the game but has big stutters. Best performance by far.

Readback accuracy 1 has better performance compared to main build by 5-10 fps, but it freezes the game

Readback accuracy 2 performance is the same as main build with readbacks enabled and it also freezes the game

The game freezes happen with readback enabled on main too.

fabiossussin (approved):

Arsconyl:
Can we have a clue on when will this be merged ?

raphaelthegreat:
This needs rework to make flushes thread safe, either with host reads from rebar memory or put them in dedicated transfer queues and directly DMA to host memory without staging

DarkBoy08:
@raphaelthegreat Hi Rafael. I'd like to bring up a topic, which was started a year ago. I'd like you to consider rebuilding this PR campaign for the purpose of testing some exclusives in the new emulator version? Would you allow me? I'm in advance for my poor English. I'm from Russia.

PR #3405, opened 2025-08-09
libSceAppContent: Determine entitlement labels from additional content param.sfo

StevenMiller123:
The approach I took to deriving the entitlement label for additional content files in #3375 had two limitations: it relied on entitlement ids being 16 characters long, and it relied on the end of the folder name containing the entitlement. If the former wasn't the case, my code would throw an exception, while the latter would cause the DLC to not detect.

To resolve both issues, I've created adjusted the code to search for entitlement labels from the param.sfo of the additional content instead, which should behave consistently regardless of dumping method.

I've also fixed an additional issue with sceAppContentAddcontMount, where the addcont directory wasn't mounted correctly if the folder name wasn't just the entitlement label.

This should fix any regressions seen by people who installed DLC with smaller entitlement labels prior to #3375


squidbus:
Just to check - have you considered if we should just get the entitlement ID from the DLC `param.sfo` to make it more reliable?

StevenMiller123:
I hadn't considered it too strongly, though I'll look into doing that since it sounds like a more reliable solution.
I've also noticed that `sceAppContentAddcontMount` needs updating too, so I'm marking this PR as a draft while I address these issues.

StevenMiller123:
I've updated sceAppContentInitialize to use DLC param.sfos to determine the entitlement label, and updated sceAppContentAddcontMount to search for the proper directory to mount.
I added thorough logging for errors when loading additional content in sceAppContentInitialize too, so if there are any issues, they should be visible within logs now.

Only problem I can forsee is that ftpdump doesn't seem to dump the param.sfo for DLC properly, but in terms of real hardware accuracy, I doubt a PS4 would accept DLC without a param.sfo anyway.

squidbus (approved):

PR #3414, opened 2025-08-12
Core: Add entries for . and .. in MntPoints::IterateDirectory

StevenMiller123:
On a real PS4, all directories will start with an directory entry for "." and "..". Emulating this behavior is necessary for some games, since devs can safely assume there will always be at least one entry in a given directory, and try to read said entry regardless of what we return.

Emulating these two entries fixes an exception in Grand Theft Auto V (CUSA00419) that occurs if the game's /download0 mount is empty.

For context, here's the output on real hardware, from a homebrew I wrote for testing this:
image

Here's what the current shadPS4 main build shows:
image

And here's what my PR shows:
image

### DIFF
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -161,6 +161,10 @@ void MntPoints::IterateDirectory(std::string_view guest_directory,
     // Only need to consider patch path if it exists and does not resolve to the same as base.
     const auto apply_patch = base_path != patch_path && std::filesystem::exists(patch_path);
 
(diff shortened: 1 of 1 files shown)


StevenMiller123:
With this PR, an unmodified copy of Grand Theft Auto V can now reach the loading screen without any workarounds.
Now the game crashes on:
```
Unknown data_format=45 and num_format=4
```
image
image

[CUSA00419.log](https://github.com/user-attachments/files/21745069/CUSA00419.log)

PR #3424, opened 2025-08-16
Config: toggle background controller inputs

### LINKED ISSUE #3390: [Feature Request]: Toggleable option to recieve controller input while ShadPS4 isn't focused
### Checklist

### Description

Like some other emulators, if you have a controller (or a form of input device that isn't a keyboard/mouse), there is an option to still receive player inputs in the background while the emulator isn't on focus.
(ie being able to play a game completely while on other windows)

Not needed, but would definitely be welcomed, especially by some multimonitor/widescreen folk

(Couldn't find an option in the settings)

### Reason

[Title, while it wouldn't be insanely useful, its main purpose is essentially only for convenience]
Likely, will only save you a few seconds as you could just Alt+Tab or focus the window, although its main purpose also may be useful for multitasking

### Examples

_No response_

rainmakerv3:
closes https://github.com/shadps4-emu/shadPS4/issues/3390

Gives the option to allow the emulator to process controller inputs even when the game window is not in focus.

Can be good for multi-monitor set-ups and the like.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -58,6 +58,7 @@ static bool isMotionControlsEnabled = true;
 static bool useUnifiedInputConfig = true;
 static std::string micDevice = "Default Device";
 static std::string defaultControllerID = "";
+static bool backgroundControllerInput = false;
 
 // These two entries aren't stored in the config
 static bool overrideControllerColor = false;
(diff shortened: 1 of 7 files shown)


tochnonechelovek:
Works as intended and doesn't regress anything in my titles. Really useful for two monitors.

PR #3426, opened 2025-08-17
Improve heuristic for attributes passed via ring buffers.

baggins183:
Previously a buffer load in a vertex shader could be treated like a ring access, dropping offen vgpr and possibly asserting during resource tracking because of mismatch between types (u32x2 vs U32), caused by inconsistencies in flags (index_enable and offset_enable)


tochnonechelovek:
No changes and no regressions in my titles.

PR #3431, opened 2025-08-18
Make shadps4 default repo for patches

Missake212:
Title. One issue that's left is that once downloaded, the GoldHEN patches are still above in the patch list, not sure if there's a part in the code I missed or if it's just alphabetical order, if anyone knows how to change that I would appreciate it but I don't think it's a big deal personally.

image

### DIFF
--- a/src/qt_gui/cheats_patches.cpp
+++ b/src/qt_gui/cheats_patches.cpp
@@ -643,6 +643,14 @@ void CheatsPatches::populateFileListPatches() {
         }
     }
     QStringListModel* model = new QStringListModel(matchingFiles, this);
+    if (!matchingFiles.isEmpty()) {
+        QModelIndexList matches = model->match(model->index(0, 0), Qt::DisplayRole, "shadPS4", 1,
+                                               Qt::MatchContains | Qt::MatchCaseSensitive);
+        if (!matches.empty()) {
+            QModelIndex shadPS4Index = matches.first();
+            model->moveRow(QModelIndex(), shadPS4Index.row(), QModelIndex(), 0);
+        }
+    }
     patchesListView->setModel(model);
 
     connect(
--- a/src/qt_gui/cheats_patches.cpp
+++ b/src/qt_gui/cheats_patches.cpp
@@ -614,6 +614,7 @@ void CheatsPatches::populateFileListPatches() {
 
     QStringList folders = dir.entryList(QDir::Dirs | QDir::NoDotAndDotDot);
     QStringList matchingFiles;
+    QString shadPS4entry = "";
 
     foreach (const QString& folder, folders) {
         QString folderPath = dir.filePath(folder);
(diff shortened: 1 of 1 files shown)


kalaposfos13:
If I'm correct, the thing that needs modification is `CheatsPatches::populateFileListPatches` to get shadPS4 as the first entry in the "Select Patch File" list

Missake212:
I can't lie I don't understand this part of the code very well.

DanielSvoboda:
It's actually a little more complicated
in populateFileListPatches you would have to identify if 'folder' is equal to shadPS4 and select it.

Missake212:
Yeah then that's definitely not feasible for someone like me, I'll let someone else handle this if anyone's interested.

rainmakerv3:
I think you can do something like this to select shadPS4 patches from the listview automatically https://github.com/rainmakerv3/shadPS4/commit/8b9afbbe5a5ff6187d4be5b911444effe525ae9c

Missake212:
@rainmakerv3 Did you try it out? I gave it a try on my end and it doesn't really seem to work, personally I'm fine with the PR as is, as long as shadPS4 is first people would only really need to use the shadPS4 patches and won't have to download the GoldHEN ones (in the game where shad has patches).

rainmakerv3:
> @rainmakerv3 Did you try it out? I gave it a try on my end and it doesn't really seem to work, personally I'm fine with the PR as is, as long as shadPS4 is first people would only really need to use the shadPS4 patches and won't have to download the GoldHEN ones (in the game where shad has patches).

works ok for me, it doesn't put the shadps4 entry above goldhen one in the list, but it does select it by default

Missake212:
Idk then it really just doesn't work for me.

image

rainmakerv3:
> Idk then it really just doesn't work for me.
> 


image

To clarify, this is the part that should be automatically selected by the code. Did you have to manually select the shadPS4 file here?

Missake212:
Oh I think there's a misunderstanding then, I didn't even think about this but yeah then your code works, what I wanted was to put the "Bloodborne.xml | shadPS4" patch above the "Bloodborne-Orbis.xml | GoldHEN" one, that way not only does it look better but also it's selected as the first one by default.

rainmakerv3:
> Oh I think there's a misunderstanding then, I didn't even think about this but yeah then your code works, what I wanted was to put the "Bloodborne.xml | shadPS4" patch above the "Bloodborne-Orbis.xml | GoldHEN" one, that way not only does it look better but also it's selected as the first one by default.

https://github.com/rainmakerv3/shadPS4/commit/35b9b7504664f6f26749f31e76fc041032c5bedc I think should work for that (moving the row to the top instead of just selecting it)

Missake212:
That works perfectly! Do you want to open a separate PR once this is merged or open a PR on my fork maybe? 

image

rainmakerv3:
I'll open a PR to your branch in a bit

rainmakerv3:
ok PR submitted

rainmakerv3:
@Missake212 I made another PR to your branch with an adjustment to this (moving the row only to the top when the entry is exactly the same as filename | shadPS4 (instead of just whenever it contains "shadPS4") if you would like to merge it.

I just remembered that custom folders can now be created, so this makes it so the file from the repo will be on top even if there are other folders with shadPS4 in it

squidbus:
Is this ready for review?

Missake212:
> Is this ready for review?

Yes.

PR #3432, opened 2025-08-18
Remove unused code from the cheat repo

Missake212:
Remove everything tied to wolf2022 in the code since it seems the website isn't up anymore, opening as a draft because I had to delete a huge chunk of the code and I'm not sure if that's correct or not (i tested it and it still works as expected, I can download both GoldHEN and shadPS4's cheats and delete them), I had two theories for this:

1- I deleted something I shouldn't have.

or

2- The code I deleted was only used for Wolf2022's website and not for the GitHub's repos, which makes the most sense to me.


DanielSvoboda:
I tested it here and it seems fine, the other two are working.
Nothing broke.

PR #3437, opened 2025-08-20
Dropping "SDL audio queue backed up..." to INFO

Missake212:
Dropping "SDL audio queue backed up..." to INFO instead of WARNING, a lot of people were confused about these in the logs, thinking either they were the cause of audio stutters or crashes when they're not, I don't think it's useful to keep it as warning for most people.


squidbus:
> thinking either they were the cause of audio stutters or crashes when they're not

Just to be clear, while not the root cause, they are directly related to experiencing audio stutters. It’s a warning because it’s not supposed to happen and will result in audio problems if it does happen.

Missake212:
Yes I know, what I meant was that people thought the line in the log was responsible for the audio stutter and were trying to get rid of it, but it just informs you about the audio stutter.

squidbus:
Okay, I'm fine with it if you get clang-format passing. I think its just the indentation needs to be adjusted since the log macro name is shorter.

PR #3439, opened 2025-08-21
Allow vector and scalar offset in buffer address arg to LoadBuffer/StoreBuffer

baggins183:


baggins183:
BTW the reason the vector offset and scalar offsets are kept separate as different u32x3 members is that geometry shader pattern matching seems to need them separate

StevenMiller123:
This PR allows The Last of Us™ Part II (CUSA07820) to compile a few more shaders (on Linux), but it still crashes before showing anything.

Before it crashed on 
```
[Debug] <Critical> vector_memory.cpp:213 operator(): Assertion Failed!
Having both scalar and vector offsets is not supported
```
Now it's crashing on
```
[Debug] <Critical> resource_tracking_pass.cpp:402 FindSharpSources: Unreachable code!
Bindless sharp access detected pc=0x3894
```
[CUSA07820 PR.log](https://github.com/user-attachments/files/21910394/CUSA07820.PR.log)
[CUSA07820 Main.log](https://github.com/user-attachments/files/21910395/CUSA07820.Main.log)

baggins183:
yeah i broke bloodborne too

tochnonechelovek:
`[Debug] <Critical> operand_helper.cpp:10 operator(): Assertion Failed!` during loading:
[CUSA07007.log](https://github.com/user-attachments/files/21918909/CUSA07007.log) - NieR:Automata DEMO 120161128
[CUSA17925.log](https://github.com/user-attachments/files/21918918/CUSA17925.log) - RESIDENT EVIL 3 "Raccoon City Demo"
[CUSA05527.log](https://github.com/user-attachments/files/21919343/CUSA05527.log) - STEEP

also now Dauntless crashes instead of menus:
[CUSA15433.log](https://github.com/user-attachments/files/21919149/CUSA15433.log) - Dauntless

StevenMiller123:
Not seeing any regressions yet on my end, though I haven't tested thoroughly yet.
The Last of Us™ Part II is still improved too, same crash as before the newly pushed fixes.

src/shader_recompiler/ir/operand_helper.h
```diff
@@ -0,0 +1,34 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
```
raphaelthegreat:
```suggestion
// SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
```

src/shader_recompiler/ir/operand_helper.cpp
```diff
@@ -0,0 +1,26 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "operand_helper.h"
+
+namespace Shader::IR {
+
+static IR::U32 GetBufferAddressComponent(const Inst* buffer_inst, u32 comp) {
```
raphaelthegreat: These functions are very small, it would be better to define them as inline in the header

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -959,7 +963,8 @@ void PatchImageSampleArgs(IR::Block& block, IR::Inst& inst, Info& info,
             [[fallthrough]];
         case AmdGpu::ImageType::Color2DArray: // x, y, slice
             addr_reg = addr_reg + 3;
-            // Note we can use FixCubeCoords with fallthrough cases since it checks for image type.
+            // Note we can use FixCubeCoords with fallthrough cases since it checks for image
+            // type.
```
raphaelthegreat: Why is the comment 2 lines now?

src/shader_recompiler/ir/passes/hull_shader_transform.cpp
```diff
@@ -386,9 +387,16 @@ void HullShaderTransform(IR::Program& program, RuntimeInfo& runtime_info) {
...
                 const u32 num_dwords = u32(opcode) - u32(IR::Opcode::StoreBufferU32) + 1;
-                IR::U32 index = IR::U32{inst.Arg(1)};
-                ASSERT(index.IsImmediate());
-                const u32 gcn_factor_idx = (info.inst_offset.Value() + index.U32()) >> 2;
+                IR::Value voffset, soffset;
+                bool success =
+                    M_COMPOSITECONSTRUCTU32X3(MatchU32(0), MatchImm(voffset), MatchValue(soffset))
+                        .Match(inst.Arg(IR::StoreBufferArgs::Address));
+                // Some shaders seen using an uninitialized sgpr for soffset. Ignore undef soffset
+                // arg. TODO figure out if there's defined behavior or just UB
```
raphaelthegreat: That doesn't sound possible for compiler to emit, there must be some initialization SGPR missing

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
baggins183: Probably my clang-format messed it up

src/shader_recompiler/ir/passes/hull_shader_transform.cpp
baggins183: Theres some hull shaders in bloodborne, i.e. hs_0x00000000bc234799, that use
> /*0000000000d4*/ buffer_store_dword v4, v5, s[0:3], s8 offen glc

where s8 is not written beforehand in the shader.

I found num_user_regs == 8
SPI_SHADER_PGM_RSRC1_HS::SGPRS == 1 -> 16  sgprs allocated

so s8 should be uninitialized as far as I know from the docs.
I dont know if maybe num_user_regs < 16, the rest of the 16 up to s15 are 0 initialized instead of being loaded. I cant tell from any docs

baggins183: I mean in general, maybe s[0:15] are always either loaded or zero initialized since these are the registers that can be user data, depending on num_user_regs. I dont mean all allocated sgprs may be 0 initialized on hardware before execution (unlike my example where the number allocated happens to be 16)

raphaelthegreat: The user data regs come after the HS stage dependant SGPR initialization, check out the "Front End Shader (LS+HS)" part of "SGPR initialization" chapter in https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/rdna-shader-instruction-set-architecture.pdf

It's not exactly the same as GCN2 but I imagine something similar is going on, either s8 is part of that initialization and its missing or user data

raphaelthegreat: RADV also shows what the SGPR initialization sequence of HS stage generally looks like https://github.com/chaotic-cx/mesa-mirror/blob/79656dbcd3023574897f863c93be3a12171b0cf4/src/amd/vulkan/radv_shader_args.c#L721C7-L721C33

baggins183: Thanks for the help, I added new Attributes for the missing sgprs so stores to tess factors can be better disambiguated from regular BUFFER_STORE's and to catch if we ever encounter off-chip tess (itll show up if they cant be dead code eliminated)

baggins183:
This should probably be tested with any game with tess shaders

StevenMiller123:
I haven't encountered any new issues so far, though the only game I've tested that's actually ingame on Windows is Bloodborne. I'll retest PowerWash Simulator on Linux when I have a chance, since it's ingame on there.
[CUSA29983.log](https://github.com/user-attachments/files/21973381/CUSA29983.log)
[CUSA00900.log](https://github.com/user-attachments/files/21973380/CUSA00900.log)
[CUSA00803.log](https://github.com/user-attachments/files/21973379/CUSA00803.log)

tochnonechelovek:
No more regressions for my titles now. No noticeable changes also.

squidbus:
Is this pending any further testing or good to go?

baggins183:
Should be good enough, it should assert if there are future tess problems

Issue #3372, opened 2025-08-03
[APP BUG]: [Debug] <Critical> pthread.cpp:294 operator(): Assertion Failed! Failed to create thread with error -1073741790


jitart:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

<img width="1388" height="482" alt="Image" src="https://github.com/user-attachments/assets/63f04faf-3b59-44b3-8bde-41e52bb6a97e" />

i get this bug

[Debug] <Critical> pthread.cpp:294 operator(): Assertion Failed!
Failed to create thread with error -1073741790

### Reproduction Steps

tried many things and still didnt get a way to fix it 

### Specify OS Version

win 11

### CPU

i5-9600K

### GPU

rtx 2060  super

### Amount of RAM in GB

16 gb

jitart:
the game is BloodBorne CUSA03173

Randomuser8219:
This is a known issue with Windows Defender's Control Flow Guard on Windows 11 24H2, but disable this at your own risk. Open Windows Security, then go to App & browser control/Exploit protection settings and disable "Control flow guard (CFG)".

SAN4EZDREAMS:
In theory, I fixed it, but since you didn't leave **Reproduction Steps**, I have no way to check, you can [download](https://github.com/SAN4EZDREAMS/shadPS4/actions/runs/16938492764) it yourself and see if the error is gone or not.

Missake212:
@SAN4EZDREAMS not my log but I found someone able to test it in the Discord and it didn't work for them unfortunately.

[shad_log(3).txt](https://github.com/user-attachments/files/21798395/shad_log.3.txt)

SAN4EZDREAMS:
> [@SAN4EZDREAMS](https://github.com/SAN4EZDREAMS) not my log but I found someone able to test it in the Discord and it didn't work for them unfortunately.
> 
> [shad_log(3).txt](https://github.com/user-attachments/files/21798395/shad_log.3.txt)

It seems I wasn't entirely correct and my fix wasn't complete. Thank you, I have updated my fix based on your report. If you have a chance to test it, please let me know.
[Download](https://github.com/SAN4EZDREAMS/shadPS4/actions/runs/17034250553).

Missake212:
[...]

PR #3440, opened 2025-08-21
Fix for Failed to create thread -1073741790

SAN4EZDREAMS:
Potential fix to #3372
Due to lack of testing for me, it is impossible to know if the bug is fixed or not.
So if someone can reproduce this error with my fixes, I'd really appreciate an emulator report, thanks.

### DIFF
--- a/src/core/libraries/kernel/threads/stack.cpp
+++ b/src/core/libraries/kernel/threads/stack.cpp
@@ -45,21 +45,19 @@ int ThreadState::CreateStack(PthreadAttr* attr) {
      * If the stack and guard sizes are default, try to allocate a stack
      * from the default-size stack cache:
      */
-    bool from_cache = false;
     if (stacksize == ThrStackDefault && guardsize == ThrGuardDefault) {
         if (!dstackq.empty()) {
             /* Use the spare stack. */
             Stack* spare_stack = dstackq.top();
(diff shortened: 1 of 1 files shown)


red-prig:
Enable CFG as mandatory and test it, it looks dubious so far, since CFG protects against jumping into a memory area with code that the program itself provided (in this case, it is the result of emulating ELF dynamic linking) and does not depend on the stack in any way

SAN4EZDREAMS:
> Enable CFG as mandatory and test it, it looks dubious so far, since CFG protects against jumping into a memory area with code that the program itself provided (in this case, it is the result of emulating ELF dynamic linking) and does not depend on the stack in any way

You are correct about CFG's primary function protecting indirect calls.

However, the issue here is a broader Windows security policy that is active when CFG is enabled: the OS blocks the creation of any new thread if its stack memory is marked as executable. This is what causes the `STATUS_ACCESS_DENIED` (-1073741790) error.

My fix simply removes the execute permission from the stack memory, aligning with this security requirement and allowing the thread to be created successfully.

Missake212:
I managed to find someone else able to test, it still doesn't work unfortunately.

[shad_log.txt](https://github.com/user-attachments/files/21940057/shad_log.txt)

PR #3446, opened 2025-08-23
texture_cache: Only queue image downloads if the image address is greater than zero.

StevenMiller123:
If the image address is zero, that means we're trying to download a null image, which causes other issues down the line.

This fixes an issue in Cyberpunk 2077 (CUSA16596) when readbackLinearImages is enabled.

Credits to @squidbus for helping.


ShaggyLorean:
Looking forward to more texture cache fixes/implements ❤️

PR #3226, opened 2025-07-10
video_core: support for RT layer outputs

polybiusproxy:
Games can choose what slice of the render target to write to from within a shader—this is commonly used for drawing cubemaps in a single pass. This can be done from geometry, vertex, or tesselation eval shaders.
This works by outputting the selected slice to the Z component of the pos1 target:
```
s_load_dwordx16 s[40:55], s[6:7], 0x0
/*snip*/
v_mov_b32       v15, s40
/*snip*/
exp             pos1, off, off, v15, off done
```

Can also be seen used on the AMD compilers:
- https://godbolt.org/z/xT9nfGdcs
- https://github.com/GPUOpen-Drivers/llpc/blob/6222f1ff6ba7f00b085003c1a05fdbdc31356235/lgc/lowering/LowerInOut.cpp#L1526

PR #3447, opened 2025-08-23
shader_recompiler: Improve shader exports accuracy (part 1)

raphaelthegreat:
This includes 2 changes
* Rebased version of https://github.com/shadps4-emu/shadPS4/pull/3226. The code there is relevant to the file being refactored so it would cause more conflict. It's also an effort to prevent that PR from code-rotting a second time

* MRT swizzle is now applied to the color vector instead of the destination channels. This appears more accurate to what hardware does, I've also written an example in the commit description of a broken output because of it.

This is part 1 because this code also needs fixes to account for broken blending. The comp_swap swizzle is applied after blending on hardware, so swizzling shader output can break blending step because it swaps its inputs. In most cases when no swizzle is applied or simple BGR swizzles it's okay but when alpha and RGB components are swapped (e.g ARGB) it doesn't work, and iirc this affects some effects in Knack.

That probably needs to implement blending step in software with fragment shader interlock which is a big change in itself

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv.cpp
@@ -272,6 +272,9 @@ void SetupCapabilities(const Info& info, const Profile& profile, const RuntimeIn
     if (info.has_image_query) {
         ctx.AddCapability(spv::Capability::ImageQuery);
     }
+    if (info.has_layer_output) {
+        ctx.AddCapability(spv::Capability::ShaderLayer);
+    }
     if ((info.uses_image_atomic_float_min_max && profile.supports_image_fp32_atomic_min_max) ||
         (info.uses_buffer_atomic_float_min_max && profile.supports_buffer_fp32_atomic_min_max)) {
         ctx.AddExtension("SPV_EXT_shader_atomic_float_min_max");
--- a/src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
@@ -16,39 +16,6 @@
 namespace Shader::Backend::SPIRV {
 namespace {
 
-Id VsOutputAttrPointer(EmitContext& ctx, VsOutput output) {
-    switch (output) {
-    case VsOutput::ClipDist0:
-    case VsOutput::ClipDist1:
-    case VsOutput::ClipDist2:
-    case VsOutput::ClipDist3:
-    case VsOutput::ClipDist4:
-    case VsOutput::ClipDist5:
-    case VsOutput::ClipDist6:
-    case VsOutput::ClipDist7: {
-        const u32 index = u32(output) - u32(VsOutput::ClipDist0);
-        const Id clip_num{ctx.ConstU32(index)};
-        ASSERT_MSG(Sirit::ValidId(ctx.clip_distances), "Clip distance used but not defined");
-        return ctx.OpAccessChain(ctx.output_f32, ctx.clip_distances, clip_num);
-    }
-    case VsOutput::CullDist0:
-    case VsOutput::CullDist1:
-    case VsOutput::CullDist2:
-    case VsOutput::CullDist3:
-    case VsOutput::CullDist4:
-    case VsOutput::CullDist5:
-    case VsOutput::CullDist6:
-    case VsOutput::CullDist7: {
-        const u32 index = u32(output) - u32(VsOutput::CullDist0);
-        const Id cull_num{ctx.ConstU32(index)};
-        ASSERT_MSG(Sirit::ValidId(ctx.cull_distances), "Cull distance used but not defined");
-        return ctx.OpAccessChain(ctx.output_f32, ctx.cull_distances, cull_num);
-    }
-    default:
-        UNREACHABLE_MSG("Vertex output {}", u32(output));
-    }
-}
-
 Id OutputAttrPointer(EmitContext& ctx, IR::Attribute attr, u32 element) {
     if (IR::IsParam(attr)) {
         const u32 attr_index{u32(attr) - u32(IR::Attribute::Param0)};
@@ -76,15 +43,14 @@ Id OutputAttrPointer(EmitContext& ctx, IR::Attribute attr, u32 element) {
         }
     }
     switch (attr) {
-    case IR::Attribute::Position0: {
+    case IR::Attribute::Position0:
         return ctx.OpAccessChain(ctx.output_f32, ctx.output_position, ctx.ConstU32(element));
-    }
-    case IR::Attribute::Position1:
-    case IR::Attribute::Position2:
-    case IR::Attribute::Position3: {
-        const u32 index = u32(attr) - u32(IR::Attribute::Position1);
-        return VsOutputAttrPointer(ctx, ctx.runtime_info.vs_info.outputs[index][element]);
-    }
+    case IR::Attribute::ClipDistance:
+        return ctx.OpAccessChain(ctx.output_f32, ctx.clip_distances, ctx.ConstU32(element));
+    case IR::Attribute::CullDistance:
+        return ctx.OpAccessChain(ctx.output_f32, ctx.cull_distances, ctx.ConstU32(element));
+    case IR::Attribute::RenderTargetId:
+        return ctx.output_layer;
     case IR::Attribute::Depth:
         return ctx.frag_depth;
     default:
@@ -105,11 +71,13 @@ std::pair<Id, bool> OutputAttrComponentType(EmitContext& ctx, IR::Attribute attr
     }
     switch (attr) {
     case IR::Attribute::Position0:
-    case IR::Attribute::Position1:
-    case IR::Attribute::Position2:
-    case IR::Attribute::Position3:
+    case IR::Attribute::ClipDistance:
+    case IR::Attribute::CullDistance:
     case IR::Attribute::Depth:
         return {ctx.F32[1], false};
+    case IR::Attribute::RenderTargetId:
+    case IR::Attribute::ViewportId:
+        return {ctx.S32[1], true};
     default:
         UNREACHABLE_MSG("Write attribute {}", attr);
     }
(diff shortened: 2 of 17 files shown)


squidbus (approved):
Looks good and tested in a few games, going to wait a while for more testing though.

Missake212:
> Looks good and tested in a few games, going to wait a while for more testing though.

I'll test later on today.

Randomuser8219:
I'll try this on Knack (CUSA00068) when I get the chance.

Randomuser8219:
I'm not seeing any graphical improvements so far.

raphaelthegreat:
This PR doesn't fix the blending issue, so not sure if you were expecting that

Missake212:
Tested 47 games for about 1/2 minutes each, no regressions on my end (Nvidia GPU on Windows, tested all the games without readbacks beside Driveclub).

raphaelthegreat:
Believe I spotted another swizzling issue, the blending write mask is not swizzled to match the shader output. So for example on a RGBA8 render target with SWAP_ALT (BGRA8) if the CB_TARGET_MASK only has the R component set it would result in wrong channel being written when doing shader-side swizzling. I can fix this in this PR if wanted

Xcedf:
> Believe I spotted another swizzling issue, the blending write mask is not swizzled to match the shader output. So for example on a RGBA8 render target with SWAP_ALT (BGRA8) if the CB_TARGET_MASK only has the R component set it would result in wrong channel being written when doing shader-side swizzling. I can fix this in this PR if wanted

Don't know if this related I can only guess but it could possibly fix long standing text issue in GoW3.

StevenMiller123:
No regressions/improvements in the games I've tested, though I tested before the most recent push.

StevenMiller123:
No regressions or improvements in my tests with the new push.

raphaelthegreat:
The graphics key setup code desperately needs a cleanup and might be a good idea to bring back the swizzling bypass path by using appropriate image view formats, but probably gonna do those in a different PR

Xcedf:
Yep, confirm, last push fixes GoW3 text
image

Randomuser8219:
image
image
This fixed some of the remaining color issues in KNACK (CUSA00068)
Here's what was fixed:

- Red eyes on chapter 1 robots
- Sunstone outlines
- Skybox
- Monitors in laboratory
- Lights in laboratory

raphaelthegreat:
I can't really see anything different

StevenMiller123:
Seems like this regressed Persona 3 Reload
```
[Debug] <Critical> position.h:49 ExportPosition: Unreachable code!
Unhandled output 0 on attribute 13
```
[CUSA37521.log](https://github.com/user-attachments/files/21952792/CUSA37521.log)

Randomuser8219:
> I can't really see anything different

Here's the same spots from a build from before this was merged into main, it looks quite different:
image
image

raphaelthegreat:
> Seems like this regressed Persona 3 Reload
> 
> ```
> [Debug] <Critical> position.h:49 ExportPosition: Unreachable code!
> Unhandled output 0 on attribute 13
> ```
> 
> [CUSA37521.log](https://github.com/user-attachments/files/21952792/CUSA37521.log)

I have commited a fix for this in my other PR

squidbus:
Missed this in my testing before, but it seems the first commit here causes performance to regress a lot in some scenarios. For example, the song select menu in Persona 3 Dancing went from 60 FPS to 30 FPS.

StevenMiller123:
Sword Art Online: Lost Song now crashes on 
```
[Debug] <Critical> position.h:49 ExportPosition: Unreachable code!
Unhandled output 1 on attribute 13
```
(This was tested on the previous build of your other PR, since the newest push hasn't finished compiling yet)
[CUSA02823.log](https://github.com/user-attachments/files/21953563/CUSA02823.log)

raphaelthegreat:
> Missed this in my testing before, but it seems the first commit here causes performance to regress a lot in some scenarios. For example, the song select menu in Persona 3 Dancing went from 60 FPS to 30 FPS.

That sounds strange, there isn't really anything that would hog CPU time there. Maybe some hashing got messed up and a cache is failing?

squidbus:
> That sounds strange, there isn't really anything that would hog CPU time there. Maybe some hashing got messed up and a cache is failing?

I don't really fully understand yet but it seems to literally just be the presence of `num_layers` in `RenderState`. If I comment out all usages, it still is slow. But if I remove it from the struct entirely it goes back to normal.

raphaelthegreat:
Try to memset the struct to zero on the constructor and manually default layers to 1, its possible the additional member added some padding which would cause equality operator to always return false and break rp, I imagine that would murder perf on tiled GPUs like Apple sillicon

raphaelthegreat:
The operator should probably be changed to a more sane version, even if its not the issue
```cpp
bool operator==(const RenderState& other) const noexcept = default;
```

squidbus:
> The operator should probably be changed to a more sane version, even if its not the issue
> 
> ```c++
> bool operator==(const RenderState& other) const noexcept = default;
> ```

Adding the memset works, but changing the equals operator to the default breaks it again.

raphaelthegreat:
Hmm the default operator should do member wise comparsion to avoid padding issues, so weird that it doesn't work but memset does

squidbus:
This also seems to break the loading animation after choosing a save in Hatsune Miku: Project DIVA X, although maybe it's because of the blending stuff.

raphaelthegreat:
Can be checked if game has ARGB or ABGR swizzle and blending input uses src color/alpha

squidbus:
Actually looking closer it seems to be some kind of vertex explosions, again from the first commit.

Specifically it goes away with the attribute set for `Output::GsMrtIndex` in `ExportPosition` commented out.

squidbus:
It seems that game sets `gl_Layer = gl_InstanceIndex` in multiple shaders. Maybe it's just some kind of effect that wasn't happening before and now that it's happening it turns out to be broken, not sure. However I don't actually see multiple layers being bound to the render targets.

raphaelthegreat:
Are the draws that do this instanced? And still don't get how this would be causing vertex explosions, it should only result in missing geometry if some instances are rendered to non-existent layers

raphaelthegreat:
I think this game uses base instance/vertex in its fetch shader, so you should also check if the v_add instruction there is applied directly to v3 (plain instance id) and not some other VGPR. Currently the compiler subtly relies on this so it can use gl_InstanceIndex everywhere which includes base instance

squidbus:
> I think this game uses base instance/vertex in its fetch shader, so you should also check if the v_add instruction there is applied directly to v3 (plain instance id) and not some other VGPR. Currently the compiler subtly relies on this so it can use gl_InstanceIndex everywhere which includes base instance

I don't see any `v3` at all in the fetch shaders for the ones I'm looking at, and in the vertex shader only use of `v3` is left-shifting `v0` near the start and then the export later.

raphaelthegreat:
Does the draw have a base instance?

squidbus:
I think this may actually be a MoltenVK bug when `Layer` is used with tessellation (for RectList/QuadList). I'll look into it further myself.

raphaelthegreat:
Maybe the hardware doesn't support gl_Layer in vertex shaders and needs a geoshader passthrough?

squidbus:
> Maybe the hardware doesn't support gl_Layer in vertex shaders and needs a geoshader passthrough?

It does (especially since Metal does not support geometry shaders). MoltenVK handles translating tessellation using compute, and the issue seems to be that on the vertex compute end the `Layer` is part of the output structure, but on the receiving tessellation compute end it's not present in the structure, so I think the vertex data is getting messed up (and I assume the layer is also just discarded completely):

```
struct main0_out
{
    float4 out_attr0;
    float4 out_attr1;
    float4 gl_Position;
    uint gl_Layer;
};

```

```
struct main0_in
{
    float4 m_27;
    float4 m_28;
    float4 gl_Position;
};
```

raphaelthegreat:
Ah lol forgot about geoshader support. Nice find!

squidbus:
To be honest, it is unclear to me from the spec whether you are supposed to write `Layer` from the vertex shader when tessellation is in effect, or if you are supposed to write from the tessellation evaluation stage. The only thing I can find is some wording from the equivalent GL spec where the value is undefined if not written from the *final* vertex-tessellation-geometry stage.

DanielSvoboda:
Before, the Ōkami game would start, but some things seemed a bit transparent.
Now, with this PR, the game won't start anymore, and there's this error:
```
[Debug] <Critical> export.cpp:132 operator(): Assertion Failed!
Compressed exports only supported for render targets
```

raphaelthegreat:
This is likely a compressed depth export, before those were not handled properly, the format needs to be fetched from SPI_SHADER_Z_FORMAT

Zamaeda:
This PR has made bloodborne water to flicker and have some overlaps 
image

raphaelthegreat:
Can you post a save file for that area?

Zamaeda:
Yes just a sec

Zamaeda:
[userdata0000.zip](https://github.com/user-attachments/files/21959391/userdata0000.zip)

raphaelthegreat:
This appears to be caused by clip distance being compiled now, there is a small accuracy difference between NVIDIA and AMD that causes former to clip some vertices. I'm not super sure on how to deal with this, either need to disable clip distances on NVIDIA or possibly add a small bias to avoid the error

squidbus:
So I got clarification on the tessellation thing. It seems that when we use tessellation for rect/quad list, we should be outputting the layer from there instead of vertex stage: https://docs.vulkan.org/spec/latest/chapters/interfaces.html#interfaces-builtin-variables-layer

> The last active [pre-rasterization shader stage](https://docs.vulkan.org/spec/latest/chapters/pipelines.html#pipelines-graphics-subsets-pre-rasterization) (in pipeline order) controls the Layer that is used. Outputs in previous shader stages are not used, even if the last stage fails to write the Layer.

So the way it currently exports in this case is invalid by the Vulkan spec, and we need to instead propagate it to tess evaluation. I'm not sure myself the best way to wire this up.

raphaelthegreat:
The only way I can think of is hijacking an attribute slot from VS to pass the layer to TES. In ExportPosition check if primitive is rect/quad list and change attribute store from builin to param store and when emitting the tessellation shaders query the vs info structure if it has a layer store and propagate it. Though I also wouldn't have an issue in ignoring layer stores in this case and logging a warning instead as its a lot of work for a case that doesn't even matter in the end

squidbus:
Yeah we'll see if this ever actually causes problems I guess. Adding the skip here: https://github.com/shadps4-emu/shadPS4/pull/3457

tochnonechelovek:
Broke Destiny.
Before this PR: 
[CUSA00568 old](https://github.com/user-attachments/files/21978526/CUSA00568.log)

After this PR:
[CUSA00568 new](https://github.com/user-attachments/files/21978539/CUSA00568.log)

Was in menus before and requiring connection, but now crashes before entering menus. No graphics, no sound.

gandalfthewhite19890404:
It cause blur image (you can see it briefly on cockpit gauges) in Project Cars
image
log:
[shad_log-6d98a5a-bad.txt](https://github.com/user-attachments/files/23833546/shad_log-6d98a5a-bad.txt)


Previous version:
[shad_log-d42f4fc-good.txt](https://github.com/user-attachments/files/23833554/shad_log-d42f4fc-good.txt)
image

PR #3449, opened 2025-08-23
vk_pipeline_cache: Cleanup graphics key refresh

raphaelthegreat:
- Remove color write mask dynamic state. This needs a whole extension and checks just for it, and it doesn't even benefit compilation times because the pipeline key still hashes write masks, so triggers recompilations. It would be better to readd this when the entire color blending state can properly be made dynamic, since write masks are not a "hot" pipeline state that triggers often recompilations

- Moved MoltenVK unorm fixup into shader recompiler. The color buffer struct has contained both data and number formats for a while, so little reason to have pipeline cache do the check for it, instead of the intended user

- Split out shader stage binding from graphics pipeline key refresh, to make the whole function easier to parse

- Remove the concept of color buffer compaction. This never made much sense to me, maybe this change is wrong and this will break stuff but I believe hardware has 8 fixed MRT slots, shaders can export to those slots so the mapping between those is 1-1. I feel any compacting will break this if there are gaps between render targets.

The binding code has been rewritten, so first pass iterates all the slots and binds whatever is valid, then mrt_mask is used to fetch the number of render targets bound, which feels more correct then just counting the number of valid targets, as it will include gaps properly, and masks out render targets that don't appear in the mask (the gaps essentially)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -918,6 +918,7 @@ set(SHADER_RECOMPILER src/shader_recompiler/exception.h
                       src/shader_recompiler/ir/opcodes.inc
                       src/shader_recompiler/ir/patch.cpp
                       src/shader_recompiler/ir/patch.h
+                      src/shader_recompiler/ir/position.h
                       src/shader_recompiler/ir/post_order.cpp
                       src/shader_recompiler/ir/post_order.h
                       src/shader_recompiler/ir/program.cpp
--- a/src/shader_recompiler/frontend/translate/export.cpp
+++ b/src/shader_recompiler/frontend/translate/export.cpp
@@ -93,17 +93,24 @@ void Translator::ExportRenderTarget(const GcnInst& inst) {
         }
     }
 
+    // Metal seems to have an issue where 8-bit unorm/snorm/sRGB outputs to render target
+    // need a bias applied to round correctly; detect and set the flag for that here.
+    const auto needs_unorm_fixup = profile.needs_unorm_fixup &&
+                                   (color_buffer.num_format == AmdGpu::NumberFormat::Unorm ||
+                                    color_buffer.num_format == AmdGpu::NumberFormat::Snorm ||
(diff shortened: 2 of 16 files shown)


StevenMiller123:
Can confirm that the Persona 3 Reload regression is fixed with the latest push.
[CUSA37521.log](https://github.com/user-attachments/files/21953592/CUSA37521.log)

I haven't run into any regressions or relevant improvements from this PR yet.

squidbus (approved):

GHU7924:
I don't know for sure if there are any other graphical changes in the games, but the most noticeable one for me was in [CUSA30992] Teenage Mutant Ninja Turtles: Shredder's Revenge.

Main before the merger



PR 
[Video.zip](https://github.com/user-attachments/files/21956092/Video.zip)

[CUSA30992.log](https://github.com/user-attachments/files/21956107/CUSA30992.log)

raphaelthegreat:
Is it a regression or improvement I can't tell

raphaelthegreat:
Is the background missing now?

GHU7924:
> Is the background missing now?

Yes, the background has disappeared and the picture is superimposed on the picture or the transparency has been lost, I don’t know what to call it correctly.

Last Main (You can compare with the screenshots of the previous Main, everything was fine there.)

raphaelthegreat:
I'll look at this thx

raphaelthegreat:
I don't see anything wrong caused by changes in this PR, but for some reason shader side discards just stopped working with this PR on NVIDIA specifically. It looks like a driver bug triggered by an unrelated change here, but idk exactly. If someone with older NVIDIA driver can test would help. RADV works fine and I didn't see any validation errors

raphaelthegreat:
Actually I just realized your test was with 566.36 which is quite old, so maybe it's not a driver bug? I cannot explain yet, need to bisect the exact code line that triggers it

raphaelthegreat:
It's caused by the removal of color write dynamic state, seems the NVIDIA driver isn't working correctly when its part of the pipeline

PR #3450, opened 2025-08-24
video_core: Fix some struct comparisons.

squidbus:
* In `specialization`, replace some three-way comparison operator declarations with equality operators, as some of the contents are not three-way comparable.
* For state structs that are compared using `memcmp`, initialize full struct (including padding) to zero in constructor.

Fixes some lag due to state being constantly deemed to be dirty.


squidbus:
@raphaelthegreat Can you double check this?

raphaelthegreat:
Looks fine to me

PR #3453, opened 2025-08-24
renderer_vulkan: Restore color write dynamic state

raphaelthegreat:
Should fix https://github.com/shadps4-emu/shadPS4/pull/3449#issuecomment-3217975161
It appears shader side discards break when color write masks are not dynamic on NVIDIA, I have no idea why but restore it for now

### DIFF
--- a/src/video_core/amdgpu/pixel_format.h
+++ b/src/video_core/amdgpu/pixel_format.h
@@ -127,6 +127,14 @@ union CompMapping {
         };
     }
 
+    [[nodiscard]] u32 ApplyMask(u32 mask) const {
+        u32 swizzled_mask{};
+        for (u32 i = 0; i < 4; ++i) {
+            swizzled_mask |= ((mask >> i) & 1) << Map(i);
+        }
+        return swizzled_mask;
+    }
+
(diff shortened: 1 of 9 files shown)


GHU7924:
This PR fixed the problem.

изображение

larfy298:
This PR solves two critical performance regressions on RDNA3 and RDNA4 GPUs, not sure how important it is but thought it was worth mentioning

raphaelthegreat:
These performance regressions were caused by the PR that removed it?

larfy298:
There was a regression from 3449, the PR that this fixes, but also another from a few weeks ago that I don't remember the exact PR of. This PR resolves both of those issues so AMD is improved a ton on this PR

squidbus:
> It appears shader side discards break when color write masks are not dynamic on NVIDIA

Is this really true? Surely there are plenty of users of discards without dynamic color write.

raphaelthegreat:
To be more specific it doesn't break the color output itself, but rather the depth buffer is written as if the discard never happened (even though it does get executed)

squidbus:
Okay, well that's unfortunate.

PR #2972, opened 2025-05-22
Windows filesystem fixes

georgemoralis:
-Remove trailing separator when opening dirs if exist (windows doesn't like it)
-Support for opening dirs in windows

This is not probably the best code but might give better results in games that reports "Opening directories is not fully supported on this platform"

PR #3455, opened 2025-08-24
Filesystem: Abstract handling of directory files

StevenMiller123:
Based on similar work done for device files in the past, this PR separates directory file-related code into separate classes, designed to handle the different types of directories present on a real PS4.

This PR adds cross-platform support for `read` and `lseek` operations on directories, and fixes remaining inaccuracies with `getdents` behavior. I've also touched up a few issues in likernel filesystem code while I was there.

So far I've got two types of folders implemented, based on the two types I've seen while testing, but the code leaves room for additional directory types if those become necessary.

In my testing, this fixes IxSHE Tell (CUSA17112) and anywhereVR (CUSA08643) crashing on startup for Windows devices, and fixes a regression in updated versions of Final Fantasy XV (CUSA01633) caused by #3416


StevenMiller123:
If this gets merged, this will effectively replace #2972

georgemoralis:
> If this gets merged, this will effectively replace #2972

;/

Missake212:
Seems like this fixes the crash UFC 2 and 4 used to experience before, now they both hang spamming net stuff instead.

UFC 2:

[CUSA01936PR.log](https://github.com/user-attachments/files/21958473/CUSA01936.log)

[CUSA01936main.log](https://github.com/user-attachments/files/21958479/CUSA01936.log)

UFC 4:

[CUSA14209PR.log](https://github.com/user-attachments/files/21958477/CUSA14209.log)

[CUSA14209main.log](https://github.com/user-attachments/files/21958481/CUSA14209.log)

Randomuser8219:
image
This brings Pinball FX3 (CUSA04915) back in-game.
I know The Sims 4 had a filesystem issue that caused it to softlock when trying to select a house, but I can't confirm if it's fixed because the game appears to have regressed in later main builds as that BlendFactor error always appears.

StevenMiller123:
@Randomuser8219 You could try using nullGPU for The Sims 4, I don't think it's possible to hit the error you're describing when that's enabled.

StevenMiller123:
New push fixes a regression I spotted in Grand Theft Auto V (CUSA00419).

StevenMiller123:
Latest push fixes memory corruption issues in anywhereVR (CUSA08643)

StevenMiller123:
Newest push fixes a regression in some versions of Minecraft (CUSA00744).
I've fixed all the issues I've run into, so I'm marking this as ready for review.

src/core/directories/base_directory.cpp
```diff
@@ -0,0 +1,15 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
```
squidbus: Should have 2025 as year. Same for any new file.

squidbus (commented):
Have not fully looked over yet. General comment though, we should probably have all the filesystem device/directory/etc stuff centralized under the `file_sys` directory.

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,140 @@
...
+                                          4);
+
+        directory_size += dirent.d_reclen;
+    });
+
+    // The last entry of a normal directory should have d_reclen covering the remaining data.
+    // Since the dirents of a folder are constant by this point, we can modify the last dirent
+    // before creating the emulated file buffer.
+    u64 filler_count = Common::AlignUp(directory_size, 512) - directory_size;
+    dirents[dirents.size() - 1].d_reclen += filler_count;
```
squidbus, line 43: Is size guaranteed to be >= 1 here?

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,140 @@
...
+    // Reading from standard directories seems to be based around file pointer logic.
+    // Keep an internal buffer representing the raw contents of this file descriptor,
+    // then emulate the various read functions with that.
+    directory_size = Common::AlignUp(directory_size, 512);
+    data_buffer = (char*)malloc(directory_size);
+    memset(data_buffer, 0, directory_size);
+
+    char* current_dirent = data_buffer;
+    for (NormalDirectoryDirent dirent : dirents) {
+        NormalDirectoryDirent* dirent_to_write = (NormalDirectoryDirent*)current_dirent;
```
squidbus: Prefer C++ style `static_cast`. Same anywhere else this is done.

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,140 @@
...
+    // Since the dirents of a folder are constant by this point, we can modify the last dirent
+    // before creating the emulated file buffer.
+    u64 filler_count = Common::AlignUp(directory_size, 512) - directory_size;
+    dirents[dirents.size() - 1].d_reclen += filler_count;
+
+    // Reading from standard directories seems to be based around file pointer logic.
+    // Keep an internal buffer representing the raw contents of this file descriptor,
+    // then emulate the various read functions with that.
+    directory_size = Common::AlignUp(directory_size, 512);
+    data_buffer = (char*)malloc(directory_size);
```
squidbus: This doesn't appear to be freed anywhere. Should probably use something like a `std::vector<u8>` with pre-reserved size which will auto-free when the `NormalDirectory` is destructed.

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,140 @@
...
+    if (file_offset >= directory_size) {
+        return ORBIS_OK;
+    }
+
+    s64 remaining_data = directory_size - file_offset;
+    s64 bytes = nbytes > remaining_data ? remaining_data : nbytes;
+    memset(buf, 0, bytes);
+
+    char* read_pointer = data_buffer + file_offset;
+    memcpy(buf, read_pointer, bytes);
```
squidbus: Why do we need to memset to 0 and then copy to the buffer, when the number of bytes is the same for both?

src/core/file_sys/directories/normal_directory.cpp
StevenMiller123: Yes, since every directory has an entry for `.` and `..`

src/core/file_sys/directories/normal_directory.cpp
StevenMiller123: Swapped all C-style type casts for C++ ones, though a few of my type casts needed to use `reinterpret_cast` instead of `static_cast`.

src/core/file_sys/directories/normal_directory.cpp
StevenMiller123: Swapped the buffer to a `std::vector<u8>` as per your suggestion, seems to work fine in my tests.

src/core/file_sys/directories/normal_directory.cpp
StevenMiller123: Was a remnant from old code, I've removed it in my latest push.

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "common/logging/log.h"
+#include "common/singleton.h"
+#include "core/file_sys/fs.h"
+#include "normal_directory.h"
+
+namespace Core::Directories {
+
+std::shared_ptr<BaseDirectory> NormalDirectory::Create(std::string_view guest_directory) {
+    return std::shared_ptr<BaseDirectory>(
+        reinterpret_cast<Directories::BaseDirectory*>(new NormalDirectory(guest_directory)));
```
squidbus: Missed this on other pass, can we use `std::make_shared` here? Possibly with `std::static_pointer_cast` it would work for your case.

src/core/file_sys/directories/pfs_directory.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include "common/logging/log.h"
+#include "common/singleton.h"
+#include "core/file_sys/fs.h"
+#include "pfs_directory.h"
+
+namespace Core::Directories {
+
+std::shared_ptr<BaseDirectory> PfsDirectory::Create(std::string_view guest_directory) {
+    return std::shared_ptr<BaseDirectory>(
+        reinterpret_cast<Directories::BaseDirectory*>(new PfsDirectory(guest_directory)));
```
squidbus: Same here.

src/core/file_sys/directories/normal_directory.cpp
StevenMiller123: Addressed with latest push.

squidbus (approved):
Looks good if the small clang-format issue is fixed.

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,139 @@
...
+    file_offset = offset;
+    s64 bytes_read = readv(iov, iovcnt);
+    file_offset = old_file_pointer;
+    return bytes_read;
+}
+
+s64 NormalDirectory::lseek(s64 offset, s32 whence) {
+    switch (whence) {
+    case 0: {
+        file_offset = 0;
```
raphaelthegreat: Is missing break here intentional?

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,139 @@
...
+s64 NormalDirectory::read(void* buf, u64 nbytes) {
+    // Nothing left to read.
+    if (file_offset >= directory_size) {
+        return ORBIS_OK;
+    }
+
+    s64 remaining_data = directory_size - file_offset;
+    s64 bytes = nbytes > remaining_data ? remaining_data : nbytes;
+
+    memcpy(buf, data_buffer.data() + file_offset, bytes);
```
raphaelthegreat:
```suggestion
    std::memcpy(buf, data_buffer.data() + file_offset, bytes);
```

src/core/file_sys/directories/pfs_directory.cpp
```diff
@@ -0,0 +1,220 @@
...
+                                              sizeof(dirent.d_namlen) + sizeof(dirent.d_reclen) +
+                                              (dirent.d_namlen + 1),
+                                          8);
+
+        // To handle some obscure dirents_index behavior,
+        // keep track of the "actual" length of this directory.
+        directory_content_size += dirent.d_reclen;
+    });
+
+    directory_size = Common::AlignUp(directory_content_size, 0x10000);
```
raphaelthegreat: Should probably have a constant instead of magic number

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+s64 NormalDirectory::read(void* buf, u64 nbytes) {
+    // Nothing left to read.
+    if (file_offset >= directory_size) {
+        return ORBIS_OK;
+    }
+
+    s64 remaining_data = directory_size - file_offset;
+    s64 bytes = nbytes > remaining_data ? remaining_data : nbytes;
```
raphaelthegreat:
```suggestion
    const s64 remaining_data = directory_size - file_offset;
    const s64 bytes = nbytes > remaining_data ? remaining_data : nbytes;
```
Also apply in other places where const is applicable

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+    // Reading from standard directories seems to be based around file pointer logic.
+    // Keep an internal buffer representing the raw contents of this file descriptor,
+    // then emulate the various read functions with that.
+    directory_size = Common::AlignUp(directory_size, 512);
+    data_buffer.reserve(directory_size);
+    memset(data_buffer.data(), 0, directory_size);
+
+    u8* current_dirent = data_buffer.data();
+    for (NormalDirectoryDirent dirent : dirents) {
```
raphaelthegreat:
```suggestion
    for (const NormalDirectoryDirent& dirent : dirents) {
```
Avoids copy

src/core/file_sys/directories/normal_directory.cpp
```diff
@@ -0,0 +1,139 @@
+// SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "common/alignment.h"
+#include "common/assert.h"
+#include "common/logging/log.h"
+#include "common/singleton.h"
+#include "core/file_sys/fs.h"
+#include "normal_directory.h"
```
raphaelthegreat:
```suggestion
#include "core/file_sys/directories/normal_directory.h"
```
Adjust in other files as well to have full path for consistency

src/core/file_sys/directories/normal_directory.cpp
StevenMiller123: Was intentional at one point, but it's unnecessary now. I'll get that fixed.

georgemoralis (approved):

Missake212:
Seems like one of the commit after I tested the UFC games made them go back to crashing, not sure which one.

StevenMiller123:
@Missake212 Given the circumstances, I see two potential actions for commits that could've regressed the title:
https://github.com/StevenMiller123/shadPS4/actions/runs/17194081435
https://github.com/StevenMiller123/shadPS4/actions/runs/17195215677

If you have the time and are willing, could you let me know if these UFC games work on either of these two actions?

Missake212:
> @Missake212 Given the circumstances, I see two potential actions for commits that could've regressed the title: https://github.com/StevenMiller123/shadPS4/actions/runs/17194081435 https://github.com/StevenMiller123/shadPS4/actions/runs/17195215677
> 
> If you have the time and are willing, could you let me know if these UFC games work on either of these two actions?

Sorry it was pretty late for me, tested both actions and same crash as usual (the second action is called "Merge branch 'main' into abstract-directories", so I don't think it's the one you meant to send me ?).

PR #3456, opened 2025-08-25
Certain changes to README

Randomuser8219:
- As the Ko-Fi has existed for almost a year, I reworded the mention in README to sound less like it was just made.
- Replaced Bloodborne video link with newer one.
- Fixed grammar in "Why?" section
- Removed duplicated sentence in Firmware Files section, will revert this change if it was intentional.


tvoracek:
Should probably have CI Skip so it doesn't trigger a release...

PR #3459, opened 2025-08-25
Fix resetting patches when the isEnabled field is not present

kalaposfos13:
Fixes the issue when you have two patches like so:
```xml
<Metadata Title="game" Name="patch1" Note="" Author="kalaposfos" PatchVer="1.0" AppVer="01.00" AppElf="eboot.bin" isEnabled="true">
    <PatchList>
        <Line Type="bytes" Address="0x0068bf60" Value="31c0c3"/>
    </PatchList>
</Metadata>
<Metadata Title="game" Name="patch2" Note="" Author="kalaposfos" PatchVer="1.0" AppVer="01.00" AppElf="eboot.bin">
    <PatchList>
        <Line Type="bytes" Address="0x0068bf60" Value="31c0c3"/>
    </PatchList>
</Metadata>
```
The first has the isEnabled field, but the second does not. The expected behaviour would be that the first is turned on, but the second isn't, but what happens is that both get turned on, because the internal isEnabled flag doesn't get reset, so the first "true" gets carried over.

### DIFF
--- a/src/common/memory_patcher.cpp
+++ b/src/common/memory_patcher.cpp
@@ -287,6 +287,7 @@ void OnGameLoaded() {
                     QString appVer = xmlReader.attributes().value("AppVer").toString();
 
                     // Check and update the isEnabled attribute
+                    isEnabled = false;
                     for (const QXmlStreamAttribute& attr : xmlReader.attributes()) {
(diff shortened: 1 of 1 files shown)


kalaposfos13:
Should've checked that this already has an open PR for it before I opened this

PR #3460, opened 2025-08-26
Libraries: libSceHmdSetupDialog stubs

StevenMiller123:
These are some basic stubs designed to get some VR-optional titles further by simulating a user canceling the dialog.

VR-required titles are expected to endlessly spam these calls, at least based on Sony's "When you see this screen repeatedly even after canceling, press the PS button to go to the home screen" message on the actual HmdSetupDialog.


squidbus (approved):

StevenMiller123:
@Randomuser8219 Based on the log you've provided, MiniWood VR is a VR-required title behaving as expected. While it's behavior may have regressed, it's now behaving like it would on a PS4 without a PSVR attached, which is currently what we're emulating.
It will probably remain like this until someone tackles proper VR support, which I doubt will happen anytime soon.

PR #3464, opened 2025-08-27
Replace direct usage of wrgsbase and rdgsbase with a more portable solution 

kalaposfos13:
From what I found online (https://www.kernel.org/doc/html/next/x86/x86_64/fsgs.html), using these syscalls are the intended way to access these instructions on Linux. However, the syscall also supports setups where this instruction is missing or is unsupported by the kernel, namely, PS4 Linux in my case (and with this, the only modification required for the emulator to run there is recompiling with an older march that's supported by the console). From testing by me and bread, this also doesn't come with any noticeable performance drop either.

### DIFF
--- a/src/core/tls.cpp
+++ b/src/core/tls.cpp
@@ -19,8 +19,8 @@
 #include <pthread.h>
 #endif
 #if defined(__linux__) && defined(ARCH_X86_64)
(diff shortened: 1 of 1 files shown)


squidbus (approved):
Needs clang-format fix

PR #3468, opened 2025-08-28
Net Fixes

georgemoralis:
1. Fix handling of ORBIS_NET_SO_ONESBCAST to support proper broadcast address translation on send.
2. Fix setsockopt for SO_(RCV/SND)TIMEO on Windows/Linux.


src/core/libraries/network/posix_sockets.cpp
```diff
@@ -281,16 +282,42 @@ int PosixSocket::SetSocketOptions(int level, int optname, const void* optval, u3
...
+            }
+            std::vector<char> val;
+            const auto optname_nat = (optname == ORBIS_NET_SO_SNDTIMEO) ? SO_SNDTIMEO : SO_RCVTIMEO;
+            int timeout_us = *(const int*)optval;
+#ifdef _WIN32
+            DWORD timeout = timeout_us / 1000;
+            val.insert(val.end(), (char*)&timeout, (char*)&timeout + sizeof(timeout));
+            optlen = sizeof(timeout);
+#else
+            timeval timeout{.tv_sec = timeout_us / 1000000, .tv_usec = timeout_us % 1000000};
```
mikusp, line 354: Only the timeout declaration needs to be ifdef'd, the rest is the same between branches

StevenMiller123:
Seems like the `setsockopt` changes don't work properly on Linux?
I tested with libSceHttp LLE, and while it's no longer returning `EINVAL`, `setsockopt` returns `ENOPROTOOPT` instead, which causes libSceHttp to close the socket and fail.

libSceHttp seems to use both `SO_RCVTIMEO` and `SO_SNDTIMEO`, and both hit the same issues.

src/core/libraries/network/posix_sockets.cpp
```diff
@@ -281,16 +282,40 @@ int PosixSocket::SetSocketOptions(int level, int optname, const void* optval, u3
...
+            const auto optname_nat = (optname == ORBIS_NET_SO_SNDTIMEO) ? SO_SNDTIMEO : SO_RCVTIMEO;
+            int timeout_us = *(const int*)optval;
+#ifdef _WIN32
+            DWORD timeout = timeout_us / 1000;
+#else
+            timeval timeout{.tv_sec = timeout_us / 1000000, .tv_usec = timeout_us % 1000000};
+#endif
+            val.insert(val.end(), (char*)&timeout, (char*)&timeout + sizeof(timeout));
+            optlen = sizeof(timeout);
+            return ConvertReturnErrorCode(setsockopt(sock, level, optname_nat, val.data(), optlen));
```
StevenMiller123: Should be using native_level here, not level.

src/core/libraries/network/posix_sockets.cpp
```diff
@@ -281,16 +282,40 @@ int PosixSocket::SetSocketOptions(int level, int optname, const void* optval, u3
...
+        }
+        case ORBIS_NET_SO_ONESBCAST: {
+
+            if (optlen != sizeof(sockopt_so_onesbcast)) {
+                *Libraries::Kernel::__Error() = ORBIS_NET_ERROR_EFAULT;
+                return -1;
+            }
+            memcpy(&sockopt_so_onesbcast, optval, optlen);
+            return ConvertReturnErrorCode(
+                setsockopt(sock, level, SO_BROADCAST, (const char*)optval, optlen));
```
StevenMiller123: Same issue here, native_level instead of level.

StevenMiller123 (changes requested):
This fixes the issue I commented on with libSceHttp.

src/core/libraries/network/posix_sockets.cpp
georgemoralis: fixed

PR #3485, opened 2025-08-30
renderer_vulkan: Ignore blend parameters when disabled.

squidbus:
* When blending is disabled, ignore the configuration values instead of attempting to translate them.
* Add warning for unsupported scaled min/max blending.

Should fix cases of invalid blend factor/op assert.

### DIFF
--- a/src/video_core/renderer_vulkan/liverpool_to_vk.cpp
+++ b/src/video_core/renderer_vulkan/liverpool_to_vk.cpp
@@ -210,7 +210,7 @@ vk::BlendFactor BlendFactor(Liverpool::BlendControl::BlendFactor factor) {
     case BlendFactor::OneMinusConstantAlpha:
         return vk::BlendFactor::eOneMinusConstantAlpha;
     default:
-        UNREACHABLE();
+        UNREACHABLE_MSG("Unknown blend factor: {}", static_cast<u32>(factor));
     }
 }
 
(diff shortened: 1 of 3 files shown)


Randomuser8219:
The Sims 4 (CUSA09209) now gives this:
```
[Debug] <Critical> spirv_emit_context.cpp:662 operator(): Assertion Failed!
Dual source blending enabled, there must be exactly two MRT exports
```

squidbus:
Do you know what happens if you comment out that other assert?

Randomuser8219:
image
By using WinDBG, it works again

Randomuser8219:
[Debug] <Critical> position.h:67 ExportPosition: Unreachable code!
Unhandled output 6 on attribute Position1
This appeared when I tried to open Create-A-Sim though

squidbus:
> [Debug] position.h:67 ExportPosition: Unreachable code! Unhandled output 6 on attribute Position1

This is for viewport index, I believe implementing the export in the shader is simple but there's some extra viewport stuff with push constants that needs to support multiple as well.

squidbus:
@Randomuser8219 Had something else to try, could you check again with the new commit? If you hit the `ExportPosition` assert again that's expected.

Randomuser8219:
Once it compiles for Windows QT

Randomuser8219:
[Debug] <Critical> position.h:67 ExportPosition: Unreachable code!
Unhandled output 6 on attribute Position1
Same error on latest build, however I don't need WinDBG anymore to get past the copyright screen

squidbus:
Thanks, I'm a lot more satisfied with this fix so that's good.

PR #3487, opened 2025-08-30
Implement send/recvmsg

mikusp:


georgemoralis:
fixed windows(?)

src/core/libraries/network/posix_sockets.cpp
```diff
@@ -196,6 +222,32 @@ int PosixSocket::SendPacket(const void* msg, u32 len, int flags, const OrbisNetS
...
+#ifdef _WIN32
+    static LPFN_WSASENDMSG wsasendmsg = nullptr;
+    if (!wsasendmsg) {
+        GUID guid = WSAID_WSASENDMSG;
+        DWORD bytes = 0;
+        if (WSAIoctl(sock, SIO_GET_EXTENSION_FUNCTION_POINTER, &guid, sizeof(guid), &wsasendmsg,
+                     sizeof(wsasendmsg), &bytes, nullptr, nullptr) != 0) {
+            return ConvertReturnErrorCode(WSAGetLastError());
+        }
+    }
```
StevenMiller123: Is there a point in having this if?
Seems like it would always be true due to the line above.

mikusp: Not to mention it should be recvmsg 😄

StevenMiller123: That's a separate, but more important issue :smile:

StevenMiller123 (approved):
Code looks good to me, I haven't been able to test these Windows implementations yet though.

src/core/libraries/network/posix_sockets.cpp
```diff
@@ -183,6 +183,32 @@ int PosixSocket::Listen(int backlog) {
...
+                         &bytesSent, nullptr, nullptr);
+
+    if (res == SOCKET_ERROR) {
+        return ConvertReturnErrorCode(-1);
+    }
+    return static_cast<int>(bytesSent);
+#else
+    int res = sendmsg(sock, reinterpret_cast<const msghdr*>(msg), flags);
+#endif
+    return ConvertReturnErrorCode(res);
```
StevenMiller123: Minor nitpick here, but this return isn't reachable on Windows.
It could probably just be moved into the else block instead.

Same applies to the ReceiveMessage function too.

georgemoralis: done

PR #3490, opened 2025-08-31
semaphore: Fix determining wait status when canceled/deleted

squidbus:
Current logic for determining the result will always return `ETIMEDOUT` when not signaled, which causes the canceled and deleted statuses to get lost. In turn this can cause the main `Wait` function to erroneously try to remove the waiter from the wait list when it already has been and/or the semaphore has already been deleted.

Fix up the logic to make sure that these statuses are not ignored. Also, clean up some types to follow our standard here and fix a typo.

### DIFF
--- a/src/core/libraries/kernel/threads/semaphore.cpp
+++ b/src/core/libraries/kernel/threads/semaphore.cpp
@@ -19,7 +19,7 @@
 
 namespace Libraries::Kernel {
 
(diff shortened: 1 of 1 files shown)


kalaposfos13:
This PR fixes random semaphore crashes in LittleBigPlanet 3.

Issue #2508, opened 2025-02-23
[Feature Request]: Move the Qt plugin folders into QtPlugins on Windows


RuskyDed:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

Just a minor suggestion to place all Qt plugin folders in a dedicated one

![Image](https://github.com/user-attachments/assets/91aa0853-e9d0-4453-a323-de8f2fd9ee1b)
![Image](https://github.com/user-attachments/assets/ef052776-77f9-4096-ac83-cf86ceafd01e)

### Reason

For convenience

### Examples

In some popular emulators with Qt-based interfaces, the folder structure is organized in a same way. 
Dolphin, PCSX2:

![Image](https://github.com/user-attachments/assets/2a16bce1-86f0-4f94-b522-a4059242322a)
![Image](https://github.com/user-attachments/assets/633c78c5-f5e3-4f78-a7cd-b403d9b27063)

VeraPhobia:
a serious, how is it more convenient? like why would anyone even go and look at those folders? and a counter argument i guess would be wouldn't having less folders and files make it easier to find the ones we need? like log, user or save folders?

ghost:
ok my bad i made a mistake with reading you post in the first time. sorry.

squidbus:
Implemented in https://github.com/shadps4-emu/shadPS4/pull/3491

PR #3491, opened 2025-08-31
Move QT plugins to the qtplugins folder on Windows QT builds

UltraDaCat:
This addresses issue [#2508](https://github.com/shadps4-emu/shadPS4/issues/2508).

Here's a before and after comparison:
image
image


.github/workflows/build.yml
```diff
@@ -141,8 +141,17 @@ jobs:
...
         move build/shadPS4.exe upload
+        cp qt.conf upload/qt.conf
         windeployqt --no-compiler-runtime --no-system-d3d-compiler --no-system-dxc-compiler --dir upload upload/shadPS4.exe
+        move upload/generic upload/qtplugins
+        move upload/imageformats upload/qtplugins
+        move upload/multimedia upload/qtplugins
+        move upload/networkinformation upload/qtplugins
+        move upload/platforms upload/qtplugins
+        move upload/styles upload/qtplugins
+        move upload/tls upload/qtplugins
```
squidbus: According to the [docs](https://doc.qt.io/qt-6/windows-deployment.html), `windeployqt` should have a `--plugindir` option for where to output the plugins, have you tried using that instead of manually moving them all after?

qt.conf
```diff
@@ -0,0 +1,5 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
```
squidbus: Should probably be in the `dist` folder. Also it's possible to not need to add the license comments here if you add the file to the `REUSE.toml` with the license information (see some examples in that file).

.github/workflows/build.yml
UltraDaCat: Made that change, will have to wait until windows-qt is finished compiling to see if I messed something up or not.

qt.conf
UltraDaCat: Done

squidbus (approved):
Looks okay to me, if you can confirm that the latest version works fine on Windows still (since I can't test it right now).

UltraDaCat:
It works for me with both fresh and existing installs of shadPS4 on Windows

Issue #1292, opened 2024-10-08
log: Add ability to just disable it (by setting checkbox/arguments/config/whatever)


cvertee:
I know there is a trick to "disable" it by setting Filter to `*:Critical`
Maybe it'll be better to add some checkbox (?) that does this trick automatically?

Or maybe disables this log completely, like this in i guess `Impl::Instance().PushEntry()`
```c
if (!somethingEnabled) return;
```
with this no checking for filter, no checking for logtype, CPU can just chill here

cvertee:
Maybe that would require too much of workarounds. Idk maybe if its easy then yeah

Hermiten:
Could be a good idea, maybe someone interested to do that ?

k3dr1:
I've implemented this locally by adding a ``logEnabled`` boolean value in the config and checking against it in ``PushEntry`` and it works. However, I'm not sure whether letting users simply disable logging will hinder debuggability. If this is the right way of doing it, let me know and I'll open a PR when/if my other PR gets merged.

![Image](https://github.com/user-attachments/assets/139159b2-d282-4e64-bd8f-6f9400054a6b)

![Image](https://github.com/user-attachments/assets/201d03b2-2f69-4876-9fc7-21ee35c10c47)

![Image](https://github.com/user-attachments/assets/7aa34347-d307-4b1e-8974-73fb970e0ab9)

Hermiten:
If I understand, something with a menu or a checkbox to add filter easily
[] *:Debug[] *:Error [] *:Critical , things like this ?

topsex88:
ر

squidbus:
Added in https://github.com/shadps4-emu/shadPS4/pull/3493

PR #3493, opened 2025-08-31
Toggle to enable/disable logging

UltraDaCat:
This addresses issue [#1292](https://github.com/shadps4-emu/shadPS4/issues/1292).

### DIFF
--- a/src/qt_gui/settings_dialog.cpp
+++ b/src/qt_gui/settings_dialog.cpp
@@ -458,7 +458,7 @@ void SettingsDialog::LoadValuesFromConfig() {
         ifs.exceptions(std::ifstream::failbit | std::ifstream::badbit);
         const toml::value data = toml::parse(userdir / "config.toml");
     } catch (std::exception& ex) {
-        fmt::print("Got exception trying to load config file. Exception: {}n", ex.what());
+        fmt::print("Got exception trying to load config file. Exception: {}\n", ex.what());
         return;
     }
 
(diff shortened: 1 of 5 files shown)


src/qt_gui/settings_dialog.cpp
```diff
@@ -457,7 +458,7 @@ void SettingsDialog::LoadValuesFromConfig() {
         ifs.exceptions(std::ifstream::failbit | std::ifstream::badbit);
         const toml::value data = toml::parse(userdir / "config.toml");
     } catch (std::exception& ex) {
-        fmt::print("Got exception trying to load config file. Exception: {}\n", ex.what());
+        fmt::print("Got exception trying to load config file. Exception: {}n", ex.what());
```
StevenMiller123: Was this intentional?

UltraDaCat: It was not

Missake212:
Might be a bit of a nitpick but wouldn't it be better to have this in the logger tab instead of the experimental one? Other than that it works as expected (to be honest even the separate log toggle should be in the logger tab imo so it's not that big of a deal).

image

DanielSvoboda:
I think these options are better all together there in Logger
image

UltraDaCat:
Pushed a commit that should move those 2

Missake212:
Looks perfect to me UI wise.

image

Hermiten:
This doesn't adresse the issue #1292, it ask to put some checkbox or something to filter easily the log

StevenMiller123:
Based on the issue name and description, all @cvertee requests in #1292 is an option to disable logging, which this PR does provide.

Issue #2553, opened 2025-02-27
[Feature Request]: use the cover in the update if available


ghost:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

some updates bring new covers for the game, but chadps4 only uses the covers from the base game and not the updates

### Reason

i guess consistency?

ghost:
oh i forgot to say, it is the same behavior for the background image as well

imnltsa:
Do any games have new title music in their updates? Would be something to note as well if so. (`sce_sys/snd0.at9`)

ghost:
this game doesn't have any music even for the main game so i cant say for sure, i don't have any other game with this problem to check as well

ghost:
ok wait i copied the music file from 2 different game to the main game and the update each, it plays the main game

imnltsa:
> ok wait i copied the music file from 2 different game to the main game and the update each, it plays the main game

That should be expected behavior if no game updates contain new title music; <span title="God, I miss rePatch.">the emulator shouldn't really care for title music from updates if they aren't to be expected in the first place.</span>

I feel like if any game were to change title music, it would be Minecraft. I know they did that whole background image change back on the PS3 with v1.81(?), dunno if that was in line with any Minecraft PS4 Edition update but probably a good starting point if any for anyone who'd wanna perfect shadPS4 down to a T lol

StevenMiller123:
Should be resolved with #3494

PR #3494, opened 2025-08-31
Use icons, music and background images of updates if available

UltraDaCat:
This addresses issue [#2553](https://github.com/shadps4-emu/shadPS4/issues/2553).
-UPDATE folders take priority, followed by -patch and then base game

### DIFF
--- a/src/qt_gui/game_info.h
+++ b/src/qt_gui/game_info.h
@@ -19,13 +19,16 @@ public:
     QVector<GameInfo> m_games;
     QVector<GameInfo> m_games_backup;
 
-    static void SceUpdateChecker(std::string sceItem, std::filesystem::path& gameItem,
+    static void SceUpdateChecker(const std::string sceItem, std::filesystem::path& gameItem,
                                  std::filesystem::path& update_folder,
-                                 std::filesystem::path& patch_folder) {
+                                 std::filesystem::path& patch_folder,
+                                 std::filesystem::path& game_folder) {
         if (std::filesystem::exists(update_folder / "sce_sys" / sceItem)) {
             gameItem = update_folder / "sce_sys" / sceItem;
         } else if (std::filesystem::exists(patch_folder / "sce_sys" / sceItem)) {
             gameItem = patch_folder / "sce_sys" / sceItem;
+        } else {
+            gameItem = game_folder / "sce_sys" / sceItem;
         }
     }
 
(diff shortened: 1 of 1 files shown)


Missake212:
Works as expected for icons and backgrounds for me, would be nice to have someone test a game which change the title music, I don't think I have one.

Silver-Skyy:
i have replaced the main files with the updates of my games so i cant test it but i think you can try it by music file of another game with the update one?

src/qt_gui/game_info.h
```diff
@@ -41,15 +41,36 @@ class GameInfoClass : public QObject {
...
             }
         }
-
         PSF psf;
         if (psf.Open(sce_folder_path)) {
-            game.icon_path = game.path / "sce_sys" / "icon0.png";
+            std::filesystem::path game_update_path = filePath;
+            game_update_path += "-UPDATE";
+            std::filesystem::path game_patch_path = filePath;
+            game_patch_path += "-patch";
```
squidbus: I think you should consolidate this with the path logic above here for `param.sfo`, so only define these paths once for this whole function and then find the `param.sfo` in the same way as you are for the other files.

src/qt_gui/game_info.h
```diff
@@ -41,15 +41,36 @@ class GameInfoClass : public QObject {
...
+            if (std::filesystem::exists(game_update_path / "sce_sys" / "icon0.png")) {
+                game.icon_path = game_update_path / "sce_sys" / "icon0.png";
+            } else if (std::filesystem::exists(game_patch_path / "sce_sys" / "icon0.png")) {
+                game.icon_path = game_patch_path / "sce_sys" / "icon0.png";
+            } else {
+                game.icon_path = game.path / "sce_sys" / "icon0.png";
+            }
```
squidbus: Could move this to a separate function or macro so we can just set each path with 1 function call instead of duplicating the checks for each path for every file.

UltraDaCat:
@squidbus I've made some changes, now icons, background images and sounds use the same game_update_path and game_patch_path paths as param.sfo and instead of doing if else statements every time, now the function is being called with the needed parameters. I opted out for using 2 separate paths for -UPDATE and -patch folders due to it being easier to use imo. I can change it back to just 1 path for both if you'd like. When building locally it worked well on my end, but I would like if someone else tested this too tho.

src/qt_gui/game_info.h
```diff
@@ -32,24 +42,21 @@ class GameInfoClass : public QObject {
...
-        } else {
-            game_update_path = filePath;
-            game_update_path += "-patch";
-            if (std::filesystem::exists(game_update_path / "sce_sys" / "param.sfo")) {
-                sce_folder_path = game_update_path / "sce_sys" / "param.sfo";
-            }
-        }
+        std::filesystem::path game_patch_path = filePath;
+        game_patch_path += "-patch";
+        SceUpdateChecker("param.sfo", sce_folder_path, game_update_path, game_patch_path);
```
squidbus: I think to further reduce duplication for these cases you could put the base game path in the function as well.

So basically the paths would look something like this:
```
std::filesystem::path param_sfo_path = GetSceSysPath("param.sfo", game.path, game_update_path, game_patch_path);
std::filesystem::path icon0_path = GetSceSysPath("icon0.png", game.path, game_update_path, game_patch_path);
...
```

UltraDaCat: what do you think of me doing something like this:

std::filesystem::path game_update_path = filePath.string() + "-UPDATE";
std::filesystem::path game_patch_path = filePath.string() + "-patch";

this would bring the line count from 97 to 95

squidbus: I recall there being issues using `.string()` if certain languages or other characters are in the path so it would need special care.

My main concern with the review comments is just to keep the duplication down resolving the `sce_sys` paths and the file names.

squidbus (approved):

UltraDaCat:
Now that clang is fixed, it should be ready to merge

PR #3502, opened 2025-09-01
config: Add present mode option. 

squidbus:
Adds a config option to control the present mode. Available options are Mailbox (default), Fifo, and Immediate. If the selected mode is not supported, we fall back to FIFO as the Vulkan spec guarantees it is supported.

screenshot

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -79,6 +79,7 @@ static bool shouldPatchShaders = false;
 static u32 vblankDivider = 1;
 static bool isFullscreen = false;
 static std::string fullscreenMode = "Windowed";
+static std::string presentMode = "Mailbox";
 static bool isHDRAllowed = false;
 
 // Vulkan
@@ -192,6 +193,10 @@ bool getIsFullscreen() {
 
 std::string getFullscreenMode() {
     return fullscreenMode;
+}
+
+std::string getPresentMode() {
+    return presentMode;
 }
 
 bool getisTrophyPopupDisabled() {
@@ -466,6 +471,10 @@ void setFullscreenMode(std::string mode) {
     fullscreenMode = mode;
 }
 
+void setPresentMode(std::string mode) {
+    presentMode = mode;
+}
+
 void setisTrophyPopupDisabled(bool disable) {
     isTrophyPopupDisabled = disable;
 }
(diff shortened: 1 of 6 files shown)


squidbus:
Updated to allow specifying more present modes.

RuskyDed:
Working as expected on my end.

Missake212:
I think it would be nice to have a small explanation about what each mode does at the bottom if possible.

Solitair69:
Could you maybe include shadps4-emu/shadps4-qtlauncher#51

squidbus:
> Could you maybe include shadps4-emu/shadps4-qtlauncher#51

No, this is a separate feature entirely.

squidbus:
> I think it would be nice to have a small explanation about what each mode does at the bottom if possible.

Added:
screenshot

UltraDaCat:
On Windows, the description doesn't fit in the box properly, making the window bigger horizontally helps, but a part of the description remains out of view
image

StevenMiller123:
A scroll bar is visible in the description box, maybe see if that scroll bar works?

squidbus:
> On Windows, the description doesn't fit in the box properly, making the window bigger horizontally helps, but a part of the description remains out of view

I don't really know how to make it more concise than that while still fully explaining it, I think we would need to expand the box.

> A scroll bar is visible in the description box, maybe see if that scroll bar works?

Given that you have to hover the setting, I don't think that will work that well.

StevenMiller123:
Ah, good to know 😅

squidbus:
I'm not able to test on Windows right now by the way, so if someone could figure out a good size for the box to not scroll and adjust it, would be appreciated.

GHU7924:
>I'm not able to test on Windows right now by the way, so if someone could figure out a good size for the box to not scroll and adjust it, would be appreciated.

Previously, this box with explanations contained 3 lines, I think, later I asked to increase the number to 5 lines, because there were also descriptions that did not fit and on Steam Deck this changed the size of the window in game mode, which was very inconvenient. I do not know how this box might look if the text is translated into other languages, but conditionally you can try as a test to increase the number of lines to 7, and expand the window itself horizontally by 4 cm, for example, but all this needs to be checked.

>Given that you have to hover the setting, I don't think that will work that well.

Yes, if you remove the cursor from the setting, the description will disappear.

Issue #2756, opened 2025-04-07
[Feature Request]: Add the ability to configure FSR through the Qt "Graphics" tab.


brad0demx:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

With current releases, toggling and configuring FSR is only possible through the video debug information. I believe a more user friendly option would be to include these options within the Graphics tab of the emulator. 

### Reason

Opening the video debug information to disable this is less intuitive and in some games causes framerate instabilities while the menu is open. 

### Examples

Proposed UI change of the Graphics tab
![Image](https://github.com/user-attachments/assets/d401acf2-3872-4ed6-b049-123cd2b81f9f)

MajorP93:
Instead it may be a good idea to call the menu option "upscaling" and make FSR one of it's options.
For now (until) more methods get added the options could be "disabled" and "FSR".

Got already proposed here: https://github.com/shadps4-emu/shadPS4/issues/1303

brad0demx:
Yeah I was originally thinking something like a dropdown box to toggle between different upscaling methods, like what is mentioned in the issue you tagged. Since there are no other upscaling methods currently, I decided to just mirror the layout and contents of the menu from the developer menu.

squidbus:
Resolved by https://github.com/shadps4-emu/shadPS4/pull/3504

PR #2624, opened 2025-03-08
FidelityFX FSR implementation

viniciuslrangel:
Dive into the 4k res

Ref: https://github.com/GPUOpen-Effects/FidelityFX-FSR

PR #3504, opened 2025-09-01
Qt: Add FSR options to settings GUI

rainmakerv3:
Adds the 3 FSR options to the settings dialog for better user discoverability.

image

### DIFF
--- a/src/core/devtools/options.cpp
+++ b/src/core/devtools/options.cpp
@@ -5,6 +5,7 @@
 
 #include <memory>
 #include <imgui.h>
+#include "common/config.h"
 
 #include "video_core/renderer_vulkan/vk_presenter.h"
 
@@ -30,26 +31,17 @@ void LoadOptionsConfig(const char* line) {
         Options.frame_dump_render_on_collapse = i != 0;
         return;
     }
-    if (sscanf(line, "fsr_enabled=%d", &i) == 1) {
-        presenter->GetFsrSettingsRef().enable = i != 0;
-        return;
-    }
-    if (sscanf(line, "fsr_rcas_enabled=%d", &i) == 1) {
-        presenter->GetFsrSettingsRef().use_rcas = i != 0;
-        return;
-    }
-    if (sscanf(line, "fsr_rcas_attenuation=%f", &f) == 1) {
-        presenter->GetFsrSettingsRef().rcas_attenuation = f;
-    }
+
+    presenter->GetFsrSettingsRef().enable = Config::getFsrEnabled();
+    presenter->GetFsrSettingsRef().use_rcas = Config::getRcasEnabled();
+    presenter->GetFsrSettingsRef().rcas_attenuation =
+        static_cast<float>(Config::getRcasAttenuation() / 1000.f);
 }
 
 void SerializeOptionsConfig(ImGuiTextBuffer* buf) {
     buf->appendf("disassembler_cli_isa=%s\n", Options.disassembler_cli_isa.c_str());
     buf->appendf("disassembler_cli_spv=%s\n", Options.disassembler_cli_spv.c_str());
     buf->appendf("frame_dump_render_on_collapse=%d\n", Options.frame_dump_render_on_collapse);
-    buf->appendf("fsr_enabled=%d\n", presenter->GetFsrSettingsRef().enable);
-    buf->appendf("fsr_rcas_enabled=%d\n", presenter->GetFsrSettingsRef().use_rcas);
-    buf->appendf("fsr_rcas_attenuation=%f\n", presenter->GetFsrSettingsRef().rcas_attenuation);
 }
 
 } // namespace Core::Devtools
--- a/src/core/devtools/options.cpp
+++ b/src/core/devtools/options.cpp
@@ -5,7 +5,6 @@
 
 #include <memory>
 #include <imgui.h>
-#include "common/config.h"
 
 #include "video_core/renderer_vulkan/vk_presenter.h"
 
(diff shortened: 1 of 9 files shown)


src/qt_gui/settings_dialog.cpp
```diff
@@ -978,4 +987,102 @@ void SettingsDialog::setDefaultValues() {
...
+
+    ui->FSRCheckBox->setChecked(true);
+    ui->RCASCheckBox->setChecked(true);
+    ui->RCASSlider->setValue(250);
+    SaveFsrSettings();
+}
+
+void SettingsDialog::SaveFsrSettings() {
+    std::filesystem::path DevSettingsFile =
+        Common::FS::GetUserPath(Common::FS::PathType::UserDir) / "imgui.ini";
```
squidbus: I’m not so sure about having to introduce modifying the Imgui settings here, can we maybe move it to the main config?

rainmakerv3: I remember an old PR https://github.com/shadps4-emu/shadPS4/pull/2652, @viniciuslrangel wanted to keep the FSR settings in that file if I understand correctly.

squidbus: If you’re referring to the review comment there, I believe it’s just saying that the parameters should still be passed into and read by the FSR pass using its settings structure, not that it has to be in the imgui settings file.

rainmakerv3: ah I see. Ok, I can create new config entries and pass that to settings.rcas then

rainmakerv3: just to be sure, is it okay to remove fsr from the imgui settings then?

viniciuslrangel: Yeah, the last PR was mixing both config files, feel free to remove completely from imgui.ini

src/core/devtools/options.cpp
```diff
@@ -30,26 +31,17 @@ void LoadOptionsConfig(const char* line) {
...
-        return;
-    }
-    if (sscanf(line, "fsr_rcas_attenuation=%f", &f) == 1) {
-        presenter->GetFsrSettingsRef().rcas_attenuation = f;
-    }
+
+    presenter->GetFsrSettingsRef().enable = Config::getFsrEnabled();
+    presenter->GetFsrSettingsRef().use_rcas = Config::getRcasEnabled();
+    presenter->GetFsrSettingsRef().rcas_attenuation =
+        static_cast<float>(Config::getRcasAttenuation() / 1000.f);
```
squidbus: This should probably be moved since it's not part of dev tools config anymore, @viniciuslrangel would you have any preference on this?

rainmakerv3: how about here? (devtoools/layer.cpp)

image

squidbus: I think it would make sense to load it in the constructor for Presenter or something along those lines

rainmakerv3: yeah this seems to work fine. i assume i have to pass it before fsr_pass.create

image

viniciuslrangel: Yeah. But currently is does not update real-time, and it should be to have a quick feedback

edit: Just saw the menu bar still updates the setting directly. It kinda works, but isn't saving back to the config tho

src/core/devtools/options.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include <memory>
 #include <imgui.h>
+#include "common/config.h"
```
squidbus: This seems to be left over from previous commits, can be removed.

src/core/devtools/options.cpp
rainmakerv3: for imgui, i just added a button to save fsr settings back to config

for qt, added connect functions to update fsr settings in real time (only when a game is running)

src/qt_gui/settings_dialog.cpp
```diff
@@ -963,6 +1001,13 @@ void SettingsDialog::SyncRealTimeWidgetstoConfig() {
 
         Config::setAllGameInstallDirs(settings_install_dirs_config);
     }
+
+    if (GameRunning) {
```
squidbus: Couldn't these `GameRunning` checks just check if `presenter` is not null?

rainmakerv3: yup, kinda feel silly now for doing it how i did before lol. changed in latest commit

squidbus (approved):

brad0demx:
Resolves #2756

Silver-Skyy:
sorry to crash here just wanted to say thank you and,,,,, wait there was an FSR option all along?

StevenMiller123:
Yes, though the option was previously hidden away in the dev ui/imgui config.

Missake212:
Sorry I'm a bit late and I didn't test earlier but wouldn't it be better to have this off by default and let the user chose to enable it or not?

squidbus:
> Sorry I'm a bit late and I didn't test earlier but wouldn't it be better to have this off by default and let the user chose to enable it or not?

I wouldn't be opposed to changing it but I think it makes sense as is given it's already been default on.

wantija:
When was FSR added as default? I Would assume users would want the original image quality without the use of cheats/patches (I do).

StevenMiller123:
FSR has defaulted to on since #2624 was merged, for users who have the window larger than their game resolution.
Given how many people seem to be unaware that it's on, it's not that big a deal.

PR #3509, opened 2025-09-03
Log presets feature

w1naenator:
- Add “Load Presets…” button in Logger → Log Filter (settings_dialog.ui)
- Implement LogPresetsDialog (table: Comment, Filter)
- Support add (+), multi-remove (−), load via button or double-click
- Persist presets via QSettings at logger_presets/entries (qt_ui.ini)
- Wire button to open dialog and set logFilterLineEdit
- Update CMakeLists.txt to include new dialog sources

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1097,6 +1097,8 @@ set(QT_GUI src/qt_gui/about_dialog.cpp
            src/qt_gui/kbm_help_dialog.h
            src/qt_gui/main_window_themes.cpp
            src/qt_gui/main_window_themes.h
+           src/qt_gui/log_presets_dialog.cpp
+           src/qt_gui/log_presets_dialog.h
            src/qt_gui/settings_dialog.cpp
            src/qt_gui/settings_dialog.h
            src/qt_gui/settings_dialog.ui
--- a/src/qt_gui/log_presets_dialog.cpp
+++ b/src/qt_gui/log_presets_dialog.cpp
@@ -4,6 +4,7 @@
 #include "log_presets_dialog.h"
 
 #include <algorithm>
+#include <QCheckBox>
 #include <QHBoxLayout>
 #include <QHeaderView>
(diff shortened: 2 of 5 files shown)


w1naenator:
Way to manage log filter.

Screenshot_20250904-012602 (1)

GHU7924:
As a regular user, I don't understand much about filters, because I simply don't know what and how to write correctly. 

This PR is very good, but it would be great if there were already ready-made presets, although I don't know if there is a full list or some of the most useful ones. These are just my thoughts from the point of view of a regular user and I may be wrong.

We might also need button `Reset` for `Log Filter`, the small button to the right of button `Load Presets...`
Yes, I understand that it is not difficult to manually clear the filter field, but I think that since we can load a preset, we can also reset it, that is, clear the filter field.

At the moment it works like this:

I added filter `*Critical` (just to check the addition of information), loaded it, it appeared in line Log Filter. If you then press button `Close` and open the window again, there will be nothing in Log Filter, but if you pressed button `Save` or `Apply` instead of `Close`, the line will contain your filter. This filter will remain in the Log Filter until you clear it manually, even if you clear the entire preset list (yes i understand that Presets and Log Filter are different things, but i'm just saying how it works now). So, right now `Log Filter` depends on the `Save`, `Close`, `Apply` buttons in the settings window itself.

Again, I say this from the perspective of a regular user who may not use this setting unless asked, but it may be useful.

w1naenator:
This needed for developers only to skip unneeded log records, regular user can simply disable log.

w1naenator:
Screenshot_20250904-235800_1
Screenshot_20250904-235752_1
Screenshot_20250904-235739_1
Screenshot_20250904-235729_1

w1naenator:
Screenshot_20250905-005902_1
Screenshot_20250905-005857_1
Screenshot_20250905-005852_1

w1naenator:
I think it is usable now

PR #3511, opened 2025-09-04
SaveData fixes IXXXXXX

georgemoralis:


viniciuslrangel (approved):

Missake212:
Fixes the loading issue in Valkyria Revolution, also tested Bloodborne and Velocity 2X (Velocity 2X had issues with save files a while ago so I wanted to check if this didn't regress it) and both of these games save files still work, didn't test more than that though.

image

georgemoralis:
well fix only applies if save_data_dir and titleId specified in SFO is different . i don't think that is the case for 99% of the games

PR #3513, opened 2025-09-04
Add configurable extra memory

kalaposfos13:
This setting allows you to increase the available direct memory for games, similarly to what isNeo and isDevKit do, except here you can control the amount of extra memory given instead of only being able to add set amounts. This setting is ideally used only in game specific configs, due to the nature of increased direct memory size adversely affecting some games.


src/common/config.h
```diff
@@ -111,6 +111,8 @@ bool isDevKitConsole();
 void setDevKitConsole(bool enable, bool is_game_specific = false);
 
 bool vkValidationGpuEnabled(); // no set
+int GetExtraDmemInMbytes();
+void SetExtraDmemInMbytes(int value);
```
squidbus: nit: Lower case first letter in these to match the rest of config.

squidbus (commented):
Not sure if this needs updates for the recent memory changes. Also can’t seem to see why the Mac build failed as it expired, I assume maybe it was when we had the MVK build issue?

georgemoralis:
@kalaposfos13 is it ready? Fix it or close it

kalaposfos13:
With the latest push, going above the hardcoded memory limit (that was previously the maximum memory a PS4 Pro devkit model would have) now correctly expands the amount of memory shadPS4 will use, and won't crash in initialization code anymore. I tested a few games, and games still work fine with 0 MB-s of extra dmem, and 4K Bloodborne was also working fine, even with 8 GB of extra direct memory allocated. Still, just in case I missed something, I'd prefer if our resident memory expert @StevenMiller123 checked the code out before it gets merged.

Missake212:
2990 seems to be the maximum I can go up to before getting an error.

``[Debug] <Critical> address_space.cpp:182 operator(): Assertion Failed!``
``MapViewOfFile3 failed. Access is denied.``

[CUSA03173.log](https://github.com/user-attachments/files/22582258/CUSA03173.log)

Missake212:
Works now, even managed to boot up 6K (8K doesn't work sadly).

Screenshot 2025-09-28 131254

src/core/memory.cpp
```diff
@@ -14,6 +14,8 @@
 
 namespace Core {
 
+u64 total_memory_to_use = ORBIS_KERNEL_TOTAL_MEM_DEV_PRO;
```
StevenMiller123: Is there a reason to have this variable at all?
From what I can see, it's the same value as `total_size`

StevenMiller123 (commented):
Just a minor nitpick, otherwise code looks decent.
It would be nice if this setting could be restricted to game-specific configs, but I'm not entirely sure how possible that is since I haven't really looked at that implementation.

StevenMiller123 (approved):
Assuming this all works properly, I think it's fine.

PR #3514, opened 2025-09-04
Open a dummy audio input device if none is present in the system

mikusp:
Allows L.A. Noire to progress past intros

### DIFF
--- a/src/core/libraries/audio/sdl_in.cpp
+++ b/src/core/libraries/audio/sdl_in.cpp
@@ -48,23 +48,27 @@ int SDLAudioIn::AudioInOpen(int type, uint32_t samples_num, uint32_t freq, uint3
             std::string micDevStr = Config::getMicDevice();
             uint32_t devId;
 
+            bool nullDevice = false;
             if (micDevStr == "None") {
-                return ORBIS_AUDIO_IN_ERROR_INVALID_PORT;
+                nullDevice = true;
             } else if (micDevStr == "Default Device") {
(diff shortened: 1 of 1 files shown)


src/core/libraries/audio/sdl_in.cpp
```diff
@@ -48,7 +50,7 @@ int SDLAudioIn::AudioInOpen(int type, uint32_t samples_num, uint32_t freq, uint3
             std::string micDevStr = Config::getMicDevice();
             uint32_t devId;
 
-            if (micDevStr == "None") {
+            if (micDevStr == "None" && dummyDeviceOpened) {
```
squidbus: Is there a reason to only open once? I thought you could open many ports to the same device

mikusp: Not really, I think the behaviour seen in la noire is a bug in fmod that do not check the return value, and faking only the first call is enough

squidbus: I think if we need to fake devices for `None` I'd rather we just fake all of them. So like:
```
if (micDevStr == "None") {
    return id + 1;
}
```

mikusp: It does work for the rest of devices in the config as is

squidbus: I mean fake all ports if the device is `None`, instead of adding a static dummy flag and only faking one port.

PR #3516, opened 2025-09-04
Log SIGBUS signals

kalaposfos13:
Currently, if SIGBUS is hit, it'll just get silently skipped, which is not desirable if for example a game enters into an infinite loop of raising SIGBUS on a single instruction. I wasn't sure if I should make this an unreachable though, but given that so far no one complained about randomly hitting this while using a debugger, it might be the better choice, especially given that all three cases I've this this being hit the games either crashed anyway immediately after or hung. It should be mentioned that out of the three cases, two was messing around with things that changed memory allocations such as isNeo, and the third time was on PS4 Linux

### DIFF
--- a/src/core/signals.cpp
+++ b/src/core/signals.cpp
@@ -76,8 +76,10 @@ static void SignalHandler(int sig, siginfo_t* info, void* raw_context) {
     auto* code_address = Common::GetRip(raw_context);
 
     switch (sig) {
-    case SIGSEGV:
-    case SIGBUS: {
+    case SIGBUS:
+        LOG_ERROR(Debug, "SIGBUS in thread '{}' at code address {}", GetThreadName(),
+                  fmt::ptr(code_address));
+    case SIGSEGV: {
         const bool is_write = Common::IsWriteError(raw_context);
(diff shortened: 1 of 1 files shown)


squidbus:
I see two problems with this:
* This signal handler already uses `UNREACHABLE` if the signal was not from accessing GPU-tracked memory or from a unpatched & patchable instruction. Neither of these cases should log due to performance reasons.
* Some OS use `SIGBUS` for the above mentioned cases so you will be logging errors on every single GPU-tracked page signal.

kalaposfos13:
Is there any way to differentiate SIGBUS coming from the GPU side and the CPU side? My examples were all ones that happened right after booting and before graphics were initialized

squidbus:
> Is there any way to differentiate SIGBUS coming from the GPU side and the CPU side? My examples were all ones that happened right after booting and before graphics were initialized

Check through to the signal handlers in `cpu_patches` and `page_manager`, they should return `false` for cases that don't apply to them which would then fall through to the `UNREACHABLE`. If a case is applicable for those handlers it becomes very difficult to distinguish.

georgemoralis:
is that pr needed ?

kalaposfos13:
I guess not

PR #3524, opened 2025-09-05
rtc: Fix date parsing detection.

squidbus:
Fix time format detection in `sceRtcParseDateTime` by changing the format key index to 19.

* For RFC3339, index of 22 for the timezone specifier assumed a milliseconds component, which may not be present. We can use 19 instead, where the timezone indicator is when there is no milliseconds component, and add the `.` to the keys list for when there are milliseconds.
* For RFC2822, there will also be a `:` at index 19 so there is no change needed with the new index.

### DIFF
--- a/src/core/libraries/rtc/rtc.cpp
+++ b/src/core/libraries/rtc/rtc.cpp
@@ -748,10 +748,10 @@ int PS4_SYSV_ABI sceRtcParseDateTime(OrbisRtcTick* pTickUtc, const char* pszDate
 
     std::string dateTimeString = std::string(pszDateTime);
 
-    char formatKey = dateTimeString[22];
+    char formatKey = dateTimeString[19];
     OrbisRtcDateTime dateTime;
 
-    if (formatKey == 'Z' || formatKey == '-' || formatKey == '+') {
+    if (formatKey == 'Z' || formatKey == '-' || formatKey == '+' || formatKey == '.') {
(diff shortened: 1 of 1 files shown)


src/core/libraries/rtc/rtc.cpp
```diff
@@ -748,10 +748,10 @@ int PS4_SYSV_ABI sceRtcParseDateTime(OrbisRtcTick* pTickUtc, const char* pszDate
 
     std::string dateTimeString = std::string(pszDateTime);
 
-    char formatKey = dateTimeString[22];
+    char formatKey = dateTimeString[19];
```
kalaposfos13, line 751: This just replaces one hard-coded value with another, as far as I know, both are valid according to RFC-3339, so ideally we'd have to check both, no?

squidbus: No, because I added the `.` to the key list, which will be there if there is an `RFC3339` date with a milliseconds component.

kalaposfos13: Ah I see, it's all right then

squidbus: I will say, while I think this should be okay for detecting which format it is, ultimately we should probably use a proper parsing library for the actual parsing sometime down the line as our own rolled implementation feels a bit rough currently.

Issue #2168, opened 2025-01-17
[Feature Request]: Arbitrary framerates


birthtothunder:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

You can currently only cap framerate in integer multiples. This would allow you to set the framerate cap to whatever you want.

### Reason

I tried using external software (goverlay/Mangohud) to cap the framerate which caused it to play stuttery/suboptimally. I assumed this was an issue with Linux/goverlay but someone else also reported this as an issue on Windows/RTSS.
With cap
![Image](https://github.com/user-attachments/assets/8d7a580f-48d2-47e4-936f-a017b93daf87)
Without cap
![Image](https://github.com/user-attachments/assets/14511489-f767-48f4-8a0f-b41b6b318720)

### Examples

RPCS3 
![Image](https://github.com/user-attachments/assets/f90cd895-ab34-4304-810a-4b10e76c7bf1)

squidbus:
Implemented in https://github.com/shadps4-emu/shadPS4/pull/3532

PR #3527, opened 2025-09-05
Allow for custom framerate caps

UltraDaCat:
This addresses issue [#2168](https://github.com/shadps4-emu/shadPS4/issues/2168). To adjust the FPS limit you need to change it through the config file. The default value is 60. Vblank takes priority, so if you want to cap a 60FPS game to 144 you'll also need to increase Vblank to at least 3. QT stuff was scrapped due to the planned QT launcher separation.

### DIFF
--- a/src/core/devtools/widget/frame_graph.cpp
+++ b/src/core/devtools/widget/frame_graph.cpp
@@ -13,6 +13,7 @@ using namespace ImGui;
 
 namespace Core::Devtools::Widget {
 
+constexpr float TARGET_FPS = 60.0f;
 constexpr float BAR_WIDTH_MULT = 1.4f;
 constexpr float BAR_HEIGHT_MULT = 1.25f;
 constexpr float FRAME_GRAPH_PADDING_Y = 3.0f;
@@ -30,8 +31,8 @@ void FrameGraph::DrawFrameGraph() {
     }
 
     float target_formula;
-    if ((Config::vblankDiv() * 60) < Config::getFpsLimit()) {
-        target_formula = 60.0f * (float)Config::vblankDiv();
+    if ((Config::vblankDiv() * TARGET_FPS) < Config::getFpsLimit()) {
+        target_formula = TARGET_FPS * (float)Config::vblankDiv();
     } else {
         target_formula = (float)Config::getFpsLimit();
     }
--- a/src/core/devtools/widget/frame_graph.cpp
+++ b/src/core/devtools/widget/frame_graph.cpp
@@ -13,7 +13,6 @@ using namespace ImGui;
 
 namespace Core::Devtools::Widget {
 
-constexpr float TARGET_FPS = 60.0f;
 constexpr float BAR_WIDTH_MULT = 1.4f;
 constexpr float BAR_HEIGHT_MULT = 1.25f;
 constexpr float FRAME_GRAPH_PADDING_Y = 3.0f;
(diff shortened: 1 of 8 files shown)


Missake212:
I don't know if that's just me but I think that having two different options for FPS is kinda counter intuitive, I think that having one setting for both would be better (and to be completely honest I feel like having a frame limiter built in is pretty useless since you can cap externally anyway), I thought either modifying the current Vblank divider and change it to something like RPCS3 does, or have you frame limiter use a logic like this (without having to manually change the vblank value):

if over 60 fps > vblank set to 2
if over 120 fps > vblank set to 3

and so on would be better in my opinion, not sure what's the general opinion on this though.

src/core/devtools/widget/frame_graph.cpp
```diff
@@ -30,7 +29,13 @@ void FrameGraph::DrawFrameGraph() {
         return;
     }
 
-    float target_dt = 1.0f / (TARGET_FPS * (float)Config::vblankDiv());
+    float target_formula;
+    if ((Config::vblankDiv() * 60) < Config::getFpsLimit()) {
+        target_formula = 60.0f * (float)Config::vblankDiv();
```
StevenMiller123: For purposes like these, it's generally preferred to keep the constant to improve readability (so swap these hardcoded 60's for `TARGET_FPS`)

StevenMiller123 (changes requested):
Minor code nitpick, but I find it weird having two options that essentially do the same thing.
I personally think the vblank option should be removed in favor of this, but not sure how others may feel.

Also if you're updating these files, it wouldn't hurt to update the copyright date too. Not a requirement, but it would be nice.

src/core/libraries/videoout/driver.cpp
```diff
@@ -269,13 +270,18 @@ void VideoOutDriver::SubmitFlipInternal(VideoOutPort* port, s32 index, s64 flip_
...
-    static constexpr std::chrono::nanoseconds VblankPeriod{16666667};
-    const auto vblank_period = VblankPeriod / Config::vblankDiv();
+    int fps_cap_value;
+    if ((Config::vblankDiv() * 60) <= Config::getFpsLimit()) {
+        fps_cap_value = 16666667 / Config::vblankDiv();
+    } else {
+        fps_cap_value = 1000000000 / Config::getFpsLimit();
+        isFpsLimited = true;
+    }
+    const std::chrono::nanoseconds FpsCap{fps_cap_value};
```
squidbus, line 280: I don't think these variables should be renamed, and I also don't see a reason to reorder and make the base period non-constexpr.

src/core/libraries/videoout/driver.cpp
```diff
@@ -298,20 +304,36 @@ void VideoOutDriver::PresentThread(std::stop_token token) {
...
+                DrawLastFrame();
+            }
+        };
+
+        auto endFrame = [&](const auto request) {
+            Flip(request);
+            FRAME_END;
+        };
+
+        auto setFramerateCap = [&](bool limitCheck) {
```
squidbus, line 321: I don't think there's a need to pull these out into lambdas, you could just keep the code the same and add `isFpsLimited` to the `if` conditions.

Although I'm not clear what adding this flag is for.

squidbus (commented):
Other general comment: couldn't we just make the vblank divider a decimal value?

UltraDaCat:
Closed in favor of PR [#3532](https://github.com/shadps4-emu/shadPS4/pull/3532).

PR #3531, opened 2025-09-05
minor adjustments to the interface and translation

DanielSvoboda:
*This is a draft so I can adjust the text if it isn't correct(or improvement), as it's not my native language.

These texts will no longer be translated:
```
L1,L2,L3,R1,R2,R3
0.250
100%
```
*Do not include 'Options' as there is disagreement about this, so each language must decide, so there will still be a translation...

These descriptions have been added:
```
Open Log Location:
Open the folder where the log file is saved.

Microphone:
None: Does not use the microphone.
Default Device: Will use the default device defined in the system.
Or manually choose the microphone to be used from the list.

Volume:
Adjust volume for games on a global level, range goes from 0-500% with the default being 100%.

Default tab when opening settings:
Choose which tab will open, the default is General.

Show Game Size In List:
There is the size of the game in the list.

Enable Motion Controls:
When enabled it will use the controller's motion control if supported.
```

The debug screen has been revamped, as the log options have been added to a new tab.
| Before | After |
|-------------|-------------|
image | image

image

Some text on these screens has been centered
| Before | After |
|-------------|-------------|
image | image

| Before | After |
|-------------|-------------|
image | image

Values ​​have been separated from the text
image

RGB is now written in full and the values ​​are separated to the other side
image

### DIFF
--- a/src/qt_gui/control_settings.cpp
+++ b/src/qt_gui/control_settings.cpp
@@ -95,22 +95,19 @@ ControlSettings::ControlSettings(std::shared_ptr<GameInfoClass> game_info_get, b
 
     connect(ui->RSlider, &QSlider::valueChanged, this, [this](int value) {
         QString RedValue = QString("%1").arg(value, 3, 10, QChar('0'));
-        QString RValue = tr("R:") + " " + RedValue;
-        ui->RLabel->setText(RValue);
+        ui->RLabel->setText(RedValue);
         UpdateLightbarColor();
     });
 
     connect(ui->GSlider, &QSlider::valueChanged, this, [this](int value) {
         QString GreenValue = QString("%1").arg(value, 3, 10, QChar('0'));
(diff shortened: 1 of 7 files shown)


GHU7924:
I have already suggested these amendments, maybe they will be included in this PR, if everyone is satisfied.

config2

I would personally remove the mouse from the keyboard image and just add a line saying that the mouse can also be assigned controls, but just make sure to enable option `Mouse to Joystick` and `F7` so that you have mouse movement controls available.

On tab `Log`, it might be worth changing the position of the elements somehow so that it looks harmonious.

Otherwise, very good changes. @DanielSvoboda 👍

DanielSvoboda:
@GHU7924 I adjusted the Log's position, it looks a little better...
and I put the values ​​separately as you said.

image
image

GHU7924:
Yes, it looks better.

PR #3532, opened 2025-09-05
Replace Vblank Divider with Vblank Frequency

UltraDaCat:
This should address issue [#2168](https://github.com/shadps4-emu/shadPS4/issues/2168). Replaces Vblank Divider with Vblank Frequency. Vblank Frequency set to 60Hz is the equivalent to Vblank Divider set to 1. This allows you to set framerate limits to numbers such as 144, 100, 90, etc. It's a replacement for pull request [#3527](https://github.com/shadps4-emu/shadPS4/pull/3527)


src/core/libraries/videoout/driver.cpp
```diff
@@ -269,8 +269,8 @@ void VideoOutDriver::SubmitFlipInternal(VideoOutPort* port, s32 index, s64 flip_
 }
 
 void VideoOutDriver::PresentThread(std::stop_token token) {
-    static constexpr std::chrono::nanoseconds VblankPeriod{16666667};
-    const auto vblank_period = VblankPeriod / Config::vblankDiv();
+    static constexpr std::chrono::nanoseconds VblankPeriod{1000000000};
+    const auto vblank_period = VblankPeriod / Config::vblankFreq();
```
squidbus:
```suggestion
    const std::chrono::nanoseconds vblank_period(1000000000 / Config::vblankFreq());
```

Since `VblankPeriod` is no longer actually a vblank period, just the number of nanoseconds in a second.

squidbus (approved):

red-prig:
On the one hand, it is of course good that it is configurable, on the other hand, the user can break some games that clearly rely on the frequency of the vblank timer, just as another point for incorrect reports

squidbus:
> On the one hand, it is of course good that it is configurable, on the other hand, the user can break some games that clearly rely on the frequency of the vblank timer, just as another point for incorrect reports

This is already true of the vblankDivider setting.

red-prig:
by the way, what about fliprate? It turns out it will divide the custom frequency into parts?

georgemoralis:
this needs to be update to latest main before merging

PR #3533, opened 2025-09-05
Qt: Add GUI for game-specific settings

### LINKED ISSUE #3512: [Feature Request]: Add the per game config option to the GUI
### Checklist

### Description

Like most other emulators, have an option to right click and change the settings for a specific game by going off the default settings, for example if I want to have Readbacks enabled but only for Driveclub, I could right click Driveclub, create a custom config file and only enable Readbacks for this game.

### Reason

Currently the feature has been implemented but without being included in the QT GUI, meaning that you have to browse the emulator files and create them yourself if you want to use it, adding them to the GUI would make this process easier and more intuitive in my opinion. I believe most users won't bother browsing the emulator's files to make a custom config for their games, and I even think that in this state most people will miss the fact that this feature has been implemented, I think adding it to the GUI at some point is mandatory to facilitate the user experience.

### Examples

PCSX2:

Image

Image

RPCS3:

Image

Image

rainmakerv3:
Closes https://github.com/shadps4-emu/shadPS4/issues/3512

This incredibly tedious pain of PR adds a GUI for configuration of game-specific settings, accessible for the right-click context menu.

Quick initial tests are ok, but probably needs a lot of testing since this touches config files a lot. (You may want to backup your configs before testing, there's high potential for error here lol)

It's also just kinda ugly since all the non game-specific UI elements were hidden. In the future maybe we can do it like RPCS3 and group all the game-specific and non game-specific stuff in the same tabs.

image
 
EDIT: Summary of changes

1) Added game-specific gui options from context menu, this will open the settings dialog with all the global-only settings hidden and will show the game-specific experimental options. All options are in a sub-menu and has separate entries to create/configure/delete game configs

2) save and restore values functions changed to save/set default only game-specific values if called from the game-specific GUI

3) Added icons in the GUI to denote whether a game in the list/grid has a custom config

4) Added logging for isDevkit, and also if custom game config exists

5) Added the other experimental settings only to the game-specific settings GUI


GHU7924:
Frankly, I don't like it, although you put a lot of effort into this PR. Don't be offended, I'll explain why now.

Initially, when I saw issue 3512, I, like you, thought of copying a settings window similar to yours, but I decided that it was better to avoid duplication, because some general emulator settings are not needed here, but some specific ones are, so I made a sketch based on the cheats and patches window. I even thought that I should somehow transfer the changed configurations for the gamepad and keyboard here, because as a player I have to customize the game for myself, for this I need to customize the controls (keyboard or gamepad), someone might find cheats, patches, activation of certain functions useful for the correct operation of the game, say, only a certain game requires me to activate the `Readbacks` or you want to run some game in PS4 Pro mode, because the game allows it, for example.

I'm not saying that my sketches are better, maybe the bar is too high or they are not quite correct, but in your PR there is a lot of unnecessary stuff for setting up, say, a specific game, in my opinion.

This is my personal opinion, but it is normal that our views on this can be opposite.

I also think you should wait for PR 3531.

P.S. By the way, it turns out that the configuration file is created, but it can only be deleted manually, as far as I understand.

bigol83:
Rpcs3 does this, Pcsx2 does this too. The game configuration windows are almost exactly the same as the general one, i see nothing wrong there.

rainmakerv3:
One reason why it's done like this (game-specific settings window is the same as the general one, just with some settings hidden) is to reduce maintenance burden. If you put game-specific settings in another window, you have edit the GUI for the global settings and also the new game-specific window whenever you add a new setting, so it's twice the work. I understand it might be better in terms of usability if you design something from the ground-up just for game-specific settings, but we also have to consider how it affects development.

GHU7924:
>Rpcs3 does this, Pcsx2 does this too. The game configuration windows are almost exactly the same as the general one, i see nothing wrong there.

That's the thing, these emulators have been developed for a long time and there are many different functions, but it already looks old-fashioned, cluttered and confusing. I think everything should be more neat, free and easy.

The games themselves on the original hardware do not particularly require any settings, except for the resolution, sound and something in the interface, and since this is an emulator, then here the comparison is more with a PC, where you need to configure something.

On the console you just launch the game and play, everything is done through the shell, and on the emulator it is equivalent to having several games installed on your PC and each one needs to be configured differently. You change the controls or use what is there, but it is already configured for you (on the PC), you essentially select a preset, select the graphics level, here instead you need to activate or deactivate some setting.

These emulators can be taken as an example, and even should be, but we need to move with the times, now everything should be simpler and more convenient, in some ways even minimalistic.

You open the settings and your eyes run in different directions, you get lost in what you need and what you don’t need to properly emulate the game on your computer (and everyone’s computer is different).

I think it's worth moving towards more user accessibility, both in terms of usability and understanding, at least in Qt, so that fewer questions arise.

GHU7924:
>One reason why it's done like this (game-specific settings window is the same as the general one, just with some settings hidden) is to reduce maintenance burden. If you put game-specific settings in another window, you have edit the GUI for the global settings and also the new game-specific window whenever you add a new setting, so it's twice the work. I understand it might be better in terms of usability if you design something from the ground-up just for game-specific settings, but we also have to consider how it affects development.

I don't know all the details, yes, but perhaps there is a way to separate general emulator settings and those that can be used as specific to a particular game. I don't know, but maybe there is a loophole to make it so that even with a separate window you don't have to duplicate the code. In any case, I have expressed my opinion on this issue, and it is up to the developers to decide what to do. If there is no way to do it differently, then okay, I have nothing against it, but I would like the emulator to be even cooler))

Missake212:
First thing I noticed is that the settings are set to default when going to the game specific settings, I know RPCS3 lets you have the choice on whether you want to use the default settings or your current settings but I think the second one makes more sense personally, also I'm not sure if you can just get rid of some tabs but the GUI one is kinda useless imo.

image

rainmakerv3:
> First thing I noticed is that the settings are set to default when going to the game specific settings, I know RPCS3 lets you have the choice on whether you want to use the default settings or your current settings but I think the second one makes more sense personally, also I'm not sure if you can just get rid of some tabs but the GUI one is kinda useless imo.
> 
> image

Yeah this would probably be better. I changed it to create game config from global settings instead (anyway it can easily be reset to default with the restore defaults button).

EDIT: you can hide the tabs (I hid the paths one since none of the items there are game-specific), but I think we can just leave the GUI in just in case someone may find it useful

Missake212:
Found another problem,in the emulator readbacks are off for me so I created a custom config for BB where I selected readbacks enabled, after that I created a custom config for The Last Guardian and in that custom config instead of having readbacks off they were on (checked the global config file and they were off too) so if you create a custom config after another custom config the second one reads from the first one. Restarting the emulator and creating a custom config will go back to reading the global config (I made sure to test this with all the settings so it's not just readbacks).

rainmakerv3:
> Found another problem,in the emulator readbacks are off for me so I created a custom config for BB where I selected readbacks enabled, after that I created a custom config for The Last Guardian and in that custom config instead of having readbacks off they were on (checked the global config file and they were off too) so if you create a custom config after another custom config the second one reads from the first one. Restarting the emulator and creating a custom config will go back to reading the global config (I made sure to test this with all the settings so it's not just readbacks).

Nice catch, latest commit should fix this, if you can test

Missake212:
Yeah that's fixed, I'll give it more testing later today but from a quick check it works as expected, every game I launched that had a modified custom config used it as intended.

GHU7924:
I looked at the latest commit and here are my thoughts:

1) We need to add an "Audio" tab for global and specific settings, where we will put the settings: "Volume" from the "General" tab and "Microphone" from the "Input" tab.

2) Leave only the following tabs for specific settings: "Graphics", "Input", "Log", "Debug". I don't know if we should leave the "User" tab, but let's say if a new "Audio" tab is added, then it should be added for specific settings too.

3) Maybe we should come up with something so that the game is marked with a special label that will notify the user that he currently has a certain configuration applied. (This is a kind of protection, because during testing or reporting someone can write what is happening with my game, and then it turns out that a certain setting was used, which can thus affect the game.)

Implementation: For example, it can be a small gear icon on the game icon, for example, in the upper right corner. You can also write to the Log, this is not necessary, just in case (yes, I know that Readbacks and such are noted in the Log).

4) Perhaps in the context menu it is worth creating an item "Game Profile" and then putting the creation of specific settings in it, so that a person does not accidentally create a specific setting that will work differently, provided that global settings can be changed, but they will not affect this game due to a specific setting.

5) I do not know all the experimental features, but perhaps in the "Debug" tab it is worth expanding them. Add DMA, PS4 Pro, etc.

rainmakerv3:
> I looked at the latest commit and here are my thoughts:
> 
> 1. We need to add an "Audio" tab for global and specific settings, where we will put the settings: "Volume" from the "General" tab and "Microphone" from the "Input" tab.
> 2. Leave only the following tabs for specific settings: "Graphics", "Input", "Log", "Debug". I don't know if we should leave the "User" tab, but let's say if a new "Audio" tab is added, then it should be added for specific settings too.
> 3. Maybe we should come up with something so that the game is marked with a special label that will notify the user that he currently has a certain configuration applied. (This is a kind of protection, because during testing or reporting someone can write what is happening with my game, and then it turns out that a certain setting was used, which can thus affect the game.)
> 
> Implementation: For example, it can be a small gear icon on the game icon, for example, in the upper right corner. You can also write to the Log, this is not necessary, just in case (yes, I know that Readbacks and such are noted in the Log).
> 
> 4. Perhaps in the context menu it is worth creating an item "Game Profile" and then putting the creation of specific settings in it, so that a person does not accidentally create a specific setting that will work differently, provided that global settings can be changed, but they will not affect this game due to a specific setting.
> 5. I do not know all the experimental features, but perhaps in the "Debug" tab it is worth expanding them. Add DMA, PS4 Pro, etc.

I added number 3 in the latest commit (implemented an icon to mark if the game has a custom config) as seen here:

image

image

I think I will leave 1/2 (re-arranging tabs) for a future PR as that requires a lot more discussion.

For number 5, the options that are not in the GUI are intentionally left out, the main developers would prefer those things not to be in the GUI yet (and putting them in the GUI is out of scope for this PR)

What do you mean by number 4, can you give an example?

GHU7924:
> I added number 3 in the latest commit (implemented an icon to mark if the game has a custom config) as seen here:
> image image

Looks great 👍 

> I think I will leave 1/2 (re-arranging tabs) for a future PR as that requires a lot more discussion.

It so happened that I didn't have time to propose such a change (I mean point 1) in PR 3531, since I thought about it much later, but the PR was already merged into Main. Point 2 should contain fewer settings, so that it only contains things that can actually affect the game.
 
> For number 5, the options that are not in the GUI are intentionally left out, the main developers would prefer those things not to be in the GUI yet (and putting them in the GUI is out of scope for this PR)

Ок

> What do you mean by number 4, can you give an example?



I thought this would provide additional protection against accidental creation of certain settings. For example, a user wanted to add something to favorites, but accidentally created a certain configuration, but if you did it like in the screenshot, you would have to intentionally activate this option. On the other hand, since you added a note that a certain setting was applied to the game, this may be redundant.

Missake212:
> > What do you mean by number 4, can you give an example?
> 
> 
> 
> I thought this would provide additional protection against accidental creation of certain settings. For example, a user wanted to add something to favorites, but accidentally created a certain configuration, but if you did it like in the screenshot, you would have to intentionally activate this option. On the other hand, since you added a note that a certain setting was applied to the game, this may be redundant.

I think this is a good change, down the road we will probably be able to have a "create game-specific settings from default settings" options like other emulators (and also have the delete game specific config there), also it will save a lot of space so I think that's for the better.

Missake212:
>For number 5, the options that are not in the GUI are intentionally left out, the main developers would prefer those things not to >be in the GUI yet (and putting them in the GUI is out of scope for this PR)

Considering what #3513 is trying to do, I think adding the option to have DMA, isDevKit ... available only in the custom config wouldn't be a bad thing, but best to have the opinion of one of the devs.

EDIT (don't want to comment 5 times in a row): I still believe the GUI tab is kind of useless for now, all the settings there don't impact the games in any way (and probably never will), other than that I don't think I've found any issue when trying out the PR so all that's left are the suggestions (and having a dev give us his opinion on DMA and other things).

GHU7924:
>For number 5, the options that are not in the GUI are intentionally left out, the main developers would prefer those things not to be in the GUI yet (and putting them in the GUI is out of scope for this PR)

If the developers approve, then we can do this:

rainmakerv3:
> > I added number 3 in the latest commit (implemented an icon to mark if the game has a custom config) as seen here:
> > image image
> 
> Looks great 👍
> 
> > I think I will leave 1/2 (re-arranging tabs) for a future PR as that requires a lot more discussion.
> 
> It so happened that I didn't have time to propose such a change (I mean point 1) in PR 3531, since I thought about it much later, but the PR was already merged into Main. Point 2 should contain fewer settings, so that it only contains things that can actually affect the game.
> 
> > For number 5, the options that are not in the GUI are intentionally left out, the main developers would prefer those things not to be in the GUI yet (and putting them in the GUI is out of scope for this PR)
> 
> Ок
> 
> > What do you mean by number 4, can you give an example?
> 
> 
> I thought this would provide additional protection against accidental creation of certain settings. For example, a user wanted to add something to favorites, but accidentally created a certain configuration, but if you did it like in the screenshot, you would have to intentionally activate this option. On the other hand, since you added a note that a certain setting was applied to the game, this may be redundant.



image


Added this submenu in the latest commit.

Regarding other possible experimental options in the game-specific settings (DMA, isDevKit), I'm personally ok with it since that section is clearly marked as experimental but I will wait for one of the main devs to weigh in. But that can also be another PR anyway if there's gonna be a lot of back and forth discussion for that, as I think this would be good to try and have this merged before the release.

For the GUI tab, there is a very slight use to selecting the home tabs on a per game basis. For instance if only on a certain game, you test readbacks on/off a lot, you can set home tab to debug only for that particular game. It's set-up an a game-specific config entry and someone out there might find it useful so I just kept it. 

Like I said though, in the future, I would like the tabs to be grouped into global items and game-specific items to avoid these things where a tab only has one or two elements (plus, that will be easier to maintain compared to having to manually hide individual elements in the game-specific GUI)

GHU7924:
>For the GUI tab, there is a very slight use to selecting the home tabs on a per game basis. For instance if only on a certain game, you test readbacks on/off a lot, you can set home tab to debug only for that particular game. It's set-up an a game-specific config entry and someone out there might find it useful so I just kept it.

I selected the Path tab as the main one, but it resulted in this:

изображение

Therefore, I personally do not see the point in keeping an `GUI` tab for specific settings, so I will not be at all against it if the `Graphics` tab becomes the main one for them, because the first thing a user usually sets is the screen resolution, and then everything else. It may not be as convenient, but for regular users it's a pretty reasonable solution, in my opinion.



Most likely, there will be more changes to the tabs themselves and their contents in the future, but I think that at this stage, as a first step, this solution is acceptable.

rainmakerv3:
> > For the GUI tab, there is a very slight use to selecting the home tabs on a per game basis. For instance if only on a certain game, you test readbacks on/off a lot, you can set home tab to debug only for that particular game. It's set-up an a game-specific config entry and someone out there might find it useful so I just kept it.
> 
> I selected the Path tab as the main one, but it resulted in this:
> 
> изображение
> Therefore, I personally do not see the point in keeping an `GUI` tab for specific settings, so I will not be at all against it if the `Graphics` tab becomes the main one for them, because the first thing a user usually sets is the screen resolution, and then everything else. It may not be as convenient, but for regular users it's a pretty reasonable solution, in my opinion.
> 
> 
> Most likely, there will be more changes to the tabs themselves and their contents in the future, but I think that at this stage, as a first step, this solution is acceptable.

The game-specific config doesn't load anything in the paths tab because none of those items are game-specific,  I just removed the option to select paths when on the game specific gui so this does not happen. I just think it's easier if we just include everything in the GUI that's set-up as game-specific to be part of the game-specific settings (anyway it can just be ignored) rather than have to deliberate if each individual setting should or should not be included.

Also added logging to check if game-specific config exists.

Anyway, as Stephen has voiced his agreement, I will add some experimental features to the game-specific GUI, but I think that should be enough alraedy for the scope of this PR after I do that hopefully by tomorrow. I'm concerned that this is already getting quite big to review.

rainmakerv3:
Found myself with a bit of time, so I went ahead with this:

image

That should do it for all the suggestions (except rearranging all the items and tabs, which I think should be in another PR as that might take a while and probably a bit of a long discussion)

GHU7924:
>The game-specific config doesn't load anything in the paths tab because none of those items are game-specific, I just removed the option to select paths when on the game specific gui so this does not happen.

Actually I just wanted to show with an example why we should abandon the option of choosing a default tab for specific settings))

>That should do it for all the suggestions (except rearranging all the items and tabs, which I think should be in another PR as that might take a while and probably a bit of a long discussion)

There were quite a few important changes made to this PR, so I agree that we can stop here and provide the PR for review. Good job 👍

georgemoralis:
if thats ready it should be merged before friday's feature freeze

Missake212:
Noticed a bug, when making a custom config for one game then launching another game (without a custom config), when you launch shad again the second game will have the second wheel like it has a custom config and you'll also get a custom config file in the folder, so even if you don't want a custom config you get one generated. In this example I made a custom config for Bloodborne, then launched 13 Sentinels without making a custom config and when I quit the game and launched the emu again it has the wheel and a custom config inside the folder

image

rainmakerv3:
> Noticed a bug, when making a custom config for one game then launching another game (without a custom config), when you launch shad again the second game will have the second wheel like it has a custom config and you'll also get a custom config file in the folder, so even if you don't want a custom config you get one generated. In this example I made a custom config for Bloodborne, then launched 13 Sentinels without making a custom config and when I quit the game and launched the emu again it has the wheel and a custom config inside the folder
> 
> image

Another nice catch, thanks. Last push should fix this if you can test again

Missake212:
I was about to say this looks perfect and that the only nitpick I have is that new settings don't have descriptions but they could be handled in another PR or something but I'll test the new push and see if that's also good.

Missake212:
Last push broke something again, games now read the last modified custom config, so modifying game A's config then launching game B will make game B read game A's config instead of the global one.

rainmakerv3:
> Last push broke something again, games now read the last modified custom config, so modifying game A's config then launching game B will make game B read game A's config instead of the global one.

Yeah I goofed the squash somehow LOL.

Anyhow, I just thought of how to refactor this anyway to make it a lot cleaner and remove all additional args that I used to have. This time I spent a bit more time testing so hopefully I didn't mess anything else up, but hoping you can test as well

Missake212:
Seems to work fine on my end but I just noticed when you launch a game without a config file it says this at the top of the log:

``Got exception trying to load config file. Exception: ios_base::failbit set: iostream stream error``

rainmakerv3:
> Seems to work fine on my end but I just noticed when you launch a game without a config file it says this at the top of the log:
> 
> `Got exception trying to load config file. Exception: ios_base::failbit set: iostream stream error`

I carelessly put a return command in the wrong place when I was re-adding it :despair:

Next push should fix that and shouldn't really affect anything else. But since I'm obviously pretty sloppy today (maybe cuz of my pretty intense work day lol) I'm gonna do some more testing tomorrow before saying this is ready, though so far it looks ok for me as well

GHU7924:
I haven't had time to test everything properly yet, but there is no fix for my problem yet.

What I meant was that in the global settings (config.toml) I have the "Path" tab selected, so even if a specific configuration is created, that selection is saved, which is why I suggested changing the selected tab for a specific configuration to something else. You can temporarily select the General tab.

Edit: More precisely, this happens when creating a certain configuration, when you restart these settings everything will be fine, but is it possible to somehow fix this so that users do not complain about it in the future?

rainmakerv3:
> I haven't had time to test everything properly yet, but there is no fix for my problem yet.
> 
> What I meant was that in the global settings (config.toml) I have the "Path" tab selected, so even if a specific configuration is created, that selection is saved, which is why I suggested changing the selected tab for a specific configuration to something else. You can temporarily select the General tab.
> 
> Edit: More precisely, this happens when creating a certain configuration, when you restart these settings everything will be fine, but is it possible to somehow fix this so that users do not complain about it in the future?

Ok, fixed this in latest push

GHU7924:
>Ok, fixed this in latest push

Unfortunately no. I provide you with a video of the previous commit (I apologize for the quality of the video, but even so you can understand what is happening on the screen):

https://github.com/user-attachments/assets/10527bf2-1b71-4367-9b00-123f9b0a94fe

The last commit affected global settings, which breaks everything.

https://github.com/user-attachments/assets/724e9a35-fadc-4f30-9718-7cc7a0223b4d

P.S. In my case, I would like to have the Path tab in the global settings, but when creating a specific configuration, the General tab was selected by default for that specific configuration. I understand that this is a copy, but it would be great if it were possible to replace the tab during creation. I'm going to state the obvious, but this applies to all default tabs in global settings because someone might have a different one selected.

GHU7924:
@DanielSvoboda if you can give a hint or study the code of this PR and also express your thoughts or suggestions, it would be very helpful.

rainmakerv3:
> > Ok, fixed this in latest push
> 
> Unfortunately no. I provide you with a video of the previous commit (I apologize for the quality of the video, but even so you can understand what is happening on the screen):
> 
>  2025-09-10.00-42-35.mp4 
> The last commit affected global settings, which breaks everything.
> 
>  2025-09-10.00-45-13.mp4 
> P.S. In my case, I would like to have the Path tab in the global settings, but when creating a specific configuration, the General tab was selected by default for that specific configuration. I understand that this is a copy, but it would be great if it were possible to replace the tab during creation. I'm going to state the obvious, but this applies to all default tabs in global settings because someone might have a different one selected.

Should really be fixed last push.

When creating a new game config, the general tab will always be selected now. But opening it after that will select the selected tab

GHU7924:
>Should really be fixed last push.

>When creating a new game config, the general tab will always be selected now. But opening it after that will select the selected tab

I would say that it works perfectly now 👍 

Personally, I haven't noticed any more problems, everything works as it should. If someone wants to test it in more detail, please do. In its current form, I am completely satisfied with both the interface and the work, all other improvements and changes would be more logical to make separately from this PR. I think this PR is ready for merging, if there are no problems with the code or other errors.

@rainmakerv3 , I give you a standing ovation, you did a great job 🥇

DanielSvoboda:
> @DanielSvoboda if you can give a hint or study the code of this PR and also express your thoughts or suggestions, it would be very helpful.

I haven't analyzed the code, but using it seems good to me, it's functional.

rainmakerv3:
I did a longer test session with everything I can think of and everything I tested works. Tests from Daniel, GHU, and Missake's earlier testing went ok, so I think this is ready now

GHU7924:
I have a question: are all flags displayed in the log? For example, since we have given the option to activate the PS4 Pro mode, it should be checked, as it can affect the game.

rainmakerv3:
> I have a question: are all flags displayed in the log? For example, since we have given the option to activate the PS4 Pro mode, it should be checked, as it can affect the game.

the ones added in the GS GUI are all logged, yes, now that I added the devkit logging

RuskyDed:
This seems to have broken the display mode setting in the global interface. It's always set to windowed, even if the file specifies a different value. The settings work correctly if you don't save anything though.

RuskyDed:
Thanks for your quick response

rainmakerv3:
Looks like you know this already, but this should be fixed by https://github.com/shadps4-emu/shadPS4/pull/3570

PR #3535, opened 2025-09-06
Libraries: Np libraries cleanup

StevenMiller123:
This PR:
- Moves all libSceNp* implementations into `src/core/libraries/np`. 
- Adds `np_error.h` and `np_types.h` for storing errors and types frequently seen in multiple Np libraries. 
- Removes the thousands of auto-stubs generated in libSceNpManager and libSceNpCommon, since most are used internally by VSH/other Np libs and will never need a proper implementation.
- Implements basic logic for handling NpRequests.
- Implements several basic libSceNpManager functions.
- Cleans up the implementation for the isPSNSignedIn setting.

This PR could use some testing, though I don't expect much to change in games.


Missake212:
No changes/regressions on my end.

PR #3476, opened 2025-08-29
Epoll

mikusp:
Initial net epoll implementations. Windows support waiting for @georgemoralis to fix a submodule lib

Things that are not used by anything in my library I've left as unreachables, if you hit those let me know. While testing please set connectedToNetwork to true in your config and enable Debug logs (ideally for everything, but at least for Lib.Net)

PR #3231, opened 2025-07-11
Network WIP

mikusp:
Only for Linux as of now as that's my usual dev environment

PR #2761, opened 2025-04-08
[Libs] Font lib HLE implementation

georgemoralis:
Just some functions atm , not useable

PR #3540, opened 2025-09-06
Core: Remove checks for symbol version_major and version_minor

StevenMiller123:
On top of the added complexity of needing to specify these numbers every time we generated a library, we've also run into multiple cases where homebrew that work on real hardware won't work on here because their imports use different versions over what actual firmware libraries may specify.

This PR fixes problems like this by removing the checks entirely, and removing the versions from the LIB_FUNCTION and LIB_OBJ macros. I've also removed the duplicate exports in libSceVrTracker and libSceVideoOut, since those won't be necessary if this is merged.

Based on my local tests, this shouldn't cause any regressions, and should fix the issue with libScePigletv2VSH.sprx that @kalaposfos13 ran into on the Discord server.

This will likely cause merge conflicts with #3476, #3290, #3259, #3231, and #2761


StevenMiller123:
Marked as draft since I need to figure out proper behavior for libSceSsl/libSceSsl2, the only difference between those libraries is the reported module major and minor versions (despite function implementations being different.

StevenMiller123:
Based on decompiling libSceSysmodule, I've determined that real hardware just never loads libSceSsl.
In case Sony ever changes that (or we encounter a game manually loading libSceSsl.sprx from `/sys/common/lib`), I've left the library HLE intact

PR #3541, opened 2025-09-06
shader_recompiler: Support PointSize and ViewportIndex attributes.

squidbus:
Adds support for the `PointSize` and `ViewportIndex` pre-rasterization attribute exports, with the appropriate capabilities set for each pre-rasterization stage.

Note that it asserts for `ViewportIndex` if clip space conversion in-shader is needed, as the push constants only fit the first viewport currently. Hopefully a rare case if hit at all.

Fixes `Unhandled output 1 on attribute Position1` and `Unhandled output 6 on attribute Position1` asserts.


StevenMiller123:
image
Fixes the regression in Sword Art Online: Lost Song (CUSA02823).

PR #3545, opened 2025-09-07
extend IPC functionalities

Fire-Cube:
This PR adds more IPC functionalities that will be required by the GUI separation

### DIFF
--- a/src/core/ipc/ipc.cpp
+++ b/src/core/ipc/ipc.cpp
@@ -6,7 +6,7 @@
 #include <iostream>
 #include <string>
 
-#include "SDL3/SDL.h"
+#include <SDL3/SDL.h>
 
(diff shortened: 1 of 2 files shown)


viniciuslrangel (approved):

src/core/ipc/ipc.cpp
```diff
@@ -6,9 +6,14 @@
 #include <iostream>
 #include <string>
 
+#include "SDL3/SDL.h"
```
squidbus: nit: Should use `<>` include style for this

squidbus (commented):
Other than very small comment looks good, is this intended to still be a draft?

Fire-Cube:
> Other than very small comment looks good, is this intended to still be a draft?

Can be merged

PR #3551, opened 2025-09-07
externals: Update libusb to fix more ABI issues.

squidbus:
Moved the ABI declaration from a cast when calling the callback, to using the existing `LIBUSB_CALL` define to put it everywhere.

This should hopefully fix the issue that appeared with Tekken 7 by making sure that libusb's internal callback is declared with the right ABI. It also makes sure than any other potential users of libusb (e.g. SDL) will inherit the right ABI for callbacks.


squidbus:
@Missake212 Please check with your issue

Missake212:
That fixes the regression.

image

PR #3557, opened 2025-09-08
amdgpu: Report GPU perf counter in GPU cycles.

squidbus:
Currently the perf counter from the GPU is reported as the plain TSC value from the CPU, but the perf counter is expected to be in GPU cycles. Instead, scale the value from the TSC frequency to the GPU frequency, using `sceGnmGetGpuCoreClockFrequency()` to make sure regular vs Neo mode is accounted for.

Fixes Valkyria Revolution (CUSA06810) slowing itself down due to incorrect GPU time measurements.


StevenMiller123:
No regressions to report on my end.

PR #3559, opened 2025-09-09
video_core: Address various UE bugs

raphaelthegreat:


src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
```diff
@@ -370,13 +370,14 @@ void EmitContext::DefineInputs() {
...
-        if (info.stores.Get(IR::Attribute::Depth)) {
-            frag_depth = DefineVariable(F32[1], spv::BuiltIn::FragDepth, spv::StorageClass::Output);
-        }
         if (info.loads.Get(IR::Attribute::IsFrontFace)) {
             front_facing =
                 DefineVariable(U1[1], spv::BuiltIn::FrontFacing, spv::StorageClass::Input);
         }
+        if (info.loads.GetAny(IR::Attribute::RenderTargetIndex)) {
+            output_layer = DefineVariable(S32[1], spv::BuiltIn::Layer, spv::StorageClass::Input);
+            Decorate(output_layer, spv::Decoration::Flat);
```
squidbus, line 379: Should we rename to just `layer` or something since its not longer just used for output?

src/shader_recompiler/backend/spirv/emit_spirv.cpp
```diff
@@ -305,6 +305,9 @@ void SetupCapabilities(const Info& info, const Profile& profile, const RuntimeIn
             runtime_info.fs_info.addr_flags.persp_sample_ena) {
             ctx.AddCapability(spv::Capability::SampleRateShading);
         }
+        if (info.has_layer_output) {
+            ctx.AddCapability(spv::Capability::Geometry);
```
squidbus, line 309: I'm not quite sure I understand this, fragment shaders can use the Geometry execution model?

src/shader_recompiler/ir/passes/shader_info_collection_pass.cpp
```diff
@@ -160,7 +160,8 @@ void CollectShaderInfoPass(IR::Program& program, const Profile& profile) {
         }
     }
 
-    if (info.stores.GetAny(IR::Attribute::RenderTargetIndex)) {
+    if (info.stores.GetAny(IR::Attribute::RenderTargetIndex) ||
+        info.loads.GetAny(IR::Attribute::RenderTargetIndex)) {
         info.has_layer_output = true;
```
squidbus: Same for this, should probably rename since its also input now.

src/video_core/amdgpu/pixel_format.h
```diff
@@ -102,6 +102,7 @@ enum class NumberConversion : u32 {
     Sint8ToSnormNz = 4,
     Sint16ToSnormNz = 5,
     Uint32ToUnorm = 6,
+    SrgbToNorm = 7,
```
squidbus, line 105: The order in these conversion op names is usually in the direction from the base format to what the shader sees, so I think this would be `SnormToSrgb` (or just `NormToSrgb` if the same code works for both sign types).

src/shader_recompiler/backend/spirv/emit_spirv.cpp
raphaelthegreat: Yeah I compiled a simple example with glslang and it added Geometry capability. Validation layers also complain if its missing
```
     OpCapability Geometry
%1 = OpExtInstImport "GLSL.std.450"
     OpMemoryModel Logical GLSL450
     OpEntryPoint Fragment %main "main" %color %gl_Layer
```

src/video_core/amdgpu/pixel_format.h
raphaelthegreat: Isn't this the case already however, since srgb is the stored format and the shader will "see" linear norm values

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
raphaelthegreat: Technically output could be interpreted as the output render target, which makes sense in either stage which is why I didn't rename it. But I can do this if you prefer

src/video_core/amdgpu/pixel_format.h
squidbus: It seems a bit confusing in this case. Normally for example the Uscaled conversion is a `Uint` Vulkan format converted to a `Uscaled` value in the shader, so it is `UintToUscaled`, and here we use `Snorm` on the Vulkan side. But I suppose you're right that in the end we're degammaing it and giving it to the shader as a linear value, so I guess it's fine?

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
squidbus: Sure, should probably at least rename the info flag to `has_output_layer` to make more sense then.

src/shader_recompiler/backend/spirv/emit_spirv.cpp
squidbus: That's a weird case. The Vulkan spec side says that `Fragment` execution model should be fine (https://registry.khronos.org/vulkan/specs/latest/man/html/Layer.html), but if validation is complaining, glslang does it, and this works, then no reason not to I guess.

raphaelthegreat: Here is the exact error for future reference
```
[Render.Vulkan] <Info> vk_pipeline_cache.cpp:502 CompileModule: Compiling fs shader 0xbf93e1b9a9b13256 
[Render.Vulkan] <Error> vk_platform.cpp:54 DebugUtilsCallback: VUID-VkShaderModuleCreateInfo-pCode-08737: vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
Operand 3 of Decorate requires one of these capabilities: Geometry ShaderLayer ShaderViewportIndexLayerEXT MeshShadingNV MeshShadingEXT 
  OpDecorate %gl_Layer BuiltIn Layer
The Vulkan spec states: If pCode is a pointer to SPIR-V code, pCode must adhere to the validation rules described by the Validation Rules within a Module section of the SPIR-V Environment appendix (https://docs.vulkan.org/spec/latest/chapters/shaders.html#VUID-VkShaderModuleCreateInfo-pCode-08737)
```

src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -285,6 +285,12 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
         scheduler.CurrentTick() - tex_cache_image.tick_accessed_last > NumFramesBeforeRemoval;
 
     if (image_info.guest_address == tex_cache_image.info.guest_address) { // Equal address
+        if (image_info.guest_size == tex_cache_image.info.guest_size &&
+            (image_info.type == AmdGpu::ImageType::Color3D ||
+             tex_cache_image.info.type == AmdGpu::ImageType::Color3D)) {
```
squidbus: Noticed the commit message for this change, if we're doing this for 2D array / 3D aliases shouldn't we check that one of them is 2D / 2D array?

There's `IsViewTypeCompatible` in `image_view.cpp` if you want to reuse some of that as well but for image types.

raphaelthegreat: Yeah I'm still not too happy with this, needs some extra checks and properly be merged into some other cases bellow. That path in general needs some cleanup (i.e tiling check could be removed in favor of checking array_mode when searching overlaps). Also might be better to split this into cases of >, =, < guest size and do actions depending on each one (copy, or make image view). Some of the code here is remnant from when this function only checked the size and not the more proper block dim/block size/tile mode checks that ensure proper compatibility

Missake212:
Cars 3 (not UE but suffers from black screen in game on main):

Nvidia
Screenshot 2025-09-09 141846

Intel
Screenshot 2025-09-09 141953


Tekken 7:

Nvidia
image

Intel
image

georgemoralis:
3d in ace combat

εικόνα

εικόνα

stops ingame with that stubs

[Lib.GnmDriver] <Error> gnmdriver.cpp:1817 sceGnmSetWaveLimitMultipliers: (STUBBED) called
[Lib.GnmDriver] <Error> gnmdriver.cpp:1817 sceGnmSetWaveLimitMultipliers: (STUBBED) called 

(never seen them before)

georgemoralis:
Killing floor 2 

εικόνα

εικόνα

εικόνα

georgemoralis:
Strike suit zero

εικόνα

georgemoralis:
resident evil revelations black screen fixed

εικόνα

Xcedf:
This PR fixes gow3 blocky lighting/shadows on nvidia
image
image

StevenMiller123:
DRAGON BALL FighterZ is now more visible, though there are stll clear rendering issues.
image
image
[CUSA09072.zip](https://github.com/user-attachments/files/22239297/CUSA09072.zip)

DRAGON BALL Z: KAKAROT now shows some rendering before crashing on a non-GPU memory assert.
image
[CUSA14655.log](https://github.com/user-attachments/files/22239301/CUSA14655.log)

KINGDOM HEARTS Ⅲ is rendering decently, but crashes silently before I can get to anything too complicated. Camera sensitivity is also broken.
image
[CUSA12031.zip](https://github.com/user-attachments/files/22239477/CUSA12031.zip)

Loading from a save reveals a severe "over-exposure" issue present regardless of whether that Bc6 Srgb test is applied. The test commit does seem to help with a discoloration issue though.
Current PR build:
image
Previous PR build:
image
[CUSA12031.zip](https://github.com/user-attachments/files/22239679/CUSA12031.zip)

KINGDOM HEARTS 0.2 renders pretty well ingame, but crashes frequently from internal out of memory asserts.
image
image
[CUSA05795.log](https://github.com/user-attachments/files/22239720/CUSA05795.log)

This PR regresses KNACK™
image
Main for reference:
image

StevenMiller123:
Persona 3 Reload now renders 3d graphics, though the colors seem to be off? The srgb commit seems to have no effect on this issue.
image
image

StevenMiller123:
This PR regresses Sakura Wars. It now crashes my GPU driver when trying to load ingame.
[CUSA16429.log](https://github.com/user-attachments/files/22239943/CUSA16429.log)
[cdl_dump.yaml](https://github.com/user-attachments/files/22239944/cdl_dump.yaml)

StevenMiller123:
SpongeBob SquarePants: Battle For Bikini Bottom - Rehydrated now renders 3d, though some parts of characters seem to be transparent.
image
image
[CUSA14898.log](https://github.com/user-attachments/files/22240032/CUSA14898.log)

StevenMiller123:
CODE VEIN now renders 3d, but character eyes seem to be broken. The game softlocks in character creation, so I can't test any further than this right now.
image

StevenMiller123:
SWORD ART ONLINE: FATAL BULLET now renders 3d, though visual effects are flickering, shadows are broken, and some textures appear to be corrupting. The game eventually crashed on `[Debug] <Critical> image.cpp:347 operator(): Assertion Failed!`
image
image
image
image

[CUSA10093.zip](https://github.com/user-attachments/files/22240486/CUSA10093.zip)

pl-k4:
Mailapp Now renders 3D but has a pretty bad rendering issue.
image
it also spams
```
[Render.Vulkan] <Error> im
age.cpp:161 Image: image format Bc5UnormBlock type 2D is not supported (flags { MutableFormat | BlockTexelViewCompatible | ExtendedUsage }, usage { TransferSrc | TransferDst | Sampled | Storage }) 
```

[shad_log.txt](https://github.com/user-attachments/files/22240713/shad_log.txt)

StevenMiller123:
Rendering is still completely broken in SWORD ART ONLINE Last Recollection and Tower of Fantasy.
image
[CUSA32810.log](https://github.com/user-attachments/files/22241207/CUSA32810.log)

image

[CUSA35049.log](https://github.com/user-attachments/files/22241283/CUSA35049.log)

georgemoralis:
broke ace combat title screen

εικόνα

georgemoralis:
but aircraft looks less buggy

εικόνα

georgemoralis:
Although i still get

[Lib.GnmDriver] gnmdriver.cpp:1817 sceGnmSetWaveLimitMultipliers: (STUBBED) called
[Lib.GnmDriver] gnmdriver.cpp:1817 sceGnmSetWaveLimitMultipliers: (STUBBED) called

That doesn't seem to be the problem (mostly) as it goes a few more frames before crash

εικόνα

StevenMiller123:
I don't think sceGnmSetWaveLimitMultipliers will pose much of a problem, as the function just runs an ioctl (as opposed to writing a PM4 command). Final Fantasy XV also spams it, but it's menus look fine on main

raphaelthegreat:
What commit broke the title screen?

georgemoralis:
this it seems as i tested the previous based on my screenshots ( [video_core: Use attachment feedback layout instead of general if poss…](https://github.com/shadps4-emu/shadPS4/pull/3559/commits/81fc32549dc71c7be9df9557303a6dca14fb857b)

georgemoralis:
re-run it and seem ok.. (driver issue?)

StevenMiller123:
Seems like the newest commit to this PR fixed blocky artifacts in LUA games?
| Main | ue-fixes |
|------|----------|
| image | image |

They now spam a ton of validation errors though.
[CUSA17112-ue.zip](https://github.com/user-attachments/files/22259721/CUSA17112-ue.zip)
[CUSA17112-main.log](https://github.com/user-attachments/files/22259722/CUSA17112-main.log)

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -755,7 +754,9 @@ void Rasterizer::BindTextures(const Shader::Info& stage, Shader::Backend::Bindin
             auto& image_view = texture_cache.FindTexture(image_id, desc);
 
             if (image.binding.force_general || image.binding.is_target) {
-                image.Transit(vk::ImageLayout::eGeneral,
+                image.Transit(instance.IsAttachmentFeedbackLoopLayoutSupported()
+                                  ? vk::ImageLayout::eAttachmentFeedbackLoopOptimalEXT
```
squidbus, line 758: You need to set `VK_IMAGE_USAGE_ATTACHMENT_FEEDBACK_LOOP_BIT_EXT` usage on the image to use this.

raphaelthegreat: Looks like that also needs a pipeline compile flag too hmm. Should I also make VK_EXT_attachment_feedback_loop_dynamic_state mandatory to use feedback loops at all

squidbus:
```
Feedback loop is enabled for an attachment if either of the following conditions are satisfied:
* The corresponding image is in the VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT layout
* For the attachment identified by VkRenderingAttachmentInfo::feedbackLoopEnable is VK_TRUE
```

~~Seems to me like it's an either/or from the spec?~~

I think I'm looking at the wrong thing but the pipeline flag doesn't seem to be mentioned in the feedback loop section.

raphaelthegreat: So probably need both pipeline flag and image usage flag seems like. The flag I'm referring to is VK_PIPELINE_CREATE_COLOR_ATTACHMENT_FEEDBACK_LOOP_BIT_EXT

squidbus: Probably should be dynamic then yes

raphaelthegreat: Newest commit should be a lot better. I'm a bit worried about performance impact of setting that usage flag on many color images, it should be checked ig

StevenMiller123:
Validation error spam is gone, and IxSHE Tell remains fixed.
image
[CUSA17112.log](https://github.com/user-attachments/files/22264656/CUSA17112.log)

StevenMiller123:
This PR fixes some graphical artifacting in Persona 5 Strikers.
| Main | `ue-fixes` |
|------|------------|
| image | image |

squidbus:
@StevenMiller123 Are the issues you found with Sakura Wars and Knack still present? (Also is it a specific GPU vendor only as I don't seem to have the Sakura Wars issue)

StevenMiller123:
@squidbus I can confirm Knack is still regressed, though @Randomuser8219 informed me that the issue is AMD-only.
I'll retest Sakura Wars.

StevenMiller123:
Sakura Wars appears to be working properly now, I can get ingame without any device lost crash now.

Knack is still visibly broken, here's a screenshot from the latest action of this PR.
image

[CUSA00068.zip](https://github.com/user-attachments/files/22265829/CUSA00068.zip)
[CUSA16429.log](https://github.com/user-attachments/files/22265831/CUSA16429.log)

StevenMiller123:
@squidbus Seems like Sakura Wars only works with readbacks enabled on this PR?
I'm consistently able to reproduce the device lost after disabling readbacks, which is a regression since main works with readbacks disabled.
[CUSA16429-no-readbacks.log](https://github.com/user-attachments/files/22266017/CUSA16429-no-readbacks.log)
[CUSA16429-readbacks.log](https://github.com/user-attachments/files/22266022/CUSA16429-readbacks.log)
[CUSA16429-main.log](https://github.com/user-attachments/files/22266033/CUSA16429-main.log)

squidbus:
> Seems like Sakura Wars only works with readbacks enabled on this PR? I'm consistently able to reproduce the device lost after disabling readbacks, which is a regression since main works with readbacks disabled.

Is this on any specific vendor(s) or all of them?

StevenMiller123:
I've only tested AMD so far, give me a bit and I can test Intel too.

RealSayyadina:
I can test on NVIDIA, be right back.

StevenMiller123:
Seems like it's safe to assume Intel isn't affected, I'm not getting a device lost, and while I'm not testing the same scene, the same shaders involved in the GPU-crashing submit seem to have compiled fine on here.

Randomuser8219:
IMG_1434
This PR allows Sackboy: A Big Adventure to (barely) show graphics on AMD. The pink screen is the title screen background.

RealSayyadina:
Sorry it took me so long, had to redump. NVIDIA hits an "unresolvable image overlap" error on the load screen after Sumire Kanzaki's introduction at the Grand Imperial Theater. Different error from what Steven hit. (Also possible my dump might be bad, I was having internet issues last night and my PS4 is...cranky...)
image

RealSayyadina:
> Sorry it took me so long, had to redump. NVIDIA hits an "unresolvable image overlap" error on the load screen after Sumire Kanzaki's introduction at the Grand Imperial Theater. Different error from what Steven hit. (Also possible my dump might be bad, I was having internet issues last night and my PS4 is...cranky...) image

(Clarification: This was with readbacks off, so Sakura Wars does run, it just hits a different error that's also a problem in games on main)

StevenMiller123:
That texture cache issue is present on current main too, if you're able to reach that then the crash I saw is AMD-only.

RealSayyadina:
Yeah, that's pretty much what I thought.

raphaelthegreat:
Latest push add various things
* More proper implementation of depth export that takes account Z_EXPORT_FORMAT and implements gl_SampleMask
* Fixed a bug where mul omod would be ignored with clamp was also present (might fix shader bugs in some cases)
* Sampling from MSAA images now behaves properly and isn't UB

georgemoralis:
i guess pr shouldn't be called UE bugs now :D

ElBread3:
Likely either due to the depth fix or the MSAA sampling fix, Orochi 3 engine games now render correctly

World of FF before:
woff-before
after:
woff-after
Kyoei Toshi (low fps is due to validation on btw) before:
kyoei-toshi-before
Kyoei Toshi after:
kyoei-toshi-after

RealSayyadina:
Ratchet & Clank (CUSA01047) now attempts to play the opening studio logos and crashes on the following error. Not actually sure if this is a regression or not, as it usually doesn't even boot for me. 
image

squidbus:
> Ratchet & Clank (CUSA01047) now attempts to play the opening studio logos and crashes on the following error. Not actually sure if this is a regression or not, as it usually doesn't even boot for me.

This is not a regression, it does that sometimes.

RealSayyadina:
Sakura Wars now gets past the previous "resolve overlap" crash, but performance past that point is absolutely horrendous, less than 10 FPS compared to a fairly stable 30 before it.

StevenMiller123:
Sakura Wars behavior was always a tad random, I would avoid claiming the issue is fixed unless you've performed a couple tests to make sure you didn't get lucky.

RealSayyadina:
> Sakura Wars behavior was always a tad random, I would avoid claiming the issue is fixed unless you've performed a couple tests to make sure you didn't get lucky.

You were right. I ran it several more times to test and found that it consistently gets past the crash if you skip the cutscene immediately before the loading screen where it happens, but crashes if you let the cutscene play. So, not fixed. Sorry.

StevenMiller123:
Metaphor: ReFantazio seems to suffer from a similar regression as KNACK (so probably AMD-only too).
image
[CUSA47038.log](https://github.com/user-attachments/files/22280926/CUSA47038.log)

Haven't spotted any differences from the new commits so far, and KNACK is still broken. I haven't tested Sakura Wars again just yet.

raphaelthegreat:
Can you check which commit caused it, would help a lot

StevenMiller123:
@raphaelthegreat Seems like the regression came from https://github.com/shadps4-emu/shadPS4/pull/3559/commits/10129005c34aa3bbeb228f1897fc133ad66c6fca

raphaelthegreat:
Can you try removing small pieces of the commit to pinpoint the exact change it in it

RainKikyou:
GUNDAM VERSUS can now render graphics
QQ图片20250912023433

StevenMiller123:
@raphaelthegreat Seems like Metaphor: ReFantazio broke from https://github.com/raphaelthegreat/shadPS4/blob/8aeb2937c959dd5512ab8d21ab15a0e2564af747/src/shader_recompiler/frontend/translate/translate.cpp#L177

raphaelthegreat:
If you comment out only this line does it fix it https://github.com/raphaelthegreat/shadPS4/blob/8aeb2937c959dd5512ab8d21ab15a0e2564af747/src/shader_recompiler/frontend/translate/translate.cpp#L180

StevenMiller123:
@raphaelthegreat Commenting that one out had no effect.

raphaelthegreat:
If you replace the set that broke it with the zero set from the highlighted line, what effect does it have?

StevenMiller123:
@raphaelthegreat That seems to fix the issue.
image

RainKikyou:
fatal frame 4 graphic fix
image

StevenMiller123:
All of the regressions I've reported so far are fixed now, I'll keep testing in case there's anything I've missed.

Missake212:
God of War III Remastered now hits the following UC on both Nvidia and Intel.

``[Debug] <Critical> emit_spirv_context_get_set.cpp:198 EmitGetAttributeU32: Unreachable code!``
``Read U32 attribute PackedAncillary``

[CUSA01715.log](https://github.com/user-attachments/files/22284626/CUSA01715.log)

Randomuser8219:
With the newest build, the blocky texture on the water in KNACK (CUSA00068) is fixed, but the game is having depth issues similar to how it looked last year.
image

StevenMiller123:
Seems like this depth issue in KNACK is exclusive to v1.00 of the game?
I'll see if I can figure out where it regressed.

Randomuser8219:
> Seems like this depth issue in KNACK is exclusive to v1.00 of the game? I'll see if I can figure out where it regressed.

It regressed in https://github.com/shadps4-emu/shadPS4/pull/3559/commits/8467ac88f3708d6b80612962511a8f158d04b8ed

StevenMiller123:
Yeah, seems like it.

StevenMiller123:
And it's specifically related to the depth export part of that commit too. A small change like modifying the last loop in `Translator::ExportDepth` to make it skip the depth attribute is enough to prevent the problem.

raphaelthegreat:
Depth exports weren't skipped before this commit either tho. @StevenMiller123 can you compare the shader that exports depth before and after this and see how depth exports changed and regressed

georgemoralis:
Tony hawk 5 (UE game ) seems playable

εικόνα

coolllman:
Knack (regression) now looks
Screenshot_20250913_022208_Moonlight
Nvidia 3060
Before pr
Screenshot_20250913_023415_Moonlight

Ratchet go playable)
Screenshot_20250913_023914_Moonlight

StevenMiller123:
This bug has already been reported, we're still debugging the causes.

PR #3561, opened 2025-09-09
Minimum Vblank Frequency of 60Hz

UltraDaCat:
Stops the Vblank Frequency from being below 60Hz.


rainmakerv3:
Can't we just set the minimum value of the spinbox to 60? That way the user can see that there's a minimum value

UltraDaCat:
The spinbox's minimum value is also changed to 60, the setVblankFreq part in emulator.cpp is used in case a user has already changed it to below 60Hz or if they try manually changing the value via the config(seeing as many users tend to just change stuff in the config without knowing what they do)

Missake212:
One thing I thought of was to use a slider instead of a spinbox to clearly show there's a minimum value in the UI but I thought this would be too much work and might look weird currently in the UI.

src/emulator.cpp
```diff
@@ -62,6 +62,10 @@ Emulator::Emulator() {
 Emulator::~Emulator() {}
 
 void Emulator::Run(std::filesystem::path file, const std::vector<std::string> args) {
+    if (Config::vblankFreq() < 60) {
+        Config::setVblankFreq(60);
+    }
```
squidbus: Think you could just put this in the `vblankFreq()` function itself, or config load/save.

src/qt_gui/settings_dialog.cpp
```diff
@@ -817,7 +817,7 @@ void SettingsDialog::updateNoteTextEdit(const QString& elementName) {
     } else if (elementName == "windowSizeGroupBox") {
         text = tr("Width/Height:\\nSets the size of the emulator window at launch, which can be resized during gameplay.\\nThis is different from the in-game resolution.");
     } else if (elementName == "heightDivider") {
-        text = tr("Vblank Frequency:\\nThe frame rate at which the emulator refreshes at (60hz is the baseline, whether the game runs at 30 or 60fps). Changing this may have adverse effects, such as increasing the game speed, or breaking critical game functionality that does not expect this to change!");
```
squidbus: Why is this removed?

Missake212: That was my recommendation, I added this to let users know there's no reason to go below 60hz so I told Ultra to remove it if we don't let users go below 60hz anyway.

squidbus: I think it is still worth noting in some way that setting it to some value will not automatically make the game run at that FPS.

UltraDaCat:
Should be fine now

PR #3565, opened 2025-09-10
Simple implementation of sceGnmUnmapComputeQueue in gnmdriver

zf06gaius:
just as the tile says, this is just a simple implementation of sceGnmUnmapComputeQueue
only known games to call this function is Uncharted 4 v1.00 according to Stephen on the discord, and a test renderer homebrew I was using to test iDTech2 3D rendering for later work on uHexenII porting


zf06gaius:
newest commit removes 2 comments I felt weren't proper since they mention specific people outside of Github or talk about a homebrew app nobody else other than me has as it's not public

squidbus (approved):

PR #3573, opened 2025-09-11
Qt: Add descriptions for new experimental settings in GUI

rainmakerv3:
Title. Very much open to suggestions, here's what I have now:

DMA:
Enable Direct Memory Access:
Enables arbitrary memory access from the GPU to CPU memory

Neo:
Enable PS4 Pro Mode:
Adds support for PS4 Pro emulation and memory size

Devkit:
Enable Devkit Console Mode:
Adds support for Devkit console memory size

PSN:
Set PSN Signed-in to True:
Forces games to detect an active PSN sign-in. Actual PSN capabilities are not supported

Network Connected:
Set Network Connected to True:
Forces games to detect an active network connection. Actual online capabilities are not yet supported

### DIFF
--- a/src/qt_gui/settings_dialog.cpp
+++ b/src/qt_gui/settings_dialog.cpp
@@ -952,7 +952,7 @@ void SettingsDialog::updateNoteTextEdit(const QString& elementName) {
     } else if (elementName == "dmaCheckBox") {
         text = tr("Enable Direct Memory Access:\\nEnables arbitrary memory access from the GPU to CPU memory.");
     } else if (elementName == "neoCheckBox") {
-        text = tr("Enable PS4 Pro Mode:\\nAdds support for PS4 Pro emulation and memory size.");
+        text = tr("Enable PS4 Neo Mode:\\nAdds support for PS4 Pro emulation and memory size. Currently causes instability in a large number of tested games.");
     } else if (elementName == "devkitCheckBox") {
         text = tr("Enable Devkit Console Mode:\\nAdds support for Devkit console memory size.");
     } else if (elementName == "networkConnectedCheckBox") {
--- a/src/qt_gui/settings_dialog.cpp
+++ b/src/qt_gui/settings_dialog.cpp
@@ -522,6 +522,11 @@ SettingsDialog::SettingsDialog(std::shared_ptr<gui_settings> gui_settings,
         ui->readbacksCheckBox->installEventFilter(this);
         ui->readbackLinearImagesCheckBox->installEventFilter(this);
         ui->dumpShadersCheckBox->installEventFilter(this);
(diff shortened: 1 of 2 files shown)


StevenMiller123:
EDIT: Disregard this comment, clearly I need sleep before I do any reviewing 😅 

I'm still personally against adding PS4 Pro mode to the GUI, but I am curious what other devs think about this.

In my time in the shadPS4 community, back when the PS4 Pro mode setting was originally in the GUI, we ran into people enabling the setting to enable PS4 Pro-related enhancements in their games. They would then complain in help because of some type of crash caused by some PS4 Pro-specific behavior we don't handle properly yet, or from the more complex shaders some games might toss out with the setting enabled. While we originally ran into many people struggling to disable the setting after we removed it from the GUI, this problem has died out in the time since.

In my testing, while our emulation has drastically improved, many of these Pro-specific issues still remain, so I fear we might run into the same issues all over again (but with the larger community we have now).

The rest of the settings I don't mind, DMA is necessary for some games, and is at this point more stable than readbacks. isDevKit doesn't cause too many issues so long as you avoid other dmem size-related hacks. PSN signed-in does cause trouble, but only in a select few games. isNetworkConnected was already a mostly harmless setting anyway, since it only blocks one of ways games connect to the internet.

Missake212:
Not sure if you know but on current nightly ps4Pro is in the GUI, you can only access it if you actually create a custom config so all this PR does is give a description to those settings "hidden" in the custom config GUI.

StevenMiller123:
I don't mind it with custom configs, since people creating those will already be used to tinkering with settings to improve their experience

StevenMiller123:
Oh, I see now, I misunderstood the pr 🤦

kalaposfos13:
If we rename ps4pro to ps4neo in the GUI, people will probably be even less likely to turn it on without knowing what it actually does, since if they don't know what it means at a first glance, they're more likely to read the tooltip for it

GHU7924:
I won't speak for the correctness and clarity of the descriptions, you can do it in such a way that it will be clear only to the developers, but for PS4 Pro I agree with kalaposfos13, so I suggest doing it this way:

>Enable PS4 Pro Mode:

Enable Neo Mode

>Adds support for PS4 Pro emulation and memory size

Here you can add that this option is still unstable and it breaks games, so it is only needed by developers to find new and more complex shaders. Now this option will not make your PC a PS4 Pro console, but rather will break your games.

rainmakerv3:
Applied comments:

1) Changed PS4 pro mode to PS4 neo mode in the GUI
2) added note on PS4 pro instability. 

Enable PS4 Neo Mode:
Adds support for PS4 Pro emulation and memory size. Currently causes instability in a large number of tested games

Rebased to main. Should now be ready

PR #3574, opened 2025-09-11
Qt: Group game-specific and non-game-specific items in the same tabs

rainmakerv3:
Right now in the game-specific GUI, each individual item has to be manually hidden if it shares a tab with the non-game-specific item, and vice versa.

This simplifies adding items to the GUI (just add them to the appropriate tabs if game specific or non game-specific), and everything that needs to be hidden will already be hidden.

Hiding individual items within tabs also causes certain elements to stretch, so it's also kinda ugly that way.

This however means I kinda had to force certain entries in certain tabs. A couple of things that some might not like:
1) Trophy key put in the Frontend Tab (formerly the GUI tab)
2) Default settings tab put in the User Tab

Also, I moved the readbacks and readbacklinearimages to the experimental tab that is only accessible via custom configs. I think it makes sense since you want to enable those on a per-game basis anyway and it's not something probably that we want the casual users touching a lot since we acknowledge that it's experimental

Let me know if this seems iffy

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -100,8 +100,12 @@ public:
     void setTomlValue(toml::ordered_value& data, const std::string& header, const std::string& key,
                       bool is_game_specific = false) {
         if (is_game_specific) {
-            if (game_specific_value == std::nullopt)
+            if (game_specific_value == std::nullopt) {
+                fmt::print("Attempted to save std::nullopt value to {}-{}, matching config entry "
+                           "may not be correctly set-up\n",
+                           header, key);
                 return;
+            }
             data[header][key] = game_specific_value.value_or(base_value);
             game_specific_value = std::nullopt;
         } else {
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -100,6 +100,8 @@ public:
     void setTomlValue(toml::ordered_value& data, const std::string& header, const std::string& key,
                       bool is_game_specific = false) {
         if (is_game_specific) {
+            if (game_specific_value == std::nullopt)
+                return;
             data[header][key] = game_specific_value.value_or(base_value);
             game_specific_value = std::nullopt;
         } else {
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -1110,6 +1110,8 @@ void setDefaultValues(bool is_game_specific) {
     // Entries with game-specific settings that are in the game-specific setings GUI but not in
     // the global settings GUI
     if (is_game_specific) {
+        readbacksEnabled.set(false, is_game_specific);
+        readbackLinearImagesEnabled.set(false, is_game_specific);
         isNeo.set(false, is_game_specific);
         isDevKit.set(false, is_game_specific);
         isPSNSignedIn.set(false, is_game_specific);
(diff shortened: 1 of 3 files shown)


GHU7924:
Personally, I find this really questionable. So, here are my thoughts on these changes:

>Trophy key put in the Frontend Tab (formerly the GUI tab)

It seems illogical.

>Default settings tab put in the User Tab

This option can also be left untouched as it is located on the GUI tab, which is hidden for specific settings.

>Also, I moved the readbacks and readbacklinearimages to the experimental tab that is only accessible via custom configs. I think it makes sense since you want to enable those on a per-game basis anyway and it's not something probably that we want the casual users touching a lot since we acknowledge that it's experimental

These changes seem logical. But let the developers speak out.

Next

I'll repeat once again, we need an Audio tab, this needs to be discussed, but I know at least two options that can be moved there, these are Volume and Microphone, and I also saw PR related to the controller speaker, so maybe new elements will be added.

P.S. In my opinion, we should close this PR, because we have just introduced new features, we need time to test all the changes at least until the next release (I mean v0.12.0), to understand what works, what does not work, what is convenient, what is not convenient, what is superfluous or what is missing, we need to get feedback from both the developers and the community. I believe that the time for this kind of PR has not yet come.

rainmakerv3:
I think it's better to have the current audio settings where they are instead of just having an Audio tab with only two things in it. 

The default settings tab can actually be set on a per-game basis, which someone out there may find useful, but if this PR is applied it can only be used if placed together with other game-specific settings.

I understand your other point and I myself pointed that out, but I think it the benefits of grouping tabs outweight a couple of somewhat oddly placed settings.

Rebased to main. This PR does come with downside, so I will leave it to the reviewer to decide if this should be merged or if it's better the way it is before these changes.

georgemoralis:
pr has some conflicts to resolved before merging . Otherwise looks good to me

GHU7924:
@rainmakerv3, since @georgemoralis approved your PR, I accept it, but maybe we should move the "trophy key" and "default tab" to the "General" tab as it was before?

rainmakerv3:
I think it's a good idea to transfer settings tab to general, so I did that, but since trophy is non-game-specific, I left it with the other non-game-specific items.

Still need one minor cosmetic fix (frame size of the experimental tab)

rainmakerv3:
Ok, should be done now unless there are more concerns

EDIT: took this chance to add a check to the save function for game-specific items to only save when there's a valid value. Trying to save an invalid value could potentially happen if a config entry is incorrectly set-up. Also added missing game-specific arg I missed before

GHU7924:
I only looked at the interface and everything is done quite logically, I am satisfied.

Found a problem, if you select the default "Experimental" tab, then the "General" tab is selected.

tvoracek:
Seems it broke the "Create Game-specific settings". Click on it causes crash of emulator. Don't have any logs as it crashes so fast no logging happens with this error...

Ok, redirected Console output to file and error is:
"Attempted to save std::nullopt value to General-volumeSlider, matching config entry may not be correctly set-up" etc for each setting.

kalaposfos13:
@tvoracek if you haven't noticed it yet, a hotfix for this has already been made for thus 2 hours ago, which is visible right above your comment

PR #3576, opened 2025-09-11
Fix Warnings with Clang 21

Xphalnos:
Fixed warnings seen when compiling with **Visual Studio 2026 (Insiders)** and **Clang 21.1.1**

### DIFF
--- a/src/core/libraries/kernel/threads/thread_state.cpp
+++ b/src/core/libraries/kernel/threads/thread_state.cpp
@@ -100,7 +100,7 @@ Pthread* ThreadState::Alloc(Pthread* curthread) {
         tcb = TcbCtor(thread, 1 /* initial tls */);
     }
     if (tcb != nullptr) {
-        memset(thread, 0, sizeof(Pthread));
+        memset(static_cast<void*>(thread), 0, sizeof(Pthread));
         std::construct_at(thread);
         thread->tcb = tcb;
         thread->sleepqueue = new SleepQueue{};
--- a/src/video_core/renderer_vulkan/vk_presenter.cpp
+++ b/src/video_core/renderer_vulkan/vk_presenter.cpp
@@ -461,7 +461,7 @@ void Presenter::Present(Frame* frame, bool is_reusing_frame) {
     // Reset fence for queue submission. Do it here instead of GetRenderFrame() because we may
     // skip frame because of slow swapchain recreation. If a frame skip occurs, we skip signal
     // the frame's present fence and future GetRenderFrame() call will hang waiting for this frame.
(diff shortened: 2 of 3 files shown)


src/core/libraries/kernel/threads/thread_state.cpp
```diff
@@ -100,7 +100,7 @@ Pthread* ThreadState::Alloc(Pthread* curthread) {
         tcb = TcbCtor(thread, 1 /* initial tls */);
     }
     if (tcb != nullptr) {
-        memset(thread, 0, sizeof(Pthread));
+        memset((void*)thread, 0, sizeof(Pthread));
```
squidbus: If anything should be using a C++ style cast here, but not sure why this is needed to begin with

src/video_core/renderer_vulkan/vk_presenter.cpp
```diff
@@ -461,7 +461,7 @@ void Presenter::Present(Frame* frame, bool is_reusing_frame) {
     // Reset fence for queue submission. Do it here instead of GetRenderFrame() because we may
     // skip frame because of slow swapchain recreation. If a frame skip occurs, we skip signal
     // the frame's present fence and future GetRenderFrame() call will hang waiting for this frame.
-    instance.GetDevice().resetFences(frame->present_done);
+    (void)instance.GetDevice().resetFences(frame->present_done);
```
squidbus: This is not the proper way to address this, the result should be handled.

src/video_core/renderer_vulkan/vk_resource_pool.cpp
```diff
@@ -156,7 +156,7 @@ vk::DescriptorSet DescriptorHeap::Commit(vk::DescriptorSetLayout set_layout) {
     if (const auto [pool, tick] = pending_pools.front(); master_semaphore->IsFree(tick)) {
         curr_pool = pool;
         pending_pools.pop_front();
-        device.resetDescriptorPool(curr_pool);
+        (void)device.resetDescriptorPool(curr_pool);
```
squidbus: Same here

src/core/libraries/kernel/threads/thread_state.cpp
Xphalnos: From Clang:
image

squidbus: The issue here is probably that `Pthread` is not trivially copyable then.

squidbus:
Note that the only warning left here should be the memset one, still not exactly sure what to do about that.

squidbus:
Taking care of last remaining item in https://github.com/shadps4-emu/shadPS4/pull/3625

PR #3580, opened 2025-09-12
video_core: Upload buffer memory around unmapped pages

squidbus:
* Use `CopySparseMemory` to copy memory for buffer upload, to copy around unmapped pages.
  * Remove `prt_areas` check optimization in `CopySparseMemory` to allow this use case.
* Bump `ClampRangeSize` threshold back up as this should handle more cases that would have applied to.
* Downgrade GPU memory assert to warning.

Should take care of GPU memory assert in Ratchet & Clank (2016).

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -68,7 +68,7 @@ void MemoryManager::SetupMemoryRegions(u64 flexible_size, bool use_extended_mem1
 }
 
 u64 MemoryManager::ClampRangeSize(VAddr virtual_addr, u64 size) {
-    static constexpr u64 MinSizeToClamp = 2_MB;
+    static constexpr u64 MinSizeToClamp = 3_GB;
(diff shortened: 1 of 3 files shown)


StevenMiller123 (changes requested):
As-is this needs a minor fix to `CopySparseMemory` to work properly for several cases. The function checks if a VMA is mapped using IsFree, which causes it to perform memcpys from Reserved/PoolReserved memory and throw access violations on Windows devices.

Once that's fixed, this should be good to go.

squidbus:
> As-is this needs a minor fix to `CopySparseMemory` to work properly for several cases. The function checks if a VMA is mapped using IsFree, which causes it to perform memcpys from Reserved/PoolReserved memory and throw access violations on Windows devices.
> 
> Once that's fixed, this should be good to go.

Is the current state of the code what you were looking for?

StevenMiller123 (approved):
Looks good to me.

PR #3592, opened 2025-09-13
kernel: Improvements to condvars.

squidbus:
Some misc changes I made while investigating a condvar issue:
* Fix an oversight in Windows semaphore `try_acquire_for` where the time since the start of the acquire would be subtracted from the remaining time each iteration instead of the time since that iteration started.
* Simplify semaphore `try_acquire_until` to utilize other existing functions instead of requiring its own platform implementations.
* Few type and typo fixes in condvar and mutex code.
* Misc change in time code to check `__APPLE__` definition properly.

### DIFF
--- a/src/core/libraries/kernel/sync/semaphore.h
+++ b/src/core/libraries/kernel/sync/semaphore.h
@@ -87,10 +87,9 @@ public:
     template <class Rep, class Period>
     bool try_acquire_for(const std::chrono::duration<Rep, Period>& rel_time) {
 #ifdef _WIN64
-        const auto start_time = std::chrono::high_resolution_clock::now();
         auto rel_time_ms = std::chrono::ceil<std::chrono::milliseconds>(rel_time);
-
         do {
+            const auto start_time = std::chrono::high_resolution_clock::now();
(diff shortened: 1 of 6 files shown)


Osyotr:
Since you are touching these files it's a good opportunity to replace operator new calls in murex and condvar code with its non-throwing variant. Otherwise the checks on returned pointers are invalid.
Not that it matters when there are no memory though.

squidbus:
> Since you are touching these files it's a good opportunity to replace operator new calls in murex and condvar code with its non-throwing variant. Otherwise the checks on returned pointers are invalid. Not that it matters when there are no memory though.

Not currently planning on that for this PR, feel free to submit something like this though

StevenMiller123:
No improvements or regressions to note on my end, testing on Windows.
I haven't checked too thoroughly for the condvar unreachable itself, but based on your PR description, I assume it's not addressed yet?

Randomuser8219:
I tested R&C on this PR earlier, it still crashed.

squidbus:
> I haven't checked too thoroughly for the condvar unreachable itself, but based on your PR description, I assume it's not addressed yet?

Nope, haven't figured it out yet.

PR #3593, opened 2025-09-13
texture_cache: Make sure that readback images are downloaded in time

squidbus:
Use a thread for deferring copying readback images to memory, instead of a scheduler deferred operation, to make sure that the copy executes in time for the guest to see.

Ratchet & Clank (2016) was hanging on load because it was stalling the main thread expecting a shadow-related image to be read back from the GPU, but the actual memory copy deferred operation did not execute yet. With a thread in the background making sure the copy happens when the tick is reached, instead of whenever the scheduler is next updated to pop deferred operations, the game now progresses without any patches (with readback and readbackLinearImages enabled).

PR #3596, opened 2025-09-13
liverpool: Finish GPU work before signaling idle.

squidbus:
Finish pending GPU work before signaling the `GpuIdle` interrupt, to make sure the guest sees the results in case of e.g. readbacks.

This also fixes Ratchet & Clank (2016) same as https://github.com/shadps4-emu/shadPS4/pull/3593, but with a simpler solution that allows continuing to use scheduler deferred operations.

Performance testing compared to main is appreciated to see if this significantly impacts performance or not.

### DIFF
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -139,7 +139,7 @@ void Liverpool::Process(std::stop_token stoken) {
             VideoCore::EndCapture();
             if (rasterizer) {
                 rasterizer->OnSubmit();
-                rasterizer->Flush();
+                rasterizer->Finish();
             }
             submit_done = false;
         }


StevenMiller123:
Seems to be quite a painful performance drop in KINGDOM HEARTS Ⅲ (readbacks and readbackLinearImages enabled, DMA is disabled).
| Main (~270 ms per frame) | PR (~290 ms per frame) |
|------|----|
| image | image |

These were tested with a locally compiled debug build, so that may skew results?
[CUSA12031 Main.log](https://github.com/user-attachments/files/22313919/CUSA12031.Main.log)
[CUSA12031 PR.log](https://github.com/user-attachments/files/22313920/CUSA12031.PR.log)

squidbus:
> These were tested with a locally compiled debug build, so that may skew results?

Are you able to check with release build in case?

StevenMiller123:
@squidbus It's still a pretty noticeable drop with release actions.
| Main (~27 ms per frame) | PR (~35 ms per frame) |
|------|----|
| image | image |

squidbus:
Okay thanks, was worth checking into at least. Maybe there's some better sync point we can use in the future.

StevenMiller123:
Should I check against your other PR for this issue?

squidbus:
> Should I check against your other PR for this issue?

You can if you want but it shouldn't be an issue, the other one writes out in a separate thread.

PR #3599, opened 2025-09-14
kernel: More thread code clean-up.

squidbus:
Mostly just style warnings like `nullptr` instead of `NULL`, use of C++-style casts, using const where possible, type consistency, etc. Added in some `std::nothrow` constructions as well to make the null checks actually useful.

### DIFF
--- a/src/core/libraries/kernel/threads/condvar.cpp
+++ b/src/core/libraries/kernel/threads/condvar.cpp
@@ -13,16 +13,16 @@ namespace Libraries::Kernel {
 
 static std::mutex CondStaticLock;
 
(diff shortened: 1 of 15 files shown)


tvoracek:
Seems it broke (or other latest change) the "Create Game-specific settings". Click on it causes crash of emulator. Is anyone even testing this before push?

squidbus:
This PR has nothing to do with settings, you’re looking in the wrong place.

kalaposfos13:
It could be this PR: https://github.com/shadps4-emu/shadPS4/pull/3574

tvoracek:
> It could be this PR: #3574

Ok, moved comment there.

PR #3607, opened 2025-09-15
Ds4 Speaker Audio Rebase Fix

nickci2002:
A new PR with a proper rebase.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -1078,14 +1078,17 @@ void save(const std::filesystem::path& path, bool is_game_specific) {
 
         std::vector<DirEntry> sorted_dirs;
         for (const auto& dirInfo : settings_install_dirs) {
-            sorted_dirs.push_back({string{fmt::UTF(dirInfo.path.u8string()).data}, dirInfo.enabled});
+            sorted_dirs.push_back(
+                {string{fmt::UTF(dirInfo.path.u8string()).data}, dirInfo.enabled});
         }
 
         // Sort directories alphabetically
         std::sort(sorted_dirs.begin(), sorted_dirs.end(), [](const DirEntry& a, const DirEntry& b) {
             return std::lexicographical_compare(
                 a.path_str.begin(), a.path_str.end(), b.path_str.begin(), b.path_str.end(),
-                [](char a_char, char b_char) { return std::tolower(a_char) < std::tolower(b_char); });
+                [](char a_char, char b_char) {
+                    return std::tolower(a_char) < std::tolower(b_char);
+                });
         });
 
         for (const auto& entry : sorted_dirs) {
@@ -1101,7 +1104,8 @@ void save(const std::filesystem::path& path, bool is_game_specific) {
         data["GUI"]["installDirsEnabled"] = install_dirs_enabled;
         data["GUI"]["saveDataPath"] = string{fmt::UTF(save_data_path.u8string()).data};
         data["GUI"]["loadGameSizeEnabled"] = load_game_size;
-        data["GUI"]["addonInstallDir"] = string{fmt::UTF(settings_addon_install_dir.u8string()).data};
+        data["GUI"]["addonInstallDir"] = 
+            string{fmt::UTF(settings_addon_install_dir.u8string()).data};
         data["Debug"]["ConfigVersion"] = config_version;
         data["Keys"]["TrophyKey"] = trophyKey;
 
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -302,6 +302,14 @@ string getMicDevice() {
     return micDevice.get();
 }
 
+std::string getMainOutputDevice() {
+    return mainOutputDevice;
+}
+
+std::string getPadSpkOutputDevice() {
+    return padSpkOutputDevice;
+}
+
 double getTrophyNotificationDuration() {
     return trophyNotificationDuration.get();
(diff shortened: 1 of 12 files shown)


src/common/config.cpp
```diff
@@ -194,6 +196,10 @@ static string trophyKey = "";
 // Config version, used to determine if a user's config file is outdated.
 static string config_version = Common::g_scm_rev;
 
+// Output
+static std::string mainOutputDevice = "Default Device";
+static std::string padSpkOutputDevice = "Default Device";
```
squidbus: Should use ‘ConfigEntry’ similar to how the mic device does now (and also probably move mic device together with these definitions here). I see you do already have some TODO items for this.

src/common/config.cpp
```diff
@@ -1036,67 +1067,60 @@ void save(const std::filesystem::path& path, bool is_game_specific) {
 
     m_language.setTomlValue(data, "Settings", "consoleLanguage", is_game_specific);
 
-    // All other entries
```
squidbus, line 1039: I’m not sure if I’m just looking at the diff wrong but I think this may have been rebased wrong, seems a lot of stuff was changed unrelated.

src/core/libraries/audio/sdl_audio.cpp
```diff
@@ -21,8 +21,32 @@ class SDLPortBackend : public PortBackend {
...
+
+        if (port_name == "None") {
+            stream = nullptr;
+            return;
+        } else if (port_name == "Default Device") {
+            devId = SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;
+        } else {
+            try {
+                devId = static_cast<u32>(std::stoul(port_name));
+                if (port_name != SDL_GetAudioDeviceName(devId)) {
```
squidbus: I don’t see how this can work, how can the config value be both the ID number and device name?

I think what you can do here is loop though 0 to SDL_GetNumAudioDevices(), use SDL_GetAudioDeviceName(index) to get the name of each, and compare that name to the config value until you find a match. That way the config can store the actual device name and keep the right device even if the IDs change.

squidbus (commented):
I’ll try to keep this moving with reviews so it doesn’t get out of sync like before. Left some comments.

src/common/config.cpp
nickci2002: This diff is really cursed, but a bit is missing, I'll fix that in my next commit

src/common/config.cpp
nickci2002: Done

georgemoralis:
clang issues

nickci2002:
Can clang-format change the concerns without having to make a new commit myself? I hate having to recompile because I added an extra space somewhere.

squidbus:
> Can clang-format change the concerns without having to make a new commit myself? I hate having to recompile because I added an extra space somewhere.

You can install the clang-format tool locally and run it on your changed files before compiling.

src/core/libraries/audio/sdl_audio.cpp
```diff
@@ -21,8 +22,41 @@ class SDLPortBackend : public PortBackend {
...
+        if (port_name == "None") {
+            stream = nullptr;
+            return;
+        } else if (port_name == "Default Device") {
+            dev_id = SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;
+        } else {
+            try {
+                SDL_AudioDeviceID* dev_array = SDL_GetAudioPlaybackDevices(nullptr);
+                for (; dev_array != 0;) {
+                    std::string name = static_cast<std::string>(SDL_GetAudioDeviceName(*dev_array));
```
squidbus:
```suggestion
                    std::string name(SDL_GetAudioDeviceName(*dev_array));
```

Should just construct from the result instead of casting a `const char*` to `std::string`.

src/core/libraries/audio/sdl_audio.cpp
```diff
@@ -21,8 +22,41 @@ class SDLPortBackend : public PortBackend {
...
+                    if (name == port_name) {
+                        dev_id = *dev_array;
+                        break;
+                    } else {
+                        dev_array++;
+                    }
+                }
+                if (dev_id == 0) {
```
squidbus: Do we know for sure that the device ID cannot be 0? Maybe should use something like max possible value to denote this if it can be 0.

Should probably also either set stream to nullptr here and return (same as `"None"`) or set `dev_id = SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;`. I don't really have a preference on the fallback here as long as we do something that makes sense.

squidbus (commented):
Just two more comments from me on the emulation side of things. I'm not as familiar with some of the stuff changing on the Qt side so may need someone else to review that.

src/core/libraries/audio/sdl_audio.cpp
nickci2002: An AudioDeviceID of zero represents an invalid id according to the SDL docs https://wiki.libsdl.org/SDL3/SDL_AudioDeviceID

nickci2002:
From testing, the GUI works and looks fine. I asked rainmaker to make any changes to the layout if he wanted to.

squidbus:
If the GUI testing is good then it looks fine to me, I’ll leave it up to @georgemoralis whether to pull now or after release.

PR #3608, opened 2025-09-16
QT: Note Vulkan SDK requirement for validation tooltips

Randomuser8219:
The Vulkan SDK is required for validation layers to work.


squidbus:
Can you add it to crash diagnostics as well?

Randomuser8219:
> Can you add it to crash diagnostics as well?

Crash diagnostics already note the requirement.
```
else if (elementName == "crashDiagnosticsCheckBox") {
        text = tr("Crash Diagnostics:\\nCreates a .yaml file with info about the Vulkan state at the time of crashing.\\nUseful for debugging 'Device lost' errors. If you have this enabled, you should enable Host AND Guest Debug Markers.\\nYou need Vulkan Validation Layers enabled and the Vulkan SDK for this to work.");
```

squidbus (approved):

PR #3615, opened 2025-09-17
Some new README images for 0.11.0

Randomuser8219:
Stephen suggested that the Bloodborne image could be changed to an in-game one yesterday on the Discord server.
I also replaced P4G with DRIVECLUB, just in case Atlus tries to go ham on us.
image


raphaelthegreat:
Imo the images should just be removed from README, no point in having them

Randomuser8219:
> Imo the images should just be removed from README, no point in having them

My next PR will be doing that

PR #3616, opened 2025-09-18
Fix: Extend strlcpy implementation to Linux builds

gabrielmrts:
- Extend existing Windows strlcpy implementation to cover Linux systems
- Resolves undefined reference to strlcpy on glibc systems

Uses the project's existing internal implementation instead of adding
external dependencies, maintaining consistency with current approach.

Fixes build error:
undefined reference to symbol 'strlcpy@@LIBBSD_0.0'

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1169,12 +1169,6 @@ if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
         target_compile_definitions(shadps4 PRIVATE ENABLE_USERFAULTFD)
     endif()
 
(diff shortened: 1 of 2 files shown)


squidbus:
We already have a `strlcpy` implementation in the file for Windows, I'd rather you just extend that to be used in this case than introduce a whole library dependency for it.

gabrielmrts:
> We already have a `strlcpy` implementation in the file for Windows, I'd rather you just extend that to be used in this case than introduce a whole library dependency for it.

Makes sense, will do that instead.

gabrielmrts:
@squidbus done.

PR #3622, opened 2025-09-18
Epolls are file descriptors

mikusp:
- registers epolls and resolvers in the file descriptor table (on the real hardware, epolls are kqueues and resolvers are sockets)
- fakes async DNS resolution by storing the call parameters and resolving the address synchronously in sceNetEpollWait
- handles epoll commands for both sockets and resolvers

### DIFF
--- a/src/core/libraries/network/net.cpp
+++ b/src/core/libraries/network/net.cpp
@@ -22,6 +22,7 @@
 #include "core/libraries/network/net.h"
 #include "net_epoll.h"
 #include "net_error.h"
+#include "net_resolver.h"
 #include "net_util.h"
 #include "netctl.h"
 #include "sockets.h"
@@ -645,17 +646,32 @@ int PS4_SYSV_ABI sceNetEpollControl(OrbisNetId epollid, OrbisNetEpollFlag op, Or
             return ORBIS_NET_ERROR_EEXIST;
         }
 
-        auto file = FDTable::Instance()->GetSocket(id);
+        auto file = FDTable::Instance()->GetFile(id);
         if (!file) {
             *sceNetErrnoLoc() = ORBIS_NET_EBADF;
-            LOG_ERROR(Lib_Net, "socket id is invalid = {}", id);
+            LOG_ERROR(Lib_Net, "file id is invalid = {}", id);
             return ORBIS_NET_ERROR_EBADF;
         }
-        epoll_event native_event = {.events = ConvertEpollEventsIn(event->events),
-                                    .data = {.fd = id}};
-        ASSERT(epoll_ctl(epoll->epoll_fd, EPOLL_CTL_ADD, *file->socket->Native(), &native_event) ==
-               0);
-        epoll->events.emplace_back(id, *event);
+
+        switch (file->type) {
+        case Core::FileSys::FileType::Socket: {
+            epoll_event native_event = {.events = ConvertEpollEventsIn(event->events),
+                                        .data = {.fd = id}};
+            ASSERT(epoll_ctl(epoll->epoll_fd, EPOLL_CTL_ADD, *file->socket->Native(),
+                             &native_event) == 0);
+            epoll->events.emplace_back(id, *event);
+            break;
+        }
+        case Core::FileSys::FileType::Resolver: {
+            epoll->async_resolutions.emplace_back(id);
+            epoll->events.emplace_back(id, *event);
+            break;
+        }
+        default: {
+            LOG_ERROR(Lib_Net, "file type {} ({}) passed", magic_enum::enum_name(file->type.load()), file->m_guest_name);
+            break;
+        }
+        }
         break;
     }
     case ORBIS_NET_EPOLL_CTL_MOD: {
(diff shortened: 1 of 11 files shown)


solvise:
STAR WARS™ Battlefront™

log:
[shad_log.txt](https://github.com/user-attachments/files/22414745/shad_log.txt)

mikusp:
@solvise thanks for testing, I've pushed a change that should fix the crash

solvise:
STAR WARS™ Battlefront™

log:
[shad_log.txt](https://github.com/user-attachments/files/22415595/shad_log.txt)

now crashes at: [Debug] <Critical> signals.cpp:86 SignalHandler: Unreachable code!
Unhandled access violation in thread '50_Heartbeat' at code address 0x800005f62: Write to address 0x0

Missake212:
UFC 1:

[CUSA00264.log](https://github.com/user-attachments/files/22415941/CUSA00264.log)

UFC 2 behaves the same:

[CUSA01936.log](https://github.com/user-attachments/files/22415885/CUSA01936.log)

UFC 4 behaves the same:

[CUSA14209.log](https://github.com/user-attachments/files/22415902/CUSA14209.log)

FIFA 18 crashes for something unrelated I believe:

[CUSA07994.log](https://github.com/user-attachments/files/22416050/CUSA07994.log)

FIFA 19 behaves the same:

[CUSA11608.log](https://github.com/user-attachments/files/22415829/CUSA11608.log)

FIFA 20 behaves the same:

[CUSA15545.log](https://github.com/user-attachments/files/22415830/CUSA15545.log)

Little Big Planet 3 crashes but I'm pretty sure that's expected: 

[CUSA00063.log](https://github.com/user-attachments/files/22415836/CUSA00063.log)

The Sims 4 behaves the same:

[CUSA09216.log](https://github.com/user-attachments/files/22415842/CUSA09216.log)

tochnonechelovek:
STAR WARS™ Battlefront™ now increments epollid (didn't before this PR and it was just 0) and then crashes.
[CUSA00634.log](https://github.com/user-attachments/files/22430829/CUSA00634.log)

GUNS UP!™ now spams this, which it didn't before. It didn't mention epoll at all.
`[Lib.Net] <Info> net.cpp:798 sceNetEpollWait: called, epollid = 10 (threadpool_epoll), maxevents = 32, timeout = -1`
[CUSA01483.log](https://github.com/user-attachments/files/22430673/CUSA01483.log)

Everything else is not regressed nor improved.

src/core/libraries/network/net.cpp
```diff
@@ -644,17 +646,33 @@ int PS4_SYSV_ABI sceNetEpollControl(OrbisNetId epollid, OrbisNetEpollFlag op, Or
...
-                                    .data = {.fd = id}};
-        ASSERT(epoll_ctl(epoll->epoll_fd, EPOLL_CTL_ADD, *file->socket->Native(), &native_event) ==
-               0);
-        epoll->events.emplace_back(id, *event);
+
+        switch (file->type) {
+        case Core::FileSys::FileType::Socket: {
+            epoll_event native_event = {.events = ConvertEpollEventsIn(event->events),
+                                        .data = {.fd = id}};
+            ASSERT(epoll_ctl(epoll->epoll_fd, EPOLL_CTL_ADD, *file->socket->Native(),
```
aaalloc, line 660: Testing with The Little big planet 3, I'm having a stop right here because the socket is P2P. What I've seen nothing is implemented right now for that so the value of std::optional.has_value() is false.

I think things should be :

```suggestion
            if(file->socket->Native().has_value()){
                int res = epoll_ctl(epoll->epoll_fd, EPOLL_CTL_ADD, *file->socket->Native(), &native_event);
                ASSERT_MSG(res == 0,
                       "epoll_ctl(ADD) failed for socket fd {}, error code {}", errno);
            } else {
                ASSERT_MSG(false, "Socket has no native handle {}", file->m_guest_name);
            }
```

mikusp: Thanks @aaalloc, you're correct P2P sockets are not implemented right now, that is the next thing I'm working on

src/core/libraries/network/net.cpp
```diff
@@ -703,46 +763,58 @@ int PS4_SYSV_ABI sceNetEpollCreate(const char* name, int flags) {
         return ORBIS_NET_ERROR_EINVAL;
     }
 
-    auto epoll = Common::Singleton<EpollTable>::Instance()->CreateHandle(name);
-
-    return epoll;
+    auto fd = FDTable::Instance()->CreateHandle();
```
aaalloc, line 766: It seems that `sceNetEpoll<Create,Destroy,Control>` from source files has the need to be wrapped with `NetErrorHandler` lambda like
```suggestion
int PS4_SYSV_ABI sceNetEpollCreate(const char* name, int flags) {

...
    return NetErrorHandler([&] {
        auto fd = FDTable::Instance()->CreateHandle();
        auto* epoll = FDTable::Instance()->GetFile(fd);
        epoll->is_opened = true;
        epoll->type = Core::FileSys::FileType::Epoll;
        epoll->epoll = std::make_shared<Epoll>(name);
        epoll->m_guest_name = name;
        return fd;
    });
}
```
for example

PR #3635, opened 2025-09-20
Fixing typos in the GUI and adding an extra line about Null GPU

Missake212:


Missake212:
Forgot about the shadps4-qt launcher, will move this PR over there instead.

PR #3644, opened 2025-09-23
avplayer: fix play request state handling

roamic:
`Katamari Damacy Reroll` videos now play properly. Also re-initialized codecs on the subsequent play requests.


StevenMiller123:
Fixes cutscene playback in KINGDOM HEARTS Melody of Memory
image

PR #3653, opened 2025-09-25
Trophy fixes IXXXX

georgemoralis:
- Fixing missing png extraction from trophy files
- Implemented sceNpTrophyGetGameIcon

### DIFF
--- a/src/core/libraries/np/np_trophy.cpp
+++ b/src/core/libraries/np/np_trophy.cpp
@@ -230,7 +230,8 @@ s32 PS4_SYSV_ABI sceNpTrophyDestroyHandle(OrbisNpTrophyHandle handle) {
     LOG_INFO(Lib_NpTrophy, "Handle {} destroyed", handle);
     return ORBIS_OK;
 }
-size_t ReadFile(Common::FS::IOFile& file, void* buf, size_t nbytes) {
+
+u64 ReadFile(Common::FS::IOFile& file, void* buf, u64 nbytes) {
     const auto* memory = Core::Memory::Instance();
     // Invalidate up to the actual number of bytes that could be read.
     const auto remaining = file.GetSize() - file.Tell();
@@ -240,7 +241,7 @@ size_t ReadFile(Common::FS::IOFile& file, void* buf, size_t nbytes) {
 }
 
 int PS4_SYSV_ABI sceNpTrophyGetGameIcon(OrbisNpTrophyContext context, OrbisNpTrophyHandle handle,
-                                        void* buffer, size_t* size) {
+                                        void* buffer, u64* size) {
     ASSERT(size != nullptr);
     const auto trophy_dir =
         Common::FS::GetUserPath(Common::FS::PathType::MetaDataDir) / game_serial / "TrophyFiles";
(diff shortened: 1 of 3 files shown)


georgemoralis:
@StevenMiller123 i remember you had a game that uses that

StevenMiller123:
This PR brings SWORD ART ONLINE: HOLLOW REALIZATION ingame with minor graphical issues.
image
image
image

This PR also brings SWORD ART ONLINE Re: Hollow Fragment ingame. Not sure if the low resolution is normal or not, it's been a while since I played it on real hardware.
image

src/core/libraries/np/np_trophy.cpp
```diff
@@ -229,10 +230,29 @@ s32 PS4_SYSV_ABI sceNpTrophyDestroyHandle(OrbisNpTrophyHandle handle) {
...
+    auto icon_file = trophy_dir / "trophy00" / "Icons" / "ICON0.PNG";
+
+    Common::FS::IOFile icon(icon_file, Common::FS::FileAccessMode::Read);
+    size_t icon_size = icon.GetSize();
+
+    if (buffer != nullptr) {
+        ReadFile(icon, buffer, *size);
+    }
+
+    *size = icon_size;
```
StevenMiller123: Not sure if firmware checks this or not, but you should probably make sure size isn't a nullptr here before using it.

georgemoralis: according to RE , size only updates when calling with buffer = null to get the bytes needed for allocation. Now it should be proper

StevenMiller123: You should still add a check for whether size is nullptr or not. If there's no error return for it in the library, add an assert instead to make sure it's a valid pointer.

Otherwise, if size is a nullptr when it shouldn't be, games will just silently crash from an access violation.

georgemoralis: done

src/core/libraries/np/np_trophy.cpp
```diff
@@ -229,10 +230,34 @@ s32 PS4_SYSV_ABI sceNpTrophyDestroyHandle(OrbisNpTrophyHandle handle) {
     LOG_INFO(Lib_NpTrophy, "Handle {} destroyed", handle);
     return ORBIS_OK;
 }
+size_t ReadFile(Common::FS::IOFile& file, void* buf, size_t nbytes) {
```
StevenMiller123: Minor nit, add a new line to separate functions.
Also preferred to use u64 over size_t

StevenMiller123 (changes requested):
Just one minor nitpick here

src/core/libraries/np/np_trophy.cpp
georgemoralis: done

StevenMiller123 (approved):
Looks good to me.

src/core/libraries/np/np_trophy.cpp
```diff
@@ -230,9 +231,34 @@ s32 PS4_SYSV_ABI sceNpTrophyDestroyHandle(OrbisNpTrophyHandle handle) {
...
+    const auto remaining = file.GetSize() - file.Tell();
+    memory->InvalidateMemory(reinterpret_cast<VAddr>(buf), std::min<u64>(nbytes, remaining));
+
+    return file.ReadRaw<u8>(buf, nbytes);
+}
+
 int PS4_SYSV_ABI sceNpTrophyGetGameIcon(OrbisNpTrophyContext context, OrbisNpTrophyHandle handle,
-                                        void* buffer, size_t* size) {
-    LOG_ERROR(Lib_NpTrophy, "(STUBBED) called");
+                                        void* buffer, u64* size) {
```
squidbus, line 244: Change to signature needs to be reflected in the headers as well. This is probably the compile error in CI.

squidbus (approved):

PR #3655, opened 2025-09-25
Core: Memory code cleanup and further direct memory fixes

StevenMiller123:
This PR changes some more logic, primarily related to direct memory handling and the dmem map.

- Reimplemented `MemoryManager::SetDirectMemoryType` to search through VMAs to find the physical memory areas to update.
- Updated `sceKernelMtypeprotect` to apply mtype using `MemoryManager::SetDirectMemoryType`.
- Updated `sceKernelMemoryPoolCommit` to apply mtype during `MemoryManager::PoolCommit`.
- Restored overlapping direct memory check, but properly restricted it to `sceKernelMapDirectMemory2` calls.
- Fixed detection of executable memory mappings in libkernel memory functions.
- Fixed issues with backed executable mappings on Windows.
- Fixed a variety of smaller things I've noticed while looking at memory code.

Everything I've changed appears to work fine on my end, but it would be good to have some other tests before merging.

### DIFF
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -103,17 +103,18 @@ struct AddressSpace::Impl {
                         MemoryRegion{system_managed_addr, virtual_size, false});
 
         // Allocate backing file that represents the total physical memory.
-        backing_handle =
-            CreateFileMapping2(INVALID_HANDLE_VALUE, nullptr, FILE_MAP_WRITE | FILE_MAP_READ,
-                               PAGE_READWRITE, SEC_COMMIT, BackingSize, nullptr, nullptr, 0);
+        backing_handle = CreateFileMapping2(INVALID_HANDLE_VALUE, nullptr, FILE_MAP_ALL_ACCESS,
+                                            PAGE_EXECUTE_READWRITE, SEC_COMMIT, BackingSize,
+                                            nullptr, nullptr, 0);
         ASSERT_MSG(backing_handle, "{}", Common::GetLastErrorMsg());
         // Allocate a virtual memory for the backing file map as placeholder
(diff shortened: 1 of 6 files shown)


red-prig:
I'll put in my 2 cents, everything is a bit more complicated and by default dmem can not have overlapping mappings of the same dmem address, but it can be enabled by the game through the `sys_set_chicken_switches` system call or the `sceKernelEnableDmemAliasing` API

StevenMiller123:
I made my own homebrew test for that and overlapping physical addresses in sceKernelMapDirectMemory didn't cause any errors by default 🤔

red-prig:
two sceKernelMapDirectMemory on the same dmem right?

StevenMiller123:
Yup

StevenMiller123:
Marking as draft for now while I run some extra tests.

StevenMiller123:
Ran some deeper tests on the actual memory contents and everything looks good.

red-prig:
It seems that the overlapping check is only in `sceKernelMapDirectMemory2 ` and in `sceKernelMapDirectMemory`, provided that it is a system application.

StevenMiller123:
I've added back in logic to check for physical memory overlaps, but left it tied to a flag specified when calling MapMemory (so only `sceKernelMapDirectMemory2` will trigger it). In the future, when we have proper checks for things like system applications, this code can be used for that too.

I've also added checks for executable mappings where necessary, and tied actual executable mappings to the inputted permissions instead of a separate flag. In the process, I've also fixed executable mappings on Windows too, since RESIDENT EVIL 2 uses it 😅

tochnonechelovek:
No changes in my titles.

Missake212:
No regressions on my end.

squidbus (approved):

PR #3657, opened 2025-09-26
make sys_modules folder configurable

georgemoralis:
This will make sys_modules folder configurable so if you have multiple copies of the emu , not gonna need to have sys_modules in several places. If not set default place will be used


src/common/config.cpp
```diff
@@ -128,6 +128,7 @@ static ConfigEntry<bool> isConnectedToNetwork(false);
 static bool enableDiscordRPC = false;
 static bool checkCompatibilityOnStartup = false;
 static bool compatibilityData = false;
+std::filesystem::path sys_modules_path = {};
```
squidbus:
```suggestion
static std::filesystem::path sys_modules_path = {};
```

squidbus (approved):

PR #3639, opened 2025-09-21
Core: physical backing for flexible and pooled memory allocations

StevenMiller123:
This PR handles partial unmaps of flexible memory. I've created a separate flexible memory map used to track flexible "physical" allocations, calculate free physical addresses to map, and implemented logic in MapMemory and UnmapBytesFromEntry for properly updating this flexible map.

I also added a fix for the isDevKit logic, the backing size wasn't properly updated for the increased direct memory size of isDevKit, which could cause weird issues with direct memory allocations that use the extra space.

This improves Uncharted 4: A Thief’s End™ (CUSA00341), The Last of Us™ Part II (CUSA07820), and FINAL FANTASY XV (CUSA01633) on Windows devices.

PR #3660, opened 2025-09-26
Core: Handle various edge cases related to executable permissions.

StevenMiller123:
As I tested various forms of protections on memory, I found a handful of inaccuracies this PR will address.
- Execute permissions are hidden from direct memory mappings.
- Attempting to map backed memory with execute permssions could fail on Windows, due to an edge case related to `MapViewOfFile3`. Fixing this is necessary for `sceKernelMprotect` to work properly on flexible memory after #3639 
- Protecting memory with read | execute (and no write permissions) would behave strangely on Windows. This was due to an issue with the flags we applied in `AddressSpace::Impl::Protect`, so I cleaned up the logic there too.
- Internally, the PS4 kernel seems to treat read and write permissions separately, but applies read permissions whenever only write is specified. Handling this properly fixes the raw value of VMA protections occasionally being exactly one less than equivalent mappings on real hardware.
- `sceKernelMapDirectMemory2` returns EINVAL when requesting execute permissions, not EACCES.

### DIFF
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -157,7 +157,7 @@ struct AddressSpace::Impl {
                 ptr = MapViewOfFile3(backing, process, reinterpret_cast<PVOID>(virtual_addr),
                                      phys_addr, size, MEM_REPLACE_PLACEHOLDER,
                                      PAGE_EXECUTE_READWRITE, nullptr, 0);
-                ASSERT_MSG(ptr, "{}", Common::GetLastErrorMsg());
+                ASSERT_MSG(ptr, "MapViewOfFile3 failed. {}", Common::GetLastErrorMsg());
                 DWORD resultvar;
                 bool ret = VirtualProtect(ptr, size, prot, &resultvar);
                 ASSERT_MSG(ret, "VirtualProtect failed. {}", Common::GetLastErrorMsg());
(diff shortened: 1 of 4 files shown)


src/core/address_space.cpp
```diff
@@ -155,7 +155,12 @@ struct AddressSpace::Impl {
                 ASSERT_MSG(ret, "VirtualProtect failed. {}", Common::GetLastErrorMsg());
             } else {
                 ptr = MapViewOfFile3(backing, process, reinterpret_cast<PVOID>(virtual_addr),
-                                     phys_addr, size, MEM_REPLACE_PLACEHOLDER, prot, nullptr, 0);
+                                     phys_addr, size, MEM_REPLACE_PLACEHOLDER,
+                                     PAGE_EXECUTE_READWRITE, nullptr, 0);
+                ASSERT_MSG(ptr, "{}", Common::GetLastErrorMsg());
```
squidbus: Probably should have some kind of message with this similar to the one above, like `MapViewOfFile3 failed. {}` maybe.

squidbus (approved):
Presumably needs testing but code looks good to me.

StevenMiller123:
I've located some regressions, so I'm marking this as draft for now. Will probably fix sometime tomorrow.

StevenMiller123:
I've copied my newer protection flag logic over to the prot converting functions, and that fixes the regression I hit. I doubt this will fix any new bugs though.

tochnonechelovek:
No regressions on my end.

GHU7924:
I also didn't notice any regressions.

StevenMiller123:
I found another regression, marking as draft while I investigate.

StevenMiller123:
Seems like it was a regression from #3655 😅 
I've pushed some more fixes here, I think everything on my end should be fixed now?

georgemoralis:
Seems fine to me . Let's go for general testing

PR #3666, opened 2025-09-27
vulkan: fix resetDescriptorPool and free ImGui descriptor sets

ADEMOLA200:
This change fixes two build/runtime issues related to Vulkan usage:

Build & test:
- Built successfully on macOS (Apple M2, MoltenVK) after these changes.

### DIFF
--- a/src/imgui/renderer/imgui_impl_vulkan.cpp
+++ b/src/imgui/renderer/imgui_impl_vulkan.cpp
@@ -10,6 +10,10 @@
 
 #include "imgui_impl_vulkan.h"
 
+#include <vulkan/vulkan.hpp>
+#include <fmt/core.h>
+#include <fmt/chrono.h>
+
 #ifndef IM_MAX
 #define IM_MAX(A, B) (((A) >= (B)) ? (A) : (B))
 #endif
@@ -471,7 +475,7 @@ void RemoveTexture(ImTextureID texture) {
     IM_ASSERT(texture != nullptr);
     VkData* bd = GetBackendData();
     const InitInfo& v = bd->init_info;
-    CheckVkErr(v.device.freeDescriptorSets(bd->descriptor_pool, {texture->descriptor_set}));
+    v.device.freeDescriptorSets(bd->descriptor_pool, {texture->descriptor_set});
     delete texture;
 }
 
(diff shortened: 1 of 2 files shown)


src/imgui/renderer/imgui_impl_vulkan.cpp
```diff
@@ -10,6 +10,10 @@
 
 #include "imgui_impl_vulkan.h"
 
+#include <vulkan/vulkan.hpp>
+#include <fmt/core.h>
+#include <fmt/chrono.h>
+
```
kalaposfos13, line 16: Why did you add extra includes when you don't use them at all? You only removed code from this file and didn't add anything new other than these includes.

src/video_core/renderer_vulkan/vk_resource_pool.cpp
```diff
@@ -8,6 +8,8 @@
 #include "video_core/renderer_vulkan/vk_instance.h"
 #include "video_core/renderer_vulkan/vk_master_semaphore.h"
 #include "video_core/renderer_vulkan/vk_resource_pool.h"
+#include <type_traits>
+#include <utility>
```
kalaposfos13, line 12: Same here, unless building magically broke in the last few minutes, this isn't needed.

src/video_core/renderer_vulkan/vk_resource_pool.cpp
```diff
@@ -152,14 +154,15 @@ vk::DescriptorSet DescriptorHeap::Commit(vk::DescriptorSetLayout set_layout) {
     ASSERT_MSG(result == vk::Result::eErrorOutOfPoolMemory ||
                    result == vk::Result::eErrorFragmentedPool,
                "Unexpected error during descriptor set allocation: {}", vk::to_string(result));
+
```
kalaposfos13, line 157:
```suggestion
```

squidbus:
* I haven’t seen myself if there’s actually an older Vulkan-Hpp without the return values, but if there is it’s not supported since that’s wrong, and you should just build with the copy bundled here as a submodule. It certainly is not needed for MoltenVK/macOS as you can see in our macOS CI build.
* This looks AI generated

src/imgui/renderer/imgui_impl_vulkan.cpp
ADEMOLA200: > Why did you add extra includes when you don't use them at all? You only removed code from this file and didn't add anything new other than these includes.

My apologies I was experimenting with something locally and forgot to remove these includes before committing I'll clean this up.

src/video_core/renderer_vulkan/vk_resource_pool.cpp
ADEMOLA200: > Same here, unless building magically broke in the last few minutes, this isn't needed.

I will clean this up in the next commit

PR #3096, opened 2025-06-13
Handle mixed samples attachments during draws

mikusp:
Another shot at fixing the issue mentioned in https://github.com/shadps4-emu/shadPS4/pull/3020, when a game tries to do a draw into multiple color attachments that differ in their sample counts.

No longer requires VK_AMD/NV_* extension as the previous one, but VK_EXT_dynamic_rendering_unused_attachments and extendedDynamicState3RasterizationSamples - both handled by three major GPU vendors. Not supported by MoltenVK, that would need to wait for another PR. Now it gets ingame with HEAVY RAIN and renders the whole framebuffer
Screenshot From 2025-06-13 19-33-44

PR #3667, opened 2025-09-27
Handle mixed samples attachments (V2)

### LINKED ISSUE #3627: [GAME BUG]: God of War III Remastered device lost on Nvidia GPUs
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [x] I have disabled all patches and cheats and the issue is still present.
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadPS4/wiki/I.-Quick-start-%5BUsers%5D#4-adding-modules) installed.

### Describe the Bug

Since we can't use validation layer in the compatibility list I decided to open this here, God of War III Remastered goes to the main menu on both Intel and AMD GPUs but not on Nvidia (which is unfair), it hits a device lost before that, I know PR #3096 helped this issue (it also helped a lot of other games like Just Dance 2019, NBA 2K15, The Order 1886 ... so fixing this might fix other games) but it seems like it has been abandoned since, I believe this is the reason of the crash:

``[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-vkCmdDraw-multisampledRenderToSingleSampled-07285: vkCmdDraw(): Color attachment (0) sample count (VK_SAMPLE_COUNT_1_BIT) must match the corresponding VkPipelineMultisampleStateCreateInfo sample count (VK_SAMPLE_COUNT_2_BIT).``
``The Vulkan spec states: If the bound pipeline was created without a VkAttachmentSampleCountInfoAMD or VkAttachmentSampleCountInfoNV structure, and the multisampledRenderToSingleSampled feature is not enabled, and the current render pass instance was begun with vkCmdBeginRendering with a VkRenderingInfo::colorAttachmentCount parameter greater than 0, then each element of the VkRenderingInfo::pColorAttachments array with a imageView not equal to VK_NULL_HANDLE must have been created with a sample count equal to the value of rasterizationSamples for the bound graphics pipeline (https://vulkan.lunarg.com/doc/view/1.4.321.1/windows/antora/spec/latest/chapters/drawing.html#VUID-vkCmdDraw-multisampledRenderToSingleSampled-07285)``

Also for some reason using crash diagnostic seems to bypass this crash, not sure if that's useful but I thought I would leave that here. The device lost also happens on a 1650 Super on Linux so I think it's safe to assume it happens on every Nvidia GPU that are above the minimum specs for the emulator.

### Reproduction Steps

Launch the game on an Nvidia GPU then crash on the device lost.

### Specify OS Version

Windows 11

### CPU

Intel Core I9 13900K

### GPU

Nvidia RTX 4090

### Amount of RAM in GB

32 GB

### Amount of VRAM in GB

24 GB

### Log File

[CUSA01715.log](https://github.com/user-attachments/files/22429418/CUSA01715.log)

raphaelthegreat:
AMD GCN has fairly advanced and convoluted multisampling capabilities. There is a published GL [extension](https://registry.khronos.org/OpenGL/extensions/AMD/AMD_framebuffer_multisample_advanced.txt) that exposes some of that capability and has a good description of the basics. In addition the mesa OpenGL driver that implements EQAA contains a more [in-depth](https://github.com/chaotic-cx/mesa-mirror/blob/4531b8ace44342be53311d4abe54fe21632f28c1/src/gallium/drivers/radeonsi/si_state.c#L3489) comment about what each register is responsible for. The TLDR is that the hardware supports having separate coverage, depth and color samples.

To avoid this large complexity the current code only considers the depth buffer samples and the fragments of each bound color target (i.e storage samples). The MSAA_NUM_SAMPLES bitfield from PA_SC_AA_CONFIG is ignored because it contains the coverage samples that can be higher than both depth and color samples in case of EQAA. But even with this, there are still cases of color targets having mixed storage samples or depth mismatching or both. There have been several attempts of trying to address this problem, all with separate problems (including this one) as there isn't a perfect solution, and adding a config to disable MSAA entirely feels hacky.

First solution from https://github.com/shadps4-emu/shadPS4/pull/3020 used the AMD/NV extensions for mixed samples, similar to this PR. The vulkan spec only intends that you use these for mismatched depth samples, but AMD hardware supports mismatched color samples too, so it works anyway with that extension. NVIDIA hardware only supports depth mismatch, so it still crashes if color samples mismatch.

The other attempt in https://github.com/shadps4-emu/shadPS4/pull/3096 from what I understood took a different approach of splitting the draw based on sample count, essentially masking out attachments that mismatched. The problem with this, is that it can affect performance depending on the number of draws affected (as they are duplicated) and also that draws can be statefull (vertex/fragment shaders can load/store from images and buffers)

This attempt is similar to the 1st one, in that it uses the extensions when possible to avoid doing extra work in cases that are covered by those. In the absence of those a new system has been placed that allows images to swap their backing samples to match what is requested by the pipeline.

The first commit refactors the render target binding code to be have a similar 2-step bind process like textures and buffers. Technically it was already 2-step on main, but the first step did 90% of the work of fetching render targets and setting up render state, with the second step only doing additional layout transitions if needed. Now most of the work has been deferred there, with the first step only fetching images from texture cache. This is useful as it allows us to swap the VkImage before any image views are involved.

Each texture cache image now has an array of backing images, which are the vulkan allocations with the necessary state tracking and a call for swapping to a new backing image. If the pipeline cache detects on unsupported MSAA configuration it will decide, depending on the extensions it has available to modify the samples to avoid crashing. In all cases it will try to match the depth samples, as this avoids needing to also swap the depth image (I believe copying to multisampled stencil is impossible on NVIDIA, as the hardware doesn't support stencil export, nor storage capability on D32S8). When the image is later bound as texture or part of overlap resolution it will swap back to its guest advertised samples to avoid issues.

Using this backing system means the image can still advertise itself with the correct guest samples. While multsampled images can't be worked with easily without a resolve, it is possible for the guest to alias them with non multisampled images, as it is aware of HW image layouts, so its usually best to avoid messing with that aspect. The problem with this approach is obviously the extra copies it adds and potentially some state tracking errors from all the extra logic. I don't know if its better than the previous one, but it should work on all platforms


Xcedf:
Finally gow3 ingame with no hacks

SixSeven-off:
what about white light too bright scenes? Like in The Forum vs Hercules?

SixSeven-off:
Oh another fix, what this includes forgive me?

StevenMiller123:
@ThunderNowX This just fixes a regression reported on Discord.

src/video_core/texture_cache/image.cpp
```diff
@@ -491,13 +486,86 @@ void Image::CopyMip(const Image& src_image, u32 mip, u32 slice) {
...
+    cmdbuf.copyImage(src_image.GetImage(), src_image.backing->state.layout, GetImage(),
+                     backing->state.layout, image_copy);
 
     Transit(vk::ImageLayout::eGeneral,
             vk::AccessFlagBits2::eShaderRead | vk::AccessFlagBits2::eTransferRead, {});
 }
 
-Image::~Image() = default;
+void Image::SetBackingSamples(u32 num_samples) {
+    if (backing->num_samples == num_samples) {
```
squidbus: Got a crash here in Gravity Rush Remastered, where `backing` is `nullptr`.

raphaelthegreat: Oh I think is from stencil images that don't have allocations, should add a null check here

raphaelthegreat: Or actually, this is resolved to the depth attachment before the call, so not sure. Can you check the callstack

squidbus: If it helps I believe it's specifically the `info.pixel_format == vk::Format::eUndefined` killing the constructor that leads to the nullptr here.

squidbus: > Or actually, this is resolved to the depth attachment before the call, so not sure. Can you check the callstack

stack

raphaelthegreat: I see, yeah its def stencil and that can be fixed with a null check (though I'm curious how come a stencil attachment is part of an overlap resolve)

squidbus: Not sure but it happens pretty immediately on boot so it's pretty easy to reproduce.

SixSeven-off: @raphaelthegreat can u fix white sky in gow3 scenes? Don't know if u seen something
Thx in advance

raphaelthegreat: I've seen your message you don't have to write under random review comments

SixSeven-off: Sorry my bad
Where so?

raphaelthegreat: I'm not making any promises on that front that is all I'll say

diogoaduartereis:
The order 1886 goes in-game with this PR. In my rtx 4060, crashes soon after though

Missake212:
God of War III Remastered now goes in game on both Nvidia and Intel

Nvidia
image

Intel
image


Just Dance 2019 now gets to menus on both Nvidia and Intel

image


NBA 2K15 doesn't crash anymore when selecting the Blacktop game mode on Nvidia

image

On Intel the game gets past the initial device lost but hits another one after, tried to reproduce it once with validation enabled but couldn't get it to happen


The Order 1886 now gets to menus on Nvidia but crashes on what I think is garbage data? I'm still not sure about my game dump for this one to be honest.

image

``[Debug] <Critical> liverpool_to_vk.cpp:763 operator(): Assertion Failed!``
``Unknown data_format=36 and num_format=14``

[CUSA00076.log](https://github.com/user-attachments/files/22580183/CUSA00076.log)

Similar thing for Intel

``[Debug] <Critical> liverpool_to_vk.cpp:763 operator(): Assertion Failed!``
``Unknown data_format=27 and num_format=0``

[CUSA00076.log](https://github.com/user-attachments/files/22580204/CUSA00076.log)

Will test for regression later today.

Xcedf:
Order 1886 has 2 shaders that gives random Unknown data_format and num_format every time, not consistent

tochnonechelovek:
Somehow, this amount of changes didn't change anything in my titles. No regressions.

isshininu:
Снимок экрана (2064)
1886 goes further, crash on loading save:
Unknown data_format=28 and num_format=0

[shad_log.zip](https://github.com/user-attachments/files/22582140/shad_log.zip)

GHU7924:
Order 1886

With `readbacks: true` I get the same picture as **_Missake212_** 
If you also include `readbackLinearImages: true` and `directMemoryAccess: true`, it will be like **_isshininu_** 
The error is the same, but `data_format` and `num_format` change, as mentioned above, so I will not attach the log.

In Main, with the same parameters, there was a different error:

[Debug] <Critical> vk_scheduler.cpp:174 operator(): Assertion Failed!
Device lost during submit

P.S. I didn't notice any changes in the other games.

raphaelthegreat:
Latest commit is a relatively large refactor to make this code more robust in general
- Moved several operations like clear, resolve, image_to_buffer and buffer_to_image copies inside the image class as member functions that users need to call. These functions ensure that the backing samples are correct before any operation, so callers don't need to manually do that everywhere anymore

- Cleaned up presenter, as it had various code leftover from previous implementations that aren't true anymore. Mainly, PrepareFrame is always called from GPU thread now, so the usage of flip scheduler is not needed. FlushDraw is also unnecessary due to this. The PrepareFrameInternal function also grew large and complicated so split the normal and blank frame version and moved some debug labeling inside the post processing passes

- PrepareRenderState now calls UpdateImage on color buffers to ensure any copies from compute generated buffers are flushed before it swaps the backing samples. The first change ensures that the backing samples will be correct when RefreshImage attempts the copy

kvicken:
METAL GEAR SOLID 3: Snake Eater - Master Collection Version Ingame
image

SixSeven-off:
tried last build fix degamma, in gow3, still some points too bright,and texture are all messed up

src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
```diff
@@ -259,7 +259,20 @@ GraphicsPipeline::GraphicsPipeline(
         color_formats[i] = color_format;
     }
 
+    std::array<vk::SampleCountFlagBits, Liverpool::NumColorBuffers> color_samples;
+    std::ranges::transform(key.color_samples, color_samples.begin(), [&instance](u8 num_samples) {
+        return num_samples ? LiverpoolToVK::NumSamples(num_samples, instance.GetColorSampleCounts())
+                           : vk::SampleCountFlagBits::e1;
+    });
+    const vk::AttachmentSampleCountInfoAMD mixed_samples = {
```
squidbus, line 267: Could you include a comment this is the same as the NVIDIA struct, confused me a bit that we were including it for both extensions.

SixSeven-off: Saying to me or raphael?

squidbus:
Seems to be behaving a lot worse for me since refactor, for example:
* Bloodborne has some white flickering and everything looks like it's covered in charcoal.
* Gravity Rush Remastered has black flickering often when rotating the camera
* Sakura Wars has some missing UI graphics and then ingame the world does not render, just black screen with UI.

raphaelthegreat:
I seem to have missed a scheduler->EndRendering() in Image::Download but those glitches sound worse than just that. Are you testing on moltenvk?

squidbus:
> I seem to have missed a scheduler->EndRendering() in Image::Download but those glitches sound worse than just that. Are you testing on moltenvk?

Yes. I can't test on Windows until later today.

It seems like the latest commit improved things a lot but some flickering is still there.

raphaelthegreat:
Are you seeing any other validation errors?

squidbus:
Seems related to the flickering at least in Bloodborne:
```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-vkCmdDispatch-None-10672: vkCmdDispatch(): It is invalid to issue this call inside an active VkRenderPass 0x0.
The Vulkan spec states: If the per-tile execution model is not enabled, this command must be called outside of a render pass instance (https://vulkan.lunarg.com/doc/view/1.4.321.0/mac/antora/spec/latest/chapters/dispatch.html#VUID-vkCmdDispatch-None-10672)
```

raphaelthegreat:
Found a possible source for this validation error

raphaelthegreat:
Though this whole new code shouldn't really trigger for games like bloodborne so I'm not sure what is happening there

squidbus:
Still seems to occur, this is where it's coming from (using backtrace in the debug utils callback):
```
8   shadps4                             0x00000200007cea08 _ZN9VideoCore11TileManager11DetileImageEN2vk6BufferEjRKNS_9ImageInfoE + 1160
9   shadps4                             0x00000200007c32d2 _ZN9VideoCore12TextureCache12RefreshImageERNS_5ImageE + 1682
10  shadps4                             0x00000200007c493b _ZN9VideoCore12TextureCache11UpdateImageEN6Common6SlotIdE + 267
11  shadps4                             0x00000200007c47ff _ZN9VideoCore12TextureCache11FindTextureEN6Common6SlotIdERKNS0_8BaseDescE + 127
12  shadps4                             0x00000200007820c9 _ZN6Vulkan10Rasterizer12BindTexturesERKN6Shader4InfoERNS1_7Backend8BindingsE + 1737
13  shadps4                             0x000002000077f0ad _ZN6Vulkan10Rasterizer13BindResourcesEPKNS_8PipelineE + 557
14  shadps4                             0x000002000077eb77 _ZN6Vulkan10Rasterizer4DrawEbj + 247
```

raphaelthegreat:
Third time's the charm ig

squidbus:
Seems to render okay now.

There's one more validation error I think related to the host shader changes and the SPIR-V version used by glslang:
```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-VkShaderModuleCreateInfo-pCode-08740: vkCreateShaderModule(): SPIR-V Capability ShaderViewportIndexLayerEXT was declared, but one of the following requirements is required (VK_EXT_shader_viewport_index_layer OR VK_NV_viewport_array2).
The Vulkan spec states: If pCode is a pointer to SPIR-V code, and pCode declares any of the capabilities listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied (https://vulkan.lunarg.com/doc/view/1.4.321.0/mac/antora/spec/latest/chapters/shaders.html#VUID-VkShaderModuleCreateInfo-pCode-08740)
```

Using a different EXT version instead of `ShaderViewportIndex` which was introduced in SPIR-V 1.5, while we compile for 1.3.

Also a bunch like this which I'm not sure if it's related to this PR or not:
```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-vkCmdCopyImage-srcImage-08793: vkCmdCopyImage(): pRegions[0].srcSubresource.layerCount (1) is different from dstSubresource.layerCount (2). Both images are VK_IMAGE_TYPE_2D.
The Vulkan spec states: If srcImage and dstImage have the same VkImageType, for each element of pRegions, if neither of the layerCount members of srcSubresource or dstSubresource are VK_REMAINING_ARRAY_LAYERS, the layerCount members of srcSubresource or dstSubresource must match (https://vulkan.lunarg.com/doc/view/1.4.321.0/mac/antora/spec/latest/chapters/copies.html#VUID-vkCmdCopyImage-srcImage-08793)
```

Both in Bloodborne.

raphaelthegreat:
Fixed the first by removing the unused shader that triggers it. Second I'm not sure, would check if its also present in main

squidbus:
> Fixed the first by removing the unused shader that triggers it. Second I'm not sure, would check if its also present in main

Yeah seems that one exists on main. Specifically in `ResolveDepthOverlap` image copy. I'll submit my own fix for this after this PR since it touches some of the same code.

georgemoralis:
so your problems solved now @squidbus ?

squidbus:
There seems to be one more validation error surfacing in my games from this PR:

```
[Render.Vulkan] <Error> vk_platform.cpp:56 DebugUtilsCallback: VUID-vkCmdDraw-dynamicRenderingUnusedAttachments-08910: vkCmdDraw(): VkRenderingInfo::pColorAttachments[0].imageView format (VK_FORMAT_R8G8B8A8_UNORM) must match the corresponding format in VkPipelineRenderingCreateInfo::pColorAttachmentFormats[0] (VK_FORMAT_UNDEFINED).
The Vulkan spec states: If the dynamicRenderingUnusedAttachments feature is not enabled, and the current render pass instance was begun with vkCmdBeginRendering and VkRenderingInfo::colorAttachmentCount greater than 0, then each element of the VkRenderingInfo::pColorAttachments array with an imageView not equal to VK_NULL_HANDLE must have been created with a VkFormat equal to the corresponding element of VkPipelineRenderingCreateInfo::pColorAttachmentFormats used to create the bound graphics pipeline (https://vulkan.lunarg.com/doc/view/1.4.321.0/mac/antora/spec/latest/chapters/drawing.html#VUID-vkCmdDraw-dynamicRenderingUnusedAttachments-08910)
```

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -107,125 +107,58 @@ bool Rasterizer::FilterDraw() {
...
     const auto& regs = liverpool->regs;
-
     if (regs.color_control.degamma_enable) {
         LOG_WARNING(Render_Vulkan, "Color buffers require gamma correction");
     }
 
     const bool skip_cb_binding =
         regs.color_control.mode == AmdGpu::Liverpool::ColorControl::OperationMode::Disable;
-
     for (s32 cb = 0; cb < state.num_color_attachments && !skip_cb_binding; ++cb) {
```
squidbus:
```suggestion
    for (s32 cb = 0; cb < state.num_color_attachments; ++cb) {
```

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -107,125 +107,58 @@ bool Rasterizer::FilterDraw() {
...
+        auto& [image_id, desc] = cb_descs[cb];
         const auto& col_buf = regs.color_buffers[cb];
-        if (!col_buf) {
-            state.color_attachments[cb].imageView = VK_NULL_HANDLE;
+        if (!col_buf || (mrt_mask & (1 << cb)) == 0 || !regs.color_target_mask.GetMask(cb)) {
```
squidbus:
```suggestion
        if (skip_cb_binding || !col_buf || (mrt_mask & (1 << cb)) == 0 || !regs.color_target_mask.GetMask(cb)) {
```

squidbus (commented):
For the last validation error; the `cb_descs` image IDs are not being cleared when color is disabled.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
raphaelthegreat: Thanks for finding this, applied

SixSeven-off:
immagine
bright gow3 in last mixed-msaa

src/video_core/texture_cache/image.cpp
```diff
@@ -491,13 +550,161 @@ void Image::CopyMip(const Image& src_image, u32 mip, u32 slice) {
...
+    if (!backing || backing->num_samples == num_samples) {
+        return;
+    }
+    ASSERT_MSG(!info.props.is_depth, "Swapping samples is only valid for color images");
+    BackingImage* new_backing;
+    auto it = std::ranges::find(backing_images, num_samples, &BackingImage::num_samples);
+    if (it == backing_images.end()) {
+        auto new_image_ci = backing->image.image_ci;
+        new_image_ci.samples =
+            LiverpoolToVK::NumSamples(num_samples, instance->GetColorSampleCounts());
```
squidbus:
```suggestion
            LiverpoolToVK::NumSamples(num_samples, supported_samples);
```

Image has `supported_samples` which records the supported sample count queried from format properties.

SixSeven-off:
Does this forced push, include something new? Like what i posted up here?

tochnonechelovek:
Still no regressions and no noticable changes in my titles on the latest commit.

PR #3669, opened 2025-09-28
Ime Fix IXX

georgemoralis:
This disable the check for unsupported bits not because it is wrong but probably because we don't handle them

### DIFF
--- a/src/core/libraries/ime/ime_dialog.cpp
+++ b/src/core/libraries/ime/ime_dialog.cpp
@@ -34,7 +34,7 @@ static bool IsValidOption(OrbisImeOption option, OrbisImeType type) {
     }
 
     if (True(option & OrbisImeOption::NO_AUTO_CAPITALIZATION /* NoAutoCompletion */) &&
-        type != OrbisImeType::Number && type != OrbisImeType::BasicLatin) {
+        type != OrbisImeType::Default && type != OrbisImeType::BasicLatin) {
         return false;
     }
 
--- a/src/core/libraries/ime/ime_dialog.cpp
+++ b/src/core/libraries/ime/ime_dialog.cpp
@@ -21,11 +21,13 @@ static ImeDialogState g_ime_dlg_state{};
 static ImeDialogUi g_ime_dlg_ui;
 
 static bool IsValidOption(OrbisImeOption option, OrbisImeType type) {
+#if 0
+    // Reject if option contains any unsupported bits (unfinished should check for rest of the bits in ImeOption as well)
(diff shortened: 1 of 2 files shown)


georgemoralis:
https://github.com/shadps4-compatibility/shadps4-game-compatibility/issues/1713 should be fixed now

gandalfthewhite19890404:
Yes, Harvest Moon: Light of Hope SE Complete fixed now. Stardew Valley, FF9, Diablo 3 - works fine, without regressions

src/core/libraries/ime/ime_dialog.cpp
```diff
@@ -21,18 +21,20 @@ static ImeDialogState g_ime_dlg_state{};
 static ImeDialogUi g_ime_dlg_ui;
 
 static bool IsValidOption(OrbisImeOption option, OrbisImeType type) {
+#if 0
+    // Reject if option contains any unsupported bits (unfinished should check for rest of the bits in ImeOption as well)
```
squidbus, line 25: Can’t we just add the bits that were missing here?

georgemoralis: there are 15 options and i am not sure if that check is even neccesary...

squidbus: Well it would be nice to know what this check was for before it's removed... I don't think the comment here you've added is correct, as this check seems to be saying: if both `MULTILINE` and `NO_AUTO_CAPITALIZATION` are set, return false (invalid).

georgemoralis: yah i didn't write that code so i don't know why this check was added and the check doesn't make sense to me either since i suppose multiline and auto_capitalization is possible co-exist..

squidbus: Might as well ask @LNDF I guess (going by the git blame).

georgemoralis: or perhaps @w1naenator as he did most of the ime rewrite

w1naenator: This check was originally there, and I didn’t change it — I just left it without really considering whether it is necessary or correct.

w1naenator: I think that determining the correct checks is essential when the emulator is intended for homebrew testing.

w1naenator:
I think I found reason of crash, please check it:
[https://github.com/shadps4-emu/shadPS4/pull/3680](https://github.com/shadps4-emu/shadPS4/pull/3680)

PR #3671, opened 2025-09-29
Update links to game showcases in the README

UltraDaCat:
Updates the links and replaces DSR and RDR1 with games that seem a lot more impressive. Some of the videos were pretty old(0.1.1 and 0.2.1 old). If y'all think a different game should be put just let me know.

Before and after comparison:
image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ SPDX-License-Identifier: GPL-2.0-or-later
         <img src="https://img.shields.io/github/stars/shadps4-emu/shadPS4" width="120">
 </h1>
 
-|               Bloodborne by From Software                   |                     Hatsune Miku Project DIVA Future Tone by SEGA                         |
+|               borne by From Software                   |                     Hatsune Miku Project DIVA Future Tone by SEGA                         |
 | :-----------------------------------------------------------: | :--------------------------------------------------------------------------------------------: |
-| ![Bloodborne screenshot](./documents/Screenshots/1.png) | ![Project DIVA screenshot](./documents/Screenshots/2.png) |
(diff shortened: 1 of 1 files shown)


Randomuser8219:
The DRIVECLUB video is from a fork, I think this video I made will work better since it's not on a fork: [here](https://www.youtube.com/watch?v=-ZvhIGtYjtc)

UltraDaCat:
replaced, anything else left?

Missake212:
Wouldn't it be better to get better videos first then update the links? I'm not sure what's the common opinion but I think it's better to show videos of stuff that's already running good instead, also I think the Bloodborne video has like 1000 mods in it

UltraDaCat:
RDR1 was plagued with issues(still is I'm pretty sure) yet it was showcased. I think DRIVECLUB and Bloodborne are in good enough states(much better than RDR1 at least) as to where it's okay to showcase them. Regarding mods used in the Bloodborne video, it seems to be in the same boat as the previous one. It looks like the mods used are mostly UI related(which is disabled in the video anyway), but you'd have to ask Kyo to see what mods he used to be 100% sure. I think if anything maybe Borderlands 3 should be replaced with some other game that's in a better state, but I'll leave the decision up to any of the reviewers of this.

Missake212:
Yeah I was talking mainly about Borderlands here (and the old Driveclub video before it got changed), I still think going forward it's better to not showcase mods, not sure what's being used obviously but throughout the whole video there's not a single enemy on the map and some items were removed, I think that touches a little bit more than UI, the best change would probably be to remove the word "successfully" I guess.

hgh32:
I am in favor of removing the DSR because it has issues on Intel CPUs

PR #3675, opened 2025-09-30
VK Wayland RenderDoc log incompatible

Niram7777:


src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -86,6 +86,10 @@ vk::SurfaceKHR CreateSurface(vk::Instance instance, const Frontend::WindowSDL& e
             UNREACHABLE();
         }
     } else if (window_info.type == Frontend::WindowSystemType::Wayland) {
+        if (Config::isRdocEnabled()) {
+            LOG_ERROR(Render_Vulkan, "As of today RenderDoc and Wayland are incompatible!");
```
squidbus:
```suggestion
            LOG_ERROR(Render_Vulkan, "RenderDoc is not compatible with Wayland, use an X11 window instead.");
```

nit: `today` is a bit vague and easily outdated, and can suggest the alternative here.

StevenMiller123: Nit of the nit, I assume "not incompatible" should probably be "not compatible"

squidbus: > Nit of the nit, I assume "not incompatible" should probably be "not compatible"

Fixed the suggestion

PR #3676, opened 2025-09-30
VK Layer Settings remove deprecated

Niram7777:


squidbus:
Can you link to any information on these layer settings being deprecated as I cannot find any indication of it.

Niram7777:
https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/f28c0df8a5170cbcc66f7cb6b49674f2c13983e7

https://github.com/KhronosGroup/VulkanSC-ValidationLayers/blob/sc_main/layers/VkLayer_khronos_validation.json.in#L477
```
"status": "DEPRECATED",
```

https://github.com/KhronosGroup/VulkanSC-ValidationLayers/blob/sc_main/layers/layer_options.cpp#L143-L144
```
// These were deprecated after the 1.3.296 SDK release (because it was a flag and now is a boolean)
...
// Plan to remove after deprecated for July 2025 SDK
```

I couldn't find the exact matching option but it is replaced by the `gpuav_enable` apparently

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -300,34 +298,13 @@ vk::UniqueInstance CreateInstance(Frontend::WindowSystemType window_type, bool e
             .valueCount = 1,
             .pValues = &enable_sync,
         },
-        vk::LayerSettingEXT{
-            .pLayerName = VALIDATION_LAYER_NAME,
-            .pSettingName = "validate_gpu_based",
```
squidbus, line 310: For this - if it's replaced by `gpuav_enable`, should probably update it instead of removing completely.

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -300,34 +298,13 @@ vk::UniqueInstance CreateInstance(Frontend::WindowSystemType window_type, bool e
...
         vk::LayerSettingEXT{
             .pLayerName = VALIDATION_LAYER_NAME,
             .pSettingName = "gpuav_descriptor_checks",
             .type = vk::LayerSettingTypeEXT::eBool32,
             .valueCount = 1,
             .pValues = &enable_gpuav,
         },
-        vk::LayerSettingEXT{
-            .pLayerName = VALIDATION_LAYER_NAME,
-            .pSettingName = "gpuav_validate_indirect_buffer",
```
squidbus, line 331: Similar here - there seem to be a few indirect buffer related settings added at the same time as this was deprecated, so should likely be replaced.

src/video_core/renderer_vulkan/vk_platform.cpp
Niram7777: the issue I had when activating `gpuav_enable` was  that normal Core validation and GPU-AV got both activated

https://vulkan.lunarg.com/doc/view/latest/windows/gpu_validation.html
```
Note - it is highly recommended to not have both normal Core validation and GPU-AV on together as performance will be slow.
```

I will check if it is fine to have both or if I should keep the minimum

src/video_core/renderer_vulkan/vk_platform.cpp
Niram7777: I will try to find the equivalent, but it is not always the case apparently, there is a profile called "GPU-AV" maybe I can start with that but deactivate some settings

src/video_core/renderer_vulkan/vk_platform.cpp
Niram7777: I have added gpuav_enable and created a core variable to be able separate the two if needed

src/video_core/renderer_vulkan/vk_platform.cpp
Niram7777: I have added the different indirect buffers but they are on by default when gpuav_enable is on

squidbus (approved):

PR #3677, opened 2025-09-30
Lock graphics queue MultipleThreads-Write: vkQueueSubmit(): THREADING…

Niram7777:
… ERROR : object of type VkQueue is simultaneously used in current thread...


squidbus:
I would question more here on why the submits are happening simultaneously in multiple threads, rather than just sticking a lock on it.

Niram7777:
closing since not able to reproduce the issue with latest verion

PR #3678, opened 2025-09-30
Lock device vkDeviceWaitIdle(): THREADING ERROR : object of type VkDe…

Niram7777:
…vice is simultaneously used in current thread...

### DIFF
--- a/src/imgui/renderer/imgui_core.cpp
+++ b/src/imgui/renderer/imgui_core.cpp
@@ -20,6 +20,12 @@
 #include "imgui_fonts/notosansjp_regular.ttf.g.cpp"
 #include "imgui_fonts/proggyvector_regular.ttf.g.cpp"
 
+namespace Vulkan {
+
+extern std::mutex g_mutex_device;
+
+}
+
 static void CheckVkResult(const vk::Result err) {
     LOG_ERROR(ImGui, "Vulkan error {}", vk::to_string(err));
 }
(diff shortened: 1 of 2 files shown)


squidbus:
Judging by where you placed the locks, this seems like it could only happen if HDR is being set up at the same time as the emulator is being exited? Seems like quite an edge case, and aside from that would rather not have this new mutex global.

Niram7777:
I was not using HDR.
I was testing with Wipeout and GOW3 with the following settings:
1920 x 1080
present mode FIFO
Vulkan validation and crash diagnostic activated
linux wayland with Nvidia RTX PRO 6000

Both for this PR and this one https://github.com/shadps4-emu/shadPS4/pull/3677 I was using this branch trying to get around the sampling issue
https://github.com/Niram7777/shadPS4/tree/fix-vulkan
I have just extracted the interesting commits and rebased out of my mess if it could help others.

If you prefer to face the issue before merging I understand, don´t hesitate to close

squidbus:
It's a similar sort of thing here to the other PR - would like to understand why the threading errors are occurring and if there's anything to be done about that before just sticking a lock on it.

squidbus:
I just moved the HDR set part to the present thread so maybe this is gone?

Niram7777:
> I just moved the HDR set part to the present thread so maybe this is gone?

I am not using HDR

I will investigate tomorrow if I can reproduce the error with your branch, i will give you a backtrace otherwise I will close my PR

Niram7777:
closing since not able to reproduce the issue with latest verion

PR #3680, opened 2025-09-30
IME: guard null params for CUSA04909

w1naenator:
- Add null checks in IME constructors to prevent crashes seen in CUSA04909.
- Leave a clear note about deferring keyboard event dispatch until guest-space translation is ready.


StevenMiller123:
Fixes the regression in Metaphor: ReFantazio.
[CUSA47038.log](https://github.com/user-attachments/files/22624302/CUSA47038.log)

w1naenator:
Fixed the IME pop-up in GE2. The game now behaves the same as before my IME changes; it still doesn't work.
It hangs when text input is completed or canceled.

PR #3681, opened 2025-09-30
image: Improve enforcement of image copy layer rules.

squidbus:
Previous go at this did not account for things like 2D<->3D copies where layer count should equal depth. This version adds more robust copy sanitization following all of the Vulkan restriction cases.


StevenMiller123:
Seems like this fixes a regression in Kingdom Hearts III on Intel GPUs caused by https://github.com/shadps4-emu/shadPS4/pull/3673. The game broke to just showing a black screen, now it's back to the typical UE mess.

[CUSA12031.log](https://github.com/user-attachments/files/22628263/CUSA12031.log)

PR #3682, opened 2025-10-01
Kernel.Fs: Device file cleanup and /dev/rng implementation

StevenMiller123:
This PR does two major things, cleaning up various issues with device files, and implementing RngDevice.
For the former, I've adjusted includes to better suit our current code preferences, fixed various incorrect function types, and adjusted the Device Create functions to use similar logic to the Directory Create functions.

For the latter, I've implemented the RngDevice, which is required for libSceSsl2 to run LLE (I have not actually LLE'd the library here, but this fixes internal library issues when trying to run it LLE).
The actual implementation for RngDevice ioctls is based entirely around code from fpPS4, though I've confirmed that libSceSsl2 works with what I've implemented.

I'm open to feedback, especially with my ioctls for the RngDevice, as I'm not too confident in how that code looks.

### DIFF
--- a/src/core/file_sys/devices/rng_device.cpp
+++ b/src/core/file_sys/devices/rng_device.cpp
@@ -1,10 +1,9 @@
 // SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
-#include <cstdlib>
 #include <ctime>
 #include "common/logging/log.h"
-#include "rng_device.h"
+#include "core/file_sys/devices/rng_device.h"
 
 namespace Core::Devices {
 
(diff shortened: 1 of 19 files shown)


squidbus:
For the build issue, you need to restore the `#include <cstdlib>` to be in the following files:
```
random_device.cpp
rng_device.cpp
urandom_device.cpp
srandom_device.cpp
```

src/core/file_sys/devices/rng_device.cpp
```diff
@@ -0,0 +1,86 @@
...
+    return std::static_pointer_cast<BaseDevice>(std::make_shared<RngDevice>(handle));
+}
+
+s32 RngDevice::ioctl(u64 cmd, Common::VaCtx* args) {
+    LOG_INFO(Kernel_Pthread, "called, cmd = {:#x}", cmd);
+    // Both commands are for generating a random number
+    if (cmd == 0x40445301 || cmd == 0x40445302) {
+        auto& data = *vaArgPtr<GetRandomArgs>(&args->va_list);
+        data.result = 0;
+        for (u64 i = 0; i < 16; i++) {
```
red-prig: The result buffer is 64 bytes, not 16.

PR #3683, opened 2025-10-01
Update Qt to 6.9.3

Xphalnos:
Update Qt to 6.9.3

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,14 +17,14 @@ jobs:
     runs-on: ubuntu-24.04
     continue-on-error: true
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
     - uses: fsfe/reuse-action@v5
 
   clang-format:
     runs-on: ubuntu-24.04
     continue-on-error: true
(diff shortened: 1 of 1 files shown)


Xphalnos:
macOS CIs seems to have internal errors

squidbus:
> macOS CIs seems to have internal errors

It seems like GitHub Actions is having issues with capacity, it's not getting a runner in time. Seems fine since I got Qt job to run though.

kalaposfos13:
Shouldn't this PR be submitted to https://github.com/shadps4-emu/shadps4-qtlauncher instead?

squidbus:
> Shouldn't this PR be submitted to https://github.com/shadps4-emu/shadps4-qtlauncher instead?

I'm not opposed to applying most of these CI updates here but yes Qt version update should also go to the Qt launcher.

georgemoralis:
i think is better not updating qt relative stuff , since soon qt actions will be removed

squidbus:
> i think is better not updating qt relative stuff , since soon qt actions will be removed

Yeah it's just there are other action updates here that are useful.

PR #3688, opened 2025-10-01
Add CLI argument to launch the emulator with global config or with default settings

kalaposfos13:
```
  --config-clean           Run the emulator with the default config values, ignores the config file(s) entirely.
  --config-global          Run the emulator with the base config file only, ignores game specific configs.
  ```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -37,6 +37,7 @@ int main(int argc, char* argv[]) {
     std::optional<std::filesystem::path> game_folder;
 
     bool waitForDebugger = false;
+    bool load_game_specific_config = true;
     std::optional<int> waitPid;
 
     // Map of argument strings to lambda functions
@@ -62,6 +63,10 @@ int main(int argc, char* argv[]) {
                     "parent of game path\n"
                     "  --wait-for-debugger           Wait for debugger to attach\n"
                     "  --wait-for-pid <pid>          Wait for process with specified PID to stop\n"
+                    "  --config-clean                Run the emulator with the default config "
+                    "values, ignores the config file(s) entirely.\n"
+                    "  --config-global               Run the emulator with the base config file "
+                    "only, ignores game specific configs.\n"
                     "  -h, --help                    Display this help message\n";
              exit(0);
          }},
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/main.cpp [hunk 1]
```diff
@@ -151,6 +156,12 @@ int main(int argc, char* argv[]) {
              exit(0);
          }},
         {"--log-append", [&](int& i) { Common::Log::SetAppend(); }},
+        {"--config-clean",
+         [&](int& i) {
+             Config::setDefaultValues(/*is_game_specific*/ false);
```
squidbus: Just to confirm: is there a possibility the default values would save and overwrite the config?

kalaposfos13: I don't see how that could happen, and testing showed no signs of it happening, but it might be possible to somehow achieve that via the GUI, that part I don't know.

kalaposfos13: There is a place where the config gets automatically updated if the commit value in it isn't the same as the emulator's version, but that runs before the arguments get parsed, so it will not affect this.

squidbus: https://github.com/shadps4-emu/shadPS4/blob/8a8ee395c5dd31fd437e1f39541751077f193bbe/src/core/devtools/layer.cpp#L108-L115

kalaposfos13: The latest push should address this issue, at the cost of the devtools menu not being able to update settings live, if the user does a clean run.

squidbus (approved):

src/main.cpp
```diff
@@ -249,7 +255,6 @@ int main(int argc, char* argv[]) {
     // Run the emulator with the resolved eboot path
     Core::Emulator* emulator = Common::Singleton<Core::Emulator>::Instance();
     emulator->executableName = argv[0];
-    emulator->waitForDebuggerBeforeRun = waitForDebugger;
```
Niram7777, line 252: I think it should be kept, we could still need to attach with some dbg tools

kalaposfos13: Oops this was not intentional, my bad.

PR #3689, opened 2025-10-01
audioout: Do not wait for data within timer.

squidbus:
Should fix cases where games not outputting audio data throws off the timer compensation and speeds up the next audio.

### DIFF
--- a/src/core/libraries/audio/audioout.cpp
+++ b/src/core/libraries/audio/audioout.cpp
@@ -301,7 +301,7 @@ static void AudioOutputThread(PortOut* port, const std::stop_token& stop) {
         timer.Start();
         {
             std::unique_lock lock{port->mutex};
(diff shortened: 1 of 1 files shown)


roamic (approved):

georgemoralis:
every game that doesn't have audio in videos (and general ) without using some strange audio lib should be tested

SixSeven-off:
is this also for gow3? That has audio issues too, which Lib has audio and should be removed?

squidbus:
> is this also for gow3? That has audio issues too, which Lib has audio and should be removed?

This fixes only cases where after a period of no audio playing, new audio playing starts sped up, usually accompanied by SDL queue full log messages.

georgemoralis:
fixed 2015 F1 quick race at least

εικόνα

PR #3690, opened 2025-10-01
videoout: Move HDR swapchain configuration to present thread.

squidbus:
On current main I'm hitting a crash when HDR is set up due to threading issues with recreating the Vulkan swapchain during present. Possibly due to recent changes to presenter logic.

Fixing this by moving the HDR swapchain re-configuration into the present thread. Note that we can safely call `SetHDR` each time as it is a no-op if the current HDR state is the same as input.


Exhigh:
Does this PR, replace PR #3678 ?

PR #3691, opened 2025-10-01
VK settings activation based on layers

Niram7777:
1. Fix nullptr log when accessing emulator GUI settings with extension missing causing signal handler to call stop (it would be better to setup logs before anything else, but here it is a quick fix).

2. Fix VK extension warning:
```
[Render.Vulkan] <Info> vk_platform.cpp:227 CreateInstance: Creating vulkan instance
[Render.Vulkan] <Info> vk_platform.cpp:183 operator(): Candidate instance extension VK_EXT_layer_settings is not available
[Render.Vulkan] <Info> vk_platform.cpp:270 CreateInstance: Enabled instance extensions: VK_KHR_wayland_surface, VK_KHR_surface, VK_EXT_debug_utils
```

Based on
- https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8760#issuecomment-2447462238
- https://github.com/KhronosGroup/Vulkan-Samples/pull/1366/files#diff-590ca11c17edcf2236215b1326e168d8b2d515f0d4469af2cd22caf6c89b1f13
- https://registry.khronos.org/vulkan/specs/latest/man/html/vkEnumerateInstanceExtensionProperties.html#_description

Normal start
```
[Render.Vulkan] <Info> vk_platform.cpp:256 CreateInstance: Creating vulkan instance
[Render.Vulkan] <Info> vk_platform.cpp:299 CreateInstance: Enabled instance extensions: VK_KHR_wayland_surface, VK_KHR_surface, VK_EXT_swapchain_colorspace, VK_EXT_debug_utils, VK_EXT_layer_settings
[Render.Vulkan] <Info> vk_platform.cpp:300 CreateInstance: Enabled instance layers: VK_LAYER_KHRONOS_validation, VK_LAYER_LUNARG_crash_diagnostic
[Render.Vulkan] <Info> vk_instance.cpp:105 Instance: Found 2 physical devices
[Render.Vulkan] <Info> vk_instance.cpp:651 CollectDeviceParameters: GPU_Vendor: NVIDIA
[Render.Vulkan] <Info> vk_instance.cpp:652 CollectDeviceParameters: GPU_Model: NVIDIA RTX PRO 6000 Blackwell Workstation Edition
[Render.Vulkan] <Info> vk_instance.cpp:653 CollectDeviceParameters: GPU_Integrated: No
[Render.Vulkan] <Info> vk_instance.cpp:654 CollectDeviceParameters: GPU_Vulkan_Driver: NVIDIA 580.82.9.0
[Render.Vulkan] <Info> vk_instance.cpp:655 CollectDeviceParameters: GPU_Vulkan_Version: 1.4.312
```

With RenderDoc
```
[Render.Vulkan] <Info> vk_platform.cpp:256 CreateInstance: Creating vulkan instance
[Render.Vulkan] <Error> vk_platform.cpp:210 GetInstanceExtensions: Settings for layer VK_LAYER_KHRONOS_validation not available.
[Render.Vulkan] <Error> vk_platform.cpp:210 GetInstanceExtensions: Settings for layer VK_LAYER_LUNARG_crash_diagnostic not available.
[Render.Vulkan] <Info> vk_platform.cpp:299 CreateInstance: Enabled instance extensions: VK_KHR_xlib_surface, VK_KHR_surface, VK_EXT_swapchain_colorspace, VK_EXT_debug_utils
[Render.Vulkan] <Info> vk_platform.cpp:300 CreateInstance: Enabled instance layers: VK_LAYER_KHRONOS_validation, VK_LAYER_LUNARG_crash_diagnostic
[Render.Vulkan] <Info> vk_instance.cpp:105 Instance: Found 3 physical devices
[Render.Vulkan] <Info> vk_instance.cpp:651 CollectDeviceParameters: GPU_Vendor: NVIDIA
[Render.Vulkan] <Info> vk_instance.cpp:652 CollectDeviceParameters: GPU_Model: NVIDIA RTX PRO 6000 Blackwell Workstation Edition
[Render.Vulkan] <Info> vk_instance.cpp:653 CollectDeviceParameters: GPU_Integrated: No
[Render.Vulkan] <Info> vk_instance.cpp:654 CollectDeviceParameters: GPU_Vulkan_Driver: NVIDIA 580.82.9.0
[Render.Vulkan] <Info> vk_instance.cpp:655 CollectDeviceParameters: GPU_Vulkan_Version: 1.4.0
```

I am guessing RenderDoc is messing with layer path to not have my VK settings available there

The next improvement would be to use vk_layer_settings.txt instead IMO
https://vulkan.lunarg.com/doc/sdk/1.4.321.1/windows/layer_configuration.html

### DIFF
--- a/src/video_core/renderer_vulkan/vk_platform.cpp
+++ b/src/video_core/renderer_vulkan/vk_platform.cpp
@@ -123,6 +123,37 @@ vk::SurfaceKHR CreateSurface(vk::Instance instance, const Frontend::WindowSDL& e
     return surface;
 }
 
+static auto GetLayerExtensions(std::vector<const char*>&& extensions,
+                               const std::vector<const char*>& layers) {
+    auto all_missing_vk_settings = true;
+
+    for (const auto& layer_name : layers) {
+        const auto [layer_properties_result, layer_extensions] =
+            vk::enumerateInstanceExtensionProperties(std::string(layer_name));
+        if (layer_properties_result != vk::Result::eSuccess) {
+            LOG_ERROR(Render_Vulkan, "Failed to query extension properties of {}: {}",
+                      layer_name,
+                      vk::to_string(layer_properties_result));
+        }
+        auto found = false;
+        for (const auto& extension : layer_extensions) {
+            if (extension.extensionName == std::string_view(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME)) {
+                found = true;
+                all_missing_vk_settings = false;
+            }
+        }
+        if (!found) {
+            LOG_ERROR(Render_Vulkan, "Settings for layer {} not available.", layer_name);
+        }
+    }
+
+    if (!all_missing_vk_settings) {
+        extensions.push_back(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME);
+    }
+
+    return extensions;
+}
+
 std::vector<const char*> GetInstanceExtensions(Frontend::WindowSystemType window_type,
                                                bool enable_debug_utils) {
     const auto [properties_result, properties] = vk::enumerateInstanceExtensionProperties();
@@ -135,7 +166,6 @@ std::vector<const char*> GetInstanceExtensions(Frontend::WindowSystemType window
     // Add the windowing system specific extension
     std::vector<const char*> extensions;
     extensions.reserve(7);
-    extensions.push_back(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME);
 
     switch (window_type) {
     case Frontend::WindowSystemType::Headless:
@@ -258,7 +288,8 @@ vk::UniqueInstance CreateInstance(Frontend::WindowSystemType window_type, bool e
                VK_VERSION_MAJOR(TargetVulkanApiVersion), VK_VERSION_MINOR(TargetVulkanApiVersion),
                VK_VERSION_MAJOR(available_version), VK_VERSION_MINOR(available_version));
 
-    const auto extensions = GetInstanceExtensions(window_type, true);
+    const auto layers = GetInstanceLayers(enable_validation, enable_crash_diagnostic);
+    const auto extensions = GetLayerExtensions(GetInstanceExtensions(window_type, true), layers);
 
     const vk::ApplicationInfo application_info = {
         .pApplicationName = "shadPS4",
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -191,6 +191,32 @@ std::vector<const char*> GetInstanceExtensions(Frontend::WindowSystemType window
...
+            }
+        }
+        if (!found) {
+            LOG_ERROR(Render_Vulkan, "Settings for layer {} not available.", layer_name);
+        }
+    }
+
+    if (!all_missing_vk_settings) {
+        extensions.push_back(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME);
+    }
```
squidbus: I don't really see the need to do all this, all this seems (from the PR description) to be to avoid the warning print when it is not found, but we want that information to be in the log.

src/common/logging/backend.cpp
```diff
@@ -156,10 +156,16 @@ class Impl {
     }
 
     static void Start() {
+        if (!IsActive()) {
+            Log::Initialize();
```
squidbus: I don't think this is a suitable workaround because then something could silently initialize the logger with no log file.

Niram7777: I will remove the changes of this file, and we can discuss/redesign another time

src/video_core/renderer_vulkan/vk_platform.cpp
Niram7777: the issue is that enumerateInstanceExtensionProperties never provides VK_EXT_layer_settings (because explicit layer)

its up to each layer to say if it supports it, we can not know before (unless forcing with vkconfig)

the first link also explains for example the design is the same for VK_EXT_debug_utils on android which is not implicit layer

tldr: we always get not found when actually it is there, activated by the validation/CDL layers
-the not found print is too early and not retrieving explicit layers

let me know if still not clear, and dont hesitate to read the 3 links, it explains quite well

squidbus: Your provided new logs still show it not being enabled at all though with RenderDoc. So what is this fixing?

squidbus: I think I understand the problem better now after further reading. But I think this could be organized a bit different, for example maybe split this out to a ‘GetLayerExtensions’ and then add on to the extension list using that in ‘CreateInstance’.

squidbus (approved):

PR #2872, opened 2025-05-02
Core: Memory Fixes

StevenMiller123:
This PR makes our memory code more robust, preventing a variety of exceptions and strange behavior I observed when using my recently shared memory stress test homebrew on shadPS4.

- Fixes `OrbisVirtualQueryInfo` struct. My homebrew was reading garbage data with the BitFields and array in place. I've replaced the BitField class use with C-style bitfields, and the std::array with a primitive char array instead.
- Proper low address handling in `sceKernelVirtualQuery`. Our previous code didn't check if the inputted address was below our vma map, which led to OOB memory accesses. My edge case error returns are based on hardware tests here.
- Edge case handling for `SearchFree`. Our previous code encountered unpredictable behavior if the inputted address was too high, or if there were no free memory areas to map.
- Changed default nameless memory mapping behavior. On real hardware, nameless mappings will have the name "anon" followed by the code address responsible for the memory call. For simplicity sake, I just used "anon" as the default name instead.
- Fixes a variety of broken memory asserts, caused by a fundamental misunderstanding of how `FindVMA` and `FindDmemArea` behave.
- Fixes a long unnoticed regression caused by #2080 
- Fixes an issue where `MapMemory` would assert when the area to map was only partially reserved.

With these fixes, my memory homebrew now executes all tests on shadPS4. Many of the returned results are still inaccurate, but these are due to deeper issues that I believe are out of scope for this PR.

Currently:
- Regresses EA SPORTS™ UFC® (CUSA00264) 
     - The only reason this game works in main is due to a bug in our code. Fixing this bug broke the game, and attempting to properly fix the regression broke Windows support.
- Improves Evolve (CUSA00432)
- Improves Cars 3: Driven to Win (CUSA07027)
- Improves Shadow of the Tomb Raider (CUSA10872)
- Improves The Order: 1886 (CUSA00076)
- Improves Final Fantasy XV (CUSA01633)

PR #3697, opened 2025-10-03
Core: Increase address space limits and rework Windows address space initialization.

StevenMiller123:
Current Windows builds of shadPS4 are prone to asserting in address_space on emulator initialization due to Windows' address space randomization. This also prevents expanding the address space to higher addresses (which is needed for a handful of EA titles).

This PR introduces logic to detect and map around system mappings on Windows devices, pass information about these gaps to our memory code, and more robust address validation logic to ensure gaps are accounted for when detecting if requested memory areas are within the vma map.

For Mac users: this PR should make asserts trigger more consistently when games try mapping to addresses that we can't reserve.

For Windows users: this PR should fix those random-ish cases of
```
[Debug] <Critical> address_space.cpp:105 operator(): Assertion Failed!
Unable to reserve virtual address space: Not enough memory resources are available to process this command.
```

For all users: this PR should address cases of the following assert triggering on addresses greater than 0x11000000000 (mainly seen in earlier titles published by EA).

```
[Debug] memory.cpp:902 operator(): Assertion Failed!
Input address .... is out of bounds
```
I'm opening this as a draft for now since I need to test edge cases further, and ideally make sure our own HLE mappings aren't going to hit asserts in the more extreme cases.

I would also appreciate some tests from the aforementioned EA games, the only thing I have hitting that out of bounds assert is an app, and it dies from a Windows-specific issue with code patches.

### DIFF
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -72,11 +72,8 @@ static int sdk_version{0};
 
 static u32 asc_next_offs_dw[Liverpool::NumComputeRings];
 
-// We need to get an instance of our address space to derive a proper ring address here.
-auto* memory = Core::Memory::Instance();
-auto& address_space = memory->GetAddressSpace();
-static VAddr tessellation_factors_ring_addr = address_space.SystemReservedVirtualBase() +
-                                              address_space.SystemReservedVirtualSize() - 0xFFFFFFF;
+// This address is initialized in sceGnmGetTheTessellationFactorRingBufferBaseAddress
+static VAddr tessellation_factors_ring_addr = -1;
 static constexpr u32 tessellation_offchip_buffer_size = 0x800000u;
 
 static void ResetSubmissionLock(Platform::InterruptId irq) {
@@ -1002,7 +999,14 @@ int PS4_SYSV_ABI sceGnmGetShaderStatus() {
 }
 
 VAddr PS4_SYSV_ABI sceGnmGetTheTessellationFactorRingBufferBaseAddress() {
+    auto* memory = Core::Memory::Instance();
+    auto& address_space = memory->GetAddressSpace();
     LOG_TRACE(Lib_GnmDriver, "called");
+    if (tessellation_factors_ring_addr == -1) {
+        tessellation_factors_ring_addr = address_space.SystemReservedVirtualBase() +
+                                         address_space.SystemReservedVirtualSize() - 0xFFFFFFF;
+    }
+
     return tessellation_factors_ring_addr;
 }
 
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -17,6 +17,7 @@
 #include "core/libraries/kernel/process.h"
 #include "core/libraries/libs.h"
 #include "core/libraries/videoout/video_out.h"
+#include "core/memory.h"
 #include "core/platform.h"
 #include "video_core/amdgpu/liverpool.h"
 #include "video_core/amdgpu/pm4_cmds.h"
@@ -70,7 +71,12 @@ static bool send_init_packet{true}; // initialize HW state before first game's s
 static int sdk_version{0};
 
 static u32 asc_next_offs_dw[Liverpool::NumComputeRings];
-static constexpr VAddr tessellation_factors_ring_addr = Core::SYSTEM_RESERVED_MAX - 0xFFFFFFF;
+
+// We need to get an instance of our address space to derive a proper ring address here.
+auto* memory = Core::Memory::Instance();
+auto& address_space = memory->GetAddressSpace();
+static VAddr tessellation_factors_ring_addr = address_space.SystemReservedVirtualBase() +
+                                              address_space.SystemReservedVirtualSize() - 0xFFFFFFF;
 static constexpr u32 tessellation_offchip_buffer_size = 0x800000u;
 
 static void ResetSubmissionLock(Platform::InterruptId irq) {
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -999,10 +999,10 @@ int PS4_SYSV_ABI sceGnmGetShaderStatus() {
 }
 
 VAddr PS4_SYSV_ABI sceGnmGetTheTessellationFactorRingBufferBaseAddress() {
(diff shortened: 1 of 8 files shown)


Missake212:
EA SPORTS™ UFC® doesn't crash on the memory assert anymore.

[CUSA00264.log](https://github.com/user-attachments/files/22678369/CUSA00264.log)

src/core/memory.cpp
```diff
@@ -138,9 +132,9 @@ void MemoryManager::CopySparseMemory(VAddr virtual_addr, u8* dest, u64 size) {
 }
 
 bool MemoryManager::TryWriteBacking(void* address, const void* data, u32 num_bytes) {
-    ASSERT_MSG(IsValidAddress(address), "Attempted to access invalid address {}",
-               fmt::ptr(address));
     const VAddr virtual_addr = std::bit_cast<VAddr>(address);
+    ASSERT_MSG(IsValidMapping(virtual_addr, num_bytes), "Attempted to access invalid address {}",
```
Exhigh:
```suggestion
    ASSERT_MSG(IsValidMapping(virtual_addr, num_bytes), "Attempted to access invalid address {:#x}",
```

StevenMiller123:
I've abused AMD's Windows drivers to test under extreme circumstances and everything seems to work as expected there (most of the system reserved area and a chunk of the user area is completely inaccessible if those drivers initialize before us).
Games that try performing fixed mappings in that inaccessible space hit asserts, but that's intended behavior in this PR.

Since it seems to work fine for me, I'm marking this as ready for review.

StevenMiller123:
Linux is not a fan of the increased user size, so marking as draft while I figure this out.

StevenMiller123:
Latest push is just to fix Linux on my end. From what I've seen in logs, nothing should regress from this change.

StevenMiller123:
Seems like Mac doesn't behave properly with the increased user size. Leaving this as a draft until that can be addressed.

StevenMiller123:
Windows 10 devices seem to completely hang with this PR. Until I've fixed that issue, I suggest Windows 10 users avoid testing this.

StevenMiller123:
The Windows issue comes down to either dropping Windows 10 support entirely or breaking a couple EA titles. Needless to say, I've chosen the latter option.

I need Windows 10 testers to test for performance issues, that may still end up a problem with this current push.

StevenMiller123:
UFC 1 should still work, hopefully.

StevenMiller123:
This probably also fixes any regressions on Mac OS too, I believe this level of address space expanding was tested on there during #2872

StevenMiller123:
I've reverted the "fun" part of this PR for now due to limitations with Windows 10.

The improved logic will still help for Windows users hitting those random asserts on emulator startup, and make asserts more consistent for Mac users. Games like UFC will be back to their compatibility on main though.

StevenMiller123:
With the help of @squidbus, the high user max should be working for Mac OS devices now.
I've also added checks to allow high user max for Windows 11 devices, and prevent it on Windows 10 devices where VirtualAlloc2 is prone to hanging.

I still need to take a better look at the issues with higher user max on Linux, so this will remain a draft for now. I also need someone to test Windows 10 behavior.

src/core/libraries/gnmdriver/gnmdriver.cpp
```diff
@@ -70,7 +71,12 @@ static bool send_init_packet{true}; // initialize HW state before first game's s
 static int sdk_version{0};
 
 static u32 asc_next_offs_dw[Liverpool::NumComputeRings];
-static constexpr VAddr tessellation_factors_ring_addr = Core::SYSTEM_RESERVED_MAX - 0xFFFFFFF;
+
+// We need to get an instance of our address space to derive a proper ring address here.
+auto* memory = Core::Memory::Instance();
+auto& address_space = memory->GetAddressSpace();
+static VAddr tessellation_factors_ring_addr = address_space.SystemReservedVirtualBase() +
+                                              address_space.SystemReservedVirtualSize() - 0xFFFFFFF;
```
squidbus: Not sure if it's safe to initialize this statically, may be safer to initialize this on first `sceGnmGetTheTessellationFactorRingBufferBaseAddress` call.

StevenMiller123:
With the newest changes, all OSes now load the shadPS4 executable at address `0x700000000000`, which allows Linux to use the full USER_MAX we've defined, and helps reserve some extra lower addresses on the Windows side.

This needs testing on Windows 10 though, I don't know if the same VirtualAlloc2 performance issue will end up applying to this too.

StevenMiller123:
Local tests suggest Windows 10 works fine, and that my Windows 10 detection code is functioning.
Just need to fix Linux actions now

StevenMiller123:
I've stuck to just reducing USER_MAX on Linux for now, this shouldn't impact too much since it's still high enough to include the rather extreme fixed mappings EA's titles perform.

Like with the Windows 10 issue, I'm open to changing this if anyone's able to suggest a working solution.

StevenMiller123:
Everything seems to be working properly on my end, so I'm marking this as ready for review.

StevenMiller123:
Given the variety of games hitting this particular memory assert, here's a list of games that this PR should affect based on compatibility reports:
- Need for Speed™
- FIFA 14
- NHL™ 15
- Madden NFL 16
- FIFA 16
- STAR WARS™ Battlefront™ - Only version 1.00 was affected.
- EA SPORTS™ UFC® - Confirmed improvements.

None of these will improve on Windows 10, since I've had to restrict the user size to accomodate OS bugs there. Windows 11, Mac OS, and Linux should all see improvements though.

This PR should also improve similar Windows-specific issues (for both Windows 10 and Windows 11) in:
- Saints Row: Gat out of Hell
- Saints Row IV: Re-Elected

These two games were having issues due to the old Windows address space logic, my tests suggest the addresses they're trying to use should be in the vma_map with this PR now.

Missake212:
Can confirm EA SPORTS™ UFC® status is still improved by this PR.

src/core/libraries/gnmdriver/gnmdriver.cpp
```diff
@@ -996,7 +999,14 @@ int PS4_SYSV_ABI sceGnmGetShaderStatus() {
 }
 
 VAddr PS4_SYSV_ABI sceGnmGetTheTessellationFactorRingBufferBaseAddress() {
+    auto* memory = Core::Memory::Instance();
+    auto& address_space = memory->GetAddressSpace();
```
squidbus: Can move these into the `if` case.

squidbus (commented):
Just one small comment

squidbus (approved):

PR #3698, opened 2025-10-03
Vk validation core on for game specific + optimize search extension

Niram7777:


src/common/config.cpp
```diff
@@ -1236,7 +1236,7 @@ void setDefaultValues(bool is_game_specific) {
     // GS - Vulkan
     gpuId.set(-1, is_game_specific);
     vkValidation.set(false, is_game_specific);
-    vkValidationCore.set(false, is_game_specific);
+    vkValidationCore.set(true, is_game_specific);
```
StevenMiller123, line 1239: We default to validation disabled because it reduces game performance.
I'd recommend just changing the setting in your game specific configuration instead.

StevenMiller123 (changes requested):

src/common/config.cpp
squidbus: This setting doesn’t take effect unless the one above it is also on, which defaults to false.

squidbus (approved):

PR #3700, opened 2025-10-03
image_info: Fix guest size calculation for linear render targets

raphaelthegreat:
The linear aligned tile mode doesn't require image height to be tile aligned only the pitch [per addrlib](https://github.com/chaotic-cx/mesa-mirror/blob/52c7b0d20c88620fd55d21a62468211674559978/src/amd/addrlib/src/r800/egbaddrlib.cpp#L666)

This is an issue with render targets because the registers that hold size information store it in units of tiles, so using these makes the guest size appear larger than it actually is. For example a 16128x1 RGBA32 render target should have a size of 258048 bytes (16128 * 16), but the color buffer registers will report 2064384 bytes (16128 * 8 * 16) which would be correct if the height were to be padded to tile dimension (8). The fix is manually computing the size if the render target is not tiled. Also removed handling of ArrayLinearGeneral because this doesn't align at all, so the current code is wrong for it anyway and seems entirely unused.

This should resolve some "Encountered unresolvable image overlap with equal memory address." asserts like [this](https://github.com/shadps4-compatibility/shadps4-game-compatibility/issues/1701) and might also fix random crashes when readback linear images is enabled, because with previous code it could overwrite additional memory, especially if the image size was 1x1 which isn't uncommon

This issue was reported to me by @Xcedf who also provided needed debugging information

### DIFF
--- a/src/video_core/texture_cache/image_info.cpp
+++ b/src/video_core/texture_cache/image_info.cpp
@@ -72,9 +72,15 @@ ImageInfo::ImageInfo(const AmdGpu::Liverpool::ColorBuffer& buffer,
     meta_info.fmask_addr = buffer.info.compression ? buffer.FmaskAddress() : 0;
 
     guest_address = buffer.Address();
-    const auto color_slice_sz = buffer.GetColorSliceSize();
-    guest_size = color_slice_sz * buffer.NumSlices();
-    mips_layout.emplace_back(guest_size, pitch, buffer.Height(), 0);
+    if (props.is_tiled) {
+        guest_size = buffer.GetColorSliceSize() * resources.layers;
+        mips_layout.emplace_back(guest_size, pitch, buffer.Height(), 0);
+    } else {
+        std::tie(std::ignore, std::ignore, guest_size) =
+            ImageSizeLinearAligned(pitch, size.height, num_bits, num_samples);
+        guest_size *= resources.layers;
+        mips_layout.emplace_back(guest_size, pitch, size.height, 0);
+    }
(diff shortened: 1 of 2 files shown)


squidbus (approved):
Yeah this lines up with what I remember hitting before. I remember adding the tile mode check from a case of this I hit but I don’t remember the exact details: https://github.com/shadps4-emu/shadPS4/pull/2981 Or maybe I'm misremembering what I hit this issue from.

raphaelthegreat:
Would be nice to check so the extra case can be removed if possible

squidbus:
> Would be nice to check so the extra case can be removed if possible

At least in the case of the linked PR I wasn't the one with the game in question. I think it was some PS1 game conversion?

Thinking about it more though I think that case remains valid as for overlapping linear and tiled textures, you will still have a difference in byte size with all other image parameters being the same.

locsucc:
Stray
[CUSA24899.txt](https://github.com/user-attachments/files/22691831/CUSA24899.txt)
[Debug] <Critical> texture_cache.cpp:375 ResolveOverlap: Unreachable code!

shinra-electric:
The Evil Within still hits it with this PR.
[CUSA00375.txt.zip](https://github.com/user-attachments/files/22691968/CUSA00375.txt.zip)
```
[Debug] <Critical> texture_cache.cpp:375 ResolveOverlap: Unreachable code!
```

Treshco:
It works for Need for Speed Rivals. I can reach in-game with this pr (with some small graphical issues).

cusa00168 t
cusa00168 ig

[CUSA00168.log](https://github.com/user-attachments/files/22698610/CUSA00168.log)

Xcedf:
Confirm battlefield 4 resolve overlap fixed

PR #3702, opened 2025-10-04
Fix saving audio config

kalaposfos13:


georgemoralis:
that should be fixed in qtlauncher i think..

PR #3703, opened 2025-10-05
NpManager: Implement more request-related behavior

StevenMiller123:
This PR adds implementations for `sceNpCreateAsyncRequest`, `sceNpCheckPlus`, `sceNpAbortRequest`, `sceNpPollAsync`, and `sceNpWaitAsync`.

I've also adjusted logic for storing np requests to accomodate async requests, moved them to np_manager.cpp since they're purely for internal use, and renamed them to better show that they're for internal use.

For the async functions, I've simulated their behavior for now. I suspect the behavior when connected to PSN might be innaccurate, but since we can't exactly connect to PSN on jailbroken consoles, I've just done what I can based on library decompilation.

I also added a mutex to the request-related functions, which should prevent some potential weirdness, and better matches the actual library's behavior too.

This could help improve titles that use `sceNpCreateAsyncRequest`, since games would receive unexpected error codes due to the stubs we had before. It might also help bypass more PSN checks in some titles when `isConnectedToPSN` is enabled, though most titles like this also use other Np libraries we don't have emulated yet.


src/core/libraries/np/np_manager.cpp
```diff
@@ -13,28 +15,87 @@ namespace Libraries::Np::NpManager {
...
+        NpRequest new_request{NpRequestState::None, false, 0};
+        g_requests.emplace_back(new_request);
+    }
+
+    // Offset by one, first returned ID is 0x20000001
+    g_active_requests++;
+    return req_index + ORBIS_NP_MANAGER_REQUEST_ID_OFFSET + 1;
+}
+
+s32 PS4_SYSV_ABI sceNpCreateAsyncRequest(const OrbisNpCreateAsyncRequestParameter* param) {
```
mikusp, line 71: Both CreateRequest functions do 80% of the same stuff, you could extract the common code to an internal function

mikusp (approved):

squidbus (approved):

PR #3705, opened 2025-10-05
CI: Make the Cache Functional

Xphalnos:
The `--fresh` option ignore the cache.

Pros: Much faster CI
Cons: More likely to have cache misses or conflicts when updating submodules or packages

Main:
image

PR:
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2024 shadPS4 Emulator Project
+# SPDX-FileCopyrightText: 2025 shadPS4 Emulator Project
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 name: Build and Release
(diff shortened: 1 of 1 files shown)


squidbus:
The fresh option was explicitly added because otherwise it was unstable.

Xphalnos:
> The fresh option was explicitly added because otherwise it was unstable.

So this means that we don't need the cache anymore?

squidbus:
> So this means that we don't need the cache anymore?

Not sure, `--fresh` forces it to reconfigure fresh but I figured that anything that configures to the same result might still benefit from the cache.

Niram7777 (commented):
LGTM
hard to judge on a single run but only Windows build step is slower.
any other step or platform is faster.
what would be useful IMO is to print the cache statistics at the end.
we could revert if facing previous issue again

georgemoralis:
there are different opinions about this , so closing

PR #3707, opened 2025-10-05
amdgpu: Split liverpool registers and cleanup

raphaelthegreat:
This is something I've been wanting for a while but never found motivation to do so. The liverpool header has grown pretty huge over time and because all the register structures and enums are nested inside it, they cannot be forward declared, forcing users to include the full heavyweight header to just use a small part of it. So the proposed patch does the following:

* Pull out and split register structures and enums into headers for every register category as defined in the AMD southern islands register reference manual: CB, DB, VGT, PA, TP, SPI/COMP (compute registers are just a single struct and share a lot of code with SPI)
* With above done, attempt to replace liverpool.h includes with forward declarations or inclusions of specific necessary headers

While implementing that I've also done some additional cleanup where appropriate

* Unified texture cache descriptors into a single ImageDesc with multiple constructors, simpler than a bunch of child classes.
* Split out resource structures from info.h in recompiler so image_info doesn't need to include all of it
* Replaced BitField usage in liverpool registers with normal bitfields. While the bitfield class has advantages, it is annoying not being able to see the values of bitfields when debugging
* Removed exceptions from recompiler entirely. The initial plan for those was to catch exceptions when recompiling shaders and skip the draw if anything failed but this never materialized and isn't useful anymore
* GetParams now avoids calling SearchBinaryInfo twice (on main, it gets called once from Code function and another from GetParams which also called Code.

Probably needs a ton of testing to ensure nothing broke


StevenMiller123:
Some brief tests of my games suggest nothing's broken, though there's quite a bit I can't test at the moment 😅

bigol83:
I checked every game i own and there are no regressions. Performance and compatibility is the same as main build for me.

squidbus (approved):

PR #3708, opened 2025-10-05
Add missing function exports for OpenOrbis

kalaposfos13:
OpenOrbis seems to use the libkernel-libkernel exports of posix functions, of which quite a few were missing from shadPS4. Also, posix_pthread_attr_getguardsize wasn't exported in any way.

### DIFF
--- a/src/core/libraries/kernel/threads/pthread.cpp
+++ b/src/core/libraries/kernel/threads/pthread.cpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include "common/assert.h"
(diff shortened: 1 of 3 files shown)


src/core/libraries/kernel/threads/pthread.cpp
```diff
@@ -681,6 +681,7 @@ void RegisterThread(Core::Loader::SymbolsResolver* sym) {
     LIB_FUNCTION("3PtV6p3QNX4", "libkernel", 1, "libkernel", posix_pthread_equal);
     LIB_FUNCTION("T72hz6ffq08", "libkernel", 1, "libkernel", posix_pthread_yield);
     LIB_FUNCTION("EI-5-jlq2dE", "libkernel", 1, "libkernel", posix_pthread_getthreadid_np);
+    LIB_FUNCTION("3eqs37G74-s", "libkernel", 1, "libkernel", posix_pthread_getthreadid_np);
```
StevenMiller123: Minor nit, but following the current formatting here, this export should be in the Posix-Kernel section instead of the Orbis section

PR #3722, opened 2025-10-08
Add stubbed get_authinfo

kalaposfos13:
The struct was taken from https://github.com/RPCSX/rpcsx/blob/master/orbis-kernel/include/orbis/AuthInfo.hpp.
The function is completely one-to-one with real hardware, as long as it is called from game code. If it gets called from an LLEd system library, then we shouldn't reject the call with EPERM, however, LLE HTTP works with Driveclub even like this, so I don't really care.

### DIFF
--- a/src/core/libraries/kernel/kernel.h
+++ b/src/core/libraries/kernel/kernel.h
@@ -44,6 +44,13 @@ struct SwVersionStruct {
     u32 hex_representation;
 };
 
+struct AuthInfoData {
+    u64 unk0;
+    u64 caps[4];
+    u64 attrs[4];
+    u64 ucred[8];
+};
+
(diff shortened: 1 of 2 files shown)


src/core/libraries/kernel/kernel.h
```diff
@@ -44,6 +44,13 @@ struct SwVersionStruct {
     u32 hex_representation;
 };
 
+struct AuthInfoData {
+    uint64_t unk0;
+    uint64_t caps[4];
+    uint64_t attrs[4];
+    uint64_t ucred[8];
```
Niram7777: u64 I guess in our code?

Niram7777 (approved):

PR #3727, opened 2025-10-11
Add IPC command for cheats

rainmakerv3:
This would allow cheats to work on qtlauncher (once the command is applied there as well).

Seems working with a locally modified launcher


viniciuslrangel:
This is the same as PATCH_MEMORY

rainmakerv3:
apparently I misunderstood how queued patches are applied, and this is not needed after all

PR #3730, opened 2025-10-12
Mark linux-sdl and macos-sdl binaries as executable

tomboylover93:
Title.
Should fix an issue described by @brad0demx [here](https://github.com/shadps4-emu/shadps4-qtlauncher/pull/96#issuecomment-3393805347) and on Discord regarding the Qt launcher and how SDL pre-release binaries for the aforementioned platforms aren't marked as executable, thus refusing to run.
Fixing it here during the packaging process as opposed to marking it as executable with a different ifdef for macOS on [shadps4-qtlauncher](https://github.com/shadps4-emu/shadps4-qtlauncher/blob/main/src/qt_gui/version_dialog.cpp) is more efficient.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,26 +17,26 @@ jobs:
     runs-on: ubuntu-24.04
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v5
-      - uses: fsfe/reuse-action@v5
+    - uses: actions/checkout@v5
+    - uses: fsfe/reuse-action@v5
 
   clang-format:
     runs-on: ubuntu-24.04
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v5
-        with:
-          fetch-depth: 0
-      - name: Install
-        run: |
-          wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
-          sudo add-apt-repository 'deb http://apt.llvm.org/noble/ llvm-toolchain-noble-19 main'
-          sudo apt update
-          sudo apt install clang-format-19
-      - name: Build
-        env:
-          COMMIT_RANGE: ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }}
-        run: ./.ci/clang-format.sh
+    - uses: actions/checkout@v5
+      with:
+        fetch-depth: 0
+    - name: Install
+      run: |
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+        sudo add-apt-repository 'deb http://apt.llvm.org/noble/ llvm-toolchain-noble-19 main'
+        sudo apt update
+        sudo apt install clang-format-19
+    - name: Build
+      env:
+        COMMIT_RANGE: ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }}
+      run: ./.ci/clang-format.sh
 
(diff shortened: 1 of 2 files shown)


squidbus:
Have you checked the results for this? I think they should already be executable, and your issue might be that zip tools don't have consistent support for attributes like Unix file permissions.

squidbus:
For example, I downloaded the current 0.11.0 release and they show and extract as executable for me already.

```
% unzip -Z shadps4-linux-sdl-0.11.0.zip 

Archive:  shadps4-linux-sdl-0.11.0.zip
Zip file size: 16702477 bytes, number of entries: 1
-rwxr-xr-x  4.5 unx 17185272 bl defN 25-Sep-18 07:08 Shadps4-sdl.AppImage
1 file, 17185272 bytes uncompressed, 16702323 bytes compressed:  2.8%

% unzip -Z shadps4-macos-sdl-0.11.0.zip

Archive:  shadps4-macos-sdl-0.11.0.zip
Zip file size: 18587192 bytes, number of entries: 3
-rw-r--r--  4.5 unx      179 bl defN 25-Sep-18 07:00 MoltenVK_icd.json
-rwxr-xr-x  4.5 unx  5738984 bl defN 25-Sep-18 07:00 libMoltenVK.dylib
-rwxr-xr-x  4.5 unx 47053008 bl defN 25-Sep-18 07:06 shadps4
3 files, 52792171 bytes uncompressed, 18586812 bytes compressed:  64.8%
```

brad0demx:
@squidbus For some reason the issue only affects pre-release builds downloaded from `/shadps4-emu/shadPS4/releases` (Which is where Qt Launcher gets its binaries from I believe), releases like 0.11.0, 0.10.0 etc don't have this issue. Downloading the same commit from the pre-release "branch" through `/shadps4-emu/shadPS4/actions` doesn't have the issue

squidbus:
> Downloading the same commit from the pre-release "branch" through `/shadps4-emu/shadPS4/actions` doesn't have the issue

Then the PR as-is won't fix the issue, as it impacts the zips uploaded to the actions artifacts. You will need to look at the pre-release upload specifically.

brad0demx:
Could marking the binaries as executable [here in build.yml](https://github.com/shadps4-emu/shadPS4/blob/main/.github/workflows/build.yml#L462) be a potential solution?

tomboylover93:
Should be fixed now.

squidbus (changes requested):
I can't tell the diff at all because you've re-indented the whole file.

tomboylover93:
Fixed that, Zed Editor didn't like the indentation for some reason.

.github/workflows/build.yml
```diff
@@ -459,6 +459,11 @@ jobs:
       uses: actions/download-artifact@v5
       with:
         path: ./artifacts
+
+    - name: Make SDL artifacts executable
+      run: |
+        chmod a+x ./artifacts/shadps4-linux-sdl-*
+        chmod a+x ./artifacts/shadps4-macos-sdl-*
```
squidbus: It seems here like you are changing the permissions on the directory and not the executables inside (the appimage/binary)

squidbus: To clarify: I think you would either need `-R` for recursive or to do the path to the binary files.

squidbus (approved):
Since PR CI doesn't run pre-release, guess we'll see if it works.

squidbus:
Seems to be good now.

PR #3731, opened 2025-10-12
Ime fixes

w1naenator:
- Massive Ime library cleaning
- Implemented Ime UI input field (work buffer) text and caret sync with game using events
- Added data payload to enter and close events.


w1naenator:
Now text deletion no longer breaks text input.
After adding the input data payload to the Enter and Close events, it seems to have fixed the IME issues in God Eater 2, but the game still cannot exit the menus and gets stuck on a black screen.
On my side, I haven’t found any regressions.

NieR: Automata and Metaphor: ReFantazio also use IME, but I don’t have these games. Can someone test them?

StevenMiller123:
I'll test Metaphor in a moment here.
Someone else will need to test NieR, I don't have that one.

StevenMiller123:
Seems like Metaphor: ReFantazio still works.
image
image

Missake212:
I don't have NieR Automata, but NieR Replicant also uses IME and it still works.

image

[CUSA18742.log](https://github.com/user-attachments/files/22873546/CUSA18742.log)

w1naenator:
image

I have owned God Eater 2, it now receive input and do not freezes on enter or close event.
There is one strange thing, in parameter max text size is 8, but game show only 7 in first field max size in second is 10, but typed can be only 9.

w1naenator:
> I don't have NieR Automata, but NieR Replicant also uses IME and it still works.
> 
> image
> [CUSA18742.log](https://github.com/user-attachments/files/22873546/CUSA18742.log)

Seams this one uses ImeDialog, because Ime do not have title.

brad0demx:
image
Nier Automata popup works but seems to get desynced if I type something and remove it after

w1naenator:
> image Nier Automata popup works but seems to get desynced if I type something and remove it after

Can you provide log?

brad0demx:
[CUSA04480.log](https://github.com/user-attachments/files/22873617/CUSA04480.log)

w1naenator:
Recovered accidental code replacement while pull request preparing and code cleaning. Now  deletion and typing may work correctly.
Next step is limit UI text lenght.

w1naenator:
@brad0demx, please check again.

brad0demx:
image
Works better now, adding and deleting works as expected.
[CUSA04480.log](https://github.com/user-attachments/files/22874399/CUSA04480.log)

No idea why the log won't embed, opening the github link in a new tab works

w1naenator:
Text input length will not resolve the problem of overly long text entries, because the game does not strictly enforce its own limits. According to the provided log, Metaphor: ReFantazio requests a limit of 16 characters, but the game’s text field only has space for 10 characters. On real hardware, the IME UI popup has no visible text field (only the game’s own), so any extra characters are simply ignored. However, in the current emulator implementation (which synchronizes the UI text with the game’s buffer), extra characters become desynchronized, since there is no way to retrieve the text directly from the game’s internal buffer.

So, if no regressions are found during the final checks, I believe this can be merged. Multiple issues have been fixed, and I think everything should work correctly as long as the text limits are not exceeded. This is the best solution I can offer for now.

I checked, in God Eater 2 Ime still working.

PR #3735, opened 2025-10-13
GE2: Fix IME text conversion length handling

w1naenator:
- Reserve an extra space for the terminating character, resolving an issue in GE2 where the last character did not appear when input reached the maximum length.


w1naenator:
image
image

PR #3736, opened 2025-10-14
Min max deadzone

tlarok:
Noticed that deadzone settings only has minimum distance toggle in ui. This pr adds maximum distance option to ui

Untitled
.


georgemoralis:
this pr should go on https://github.com/shadps4-emu/shadps4-qtlauncher

PR #3737, opened 2025-10-16
usbd: Implement usb backend system

ElBread3:
As a preliminary step to toys-to-life emulation, the goal of this PR is to implement a polymorphic USB backend class that shims all necessary functions for toys-to-life games to use emulated devices. This doesn't actually emulate any of the devices as of yet, I think it would be more fitting to leave that to @deReeperJosh, who previously contributed this feature to the other emulators.


src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,443 @@
...
+#if defined(_WIN32)
+        s_has_removed_driver = true;
+        return LIBUSB_SUCCESS;
+#else
+        return libusb_detach_kernel_driver(dev_handle, interface_number);
+#endif
+    }
+    s32 AttachKernelDriver(libusb_device_handle* dev_handle, s32 interface_number) override {
+#if defined(_WIN32)
+        s_has_removed_driver = true;
```
Niram7777: I think it should be false here

src/core/libraries/usbd/usbd.cpp
```diff
@@ -479,6 +457,12 @@ int PS4_SYSV_ABI Func_D56B43060720B1E0() {
 }
 
 void RegisterLib(Core::Loader::SymbolsResolver* sym) {
+    if (Config::getUsbDeviceBackend() == Config::UsbBackendType::SkylandersPortal) {
+        // usb_backend = std::make_shared<SkylandersPortalBackend>();
```
Niram7777:
```cpp
//TODO implement fake devices
using SkylandersPortalBackend = UsbEmulatedBackend
using InfinityBaseBackend = UsbEmulatedBackend
using DimensionsToypadBackend = UsbEmulatedBackend

switch Config::getUsbDeviceBackend() {
case SkylandersPortal
    usb_backend = make_shared SkylandersPortalBackend
    break
case InfinityBase
    usb_backend = make_shared InfinityBaseBackend
    break
case DimensionsToypad
    usb_backend = make_shared DimensionsToypadBackend
    break
default
    usb_backend = make_shared UsbRealBackend
    break
}
```

src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,443 @@
...
+    s64 GetDeviceList(libusb_device*** list) override {
+        auto** fake = new libusb_device*[2];
+        fake[0] = GetDevice(nullptr);
+        fake[1] = nullptr;
+        *list = fake;
+
+        return 1;
+    }
+    void FreeDeviceList(libusb_device** list, s32 unref_devices) override {
+        libusb_free_device_list(list, 0); // to avoid unref fake device
```
Niram7777: we should either use malloc instead of new or implement our own libusb_free fonction here (delete,delete[]).
to not mix new/free.

https://github.com/libusb/libusb/blob/master/libusb/core.c#L906
https://github.com/libusb/libusb/blob/master/libusb/core.c#L1323

src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,443 @@
...
+    s32 ClaimInterface(libusb_device_handle* dev_handle, s32 interface_number) override {
+        return LIBUSB_SUCCESS;
+    }
+    libusb_device_handle* OpenDeviceWithVidPid(u16 vendor_id, u16 product_id) override {
+        libusb_device_handle* dev_handle = nullptr;
+        OpenDevice(GetDevice(nullptr), &dev_handle);
+        return dev_handle;
+    }
+    s32 ResetDevice(libusb_device_handle* dev_handle) override {
+        UNREACHABLE_MSG("Guest decided to reset device, might be an implementation issue");
```
Niram7777: I think it can be a warning.
we return 0 (success) or LIBUSB_ERROR_NOT_FOUND if wanted to simulate disconnection (in this case guest might call close after by reading libusb L.1959)

https://github.com/libusb/libusb/blob/master/libusb/core.c#L1971

src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,443 @@
...
+        auto* _dev_handle = static_cast<UsbDeviceHandle*>(calloc(1, sizeof(libusb_device_handle*)));
+        if (!_dev_handle) {
+            return LIBUSB_ERROR_NO_MEM;
+        }
+        _dev_handle->dev = dev;
+        *dev_handle = reinterpret_cast<libusb_device_handle*>(_dev_handle);
+        return LIBUSB_SUCCESS;
+    }
+    void CloseDevice(libusb_device_handle* dev_handle) override {
+        UNREACHABLE_MSG("Guest decided to close device, might be an implementation issue");
```
Niram7777: I think it can be a warning.
And we just have to release the memory.
If more is needed, the implementation is here
https://github.com/libusb/libusb/blob/master/libusb/core.c#L1601

Niram7777 (changes requested):

src/core/libraries/usbd/usb_backend.h
ElBread3: The unreachables for ResetDevice and CloseDevice here were mainly for debugging purposes while implementing, so yeah I can definitely downgrade these

src/core/libraries/usbd/usb_backend.h
ElBread3: This was originally a delete[] but CLion randomly hung here once while testing so I changed cause I'm a lazy ass, gonna look at that again when I'm free and change that back for sure

Niram7777 (approved):

src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,445 @@
...
+    }
+    void CloseDevice(libusb_device_handle* dev_handle) override {
+        LOG_WARNING(Lib_Usbd, "Guest decided to close device, might be an implementation issue");
+        free(dev_handle);
+    }
+    libusb_device* GetDevice(libusb_device_handle* dev_handle) override {
+        const auto desc = FillDeviceDescriptor();
+        ASSERT(desc);
+
+        const auto fake = new UsbDevice();
```
Niram7777: calloc

src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    s32 GetConfiguration(libusb_device_handle* dev, s32* config) override {
+        return LIBUSB_SUCCESS;
+    }
+    s32 GetActiveConfigDescriptor(libusb_device* dev, libusb_config_descriptor** config) override {
+        const auto endpoint_descs = FillEndpointDescriptorPair();
+        const auto interface_desc = FillInterfaceDescriptor(endpoint_descs);
+
+        const auto interface = new libusb_interface();
```
Niram7777: calloc

src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,445 @@
...
+    s64 GetDeviceList(libusb_device*** list) override {
+        auto** fake = static_cast<libusb_device**>(calloc(2, sizeof(libusb_device*)));
+        fake[0] = GetDevice(nullptr);
+        fake[1] = nullptr;
+        *list = fake;
+
+        return 1;
+    }
+    void FreeDeviceList(libusb_device** list, s32 unref_devices) override {
+        libusb_free_device_list(list, 0); // to avoid unref fake device
```
Niram7777: I believe we should pass unref_devices to not leak the devices in the list (otherwise it only free the list holder)

Niram7777 (commented):
I dont decide but I think we could merge your PR if you are struggling with C++, and we clean up later

src/core/libraries/usbd/usb_backend.h
ElBread3: The FillXDescriptor virtual functions were meant to serve the function of RPCS3's USBDescriptorNode helper class found [here](https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Io/Skylander.cpp#L197) and [here](https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Io/usb_device.h#L113), do you think these would fit better being more like how USBDescriptorNode was implemented?

src/core/libraries/usbd/usb_backend.h
ElBread3: Fake devices don't count references as it would be pointless to do so, it would likely internal assert attempting that. But yes I should probably manually free these then

src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,445 @@
...
+        const auto desc = FillDeviceDescriptor();
+        ASSERT(desc);
+
+        const auto fake = new UsbDevice();
+        fake->bus_number = 0;
+        fake->port_number = 0;
+        fake->device_address = 0;
+        fake->device_descriptor = *desc;
+        fake->ctx = g_libusb_context;
+        return reinterpret_cast<libusb_device*>(fake);
```
Niram7777, line 375: this cant be casted, libusb_device and UsbDevice are not the same and it will segfault
https://github.com/libusb/libusb/blob/master/libusb/libusbi.h#L510

src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,445 @@
...
+            HandleAsyncTransfer(transfer);
+
+            const u8 flags = transfer->flags;
+            transfer->status = LIBUSB_TRANSFER_COMPLETED;
+            transfer->actual_length = transfer->length;
+            if (transfer->callback) {
+                transfer->callback(transfer);
+            }
+            if (flags & LIBUSB_TRANSFER_FREE_TRANSFER) {
+                libusb_free_transfer(transfer);
```
Niram7777, line 453: since its emulated, i dont believe you can call this function or any other libusb function

src/core/libraries/usbd/usb_backend.h
```diff
@@ -0,0 +1,445 @@
...
+        return desc->bMaxPacketSize0;
+    }
+
+    s32 OpenDevice(libusb_device* dev, libusb_device_handle** dev_handle) override {
+        auto* _dev_handle = static_cast<UsbDeviceHandle*>(calloc(1, sizeof(libusb_device_handle*)));
+        if (!_dev_handle) {
+            return LIBUSB_ERROR_NO_MEM;
+        }
+        _dev_handle->dev = dev;
+        *dev_handle = reinterpret_cast<libusb_device_handle*>(_dev_handle);
```
Niram7777, line 358: this cant be casted, libusb_device_handle and UsbDeviceHandle are not the same and it will segfault

src/core/libraries/usbd/usb_backend.h
Niram7777: yes, it's not clear to me if rpcs3 also has his guest with libusb interface and they convert to their custom interfaces.

but this is what would work best IMO:
have our  UsbDevice/UsbDeviceHandle/USBDescriptorNode/FillXDescriptor internally, and copy from/to libusb structures (but no cast)

here we only have one device but we should check the handle or vid/pid are the same in input libusb and internally in our custom interfaces.

let me know if not clear

src/core/libraries/usbd/usb_backend.h
Niram7777: if it makes things simpler, we could use static/global devices for now

src/core/libraries/usbd/usb_backend.h
ElBread3: Both of these were copied just inlining usbi_atomic_t, it didn't segfault during testing if you've seen on Discord. But I had to inline these structs in the first place because in order to include libusb_device_handle and libusb_device with '#include libusbi.h' (for setting any of their fields, as you can't do this with incomplete types without importing)  requires a config.h to be configured to it with CMake in order to build and I didn't want to have to change libusb to make that work.

src/core/libraries/usbd/usb_backend.h
ElBread3: After a search in libusb code, it seems libusb_free_transfer is actually safe to call in this case, its libusb_unref_device call is only on itransfer->dev if it exists; itransfer->dev is initialized in libusb_submit_transfer which we override, so it will never do that. The only other libusb call in the backend then is the free_device_list which I'm already taking care of.

src/core/libraries/usbd/usb_backend.h
Niram7777: you should be able to include config.h, what issue exactly are you getting?

(we can't copy/paste the interface code and cast, if one day the libusb interface changes even if unlikely, it will  breaks)

deReeperJosh:
Very early WIP for the emulated USB device backend, working with Lego Dimensions now: https://youtu.be/jgxdiLVzHS4

One thing I did notice is that the HandleEvents wasn't working properly for Lego Dimensions. The game calls handle events before transfers are even submitted, and the thread that calls HandleEvents should only respond when there is data available. If a transfer is submitted on the Out Endpoint (0x01 for Dimensions), then this should be handled immediately in a separate thread, rather than adding it to the flight list, so that we respond on the handle events thread:

game usb_polling thread calls sceUsbdHandleEvents: if something in the flight list then respond

game calls submit transfer at Out Endpoint, handle transfer immediately in a separate thread

game calls submit transfer at In Endpoint (0x81) transfer added to flight list, responds once transfer submitted to Out Endpoint (0x01)

If we respond blank to a transfer submitted to the In Endpoint because we haven't received a transfer at the Out endpoint yet, game calls reset device

ElBread3:
When does Lego Dimensions send transfers for either endpoint, like does it send transfers for both endpoints simultaneously, or like an in on startup and outs later? Anyhow will need to look into how libusb handles endpoints specifically.

deReeperJosh:
From what I saw, it would init usbd then call handle sceUsbdHandleEvents, check the device list and open the portal, then allocate ~ 8 transfers for the 0x81 endpoint and submit them, then eventualy allocate a transfer for the 0x01 endpoint and submit it. Once the 0x01 transfer was submitted and returned, one of the transfers on the 0x81 endpoint is completed, then sceUsbdHandleEvents returns 0, before it would call sceUsbdHandleEvents again, allocate another 0x81 transfer, submit it, allocate 0x01 transfer, submit it, and continue like that

ElBread3:
Some more clarifying questions: does the out-endpoint transfer have a sooner timeout than the in-transfers (or otherwise finds itself first on the list)? Also do you know why it submits 8 in-transfers, and what way does hardware receive them (do all 8 transfers get data or only the first or some?) And are you saying hardware only submits 8 in-transfers on the first go and then only 1 at a time after that?

deReeperJosh:
They all have a timeout of 0 (so no timeout at all), and what happens is that all 8 transfers to the In endpoint are submitted to the hardware, it seems like the hardware just holds on to them and doesn't send a response until either a transfer is submitted to the out endpoint, or a figure gets added/removed from the hardware. So game sends 8 transfers to the hardware, hardware holds on to them until it has a response queued up, either from a figure being added/removed or it has something to respond with from the other endpoint - the 0x01 endpoint doesn't appear to respond with anything in the HandleEvents method, just calls the transfer callback straight away.

As to why it submits 8 transfers to the in endpoint, I imagine it just queues them up in that way so that it can be nice and quick to respond exactly when the events happen

deReeperJosh:
Skylanders doesn't use interrupt endpoints to send commands which is also confusing, it uses control transfers to send requests (endpoint 0x02) and when it submits an interrupt transfer it doesn't need to queue it up to respond, the portal always has a response to make (it either sends a response to a control transfer request or just the current status of the portal)

ElBread3:
So unless we want an accurate sim with a event handler thread and usb impl event condition driven processing, there's either call HandleEvents immediately on submitting an out-transfer, or your RPCS3 solution of returning a status of LIBUSB_TRANSFER_STALL in this case, which seems better?

deReeperJosh:
Yeah, rather than returning success in HandleEvents we should check to see if the transfer actually completed and return stalled or success

ElBread3:
Lastly would you rather work with separate LOAD_SKYLANDER/LOAD_DIMENSIONS and REMOVE_SKYLANDER/REMOVE_DIMENSIONS IPC commands or the abstracted LOAD_FIGURE and REMOVE_FIGURE?

deReeperJosh:
Ideally we would be able to use the abstracted version, but Dimensions is the game that needs slightly more. Skylanders and Disney only need a file path name and an index, but Dimensions also needs to have the ability to move figures around the Toypad easily. The way I did this on Cemu/RPCS3 was 3 extra methods, a Temporary Move when "picking" a figure up, a Cancel Remove when you place it back in the same space, and then a Move method which would need to take in the old index and new index. But we can try to keep it abstract and all 3 backends should be able to use Load with a string and u8 index, Remove just being a u8 index

PR #3740, opened 2025-10-17
Add informative update message for Qt build deprecation

DanielSvoboda:
Starting from version 0.12.0, the Qt version of the emulator will no longer receive direct updates.

However, the Qt interface remains available through the new official launcher:
[shadps4-qtlauncher](https://github.com/shadps4-emu/shadps4-qtlauncher/releases/)
 – based on the original shadPS4 source code.

We recommend migrating to this launcher to continue receiving updates.

Previously, when no update was available, users would see the generic message:
"No download URL found for the specified asset."

Now, they will instead see the custom message from the screenshot, which makes the situation clearer and includes a clickable button to open the appropriate GitHub release page.

Note: This message only appears when the emulator fails to find a Qt version release on GitHub.

image

*This message was also translated into all languages.

### DIFF
--- a/src/qt_gui/check_update.cpp
+++ b/src/qt_gui/check_update.cpp
@@ -153,8 +153,12 @@ tr("The Auto Updater allows up to 60 update checks per hour.\\nYou have reached 
         }
 
         if (!found) {
-            QMessageBox::warning(this, tr("Error"),
-                                 tr("No download URL found for the specified asset."));
+            QMessageBox::warning(
+                this, tr("Auto Updater"),
+                // clang-format off
+tr("<b>Notice:</b><br><br> Starting from version <b>0.12.0</b>, the Qt version of the emulator will no longer receive direct updates.<br><br>However, the Qt interface remains available through the new official launcher:<br><br><a href='https://github.com/shadps4-emu/shadps4-qtlauncher/releases/'>Qt Launcher</a> - based on the original shadPS4 source code.<br><br>We recommend switching to this launcher to continue receiving updates."));
+            // clang-format on
+
             reply->deleteLater();
             return;
         }
--- a/src/qt_gui/translations/ar_SA.ts
+++ b/src/qt_gui/translations/ar_SA.ts
@@ -276,8 +276,8 @@
       <translation>بيانات الإصدار غير صالحة.</translation>
     </message>
     <message>
-      <source>No download URL found for the specified asset.</source>
-      <translation>لم يتم العثور على عنوان URL للتنزيل للأصل المحدد.</translation>
+      <source>&lt;b&gt;Notice:&lt;/b&gt;&lt;br&gt;&lt;br&gt; Starting from version &lt;b&gt;0.12.0&lt;/b&gt;, the Qt version of the emulator will no longer receive direct updates.&lt;br&gt;&lt;br&gt;However, the Qt interface remains available through the new official launcher:&lt;br&gt;&lt;br&gt;&lt;a href='https://github.com/shadps4-emu/shadps4-qtlauncher/releases/'&gt;Qt Launcher&lt;/a&gt; - based on the original shadPS4 source code.&lt;br&gt;&lt;br&gt;We recommend switching to this launcher to continue receiving updates.</source>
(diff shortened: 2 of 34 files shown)


src/qt_gui/check_update.cpp
```diff
@@ -153,8 +153,12 @@ tr("The Auto Updater allows up to 60 update checks per hour.\\nYou have reached
...
-            QMessageBox::warning(this, tr("Error"),
-                                 tr("No download URL found for the specified asset."));
+            QMessageBox::warning(
+                this, tr("Auto Updater"),
+                // clang-format off
+tr("<b>Notice:</b><br><br> Starting from version <b>0.12.0</b>, the Qt version of the emulator will no longer receive direct updates.<br><br>However, the Qt interface remains available through the new official launcher:<br><br><a href='https://github.com/shadps4-emu/shadps4-qtlauncher/releases/'>Qt Launcher</a> - based on the original shadPS4 source code.<br><br>We recommend switching to this launcher to continue receiving updates."));
+            // clang-format on
+
             reply->deleteLater();
             return;
```
StevenMiller123, line 163: I wonder if it would be better to have this as a warning that always appears?
As-is, this will only appear for people with bad internet or once we end the Qt precompiled builds (if I understand this correctly).

Also, not-so-important nitpick here, but don't disable clang-format for something so minor like this. It's fine for areas where clang destroys readability or where the formatting suggestions break code, but for something simple like this, you're just bypassing a system we have in place to ensure our code is more readable.

GHU7924:
1) Correct me if I'm wrong, but since these are the latest Qt builds, the update channels for them should be removed.
That is, configure the emulator to only look for the new Release 0.12.0 (Qt) build when it starts, and if it doesn't find one, it displays your message.

Furthermore, the update-related interface should be hidden (or removed) from the interface in the Fronted tab and from the Help menu.

2) Instead of update channels, you can add a link to Qt Launcher.

3) If someone wants to continue using this Qt build, they should be able to disable this message. For example, by checking the box next to "I have read this notice," which will then add another checkbox: "Do not show this message again" (which disables checking for updates) and an OK button (which closes the message).
It would also be better to remove the close button (the cross in the upper-right corner) from this window.

Next

>However, the Qt interface remains available through the new official launcher:
shadps4-qtlauncher
– based on the original shadPS4 source code.

Replace, for example: If you plan to continue receiving new emulator updates and still use the graphical interface, we strongly recommend replacing your Qt build with our new official launcher: Qt Launcher.

There's no point in describing Qt Launcher here. It's better to add information at https://github.com/shadps4-emu/shadps4-qtlauncher/releases/ that, starting with version 0.12.0, we've abandoned the Qt emulator builds and replaced them with Qt Launcher. The reason for the change follows. Then you can briefly explain what Qt Launcher is.

>We recommend migrating to this launcher to continue receiving updates.

This part can be removed.

It should look something like this:

502657393-cd878a53-cb85-400c-abf8-7c7ba2aca2b0

@tochnonechelovek I would be grateful if you could check the Russian localization.

StevenMiller123 (changes requested):

src/qt_gui/check_update.cpp
DanielSvoboda: This message will only appear when the user tries to update in any of the channels and does not have the qt file, for example "win64-qt" = "shadps4-win64-qt-2025-10-14-db9921b.zip". If there is a problem with the internet, it is another message.

If I don't put the clang code, that message will be separated into several lines, and that way it's not possible to locate the translations.

DanielSvoboda:
@GHU7924 
1. All QT-related information will be removed soon. Therefore, Release 0.12.0 will not have QT. This message is for when this happens. The user will try to update and will only receive the message "No download URL found for the specified asset."

2. In the message, the blue text is a link.

3. If the user still wants to use the last pre-release version before 0.12 without receiving the message, simply disable the automatic update option, as this message will only appear when they try to download and there is no QT option available.

GHU7924:
@DanielSvoboda 

>All QT-related information will be removed soon. Therefore, Release 0.12.0 will not have QT. This message is for when this happens. The user will try to update and will only receive the message "No download URL found for the specified asset."

I forgot that we have PR `Remove Qt from emulator` 😅. 

>In the message, the blue text is a link.

I was hasty, you are right, when updating the user will see a message and a link, I apologize.

>If the user still wants to use the last pre-release version before 0.12 without receiving the message, simply disable the automatic update option, as this message will only appear when they try to download and there is no QT option available.

Yes, but perhaps it would be more convenient if the update could be disabled immediately after reading the notification.
As far as I understand, this will be relevant for Qt build 0.11.0 (0.11.1), then you can add text that `if you want to continue using this build, then disable the update check so as not to see this message`.

tochnonechelovek:
> @tochnonechelovek I would be grateful if you could check the Russian localization.

@GHU7924 Yeah, it's great.

PR #3764, opened 2025-11-01
buffer_cache: smaller regions

Kyoskii:
this was a change back between v0.9.0 to v0.10.0
https://github.com/raphaelthegreat/shadPS4/commit/9f37ede336f589abfd4ac34f8194ce2e8db30d4e

reverting the TRACKER_HIGHER_PAGE_BITS from 24 to 22 gives a notable increase to performance.


kalaposfos13 (changes requested):
Please update the copyright year at the top of the file.

kalaposfos13:
I found no regressions on my end.

raphaelthegreat (approved):

kalaposfos13 (approved):

raphaelthegreat:
Ideally region manager should be reworked to have as small lock granularity as possible (a lock per 64 pages is easy to achieve but it is possible to get a lock per 8 pages for each byte) but this is good in the meantime. The smaller the granularity the less chance of threads fighting over a region, which is the cause of slowdown in this case

PR #3773, opened 2025-11-05
Net: Better return stub for sceNetPoolCreate

StevenMiller123:
sceNetPoolCreate never returns 0 on success, it returns a positive value as an ID for the net pool. Some games check to make sure the return is positive, these games wouldn't work properly before.

This PR sets the return value of sceNetPoolCreate to a positive, increasing value.
This gets THE PLAYROOM (CUSA00001) further, though now it dies on 
```
[Debug] <Critical> fiber.cpp:42 _sceFiberCheckStackOverflow: Unreachable code!
Stack overflow detected in fiber with size = 0x400
```


georgemoralis:
yah seems valid

NinjaGaiden529:
thanks george

PR #3774, opened 2025-11-05
Libkernel: Implement/stub some functions

StevenMiller123:
This PR does a few things:

- Create a define for the latest firmware version number, this can be used in place of game firmware for sceKernelGetSystemSwVersion, and I used it for sceKernelGetAllowedSdkVersionOnSystem.
- Fix get_authinfo with null authinfo struct. When I ran this edge case on my console, I got a -1 return with errno set to EPERM.
- Stub sceKernelGetAppInfo, currently only populating the cusa name and has_param_sfo fields.
- Implement sceKernelHasNeoMode, this returns true if you're running on a PS4 Pro (and doesn't require the param.sfo flag)
- Implement sceKernelGetMainSocId based on hardware observations. This returns a chunk of data from a cpuid instruction, and is used by some libraries to determine if they're running on PS4 Pro hardware.
- Stub sceKernelGetCurrentCpu, this is spammed by a few games.

### DIFF
--- a/src/core/libraries/kernel/kernel.cpp
+++ b/src/core/libraries/kernel/kernel.cpp
@@ -249,7 +249,6 @@ s32 PS4_SYSV_ABI sceKernelGetSystemSwVersion(SwVersionStruct* ret) {
     if (ret == nullptr) {
         return ORBIS_OK;
     }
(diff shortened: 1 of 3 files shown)


squidbus (approved):

squidbus:
Looks good if clang-format is fixed

PR #3779, opened 2025-11-06
usbd: Emulate Dimensions Toypad

deReeperJosh:
Last but not least, Emulated Dimensions Toypad backend. Again ported code that I wrote for RPCS3/Cemu.

One thing to note here is that the game called reset device constantly if transfers submitted to at the 0x81 endpoint were responded too with no data (even if the transfer was returned with a status of stalled) so I overwrite the submitTransfer method in this class, and ensure that 0x01 transfers are handled immediately, while 0x81 transfers are held on to until data is ready to be responded with.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -539,6 +539,8 @@ set(RANDOM_LIB src/core/libraries/random/random.cpp
 set(USBD_LIB src/core/libraries/usbd/usbd.cpp
              src/core/libraries/usbd/usbd.h
              src/core/libraries/usbd/usb_backend.h
+             src/core/libraries/usbd/emulated/dimensions.cpp
+             src/core/libraries/usbd/emulated/dimensions.h
              src/core/libraries/usbd/emulated/infinity.cpp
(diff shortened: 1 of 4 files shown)


georgemoralis:
omg that was fast ;D , do we have other devices to emulate next?

deReeperJosh:
I don't think so! Unless there are some other games I am not thinking of then this is the last one

ElBread3:
It seems you probably didn't notice, but the idea was in that condition you'd return LIBUSB_TRANSFER_STALL in HandleAsyncTransfer like the RPCS3 comment suggests so that you wouldn't have to override SubmitTransfer

deReeperJosh:
I did try to return the LIBUSB_TRANSFER_STALL in that scenario, but the game would call reset device in any case if the 0x81 endpoint returned before there was any actual data to respond with, so I had to overwrite the method. I will see if I can spend more time investigating but I believe PS3 only uses 1 thread for USB polling where it appears that PS4 has 2? In any case, have updated the comment in the interim

PR #3790, opened 2025-11-11
usbd: calloc UsbDevice instead of pointer

deReeperJosh:
Issue was raised in the Discord server, where when using an emulated backend caused the Skylander games to crash on launch. Bisected this issue to the FreeDeviceList method (specifically line 296) where calling free on the emulated device pointer was causing the crash. I saw that when allocating the emulated device in GetDevice we were using calloc to initialize a pointer with a sizeof UsbDevice pointer rather than size of UsbDevice. Some testing would be appreciated here, as I still get some crashes on Windows with this change (however instead of 100% of the time it is only 33% of the time)

### DIFF
--- a/src/core/libraries/usbd/usb_backend.h
+++ b/src/core/libraries/usbd/usb_backend.h
@@ -366,7 +366,7 @@ public:
         const auto desc = FillDeviceDescriptor();
         ASSERT(desc);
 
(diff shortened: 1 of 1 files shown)


georgemoralis:
tested it seems ok

PR #3793, opened 2025-11-12
usbd: calloc libusb_interface instead of pointer

StevenMiller123:
The broken calloc will leave `interface->num_altsetting` out of bounds, which will cause crashing on Windows.


georgemoralis:
ahhh pr

PR #3798, opened 2025-11-14
buffer_cache: Begin memory tracker rework

raphaelthegreat:
This is split from buffer-cleanup branch to test it separately

Current memory tracker was a rewrite from https://github.com/shadps4-emu/shadPS4/pull/3070. The main advantage of that rework was batching memory protection actions for the whole region instead of just the modified areas. However the introduction of the bitset also restricted locking flexibility; a single lock for the entire region was required.

This rework aims to solve the second issue, that of lock granularity, while retaining the benefits. It is a mix of older and newer code, by taking the faster and more readable approaches from either case. Locking is now per-page; each word also has a lock bitmask that is set/cleared to acquire/release the locks. For uncontested case, acquiring a lock is a bitwise or and a cmpxchg instruction. Locks for up to 64 pages can also be acquired immediately for said case, so iterating larger memory regions remains fast. When the lock is contested, waitable atomics are used which (at least in libstdc++) will perform a spinloop and then sleep, behaving in similar fashion to an adaptive mutex.

Readbacks are still unimplemented because the new locking makes the old invalidate method annoying to reimplement. WIll be added again before this is done


src/video_core/buffer_cache/memory_tracker.h
```diff
@@ -5,158 +5,117 @@
...
-                }
-            });
+        IteratePages(cpu_addr, size, [&on_flush](RegionManager* manager, u64 offset, u64 size) {
+            const bool should_flush = [&] {
+                // TODO
+                /*std::scoped_lock lk{manager->lock};
+                if (Config::readbacks() &&
+                    manager->template IsRegionModified<Type::GPU>(offset, size)) {
+                    return true;
+                }*/
```
squidbus: Is this meant to be resolved before this PR goes in?

squidbus (commented):
Looks okay aside from comment. Is this still pending testing?

georgemoralis:
This got updated ppl go test!

StevenMiller123:
@raphaelthegreat Could you rebase this so Windows users can test?

StevenMiller123:
Battlefield V still crashes on
```
[Debug] <critical> (shadPS4:GpuComm) page_manager.cpp:99 operator(): Assertion Failed!
Not enough watchers
```

Same issue as main, but given this PR makes changes to that section of code, I feel it's worth mentioning.
[CUSA08724.log](https://github.com/user-attachments/files/26872833/CUSA08724.log)

I'm not seeing any noticeable performance uplift in the titles I've tested, though I haven't tested too many games yet, just Marvel's Spider-Man, SHADOW OF THE COLOSSUS, KINGDOM HEARTS III, and Call of Duty Advanced Warfare.

georgemoralis:
rebased!

georgemoralis:
reabased

georgemoralis:
rebased

georgemoralis:
rebased

Issue #3794, opened 2025-11-13
[GAME BUG]: Video Debug Info menu makes games crash if FSR is on


Missake212:
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.
- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [x] I have disabled all patches and cheats and the issue is still present.
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadPS4/wiki/I.-Quick-start-%5BUsers%5D#4-dumping-firmware-modules) installed.

### Describe the Bug

When trying to open the Video Debug Info menu in a game with FSR enabled in the settings (and with the condition of the output res being higher than the internal res of the game), games will instantly crash with the following assert:

``[Debug] <Critical> image.cpp:87 operator(): Assertion Failed!``

Turning FSR off solves the issue. This was tested on a fresh OS and shadPS4 install so I don't think that's an issue on my end, I haven't tested on Windows but I could try if asked to.

### Reproduction Steps

Turn FSR on in the settings, make sure the game internal resolution is lower than your monitor's output resolution in order for FSR to work, launch any game then press CTRL+F10, the game should crash instantly.

### Specify OS Version

Bazzite

### CPU

Intel Core I9-13900K

### GPU

Nvidia RTX 4090

### Amount of RAM in GB

32 GB

### Amount of VRAM in GB

24 GB

### Log File

[shad_log.txt](https://github.com/user-attachments/files/23533174/shad_log.txt)

andybellenie:
Also having this issue. Seems it was fixed in a fork but not ported to main branch.

RedAstaire1:
Same problem

Kravickas:
Annoying bug 😵‍💫, retest please :)

PR #3801, opened 2025-11-15
Make FSR off by default

Missake212:
I personally think it always made more sense to keep FSR disabled by default, and I think it makes even more sense since there's a [regression](https://github.com/shadps4-emu/shadPS4/issues/3794) with FSR, this obviously won't affect anyone who already has shadPS4 installed but for the future I believe it makes more sense to have this off, it can reduce performance a bit and I don't think everyone wants FSR to modify the looks of games for them by default.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -178,7 +178,7 @@ static ConfigEntry<string> fullscreenMode("Windowed");
 static ConfigEntry<string> presentMode("Mailbox");
 static ConfigEntry<bool> isHDRAllowed(false);
 static ConfigEntry<bool> fsrEnabled(false);
-static ConfigEntry<bool> rcasEnabled(false);
+static ConfigEntry<bool> rcasEnabled(true);
 static ConfigEntry<int> rcasAttenuation(250);
 
 // Vulkan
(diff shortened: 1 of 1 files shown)


src/common/config.cpp
```diff
@@ -177,8 +177,8 @@ static ConfigEntry<bool> isFullscreen(false);
 static ConfigEntry<string> fullscreenMode("Windowed");
 static ConfigEntry<string> presentMode("Mailbox");
 static ConfigEntry<bool> isHDRAllowed(false);
-static ConfigEntry<bool> fsrEnabled(true);
-static ConfigEntry<bool> rcasEnabled(true);
+static ConfigEntry<bool> fsrEnabled(false);
+static ConfigEntry<bool> rcasEnabled(false);
```
squidbus: RCAS has no effect if FSR is off so I don't think changing this is necessary based on the justification here. Should still be default on if FSR is enabled.

squidbus (commented):
Not commenting on the premise here as I'll leave that choice to others, but had one comment.

src/common/config.cpp
Missake212: That's fair, I changed it back to true.

PR #3809, opened 2025-11-16
buffer_cache: Split DMA fault handling code from buffer cache

raphaelthegreat:
Pulls out the raw vulkan code from the buffer cache into a separate class that is easier to reason with. Also comes with some small improvements:

* Faults are now processed on a separate buffer, removing the dependency on the buffer cache download buffer which necessitates the hack in Scheduler::Wait.
* Bunch of redundant vkCmdFillBuffer operations have been removed
* Use FindBuffer instead of CreateBuffer when processing fault ranges. Avoids potentially recreating buffer if next frame also faults on that range before its created.


LNDF (approved):
I just thought that it might be a good idea to rework the infra of host shader compilation and compile the glsl shaders into spirv and ship those with the emulator insteed.

lgtm

squidbus (approved):

PR #3814, opened 2025-11-17
audio3d: sceAudio3dPortGetAttributesSupported() - init num_capabilities variable to zero

mailwl:
when `capabilities` is `nullptr`, sceAudio3dPortGetAttributesSupported() should return count of available `capabilities`
this PR initialized `num_capabilities` variable to zero, to return empty `capabilities`

### DIFF
--- a/src/core/libraries/audio3d/audio3d.cpp
+++ b/src/core/libraries/audio3d/audio3d.cpp
@@ -371,10 +371,13 @@ s32 PS4_SYSV_ABI sceAudio3dPortFreeState() {
 s32 PS4_SYSV_ABI sceAudio3dPortGetAttributesSupported(OrbisAudio3dPortId port_id,
                                                       OrbisAudio3dAttributeId* capabilities,
                                                       u32* num_capabilities) {
-    LOG_DEBUG(Lib_Audio3d, "(STUBBED) called caps = {}, num = {}", static_cast<void*>(capabilities),
+    LOG_ERROR(Lib_Audio3d, "(STUBBED) called caps = {}, num = {}", static_cast<void*>(capabilities),
               static_cast<void*>(num_capabilities));
     if (capabilities == nullptr) {
-        *num_capabilities = 0;
+        // we use ORBIS_AUDIO3D_ATTRIBUTE_PCM, so return only one capability
+        *num_capabilities = 1;
+    } else {
+        capabilities[0] = OrbisAudio3dAttributeId::ORBIS_AUDIO3D_ATTRIBUTE_PCM;
     }
     return ORBIS_OK;
 }
--- a/src/core/libraries/audio3d/audio3d.cpp
+++ b/src/core/libraries/audio3d/audio3d.cpp
@@ -371,8 +371,8 @@ s32 PS4_SYSV_ABI sceAudio3dPortFreeState() {
 s32 PS4_SYSV_ABI sceAudio3dPortGetAttributesSupported(OrbisAudio3dPortId port_id,
                                                       OrbisAudio3dAttributeId* capabilities,
                                                       u32* num_capabilities) {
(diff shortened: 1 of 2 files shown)


georgemoralis:
is that tested with specific game?

StevenMiller123:
I don't think we should reduce the log severity to debug when the function is still mostly stubbed out.

src/core/libraries/audio3d/audio3d.cpp
```diff
@@ -368,8 +368,14 @@ s32 PS4_SYSV_ABI sceAudio3dPortFreeState() {
...
-s32 PS4_SYSV_ABI sceAudio3dPortGetAttributesSupported() {
-    LOG_ERROR(Lib_Audio3d, "(STUBBED) called");
+s32 PS4_SYSV_ABI sceAudio3dPortGetAttributesSupported(OrbisAudio3dPortId port_id,
+                                                      OrbisAudio3dAttributeId* capabilities,
+                                                      u32* num_capabilities) {
+    LOG_DEBUG(Lib_Audio3d, "(STUBBED) called caps = {}, num = {}", static_cast<void*>(capabilities),
+              static_cast<void*>(num_capabilities));
+    if (capabilities == nullptr) {
+        *num_capabilities = 0;
+    }
```
georgemoralis, line 381: else { //move error log here }

mailwl:
> is that tested with specific game?

saw it in CyberPunk, but there is setted to 0 before calling. But if it not set, we potentially return garbage:
```c++
__int64 __fastcall sub_3F89F20(__int64 a1)
{
  unsigned int v1; // r14d
  char *v2; // rbx
  __int64 v3; // rcx
  __int64 v5; // [rsp+0h] [rbp-40h] BYREF
  unsigned int count; // [rsp+Ch] [rbp-34h] BYREF
  __int64 v7; // [rsp+10h] [rbp-30h]

  v1 = 0;
  v7 = *(_QWORD *)_stack_chk_guard;
  count = 0;
  sceAudio3dPortGetAttributesSupported(a1, 0, &count);
  v2 = (char *)&v5 - ((4LL * count + 15) & 0xFFFFFFFFFFFFFFF0LL);
  sceAudio3dPortGetAttributesSupported((unsigned int)a1, v2, &count);
  if ( count )
  {
    v3 = 0;
    while ( *(_DWORD *)&v2[4 * v3] != 3 )
    {
      if ( ++v3 >= (unsigned __int64)count )
        return 0;
    }
    LOBYTE(v1) = 1;
  }
  return v1;
}
```

src/core/libraries/audio3d/audio3d.cpp
mailwl: looks like returning zero make audio3d off, and no need to fill zero-size array

mailwl:
looking at source, one capability in use, so i return it to caller

src/core/libraries/audio3d/audio3d.cpp
```diff
@@ -368,8 +368,17 @@ s32 PS4_SYSV_ABI sceAudio3dPortFreeState() {
     return ORBIS_OK;
 }
 
-s32 PS4_SYSV_ABI sceAudio3dPortGetAttributesSupported() {
-    LOG_ERROR(Lib_Audio3d, "(STUBBED) called");
+s32 PS4_SYSV_ABI sceAudio3dPortGetAttributesSupported(OrbisAudio3dPortId port_id,
+                                                      OrbisAudio3dAttributeId* capabilities,
+                                                      u32* num_capabilities) {
+    LOG_ERROR(Lib_Audio3d, "(STUBBED) called caps = {}, num = {}", static_cast<void*>(capabilities),
```
squidbus: Instead of casting you can use `fmt::ptr`

georgemoralis:
support for this has been added in #3884

PR #3816, opened 2025-11-19
video_core: Initial implementation of pipeline cache

psucien:
Marked as draft, as there are few tess validation warnings left to fix, yet so far the implementation is reliable enough. For the next steps it would be nice to:
- minimize latency on new PL creation (there is a room for improvement in serialization sequence)
- async schedule for IO
- single container (miniz) + zlib compression for blobs


georgemoralis:
for the ones that didn't understand the topic it's shaders caching . So testers do you work and post you results here :D

georgemoralis:
you need to add in config.toml under [Vulkan]

pipelineCacheEnable=true

larfy298:
Obligatory "Works great on bloodborne!" Post (On an AMD GPU / CPU Windows 11)
Framerate seems roughly on par with nightly but frametimes feel like a dream, had a lot of hitching and stuttering issues before which have (for the most part) cleaned up, hopefully that's real and I'm not placebo-ing myself
Edit: Eventually crashed when warping
`[Debug] <Critical> ir_emitter.cpp:758 CompositeConstruct: Unreachable code!`

raphaelthegreat:
The file format of the cache blob should probably be explained a bit. Having a single file for all cached shader blobs (separate one for pipelines tho looks okay) should also be implemented, it appears this will create a file for each blob. Also shader metas and spv files should be in the same container ideally, for each program it stores a Shader::Info and a list of (Shader::StageSpecialization, spirv)

The specialization lookup was purposefully done with comparison and not hashing to avoid recreating a shader when resources are unbound (i.e if a T# or V# becomes null you can ignore that binding and not recreate, while with hashing you are going to recompile), so this change to a hash might cause more shaders to get compiled, especially in TLG.

Also consider storing and loading the vulkan pipeline cache blob before the warm-up step, that should speed up the pre-compilation step on AMD noticeably, alongside with a potential multithreading of such operation.

src/video_core/renderer_vulkan/vk_pipeline_serialization.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+    u32 offset{};
+    std::vector<u8> container{};
+};
+
+struct Writer {
+    template <typename T>
+    void Write(const T* ptr, size_t size) {
+        if ((ar.offset + size) >= ar.container.size()) {
+            ar.Grow(size);
```
kalaposfos13:
```suggestion
            ar.Grow(size + 1);
```
I'm not sure if this is the correct solution, but it does stop the assert in the Advance call 3 lines below from triggering on Debug builds.
Edit: Actually I think the easier thing to do  would be to change the assert in Advance instead.

src/video_core/renderer_vulkan/vk_pipeline_serialization.cpp
```diff
@@ -0,0 +1,565 @@
...
+    size_t SizeBytes() const {
+        return container.size();
+    }
+
+    u8* CurrPtr() {
+        return container.data() + offset;
+    }
+
+    void Advance(size_t size) {
+        assert((offset + size) < container.size());
```
kalaposfos13:
```suggestion
        ASSERT((offset + size) <= container.size());
```
You should use our ASSERT macro for this, and not just here, but everywhere.

Missake212:
First launch works as expected, cache folder is created and is populated, can't launch the game after that though. Deleting the cache folder makes the game launch again.

``[Debug] <Critical> vk_graphics_pipeline.cpp:386 operator(): Assertion Failed!``
``Failed to create graphics pipeline: ErrorUnknown``

[CUSA03173.log](https://github.com/user-attachments/files/23626545/CUSA03173.log)

Also I think for the time being at least it would be nice to log this setting.

par274:
Welcome to the mandatory Bloodborne test! 😄  
I tested this PR and got one crash on the very first launch, but after restarting it ran perfectly fine.  
The game — which was already working with the latest updates. Now runs buttery smooth.  

The **average FPS** is about the same, but the **1% lows have improved significantly**, making the overall experience much smoother and more stable.

bigol83:
> First launch works as expected, cache folder is created and is populated, can't launch the game after that though. Deleting the cache folder makes the game launch again.
> 
> `[Debug] <Critical> vk_graphics_pipeline.cpp:386 operator(): Assertion Failed!` `Failed to create graphics pipeline: ErrorUnknown`
> 
> [CUSA03173.log](https://github.com/user-attachments/files/23626545/CUSA03173.log)
> 
> Also I think for the time being at least it would be nice to log this setting.

Same with Octopath Traveler II 
[CUSA36274.log](https://github.com/user-attachments/files/23626833/CUSA36274.log)

Other games i tested don't have this issue

src/video_core/renderer_vulkan/vk_pipeline_serialization.cpp
```diff
@@ -0,0 +1,565 @@
...
+}
+
+bool LoadShaderMeta(Serialization::Archive& ar, Shader::Info& info,
+                    std::optional<Shader::Gcn::FetchShaderData>& fetch_shader_data,
+                    size_t& spec_hash, size_t& perm_idx) {
+    Serialization::Reader meta{ar};
+
+    u64 perm_hash_ar{};
+    meta.Read(perm_hash_ar);
+    assert(perm_hash == perm_hash_ar);
```
kalaposfos13: perm_hash doesn't exist here, the compiler just stays silent due to you never actually having tested any of the asserts on account of never having built a Debug version. When a Debug version is used, or the assert is changed to the project-specific ASSERT, this error is exposed.

src/video_core/renderer_vulkan/vk_pipeline_serialization.cpp
```diff
@@ -0,0 +1,565 @@
...
+#include "video_core/renderer_vulkan/vk_pipeline_cache.h"
+#include "video_core/renderer_vulkan/vk_pipeline_storage.h"
+#include "video_core/renderer_vulkan/vk_shader_util.h"
+
+namespace Serialization {
+
+/* You should increment versions below once corresponding serialization scheme is changed. */
+static constexpr u32 ShaderBinaryVersion = 0u;
+static constexpr u32 ShaderMetaVersion = 0u;
+static constexpr u32 PipelineKeyVersion = 0u;
```
kalaposfos13: Why not invalidate the cache based on the emulator's git hash? Sure, it'll be less convenient for nightly users, but it will also be less error-prone, and guarantees that ghost bugs caused by cached shaders from a previous version can't happen.

src/video_core/renderer_vulkan/vk_pipeline_serialization.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+    auto [it_pgm, new_program] = program_cache.try_emplace(program->info.pgm_hash);
+    if (new_program) {
+        program->InsertPermut(module, spec_hash, perm_idx);
+        it_pgm.value() = std::move(program);
+    } else {
+        // Make sure the permutation doesn't exist yet
+        const auto& it =
+            std::ranges::find(it_pgm.value()->modules, spec_hash, &Program::Module::spec_hash);
+        assert(it == it_pgm.value()->modules.end());
```
kalaposfos13: When not skipped, this assert will always get hit for every game in the presence of an already generated shader cache.

src/video_core/renderer_vulkan/vk_pipeline_serialization.cpp
georgemoralis: yah we can do it like this at least for releases (isRelease flag exists on cmakelists)

src/video_core/renderer_vulkan/vk_pipeline_serialization.cpp
```diff
@@ -0,0 +1,565 @@
...
+    Writer() = delete;
+    explicit Writer(Archive& ar_) : ar{ar_} {}
+
+    Archive& ar;
+};
+
+struct Reader {
+    template <typename T>
+    void Read(T* ptr, size_t size) {
+        assert((ar.offset + size) < ar.container.size());
```
kalaposfos13:
```suggestion
        ASSERT((ar.offset + size) <= ar.container.size());
```

src/video_core/renderer_vulkan/vk_pipeline_storage.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+    case BlobType::ShaderBinary: {
+        return "spv";
+        break;
+    }
+    case BlobType::PipelineKey: {
+        return "key";
+    }
+    default:
+        return "";
```
kalaposfos13:
```suggestion
        UNREACHABLE();
```

src/shader_recompiler/resource.h
```diff
@@ -54,11 +54,9 @@ struct BufferResource {
 
     constexpr AmdGpu::Buffer GetSharp(const auto& info) const noexcept {
         AmdGpu::Buffer buffer{};
-        if (inline_cbuf) {
+        if (inline_cbuf && inline_cbuf.base_address > 1) {
```
psucien: That needs to be reverted actually

kalaposfos13 (commented):
You should update the copyright years in path_util.cpp, fetch_shader.cpp, fetch_shader.h, translate.cpp, info.h, flatten_extended_userdata_pass.cpp, hull_shader_transform.cpp, ir_passes.cpp, resource_tracking_pass.cpp, srt.h, recompiler.cpp, recompiler.h, runtime_info.h, specialization.h, pixel_format.h, vk_compute_pipeline.cpp, vk_compute_pipeline.h, vk_graphics_pipeline.cpp, vk_graphics_pipeline.h, vk_pipeline_cache.cpp, and vk_pipeline_cache.h.

aakh1361:
It had no effect on Resident Evil 4 (OLD)
It still runs at 30-40 frames but the game is in slow motion

[CUSA04704.log](https://github.com/user-attachments/files/23629950/CUSA04704.log)

src/video_core/renderer_vulkan/vk_pipeline_storage.cpp
```diff
@@ -74,17 +144,18 @@ bool DataBase::Save(BlobType type, const std::string& name, std::vector<u8>&& da
...
 bool DataBase::Save(BlobType type, const std::string& name, std::vector<u32>&& data) {
     if (!opened) {
         return false;
     }
 
-    const auto path = cache_dir / name;
-    return WriteVector(type, path.string(), data);
+    auto path = cache_dir / name;
+    return WriteVector(type, std::move(path), std::move(data));
+    return true;
```
psucien: The price of coding in notepad :/

StevenMiller123:
This PR regresses The Last of Us™ Remastered. It now crashes silently on startup instead of booting.
image

Crashes here, with sharp_idx being (u32)-1
image

[CUSA00552 main.log](https://github.com/user-attachments/files/23631795/CUSA00552.main.log)
[CUSA00552 pr.log](https://github.com/user-attachments/files/23631796/CUSA00552.pr.log)

StevenMiller123:
Disregard my comment, I forgot to enable the config option 😓 
Guess AMD drivers were responsible for what I was seeing...

src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
```diff
@@ -556,31 +572,40 @@ PipelineCache::Result PipelineCache::GetProgram(Stage stage, LogicalStage l_stag
...
     auto& program = it_pgm.value();
     auto& info = program->info;
+    info.pgm_base = params.Base(); // Needs to be actualized for inline cbuffer address fixup
+    info.user_data = params.user_data;
     info.RefreshFlatBuf();
     const auto spec = Shader::StageSpecialization(info, runtime_info, profile, binding);
+    const auto spec_hash = spec.Hash();
+
     size_t perm_idx = program->modules.size();
+    u64 perm_hash = HashCombine(params.hash, perm_idx);
```
baggins183, line 592: I'm confused why perm_idx is being mixed into a hash here. There's nothing about perm_idx that's inherent to the shader. Are you still able to encounter permutations in a different order between sessions? As in you encounter perm 0 then perm 1 on a fresh run, cache them, then encounter perm 1 then perm 0 on the next. 

I haven't read the whole PR so I'm probably missing things atm.

diogoaduartereis:
image
image
Helps alleviate the stutters of gow 3 so much.

diogoaduartereis:
The Order 1886 is not booting: 
[CUSA00785.log](https://github.com/user-attachments/files/23664456/CUSA00785.log)

src/video_core/cache_storage.h
```diff
@@ -1,20 +1,23 @@
...
 namespace Storage {
 
 enum class BlobType : u32 {
     ShaderMeta,
     ShaderBinary,
     PipelineKey,
+    ShaderProfile,
 };
 
 class DataBase {
```
raphaelthegreat, line 23: Does this class have to be a singleton? Seems to be mostly used from PipelineCache methods, so could it just be a member of that class and pass just a reference to any static functions in vk_pipeline_serialization that need it

src/video_core/cache_storage.cpp
```diff
@@ -0,0 +1,265 @@
...
+        mz_zip_writer_finalize_archive(&zip_ar);
+        mz_zip_writer_end(&zip_ar);
+    }
+
+    LOG_INFO(Render, "Cache dumped");
+}
+
+template <typename T>
+bool WriteVector(BlobType type, std::filesystem::path&& path_, std::vector<T>&& v) {
+    auto request = std::async(
```
raphaelthegreat: Could this be a std::move_only_function and have the io thread process this? std::async at least on libstdc++ has an overhead of create a std::shared_ptr for a state object and a new std::thread object when passed std::launch::async, this could be avoided with a emulator managed thread pool

Heliumrich:
I get massive flickering in BB (especially the house in the hunter's dream) seems to depend on the camera position or something.

By merging this PR with the "GC part 2" PR (though I am missing the very latest commit "[Remain stuff](https://github.com/shadps4-emu/shadPS4/pull/3816/commits/b24cc42c6a268b4663d9f20564ed02956f769594)")
But I don't get this if I test each PR individually 🤔

StevenMiller123:
The GC2 PR is fairly unstable, and likely causes these issues.

src/video_core/cache_storage.h
psucien: There are at least two cases where we want to access it outside of pl cache class: pre-exit sync and (possible) detilers preload. Also, the cache is designed to be generic and can store various types of blobs not necessarily related to pl cache functionality

src/video_core/cache_storage.cpp
psucien: That's a good point, we don't need threaded serialisation here, only fast data grab to release hot path stall. It seems `move_only_function` isn't widely supported so I changed it to `packaged_task`

src/video_core/cache_storage.h
raphaelthegreat: For exit handling, you can either add a function for rasterizer shutdown to execute before the std::quick_exit or an std::at_quick_exit handler. Caching detilers is a good idea but I don't think worth the extra complexity at the moment. About using cache for other type of blobs, that is perfectly fine but imo should be solved by better encapsulation (for example move the static variables in cpp file inside the class and construct separate database objects if usage in other places is desired). Also the class currently assumes in several places its doing pipeline caching so not being a singleton is gonna make it simpler overall

PR #3818, opened 2025-11-19
msys2: fix build

oltolm:
I fixed the MSYS2 build with clang64.

I have also fixed a bug? in `time.cpp`. [`std::chrono::sys_info`](https://en.cppreference.com/w/cpp/chrono/sys_info.html) `offset` was being multiplied by 60 even though it is already in seconds.

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -152,7 +152,7 @@ endif()
 # sirit
 add_subdirectory(sirit)
 if (WIN32)
-    target_compile_options(sirit PUBLIC "-Wno-error=unused-command-line-argument")
+    target_compile_options(sirit PRIVATE "-Wno-error=unused-command-line-argument")
 endif()
 
 # half
--- a/src/core/libraries/kernel/time.cpp
+++ b/src/core/libraries/kernel/time.cpp
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include <ctime>
 #include <thread>
 
 #include "common/assert.h"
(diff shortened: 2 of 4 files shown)


CMakePresets.json
kalaposfos13, line 1: I'm not sure if the changes in this file should be in the PR, did you push it by accident?

src/core/libraries/kernel/time.cpp
```diff
@@ -485,25 +486,41 @@ Common::NativeClock* GetClock() {
...
-#endif
+#endif // __APPLE__
     auto info = time_zone->get_info(std::chrono::system_clock::now());
 
     *local_time = info.offset.count() + info.save.count() * 60 + time;
 
     if (st != nullptr) {
         st->t = time;
-        st->west_sec = info.offset.count() * 60;
+        st->west_sec = info.offset.count();
```
kalaposfos13, line 516: Was this a general logic bug in the code that you fixed? If that's the case, you should've stated that in the PR description, as this isn't apparent from anywhere other than reading the diff.

externals/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 # SPDX-License-Identifier: GPL-2.0-or-later
 
+add_compile_options(-w)
```
kalaposfos13: This shouldn't be in the PR.

kalaposfos13 (changes requested):
Just a small nitpick, but you can also update the reuse strings in the files you changed to the current year.

CMakePresets.json
oltolm: Not really, but I will revert.

src/core/libraries/kernel/time.cpp
oltolm: Yes you are right. I think it is a bug because `info.offset` is already in [seconds](https://en.cppreference.com/w/cpp/chrono/sys_info.html).

externals/CMakeLists.txt
oltolm: I will revert.

kalaposfos13 (approved):

PR #3819, opened 2025-11-19
Avoid storing the Tcb pointer on the stack

kalaposfos13:
This PR fixes some of the random crashes in SotC on Linux.

### DIFF
--- a/src/core/tls.cpp
+++ b/src/core/tls.cpp
@@ -46,10 +46,6 @@ void SetTcbBase(void* image_address) {
     ASSERT(result != 0);
 }
 
-Tcb* GetTcbBase() {
-    return reinterpret_cast<Tcb*>(TlsGetValue(GetTcbKey()));
-}
-
 #elif defined(__APPLE__) && defined(ARCH_X86_64)
 
 // Apple x86_64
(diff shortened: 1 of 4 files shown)


src/core/tls.cpp
```diff
@@ -165,10 +165,10 @@ void SetTcbBase(void* image_address) {
 }
 
 Tcb* GetTcbBase() {
-    void* tcb = nullptr;
+    static thread_local Tcb* tcb = nullptr;
```
squidbus: Couldn’t you just use the ‘register’ keyword to make it stored in a register?

Osyotr: `register` doesn't mean anything in in C++17 and beyond.

squidbus: That’s annoying… adding a new thread_local for this seems a bit heavy handed here to me though, I would like if there’s some alternative here (we used to just use *gsbase instructions here and I’m pretty sure that was only removed to help building on an older unsupported target…) but otherwise I guess we can accept it

src/core/tls.cpp
```diff
@@ -193,12 +176,12 @@ void SetTcbBase(void* image_address) {
     ASSERT(pthread_setspecific(GetTcbKey(), image_address) == 0);
 }
 
+#endif
+
 Tcb* GetTcbBase() {
-    return static_cast<Tcb*>(pthread_getspecific(GetTcbKey()));
+    return Libraries::Kernel::g_curthread->tcb;
```
squidbus: Probably prefer if you just replace uses of GetTcbBase with this then. Previously GetTcbBase would fetch from the same place SetTcbBase stores to, now it does not so could be confusing.

squidbus (approved):

PR #3823, opened 2025-11-22
vector_alu: Fix V_CMP_U64

raphaelthegreat:
Yet another case of "this is not actually 64-bit op" so assume for now that 64-bit compare ops are always just some thread mask in src0 and 0 in src1. Fixes handing/device loss when encountered

### DIFF
--- a/src/shader_recompiler/frontend/translate/vector_alu.cpp
+++ b/src/shader_recompiler/frontend/translate/vector_alu.cpp
@@ -1043,9 +1043,17 @@ void Translator::V_CMP_U32(ConditionOp op, bool is_signed, bool set_exec, const 
 }
 
 void Translator::V_CMP_U64(ConditionOp op, bool is_signed, bool set_exec, const GcnInst& inst) {
-    ASSERT(inst.src[0].field == OperandField::ScalarGPR &&
-           inst.src[1].field == OperandField::ConstZero);
-    const IR::U1 src0 = ir.GetThreadBitScalarReg(IR::ScalarReg(inst.src[0].code));
+    ASSERT(inst.src[1].field == OperandField::ConstZero);
+    const IR::U1 src0 = [&] {
+        switch (inst.src[0].field) {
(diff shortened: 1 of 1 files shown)


georgemoralis:
clang :/

Xcedf:
TLG hits new assert here

raphaelthegreat:
Which version of the game and what operand field it uses?

Xcedf:
v1.0.3, if I remove the first part of Assert condition it crashes on another error
<Critical> ir_emitter.cpp:131 operator(): Assertion Failed!
the second part has no difference
will check which operand field exactly

raphaelthegreat:
Print the field values for the first and second part

Xcedf:
image

raphaelthegreat:
Looks like this exposed an existing shader bug in TLG, added a commit for handling VCC. Before it would wrongly create a uint64_t from garbage and use it in the condition instead of the intended.

Before
```
bool _352 = (!_270) && (packUint2x32(uvec2(_333, 495u)) != 0ul);
if (_352)
{
    uint _356 = atomicAdd(ssbo_1_1.data[0u + buf0_dword_off], 1u);
}
```

After 
```
uint _347 = 31u & uint(_345);
bool _353 = (!_270) && (0u != _347);
if (_353)
{
    uint _357 = atomicAdd(ssbo_1_1.data[0u + buf0_dword_off], 1u);
}
```

Xcedf:
Yep, TLG's back to it's usual state

squidbus (approved):

gandalfthewhite19890404:
Unravel crashes with [Debug] <Critical> vector_alu.cpp:1046 operator(): Assertion Failed!
I don't have binaries with this PR, but with 14d71a1 (couple commits after this) it crashes. Obviously it caused by this PR.
[shad_log-14d71a1.txt](https://github.com/user-attachments/files/23981499/shad_log-14d71a1.txt)

With f6ae554 (last commit before this PR) - boots.
image

PR #3824, opened 2025-11-23
vk_rasterizer: Always submit all viewports

raphaelthegreat:
There is little point in trying to "sanitize" the viewport state like attempting to determine how many viewports are used or do any form of "compaction". So simply always parse all of them


raphaelthegreat:
Thinking about this, might not be so good of an idea, shelving for now

PR #3825, opened 2025-11-24
video_core: Reimplement inline data as buffer fill

raphaelthegreat:
Seems like it behaves as a buffer fill of a dword value instead of simply copying a single dword


squidbus (approved):
Seems to line up with how it's used in pal: https://github.com/GPUOpen-Drivers/pal/blob/c5e800072a32f68b6ccc4422936d96167c6e0728/src/core/hw/gfxip/gfx9/gfx9PipelineStatsQueryPool.cpp#L497-L515

StevenMiller123:
image
image

Fixes outlines and blur issues in DRAGON BALL XENOVERSE 2 and DRAGON BALL XENOVERSE.

StevenMiller123:
image
New push is broken.

StevenMiller123:
image
Seems to be working again now.

PR #3829, opened 2025-11-25
Texture Cache: Implement delayed destruction queue to fix use-after-free in overlapping images

mikolajmikolajczyk:
##  Summary
This PR implements a  **Delayed Destruction** mechanism (`death_row`) in the Texture Cache. It resolves critical race conditions and Use-After-Free crashes observed in  *Atomic Heart* (UE4), where aggressive texture aliasing/overlap occurs.

##  Investigation Process
Here is the step-by-step analysis that led to this solution:

1.  **Initial Crash:** `ResolveOverlap: Unreachable code!`
    * **Observation:** The game re-allocated an existing image at the same address/format but with a larger size (memory padding).
    * **Attempt:** Modified `ResolveOverlap` to allow `ExpandImage` in this scenario.

2.  **Use-After-Free:** `Access violation reading 0x110` (GpuComm thread)
    * **Observation:** `ExpandImage` destroyed the old image immediately. However, the GPU command buffer still referenced this old image for rendering/flushing.
    * **Conclusion:** We cannot `delete` resources immediately if they are currently in use by the GPU.

3.  **Memory Conflicts:** `PageManager Assertion: Not enough watchers`
    * **Attempt:** Tried keeping the old image alive indefinitely (skipping delete).
    * **Observation:** Both the old and new images tried to track the same memory pages simultaneously, confusing the `PageManager`.
    * **Conclusion:** We MUST call `UntrackImage` immediately, even if we delay the physical deletion.

##  The Solution
I implemented a **Fence-Based Garbage Collector**:

1.  **Immediate Untrack:** `UntrackImage(image_id)` is called instantly to prevent memory watcher conflicts.
2.  **Queue:** The image ID is pushed to a `death_row` queue with a target fence tick.
3.  **Safety Margin:** The target tick is calculated as `scheduler.CurrentTick() + SafetyMargin`
4.  **Synchronized Delete:** `RunGarbageCollector` checks `scheduler.IsFree(tick)`. 

##  Verified On
* **Atomic Heart:** Fixed crash after intro. Now menu loads.

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -30,6 +30,7 @@ TextureCache::TextureCache(const Vulkan::Instance& instance_, Vulkan::Scheduler&
     // Create basic null image at fixed image ID.
     const auto null_id = GetNullImage(vk::Format::eR8G8B8A8Unorm);
     ASSERT(null_id.index == NULL_IMAGE_ID.index);
+
     // Set up garbage collection parameters.
     if (!instance.CanReportMemoryUsage()) {
         trigger_gc_memory = 0;
@@ -359,7 +360,7 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
         if (image_info.type == cache_image.info.type &&
             (image_info.resources > cache_image.info.resources ||
              image_info.guest_size > cache_image.info.guest_size)) {
-            return {ExpandImageWithDelayedDestruction(image_info, cache_image_id), -1, -1};
+            return {ExpandImage(image_info, cache_image_id), -1, -1};
         }
 
         // Size is greater but resources are not, because the tiling mode is different.
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -30,7 +30,6 @@ TextureCache::TextureCache(const Vulkan::Instance& instance_, Vulkan::Scheduler&
     // Create basic null image at fixed image ID.
     const auto null_id = GetNullImage(vk::Format::eR8G8B8A8Unorm);
     ASSERT(null_id.index == NULL_IMAGE_ID.index);
-
     // Set up garbage collection parameters.
     if (!instance.CanReportMemoryUsage()) {
         trigger_gc_memory = 0;
@@ -358,8 +357,9 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
 
         // Size and resources are greater, expand the image.
         if (image_info.type == cache_image.info.type &&
-            image_info.resources > cache_image.info.resources) {
-            return {ExpandImage(image_info, cache_image_id), -1, -1};
+            (image_info.resources > cache_image.info.resources ||
+             image_info.guest_size > cache_image.info.guest_size)) {
+            return {ExpandImageWithDelayedDestruction(image_info, cache_image_id), -1, -1};
         }
 
         // Size is greater but resources are not, because the tiling mode is different.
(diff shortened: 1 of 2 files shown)


src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -358,7 +358,8 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
 
         // Size and resources are greater, expand the image.
         if (image_info.type == cache_image.info.type &&
-            image_info.resources > cache_image.info.resources) {
+            (image_info.resources > cache_image.info.resources ||
+             image_info.guest_size > cache_image.info.guest_size)) {
```
squidbus, line 362: You have essentially made `image_info.resources > cache_image.info.resources` pointless here as `image_info.guest_size > cache_image.info.guest_size` is always true here. And this is not correct, the conditions are such that if the size is greater and none of these are true then either a different case is missing or something is wrong with the size calculation.

src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -431,8 +432,7 @@ ImageId TextureCache::ExpandImage(const ImageInfo& info, ImageId image_id) {
     if (src_image.binding.is_bound || src_image.binding.is_target) {
         src_image.binding.needs_rebind = 1u;
     }
-
-    FreeImage(image_id);
+    RegisterDelayedDestruction(image_id);
```
squidbus, line 435: `FreeImage` uses `DeleteImage` which already defers image destruction, if there is an issue with that then it should be fixed rather than adding another deferred deletion method.

squidbus (commented):
This PR's steps are generally incorrect and I can tell this is AI generated. Please only submit code you came up with and fully understand.

mikolajmikolajczyk:
True, it was my mistake to ask AI for PR description...

The rest of this stupid idea was mine though ;) 
I'm just a noob so I misunderstood how this works.

Basically what I found was that all parameters but size were the same for overlapping textures:

```cpp
[Render.Vulkan] <Info> texture_cache.cpp:377 ResolveOverlap: Address: 0x2010000000
[Render.Vulkan] <Info> texture_cache.cpp:383 ResolveOverlap: Size [B]: New=16777216 vs Old=15728640 (Diff: 1048576)
[Render.Vulkan] <Info> texture_cache.cpp:388 ResolveOverlap: Format:   New=97 vs Old=97
[Render.Vulkan] <Info> texture_cache.cpp:392 ResolveOverlap: Type:     New=9 vs Old=9
[Render.Vulkan] <Info> texture_cache.cpp:396 ResolveOverlap: Levels:   New=1 vs Old=1
[Render.Vulkan] <Info> texture_cache.cpp:400 ResolveOverlap: Layers:   New=1 vs Old=1
[Render.Vulkan] <Info> texture_cache.cpp:404 ResolveOverlap: TileMode: New=14 vs Old=14
[Render.Vulkan] <Info> texture_cache.cpp:409 ResolveOverlap: Extent:   New=[1632x920x1] vs Old=[1632x920x1]
[Render.Vulkan] <Info> texture_cache.cpp:413 ResolveOverlap: Resources:   New=[Layers:1 Levels:1] vs Old=[Layers:1 Levels:1]
[Debug] <Critical> texture_cache.cpp:416 ResolveOverlap: Unreachable code!
```

This led me into doing(wrongly as from what you say - its always that new texture is bigger than old one)

```cpp
        if (image_info.type == cache_image.info.type &&
            image_info.resources > cache_image.info.resources) {
            (image_info.resources > cache_image.info.resources ||
             image_info.guest_size > cache_image.info.guest_size)) {
```

And then everything further the line went wrong because I commented out `DeleteImage(image_id);`

```cpp
    void FreeImage(ImageId image_id) {
        UntrackImage(image_id);
        UnregisterImage(image_id);
        //DeleteImage(image_id);
    }
```
and game went through to main menu. This wrongly told me that I need to delay deleting image for few frames because something is still using it. 

I will educate myself more and come back with proper solution (if its not implemented already ;)) Sorry for the mess!

PR #3832, opened 2025-11-26
0.15.1 WIP branch

### LINKED ISSUE #2637: [Feature Request]: Revert the KB/M input changes from 30.01.25
### Checklist

### Description

Diagonal movement seems to be broken from [this pull](https://github.com/shadps4-emu/shadPS4/pull/1356) onwards. Ingame (tested in Bloodborne) the character stops for a moment when switching to any diagonal direction, and the changes also seem to introduce more input latency. The way the controls were handled before these changes were better.

### Reason

For a better experience using KBM controls.

### Examples

[link to a before-after video](https://imgur.com/a/iFmFouz)
In the second part of the video you can see the character staggering for a moment when changing directions

georgemoralis:
- [x] - Separate userservice handles from pad handles,
- [ ] - Deterministically pair controllers to users,
- [ ] - Add user login screen, made with ImGui,
- [x] - Implement Move controller support,
- [x] - Finish Camera support,
- [ ] - Fix LLE Pad- and MoveTracker behaviour,
- [x] - Obtain an actual Move controller for testing,
- [x] - Fix scePadOpen issues still present in some games,
- [ ] - Refactor the input backend to support Move controllers


GHU7924:
Currently, we have two folders in the home folder: savedata and trophy. Are there plans to add other folders here that will be separated for different users? I'm thinking of folders like cheats, custom_configs, custom_trophy, input_config, and patches. I don't really know if this makes sense, but as an example, I thought of this: we could have two users at once: user 0 = tester, user 1 = shadps4 (selected by default), which, in theory, should give us more options. For example, we wouldn't have to mess with settings, adding and removing patches, and by simply switching users, we can test the emulator build and test games both with and without patches. This isn't an idea, but a question, just to see what everyone thinks about it.

georgemoralis:
deprecated

kalaposfos13:
undeprecated

ghost (dismissed):
Fixes #4038

byeugenius:
Thanks for pushing the multi-controller work - the direction looks good. I did a read-through of the multi-controller/input routing parts and found a few potential issues worth checking:

1) Potential invalid controller index from SDL
- `src/input/controller.cpp` (`GameControllers::GetGamepadIndexFromJoystickId`) uses `SDL_GetGamepadPlayerIndex(SDL_GetGamepadFromID(id))` and returns `u8` directly.
- In call sites (`src/sdl_window.cpp`, `src/input/input_handler.cpp`), this index is used for controller array access. If SDL returns `-1`, this can wrap and become an out-of-range index.

2) Handle/controller mismatch still present in pad layer
- `src/core/libraries/pad/pad.cpp`: `scePadGetHandle()` still returns constant `1`, while `scePadOpen()` returns `userId`.
- Also, some output paths still use single-controller singleton (`Common::Singleton<Input::GameController>::Instance()`) instead of selected controller id (e.g. orientation reset / lightbar / vibration paths).

3) Duplicate binding insertion in config parser
- `src/input/input_handler.cpp` around normal binding parsing: for button/hotkey, `connections.insert(...)` happens immediately, and then `connections.push_back(connection)` happens again later.
- Looks like duplicated connections for those mappings.

4) Parsed `input_gamepad_id` is mostly not applied
- `input_gamepad_id` is parsed, but not propagated into binding keys in normal path (it is only used in clone-condition logic).
- This may make explicit `:id` input mappings behave inconsistently.

5) Comparisons/searches may ignore `gamepad_id`
- In `UpdatePressedKeys`, ordering/search uses tuple of `(type, sdl_id)` but not `gamepad_id`.
- With multiple pads, same SDL button/axis ids can collide logically even when from different controllers.

6) Missing `break` fallthroughs
- `src/input/input_handler.cpp`: `case HOTKEY_REMOVE_VIRTUAL_USER` falls through into `HOTKEY_VOLUME_UP`.
- `src/sdl_window.cpp`: `case SDL_EVENT_REMOVE_VIRTUAL_USER` falls through into `SDL_EVENT_RDOC_CAPTURE`.

7) Resource handling in controller discovery
- `SDL_GetGamepads()` result should be freed.
- If all 4 slots are occupied and a new `SDL_OpenGamepad(id)` succeeds, it may not be closed on no-slot path.

If helpful, I can split these into smaller follow-up patches/checks so each can be reviewed independently.

kalaposfos13:
I'd argue copy pasting code into an LLM, then pasting the results back can hardly be described as "reading through" it, but still, the observations are valid, apart from 1, which in fact is handled already in GameControllers::operator[], and won't cause UB.

StevenMiller123:
Adding to the list, if we want to keep LLE MoveTracker, we'll either need proper HLE for libSceLibcInternal mspaces or we'll want to have a less-hacky solution for sceKernelMapSystemFlexibleMemory (and keep the malloc_init stuff in here).

Fixing up sceKernelMapSystemFlexibleMemory would be significantly easier, though in the long term we should consider the HLE mspaces route.

kalaposfos13:
At the moment, a very rudimentary Move controller support has been achieved. Currently, 4 Move controllers can be connected, but data will only be read from the first one. I also added the WIP SDL PR https://github.com/libsdl-org/SDL/pull/14719, so Move controllers can be used with emulator. Motion sensors do not work, due to the aforementioned PR not being finished, however, button and trigger data works just fine. Both myself and the original SDL PR's author are looking for testers to verify this works on different setups and on different platforms, so if you happen to be reading this and own a PS Move, please test this with my homebrew I've written to debug this: https://github.com/kalaposfos13/ps4-homebrew-base/releases/tag/move-0.01
If you also happen to have a webcam, you can try the other part of the homebrew as well, which is the Camera HLE, by setting Input - camera_id to the index of your chosen camera (0 if you only have one connected) in config.json.

(edit: the Move code got Moved to https://github.com/shadps4-emu/shadPS4/pull/4185.)

GHU7924:
The changes in this PR will significantly affect the emulator settings. Will there be conflicts with QtLauncher if you have both the old and new settings? Users may have a lot of questions about this, as they'll have to move their data to new folders, as saves and other things might not be visible. Have you resolved this issue? For example, the Flatpak version of the emulator always has only the latest (release) changes, so I wouldn't want anything to break.

georgemoralis:
updating launcher is planned

georgemoralis:
we extracted all i guess

PR #3836, opened 2025-11-29
Remove unused input code path

ngoquang2708:
This remove unused impl of 'ReadState()'.


kalaposfos13:
This PR conflicts with the User management PR, which now has my multiple controllers PR in it, as the two are closely related, and if I understand correctly, this is also doing the same thing as what I did, just reverting the controller engine class?

ngoquang2708:
Yes. If your PR already did it then I will close this.

PR #3839, opened 2025-11-29
libSceNpCommerce

thethunderturner:
Allows Sackboy a big adventure to go past the loading screen, to the menu. However, its still unstable and it works sometimes


StevenMiller123:
The actual contents are fine and could be merged, but I don't see how this would improve a game.
These are all just stubs returning zero, which is the same as what our auto-stubbing does.

thethunderturner:
> The actual contents are fine and could be merged, but I don't see how this would improve a game. These are all just stubs returning zero, which is the same as what our auto-stubbing does.

Then it could very possibly be just a false positive. I will investigate

PR #3842, opened 2025-11-30
Add basic mouse-to-touchpad emulation

kalaposfos13:
A lot of people have requested this, so here you go. It doesn't not work, but it is not very complex or robust either. The usage is that the emulator window's screen is transformed to the touchpad's input area, and a left click counts as touching the touchpad, while a right click counts as pressing it too. This is enough for simpler inputs like a swipe, and if you ever come across a situation where even this isn't enough, like an input requiring two fingers, then MAYBE CONSIDER USING A FUCKING CONTROLLER INSTEAD. The default hotkey for toggling it is delete, and there are no checks for whether you also have mouse keys bound to something else, so if you do, then that will result in both that and the touchpad being activated.

https://github.com/user-attachments/assets/72df932e-3c9e-4d5c-9922-4ed056e8100a


AussiePlick:
Don't fucking use bad language. :-)

PR #3851, opened 2025-12-02
Revert minimum CPU requirements to match the PS4's instruction set

kalaposfos13:
Originally the PR contained on optional CMake flag for using just the PS4's instruction set, and now it's changed to lowering the baseline to that level. This means that the AVX2 requirement is now dropped, and from some preliminary testing, this doesn't seem to affect performance in any way. The main target platform this PR aims to make compatible is PS4 Linux, as that is the only CPU that people using a PS4 emulator is reasonably expected to have that falls into the small list of processors that are at or above the PS4's requirements, but were under shadPS4's.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,9 +53,9 @@ else()
 endif()
 
 if (ARCHITECTURE STREQUAL "x86_64")
-    # Target x86-64-v3 CPU architecture as this is a good balance between supporting performance critical
-    # instructions like AVX2 and maintaining support for older CPUs.
-    add_compile_options(-march=x86-64-v3)
+    # Target the same CPU architecture as the PS4, to maintain the same level of compatibility.
+    # Exclude SSE4a as it is only available on AMD CPUs.
+    add_compile_options(-march=btver2 -mtune=generic -mno-sse4a)
 endif()
 
 if (APPLE AND ARCHITECTURE STREQUAL "x86_64" AND CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "arm64")


squidbus:
I'm not sure if this is worth having upstream, since that may provide an expectation it is a supported configuration. There are going to be potential differences between the compiled output that we won't be testing or supporting.

kalaposfos13:
Well from my testing done of PS4 Linux, this looks to be perfectly fine at least for the current version of the emulator, since all we use AVX2 for is free performance? With llvmpipe, the emulator does work perfectly (performance excepted) without any further modifications.

squidbus:
> Well from my testing done of PS4 Linux, this looks to be perfectly fine at least for the current version of the emulator

The point is that even if its fine for now its not a supported configuration, it could break at any time and then we're on the hook to diagnose and fix an unsupported target.

kalaposfos13:
How can this even break in the first place? I don't suppose there's anything that would require us to write Assembly code with AVX2 manually?

squidbus:
> How can this even break in the first place? I don't suppose there's anything that would require us to write Assembly code with AVX2 manually?

Any time you compile with different flags to create different output there can be differences in the resulting behavior. We just had that issue that was only showing up with interprocedural optimization enabled.

squidbus:
Let me put it this way: if we want to lower the minimum target again that's fine with me. But I don't like the idea of special flags upstream to override our minimum supported target, because then there's no point in having one.

kalaposfos13:
I guess I'll do that, then ask people on the Discord server to test how it affects performance.

kalaposfos13:
> frames seem to be identical to nightly

Doesn't affect Bloodborne's performance, we can merge it without any fear of backlash then. :kek:

LNDF:
If I understand correctly, the main use case for this is PS4 Linux.

IMO it is better to add this as a CMake option. Something like CPU_TARGET_PS4.
That way, when configuring CMake project you could do -DCPU_TARGET_PS4 and not affect the main release builds.

kalaposfos13:
@LNDF it doesn't seem to come with any obvious downsides though, and squidbus has made some valid points against making this an optional flag before.

LNDF:
> @LNDF it doesn't seem to come with any obvious downsides though, and squidbus has made some valid points against making this an optional flag before.

Ok. Just read the rest of the comments and I can see squidbis' point. So now the question comes down to whether it is worth lowering the requirements upstream in order to support ps4linux.

@raphaelthegreat was the one who raised the minimum requirements. Pinging so he can also give his POV here.

kalaposfos13:
Well technically it was [MajorP's PR](https://github.com/shadps4-emu/shadPS4/pull/2934), and it was merged by squidbus, so it wasn't raphaelthegreat, but I'm open for suggestions from anyone.

LNDF:
> Well technically it was [MajorP's PR](https://github.com/shadps4-emu/shadPS4/pull/2934), and it was merged by squidbus, so it wasn't raphaelthegreat, but I'm open for suggestions from anyone.

Turtle was the one who proposed it in my memory tracker rework pr, although MajorP did the pr

raphaelthegreat:
AVX2 is quite useful for many cases so I don't see why. Also this is a quick addition if you want to build for that platform you want so lowering the requirement for everyone just doesnt make sense. AVX2 can be used in bit tracker and in the future the readconst pass can be made to emit AVX2 loads/stores as a performance improvement

raphaelthegreat:
Also can't you pass those flags with -DCMAKE_CXX_FLAGS when configuring instead of hardcoding this into the cmake file?

squidbus:
> AVX2 can be used in bit tracker and in the future the readconst pass can be made to emit AVX2 loads/stores as a performance improvement

FWIW this is another part of why I want this to be a call on our hard minimum target rather than an optional CMake flag like the PR was originally. Because if we intend to take full advantage of AVX2, having an official option to compile for older systems will get in the way.

Heliumrich:
What's even the point of running ShadPS4 on a PS4 running Linux, when you can just reboot into the official OS 🤔 
Overhead with no added benefits (you can use mods on official OS)

Unironically curious

StevenMiller123:
Pretty sure it's mostly a "if I can, then why not?" sort of situation there.

kalaposfos13:
To quote @UltraDaCat:
> yk why I even use shad and other emus?
> alt tab
> With a PS4, I can’t just alt tab into discord :kek:

I made a joke video showcasing this, but I just now realized that without ever having seen my PS4 Linux setup before, there's no actual proof that said video really was made on my console, so I'll have to rerecord it later with fastfetch included.

PR #3852, opened 2025-12-02
Revert non-Linux parts of #3819

kalaposfos13:
This fixes Windows-specific regressions in Uncharted: The Nathan Drake Collection, Resogun, and PAYDAY 2.
It also contains some more OpenOrbis libkernel-libkernel function exports encountered in the WIP PS4 port of the eden emulator, because I'm lazy and I couldn't be bothered to make a different PR for just those 4 lines of code.

### DIFF
--- a/src/core/libraries/fiber/fiber.cpp
+++ b/src/core/libraries/fiber/fiber.cpp
@@ -6,8 +6,8 @@
 #include "common/elf_info.h"
 #include "common/logging/log.h"
 #include "core/libraries/fiber/fiber_error.h"
-#include "core/libraries/kernel/threads/pthread.h"
 #include "core/libraries/libs.h"
+#include "core/tls.h"
 
 namespace Libraries::Fiber {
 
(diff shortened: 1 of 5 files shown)


squidbus:
Do we know why using ‘Libraries::Kernel::g_curthread->tcb’ is an issue on Windows? We seem to be relying here on it working only on Linux without knowing why.

kalaposfos13:
Well, @StevenMiller123 seems to be the only one currently capable of debugging this on both Windows and Linux who also has an affected game, although currently he isn't on the most ideal setup for it.

squidbus:
Also it would be good to know if this is specifically from the linker part, fiber part, or both, and maybe narrow down where we need to use GetTcbBase() vs the current thread global.

tochnonechelovek:
PAYDAY 2 is fixed and other games have no regressions. Uncharted Nathan Drake Collection Demo is entering the game further, but not as far as it was before 3819. Something broke the game after 3819 or by something else in this PR or 3819. No way to tell without building.
[CUSA03331.log](https://github.com/user-attachments/files/23896875/CUSA03331.log)

StevenMiller123:
I personally think the whole original PR should probably be reverted, at least until we know why this happens, and why it's only Windows (there are several issues with Tcb code, particularly the cpu patches for generating accesses, but Linux is great at hiding them)

squidbus:
> I personally think the whole original PR should probably be reverted, at least until we know why this happens, and why it's only Windows (there are several issues with Tcb code, particularly the cpu patches for generating accesses, but Linux is great at hiding them)

I agree with this, main reason I was asking for more details is because it doesn’t seem sound to rely on it mysteriously only working on Linux.

PR #3857, opened 2025-12-06
cpu_patches: Patch stack canary accesses

StevenMiller123:
Some compilers emit direct reads to the stack canary (fs:[0x28]) instead of using the stack check guard provided by libkernel. This is usually done by storing the canary to a register with a `mov` instruction at the start of the function, then checking that register against the canary at the end of the function with a `xor` instruction.

In this PR, I've added code patches to handle mov and xor instructions using fs:[0x28], stubbing the returned canary value with 0. Additionally, I've updated our code to support having multiple patches per instruction, since mov instructions also need Tcb access patches.

This PR brings Red Dead Redemption 2 to menus for Windows users. It now crashes when trying to go ingame, just like how Linux behaves.
image

This is my first time working with our cpu patches, let me know if I need to change anything. Additionally, if there is a better hardcoded value I should use for the stack canary, please let me know.


src/core/cpu_patches.cpp
```diff
@@ -440,19 +465,24 @@ struct PatchInfo {
...
-
+    {ZYDIS_MNEMONIC_EXTRQ, {{FilterNoSSE4a, GenerateEXTRQ, true}}},
+    {ZYDIS_MNEMONIC_INSERTQ, {{FilterNoSSE4a, GenerateINSERTQ, true}}},
+    {ZYDIS_MNEMONIC_MOVNTSS, {{FilterNoSSE4a, ReplaceMOVNTSS, false}}},
+    {ZYDIS_MNEMONIC_MOVNTSD, {{FilterNoSSE4a, ReplaceMOVNTSD, false}}},
+
+    // fs[0x28]
+    {ZYDIS_MNEMONIC_XOR, {{FilterStackCheck, GenerateStackCheck, false}}},
+    {ZYDIS_MNEMONIC_MOV,
+     {{FilterStackCheck, GenerateStackCanary, false},
```
squidbus, line 480: I don't think these need to be enabled for Apple as we provide a usable FS segment.

squidbus (approved):

PR #3859, opened 2025-12-07
libkernel: Implement sceKernelEnableDmemAliasing, proper mapping type checks in posix_mmap

StevenMiller123:
This PR includes proper checks for dmem aliasing (implemented in a naive way for now), and adds three flags to Core::MemoryMapFlags for determining memory type.
For mmap, flag Anon signifies mappings with no backing (flex mem), flag Stack is for mapping stack memory, and flag Void is for reserving memory. These can all be verified through decompiling libkernel, as it uses these flags for mmap calls in the relevant kernel functions.

This PR needs regression testing, primarily in titles with SDK version at or above 2.50. I don't have access to the wider range of games I own, so I can't really test this myself.

Would also be nice to have someone test Red Dead Redemption 2, it is the only game I've seen sceKernelEnableDmemAliasing called in, but I can't test it right now.


georgemoralis:
rdr2 seems ok so :D

PR #3860, opened 2025-12-08
cli: add `--show-fps` to the CLI launcher

AlpinDale:
It's possible to toggle this via the config, but for quick tests, this is much more convenient. Launch the emulator with `--show-fps` to toggle the simple FPS counter.

### DIFF
--- a/src/common/config.cpp
+++ b/src/common/config.cpp
@@ -199,6 +199,7 @@ static ConfigEntry<bool> isDebugDump(false);
 static ConfigEntry<bool> isShaderDebug(false);
 static ConfigEntry<bool> isSeparateLogFilesEnabled(false);
 static ConfigEntry<bool> isFpsColor(true);
+static ConfigEntry<bool> showFpsCounter(false);
 static ConfigEntry<bool> logEnabled(true);
 
 // GUI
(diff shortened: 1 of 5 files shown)


src/common/config.cpp
```diff
@@ -969,6 +978,7 @@ void load(const std::filesystem::path& path, bool is_game_specific) {
         isSeparateLogFilesEnabled.setFromToml(debug, "isSeparateLogFilesEnabled", is_game_specific);
         isShaderDebug.setFromToml(debug, "CollectShader", is_game_specific);
         isFpsColor.setFromToml(debug, "FPSColor", is_game_specific);
+        showFpsCounter.setFromToml(debug, "ShowFpsCounter", is_game_specific);
```
kalaposfos13: I know that our config style is all over the place, but can you use camelCase here, since it's the one that's used more? Apart from this and the clang-format issue, the PR looks fine to me.

AlpinDale: Sounds good; I went with PascalCase mostly because of [this](https://github.com/shadps4-emu/shadPS4/blob/main/CONTRIBUTING.md#naming-rules)

PR #3863, opened 2025-12-10
Fixed potential UAF in `EqueueInternal::ScheduleEvent`

Rabbit0w0:
### Problem
Asynchronous timer callbacks in `ScheduleEvent()` could access the this pointer after the destruction of `EqueueInternal` object.

### Changes
- Added a life token to `EqueueInternal`.
- When scheduling a timer, we capture a weak_ptr to this token in the async_wait callback.
- Before executing the callback, check if the weak token has expired (indicating the `EqueueInternal` object has been destroyed)


ADEMOLA200:
looks solid, but clang error btw

Rabbit0w0:
Okay, style fixed

PR #3869, opened 2025-12-17
ENAMETOOLONG, posix_rename fix

marecl:
ENAMETOOLONG if path has >255 characters
Corrected posix_rename behaviour:
* handles errors only if dst exists (shouldn't error out if doesn't)
* recursively removes "old name" (would otherwise fail if it's a not-empty dir)
* actually creates target directory

[edit]
Previously: renaming dir would not create a new one nor remove the old one (no change in file structure), error when renaming dir to nonexistent target

Tested on linux so far, passes Therapist with flying colours

### DIFF
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -706,8 +706,9 @@ s32 PS4_SYSV_ABI sceKernelStat(const char* path, OrbisKernelStat* sb) {
 
 s32 PS4_SYSV_ABI sceKernelCheckReachability(const char* path) {
     if (strlen(path) > 255) {
-        return POSIX_ENAMETOOLONG;
+        return ORBIS_KERNEL_ERROR_ENAMETOOLONG;
     }
+
     auto* mnt = Common::Singleton<Core::FileSys::MntPoints>::Instance();
     std::string_view guest_path{path};
     for (const auto& prefix : available_device | std::views::keys) {
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -37,6 +37,7 @@
 #endif
 
 namespace D = Core::Devices;
+namespace fs = std::filesystem;
 using FactoryDevice = std::function<std::shared_ptr<D::BaseDevice>(u32, const char*, int, u16)>;
 
 #define GET_DEVICE_FD(fd)                                                                          \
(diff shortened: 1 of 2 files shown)


src/core/libraries/kernel/file_system.cpp
```diff
@@ -74,6 +75,12 @@ namespace Libraries::Kernel {
 
 s32 PS4_SYSV_ABI open(const char* raw_path, s32 flags, u16 mode) {
     LOG_INFO(Kernel_Fs, "path = {} flags = {:#x} mode = {:#o}", raw_path, flags, mode);
+
+    if (strlen(raw_path) > 255) {
+        *__Error() = POSIX_ENAMETOOLONG;
+        return -1;
+    }
```
StevenMiller123: Are you certain the path length is checked first here?
I can't test for this at the moment, but based on what I'd observed before, I'd expect this to come after the initial flag check.

src/core/libraries/kernel/file_system.cpp
```diff
@@ -686,6 +705,10 @@ s32 PS4_SYSV_ABI sceKernelStat(const char* path, OrbisKernelStat* sb) {
 }
 
 s32 PS4_SYSV_ABI sceKernelCheckReachability(const char* path) {
+    if (strlen(path) > 255) {
+        *__Error() = POSIX_ENAMETOOLONG;
+        return -1;
```
roamic: AFAIK you should return `ErrnoToSceKernelError(*__Error())` from all `sceKernel*` functions.

src/core/libraries/kernel/file_system.cpp
marecl: Good catch. So far it's between RW access and anything actually-file-oriented

marecl: Verified and corrected

src/core/libraries/kernel/file_system.cpp
marecl: Verified and corrected

roamic: I think there was a misunderstanding.
I meant
```cpp
    if (strlen(path) > 255) {
        *__Error() = POSIX_ENAMETOOLONG;
        return ErrnoToSceKernelError(POSIX_ENAMETOOLONG);
    }
```

marecl: oh, damn. missed this one thx

marecl:
Windows tests passed:
[TEST21370.log](https://github.com/user-attachments/files/24232367/TEST21370.log)

PR #3880, opened 2025-12-27
ajm: support for m4aac

roamic:
This is experimental.

AAC is known to be used in Dreams, Gungrave: G.O.R.E, and by AvPlayer internally (only in LLE mode).


georgemoralis:
this should be tested with games that uses directly m4aac decoder (like dreams) and games that load it through sceLibAudiodec . If you have games that loads sceLibAudiodec you can load it lle and see if it calls m4aac (gungrave does it)

brad0demx:
Using this PR, Dreams now crashes on this shortly after "An error occurred" text box pops up.
```
[Debug] <Critical> structured_control_flow.cpp:815 operator(): Assertion Failed!
Shader translation has failed
```
With NullGPU, it makes it further and I can hear that it responds to inputs. 

image

[CUSA04301-NULLGPU.log](https://github.com/user-attachments/files/24355890/CUSA04301-NULLGPU.log)
[CUSA04301.log](https://github.com/user-attachments/files/24355891/CUSA04301.log)

externals/CMakeLists.txt
```diff
@@ -258,6 +258,240 @@ if (WIN32)
 add_subdirectory(ext-wepoll)
 endif()
 
+if (NOT TARGET fdk-aac)
+    set(AACDEC_SRC
```
squidbus: This seems like a lot to inline here, would be cleaner to have a separate CMake file for fdk-aac and add that as a subdirectory. May need to move the actual git submodule one directory deeper for this, or we can fork it and add the CMake file.

roamic: Done

georgemoralis (approved):

squidbus (approved):

Issue #3291, opened 2025-07-22
[APP BUG]: Audio has brief cut offs due to frequent SDL queue overflows (on different game)


ImprovingRigmarole:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

Greetings, 

I've found those issues : #1717 and #2015 which seem related, but although they're closed, I get the same error in another game.

My setup is the following :
- shadPS4 v0.10.0
- game is CUSA01290

I'm getting the exact same symptoms, although clearly not as severe as in #1717. 
The problem occurs often enough to be noticeable, every time I think I've heard it, I check log and can see that indeed it has appeared.

```
[Lib.AudioOut] <Warning> sdl_audio.cpp:64 Output: SDL audio queue backed up (61440 queued, 61440 threshold), clearing.
[Lib.AudioOut] <Warning> sdl_audio.cpp:64 Output: SDL audio queue backed up (61440 queued, 61440 threshold), clearing.
[Lib.AudioOut] <Warning> sdl_audio.cpp:64 Output: SDL audio queue backed up (62464 queued, 61440 threshold), clearing.
[Lib.AudioOut] <Warning> sdl_audio.cpp:64 Output: SDL audio queue backed up (61440 queued, 61440 threshold), clearing.
```

I also get quite a lot of 
```
[Core] <Error> stubs.cpp:42 CommonStub: Stub: scePthreadCancel (nid: qBDmpCyGssE) called, returning zero to 0x8006b1121
```
but not sure it's related, and doesn't seem to be correlated enough to the audio warning log...

Please let me know if you have any tests, settings, or different builds I could try to help you figure the problem !

### Reproduction Steps

Run the game CUSA01290, not sure if the problem occurs in title screen nor select menu, but definitely occurs during gameplay.

### Specify OS Version

Windows 10

### CPU

AMD Ryzen 7 5700X

### GPU

NVIDIA RTX 3060

### Amount of RAM in GB

32 GB

StevenMiller123:
Could you provide a log of the issue?

ImprovingRigmarole:
sure, here's the log for my last play session

[shad_log.txt](https://github.com/user-attachments/files/21376969/shad_log.txt)

ImprovingRigmarole:
also if the info can be relevant :
[...]

Issue #2782, opened 2025-04-14
[Feature Request]: Full support for Audio3d objects


squidbus:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

The current implementation of `libSceAudio3d` primarily provides support for basic audio provided through the `sceAudio3dAudioOutOutput` and `sceAudio3dBedWrite2` functions. Some support for object audio is provided, but it is far from complete.

First off, object audio is currently just pushed to the single queue of the Audio3d port. Instead, it probably needs to be independently queued per-object and mixed down to a single audio buffer on push. If a game plays sound from a lot of object at the same time, I imagine under the current implementation it will stutter and play slowly as each object's audio buffers will be played one-by-one instead of simultaneously.

For complete object support, the implementation will need a proper 3D audio system. All objects' audio should be mixed together into the final output according to their attributes like position, gain, etc.

### Reason

Proper Audio3d support will provide working audio to more games that use it.

### Examples

N/A

ThreeDeeJay:
So is Audio3D like a reimplementation of the PS4's Gold/Platinum headset 3D audio or like an audio mixer for all games?
In my opinion, [OpenAL Soft](https://github.com/kcat/openal-soft) is hands down the best mixed for object-based [3D audio in games](https://airtable.com/appayGNkn3nSuXkaz/shrZP6E5xqQjsvplj?SEv24=b%3AWzAsWyJQM1pMeCIsNixbInNlbDBlbEhxVnZSbGRFWERFIl0sImgycHpsIl0sWyJ1YTJPdCIsNixbInNlbDBzb0xqREpMSGZiWDdPIl0sIjJZUTBPIl1d&eCzZW=recG4ZJJzBf6Xn2Yg), especially if you want to offer options for HRTF (3D audio on headphones), surround sound, or even 7.1.4 Dolby Atmos, etc.

georgemoralis:
https://github.com/shadps4-emu/shadPS4/pull/3884 should provide support for the issue soon

PR #3884, opened 2025-12-30
OpenAL sound

georgemoralis:
- [x] Added openal submodule make it build in all OS
- [x] Create an openal backend for audioOut
- [ ] Create an openal backend for AudioIn
- [ ] Replace sdl_mixer with openal
- [ ] Work on audio3d
- [x] Find a way to statically build openal
- [ ] Fix downmix for unsupported cards


georgemoralis:
tested on sonic mania it works , if you can test other games as well

georgemoralis:
fixed bloodborne sound

externals/CMakeLists.txt
```diff
@@ -268,3 +268,34 @@ add_subdirectory(json)
...
+set(ALSOFT_TESTS OFF CACHE BOOL "" FORCE)
+set(ALSOFT_INSTALL OFF CACHE BOOL "" FORCE)
+set(ALSOFT_CONFIG OFF CACHE BOOL "" FORCE)
+
+# Backends (platform-specific)
+if (WIN32)
+    set(ALSOFT_BACKEND_WASAPI ON CACHE BOOL "" FORCE)
+    set(ALSOFT_BACKEND_DSOUND OFF CACHE BOOL "" FORCE)
+    set(ALSOFT_BACKEND_WINMM OFF CACHE BOOL "" FORCE)
+else()
```
squidbus, line 292: Should not apply this for `APPLE`

georgemoralis: it shouldn't there are under a IF ?

georgemoralis:
fixed static builds no need for dll now

externals/CMakeLists.txt
squidbus: You have `if (windows) ...windows stuff... else ...linux stuff...`. Neither apply for Apple.

georgemoralis: else doesn't apply to mac and linux?

squidbus: If you want an explicit backend to enable it would be CoreAudio for macOS. But it should automatically enable it I think?

georgemoralis: don't have mac i dunno:D

squidbus: > else doesn't apply to mac and linux?

What I mean is that the ALSA/PulseAudio/PipeWire in the catch-all `else` path do not exist on macOS.

abouvier:
Can you try to use the system installed OpenAL library before unconditionally building it?

georgemoralis:
> Can you try to use the system installed OpenAL library before unconditionally building it?

I am not as good as you with cmake's (if you want to fix it i don't mind :D)

georgemoralis:
added better buffer management, should fix some cut of sounds

vzhik227:
Thanks so much for the clarification in [thread](https://github.com/shadps4-emu/shadPS4/issues/3291). I tested [this ](https://github.com/shadps4-emu/shadPS4/actions/runs/21439212957)build. On my system, it's even worse than with the current stable version 0.13. There's a significant delay in audio output, severe distortion, so it's very poor sound.
Clean build, no mods.
Windows 11 24H2 with all the latest updates. Configuration: i5 4760, 16GB, 1050ti . 30fps.
The log is littered with messages like
[Lib.AudioOut] <Warning> () openal_audio.cpp:396 ProcessBuffers: Source stopped (state=AL_STOPPED), restarting. Queued=1, Underruns=11135
[Lib.AudioOut] <Warning> () openal_audio.cpp:396 ProcessBuffers: Source stopped (state=AL_STOPPED), restarting. Queued=1, Underruns=11136
[Lib.AudioOut] <Warning> () openal_audio.cpp:396 ProcessBuffers: Source stopped (state=AL_STOPPED), restarting. Queued=1, Underruns=11137
[log-openal.txt](https://github.com/user-attachments/files/24913929/log-openal.txt)


https://github.com/user-attachments/assets/0565badf-dbf9-4c01-82d8-4bd2fc947330

I hope the information will be useful.

georgemoralis:
why not keeping @kalaposfos13 more happy

ThreeDeeJay:
Wait, does this mean OpenAL support is being abandoned, or is it being reimplemented somewhere else?
I was really looking forward to this. 🥲 
Also, I forgot to ask in https://github.com/shadps4-emu/shadPS4/issues/2782#issuecomment-2844676361: would OpenAL handle mixing on its own (and thus, support proper 3D audio mixing with HRTF) or just be used for output like in Dolphin?

georgemoralis:
There is already another pr open for this. Yes hrtf is planned as it is supported in ps4 via audio3d

PR #3900, opened 2026-01-07
common/logging: use windows only backend, in windows only

xinitrcn1:
pretty self explanatory


georgemoralis:
nothing that big but ok i will accept it :D

PR #3905, opened 2026-01-09
read compiled SDK version from eboot

roamic:
It creates an additional read pass for the elf file, but it is easier than to refactor all of the initialization sequence.


StevenMiller123 (approved):
Looks good to me

PR #3906, opened 2026-01-09
Emulator state

georgemoralis:
Moved things that are not needed on config as emulator state variables

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -840,6 +840,8 @@ set(CORE src/core/aerolib/stubs.cpp
          src/core/thread.h
          src/core/tls.cpp
          src/core/tls.h
+         src/core/emulator_state.cpp
+         src/core/emulator_state.h
 )
 
 if (ARCHITECTURE STREQUAL "x86_64")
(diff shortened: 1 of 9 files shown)


kalaposfos13 (commented):
I'm not sure if we should move all of these to EmulatorStaate, mainly showFpsCounter, as there most certainly will be people complaining that the FPS counter stopped appearing for them on startup even though they did set it in the settings.

georgemoralis:
> I'm not sure if we should move all of these to EmulatorStaate, mainly showFpsCounter, as there most certainly will be people complaining that the FPS counter stopped appearing for them on startup even though they did set it in the settings.

There isn't a setting in qtlauncher , the only way to set this is from cli

kalaposfos13 (commented):
Actually, in its current state, FpsColor is completely useless, at this point you should either get rid of it completely, or add that back to the config too, as no one would bother manually changing that every time, nor would they be able to, as there's no interface for it.

georgemoralis:
> Actually, in its current state, FpsColor is completely useless, at this point you should either get rid of it completely, or add that back to the config too, as no one would bother manually changing that every time, nor would they be able to, as there's no interface for it.

yah i don't like it too, removed it since i don't see the meaning of configure it

PR #3909, opened 2026-01-10
Kernel.Vmm: Corrected physical base handling for memory pools

StevenMiller123:
On real hardware, each pool commit can have multiple physical memory allocations, always allocating the lowest free physical addresses first.

Currently, since we currently only support one physical mapping per VMA, I create a separate VMA representing each physical allocation we perform.

This doesn't appear to help with the Unreal Engine graphical regressions triggered by https://github.com/shadps4-emu/shadPS4/pull/3639, but should fix a memory pool assert that was triggering in some games.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -314,35 +314,47 @@ s32 MemoryManager::PoolCommit(VAddr virtual_addr, u64 size, MemoryProt prot, s32
         prot |= MemoryProt::CpuRead;
     }
 
-    // Carve out the new VMA representing this mapping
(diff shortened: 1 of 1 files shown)


StevenMiller123:
This is the homebrew I used to test the behavior changes of this PR:
[pool_test_100.zip](https://github.com/user-attachments/files/24538665/pool_test_100.zip)

These are the results of running that homebrew on a 12.02 PS4, for reference.
[test.txt](https://github.com/user-attachments/files/24538667/test.txt)

PR #3913, opened 2026-01-10
Revert "common/logging: use windows only backend, in windows only"

StevenMiller123:
OutputDebugStringW is prone to crashing, this caused regressions in some titles when running outside a debugger.


StevenMiller123:
Going to make a proper fix instead, closing this.

PR #3918, opened 2026-01-11
Core: Add libSceRtc, libSceJpegDec, libSceJpegEnc, and libScePngEnc LLEs

StevenMiller123:
This PR adds three modules to our list of LLEs.

libSceRtc LLE isn't needed for anything specific right now, but works around several bugs with the HLE while we look at rewriting it.

libSceJpegDec LLE brings Trackmania (CUSA25218) ingame with broken rendering. Due to GPU emulation issues, some Linux devices don't reach ingame.

image
image

libScePngEnc LLE fixes the saving issues newer editions of Minecraft have.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,9 +148,9 @@ The following firmware modules are supported and must be placed in shadPS4's `sy
 | Modules                 | Modules                 | Modules                 | Modules                 |  
 |-------------------------|-------------------------|-------------------------|-------------------------|  
 | libSceCesCs.sprx        | libSceFont.sprx         | libSceFontFt.sprx       | libSceFreeTypeOt.sprx   |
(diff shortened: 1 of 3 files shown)


georgemoralis (approved):

kalaposfos13 (commented):
For what it's worth, if we LLE Rtc back, the HLE will probably continue being neglected and won't be getting fixes anytime soon, simply due to there being a much easier method to fix issues with it, or if a user has the LLE module dumped and loaded into the emulator, they won't come across any bugs needed fixing in the first place, stalling development.

StevenMiller123:
@georgemoralis is actively working on testing and fixing libSceRtc, which will eventually remove the need for this change.
With how messy libSceRtc decompiling is, these efforts will take time.

PR #3923, opened 2026-01-13
Lib.Audio3d: sceAudio3dGetDefaultOpenParameters fix

StevenMiller123:
sceAudio3dGetDefaultOpenParameters should only copy 0x20 bytes to the parameter output.
We were originally writing 0x28 bytes, which could corrupt the stack in titles using the function.

This fixes some cases of `kernel.cpp:81 stack_chk_fail: Unreachable code!` in games using libSceAudio3d.


StevenMiller123:
This fixes the old regression in Jackal Assault (CUSA06832) and brings it back ingame.

image
image

StevenMiller123:
I think I need to look at this closer, swapping to draft.

StevenMiller123:
Should be ready for review now.

PR #3926, opened 2026-01-15
shader_recompiler: some fixes for tess shaders

baggins183:
When walking the users of special constants which form LDS addresses:
-ignore when a user contributes to the wrong operand of an LDS inst, for example the data operand of WriteShared* instead of the address operand. This can mistakenly happen due to phi nodes.
-don't use flags to stash temp info about phis, since flags may already be in use. Use a separate map.


StevenMiller123:
Fixes `[Debug] <Critical> hull_shader_transform.cpp:247 operator(): Assertion Failed!` in Call of Duty®: Advanced Warfare (CUSA00803), bringing the title's singleplayer mode ingame.
image
image
image
image
image
image

[CUSA00803.log](https://github.com/user-attachments/files/24630223/CUSA00803.log)

PR #3932, opened 2026-01-17
Kernel.Vmm: Handle sparse physical memory usage + other fixes

StevenMiller123:
This PR rewrites chunks of address space and memory code to properly handle sparse physical memory usage for pool and flexible memory. 

Notable changes:
- AddressSpace unmaps no-longer need a physical address, or the various VMA offset parameters we used to use. All relevant information needed to properly unmap memory is stored within AddressSpace now.
- Mapped/Pooled virtual memory areas will now merge when the areas are similar enough. Previously, restrictions with Windows AddressSpace unmaps prevented this. This change should improve memory performance in games that spam small memory mappings.
- Memory code now takes advantage of our SharedFirstMutex, allowing multi-threaded memory map accesses (like virtual queries). This should theoretically help memory performance, though I don't think the change will be too noticeable.
- Virtual memory areas now store a map of all physical memory they encompass. This is needed to fix the regressions from #3909, and to properly enable merging flexible and pooled memory.

This PR affects a lot of code, so this probably needs a lot of testing before any merging can happen. I would especially like to see performance tests from games that spam memory calls, as I've made various changes that can both hurt and improve performance here.

I've tested a couple of notable edge cases in my games, and this code seems to function at least.

### DIFF
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -686,8 +686,8 @@ struct AddressSpace::Impl {
 
     void Unmap(VAddr virtual_addr, u64 size, bool) {
         // Check to see if we are adjacent to any regions.
-        auto start_address = virtual_addr;
-        auto end_address = start_address + size;
+        VAddr start_address = virtual_addr;
+        VAddr end_address = start_address + size;
         auto it = m_free_regions.find({start_address - 1, end_address + 1});
 
         // If we are, join with them, ensuring we stay in bounds.
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -93,7 +93,10 @@ static u64 BackingSize = ORBIS_KERNEL_TOTAL_MEM_DEV_PRO;
 
 struct MemoryRegion {
     VAddr base;
-    size_t size;
+    PAddr phys_base;
+    u64 size;
+    ULONG prot;
+    s32 fd;
     bool is_mapped;
 };
 
@@ -159,7 +162,8 @@ struct AddressSpace::Impl {
             // Restrict region size to avoid overly fragmenting the virtual memory space.
             if (info.State == MEM_FREE && info.RegionSize > 0x1000000) {
                 VAddr addr = Common::AlignUp(reinterpret_cast<VAddr>(info.BaseAddress), alignment);
-                regions.emplace(addr, MemoryRegion{addr, size, false});
+                regions.emplace(addr,
+                                MemoryRegion{addr, PAddr(-1), size, PAGE_NOACCESS, -1, false});
             }
         }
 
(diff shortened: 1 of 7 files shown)


StevenMiller123:
New commits include a rewrite to our Windows address space map and unmap to make them more versatile.
This should make some merging and splitting logic easier to handle, though I haven't done any updates to memory code to take advantage of this yet.

StevenMiller123:
I've made all the changes I'd like to make, so this should be ready for proper testing and review now.

StevenMiller123:
Did some local testing, patched some silly bugs my code had, hopefully it works properly?

tochnonechelovek:
CUSA01761 crashes on launch
`[Debug] <Critical> (Thread2) memory.cpp:1448 operator(): Assertion Failed!`
[CUSA01761.log](https://github.com/user-attachments/files/24752395/CUSA01761.log)

StevenMiller123:
Just ran into that one myself, I'll take a look

tochnonechelovek:
CUSA05527 and CUSA04295 crashes sooner with the same error (not a good thing btw :D )

StevenMiller123:
Thanks, I've got one of those, so that'll speed up debugging

StevenMiller123:
Fixed some cases, but broke other games.
I'll do some more testing locally to make sure, but it seems like I'll need to implement a compiled SDK version check the PS4 kernel has in order to properly fix some of the bugs I'm seeing.

brad0demx:
Tested on my library and I don't see any regressions. 
I looked out for performance gains/losses on my more spammy games, and didn't see any noticeable differences

StevenMiller123:
@tochnonechelovek When you can, could you retest to see if I fixed the issues you had?

KITT1995:
Readback linear images not working in this build. Tested with CUSA01073.

StevenMiller123:
Could you provide a log?

KITT1995:
here it is
[shad_log.txt](https://github.com/user-attachments/files/24780429/shad_log.txt)

StevenMiller123:
I can't reproduce anything like this in my titles.
Are there any others seeing a regression in Ratchet and Clank after this merged?

Also, @KITT1995, when testing PRs, please disable any patches and any unnecessary experimental settings too.

KITT1995:
The "Bypass hang" patch was necessary to be able to run this title without readback

StevenMiller123:
It's not needed to run with readbackLinearImages enabled.
Also you had readbacks and direct memory access enabled too.

KITT1995:
without readbackLinearImages it hangs after start new game. This is an old patch when readback wasn't implemented.
Disabling readbacks and direct memory access doesn't change anything, apart from faster loading time.

PR #3935, opened 2026-01-18
Introducing key_manager for storing encryption keys .

georgemoralis:
Correctly only trophy key , but can be expand if more encryption keys are neccesary in future. 
Solve also the issue of storing trophy key in the same settings file.
Also added copy from config.toml file if already set there so i don't break the current configurations (will be removed after a while)


src/core/file_format/trp.cpp
```diff
@@ -63,91 +50,232 @@ bool TRP::Extract(const std::filesystem::path& trophyPath, const std::string tit
...
+
+    std::vector<u8> ESFM(entry.entry_len - IV_LEN);
+    std::vector<u8> XML(entry.entry_len - IV_LEN);
+
+    if (!file.Read(ESFM)) {
+        LOG_ERROR(Common_Filesystem, "Failed to read encrypted XML data");
+        return false;
+    }
+
+    // Decrypt the data - FIX: Don't check return value since DecryptEFSM returns void
```
StevenMiller123, line 251: What's this comment for, the later DecryptESFM call?

StevenMiller123 (commented):
Code looks fine, though I can't test right now.

src/core/file_format/trp.cpp
georgemoralis: oh well probably didn't write it correctly , was meant that will have to add some detection if decryption fails

PR #3938, opened 2026-01-19
WebDialogBrowser module

georgemoralis:
-Currently doesn't do anything but can stop spamming on some games that only inits it but never use it (a.k.a carmageddon)


StevenMiller123 (changes requested):
Would be better to have some sort of proper state emulation, like setting status to complete after running sceWebBrowserDialogOpen.
Just swapping the log to trace will hide one potential cause of game hangs.

georgemoralis:
Yes i am planning to add more to this , although from test either of the 5 games i test never call Open. They are all do the normal procedure , checking and closing a common dialog if exists (webbrowserdialog behaves that way) but never really open a webbrowser one

PR #3941, opened 2026-01-20
video_core: Small readback optimization

raphaelthegreat:
This is an implementation of the DispatchDirect self modification elimination patch with a simple implementation on its own that shouldn't break anything hopefully. It detects when a DmaData is copying memory to the current command list, and specifically to the dimentions of a DispatchDirect packet and a) skips the copy so the region isn't marked as gpu modified and flushed b) patches the packet into an indirect dispatch to the source buffer.

Should provide a speedup to affected cases and allow them to avoid device loss errors when readbacks are disabled

### DIFF
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -929,7 +929,8 @@ Liverpool::Task Liverpool::ProcessCompute(std::span<const u32> acb, u32 vqid) {
                 const u32 num_bytes = dma_data->NumBytes();
                 const VAddr src_addr = dma_data->SrcAddress<VAddr>();
                 const VAddr dst_addr = dma_data->DstAddress<VAddr>();
-                const PM4Header* header = reinterpret_cast<const PM4Header*>(dst_addr - sizeof(PM4Header));
+                const PM4Header* header =
+                    reinterpret_cast<const PM4Header*>(dst_addr - sizeof(PM4Header));
                 if (dst_addr >= base_addr && dst_addr < base_addr + acb_size &&
                     num_bytes == sizeof(PM4CmdDispatchIndirect::GroupDimensions) &&
                     header->type == 3 && header->type3.opcode == PM4ItOpcode::DispatchDirect) {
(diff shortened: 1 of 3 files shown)


red-prig:
you forgot an important comment about the packet overflow case `return ( count + 1 ) & 0x3fff ;`

raphaelthegreat:
You mean I forgot something in the code or forgot to mention in PR description (which is true)

red-prig:
yep description

StevenMiller123:
With this PR, DRIVECLUB™ (CUSA00093) can run without enabling readbacks.
image
[CUSA00093.log](https://github.com/user-attachments/files/24751791/CUSA00093.log)

bigol83:
I checked the performance in Bloodborne and with this it is slower compared to master

same spot, 28 fps on PR, 31 fps on master

raphaelthegreat:
Bloodborne does not use this so it should not be affected. Maybe if you are running with readbacks, the extra logging added is slowing it down instead

bigol83:
> Bloodborne does not use this so it should not be affected. Maybe if you are running with readbacks, the extra logging added is slowing it down instead

disabled logging for Bloodborne and now performance is exactly the same in PR and master

georgemoralis:
if we haven't any regressions we can merge?

raphaelthegreat:
The log that was added should be removed first to avoid the perf degradation

TPepperoni666:
I'll add my regression here. Cannot get anything to show bar the UI and effects. Driveclub on a 4090. Tired combinations of readbacks on and off as well as linear images and DMA.
Screenshot 2026-01-20 100241
Screenshot 2026-01-20 080727
Screenshot 2026-01-20 074355

Randomuser8219:
This pull request regresses Marvel's Spider-Man (CUSA02299). It used to consistently crash at the PM4 SignalFence error, but now it freezes during the Marvel logo.
[shad_log before readback opt.txt](https://github.com/user-attachments/files/24798672/shad_log.before.readback.opt.txt)
[shad_log latest.txt](https://github.com/user-attachments/files/24798683/shad_log.latest.txt)

PR #3942, opened 2026-01-20
video_core: Small fixes regarding GDS

raphaelthegreat:
* First commit fixes an issue where the transformation of GDS atomics to buffer atomics would not set the buffer descriptor type, causing a silent crash in spirv backend due to missing buffer type alias
* Implements GDS to memory copy mode of ReleaseMem which affects https://github.com/shadps4-compatibility/shadps4-game-compatibility/issues/581

### DIFF
--- a/src/video_core/amdgpu/pm4_cmds.h
+++ b/src/video_core/amdgpu/pm4_cmds.h
@@ -327,6 +327,7 @@ enum class DataSelect : u32 {
     Data64 = 2,
     GpuClock64 = 3,
     PerfCounter = 4,
+    GdsMemStore = 5,
 };
 
 enum class InterruptSelect : u32 {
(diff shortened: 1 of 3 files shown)


StevenMiller123:
v1.00 of Marvel's Spider-Man now crashes on
```
[Debug] <Critical> (shadPS4:GpuCommandProcessor) vk_scheduler.cpp:164 operator(): Assertion Failed!
Device lost during submit
```

[CUSA02299.log](https://github.com/user-attachments/files/24752700/CUSA02299.log)

src/video_core/amdgpu/pm4_cmds.h
```diff
@@ -920,8 +921,8 @@ struct PM4CmdReleaseMem {
     u32 data_hi;
 
     template <typename T>
-    T* Address() const {
-        return reinterpret_cast<T*>(address_lo | u64(address_hi) << 32);
+    T Address() const {
+        return reinterpret_cast<T>(address_lo | u64(address_hi) << 32);
```
squidbus: Need `std::bit_cast`

georgemoralis: this should be ok now

PR #3946, opened 2026-01-22
Core: Miscellaneous memory fixes and slight optimizations

StevenMiller123:
After #3932, I received many mentions of performance regressions. From my own testing, these (as per usual) came from performing extra Microsoft API calls. Specifically, in order to shape placeholders properly for mapping memory (instead of requiring vma_map to match AddressSpace regions), my PR added multiple VirtualFreeEx calls over our old code.

While looking over the current Windows address space code, I found some areas that could be made faster. Specifically, this PR removes a VirtualProtect call we performed during physically backed address space mappings, and an unnecessary VirtualFreeEx call performed while coalescing regions.

In games limited by memory performance (and as a result, regressed by #3932), this should improve performance and bring it closer to where v0.13.0 was while keeping all the improvements of my PR.

I've also made two other smaller changes here, first I've updated our read-only file mmap code to properly handle file offsets, and secondly I've disabled Tracy memory tracking since it broke with VMA merging.

Opening as a draft until people test, I'd like to see the impacts of this before letting it get merged.

### DIFF
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -255,8 +255,7 @@ struct AddressSpace::Impl {
                 ASSERT_MSG(ret, "VirtualProtect failed. {}", Common::GetLastErrorMsg());
             } else {
                 ptr = MapViewOfFile3(backing, process, reinterpret_cast<PVOID>(virtual_addr),
-                                     phys_addr, size, MEM_REPLACE_PLACEHOLDER,
-                                     prot, nullptr, 0);
+                                     phys_addr, size, MEM_REPLACE_PLACEHOLDER, prot, nullptr, 0);
                 ASSERT_MSG(ptr, "MapViewOfFile3 failed. {}", Common::GetLastErrorMsg());
             }
         } else {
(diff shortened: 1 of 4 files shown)


KITT1995:
CUSA01073 still regressed with readbacklLinearImages
[shad_log.txt](https://github.com/user-attachments/files/24807247/shad_log.txt)

StevenMiller123:
@KITT1995 Given the circumstances, I consider it likely that you've found the wrong cause. A hang after compiling shaders would not be typically be caused by a PR targeting memory code.

I do not own Ratchet & Clank, so I can't debug it myself. Don't expect a fix from me unless someone else can debug the issue and identify exactly how my PR broke it (if it did).

StevenMiller123:
New commits contain a bunch of bugfixes for things I actually broke, though not sure if these will help with any of the remaining performance regressions tied to #3932

StevenMiller123:
There's still one known regression from #3932, but I'm unable to figure out the cause at this time.
I've ordered the regressed game and will look at fixing it when I've dumped it.

Until then, I'm setting this as ready for review.

StevenMiller123:
@KITT1995 Newest commits probably solve the Ratchet & Clank regression?
If you could, please give that a test when you can (and apologies for my previous statement about that issue 😅)

KITT1995:
Ratchet & Clank regression fixed with ac7f3b6 👍

PR #3952, opened 2026-01-23
Libraries: libSceCamera implementation

kalaposfos13:
This PR implements libSceCamera using SDL3's SDL_camera API. Currently WIP, but so far, all the core code is hooked up, and everything that my [test homebrew](https://github.com/kalaposfos13/ps4-homebrew-base/tree/camera) is using is implemented in enough detail for a correct image to be displayed. Actual games using the camera will likely not work just yet, and ironically enough, the PR doesn't work correctly with the actual PS4 Camera either, as the frame data from it has both eyes submitted in one buffer, which I'll need to separate into their own buffers manually. That is not to say it won't display anything, the resulting image will just be very distorted.

To get started, hook up a webcam, capture card, or anything else that can be described as a digital video input device, head to `config.toml`, and in the `[Input]` section, set the newly added `cameraId` setting to the ID of your chosen input. This is a list of indexes starting from zero, so if you only have one device, set this to 0.


src/core/libraries/camera/camera.cpp
```diff
@@ -649,6 +649,11 @@ s32 PS4_SYSV_ABI sceCameraSetConfig(s32 handle, OrbisCameraConfig* config) {
     case ORBIS_CAMERA_CONFIG_TYPE3:
     case ORBIS_CAMERA_CONFIG_TYPE4:
     case ORBIS_CAMERA_CONFIG_TYPE5:
+        int sdk_ver;
+        Libraries::Kernel::sceKernelGetCompiledSdkVersion(&sdk_ver);
+        if (sdk_ver < 0x45000000) {
```
StevenMiller123: So uh, firmware 45.0 when?
Assuming you intended to check against firmware 4.50, you should use Common::ElfInfo::FW_45 instead. If you intentionally put firmware 45 here, then I guess make a Common::ElfInfo::FW_450? 😄

kalaposfos13:
Moved to main-2, also known as https://github.com/shadps4-emu/shadPS4/pull/3832.

PR #3955, opened 2026-01-24
Refactor OrbisFiber to match PS4 layout and obfuscation

w1naenator:
- Redesign OrbisFiber struct to match PS4's internal layout, including new fields, padding, and explicit size/alignment checks.
- Store entry, arg_on_initialize, and name in obfuscated (XORed) form; add helper functions for encoding/decoding.
- Replace plain name string with XORed byte array and per-fiber randomization.
- Add random_pad field, filled with pseudo-random data on init.
- Use atomic helpers for fiber state transitions instead of direct atomic access.
- Update all code to use new accessors for obfuscated fields.
- Improve stack overflow and context size margin checks.
- Zero-initialize OrbisFiberOptParam in its initializer.
- Add <cstring> and <cstddef> includes for new memory operations.
- Miscellaneous bug fixes and code cleanup.


w1naenator:
This fixes the emulator crashing in Gravity Rush 2 on Windows, which I used to test the font library. Together with the fontlib branch in HLE mode, the font is now visible. I am not sure whether this is usable in the project, but you can test it.

src/core/libraries/fiber/fiber.cpp
```diff
@@ -39,6 +165,10 @@ void PS4_SYSV_ABI _sceFiberCheckStackOverflow(OrbisFiberContext* ctx) {
     u64* stack_base = reinterpret_cast<u64*>(ctx->current_fiber->addr_context);
     u64 stack_size = ctx->current_fiber->size_context;
     if (stack_base && *stack_base != kFiberStackSignature) {
+        if (*stack_base == 0) {
+            *stack_base = kFiberStackSignature;
```
squidbus: Seems like a hack? What is this behavior based on?

src/core/libraries/fiber/fiber.cpp
```diff
@@ -184,22 +314,21 @@ s32 PS4_SYSV_ABI sceFiberInitializeImpl(OrbisFiber* fiber, const char* name, Orb
...
     fiber->context = nullptr;
+    fiber->owner_thread = 0;
     fiber->flags = user_flags;
-
-    /*
-        A low stack area is problematic, as we can easily
-        cause a stack overflow with our HLE.
-    */
-    if (size_context && size_context <= 4096) {
-        LOG_WARNING(Lib_Fiber, "Fiber initialized with small stack area.");
```
squidbus, line 201: Why is this removed?

src/core/libraries/fiber/fiber.cpp
StevenMiller123: image

For reference, here's a basic decomp of the actual library code.
Unless I'm mistaken, this is probably a hack.

roamic: Our HLE is probably taking too much stack space in the fiber. The correct fix would be to identify the faulty HLE module and try to reduce the stack usage.

w1naenator: Yes, you are right. That was my hack; I had forgotten about it.

roamic: I have investigated and the stack breaks in ffmpeg during `avformat_open_input`. Nothing we can do for now until we replace it with mp4v2 or another mp4 library.

w1naenator:
Actually, after rewriting the fiber library, I found that the TLS hygiene (ClearStack<12_KB>()) causes a stack overflow on Windows in the fiber library. Please check this; I think the structs should better match the ABI.

tochnonechelovek:
No changes in my titles.

wuguo13842:
I'm experiencing a "picture-in-picture" rendering issue exclusively on my NVIDIA RTX 2070 Ti (Windows) while testing Gravity Rush 2 with this PR (gr2fix branch). The problem does not occur on AMD GPUs or Linux systems according to user reports.

Description
The game renders a downsized copy of the scene in the top‑left quarter of the screen, creating a recursive feedback effect. The rest of the screen renders correctly.

Investigation so far

The fragment shader that blends the depth‑of‑field layer uses correct full‑screen UV coordinates.

The compute shader (hash 0x8503bcb7) writes a full‑screen texture but only processes pixels whose coordinates are less than two values read from ssbo_1 at indices 4 and 5.

The original values at those indices are 1920.0 and 1088.0 (as floats). Even after forcing them to 4096.0 (to cover the whole screen), the problem persists.

Logs confirm that the forced values are successfully written.

This suggests the issue may lie elsewhere – possibly a missing memory barrier, incorrect image layout transition, or resource aliasing that only manifests on NVIDIA drivers.

Steps to reproduce

Build this PR on Windows with an NVIDIA GPU.

Launch Gravity Rush 2 and observe the top‑left corner after the first few frames.

I've attached relevant logs showing the forced buffer writes. Any insight or guidance on further debugging would be greatly appreciated. Thanks!

[is RenderDoc file here](https://1drv.ms/u/c/76f2fe0b033370d0/IQDQBFqC2Yu5SafVTR9YW2c7AUZ2Do3FzaDERknHSLaFSPQ?e=d653BN)
https://1drv.ms/u/c/76f2fe0b033370d0/IQDQBFqC2Yu5SafVTR9YW2c7AUZ2Do3FzaDERknHSLaFSPQ?e=d653BN
https://1drv.ms/u/c/76f2fe0b033370d0/IQDA57nZJEUIRoMFCp6v_DyjAeJFSS2lT-xzRF9ZFdE4rl8?e=6maewH

georgemoralis:
#4277 , #4274  should solve the issues

PR #3956, opened 2026-01-24
Kernel.Vmm: Attempt to address race conditions involving ClampRangeSize, CopySparseMemory, and TryWriteBacking

StevenMiller123:
This PR makes ClampRangeSize, CopySparseMemory, and TryWriteBacking thread safe to fix some issues caused by #3954, and in doing so, makes a variety of changes to unmap-related logic to prevent deadlocking.

Specifically, I've adjusted all functions with GPU unmaps to perform their GPU unmaps as soon as possible, while locking a separate mutex I've added. When that operation is done, I attempt to lock for writing before these functions will modify vma_map.

The extra mutex is also now locked for all functions that modify vma_map, since we don't want vma_map modifications occurring while maps/unmaps are performing validity checks.

By using a separate mutex instead of locking reader locks, I can avoid race conditions with trying to transition from reader to writer locks, and ensure we don't have multiple threads trying to perform maps at the same time (and potentially invalidating the results of another thread's validity checks).

Also some smaller changes: I've cleaned up unmap code a bit, removing some unused variables that got left behind in my prior PRs, and I've added a try_lock to SharedFirstMutex to enable using some extra standard library lock types with it.

While I've tested locally with some of my heavily multi-threaded games, I'm willing to bet there are a plethora of bugs that need fixing here, so I'm opening this as a draft.

### DIFF
--- a/src/common/shared_first_mutex.h
+++ b/src/common/shared_first_mutex.h
@@ -17,6 +17,15 @@ public:
         writer_active = true;
     }
 
+    bool try_lock() {
+        std::lock_guard<std::mutex> lock(mtx);
+        if (writer_active || readers > 0) {
+            return false;
+        }
+        writer_active = true;
+        return true;
+    }
+
     void unlock() {
         std::lock_guard<std::mutex> lock(mtx);
         writer_active = false;
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -709,7 +709,7 @@ struct AddressSpace::Impl {
         return ret;
     }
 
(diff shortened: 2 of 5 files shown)


StevenMiller123:
Based on what I'm hearing, it sounds like this PR works as intended.
Marking as ready for review.

StevenMiller123:
Still running into issues when testing locally. Reverting to draft for now.

StevenMiller123:
Scratch that, I was locally testing an old build 🤦

georgemoralis:
let's break something today

PR #3958, opened 2026-01-25
shader_recompiler: VS clip distance emulation for NVIDIA GPUs

psucien:
It looks like NVIDIA GPUs have precision issues with clip distances interpolation. At least the result slightly differs from what we see on AMD (and there is evidence even on Intel) GPUs, which usually results in artifacts like black stripes on tilted polygons. This PR adds emulation of VS outputted clip distances via discards on FS stage for NVIDIA cards.

### DIFF
--- a/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
+++ b/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
@@ -364,7 +364,7 @@ void EmitContext::DefineInputs() {
         }
         break;
     }
-    case LogicalStage::Fragment:
+    case LogicalStage::Fragment: {
         if (info.loads.GetAny(IR::Attribute::FragCoord)) {
             frag_coord = DefineVariable(F32[4], spv::BuiltIn::FragCoord, spv::StorageClass::Input);
         }
@@ -418,7 +418,13 @@ void EmitContext::DefineInputs() {
                                                     spv::StorageClass::Input);
             }
         }
-        for (s32 i = 0; i < runtime_info.fs_info.num_inputs; i++) {
+
+        const bool has_clip_distance_inputs = runtime_info.fs_info.clip_distance_emulation;
+        // Clip distances attribute vector is the last in inputs array
+        const auto num_inputs =
+            runtime_info.fs_info.num_inputs - (has_clip_distance_inputs ? 1 : 0);
+
+        for (s32 i = 0; i < num_inputs; i++) {
             const auto& input = runtime_info.fs_info.inputs[i];
             if (input.IsDefault()) {
                 continue;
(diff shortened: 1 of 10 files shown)


georgemoralis:
hint: water on bloodborne

raphaelthegreat:
Have you tested adding a small bias of the written value instead?

par274:
Fix water regression on Bloodborne. 

However, there's a problem: if you revert to a previous version, the game won't launch without clearing the shader cache. A warning about this could be added to the logs.

larfy298:
Shader cache crashes are a very common thing from what I have seen, it is not an issue exclusive to this pr by any means. I suggest that a "Clear Shader Cache" button should be added to the game-specific config, though there is probably a better fix possible which I am not nearly knowledgeable enough to figure out.

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
```diff
@@ -546,11 +561,16 @@ void EmitContext::DefineVertexBlock() {
...
                                  f32_zero_value, f32_zero_value, f32_zero_value, f32_zero_value};
     output_position = DefineVariable(F32[4], spv::BuiltIn::Position, spv::StorageClass::Output);
-    if (info.stores.GetAny(IR::Attribute::ClipDistance)) {
-        const Id type{TypeArray(F32[1], ConstU32(8U))};
-        const Id initializer{ConstantComposite(type, zero)};
-        clip_distances = DefineVariable(type, spv::BuiltIn::ClipDistance, spv::StorageClass::Output,
-                                        initializer);
+    const bool needs_clip_distance_emulation = l_stage == LogicalStage::Vertex &&
+                                               stage == Stage::Vertex &&
+                                               profile.needs_clip_distance_emulation;
```
squidbus, line 566: Why is this based on `profile.needs_clip_distance_emulation` and not `fs_info.clip_distance_emulation`?

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
```diff
@@ -583,16 +603,27 @@ void EmitContext::DefineOutputs() {
                 Name(output_attr_array, "out_attrs");
             }
         } else {
+            const auto has_clip_distance_outputs = info.stores.GetAny(IR::Attribute::ClipDistance);
```
squidbus, line 606: Why does this apply below changes regardless of whether clip distance emulation is enabled?

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
```diff
@@ -546,11 +561,16 @@ void EmitContext::DefineVertexBlock() {
...
-    if (info.stores.GetAny(IR::Attribute::ClipDistance)) {
-        const Id type{TypeArray(F32[1], ConstU32(8U))};
-        const Id initializer{ConstantComposite(type, zero)};
-        clip_distances = DefineVariable(type, spv::BuiltIn::ClipDistance, spv::StorageClass::Output,
-                                        initializer);
+    const bool needs_clip_distance_emulation = l_stage == LogicalStage::Vertex &&
+                                               stage == Stage::Vertex &&
+                                               profile.needs_clip_distance_emulation;
+    if (!needs_clip_distance_emulation) {
+        if (info.stores.GetAny(IR::Attribute::ClipDistance)) {
```
squidbus, line 568: This can be collapsed into one `if` block.

squidbus (commented):
Late review here but this has some things that look problematic to me.

PR #3959, opened 2026-01-25
Fake Matching2

mikusp:
Initial work for NpMatching2 used for setting up multiplayer rooms in some games. Enabled only when respective config values are set (isConnectedToNetwork and isPSNSignedIn).

### DIFF
--- a/src/core/libraries/np/np_matching2.cpp
+++ b/src/core/libraries/np/np_matching2.cpp
@@ -1,0 +1,812 @@
+// SPDX-FileCopyrightText: Copyright 2026 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <deque>
+#include <mutex>
+
+#include "common/config.h"
+#include "common/logging/log.h"
+#include "core/libraries/error_codes.h"
+#include "core/libraries/libs.h"
+#include "core/libraries/np/np_manager.h"
+#include "core/libraries/np/np_matching2.h"
+#include "core/libraries/np/np_types.h"
+#include "core/libraries/system/userservice.h"
+
+namespace Libraries::Np::NpMatching2 {
+
+static bool g_initialized = false;
+static OrbisNpMatching2ContextId contextId = 1;
+
+struct NpMatching2ContextEvent {
+    OrbisNpMatching2ContextId contextId;
+    OrbisNpMatching2Event event;
+    OrbisNpMatching2EventCause cause;
+    int errorCode;
+};
+
+struct NpMatching2LobbyEvent {
+    OrbisNpMatching2ContextId contextId;
+    OrbisNpMatching2LobbyId lobbyId;
+    OrbisNpMatching2Event event;
+    void* data;
+};
+
+struct NpMatching2RoomEvent {
+    OrbisNpMatching2ContextId contextId;
+    OrbisNpMatching2RoomId roomId;
+    OrbisNpMatching2Event event;
+    void* data;
+};
+
+static std::mutex g_events_mutex;
+static std::deque<NpMatching2ContextEvent> g_ctx_events;
+static std::deque<NpMatching2LobbyEvent> g_lobby_events;
+static std::deque<NpMatching2RoomEvent> g_room_events;
+static std::mutex g_responses_mutex;
+static std::deque<std::function<void()>> g_responses;
+
+struct OrbisNpMatching2CreateContextParameter {
+    Libraries::Np::OrbisNpId* npId;
+    void* npCommunicationId;
+    void* npPassphrase;
+    Libraries::Np::OrbisNpServiceLabel serviceLabel;
+    u64 size;
+};
+
+static_assert(sizeof(OrbisNpMatching2CreateContextParameter) == 0x28);
+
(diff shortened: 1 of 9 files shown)


src/core/libraries/np/np_matching2.cpp
```diff
@@ -0,0 +1,694 @@
...
+struct OrbisNpMatching2CreateJoinRoomResponseA {
+    OrbisNpMatching2RoomDataInternal* roomData;
+    OrbisNpMatching2RoomMemberDataInternalListA members;
+};
+
+int PS4_SYSV_ABI sceNpMatching2CreateJoinRoomA(OrbisNpMatching2ContextId ctxId,
+                                               OrbisNpMatching2CreateJoinRoomRequestA* request,
+                                               OrbisNpMatching2RequestOptParam* requestOpt,
+                                               OrbisNpMatching2RequestId* requestId) {
+    LOG_DEBUG(Lib_NpMatching2, "called, ctxId = {}, requestOpt = {}", ctxId, (void*)requestOpt);
```
StevenMiller123: Should use fmt::ptr instead of typecasting to log pointers.

src/core/libraries/np/np_matching2.cpp
```diff
@@ -0,0 +1,694 @@
...
+                  optParam->appId);
+        std::scoped_lock lk{g_responses_mutex};
+        auto reqIdCopy = *requestId;
+        auto requestCopy = *request;
+        g_responses.emplace_back([=]() {
+            OrbisNpMatching2RoomMemberDataInternalA me{
+                nullptr,
+                0,
+                {0xace104e, Libraries::Np::OrbisNpPlatformType::ORBIS_NP_PLATFORM_TYPE_PS4},
+                {{'a', 'c', 'e', 't', 'o', 'n', 'e', '\0'}, 0, {0, 0, 0}},
```
StevenMiller123: Probably use configured user names instead of hardcoding one.

StevenMiller123 (changes requested):
I see a distinct lack of checks against PSN here.
Are you sure none of these functions would encounter a PSN error?

Also some minor nitpicks.

mikusp:
From what I've seen in testing, games won't even try to call those if more basic psn checks failed before

StevenMiller123:
Doesn't matter, we need to be accurate when handling the few games that might.

StevenMiller123:
If needed, I can always write a homebrew for the sole purpose of being a homebrew to "test NpMatching2 with no PSN" 😄

mikusp:
> If needed, I can always write a homebrew for the sole purpose of being a homebrew to "test NpMatching2 with no PSN" 😄

I'm actually interested in what would the real hardware do

src/core/libraries/np/np_matching2.cpp
mikusp: Fixed

src/core/libraries/np/np_matching2.cpp
mikusp: Fixed

src/core/libraries/np/np_matching2.cpp
```diff
@@ -0,0 +1,699 @@
...
+static std::deque<NpMatching2RoomEvent> g_room_events;
+static std::mutex g_responses_mutex;
+static std::deque<std::function<void()>> g_responses;
+
+struct OrbisNpMatching2CreateContextParameter {
+    Libraries::Np::OrbisNpId npId;
+    void* npCommunicationId;
+    void* npCommunicationPassphrase;
+    Libraries::Np::OrbisNpServiceLabel serviceLabel;
+    u64 size;
```
StevenMiller123, line 54: Sword Art Online: Lost Song sets this struct size to 0x28 before calling.
Did you make a mistake while defining this struct?

georgemoralis: or maybe it is a new struct?

StevenMiller123: For what it's worth, the game just treats the npId field as a pointer to an OrbisNpId.

mikusp: I've looked into that and it cannot possibly be OrbisNpId, as that is 36 bytes long, and that would not even leave enough room for the size field? 🤔 I can only see it being OrbisNpOnlineId and 4 bytes of padding before npServiceLabel

StevenMiller123: And there's no way it could just be OrbisNpId*, which is what that game treats it as?

georgemoralis: Orbis NpMatching2CreateContextParamA uses userID and OrbisNpMatching2CreateContextParam uses OnlineId ? that doesn't feel proper

StevenMiller123: (this is from FUN_01652520)
image

image

mikusp: as OrbisNpId or a pointer to OrbisNpId?

georgemoralis:
is this good? if so needs a rebase

StevenMiller123 (approved):
There's plenty I could be picky about, but for an initial implementation of the library, I think this does a good enough job.

PR #3961, opened 2026-01-26
Implement V_LSHR_B64

mikusp:
Reported missing in https://github.com/shadps4-emu/shadPS4/issues/496#issuecomment-3596129445 

I don't have the game myself so cannot test


georgemoralis:
it looks similar to V_LSHL_B64 so i think it should be fine

PR #3972, opened 2026-01-29
Add VK_EXT_swapchain_maintenance1 support for NSM

par274:
The main idea is this:

The NSM (NVIDIA Smooth Motion) emulator can't fully control the frame rate, so it can't add extra "fake" frames.

Nvidia doesn't have any documentation on this, but according to my research, the VK_EXT_swapchain_maintenance1 ext's is needed.

For now, I've only added this. If we can do this, they can activate driver-based frame-generation. Remember, many games on PS4 can have problems when exceeding 60 FPS, but driver-based frame-generation would overcome these problems.


StevenMiller123 (changes requested):
I don't see the point in bloating logs with all these messages about Nvidia Smooth Motion, `add_extension` logs if an extension is present or not, and that should be plenty in my opinion.

par274:
> I don't see the point in bloating logs with all these messages about Nvidia Smooth Motion, `add_extension` logs if an extension is present or not, and that should be plenty in my opinion.

Right. I've only added them for testing purposes for now.

src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -335,6 +335,20 @@ bool Instance::CreateDevice() {
...
     supports_memory_budget = add_extension(VK_EXT_MEMORY_BUDGET_EXTENSION_NAME);
+    swapchain_maintenance1 = add_extension(VK_EXT_SWAPCHAIN_MAINTENANCE_1_EXTENSION_NAME);
+    vk::PhysicalDeviceSwapchainMaintenance1FeaturesEXT swapchain_maintenance1_features{};
+    if (swapchain_maintenance1) {
+        LOG_INFO(Render_Vulkan,
+                 "Enabling VK_EXT_swapchain_maintenance1 for Nvidia Smooth Motion compatibility");
+        swapchain_maintenance1_features.swapchainMaintenance1 = VK_TRUE;
+    }
+
+    // Enable additional extensions that may be required for NSM compatibility
```
kalaposfos13: This comment implies that you haven't actually tested this, and it also looks LLM-generated. Please do not submit untested code, especially if it didn't come from yourself.

par274: I'm generating the comments using AI, but I wrote the code myself. If I encounter any problems, I use AI to fix them. Is this a bad thing?

I referance this doc: https://download.nvidia.com/XFree86/Linux-x86_64/575.57.08/README/nvpresent.html

kalaposfos13: My main concern is code not being tested even once to verify that it at least works locally for your machine. If that is not the case, and you did test it and it works, then that's fine by me, but in that case, the comment should be changed.

par274: I understand the concern. The code is tested: I develop on Linux and can build it locally there.
My host machine is Windows, so I can't fully validate runtime behavior cross-platform, but CI builds and runs the Windows configuration on every update.

The comment may indeed sound speculative, true. I can reword it to explicitly state that this was tested locally and via CI. The implementation itself is based on the NVPresent / NSM documentation linked above.

But you can rest assured that I will edit these comments.

src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -335,6 +335,21 @@ bool Instance::CreateDevice() {
...
+    if (swapchain_maintenance1) {
+        swapchain_maintenance1_features.swapchainMaintenance1 = VK_TRUE;
+    }
+
+    // Enable NV_present_metering for additional Nvidia Smooth Motion support
+    const bool present_metering = add_extension(VK_NV_PRESENT_METERING_EXTENSION_NAME);
+    vk::PhysicalDevicePresentMeteringFeaturesNV present_metering_features{};
+    if (present_metering) {
+        present_metering_features.presentMetering = VK_TRUE;
+    }
```
squidbus, line 349: Both of these feature struct setups are not how things work here.

If you care about the actual feature boolean you need to add it to and get it from the `getFeatures2` chain.

But really you should be able to just set `swapchainMaintenance1` and `presentMetering` below to `true`, and usually what we do then is if the `add_extension` result was false for each extension we remove its structure from the `device_chain` (which you are also already doing). So you should just remove the `swapchain_maintenance1_features` and `present_metering_features` structs here.

src/video_core/renderer_vulkan/vk_swapchain.cpp
```diff
@@ -51,7 +51,32 @@ void Swapchain::Create(u32 width_, u32 height_) {
...
+            present_modes_info.presentModeCount = static_cast<u32>(present_modes.size());
+            present_modes_info.pPresentModes = present_modes.data();
+            use_present_modes_info = true;
+        } else {
+            LOG_WARNING(
+                Render_Vulkan,
+                "Failed to query present modes for swapchain maintenance1, NSM may not work "
+                "properly");
+        }
+    }
```
squidbus, line 70: I'd want to see some proof that this is actually doing anything, what you have here is telling Vulkan it may switch between any of the supported modes per-present, and then never actually setting any modes per-present, which sounds completely useless.

par274:
As previously requested, proof of the active present mode was needed. When the metering extension was removed, borderless fullscreen did not trigger. I have re-enabled metering and added a log to verify the swapchain mode and check for any fallback:

`[Render.Vulkan] <Info> (shadPS4:Main) vk_swapchain.cpp:246 FindPresentMode: Swapchain present mode: Immediate`

This confirms that Immediate mode is active and that enabling metering resolves the borderless fullscreen issue.

georgemoralis:
no

squidbus:
> As previously requested, proof of the active present mode was needed. When the metering extension was removed, borderless fullscreen did not trigger. I have re-enabled metering and added a log to verify the swapchain mode and check for any fallback:
> 
> `[Render.Vulkan] <Info> (shadPS4:Main) vk_swapchain.cpp:246 FindPresentMode: Swapchain present mode: Immediate`
> 
> This confirms that Immediate mode is active and that enabling metering resolves the borderless fullscreen issue.

...I'm not sure what any of this is supposed to mean. The present mode we pick is completely separate from the changes you're making here. This again just reads like AI without any understanding of the changes.

par274:
I’ve already tried to state that I don’t understand how metering fixes this problem. You can just focus on the problem and read the log output or download it from actions and try whether the problem is fixed. 

And I said it wasn’t an AI generated. Here is already the maintenance plugin etx for NSM. I also implemented it properly.

The metering ext is basically related to the low latency and yes I almost didn’t use this ext but simply even the add_extension command solves the borderless problem in a way that I don’t know why, maybe it’s indirect.

PR #3973, opened 2026-01-29
Add IME keyboard layout and panel metrics support

w1naenator:
- Introduced new header and implementation files for IME keyboard layout handling.
- Added structures for viewport metrics, keyboard grid layout, and drawing parameters.
- Implemented functions to compute viewport and panel metrics for the IME dialog.
- Enhanced the IME dialog UI to utilize the new keyboard layout and metrics.
- Updated input handling to support new keyboard interactions and layout adjustments.
- Added caret management and text normalization features in the IME dialog state.
- Improved window positioning logic to accommodate different screen resolutions.


georgemoralis:
which games uses ime keyboard?

w1naenator:
> which games uses ime keyboard?

bloodborne, code vein, nier automata, dragon balls xenoverse 1/2 (was softlocked), diablo III, god eater 2

DanielSvoboda:
In the game `Divinity: Original Sin 2 - Definitive Edition` this PR causes an error even before the game starts.
In the pre-release, it reaches the menu, but after pressing the X button, the animation loops and the menu doesn't open.

vmovdqa64 is an AVX-512 instruction, and my processor doesn't support AVX-512

image

StevenMiller123:
That sounds like it's executing garbage data, could you compile locally and debug the crash?

DanielSvoboda:
The local build works normally. I tested it both by debugging in Visual Studio and by running it through the command line.
However, the version downloaded from GitHub shows this error.
Even when testing only the SDL version from GitHub and the local SDL build, the local one works without any issues, while the GitHub version still presents this problem.
Honestly, I couldn’t figure out what is happening....

The game doesn't show any improvement with this PR. So you can disregard my comment.

georgemoralis:
is this still on work?

w1naenator:
This was implemented as a fix for Dragon Ball Xenoverse 1/2, where entering a character name would softlock the game. This issue is now resolved in this PR.

The virtual keyboard now matches the original size, uses the correct positioning, and scales according to the output video resolution.

I am not sure whether a virtual keyboard is needed on your side, so I do not know if it is worth spending additional time on it.

StevenMiller123:
And yes, I can confirm that this fixes the issue in Dragon Ball Xenoverse titles.
image

brad0demx:
Rocket League is another title which benefits from this PR
image

w1naenator:
@georgemoralis, +1

georgemoralis:
yah finish it and we gonna merge it :D

georgemoralis:
ime settings needs to be converted to new settings api

w1naenator:
> yah finish it and we gonna merge it :D

What exactly do you consider “finished” for this PR?

StevenMiller123:
Could you consider making the displayed keyboard function? Either that or hide it.
Just having that there, but non-functioning, would likely confuse users.

w1naenator:
> Could you consider making the displayed keyboard function? Either that or hide it. Just having that there, but non-functioning, would likely confuse users.

I can implement a working Latin layout as an initial step. However, fully supporting all PS4 keyboard layouts—especially Chinese and Japanese input methods—is significantly more complex, and I do not currently have a complete understanding of how those should behave.

For now, I see two practical options:

1) Implement a functional Latin layout first and extend support to additional layouts later.
2) Introduce a temporary compile-time flag (e.g., "ENABLE_FULL_LAYOUT") to switch between the old and new implementations without losing progress.

Let me know if this approach is acceptable.

StevenMiller123:
I think just having a latin layout first is fine.

w1naenator:
> I think just having a latin layout first is fine.

Sounds good, I like this approach. However, for now I will keep direct typing via the physical keyboard when the input field is focused for Arabic, Chinese, Japanese, Russian, Ukrainian, and other languages. This will serve as additional direct typing support until all layouts are fully implemented.

w1naenator:
CUSA03173_20260421_164424_844_hud_000005

georgemoralis:
ok that looks good . Give it some tests and we are ready to go :)

w1naenator:
I found several issues:

- In ImGui typing mode, focus is lost when pressing keyboard keys that are mapped to the virtual gamepad.
- The embedded font does not support certain characters.

StevenMiller123:
Two minor issues I'm seeing in CODE VEIN:
If I load the keyboard, it sometimes will automatically press a letter too.
image

If I click any keyboard button, controller interactions get wonky

https://github.com/user-attachments/assets/e23688c7-d79b-4268-a2e2-62192e82f1f9

w1naenator:
> Two minor issues I'm seeing in CODE VEIN: If I load the keyboard, it sometimes will automatically press a letter too. image
> 
> If I click any keyboard button, controller interactions get wonky
> 
>  Screen.Recording.2026-04-21.130601.mp4

Ok, I will fix tomorrow.

w1naenator:
Most of the issues should be fixed by this. The glitches between ImGui and OSK typing were caused by incorrect mode switching and handling.

brad0demx:
image
image

Seems to work great, there's a little bit of missing functionality compared to PS4, like if you press L1 + square it should delete the all the text in the field, and if you press and hold square it should keep deleting while the button is being held.

w1naenator:
> image image
> Seems to work great, there's a little bit of missing functionality compared to PS4, like if you press L1 + square it should delete the all the text in the field, and if you press and hold square it should keep deleting while the button is being held.

OK, please list any missing functionality so I can fix it. The select and copy-paste features will be added later, and additional language support will also be introduced in the future.

brad0demx:
Those two that I listed are all I can think of really, moving it around with right stick already works good

w1naenator:
Also, I won’t have access to my PS4 for a while, so it would be great if anyone could share layout screenshots. If you’d like a specific language to be supported, please include the required letter layout, both symbol layout pages, and the accents layout.
The currently implemented Latin layout was found in YouTube gameplay videos, so it may be inaccurate.

brad0demx:
image
image
image
image
image
image
Not sure if this is what you meant but hopefully it is, if theres anything else you know about you want me to screenshot just lmk

w1naenator:
> image image image image image image Not sure if this is what you meant but hopefully it is, if theres anything else you know about you want me to screenshot just lmk

Perfect, these are the Latin OSK layouts. Thank you!

w1naenator:
@georgemoralis, we should probably replace the embedded fonts with more comprehensive ones, as they don’t include all the required characters.

w1naenator:
@brad0demx, check now!

georgemoralis:
> @georgemoralis, we should probably replace the embedded fonts with more comprehensive ones, as they don’t include all the required characters.

i am not against that as long as there are public domain and compatible with GPL-2

w1naenator:
> > @georgemoralis, we should probably replace the embedded fonts with more comprehensive ones, as they don’t include all the required characters.
> 
> i am not against that as long as there are public domain and compatible with GPL-2

What about Noto Sans? I think it is a good candidate!

georgemoralis:
> > > @georgemoralis, we should probably replace the embedded fonts with more comprehensive ones, as they don’t include all the required characters.
> > 
> > 
> > i am not against that as long as there are public domain and compatible with GPL-2
> 
> What about Noto Sans? I think it is a good candidate!

You can safely use Noto Sans in your GPLv2 project.

Do:

    Ship the Noto font files in your application's directory.

    Include a copy of the OFL.txt license file in your distribution.

    Use the font to render text in your application.

Do Not:

    Modify the font files and claim the modified font is GPL-licensed (it must remain OFL-licensed).

    Attempt to relicense the font code to GPLv2.

georgemoralis:
> > > > @georgemoralis, we should probably replace the embedded fonts with more comprehensive ones, as they don’t include all the required characters.
> > > 
> > > 
> > > i am not against that as long as there are public domain and compatible with GPL-2
> > 
> > 
> > What about Noto Sans? I think it is a good candidate!
> 
> You can safely use Noto Sans in your GPLv2 project.
> 
> Do:
> 
> ```
> Ship the Noto font files in your application's directory.
> 
> Include a copy of the OFL.txt license file in your distribution.
> 
> Use the font to render text in your application.
> ```
> 
> Do Not:
> 
> ```
> Modify the font files and claim the modified font is GPL-licensed (it must remain OFL-licensed).
> 
> Attempt to relicense the font code to GPLv2.
> ```

seems ok

brad0demx:
Tried it again and now it's basically perfect great work 🚀
The L1 + Square combo works well, and holding square to delete one by one also works.

One other little thing that I forgot about is when navigating with the d-pad, if you press and hold it should keep scrolling in the direction while it's held, also holding triangle should keep entering spaces while its being held.

w1naenator:
> Tried it again and now it's basically perfect great work 🚀 The L1 + Square combo works well, and holding square to delete one by one also works.
> 
> One other little thing that I forgot about is when navigating with the d-pad, if you press and hold it should keep scrolling in the direction while it's held, also holding triangle should keep entering spaces while its being held.

Should be fixed now.

w1naenator:
@brad0demx  can you check:
- Are disabled buttons without labels still selectable, or does the selector skip over them?
- When using the left stick to move the selector, does it wrap around to the opposite side or not?

brad0demx:
Buttons with no labels are skipped over, and using the left stick the selector does wrap to the other side. 
For some reason when using the left stick, the first 2-3 cells are scrolled quite fast/sensitive then it slows down.
I notice a similar feeling with pressing L2 for shift/caps lock, it seems a little sensitive depending on how I actually pull the trigger. 

https://github.com/user-attachments/assets/596feeff-87bd-45bb-8a72-ac22520b951c

w1naenator:
> Buttons with no labels are skipped over, and using the left stick the selector does wrap to the other side. For some reason when using the left stick, the first 2-3 cells are scrolled quite fast/sensitive then it slows down. I notice a similar feeling with pressing L2 for shift/caps lock, it seems a little sensitive depending on how I actually pull the trigger.
> 
>  Screencast_20260423_171103.mp4

I don’t remember how it works on a real PS4.

brad0demx:
Got some videos of it's behaviour on PS4 and how it currently differs on emulator. 

When using D pad or left stick, scrolling all the way to the side stops when it reaches the final box, and only on an additional input will it wrap to the other side. (This is on me I should have mentioned this already, I assumed when you previously asked about this that you meant on emulator) 

When pressing and holding a direction, either with D pad or left stick, it will scroll 1 cell then after a very brief pause it will continue scrolling for the remainder of the button being held. This feature **does** work on emulator, but only when using D pad. Trying the same thing with left stick, it skips over the first cell.

And for the L2 shift, I think on console it behaves like this: You press L2 once to simulate shift, then you have to fully let go of the trigger before repressing in the time frame to simulate caps lock. 

Also on console, if you press it twice to simulate caps lock, then in a small time frame press L2 to turn it off and back on, it will go straight to caps lock. If you wait a second after turning it off though, it will go through shift > caps lock like normal. 

I think having to fully let go before repressing is the source of the inconsistency with the button on emulator. As you can see in the other video I sent above, some presses would instantly pop between shift/caps lock.

In these videos you can see the difference in the L2 and analogue stick sensitivity, as on console if I spin the stick in a circle it scrolls 2x2 cells, but on emulator it scrolls 3x3. 

https://github.com/user-attachments/assets/e344b379-8f92-4cf7-a6f8-de9a09d09bff


https://github.com/user-attachments/assets/0d21f698-48ab-4bec-8293-27163d4ac45c

w1naenator:
@brad0demx Can you check now?

georgemoralis:
@brad0demx  have you checked? . Are we ready for this?:D

src/common/config.cpp
```diff
@@ -173,6 +173,8 @@ static ConfigEntry<bool> isMotionControlsEnabled(true);
 static ConfigEntry<bool> useUnifiedInputConfig(true);
 static ConfigEntry<string> defaultControllerID("");
 static ConfigEntry<bool> backgroundControllerInput(false);
+static ConfigEntry<bool> imeAccessibilityEnabled(false);
+static ConfigEntry<bool> imeUrlMailShortPanel(false);
```
georgemoralis: this should be moved to emulator_settings.h , config is deprecated in 0.15.0+

w1naenator: ok!

w1naenator:
@georgemoralis, I think someone should check whether we might be violating Sony’s or others’ copyrights with this.

georgemoralis:
i guess if they UI isn't 100% matched we don't have an issue :d

w1naenator:
> i guess if they UI isn't 100% matched we don't have an issue :d

I made it look similar, but it’s still my own interpretation and code. However, there could still be patented elements, such as hotkeys and key combinations, table-style layouts, function button sequences, and many other features.

brad0demx:
My bad hadn't been on, I will test it now

brad0demx:
Tested again, and it now scrolls to the end and after a brief pause wraps while the button is being held, with no additional inputs required. There's still the issue with analogue stick navigation and L2 that I mentioned.

IMO I think the only really "glaring" issue is the L2 shift/capslock, because the jumps between shift and caps lock happen quite fast and inconsistently. The analogue stick issue isn't massively inconvenient because it's behaviour is consistent (and I can just use the d pad is also there which works basically perfectly from what I can tell)

brad0demx:
Also the other issue I had mentioned before like holding triangle to enter multiple spaces. However one I forgot about is holding x on a character to input it multiple times.

w1naenator:
@brad0demx, check now, I think everything fixed!

brad0demx:
Analogue stick feels a lot better now, I'm able to spin it in a circle and it scrolls 2x2 like on console. I notice when auto scrolling with the stick, it scrolls slower than when using the d-pad, though this a minor nitpick and doesn't really matter.

Holding X and Square to repeatedly enter/delete text works, holding triangle for repeatedly entering spaces works. 

L2 seems to be still inconsistent, and I also notice that when **holding L2 + triangle to access the punctuation variant of the keyboard, it flashes between both keyboard variants while triangle is being held.** 

It's kind of like the logic for holding triangle to repeatedly type spaces is also being applied to this button combo shortcut.

w1naenator:
> Analogue stick feels a lot better now, I'm able to spin it in a circle and it scrolls 2x2 like on console. I notice when auto scrolling with the stick, it scrolls slower than when using the d-pad, though this a minor nitpick and doesn't really matter.
> 
> Holding X and Square to repeatedly enter/delete text works, holding triangle for repeatedly entering spaces works.
> 
> L2 seems to be still inconsistent, and I also notice that when **holding L2 + triangle to access the punctuation variant of the keyboard, it flashes between both keyboard variants while triangle is being held.**
> 
> It's kind of like the logic for holding triangle to repeatedly type spaces is also being applied to this button combo shortcut.

OK, I have a gamepad modified for shooters. It has no range for L2/R2—just 0 or full. That is why I do not have this problem. Now the issue is clear; I will fix it.

w1naenator:
Huge refactor and optimization made, plus L2 fixes. Something might be broken now.

w1naenator:
@brad0demx, could you please check again?

brad0demx:
L2 is now working perfectly, and the issue I mentioned with L2 + triangle is also fixed. Basically everything that I have mentioned previously (Analogue stick, L2, button combos and button hold behaviours) is working now, except one bug in this new version. When pressing triangle to enter space, it doesn't type the space on the first button press and only starts typing it after the button is held briefly.

Also minor nitpick that doesn't really matter, when opening the keyboard it auto hovers over "1" at the top left when on console I believe it hovers over "G"

w1naenator:
> L2 is now working perfectly, and the issue I mentioned with L2 + triangle is also fixed. Basically everything that I have mentioned previously (Analogue stick, L2, button combos and button hold behaviours) is working now, except one bug in this new version. When pressing triangle to enter space, it doesn't type the space on the first button press and only starts typing it after the button is held briefly.
> 
> Also minor nitpick that doesn't really matter, when opening the keyboard it auto hovers over "1" at the top left when on console I believe it hovers over "G" 

Seams it is result of my optimizations, I'll fix!

w1naenator:
@brad0demx Is it working now as expected?

brad0demx:
Yes it's very close to being perfect, my only issue is currently the initial selection box is based on it's last location rather than starting on "G" when the keyboard is opened. Also when using L2 + triangle combo, it moves the selection box to the selection box of the button combo, whereas on the console the selection box stays in the same position.

Explanation might not have made much sense but I got videos of the behaviour on emu and console to potentially help 

https://github.com/user-attachments/assets/ee626543-28f9-47eb-8d8c-d571c77352e9

https://github.com/user-attachments/assets/c541826c-7e6a-4bc7-aafe-8e3f30686b4e

Really appreciate you taking the feedback and working on it so fast, glad I've been able to help 🥕
I'd say (at least for English language) its right there and could probably be merged

w1naenator:
Final fixes.
image

w1naenator:
@brad0demx Thank you, your involvement has been very helpful. I really appreciate your time, testing, and feedback - it helped a lot with debugging and improving the OSK.

w1naenator:
@georgemoralis @StevenMiller123 @DanielSvoboda please check last update.

georgemoralis:
looking good! waiting for others approval

georgemoralis (approved):

kalaposfos13:
There are still a couple things missing from the PR, however, they aren't merge blocking features. First, if you use the touchpad, the button selection automatically switches to a cursor that you can then use like a mouse based on-screen keyboard on PC. You can exit that mode by pressing any of the dpad arrows or R3 (see below). The other one is gyro based input, which is entered with R3, or the button that is already present for it but isn't hooked up yet (and exited either by initiating dpad movement or touchpad movement). While in that mode, further R3 presses recenter the cursor, and its movement is most intuitively understood as having a line coming out of the Dualshock's lightbar, and its (rotational 3dof) movement changes where that line intersects the screen, and that intersection is the pointer's location, assuming that at start, this line is pointed at the center of the keyboard.

PR #3978, opened 2026-01-30
Kernel.Vmm: Fix potential race condition involving concurrent Allocate and Free calls

StevenMiller123:
PAYDAY 2 occasionally crashes from a memory assert during dmem Free calls.
This PR makes PoolExpand and Allocate lock the extra "unmap" mutex I added in #3956, which will ensure they don't modify the dmem map after Free searches for areas to unmap.

Not fully sure this fixes the issue, as I can't consistently reproduce this issue, but this is something I should've fixed in that PR in the first place. I'm surprised Unity games didn't reveal this 😅 

Also snuck in a fix for sockets, HandleTable::GetSocket would crash if called with an invalid descriptor within the bounds of m_files. Came up while I was making my homebrew, but my socket PR got merged before I could push this fix there.

### DIFF
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -232,6 +232,9 @@ File* HandleTable::GetSocket(int d) {
         return nullptr;
     }
     auto file = m_files.at(d);
+    if (!file) {
+        return nullptr;
+    }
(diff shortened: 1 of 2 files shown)


StevenMiller123:
10 test runs, zero crashes.
I think this fixes the issue I had?

PR #3983, opened 2026-01-31
Kernel.Fs: Skip stdin for select reads

StevenMiller123:
UE uses select on stdin to signal if it should read from stdin. You can't write to stdin, so select never marks stdin as readable. (and attempting to read from stdin hangs the calling thread).


StevenMiller123:
This fixes tty error spam about failing to read from a dummy pipe in verbose UE games like Fortnite. I don't think this really affects anything behavior-wise though.

PR #3985, opened 2026-01-31
Fix thread names being set to garbage on Linux

jas0n098:
SetCurrentThreadName gets passed an std::string's c_str whose pointer gets invalidated by the assignment of g_curthread->name with a new std::string, resulting in broken thread names further down the line

### DIFF
--- a/src/common/thread.cpp
+++ b/src/common/thread.cpp
@@ -175,7 +175,7 @@ bool AccurateSleep(const std::chrono::nanoseconds duration, std::chrono::nanosec
 // Sets the debugger-visible name of the current thread.
 void SetCurrentThreadName(const char* name) {
     if (Libraries::Kernel::g_curthread) {
-        Libraries::Kernel::g_curthread->name = std::string{name};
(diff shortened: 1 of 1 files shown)


src/common/thread.cpp
```diff
@@ -175,7 +175,7 @@ bool AccurateSleep(const std::chrono::nanoseconds duration, std::chrono::nanosec
 // Sets the debugger-visible name of the current thread.
 void SetCurrentThreadName(const char* name) {
     if (Libraries::Kernel::g_curthread) {
-        Libraries::Kernel::g_curthread->name = std::string{name};
+        Libraries::Kernel::g_curthread->name = name;
```
Niram7777, line 178: if there is a concurrent access to g_curthread->name, we should rather protect where it is being used

kalaposfos13 (approved):

src/common/thread.cpp
kalaposfos13: The issue isn't concurrent access, you can't even do that, since each call to this only modifies their own thread's data.

PR #3987, opened 2026-02-02
Kernel.Pthread: Fix default pthread attributes

StevenMiller123:
Based on cross referencing libkernel decompilation with fpPS4's old code, our default pthread attributes were incorrect.

This fixes a variety of issues in Attack On Titan 2, mostly related to broken audio, as the game refuses to create an audio thread if the game's main thread priority is too low.


tochnonechelovek:
No changes in my titles.

Akeen74 (approved):

PR #3991, opened 2026-02-03
Prevent posix_pthread_mutexattr_settype from setting invalid mutex types

Spirtix:
Some games like NHL17 (CUSA04326) pass `0` as the mutex type to `posix_pthread_mutexattr_settype`, which resets its recursive mutexes and causes them to behave incorrectly. The real ps4 kernel rejects this value.

I tested this on NHL17 and managed to get past the initial loading screen without any more race conditions caused by broken mutexes. It now crashes later in some video code...


StevenMiller123 (approved):
Seems good to me, matches FreeBSD's documentation for behavior.
I haven't checked libkernel decomp though.

PR #3993, opened 2026-02-04
Libs: Miscellaneous fixes

StevenMiller123:
First change is a change to the return value of sceCompanionUtilGetEvent. Pretty sure I've rejected similar changes in the past, but with no incoming fixes, we might as well just make it a better stub for now.

Second change is a fix to libSceNpAuth's GetAuthorizationCode helper function. Mirror's Edge™ Catalyst throws a null issuer_id into the function if our PSN connectivity stub is enabled, and given the lack of null checks, I'm assuming this is considered valid for now.

Based on some debugging done by @brad0demx, these two changes should bring Mirror's Edge™ Catalyst to menus. I don't own the game, so someone else will need to showcase that working on here.

### DIFF
--- a/src/core/libraries/companion/companion_util.cpp
+++ b/src/core/libraries/companion/companion_util.cpp
@@ -29,10 +29,9 @@ u32 PS4_SYSV_ABI getEvent(sceCompanionUtilContext* ctx, sceCompanionUtilEvent* o
 }
 
 s32 PS4_SYSV_ABI sceCompanionUtilGetEvent(sceCompanionUtilEvent* outEvent) {
-    sceCompanionUtilContext* ctx = nullptr;
-    u32 ret = getEvent(ctx, outEvent, 1);
+    u32 ret = ORBIS_COMPANION_UTIL_NO_EVENT;
 
-    LOG_DEBUG(Lib_CompanionUtil, "(STUBBED) called ret: {}", ret);
+    LOG_DEBUG(Lib_CompanionUtil, "(STUBBED) called ret: {:#x}", ret);
     return ret;
 }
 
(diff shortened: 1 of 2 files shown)


StevenMiller123:
Requesting review from @georgemoralis since they rejected the last PR that changed the companion util stub

brad0demx:
image
With this PR Mirrors Edge Catalyst gets ingame, and the change to np_auth.cpp fixes the crash that you would encounter with the experimental networking options enabled. 

[CUSA01499 (1).zip](https://github.com/user-attachments/files/25061559/CUSA01499.1.zip)

brad0demx:
This PR also changes the behaviour of Battlefield 1, which now displays a loading screen briefly before crashing, rather than hanging infinitely like it does on main. 
image
[CUSA02387.log](https://github.com/user-attachments/files/25062454/CUSA02387.log)

brad0demx:
FIFA 18 is another title which changes behaviour on this PR. On main it hangs indefinitely, on here it crashes with:
```
[Debug] <Critical> (Job0) memory.cpp:1484 operator(): Assertion Failed!
Mapping cannot fit inside free region: size = 0x5c8000
```
[CUSA07994.log](https://github.com/user-attachments/files/25063345/CUSA07994.log)

georgemoralis:
> FIFA 18 is another title which changes behaviour on this PR. On main it hangs indefinitely, on here it crashes with:
> 
> ```
> [Debug] <Critical> (Job0) memory.cpp:1484 operator(): Assertion Failed!
> Mapping cannot fit inside free region: size = 0x5c8000
> ```
> 
> [CUSA07994.log](https://github.com/user-attachments/files/25063345/CUSA07994.log)

that is a different issue good to know though @StevenMiller123 new task :D

georgemoralis:
since noone seems to work on sceCompanionUtil i will accept this for now :D (also it fixes games so why not :D

PR #3994, opened 2026-02-05
filesystem: fill in timespec values for fstat

kalaposfos13:
Tearaway Unfolded has a check for whether the save files' last modification fall into a specific timespan, and if they aren't, then the save gets flagged as corrupt. By leaving these fields as zero, that check fails. By providing the correct values, this save "corruption" issue is resolved, although this doesn't help much for the fact that the game is still hanging at the end of loading.
On the Windows side, all three values (creation, modification and access) are spoofed with the last modification date, because that's the only one that std::filesystem has a cross-platform implementation for.

### DIFF
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -34,6 +34,7 @@
 #include <winsock2.h>
 #else
 #include <sys/select.h>
+#include <sys/stat.h>
 #endif
 
 namespace D = Core::Devices;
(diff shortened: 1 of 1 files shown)


StevenMiller123:
Could probably temporarily fix Windows by just zero-initializing the  struct?

kalaposfos13:
It looks like that is already done, and I slightly misunderstood the game check, and the reason why I thought it's due to memory garbage is because I put the breakpoint before the memset.

StevenMiller123 (approved):
Seems good to me, though I don't have the affected game.
I compiled and checked Windows, the code seems to work.

GHU7924:
I checked version 1.00 in the latest Main, and there was no message about corrupted saves (Windows 11). So the code works.

PR #3998, opened 2026-02-07
Kernel.Vmm: Fix bug with VMA physical area tracking

StevenMiller123:
This appears to fix memory asserts when loading ingame in Battlefield™ V (CUSA08724), which is only possible to test with nullGPU currently. This also fixes memory asserts when booting FIFA 18 (CUSA07994).


Xcedf:
Boots Battlefield 4 with NullGPU as well, can hear the sound of intro and menu.

PR #4000, opened 2026-02-07
Rewrote hotkeys in json format

georgemoralis:
As title says hotkeys are now in json format , with support for versioning , so we can add new ones without the need to erase the old settings , also more clear solution than the existing

```
{
    "hotkeys": {
        "fullscreen": "f11",
        "pause": "f9",
        "quit": "lctrl, lshift, end",
        "reload_inputs": "f8",
        "renderdoc_capture": "f12",
        "show_fps": "f10",
        "toggle_mouse_to_gyro": "f6",
        "toggle_mouse_to_joystick": "f7",
        "toggle_mouse_to_touchpad": "delete",
        "volume_down": "kpminus",
        "volume_up": "kpplus"
    },
    "version": 1
}
```

### DIFF
--- a/src/input/hotkeys_settings.cpp
+++ b/src/input/hotkeys_settings.cpp
@@ -1,0 +1,207 @@
+// SPDX-FileCopyrightText: Copyright 2026 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include "common/path_util.h"
+#include "hotkeys_settings.h"
+
+using json = nlohmann::json;
+
+std::shared_ptr<HotkeysSettingsManager> HotkeysSettingsManager::s_instance = nullptr;
+std::mutex HotkeysSettingsManager::s_mutex;
+
+// --------------------
+// ctor/dtor + singleton
+// --------------------
+HotkeysSettingsManager::HotkeysSettingsManager() {
+    // Load on construction
+    Load();
+}
+
+HotkeysSettingsManager::~HotkeysSettingsManager() {
+    Save();
+}
+
+std::shared_ptr<HotkeysSettingsManager> HotkeysSettingsManager::GetInstance() {
+    std::lock_guard<std::mutex> lock(s_mutex);
+    if (!s_instance) {
+        s_instance = std::make_shared<HotkeysSettingsManager>();
+    }
+    return s_instance;
+}
+
+void HotkeysSettingsManager::SetInstance(std::shared_ptr<HotkeysSettingsManager> instance) {
+    std::lock_guard<std::mutex> lock(s_mutex);
+    s_instance = instance;
+}
+
+// --------------------
+// Path helpers
+// --------------------
+std::filesystem::path HotkeysSettingsManager::GetDefaultConfigPath() const {
+    const std::filesystem::path userDir = Common::FS::GetUserPath(Common::FS::PathType::UserDir);
+    return userDir / "hotkeys.json";
+}
+
+std::filesystem::path HotkeysSettingsManager::GetConfigPath() const {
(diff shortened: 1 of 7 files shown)


src/input/input_handler.cpp
```diff
@@ -218,10 +222,117 @@ InputBinding GetBindingFromString(std::string& line) {
...
+                continue;
+            }
+
+            // Map to ControllerOutput
+            ControllerOutput* output = nullptr;
+
+            if (hotkeyName == "fullscreen") {
+                output = &*std::ranges::find(output_array, ControllerOutput(HOTKEY_FULLSCREEN));
+            } else if (hotkeyName == "pause") {
+                output = &*std::ranges::find(output_array, ControllerOutput(HOTKEY_PAUSE));
```
kalaposfos13, line 294: This could be a map instead.

src/input/input_handler.cpp
```diff
@@ -392,7 +503,14 @@ void ParseInputConfig(const std::string game_id = "") {
             return;
         }
 
-        // normal cases
+        // SKIP hotkey parsing from INI files - now handled by hotkeys.json
+        if (output_string.find("hotkey_") == 0) {
+            LOG_DEBUG(Input, "Hotkey '{}' is now managed by hotkeys.json, skipping INI parsing",
```
kalaposfos13, line 508: There is a much less hacky way to do this, which is removing all the hotkey_\* entries from the list of outputs.

src/input/hotkeys_settings.cpp
```diff
@@ -0,0 +1,207 @@
...
+    // Update to latest version structure
+    m_settings.UpdateToLatestVersion();
+
+    // Check for missing hotkeys from latest defaults
+    for (const auto& [key, defaultValue] : latestDefaults.bindings) {
+        if (m_settings.bindings.find(key) == m_settings.bindings.end()) {
+            // Add missing default hotkey
+            m_settings.bindings[key] = defaultValue;
+            needsSave = true;
+            std::cout << "[HotkeysSettings] Added missing default hotkey: " << key << " = "
```
kalaposfos13, line 73: You forgot to tell your LLM to use our logging for this entire file.

kalaposfos13 (changes requested):
What is the point of this PR? Is this just yet another step in your quest to JSONify every possible config thing in the emulator? In any case, I don't agree that this should be done this way. If you make all input configs be based on JSON, then fine, I can accept that, but this weird in-between state isn't something we should have in my opinion, plus this will introduce additional backwards incompatibility to both the emulator and the launcher(s), whilst with a full switch, there'd be at least a clean line between before and after. Also, this loses both having global non-hotkey mappings, and game-specific hotkey binding support.

src/input/hotkeys_settings.cpp
georgemoralis: vstudio autopilot writes them that way , the pr wasn't production ready anyway :P

georgemoralis:
> What is the point of this PR? Is this just yet another step in your quest to JSONify every possible config thing in the emulator? In any case, I don't agree that this should be done this way. If you make all input configs be based on JSON, then fine, I can accept that, but this weird in-between state isn't something we should have in my opinion, plus this will introduce additional backwards incompatibility to both the emulator and the launcher(s), whilst with a full switch, there'd be at least a clean line between before and after. Also, this loses both having global non-hotkey mappings, and game-specific hotkey binding support.

why we need game-specific hotkey support? All hotkeys are doing specific things i don't see a point of having game specific ones

georgemoralis:
also the though about this started with the fact that although you added new hotkeys for the volume that aren't available without erasing the existing global.ini and make emu write a new one . The format i purpose at least supports versioning so it will fill the keys without an issue

georgemoralis:
making @kalaposfos13 happy

PR #4001, opened 2026-02-07
Kernel.Process: Implement sceKernelGetModuleInfo2, sceKernelGetModuleList2

StevenMiller123:
Should hopefully help with FIFA 18 on Windows? Probably some other titles too.

As far as I could tell, these two functions hide system libraries from the outputted information. This is mostly a guess based on fpPS4 code and kernel decompiling, as I've yet to find a proper way to test these on real hardware.


georgemoralis:
seems to fix fifa18

PR #4011, opened 2026-02-08
Added sceAudioOutGetSystemState and reduced some loading

georgemoralis:
Helps with black ops 4

### DIFF
--- a/src/core/libraries/audio/audioout.cpp
+++ b/src/core/libraries/audio/audioout.cpp
@@ -496,11 +496,11 @@ s32 PS4_SYSV_ABI sceAudioOutGetPortState(s32 handle, OrbisAudioOutPortState* sta
 
     state->rerouteCounter = 0;
     state->flag = 0;
-    LOG_INFO(Lib_AudioOut,
-             "called, handle={:#x}, state={}, output={}, channel={}, volume={}, "
-             "rerouteCounter={}, flag={}",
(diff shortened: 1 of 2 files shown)


src/core/libraries/audio/audioout.cpp
```diff
@@ -769,6 +769,19 @@ s32 PS4_SYSV_ABI sceAudioOutSetMixLevelPadSpk(s32 handle, s32 mixLevel) {
...
+s32 PS4_SYSV_ABI sceAudioOutGetSystemState(OrbisAudioOutSystemState* state) {
+    if (lazy_init.load(std::memory_order_relaxed) == 0 || audio == nullptr) {
+        LOG_ERROR(Lib_AudioOut, "audio is not init");
+        return ORBIS_AUDIO_OUT_ERROR_NOT_INIT;
+    }
+    if (state == nullptr) {
+        return ORBIS_AUDIO_OUT_ERROR_INVALID_POINTER;
+    }
+    memset(state, 0, sizeof(*state));
+    LOG_DEBUG(Lib_AudioOut, "called");
```
StevenMiller123, line 781: My only complaint is that I think this should have a higher log level, since it seems like a stub.
Otherwise this all looks good to me.

georgemoralis: this struct is only has loudness that is used in playstation sharing , i don't think we will ever implement that feature :p

PR #4012, opened 2026-02-08
Add option to enable/disable sys_modules

### LINKED ISSUE #614: [Feature Request] Allow to enable/disable loaded sys_modules
### Quick summary
Add ability load only certain sys_modules

### Details
Similarly as RPCS3 handles it, as it might be quite useful with e.g Bloodborne (hate to say it nowadays) where removing some modules allowed the game to run better and/or avoid some issues with loading.
Would also help in detection if sys_modules were dumped correctly and none of it is broken?
Due to specifics of done things, could be in debug tab?

CleanShot 2024-08-27 at 13 03 48@2x

DanielSvoboda:
Resolves https://github.com/shadps4-emu/shadPS4/issues/614

These options are not added to the global settings, only to the per-game settings.
These options will be available in the 'Experimental' menu, and can only be activated through the 'game specific settings'.

image

If someone disables something, it will appear in the log:
image

The QtLauncher section is here: https://github.com/shadps4-emu/shadps4-qtlauncher/pull/244


src/common/config.cpp
```diff
@@ -203,6 +203,12 @@ static ConfigEntry<bool> isShaderDebug(false);
 static ConfigEntry<bool> isSeparateLogFilesEnabled(false);
 static ConfigEntry<bool> showFpsCounter(false);
 static ConfigEntry<bool> logEnabled(true);
+static const std::vector<std::string> sysModules = {
```
StevenMiller123: I'd prefer if we could have one constant list that's referenced by both emulator initialization and config code.
Introducing duplicated constants makes it easier to introduce bugs when we add or remove from the supported list

DanielSvoboda: That makes sense, I was thinking about that when I was sending the commit. I copied this from the QT version because that list isn't there, so I had to create it here, but the names already exist in the SDL version.

georgemoralis:
my feelings is mixed about this , we have to update the list each time we add or remove a new module. Also we have LLE and HLE versions is some modules. Atm we force the users to test with all these , if we make it configurable that will fix things in testing and i dunno if it makes something work better

GHU7924:
>my feelings is mixed about this , we have to update the list each time we add or remove a new module. Also we have LLE and HLE versions is some modules. Atm we force the users to test with all these , if we make it configurable that will fix things in testing and i dunno if it makes something work better

Overall, I don't see anything wrong with this. Yes, the list will have to be adjusted, so it will be necessary to open PR specifically for this (possibly right before the release of a new version). But on the other hand, if we take regular users, they either don't know about modules initially and report that their game doesn't work, or if they do know about modules, I'm not sure how often they check the module list for updates to adjust it, especially considering that their games can launch and be playable.

P.S. I haven't noticed if this already exists in the emulator, but perhaps we should make the log indicate which version of the module is being used (LLE or HLE), which would improve understanding during testing.

DanielSvoboda:
Disabled system modules now bypass LLE loading but still attempt HLE fallback
when an HLE implementation is available.
image

DanielSvoboda:
> my feelings is mixed about this , we have to update the list each time we add or remove a new module. Also we have LLE and HLE versions is some modules. Atm we force the users to test with all these , if we make it configurable that will fix things in testing and i dunno if it makes something work better

When a new module is added for LLE loading in SDL, it must also be added to the QtLauncher list so it can be toggled. Otherwise, it will simply load as enabled by default.  
Removing it from SDL prevents it from being loaded, regardless of its presence in the Qt list.

GHU7924:
>Disabled system modules now bypass LLE loading but still attempt HLE fallback
when an HLE implementation is available.

I may be wrong now, but what if the user wants to disable the module, he removes the LLE, but the HLE module is enabled, what then?

@georgemoralis The Wiki on the site is not updated and contains an outdated list of modules.
Another thing, I don't want to open an issue about this, but I have a question about official images for the emulator. If I add the emulator to SteamDeck as a third-party game, the app won't have a background, logo, or anything else. Customization is now available on Steam, and all you need is just a few images of a certain size. There are no official images. Can anyone help with this? I'm not sure what the best way to store them is: downloading them separately or including them in the settings? For example, a dedicated folder with images in the User folder.

DanielSvoboda:
@GHU7924 I'm not sure if this is exactly what you want, but if you right-click on the game, you have the option to open the game folder, and in that path you'll find images for the icon and the banner.

ICON:       \sce_sys\icon0.png
BANNER:  \sce_sys\pic0.png

GHU7924:
@DanielSvoboda No, I'm talking about the QtLauncher application itself. To make it clearer, I'll use a game as an example. You add a game to Steam that's not part of your library, then click the properties of the added application, select Customization, and it offers the option to add images of specific sizes so your application has its own background, logo, and so on. I would like the emulator to have its own images that can be used in the game mode on SteamDeck.

PR #4013, opened 2026-02-09
Fix deadlock from missed unlock call after #3946

kalaposfos13:
This fixes a regression in Shadow of the Colossus.
It also changes a heuristic that makes said game significantly more stable.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -349,7 +349,8 @@ s32 MemoryManager::Free(PAddr phys_addr, u64 size, bool is_checked) {
 }
 
 s32 MemoryManager::PoolCommit(VAddr virtual_addr, u64 size, MemoryProt prot, s32 mtype) {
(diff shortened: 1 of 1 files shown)


src/core/memory.cpp
```diff
@@ -650,6 +650,8 @@ s32 MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, u64 size, Memo
             // TRACK_ALLOC(mapped_addr, size, "VMEM");
         }
 
+        lk2.unlock();
```
StevenMiller123, line 655: This seems less like a missed unlock, more like a deadlock in Rasterizer::MapMemory?
Could you provide a stack trace?

StevenMiller123 (changes requested):
PoolCommit will need a similar change, otherwise I think this is fine for now.

StevenMiller123 (approved):

PR #4022, opened 2026-02-11
More accurate motion control events emulation

kalaposfos13:
This fixes input duplication in GRR, at least on my end.


notanonion733:
Someone please explain to me in stupid people terms because I have no idea how to fix this

PR #4024, opened 2026-02-11
Libraries: Rudp function

thethunderturner:
implementation for GetStatus; gets called by Death Stranding


src/core/libraries/np/np_manager.cpp
```diff
@@ -700,22 +682,11 @@ struct NpStateCallbackForNpToolkit {
...
-
 s32 PS4_SYSV_ABI sceNpCheckCallback() {
-    LOG_DEBUG(Lib_NpManager, "(STUBBED) called");
-
-    std::scoped_lock lk{g_np_callbacks_mutex};
-
-    for (auto i : g_np_callbacks) {
-        (i.second)();
+    if (!g_signed_in) {
+        return ORBIS_NP_ERROR_NOT_INITIALIZED;
```
kalaposfos13, line 712: This should be ORBIS_NP_ERROR_SIGNED_OUT, no?

thethunderturner: UHm, I thought i reverted this changed as its technically not part of Rudp but whatever.
I will check again and let you know

thethunderturner: No, its correct as far as i can see

StevenMiller123: This does seem wrong though. How are you checking for this?

StevenMiller123: I can't check decomp at the moment, so I could be wrong

StevenMiller123:
I'll test this with some of my titles when I have a chance, should be around 4 hours from now.

src/core/libraries/rudp/rudp.cpp
```diff
@@ -43,62 +68,102 @@ s32 PS4_SYSV_ABI sceRudpEnd() {
...
+u32 Func_0100d870(uintptr_t globalContextListPtr) {
+    if (globalContextListPtr == 0) {
+        return 0;
+    }
+
+    u32* countPtr = reinterpret_cast<u32*>(globalContextListPtr + 8);
+    return *countPtr;
+}
+
+void Func_0101a5e0(s32* sentQualityLevel4Packets, s32* rcvdQualityLevel4Packets, s32* allocs,
```
StevenMiller123: If possible, it would be nice to have a proper name for this internal function.

src/core/libraries/rudp/rudp.cpp
```diff
@@ -43,62 +68,102 @@ s32 PS4_SYSV_ABI sceRudpEnd() {
...
 
-s32 PS4_SYSV_ABI sceRudpGetStatus() {
-    LOG_ERROR(Lib_Rudp, "(STUBBED) called");
-    return ORBIS_OK;
+u32 Func_0100d870(uintptr_t globalContextListPtr) {
+    if (globalContextListPtr == 0) {
+        return 0;
+    }
+
+    u32* countPtr = reinterpret_cast<u32*>(globalContextListPtr + 8);
```
StevenMiller123, line 124: This function appears to be unused in your code, is there a reason you've added it?

src/core/libraries/rudp/rudp.cpp
```diff
@@ -43,62 +68,102 @@ s32 PS4_SYSV_ABI sceRudpEnd() {
...
+
+s32 PS4_SYSV_ABI sceRudpGetStatus(OrbisRudpStatus* status, size_t statusSize) {
+    std::lock_guard lock(g_RudpMutex);
+
+    if (!g_isRudpInitialized) {
+        return ORBIS_RUDP_ERROR_NOT_INITIALIZED;
+    }
+
+    int result = ORBIS_RUDP_ERROR_INVALID_ARGUMENT;
+    if (status != (OrbisRudpStatus*)nullptr && statusSize - 1 < 0xf8) {
```
StevenMiller123: No point in casting a nullptr, you're not going to get a compiler warning for just leaving it as-is.

StevenMiller123 (changes requested):
Some nitpicks

src/core/libraries/rudp/rudp.cpp
thethunderturner: Ok, my IDE automatically suggested it, so I went with it. I will roll it back

src/core/libraries/rudp/rudp.cpp
thethunderturner: the function should've been called by `status->currentContexts = ...` ut I forgot to change it

src/core/libraries/rudp/rudp.cpp
```diff
@@ -43,62 +68,102 @@ s32 PS4_SYSV_ABI sceRudpEnd() {
...
+s32 PS4_SYSV_ABI sceRudpGetSizeWritable(int context_id) {
     LOG_ERROR(Lib_Rudp, "(STUBBED) called");
     return ORBIS_OK;
 }
 
-s32 PS4_SYSV_ABI sceRudpGetStatus() {
-    LOG_ERROR(Lib_Rudp, "(STUBBED) called");
-    return ORBIS_OK;
+u32 RudpGetContextCount(uintptr_t globalContextListPtr) {
+    if (globalContextListPtr == 0) {
```
StevenMiller123, line 120: This is always going to return 0, since your context is null.
I suggest looking into what the context struct would look like, so this function doesn't involve arbitrary nasty pointer offsets.

src/core/libraries/rudp/rudp.cpp
```diff
@@ -43,62 +68,102 @@ s32 PS4_SYSV_ABI sceRudpEnd() {
...
+
+    if (!g_isRudpInitialized) {
+        return ORBIS_RUDP_ERROR_NOT_INITIALIZED;
+    }
+
+    int result = ORBIS_RUDP_ERROR_INVALID_ARGUMENT;
+    if ((status != (OrbisRudpStatus *)0x0) && (statusSize - 1 < 0xf8)) {
+        std::memcpy(status, &g_rudpStatusInternal, statusSize);
+
+        status->currentContexts = static_cast<s32>(RudpGetContextCount(reinterpret_cast<uintptr_t>(g_RudpContext)));
```
StevenMiller123, line 151: I'm not particularly fond of needing to typecast the rudp context to fit into a function that's designed for it.
Again, probably just needs a struct added.

src/core/libraries/rudp/rudp.cpp
```diff
@@ -43,62 +68,102 @@ s32 PS4_SYSV_ABI sceRudpEnd() {
...
+    if (allocs)
+        *allocs = g_state.allocs.load();
+    if (frees)
+        *frees = g_state.frees.load();
+}
+
+s32 PS4_SYSV_ABI sceRudpGetStatus(OrbisRudpStatus* status, size_t statusSize) {
+    std::lock_guard lock(g_RudpMutex);
+
+    if (!g_isRudpInitialized) {
```
StevenMiller123, line 143: With how barebones this implementation is, I think we should keep an error log for this function.

StevenMiller123 (changes requested):
This probably is going to break more than it helps, since you've yet to implement sceRudpInit in any form.
I suggest at least stubbing that function so sceRudpGetStatus isn't returning errors it shouldn't hit.

src/core/libraries/rudp/rudp.cpp
StevenMiller123: My point wasn't to change nullptr to 0, it was to stop casting nullptr to OrbisRudpStatus* 😅 
Modern compilers will implicitly cast the nullptr, which is easier to look at in my opinion (and matches most nullptr checks we do)

src/core/libraries/np/np_manager.cpp
StevenMiller123: Can confirm, this is incorrect. The check performed by libSceNpManager is effectively checking if the library was initialized, not if you're signed in to PSN.

Image

Image

georgemoralis:
nice try but nothing that can be merged in the near future , reopen it if you have some more progress on it

PR #4027, opened 2026-02-12
Lib.Pad: Check handle in scePadRead, scePadReadState

StevenMiller123:
Real library checks against the (up to 16) opened handles it can store. We still don't have proper handle logic though, so just reject handle values below 1.
This fixes input issues on the title screen of A KING'S TALE: FINAL FANTASY XV


StevenMiller123:
Also fixes UE input duplication.
image

PR #4033, opened 2026-02-14
GR2-win-crash-fix

w1naenator:
Added a check for fiber stacks before clearing the stack in ExecuteGuest. That fixes Gravity Rush 2 crash on Windows.

### DIFF
--- a/src/core/tls.h
+++ b/src/core/tls.h
@@ -62,7 +62,7 @@ ReturnType ExecuteGuest(PS4_SYSV_ABI ReturnType (*func)(FuncArgs...), CallArgs&&
     EnsureThreadInitialized();
     // clear stack to avoid trash from EnsureThreadInitialized
     auto* tcb = GetTcbBase();
-    if (tcb == nullptr && tcb->tcb_fiber == nullptr) {
+    if (tcb != nullptr && tcb->tcb_fiber == nullptr) {
         ClearStack<12_KB>();
     }
     return func(std::forward<CallArgs>(args)...);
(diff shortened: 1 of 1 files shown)


StevenMiller123:
This seems to break a lot of games?

src/core/tls.h
```diff
@@ -59,9 +59,11 @@ void ClearStack() {
 
 template <class ReturnType, class... FuncArgs, class... CallArgs>
 ReturnType ExecuteGuest(PS4_SYSV_ABI ReturnType (*func)(FuncArgs...), CallArgs&&... args) {
-    EnsureThreadInitialized();
```
StevenMiller123, line 62: I don't think you intended to remove the EnsureThreadInitialized call?
Removing this breaks most PS4 games.

StevenMiller123:
Adding the check for tcb breaks Catherine: Full Body, which is also the game that necessitated stack clearing in the first place.

src/core/tls.h
w1naenator: You are right, i made changes ftom phone, it is accedental remove, ill fix it.

w1naenator:
> Adding the check for tcb breaks Catherine: Full Body, which is also the game that necessitated stack clearing in the first place.

You are right!

w1naenator:
@StevenMiller123, can you check now?

StevenMiller123:
Catherine: Full Body is still broken with your latest push.
[CUSA15036.log](https://github.com/user-attachments/files/25316834/CUSA15036.log)

w1naenator:
> Catherine: Full Body is still broken with your latest push. [CUSA15036.log](https://github.com/user-attachments/files/25316834/CUSA15036.log)

🤦‍♂️ Now it should work!

StevenMiller123:
Catherine: Full Body is working properly now.
image

PR #4038, opened 2026-02-15
Lib.GnmDriver: Fix flip arg for sceGnmSubmitAndFlipCommandBuffers

StevenMiller123:
Based on some wacky crap Unity does, these functions appear to take a 64-bit value for the flip arg.
Fixing this, along with a bug in sceVideoOutGetEventData, fixes the handful of Unity games crashing on an exception in their GfxFlipThread after a sceKernelAllocateDirectMemory call fails.

This fixes crashing in:
- Bloons TD 6
- Yu-Gi-Oh! Master Duel

image

image

Based on what I know, this should also fix issues in:
- FINAL FANTASY
- FINAL FANTASY II
- FINAL FANTASY III
- FINAL FANTASY IV
- FINAL FANTASY V
- FINAL FANTASY VI

### DIFF
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -2052,7 +2052,7 @@ int PS4_SYSV_ABI sceGnmSqttWaitForEvent() {
 }
 
 static inline s32 PatchFlipRequest(u32* cmdbuf, u32 size, u32 vo_handle, u32 buf_idx, u32 flip_mode,
-                                   u32 flip_arg, void* unk) {
+                                   s64 flip_arg, void* unk) {
     // check for `prepareFlip` packet
     cmdbuf += size - 64;
     ASSERT_MSG(cmdbuf[0] == 0xc03e1000, "Can't find `prepareFlip` packet");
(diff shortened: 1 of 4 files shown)


StevenMiller123:
@PirkyP When you have a chance, do you mind testing the FINAL FANTASY Unity titles?

PirkyP:
I tested Final Fantasy 1 and 6 and they seem more stable now. In FF1, I can navigate to all of the different areas in the menu without any crashing, however once I try to go in game it still crashes. As for FF6, it still crashes before the menu, but it progressed further in the boot sequence than I've seen it before.

[FF 1_shad_log.txt](https://github.com/user-attachments/files/25320071/FF.1_shad_log.txt)
[FF 6_shad_log.txt](https://github.com/user-attachments/files/25320116/FF.6_shad_log.txt)

image

EmulationPsycho7:
Spelunky 2 still has the same problem.
Captura de Tela (379)

StevenMiller123:
@EmulationPsycho7 I doubt Spelunky 2 was affected by the bug I fixed. It isn't a Unity game, and the issue you're on seems to be a network check.

PR #4039, opened 2026-02-15
Lib.VideoOut: Fix pending flips limit

StevenMiller123:
Real hardware has a fixed limit for pending flips, it doesn't rely on the number of registered VideoOut buffers. While the kernel uses an array of 18 elements, my tests suggest the cap is around 16 pending flips.

I'm not entirely sure if our code is properly equipped to handle this change, but it appears to work fine in games that were previously asserting or spamming the flip queue is full error message.


StevenMiller123:
This change brings FINAL FANTASY VII REMAKE ingame with broken graphics.

image
image
image
image
image

StevenMiller123:
Also brings base versions of Media Player to their menu.
image

PR #4048, opened 2026-02-18
threads: initialize TLS on thread creation

roamic:
I'm not sure why this was like this before, but I assume the code in `ExecuteGuest` just initialized the TLS once per thread. I think this can be done on thread start instead of delaying it to the first guest call. This removes the need for stack cleanup hack.

I've tested this with Catherine and GR2

Let me know if this breaks something or has a potential to.

### DIFF
--- a/src/core/libraries/avplayer/avplayer_impl.cpp
+++ b/src/core/libraries/avplayer/avplayer_impl.cpp
@@ -12,28 +12,28 @@ void* PS4_SYSV_ABI AvPlayer::Allocate(void* handle, u32 alignment, u32 size) {
     const auto* const self = reinterpret_cast<AvPlayer*>(handle);
     const auto allocate = self->m_init_data_original.memory_replacement.allocate;
     const auto ptr = self->m_init_data_original.memory_replacement.object_ptr;
-    return Core::ExecuteGuest(allocate, ptr, alignment, size);
+    return allocate(ptr, alignment, size);
 }
 
 void PS4_SYSV_ABI AvPlayer::Deallocate(void* handle, void* memory) {
     const auto* const self = reinterpret_cast<AvPlayer*>(handle);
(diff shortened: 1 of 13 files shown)


squidbus:
I think it was in case we need to call guest code from a thread not started by the guest... but I don't remember exactly and I'm not sure if we do.

roamic:
> I think it was in case we need to call guest code from a thread not started by the guest... but I don't remember exactly and I'm not sure if we do.

Thank you for the explanation. I've checked all of the usages of `ExecuteGueast` and could not find a thread where `EnsureThreadInitialized` is not called. In fact, host threads that call guest code are only present in avplayer and it uses `pthread_create` from our HLE which in turn initializes the TLS.

kalaposfos13:
From what I can see, there might be one instance of a thread calling game functions while not initialising the thread properly, and looks like it wasn't doing that even before this PR?
https://github.com/shadps4-emu/shadPS4/blob/a99c8147392c806b2b7ffdf876720ba170f17c23/src/core/libraries/usbd/emulated/dimensions.cpp#L624

squidbus (approved):

roamic:
> From what I can see, there might be one instance of a thread calling game functions while not initialising the thread properly, and looks like it wasn't doing that even before this PR?
> 
> https://github.com/shadps4-emu/shadPS4/blob/a99c8147392c806b2b7ffdf876720ba170f17c23/src/core/libraries/usbd/emulated/dimensions.cpp#L624

Thanks, I've added initialization function there just to be safe.

StevenMiller123:
Fixes the TOKYO GHOUL:re CALL to EXIST regression from #2877 
image

StevenMiller123:
Breaks Catherine Full Body for me, tested on Linux.
[CUSA15036.log](https://github.com/user-attachments/files/25405461/CUSA15036.log)

PR #4054, opened 2026-02-19
Timer touchups

StevenMiller123:
This PR fixes several inaccuracies with our event queues implementation.
- Removed fflags incrementing
- Removed small timer logic
- Fixed event data for timer events
- Fixed event duplication
- Fixed data for VideoOut vblank events

These changes fix some random hanging and debugger crashes seen in titles abusing high-res timers.

Regarding my removal of small timer-specific logic: my hardware tests (posted at https://github.com/ps4emulation/integration_tests/pull/16) suggest that small timers are around the same accuracy as normal timers, somewhere between 10 and 20ms is as low as you can go before real hardware's timing inaccuracies become apparent. Even our timer event logic is fast enough to be more accurate than real hardware.

I'm opening this as a draft, I'd like to see some tests before this gets merged, primarily to make sure removing the special high accuracy small timer logic doesn't break titles using those.

### DIFF
--- a/src/core/libraries/kernel/equeue.cpp
+++ b/src/core/libraries/kernel/equeue.cpp
@@ -268,9 +268,6 @@ s32 PS4_SYSV_ABI sceKernelAddHRTimerEvent(SceKernelEqueue eq, int id, timespec* 
         return ORBIS_KERNEL_ERROR_EBADF;
     }
 
-    if (ts->tv_sec > 100 || ts->tv_nsec < 100'000) {
-        return ORBIS_KERNEL_ERROR_EINVAL;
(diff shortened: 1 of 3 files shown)


StevenMiller123:
Didn't fix what I was aiming to fix, closing.

StevenMiller123:
EA's actual equeue hang issue turned out to be pretty simple (assuming there aren't more issues than what I've already found).
Reopening for tests.

StevenMiller123:
Did some more tests, this approach to dealing with the small timer issues just wont cut it. Real hardware is too precise for this to be a viable solution.

PR #4055, opened 2026-02-20
Simulate sceKernelInternalMemory mapping during libkernel event

par274:
Refs:
* https://github.com/shadps4-compatibility/shadps4-game-compatibility/issues/65
* https://github.com/shadps4-compatibility/shadps4-game-compatibility/issues/14

This fixes flexible memory size some titles (e.g. TLOU2).

### DIFF
--- a/src/core/linker.cpp
+++ b/src/core/linker.cpp
@@ -104,6 +104,15 @@ void Linker::Execute(const std::vector<std::string>& args) {
 
     memory->SetupMemoryRegions(fmem_size, use_extended_mem1, use_extended_mem2);
 
+    // Simulate sceKernelInternalMemory mapping, a mapping usually performed during libkernel init.
+    static constexpr VAddr KernelAllocBase = 0x880000000ULL;
+    static constexpr s64 InternalMemorySize = 0x1000000;
+    void* addr_out{reinterpret_cast<void*>(KernelAllocBase)};
+
+    const s32 ret = Libraries::Kernel::sceKernelMapNamedFlexibleMemory(
+        &addr_out, InternalMemorySize, 3, 0, "SceKernelInternalMemory");
+    ASSERT_MSG(ret == 0, "Unable to perform sceKernelInternalMemory mapping");
+
     main_thread.Run([this, module, &args](std::stop_token) {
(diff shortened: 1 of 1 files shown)


StevenMiller123:
This is not accurate behavior, which is why I reverted my old PR that did this.
The sceKernelInternalMemory mapping is a system flexible mapping, which doesn't consume flexible memory.

par274:
> This is not accurate behavior, which is why I reverted my old PR that did this. The sceKernelInternalMemory mapping is a system flexible mapping, which doesn't consume flexible memory.

Sorry, I didn't see that. I'm trying everything to get rid of the black screen in TLOU2. 

I've resolved the 3dAudio problem(not exactly), but I haven't been able to overcome the problem with Vulkan yet.

`[Render.Vulkan] <Error> (shadPS4:GpuCommandProcessor) vk_platform.cpp:57 DebugUtilsCallback: VUID-vkCmdDispatchIndirect-None-06479: vkCmdDispatchIndirect(): the sampled image descriptor [VkDescriptorSet 0x82b000000082b, Set 0, Binding 18, Index 0, variable "cs_img0"] has VkImageView 0x4390000000439[ImageView 1x1x1 0x0:0x100 0:0 0:0 (Color)] with format of VK_FORMAT_R8G8B8A8_UNORM which doesn't support VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_DEPTH_COMPARISON_BIT.`

StevenMiller123:
TLOU2 isn't going to go anywhere without an implementation for bindless sharp accesses, same deal as Uncharted 4 in that regard.
As for the memory issue, the actual discrepancy is that game modules are supposed to count towards flexible memory usage, I believe that's the only thing we're really missing that would affect a title so early in the startup process.

par274:
> TLOU2 isn't going to go anywhere without an implementation for bindless sharp accesses, same deal as Uncharted 4 in that regard. As for the memory issue, the actual discrepancy is that game modules are supposed to count towards flexible memory usage, I believe that's the only thing we're really missing that would affect a title so early in the startup process.

Thanks, this was very helpful for me.

I’ll look into the flexible memory accounting and try to understand how module mappings should affect it.

PR #4058, opened 2026-02-20
Calculate flex memory size

par274:
Based on: https://github.com/shadps4-emu/shadPS4/pull/4055#issuecomment-3930957339

I'm calculating flexible memory accounting based on VMM; module code is counted towards the flexible quota; host-side trampoline region is not counted.

This can solve some startup problems in games like Uncharted 4 and TLOU2.

I'm still learning process, ttrampoline will probably cause problems in other games.

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -1426,7 +1426,10 @@ u64 MemoryManager::GetFlexibleMappedBytesInRangeLocked(VAddr virtual_addr, u64 s
     }
 
     u64 mapped_bytes = 0;
-    auto it = std::prev(vma_map.upper_bound(range_start));
+    auto it = vma_map.upper_bound(range_start);
+    if (it != vma_map.begin()) {
+        it = std::prev(it);
+    }
     while (it != vma_map.end() && it->second.base < range_end) {
         const auto& vma = it->second;
         const VAddr vma_end = vma.base + vma.size;
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -1394,7 +1394,8 @@ bool MemoryManager::IsFlexibleCommittedVma(const VirtualMemoryArea& vma) const {
     const bool has_physical_tracking =
         vma.type == VMAType::Direct || vma.type == VMAType::Flexible || vma.type == VMAType::Pooled;
     if (has_physical_tracking) {
-        // Direct/flexible/pooled mappings should expose at least one physical sub-area when committed.
+        // Direct/flexible/pooled mappings should expose at least one physical sub-area when
+        // committed.
         return !vma.phys_areas.empty();
     }
 
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -413,6 +413,7 @@ s32 MemoryManager::PoolCommit(VAddr virtual_addr, u64 size, MemoryProt prot, s32
     new_vma.prot = prot;
     new_vma.name = "anon";
     new_vma.type = Core::VMAType::Pooled;
+    new_vma.is_system_module = false;
     new_vma.phys_areas.clear();
 
     // Find suitable physical addresses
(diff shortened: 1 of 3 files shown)


par274:
[89171d8](https://github.com/shadps4-emu/shadPS4/pull/4058/commits/89171d89bf6694f8dbb4dd9bce5fec7479187c92)

This will likely resolve the patch addressing issue in other games. I'm tested on Bloodborne, everything is fine.

par274:
I think DevKit/Extra DMem is being miscalculated in games like TLOU2. I'm not using these settings right now. I think PR could be merged.

src/core/memory.cpp
```diff
@@ -1344,6 +1377,95 @@ void MemoryManager::InvalidateMemory(const VAddr addr, const u64 size) const {
...
+    const bool has_physical_tracking =
+        vma.type == VMAType::Direct || vma.type == VMAType::Flexible || vma.type == VMAType::Pooled;
+    if (has_physical_tracking) {
+        // Direct/flexible/pooled mappings should expose at least one physical sub-area when
+        // committed.
+        return !vma.phys_areas.empty();
+    }
+
+    // Non-phys-tracked mappings (code/stack/file) are committed through address-space map calls.
+    return vma.type == VMAType::Code || vma.type == VMAType::Stack || vma.type == VMAType::File;
```
StevenMiller123: System modules shouldn't count toward flex mem, only modules from the game.
Something needs to be done to separate the two properly, otherwise titles like BLEACH: Rebirth of Souls break.

src/core/memory.cpp
```diff
@@ -1344,6 +1377,95 @@ void MemoryManager::InvalidateMemory(const VAddr addr, const u64 size) const {
...
+    if (range_start >= range_end) {
+        return 0;
+    }
+
+    u64 mapped_bytes = 0;
+    auto it = vma_map.upper_bound(range_start);
+    if (it != vma_map.begin()) {
+        it = std::prev(it);
+    }
+    while (it != vma_map.end() && it->second.base < range_end) {
```
StevenMiller123, line 1473: I'm concerned about the additional overhead of a loop like this, is there a reason we need all this instead of just tracking the remaining free flex mem?

src/core/memory.h
```diff
@@ -181,8 +183,17 @@ class MemoryManager {
...
     }
 
+    u64 GetUsedFlexibleSize() const {
+        return flexible_mapped_usage;
+    }
+
     u64 GetAvailableFlexibleSize() const {
-        return total_flexible_size - flexible_usage;
+        const u64 used = GetUsedFlexibleSize();
+        return used < total_flexible_size ? total_flexible_size - used : 0;
```
StevenMiller123, line 196: Is there a reason you have these workarounds for flex usage exceeding flex total?
That case shouldn't be possible.

StevenMiller123 (changes requested):

src/core/memory.cpp
par274: Good point. I need to distinguish game allocations from system/runtime mappings

src/core/memory.cpp
par274: From what I’ve seen `vma_map` isn’t huge and this is mostly hit during init.

I tried caching here, but it quickly became inaccurate. We already have a delta-based counter, and shortcuts here just break correctness.

src/core/memory.h
par274: In theory it shouldn’t happen.

In practice, I've encountered temporary synchronization issues with flexible meters, so I added this as a safeguard against insufficient data flow and crashes.

src/core/memory.cpp
StevenMiller123: > From what I’ve seen vma_map isn’t huge and this is mostly hit during init.

This entirely depends on the game. In some titles, vma_map gets to be thousands of entries large, and said games usually run several MapMemory calls per second. Mainly comes into play with Unreal Engine or Unity titles.

> I tried caching here, but it quickly became inaccurate. We already have a delta-based counter, and shortcuts here just break correctness.

What I'm curious about is why we need all of this overhead on top of the "delta-based counter". Is there some edge case that requires all this extra work to achieve accurate flexible tracking?

par274: If I remove this loop, the issue immediately comes back, so I had to add extra works.
So far I’ve only tested this on TLOU2, but without this logic the flexible tracking becomes unreliable.

If we can guarantee that the delta is always updated correctly for all cases, then this could be simplified. But right now, removing it causes real regressions.

par274:
@StevenMiller123 

I tried to reduce unnecessary overhead. I avoided a second vma scan in flexible usage accounting. Perhaps that corrected the complexity you were trying to explain.

georgemoralis:
@StevenMiller123 can you review again this?

StevenMiller123:
I still can't see the purpose in all this extra overhead for tracking flexible memory use. "I added this because TLOU2 needs it" isn't enough reason, and to me, suggests further issues.

I'll work on a proper fix when I have more time, I need to do some other hardware tests too.

PR #4060, opened 2026-02-20
input: reduce multi-controller pad latency by trimming stale states

byeugenius:
## Problem

In UFC 4, simultaneous use of two controllers causes noticeable input delay, while each controller used alone does not.  
This points to input age/backlog under higher event pressure rather than a single-controller mapping issue.

## Root Cause

Pad state delivery used FIFO consumption without trimming stale backlog in `GameController::ReadStates()`.  
When event production outpaced reads (especially with 2 active controllers), old states accumulated and were still delivered to the game, resulting in delayed input samples.

Additionally, a 4 ms gyro/accel timer was always active, adding extra state pushes even when motion controls were not needed, increasing queue pressure.

## Changes

- `src/input/controller.h`
  - Added `RingBufferQueue::Size()`.
  - Added `RingBufferQueue::DropOldest(size_t count)` to explicitly trim stale backlog.

- `src/input/controller.cpp`
  - Reduced queue depth from 64 to 32 to lower maximum stale window.
  - Added mutex protection in `ReadState()` for a consistent snapshot.
  - Updated `ReadStates()` to:
    - trim oldest excess states when queue size exceeds requested batch,
    - return freshest relevant states,
    - fallback to current state if queue is empty.

- `src/sdl_window.cpp`
  - Reintroduced regular controller polling at 33 ms (`PollController`).
  - Enabled 4 ms gyro/accel polling only when motion controls are enabled.

## Why This Fix Works

Under dual-controller load, stale queue entries are now dropped before consumption, so the game receives fresh state data instead of delayed historical samples.  
Input path pressure is also reduced by avoiding unnecessary high-frequency motion pushes when motion controls are off.

## Scope / Risk

- Localized to input/pad pipeline.
- No API surface changes.
- Single-controller behavior remains consistent while reducing worst-case latency under high input throughput.

### DIFF
--- a/src/input/controller.cpp
+++ b/src/input/controller.cpp
@@ -61,9 +61,10 @@ void State::OnAccel(const float accel[3]) {
     acceleration.z = accel[2];
 }
 
-GameController::GameController() : m_states_queue(64) {}
+GameController::GameController() : m_states_queue(32) {}
 
 void GameController::ReadState(State* state, bool* isConnected, int* connectedCount) {
+    std::lock_guard lg(m_states_queue_mutex);
     *isConnected = m_connected;
     *connectedCount = m_connected_count;
     *state = m_state;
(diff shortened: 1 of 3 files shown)


y65yptrjjz-glitch:
onde coloca os arquivos

kalaposfos13:
This, while a valid issue, is the wrong solution for that issue, and breaks multiple other games, plus it's just AI garbage. From the only part of your description that I think isn't AI hallucinations, which is how to reproduce it (have two controllers connected), I'm pretty sure that #3832 indirectly fixed this problem already anyway.

PR #4065, opened 2026-02-22
Kernel.Event: Implement kqueue and kevent

StevenMiller123:
The bulk of this PR is a refactor, swapping our SceKernelEqueue to a 64-bit integer to match real hardware behavior, and implementing the necessary structure to store equeues instead of relying on games passing them around. I also renamed our structs to OrbisKernel prefixes to match our modern standards.
Additionally, to make this implementation reasonable, I moved timer scheduling to `EqueueInternal::AddEvent`, and changed logic for data storage in timer and small timer events to match libkernel behavior.

This needs testing, the only title I own that calls these is Assassin's Creed Unity, and that isn't really affected by this, since it's just using kevent to add a read event to their equeue.

As far as I've tested locally, I haven't broken anything involving normal equeue event usage.

### DIFF
--- a/src/core/file_sys/fs.h
+++ b/src/core/file_sys/fs.h
@@ -85,7 +85,8 @@ enum class FileType {
     Device,
     Socket,
     Epoll,
-    Resolver
+    Resolver,
+    Equeue
 };
 
 struct File {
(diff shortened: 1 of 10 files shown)


StevenMiller123:
Draft while I touchup a few things

StevenMiller123:
Nevermind, looked a little harder at Ghidra and I think this all looks fine.
Only potential issue is that kevent is supposed to be interruptible, but handling that would be rather painful currently.

PR #4066, opened 2026-02-22
Mount /data to <userdir>/data instead of <userdir>/data/gameid

kalaposfos13:
Following the behaviour of homebrew enablers, mount /data to the same location for every process, instead of making a separate folder for each. This makes working with homebrew apps easier, as it saves you one folder of indirection. People with preexisting stuff in these folders will have to manually update the contents of it to match the new mount, but given that retail apps don't (and can't on OWF) access this anyway, and are therefore unaffected, that is a sacrifice I'm willing to make.

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -341,11 +341,8 @@ void Emulator::Run(std::filesystem::path file, std::vector<std::string> args,
 
     g_window = window.get();
 
-    const auto& mount_data_dir = Common::FS::GetUserPath(Common::FS::PathType::GameDataDir) / id;
-    if (!std::filesystem::exists(mount_data_dir)) {
-        std::filesystem::create_directory(mount_data_dir);
-    }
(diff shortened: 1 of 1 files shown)


georgemoralis:
If game uses it I can't  accept it 
How ps4 handle it?

kalaposfos13:
On the PS4, while unjailbroken, this mount isn't present at all, and if you're jailbroken and have GoldHEN or HEN running, it is mounted for every process, and everything can see the exact same folder (just like a normal unsandboxed filesystem). Currently, shadPS4 does neither, and instead mounts /data as a unique folder for each app.

StevenMiller123 (approved):
Looks good to me, and matches behavior I've observed on real hardware.

georgemoralis:
Checked data folder I don't have something inside so I guess it's good

georgemoralis (approved):

PR #4067, opened 2026-02-22
Fix some logic bugs in sceHttpUriParse

kalaposfos13:
This PR drops the requirement that all paths must start with `/`, fixes the reported required buffer size for a given uri, and sets the unused members of the out OrbisHttpUriElement parameter to dummy empty strings instead of nullptr in the `file://` code branch.
This fixes issues encountered in PlayStation Vue's 01.24 version.

### DIFF
--- a/src/core/libraries/network/http.cpp
+++ b/src/core/libraries/network/http.cpp
@@ -935,18 +935,24 @@ int PS4_SYSV_ABI sceHttpUriParse(OrbisHttpUriElement* out, const char* srcUri, v
                 pathLength++;
             }
 
-            // Ensure the path starts with '/'
-            if (pathLength > 0 && pathStart[0] != '/') {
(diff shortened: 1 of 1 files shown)


georgemoralis:
Have to test it with the game I fixed this for. I think it is 7 days to die but it is quite common in unity games

StevenMiller123:
[CUSA00552.log](https://github.com/user-attachments/files/25475407/CUSA00552.log)
[CUSA00080.log](https://github.com/user-attachments/files/25475408/CUSA00080.log)

Doesn't seem to affect crashing in TLOU (when networking is enabled) or crashing in Warframe. Both crash seemingly right after a sceHttpUriParse call, so I figured I'd try them.

georgemoralis (approved):
my reference games still working so i guess it's ok

PR #4079, opened 2026-02-25
use eR8G8B8A8Unorm image format instead of eR16G16B16A16Sfloat

evill33t:
if HDR is disabled use eR8G8B8A8Unorm instead of eR16G16B16A16Sfloat 
halves bandwidth and memory for fsr pass

### DIFF
--- a/src/video_core/renderer_vulkan/host_passes/fsr_pass.cpp
+++ b/src/video_core/renderer_vulkan/host_passes/fsr_pass.cpp
@@ -161,7 +161,7 @@ vk::ImageView FsrPass::Render(vk::CommandBuffer cmdbuf, vk::ImageView input,
     }
 
     if (img.dirty) {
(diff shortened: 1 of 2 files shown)


squidbus (approved):

raphaelthegreat:
Is there a large perf benefit to this? Lower bit widths cause banding with FSR

evill33t:
the benefit is not really large, a few ns - I wasn't aware about possible banding issues

raphaelthegreat:
I wouldn't say its worth it if we are talking about ns

evill33t:
would it be worthwhile to explore FSR 3.1?

raphaelthegreat:
Later FSR versions require motion vectors which cannot be provided in general fashion

georgemoralis:
closed as probably not worthed it based on turtle's comments

PR #4084, opened 2026-02-28
WIP: Async Compute Implementation

TECHNICANGEL:
Work in progress for async compute improvements. This is not yet finished.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -418,3 +418,4 @@ FodyWeavers.xsd
 # JetBrains
 .idea
 cmake-build-*
+nul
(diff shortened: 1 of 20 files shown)


squidbus:
I think something is up with your settings (maybe line endings?) because some of the file diffs are the entire file instead of the actual changes, making it difficult to see what changed.

TECHNICANGEL:
> I think something is up with your settings (maybe line endings?) because some of the file diffs are the entire file instead of the actual changes, making it difficult to see what changed.

oh shit

raphaelthegreat:
Does this PR reroute all Dispatch functions to a separate queue? That doesn't seem correct to me. Has the performance been tested also? I feel the extra added synchronization wouldn't help

PR #4086, opened 2026-02-28
fixed an issue in sceKernelRemoveExceptionHandler

georgemoralis:


georgemoralis:
found in CUSA11616

StevenMiller123 (commented):
Makes sense, but it would be good to make sure there wasn't some sceKernelInstallExeptionHandler call that failed before

kalaposfos13:
of course something just had to use the one singular signal I reserved for HLE thinking nothing will use it anyway

Yes, I agree with Stephen, a log would be preferable here.

src/core/libraries/kernel/threads/exception.cpp
```diff
@@ -276,7 +276,12 @@ int PS4_SYSV_ABI sceKernelRemoveExceptionHandler(s32 signum) {
         return ORBIS_KERNEL_ERROR_EINVAL;
     }
     int const native_signum = OrbisToNativeSignal(signum);
-    ASSERT_MSG(Handlers[native_signum], "Invalid parameters");
+    if (!Handlers[native_signum]) {
+        LOG_WARNING(Lib_Kernel,
+                    "sceKernelRemoveExceptionHandler: removing non-installed handler for signum {}",
```
StevenMiller123: Did you forget how our logs work? There is no point in having the function name logged when the log macros add the name.

georgemoralis: done :D

PR #4091, opened 2026-03-01
changed readbacks mode to Relaxed,Precised

georgemoralis:


kalaposfos13:
@georgemoralis "precised" is not a real word. :jim:

PR #4100, opened 2026-03-05
Fix R128 flag for IMAGE_STORE

red-prig:
Make Shadow Happy!


georgemoralis:
yah what we should test with that any ideas?:D

StevenMiller123:
God of War and God of War Ragnarok both use R128 if I remember right, I plan to test those later today (though I know you have one of those)

red-prig:
Well, I only did this so that the shader test wouldn't crash.

StevenMiller123:
No regressions on my end, and fixes those shader tests.
Not seeing any improved games from my library, though my testing is limited.

squidbus (approved):

PR #4102, opened 2026-03-05
Lib.SysModule: Proper HLE implementation

StevenMiller123:
This PR adds a proper implementation for libSceSysmodule, based on decompiling the actual module. Due to the nature of this library, I had to change some notable emulator behaviors to more closely match real hardware. I'm sure there are bugs with this code, so I would like tests before any merging.

Now for some noteworthy changes:
- We no longer pre-load LLEs from a game's sce_module folder, or from our sys_modules folder.
  - Old hacks like placing modules in a game's sce_module folder to bypass our hardcoded list will no longer work due to this change.
- sceSysmoduleIsLoaded and sceSysmoduleIsLoadedInternal now function properly, this might help some titles. 

A couple implementation notes:
- Much of this library relies on hardcoded lists of modules. I dumped most of them from firmware 12.52's libSceSysmodule, and modified them slightly to fit within my internal structs.
- sceSysmoduleUnloadModule is left mostly stubbed, as we don't have a proper implementation for unloading modules. Everything outside the module unloading is handled though.
- libSceSysmodule performs a variety of checks on modules after they're loaded. I skipped most of these, since I wasn't sure how to handle them.
- The hardcoded list of modules we allow is now located in `src/core/libraries/sysmodule/sysmodule_internal.cpp`. As all system module loading is now performed in libSceSysmodule, this made sense to change.
- In my implementation of sceSysmodulePreloadModuleForLibkernel, I load libSceLibcInternal. This is technically inaccurate, as on real hardware, libSceLibcInternal is loaded before libSceSysmodule. Since this function is purely for internal use, this shouldn't cause any issues.
- libSceSysmodule was a mess to decompile in Ghidra. There will likely be inaccuracies in a couple places.

Congrats if you made it through reading all of that 😄


src/core/libraries/sysmodule/sysmodule.cpp
```diff
@@ -0,0 +1,204 @@
...
+static std::mutex g_mutex{};
+
+s32 PS4_SYSV_ABI sceSysmoduleGetModuleHandleInternal(OrbisSysModuleInternal id, s32* handle) {
+    LOG_INFO(Lib_SysModule, "called");
+    if ((id & 0x7fffffff) == 0) {
+        return ORBIS_SYSMODULE_INVALID_ID;
+    }
+
+    std::scoped_lock lk{g_mutex};
+    return isModuleLoaded(id, handle);
```
squidbus: The name `isModuleLoaded` seems a bit misleading to me given its use here to also get the handle, confused me for a bit reading this. Would like if there's some better name that can describe its semantics more completely.

squidbus (commented):
Looks pretty good to me and lines up with how I remember this working, just one comment (and assuming testing goes well).

src/core/libraries/sysmodule/sysmodule.cpp
StevenMiller123: Would `getModuleHandle` work as a better name?

squidbus: Yeah, I think that makes sense.

PR #4104, opened 2026-03-06
Lib.Ssl2: Stub sceSslGetCaCerts to return empty cert data instead of no certs.

StevenMiller123:
This is enough to get past initialization checks in Rise of the Tomb Raider.
Did a stub on sceSslFreCaCerts too so we don't leak memory from creating certificate data.

### DIFF
--- a/src/core/libraries/network/ssl2.cpp
+++ b/src/core/libraries/network/ssl2.cpp
@@ -5,6 +5,7 @@
 #include "core/libraries/error_codes.h"
 #include "core/libraries/libs.h"
 #include "core/libraries/network/ssl2.h"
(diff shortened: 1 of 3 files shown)


StevenMiller123:
For reference, Rise of the Tomb Raider v1.00 now crashes while loading ingame.
```
[Debug] <Critical> (shadPS4:GpuCommandProcessor) decode.cpp:264 operator(): Assertion Failed!
Instruction format table incomplete for opcode IMAGE_ATOMIC_CMPSWAP (1418, encoding = 0xf0000000)
```

image

Updated versions of the game still crash instantly.

PR #4105, opened 2026-03-06
VdecDecoder::Flush fix stop video

DanielSvoboda:
The game `Burnout Paradise Remastered` would show the intro video and then freeze;
with this PR, the game changes from `Boots to Playable`

This fixes the issue where the video would never finish because the decoder reached the end but wasn’t signaling completion. Now Flush properly indicates that all frames have been output.

image


src/core/libraries/videodec/videodec_impl.cpp
```diff
@@ -118,7 +118,7 @@ s32 VdecDecoder::Decode(const OrbisVideodecInputData& pInputDataIn,
 s32 VdecDecoder::Flush(OrbisVideodecFrameBuffer& pFrameBufferInOut,
                        OrbisVideodecPictureInfo& pPictureInfoOut) {
     pPictureInfoOut.thisSize = sizeof(pPictureInfoOut);
-    pPictureInfoOut.isValid = false;
+    pPictureInfoOut.isValid = true;
```
squidbus: This doesn't seem right, if there is a valid frame `isValid` should already be set to `true` below. Which case below is it exiting out on before setting it to true?

DanielSvoboda: I made a change. When the decoder reaches the end during Flush, avcodec_receive_frame returns EAGAIN. That means there are no more frames left in the decoder, so I mark isValid = true and isErrorPic = false to signal completion.

StevenMiller123: This doesn't really address the fact that isValid is supposed to imply there is a valid frame. If there are no frames, then isValid should be false.

squidbus: Yeah I think the question here would be, why are we getting no frames, and if we are supposed to get no frames then why does the game want a valid frame?

DanielSvoboda: The previous code set isValid = true when avcodec_receive_frame() returned EAGAIN/EOF, which could report a valid picture even when no frame was produced.

Now isValid is only set when a real frame is decoded. If Flush() returns no frames (frameCount == 0), we emit a final valid picture so the game can exit its decode loop.

StevenMiller123: The part we're questioning is why you need to set `isValid = true` when there are no valid frames (that `frameCount == 0` check in your current push). This seems like a hack to me.

DanielSvoboda: What I observed is that when Flush() returns with isValid = false and no frames, The game calls once again sceVideodecFlush() and the video stalls at the last frame.
Returning a final isValid = true when frameCount == 0 allows the player to exit the decode loop and continue normally.

squidbus: Have you confirmed that this is the real behavior of flush though? Because otherwise it's a hack, and my suspicion would be that the real issue is we're not giving frames when the game is supposed to get them.

PR #4106, opened 2026-03-06
Vulkan pipeline cache: finish preload state before profile-mismatch early return

collinmcg:
This fixes a cache lifecycle edge case in `PipelineCache::WarmUp()`.

If the stored pipeline profile doesn't match the current system profile, we currently return early. In archive mode that leaves cache DB state in preload/read-only mode for the rest of the run, while runtime registration paths can still call `DataBase::Save(...)`.

The change is just:
- call `Storage::DataBase::Instance().FinishPreload()` in the mismatch branch
- keep the same behavior otherwise (still ignore incompatible warmup cache)

So this is not changing compatibility logic, only cleaning up DB state before continuing.

Validation on my side:
- walked the control flow from `WarmUp()` to `RegisterShader*`/`RegisterPipelineData` save paths
- local configure succeeds
- full local build currently fails in unrelated external code (`externals/openal-soft`) in this environment

### DIFF
--- a/src/video_core/renderer_vulkan/vk_pipeline_serialization.cpp
+++ b/src/video_core/renderer_vulkan/vk_pipeline_serialization.cpp
@@ -316,6 +316,9 @@ void PipelineCache::WarmUp() {
     if (std::memcmp(profile_data.data(), &profile, sizeof(profile)) != 0) {
         LOG_WARNING(Render,
                     "Pipeline cache isn't compatible with current system. Ignoring the cache");
+        // Keep DB lifecycle consistent before early return: in archived mode, Save() later in the
+        // session may still be called and expects writable archive state.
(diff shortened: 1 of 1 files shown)


red-prig:
What's the point of save an incompatible cache? Explain in practical terms.

collinmcg:
It doesn’t save incompatible cache for use; it lets newly rebuilt compatible cache actually get written after a mismatch so later runs are smoother instead of recompiling again.

red-prig:
I looked through the code in more detail, oddly enough, your edit makes sense, but not for the reasons you indicated, the AI ​​is simply lying to you.

red-prig:
`session may still be called and expects writable archive state.`
the first comment was right

collinmcg:
That's what I get for intervening I guess

PR #4107, opened 2026-03-07
WIP: Buffer cache: run LRU GC pass when memory threshold is exceeded

collinmcg:
While checking stutter-related paths, I noticed `BufferCache::RunGarbageCollector()` was effectively a no-op under pressure.

It calculated thresholds and built a cleanup lambda, but never actually walked the LRU list, so nothing got reclaimed.

This patch does the minimal thing:
- run `lru_cache.ForEachItemBelow(gc_tick - ticks_to_destroy, clean_up)`
- make the lambda return `bool` so we can stop after `max_deletions`

No refactor or behavior changes outside this GC path.

Validation:
- verified call path from `Rasterizer::OnSubmit()` to `RunGarbageCollector()`
- local configure succeeds
- full local build still fails in unrelated external code (`externals/openal-soft`) in this environment
- checked compile of modified TU with compile commands (`buffer_cache.cpp`)

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -848,14 +848,17 @@ void BufferCache::RunGarbageCollector() {
     int max_deletions = aggressive ? 64 : 32;
     const auto clean_up = [&](BufferId buffer_id) {
         if (max_deletions == 0) {
-            return;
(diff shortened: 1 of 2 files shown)


red-prig:
Now run tests without using AI on real games and look for differences and regressions.

PR #4108, opened 2026-03-07
Improve signal emulation

kalaposfos13:
This PR technically addresses issues where games would crash on trying to remove a signal handler on signals with not yet registered handlers, but it also adds posix_sigemptyset, posix_sigaction and posix_sigkill, and refactors the sce wrappers for them to use those, plus it also fixes a few other miscallenous bugs from the previous iteration that came from me not testing thoroughly enough, and no one else giving a shit since this isn't something commercial games use much. I've tested this extensively on Linux, but Windows and macOS still need verification.

### DIFF
--- a/src/core/libraries/kernel/threads/exception.cpp
+++ b/src/core/libraries/kernel/threads/exception.cpp
@@ -317,15 +317,6 @@ s32 PS4_SYSV_ABI posix_sigaction(s32 sig, Sigaction* act, Sigaction* oact) {
             LOG_ERROR(Lib_Kernel, "Unhandled sa_mask: {:x}", act->sa_mask.bits[0]);
         }
     }
-    struct sigaction native_oact{};
-    if (act) {
-        native_oact.sa_flags = act->sa_flags;
-        native_oact.sa_sigaction = reinterpret_cast<decltype(native_oact.sa_sigaction)>(
-            act->__sigaction_handler.sigaction);
-        if (act->sa_mask.bits[0] != 0) {
-            LOG_ERROR(Lib_Kernel, "Unhandled sa_mask: {:x}", act->sa_mask.bits[0]);
-        }
-    }
     Handlers[sig] = reinterpret_cast<OrbisKernelExceptionHandler>(
         act ? act->__sigaction_handler.sigaction : nullptr);
 
@@ -336,7 +327,17 @@ s32 PS4_SYSV_ABI posix_sigaction(s32 sig, Sigaction* act, Sigaction* oact) {
         LOG_WARNING(Lib_Kernel, "We can't install a handler for native signal {}!", native_sig);
         return ORBIS_OK;
     }
+    struct sigaction native_oact{};
     s32 ret = sigaction(native_sig, act ? &native_act : nullptr, oact ? &native_oact : nullptr);
+    if (oact) {
+        oact->sa_flags = native_oact.sa_flags;
+        oact->__sigaction_handler.sigaction =
+            reinterpret_cast<decltype(oact->__sigaction_handler.sigaction)>(
+                native_oact.sa_sigaction);
+        if (native_oact.sa_mask.__val[0] != 0) {
+            LOG_ERROR(Lib_Kernel, "Unhandled sa_mask: {:x}", native_oact.sa_mask.__val[0]);
+        }
+    }
     if (ret < 0) {
         LOG_ERROR(Lib_Kernel, "sigaction failed: {}", strerror(errno));
         *__Error() = ErrnoToSceKernelError(errno);
--- a/src/core/libraries/kernel/threads/exception.cpp
+++ b/src/core/libraries/kernel/threads/exception.cpp
@@ -305,8 +305,6 @@ s32 PS4_SYSV_ABI posix_sigaction(s32 sig, Sigaction* act, Sigaction* oact) {
                   "Guest is attempting to use SIGRT signals, which aren't available on this "
                   "platform (signal: {})!",
                   sig);
-        *__Error() = POSIX_EINVAL;
-        return ORBIS_FAIL;
     }
 #endif
     LOG_INFO(Lib_Kernel, "called, sig: {}, native sig: {}", sig, native_sig);
--- a/src/core/libraries/kernel/threads/exception.cpp
+++ b/src/core/libraries/kernel/threads/exception.cpp
@@ -278,6 +278,10 @@ s32 PS4_SYSV_ABI posix_sigemptyset(Sigset* s) {
     return 0;
 }
 
+bool PS4_SYSV_ABI posix_sigisemptyset(Sigset* s) {
+    return s->bits[0] == 0 && s->bits[1] == 0;
+}
+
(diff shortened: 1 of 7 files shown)


src/core/signals.cpp
```diff
@@ -99,7 +99,7 @@ void SignalHandler(int sig, siginfo_t* info, void* raw_context) {
     }
     case SIGILL:
         if (!signals->DispatchIllegalInstruction(raw_context)) {
-            if (Libraries::Kernel::Handlers[sig]) {
+            if (Libraries::Kernel::Handlers[Libraries::Kernel::NativeToOrbisSignal(sig)]) {
```
squidbus, line 102: I guess it's technically not part of this PR since this logic was already here, but: the signal system here is designed already to support multiple handlers. So the libkernel HLE itself could just register its handler with `RegisterAccessViolationHandler` / `RegisterIllegalInstructionHandler` at lowest priority.

src/core/libraries/kernel/threads/exception.cpp
```diff
@@ -243,76 +266,99 @@ void ExceptionHandler(void* arg1, void* arg2, void* arg3, PCONTEXT context) {
...
-        act.sa_sigaction = nullptr;
-        sigemptyset(&act.sa_mask);
-        sigaction(native_signum, &act, nullptr);
+    struct sigaction native_oact{};
+    s32 ret = sigaction(native_sig, act ? &native_act : nullptr, oact ? &native_oact : nullptr);
+    if (oact) {
+        oact->sa_flags = native_oact.sa_flags;
+        oact->__sigaction_handler.sigaction =
+            reinterpret_cast<decltype(oact->__sigaction_handler.sigaction)>(prev_handler);
+        if (native_oact.sa_mask.__val[0] != 0) {
```
squidbus: Can use `!sigemptyset(native_oact.sa_mask)` for this check cross-platform, not sure if there's a good way for getting the raw value though to log. On macOS I believe `sa_mask` type just resolves to a plain `uint32_t`.

squidbus:
Copying this here: should probably clear the signal mask on the game main thread so that any signal mask inherited from the host doesn't impact game signal handling. For example my IDE masks out `SIGQUIT`. Note that this should be done once for the main thread and not subsequent threads, as threads spawned by the game should inherit from the creating thread.

Example:
```diff
diff --git a/src/core/linker.cpp b/src/core/linker.cpp
index 0b80ecacc..db1a1b646 100644
--- a/src/core/linker.cpp
+++ b/src/core/linker.cpp
@@ -21,6 +21,10 @@
 #include "core/tls.h"
 #include "ipc/ipc.h"
 
+#ifndef _WIN32
+#include <signal.h>
+#endif
+
 namespace Core {
 
 static PS4_SYSV_ABI void ProgramExitFunc() {
@@ -106,6 +110,14 @@ void Linker::Execute(const std::vector<std::string>& args) {
 
     main_thread.Run([this, module, &args](std::stop_token) {
         Common::SetCurrentThreadName("Game:Main");
+
+#ifndef _WIN32
+        // Clear any existing signal mask for game threads.
+        sigset_t emptyset;
+        sigemptyset(&emptyset);
+        pthread_sigmask(SIG_SETMASK, &emptyset, nullptr);
+#endif
+
         if (auto& ipc = IPC::Instance()) {
             ipc.WaitForStart();
         }
```

src/core/libraries/kernel/threads/exception.cpp
```diff
@@ -13,23 +14,25 @@
 #else
 #include <csignal>
 #endif
+#include <unordered_set>
+#include <core/libraries/kernel/posix_error.h>
```
squidbus: Style nit: includes out of order & `core/libraries/kernel/posix_error.h` should use quotes

src/core/libraries/kernel/threads/exception.cpp
```diff
@@ -243,76 +266,103 @@ void ExceptionHandler(void* arg1, void* arg2, void* arg3, PCONTEXT context) {
...
 #endif
-    return ORBIS_OK;
-}
+    LOG_INFO(Lib_Kernel, "called, sig: {}, native sig: {}", sig, native_sig);
+    struct sigaction native_act{};
+    if (act) {
+        native_act.sa_flags = act->sa_flags; // todo check compatibility, on Linux it seems fine
+        native_act.sa_sigaction =
+            reinterpret_cast<decltype(native_act.sa_sigaction)>(SigactionHandler);
+        if (posix_sigisemptyset(&act->sa_mask)) {
```
squidbus:
```suggestion
        if (!posix_sigisemptyset(&act->sa_mask)) {
```

Should be checking if the set is *not* empty.

src/core/libraries/kernel/threads/exception.cpp
```diff
@@ -243,76 +266,103 @@ void ExceptionHandler(void* arg1, void* arg2, void* arg3, PCONTEXT context) {
...
-        act.sa_sigaction = nullptr;
-        sigemptyset(&act.sa_mask);
-        sigaction(native_signum, &act, nullptr);
+    struct sigaction native_oact{};
+    s32 ret = sigaction(native_sig, act ? &native_act : nullptr, oact ? &native_oact : nullptr);
+    if (oact) {
+        oact->sa_flags = native_oact.sa_flags;
+        oact->__sigaction_handler.sigaction =
+            reinterpret_cast<decltype(oact->__sigaction_handler.sigaction)>(prev_handler);
+        if (sigisemptyset(&native_oact.sa_mask)) {
```
squidbus:
```suggestion
        if (!sigisemptyset(&native_oact.sa_mask)) {
```

Also my bad as this doesn't seem to actually exist on macOS (glibc extension)... you can add something like:
```
#ifdef __APPLE__
#define sigisemptyset(x) (*(x) == 0)
#endif
```

...and that should be good.

squidbus (approved):
Final comments from me, looks good otherwise.

StevenMiller123:
Breaks Unity on Windows.
image
[CUSA29953.log](https://github.com/user-attachments/files/25816052/CUSA29953.log)

Graubek:
Killzone Shadowfall crashes before reaching the menu on
`[Core] <Error> (Game:Main) cpu_patches.cpp:794 PatchesIllegalInstructionHandler: Failed to patch address 7000010b1ad5 -- mnemonic: vmovdqa64`
[shad_log.txt](https://github.com/user-attachments/files/25816319/shad_log.txt)
Edit: When launching again the game still crashes before reaching the menu but without an error this time
[shad_log1.txt](https://github.com/user-attachments/files/25816403/shad_log1.txt)

StevenMiller123:
Doesn't look like Killzone is affected by this PR. I don't see anything in the log that would suggest any changes from this PR.

Graubek:
> Doesn't look like Killzone is affected by this PR. I don't see anything in the log that would suggest any changes from this PR.

It works on newest nightly but not on this pr

StevenMiller123:
Breaks Genshin Impact on both Windows and Linux
[CUSA23681.log](https://github.com/user-attachments/files/25816526/CUSA23681.log)

Graubek:
Killzone is fixed now

PR #4117, opened 2026-03-11
video_core: buffer/render/liverpool optimizations

Kyoskii:
results seem good in bloodborne, +8 fps in areas (needs more testing on amd/intel)
this was tested with nvidia gpu amd cpu


StevenMiller123:
Compiled locally to test, and unfortunately this breaks a number of games I own. I've sent specific errors in titles to @Kyoskii directly.

PR #4122, opened 2026-03-13
psf file format: close file after Encode()

CAHbKA-IV:
"The Last Guardian" game rewrites psf file in every save, so I got "permission denied" during saving game (file was not closed after previous save).


Linear524:
Finally !
Tired of frequent crashes due to save file access issue...

PR #4124, opened 2026-03-13
Imgui: translations

rainmakerv3:
As of now, string translations in shadPS4 are only for Qt, so all imGui strings (such as the trophy popup and  the imgui widgets) are untranslated.

Since there aren't that many strings, I think we can just do it this way (basically just a bunch of hardcoded tables in one file).

Please feel free to suggest if there are better ways to do these translations.

Putting up the PR just for discussion. I'm leaving this as draft for now. I'll adapt it to the new settings struct when it gets merged and finish all the tables after that.


georgemoralis:
i think now this can be added to main :D

rainmakerv3:
Ok but I think I have to reorganize the tables first. And also I'm working on the changes to trophy viewer and open/delete savedata with the new folder structures, and probably adding the user manager dialog too.

rainmakerv3:
I think organizing by language like this is better for the translators. 

If you have a list of transatable strings I can already put them in.

PR #4130, opened 2026-03-16
V_CMP_U64 fix

georgemoralis:
handle src1 as VectorGPR in V_CMP_U64
This should fix some vector_alu.cpp asserts


src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -1136,52 +1136,105 @@ void Translator::V_CMP_U32(ConditionOp op, bool is_signed, bool set_exec, const
...
+        case OperandField::VectorGPR: {
+            IR::U32 low = ir.GetVectorReg(IR::VectorReg(inst.src[0].code));
+            IR::U32 high = ir.GetVectorReg(IR::VectorReg(inst.src[0].code + 1));
+            return ir.PackUint2x32(ir.CompositeConstruct(low, high));
+        }
+        case OperandField::VccLo: {
+            IR::U1 vcc_bit = ir.GetVcc();
+            // Fix: Cast both Imm64 calls to avoid ambiguity
+            IR::Value vcc_val =
+                ir.Select(vcc_bit, ir.Imm64(static_cast<u64>(-1)), ir.Imm64(static_cast<u64>(0)));
```
StevenMiller123: Our IR Select doesn't support U64 values. You'll need to add an IR function for that, or change your code.

squidbus:
This doesn't look right to me, we're no longer treating SGPRs as per-thread bits?

squidbus:
(cc: @raphaelthegreat )

georgemoralis:
ok added an implemenation for avoid using u64 emitter , should be more optimal and correct @squidbus if you agree with this :D

squidbus:
Currently you're still using `GetScalarReg`, before we were using `GetThreadBitScalarReg`. The difference being that the latter will use the single bit of the register belonging to each shader thread in the wave to compare, since typically that is how these 64-bit compares are used with scalar registers; they are comparing the 64 bits coming from all the wave threads.

It's possible for this to not be the case and it to be a normal 64-bit compare I suppose, but you can't really ignore the thread bit case.

raphaelthegreat:
You seem to have reimplemented this as a 64-bit op which is incorrect, in vast majority of cases this should use the thread bit values which is what I had fixed in my PR. Usage of VGPRs are inputs is indicative of needing 64-bit usage, but currently that would also break because v_mov_b32 can't handle a thread bit source.

Kravickas:
how about S_MOV_B64 as an example for this V_CMP_U64 ?

georgemoralis:
will be fixed with another pr shortly

Issue #4129, opened 2026-03-15
[GAME BUG]: "Unable to allocate guest TCB" assertion crash during 13 Sentinels: Aegis Rim gameplay.


rosenkolev1:
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.
- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [ ] I have disabled all patches and cheats and the issue is still present. (haven't tried without the specific game patch)
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadPS4/wiki/I.-Quick-start-%5BUsers%5D#4-dumping-firmware-modules) installed.

### Describe the Bug

I've been playing a game recently on the emulator for around 2-3 hours per session and occasionally I would crash 1 time per session during gameplay. I don't have any repro steps, the crash instances happened at times that seemingly don't have anything in common gameplay/game load wise + they are rare. 

* Emulator Version: 0.14.0
* Game: 13 Sentinels: Aegis Rim (CUSA19610)
* Game Patch: 1.02 
* Mods: none,
* Logs: [logs1.txt](https://github.com/user-attachments/files/26005437/logs1.txt)

Looking at the code, I am not sure why we have an assert there instead of a non-critical log, seems like if we just return nullptr when the check would fail inside "TcbCtor()", we have graceful handling at the callsite, i.e. at "ThreadState::Alloc()", afterwards. Not sure if worth noting, but at least 2458 unique threads have been made seemingly, judging from the "Thread2458" log.

Unrelated: [This TidCounter](https://github.com/shadps4-emu/shadPS4/blob/30ff9cf05045124d47843512568fdd7cfbbbff3d/src/core/libraries/kernel/threads/pthread.cpp#L245) seems to not be thread safe at a glance, shouldn't it be like atomic or something?

P.S. I don't really need help, the crashes don't impact me much anyway because the game has frequent auto save. + I cloned the repo and made a custom build of 0.14.0 where I turned off the assert and just returned nullptr instead, curious to see how that will pan out lol.

### Reproduction Steps

None, see the description of the bug.

### Specify OS Version

Windows 11

### CPU

AMD Ryzen 7 5700X 8-Core 3.40GHz

### GPU

Nvidia RTX4070 Super

[...]

PR #4132, opened 2026-03-17
Fix tls deallocation memory leak

rosenkolev1:
## The fix

Static TLS memory from the Main module + TCB memory isn't destroyed during `TcbDtor()` (now `FreeTLSForThread()`) destruction as it is never deleted, only the DTV (Dynamic Thread Vector) is deleted correctly. On the next `ThreadState::Alloc` -> `TcbCtor()` (now `InitializeTLSForThread()`) call, new static TLS + TCB memory would be allocated and the old static TLS + TCB memory would leak.

Fixed by properly freeing the static TLS + TCB memory inside `TcbDtor()` (now `FreeTLSForThread()`).

I also made the critical error into just a regular error, since we seem to have handling for null tcb inside `ThreadState::Alloc()`. Beyond that, it's up to the game to handle failed thread allocation gracefully - which may prevent a crash or at least delay it. Happy to revert this if the consensus is that failing to allocate TLS should always be considered unrecoverable.

## The goal

Hopefully solves [this](https://github.com/shadps4-emu/shadPS4/issues/4129).

I suspect what happened is that the memory pool got fragmented enough from the leaked static TLS + TCB allocations after enough time passed (and enough threads were created and destroyed), hence a TLS allocation eventually failed, hitting the assert at `TcbCtor()` for being unable to allocate guest TCB.

Played for 3 hours today with these changes on top of the [0.14.0 release](https://github.com/shadps4-emu/shadPS4/releases/tag/v.0.14.0), spawned way more threads (~3900 unique threads compared to my past crashes ~2500-2800 unique threads) during gameplay and didn't crash. Hopefully it wasn't just me being lucky 😅 

## Misc refactoring that is unrelated and should probably be in a separate PR but I am lazy.

Refactored some code around the stubs. Purpose was to avoid double look up with `AeroLib::FindByNid()` directly and then also inside `AeroLib::GetStub()`. It's doing binary search so it can't be that slow anyway but it seemed like a simple change so changed it.

### DIFF
--- a/src/core/aerolib/stubs.cpp
+++ b/src/core/aerolib/stubs.cpp
@@ -27,7 +27,7 @@ u64 UnresolvedStub() {
 }
 
 static u64 UnknownStub() {
-    LOG_ERROR(Core, "Returning zero to {}", __builtin_return_address(0));
+    LOG_ERROR(Core, "Stub: Unknown (missing nid) called, returning zero to {}", __builtin_return_address(0));
     return 0;
 }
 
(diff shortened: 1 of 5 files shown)


red-prig:
I analyzed the code. Firstly, there's a lot of syntactic sugar that doesn't change the behavior at all. The clarification of nid errors is completely irrelevant. The comments from AI about slot fragmentation are complete nonsense. Freeing the static tls `linker->FreeTlsForNonPrimaryThread(tls_base);` is also complete nonsense; it should never be deleted.

red-prig:
Removing the "Unable to allocate guest TCB" assertion masks a real error that occurs elsewhere, which is completely unacceptable.
Conclusion: teach computer science yourself, don't rely on AI.

georgemoralis:
red_prig already review it...

PR #4135, opened 2026-03-17
texture_cache: Handle image overlap when guest_size differs with same resource levels

psabhishek26:
Fixes a crash (UNREACHABLE) in ResolveOverlap when the guest requests an image at the same address with the same format, tile mode, and dimensions, but with a different array layer count that results in a larger guest_size while having fewer or equal mip levels.

When the new image has a larger guest_size, the cached image is now unconditionally freed since it is definitively superseded. When the cached image is larger or equal, it is reused as a compatible view source.

Tested with Red Dead Redemption CUSA36843


georgemoralis:
this only provides partial implementation and can lead to false results

PR #4139, opened 2026-03-17
MatchingContext

mikusp:
A lot of changes to NpMatching2:
- creates a dedicated MatchingContext class that handles the communication with multiplayer server over websockets
- all of the requests and responses are marshalled using json
- as some response structs contain arrays encoded as a pointer and the size, these are unmarshalled to ...Owned variant, and the proper struct is obtained using view() method
- only some Matching2 methods are implemented, i.e. Search, CreateJoin, Join, as well as MEMBER_JOINED event

The PR is already massive so I figured it is as good point as any to upstream even if it doesn't work end-to-end yet. The server part is also in development but not published at the moment

### DIFF
--- a/src/core/shadnet/matching_context.cpp
+++ b/src/core/shadnet/matching_context.cpp
@@ -1,0 +1,349 @@
+//  SPDX-FileCopyrightText: Copyright 2026 shadPS4 Emulator Project
+//  SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <magic_enum/magic_enum.hpp>
+
+#include "common/logging/log.h"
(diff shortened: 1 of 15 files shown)


src/core/shadnet/matching_context.cpp
```diff
@@ -0,0 +1,337 @@
+//  SPDX-FileCopyrightText: Copyright 2024-2026 shadPS4 Emulator Project
```
StevenMiller123: New files are made in 2026, not 2024.

georgemoralis:
it is working on shadnet deprecated

PR #4140, opened 2026-03-17
NpScore

mikusp:
Functions that puyo used in my testing are implemented. The general idea is to access the scoring server over HTTP, and as with matching, the server is in the works.

### DIFF
--- a/src/core/libraries/np/np_score.cpp
+++ b/src/core/libraries/np/np_score.cpp
@@ -1,12 +1,227 @@
 // SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#ifdef _WIN32
+#ifndef NOGDI
+#define NOGDI
+#endif
+#endif
+
+#include <future>
+
 #include "common/logging/log.h"
 #include "core/libraries/error_codes.h"
 #include "core/libraries/libs.h"
+#include "core/libraries/np/np_error.h"
+#include "core/libraries/np/np_manager.h"
 #include "core/libraries/np/np_score.h"
+#include "core/libraries/np/np_types.h"
+#include "core/libraries/np/object_manager.h"
+#include "core/libraries/system/userservice.h"
+#include "cppcodec/base64_rfc4648.hpp"
+#include "externals/cpp-httplib/httplib.h"
+#include "nlohmann/json.hpp"
 
 namespace Libraries::Np::NpScore {
+
+using json = nlohmann::json;
+using base64 = cppcodec::base64_rfc4648;
+
+const std::string BASE_URL = "http://127.0.0.1:3000";
+
+int PackReqId(int libCtxId, int reqId) {
+    return ((libCtxId & 0xFFFF) << 16) | (reqId & 0xFFFF);
+}
+
+std::pair<int, int> UnpackReqId(int reqId) {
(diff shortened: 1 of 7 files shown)


src/core/libraries/np/np_score.cpp
```diff
@@ -188,14 +581,58 @@ int PS4_SYSV_ABI sceNpScoreGetRankingByRangeA() {
...
+    return ORBIS_OK;
+}
+
+int PS4_SYSV_ABI sceNpScoreGetRankingByRangeAAsync(
+    int reqId, OrbisNpScoreBoardId scoreBoardId, OrbisNpScoreRankNumber startRank,
+    OrbisNpScoreRankDataA* ranks, u64 ranksBytes, OrbisNpScoreComment* comments, u64 commentsBytes,
+    OrbisNpScoreGameInfo* gameInfos, u64 gameInfosBytes, u64 ranksLen,
+    Rtc::OrbisRtcTick* lastUpdate, OrbisNpScoreRankNumber* totalRecords, void* option) {
+    // at least in 9.00 it's identical A vs non-A
+    return sceNpScoreGetRankingByRangeAsync(reqId, scoreBoardId, startRank, ranks, ranksBytes,
```
StevenMiller123: Usually these "A" variants have different inputs. Given the inputs to this function, I believe you should double check to see if `sceNpScoreGetRankingByRangeAsync` uses the same `OrbisNpScoreRankDataA` input for ranks. I doubt it does.

src/core/libraries/np/np_score.h
```diff
@@ -4,64 +4,99 @@
...
+struct OrbisNpScorePlayerRankDataA {
+    int hasData;
+    u8 pad[4];
+    OrbisNpScoreRankDataA rankData;
+};
+
+static_assert(sizeof(OrbisNpScorePlayerRankDataA) == 0x98);
+
+struct OrbisNpScoreRankData {
+    OrbisNpOnlineId onlineId;
```
StevenMiller123, line 52: Based on the information I have, this struct uses an OrbisNpId, not OrbisNpOnlineId.

StevenMiller123 (changes requested):
A couple important issues here, I also see a distinct lack of proper "isPSNSignedIn" checks too.

StevenMiller123:
I've probably got more that uses this library, but here are logs from two easy to check titles. Puyo Puyo Tetris and WE ARE DOOMED log a couple stubs.

[CUSA06509.log](https://github.com/user-attachments/files/26117166/CUSA06509.log)
[CUSA01783.log](https://github.com/user-attachments/files/26117717/CUSA01783.log)

src/core/libraries/np/np_score.cpp
StevenMiller123, line 295: Based on decomp, this should be
`int PS4_SYSV_ABI sceNpScoreCreateTitleCtx(OrbisNpCommunicationId* np_comm_id, OrbisNpCommunicationPassphrase* np_comm_pass, OrbisNpId* np_id)`

Should be relatively easy to handle if you want?

mikusp: I don't really see a point if that function is never being called in retail games 🤔

StevenMiller123: I mean, there's no reason not to 😄

georgemoralis:
already in shadnet , closing

PR #4142, opened 2026-03-18
Write after write sync hazard

Kravickas:
Be safe vulkan, no hazard.

If two commands both write to the same memory region with no barrier between them, the final memory contents are undefined. Without a barrier, the clear could execute before or during the copy.


georgemoralis:
does it affect a specific game?

StevenMiller123:
Seems to fix AMD-specific vertex explosions in GFD engine titles.

image

image

[CUSA15036.log](https://github.com/user-attachments/files/26099701/CUSA15036.log)

georgemoralis:
interesting attempting to see what else it can fix

Kravickas:
found in NHLs and MGS:GZ in theory could produce garbage (more on AMD side driver than NV)

but i forgot 1 commit /facepalm

georgemoralis:
you can always open a new pr :D

PR #4144, opened 2026-03-18
Update current firmware version

kalaposfos13:
We've improved the messages and usability on some screens.
We also patched the PlayStation Vue userland exploit by removing its system license that used to be preinstalled on the system, and we also patched yet another bug in BD-J.


StevenMiller123 (approved):
Doesn't really help much unless you're hacking around locally with LLEs, but still a good change to make.

PR #4145, opened 2026-03-18
New config mode (part1 of 0.15.1 branch series)

georgemoralis:
Since 0.15.1 branch already has a lot of stuff , this part only uses the new emulator_settings. Shouldn't break something else

### DIFF
--- a/src/core/emulator_settings.cpp
+++ b/src/core/emulator_settings.cpp
@@ -584,27 +584,48 @@ bool EmulatorSettingsImpl::TransferSettings() {
         const toml::value& gui = og_data.at("GUI");
         auto& s = m_general;
 
-        // TODO
+        // Transfer install directories
+        try {
+            const auto install_dir_array =
+                toml::find_or<std::vector<std::string>>(gui, "installDirs", {});
+            std::vector<bool> install_dirs_enabled;
 
-        // const auto install_dir_array =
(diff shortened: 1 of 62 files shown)


src/core/emulator_settings.cpp
```diff
@@ -0,0 +1,626 @@
...
+        const toml::value& settings = og_data.at("Settings");
+        auto& s = m_general;
+        setFromToml(s.console_language, settings, "consoleLanguage");
+    }
+
+    if (og_data.contains("GUI")) {
+        const toml::value& gui = og_data.at("GUI");
+        auto& s = m_general;
+
+        // TODO
```
kalaposfos13: TODO finish this before merging.

kalaposfos13 (changes requested):

src/core/emulator_settings.cpp
georgemoralis: go ahend kal

kalaposfos13 (approved):
fuck it let's go and test in production

Linear524:
This broke settings for Last Guardian (Windows)
HDR broken
Readbacks mode broken
Windows size broken
Direct Memory Access settings broken

StevenMiller123:
Make sure you're updating that new config.json instead. If you're using QtLauncher, revert to v0.15.0 until support for the new configs is added.

PR #4150, opened 2026-03-19
chore: add an RFC issue template

AlpinDale:
Based on [vLLM's](https://github.com/vllm-project/vllm/blob/main/.github/ISSUE_TEMPLATE/750-RFC.yml) RFC template.


AlpinDale:
Why did I think an issue title change would skip the CI

PR #4154, opened 2026-03-20
WAW perf fixup

Kravickas:
Don't barrier every render target (different race issue) that causes heavy performance regressions.
Vertex explosions are ''correct'' with AMD GPU in Persona etc. (different issue)


bigol83:
Performance is restored for me on nvidia rtx 3060 and Bloodborne

PR #4155, opened 2026-03-20
Initial Openal (Part 2  of 0.15.1 branch)

georgemoralis:
This adds openal backend . 

To be merged after part 1


StevenMiller123 (commented):
Why is it necessary to have two completely separate Audio3d library implementations?

georgemoralis:
I wasn't able to think a good abstraction model for this and my initial tests showed it might not be that easy. Once stabalized it might rethink to add

PR #4159, opened 2026-03-22
Fix startup crash on malformed patch indexes and short CONTENT_ID

morvaer:
Prevents early startup termination triggered by `CUSA03173` when auto-patch repositories contain missing/unreadable/malformed `files.json` by adding open checks, JSON parse exception handling, and basic schema validation.

Hardens `param.sfo` handling by guarding `CONTENT_ID` substring extraction and falling back to `TITLE_ID` , and by catching invalid `sdk_ver` parsing from `PUBTOOLINFO` with a safe fallback to firmware version.

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -132,7 +132,7 @@ void Emulator::Run(std::filesystem::path file, std::vector<std::string> args,
     std::string id;
     std::string title;
     std::string app_version;
-    u32 sdk_version;
+    u32 sdk_version{};
     u32 fw_version;
     Common::PSFAttributes psf_attributes{};
     if (param_sfo_exists) {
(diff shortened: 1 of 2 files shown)


georgemoralis:
looking good , needs clang fix and we are good for merge

src/emulator.cpp
```diff
@@ -173,7 +175,13 @@ void Emulator::Run(std::filesystem::path file, std::vector<std::string> args,
             sdk_ver_len -= sdk_ver_offset;
             std::string sdk_ver_string = pubtool_info.substr(sdk_ver_offset, sdk_ver_len).data();
             // Number is stored in base 16.
-            sdk_version = std::stoi(sdk_ver_string, nullptr, 16);
+            try {
```
StevenMiller123, line 178: Pubtoolinfo parsing shouldn't trigger exceptions, I checked with homebrew to make sure.
Could you provide more detail about this?

src/emulator.cpp
```diff
@@ -141,10 +141,12 @@ void Emulator::Run(std::filesystem::path file, std::vector<std::string> args,
...
         const auto title_id = param_sfo->GetString("TITLE_ID");
-        if (content_id.has_value() && !content_id->empty()) {
-            id = std::string(*content_id, 7, 9);
-        } else if (title_id.has_value()) {
+        if (content_id.has_value() && content_id->size() >= 16) {
+            id = content_id->substr(7, 9);
+        } else if (title_id.has_value() && !title_id->empty()) {
             id = *title_id;
+        } else if (content_id.has_value() && !content_id->empty()) {
+            LOG_WARNING(Loader, "CONTENT_ID too short to derive game id: {}", *content_id);
```
StevenMiller123, line 149: I'd prefer if we logged all cases of CONTENT_ID issues, as this shouldn't be an issue for any game dumps.

StevenMiller123 (changes requested):
Not going to judge the patching code, as that isn't something I'm particularly familiar with.

kernelstub:
Hi, I accidentaly pushed on my other acc that I didn't even use. I will fix and repush, sorry.

PR #4160, opened 2026-03-22
feat(renderer): add backend abstraction

### LINKED ISSUE #4149: [RFC]: Renderer Backend Abstraction
## TL;DR
This RFC proposes a first-class renderer backend abstraction so that Vulkan isn't the only hard-coded host rendering path. The immediate goal is to land a stable contract between the common PS4 GPU emulation code and backend-specific host rendering code, with Vulkan remaining as the primary implementation (+ a small headless backend until a proper one is introduced).

## Summary
At the moment, shadPS4 is effectively organized around a Vulkan-only renderer path. That's worked well for getting the emulator to a performant state on x86_64 Linux and Windows hosts, but it makes it very difficult to have good support for Apple Silicon (ARM64) which primarily benefits from Metal rendering.

This RFC proposes a renderer abstraction layer that defines an explicit host rendering contract for common code, keeps backend-private details private instead of leaking Vulkan types outward, allows runtime backend selection, adds a headless backend for testing and CI, and preserves current Vulkan behaviour as the default path.

Image

## Motivation

Already made largely clear in the summary.

### 1. Introduce a renderer contract layer

Add a neutral renderer namespace under `src/video_core/renderer/` that becomes the only supported interface between common emulation code and host rendering code.

The core contract is:

- `IRenderBackend`
- `IRasterizer`
- `IPresenter`
- `IFrameHandle`
- `IShaderDebugProvider`
- `BackendKind`
- `BackendCapabilities`
- `BackendCreateInfo`

This layer should express emulator-facing behaviour (NOT low-level host API objects). Common code should ask for operations such as preparing or presenting a frame, submitting draw or dispatch work, registering video-out surfaces, querying display settings, reporting capabilities, or retrieving backend-specific shader debug data through an abstract provider. It should NOT directly know about things like `vk::*` types, Vulkan swapchain objects, Vulkan command scheduling internals, Vulkan frame resources, or Vulkan-owned caches and pipeline state. 

### 2. Keep the abstraction optimized for explicit APIs

This abstraction should be designed for modern explicit backends first, which means command submission, synchronization, pipeline compilation, and host resource caches remain strictly backend-owned, and common code doesn't force a least-common-denominator API shape. This is going to be important because a premature "generic GPU object" layer would likely make the code worse. The contract should abstract ownership and behavior and not pretened Vulkan, Metal, DSD12, and future backends all share the same low-level resource model.

### 3. Move common code off direct Vulkan ownership
Subsystems such as Liverpool, GNM driver, video-out, page management, memory invalidation hooks, and shader tooling should depend only on the renderer contract, which means replacing `Vulkan::Presenter` and `Vulkan::Rasterizer` usage in common code, replacing backend-owened frame pointers with opaque frame handles, and routing backend creation through a single factory. 

### 4. Keep Vulkan as the first backend
Because it's the most tested backend so far, it should remain the default even on Apple Silicon, until we can make sure the Metal backend is as stable as the Vulkan one.

### 5. A Headless Backend
We'll include a headless backend that implements the same contract without rendering through a host graphics API. It won't be exposed to the user for obvious reasons, but it should support opaque frame handles, presentation no-ops that still preserve the lifecycle semantics, rasterizer callbacks that record or count ops, and mapped-range tracking needed by page and memory systems.

### 6. Support runtime backend selection
Can be exposed through CLI and the GUI launcher. As mention in `4.`, Vulkan will remain the default unless explicitly changed by the user. In the future, depending on backend/game compatibility, perhaps it could automatically switch to the most optimized backend, but that's for a future discussion.

### Ownership model

The ownership model is very simple; one backend object owns backend-global renderer state, one presenter object owns presentation behavior, rasterizer object owns draw and dispatch, frame handles are opaque and move-only, common systems keep interface references and not backend types, and so on.

### Backend-private code stays private
The following classes of code should remain backend-private unless there's a strong reason to extract truly shared logic:

- pipeline caches,
- shader compilers and backend code generators,
- host resource wrappers,
- buffer and image allocators,
- descriptor management,
- queue and scheduling logic,
- presentation internals,
- format translation helpers,
- and backend-specific synchronization code.

If a helper embeds Vulkan types or assumptions, it belongs on the Vulkan side of the boundary.

### Shader compilation boundary
The shader recompiler should stay split between common frontend, IR, optimization passes, and backend-specific code emission and backend-specific shader packaging. SPIR-V generation is Vulkan-only, so it shouldn't stay the universal renderer contract.

## Feedback period
1 week(s)

## CC List
...

## Any Other Things
...

AlpinDale:
Closes #4149 (RFC)

### DIFF
--- a/src/core/devtools/layer.cpp
+++ b/src/core/devtools/layer.cpp
@@ -15,14 +15,12 @@
 #include "imgui/imgui_std.h"
 #include "imgui_internal.h"
 #include "options.h"
-#include "video_core/renderer_vulkan/vk_presenter.h"
+#include "video_core/renderer/backend_factory.h"
 #include "widget/frame_dump.h"
 #include "widget/frame_graph.h"
 #include "widget/memory_map.h"
 #include "widget/module_list.h"
 #include "widget/shader_list.h"
-
-extern std::unique_ptr<Vulkan::Presenter> presenter;
 
 using namespace ImGui;
 using namespace ::Core::Devtools;
@@ -90,29 +88,33 @@ void L::DrawMenuBar() {
             ImGui::EndMenu();
         }
         if (BeginMenu("Display")) {
-            auto& pp_settings = presenter->GetPPSettingsRef();
+            auto& presenter = VideoCore::Render::GetRenderBackend().GetPresenter();
+            auto display_settings = presenter.GetDisplaySettings();
             if (BeginMenu("Brightness")) {
-                SliderFloat("Gamma", &pp_settings.gamma, 0.1f, 2.0f);
+                SliderFloat("Gamma", &display_settings.gamma, 0.1f, 2.0f);
+                presenter.SetDisplaySettings(display_settings);
                 ImGui::EndMenu();
             }
             if (BeginMenu("FSR")) {
-                auto& fsr = presenter->GetFsrSettingsRef();
-                Checkbox("FSR Enabled", &fsr.enable);
-                BeginDisabled(!fsr.enable);
+                Checkbox("FSR Enabled", &display_settings.fsr_enabled);
+                BeginDisabled(!display_settings.fsr_enabled);
                 {
-                    Checkbox("RCAS", &fsr.use_rcas);
-                    BeginDisabled(!fsr.use_rcas);
+                    Checkbox("RCAS", &display_settings.rcas_enabled);
+                    BeginDisabled(!display_settings.rcas_enabled);
                     {
-                        SliderFloat("RCAS Attenuation", &fsr.rcas_attenuation, 0.0, 3.0);
+                        SliderFloat("RCAS Attenuation", &display_settings.rcas_attenuation, 0.0,
+                                    3.0);
                     }
                     EndDisabled();
                 }
                 EndDisabled();
+                presenter.SetDisplaySettings(display_settings);
 
                 if (Button("Save")) {
-                    Config::setFsrEnabled(fsr.enable);
-                    Config::setRcasEnabled(fsr.use_rcas);
-                    Config::setRcasAttenuation(static_cast<int>(fsr.rcas_attenuation * 1000));
+                    Config::setFsrEnabled(display_settings.fsr_enabled);
+                    Config::setRcasEnabled(display_settings.rcas_enabled);
+                    Config::setRcasAttenuation(
+                        static_cast<int>(display_settings.rcas_attenuation * 1000));
                     Config::save(Common::FS::GetUserPath(Common::FS::PathType::UserDir) /
                                  "config.toml");
                     CloseCurrentPopup();
--- a/src/core/devtools/widget/shader_list.cpp
+++ b/src/core/devtools/widget/shader_list.cpp
@@ -15,10 +15,7 @@
 #include "core/devtools/options.h"
 #include "imgui/imgui_std.h"
 #include "sdl_window.h"
-#include "video_core/renderer_vulkan/vk_presenter.h"
-#include "video_core/renderer_vulkan/vk_rasterizer.h"
-
-extern std::unique_ptr<Vulkan::Presenter> presenter;
+#include "video_core/renderer/backend_factory.h"
 
 using namespace ImGui;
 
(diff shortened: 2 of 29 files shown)


georgemoralis:
Need rebase

src/video_core/renderer_vulkan/vulkan_backend.h
```diff
@@ -0,0 +1,50 @@
...
+
+private:
+    class FrameHandle;
+    class PresenterProxy;
+    class RasterizerProxy;
+    class ShaderDebugProvider;
+
+    std::unique_ptr<Presenter> presenter;
+    std::unique_ptr<PresenterProxy> presenter_proxy;
+    std::unique_ptr<RasterizerProxy> rasterizer_proxy;
```
Niram7777, line 45: Why use *Proxy instead of just Presenter/Rasterizer ?

src/core/devtools/widget/shader_list.cpp
```diff
@@ -31,12 +28,16 @@ ShaderList::Selection::Selection(int index)
     isa_editor->SetReadOnly(true);
     glsl_editor->SetPalette(TextEditor::GetDarkPalette());
     glsl_editor->SetLanguageDefinition(TextEditor::LanguageDefinition::GLSL());
-    presenter->GetWindow().RequestKeyboard();
+    if (auto* backend = VideoCore::Render::TryGetRenderBackend()) {
+        backend->GetPresenter().GetWindow().RequestKeyboard();
```
Niram7777, line 32: might as well use one line with GetRenderBackend()

src/core/devtools/widget/shader_list.cpp
```diff
@@ -31,12 +28,16 @@ ShaderList::Selection::Selection(int index)
...
+    }
 }
 
 ShaderList::Selection::~Selection() {
     if (index >= 0) {
-        presenter->GetWindow().ReleaseKeyboard();
+        if (auto* backend = VideoCore::Render::TryGetRenderBackend()) {
+            backend->GetPresenter().GetWindow().ReleaseKeyboard();
```
Niram7777, line 39: same

src/core/devtools/widget/shader_list.cpp
```diff
@@ -58,9 +59,19 @@ void ShaderList::Selection::ReloadShader(DebugStateType::ShaderDump& value) {
...
+    }
+    auto* debug_provider = backend->GetShaderDebugProvider();
+    if (debug_provider == nullptr) {
+        return;
+    }
+    const auto module_handle =
+        reinterpret_cast<uint64_t>(static_cast<VkShaderModule>(value.module));
+    if (const auto replacement = debug_provider->ReplaceShader(module_handle, spv); replacement) {
+        value.module = vk::ShaderModule{
+            reinterpret_cast<VkShaderModule>(static_cast<uintptr_t>(*replacement))};
```
Niram7777, line 74: might be cleaner to implement a ReloadShader in the Rasterizer (to avoid the cast , extra returns)

src/core/devtools/layer.cpp
```diff
@@ -90,31 +88,35 @@ void L::DrawMenuBar() {
             ImGui::EndMenu();
         }
         if (BeginMenu("Display")) {
-            auto& pp_settings = presenter->GetPPSettingsRef();
+            auto& presenter = VideoCore::Render::GetRenderBackend().GetPresenter();
+            auto display_settings = presenter.GetDisplaySettings();
```
Niram7777, line 92: you could return by reference, to not to have to call SetDisplaySettings everytime (same for ipc.cpp)

src/core/libraries/videoout/video_out.cpp
```diff
@@ -362,7 +360,8 @@ s32 sceVideoOutSubmitEopFlip(s32 handle, u32 buf_id, u32 mode, s64 flip_arg, voi
 s32 PS4_SYSV_ABI sceVideoOutGetDeviceCapabilityInfo(
     s32 handle, SceVideoOutDeviceCapabilityInfo* pDeviceCapabilityInfo) {
     pDeviceCapabilityInfo->capability = 0;
-    if (presenter->IsHDRSupported()) {
+    auto* backend = VideoCore::Render::TryGetRenderBackend();
+    if (backend != nullptr && backend->GetPresenter().IsHDRSupported()) {
```
Niram7777, line 364: one line with GetRenderBackend() (instead of silenty skip)

Niram7777 (commented):
I really like how easy currently the codebase is with only Vulkan, both when browsing the code or debugging.
I am not sure the added complexity is going to provide much benefits, but maybe you have more details on what would be the next step.
> it very difficult to have good support for Apple Silicon (ARM64) which primarily benefits from Metal rendering.

I am not sure someone would contribute significantly for Metal to the point it is better than Vulkan.

georgemoralis:
this seems to be outdated , need to be rebased if it still on works

PR #4164, opened 2026-03-23
General Misses (0.15.1 WIP part 5)

georgemoralis:


src/core/libraries/kernel/memory.cpp
```diff
@@ -318,6 +318,38 @@ s32 PS4_SYSV_ABI sceKernelMapFlexibleMemory(void** addr_in_out, u64 len, s32 pro
     return sceKernelMapNamedFlexibleMemory(addr_in_out, len, prot, flags, "anon");
 }
 
+s32 PS4_SYSV_ABI sceKernelMapNamedSystemFlexibleMemory(void** addr_in_out, u64 len, s32 prot,
```
StevenMiller123, line 321: I forgot to remove this? 😓

georgemoralis: isn't needed for move tracker lle?

StevenMiller123: It's needed for working malloc LLEs, but I'm pretty sure I reverted the malloc_init LLE stuff already.
MoveTracker LLE can't happen without malloc LLE or HLE.

georgemoralis: well pr a removal then :P , also i think sysmodules also has lle movetracker in it ;p

kalaposfos13: We aren't LLEing MoveTracker anymore, so that isn't relevant.

PR #4165, opened 2026-03-23
 General Misses (0.15.1 WIP part 6)

georgemoralis:
- Fixed a bug in settings loading
- Removed sceKernelMapNamedSystemFlexibleMemory (was a test code)
- Removed libScePadTracker.sprx loading


StevenMiller123:
PadTracker is still safe if I remember right.

kalaposfos13:
PadTracker != MoveTracker, although it's not like removing this too changes much, as while this works as in loads and starts up, it doesn't work as in it cannot actually track anything.

georgemoralis:
well we can re-add it when it is somehow useful

PR #4167, opened 2026-03-24
Lib.Vdec2: Cleaner fix for updated structs in older games

StevenMiller123:
Tested a couple older games in my library. I don't have anything that works well enough and uses the newer struct revisions, so this will need some testing.


yaya54840:
Hello StevenMiller,
 we now have videos of EA Sports UFC and Dead or Alive Xtreme 3.

StevenMiller123:
Both of those are too old to help if I recall correctly.

PR #4169, opened 2026-03-24
Local multiplayer support

kalaposfos13:
This PR started off as its own thing, found its way into the combined 0.15.1 WIP branch, and now it is a separate thing again. Functionally, it is the exact same as before, but just in case someone needs a refresher on what "exact same" entails:
- The new user management backend has been hooked up to the input engine, which now supports 4\* simultaneous virtual controllers.
- Users are managed by the users.json file in the user folder, where you can set names, player indexes, and controller colours for each user.
- Controllers should just work out of the box, however it is also possible to bind keyboard or any other inputs to a desired virtual controller output, with the `<output> : <controller index> = <input>` syntax (example: `cross : 2 = u`
- It is not possible at the moment to match controllers to users, the controller list will be populated in the order the OS presents them to the emulator.
- Special input methods like mouse-to-joystick emulation only work for the first player, however, if you ever find yourself in a place where you'd attempt to bind the mouse to player 2, you should just step back and reconsider your life choices.

\* technically 5, but the fifth one is a placeholder controller, used exclusively for a dummy TV remote input, solely because some games will misbehave if they can't open a handle for the aforementioned device.

### DIFF
--- a/src/core/libraries/pad/pad.cpp
+++ b/src/core/libraries/pad/pad.cpp
@@ -431,7 +431,7 @@ int PS4_SYSV_ABI scePadRead(s32 handle, OrbisPadData* pData, s32 num) {
     if (!controller_id) {
         return ORBIS_PAD_ERROR_INVALID_HANDLE;
     }
-    auto controllers = *Common::Singleton<GameControllers>::Instance();
+    auto& controllers = *Common::Singleton<GameControllers>::Instance();
     auto& controller = *controllers[*controller_id];
     int ret_num = controller.ReadStates(states.data(), num, &connected, &connected_count);
     return ProcessStates(handle, pData, controller, states.data(), ret_num, connected,
@@ -464,7 +464,7 @@ int PS4_SYSV_ABI scePadReadState(s32 handle, OrbisPadData* pData) {
     if (!controller_id) {
         return ORBIS_PAD_ERROR_INVALID_HANDLE;
     }
-    auto controllers = *Common::Singleton<GameControllers>::Instance();
+    auto& controllers = *Common::Singleton<GameControllers>::Instance();
     auto& controller = *controllers[*controller_id];
     int connected_count = 0;
     bool connected = false;
@@ -485,7 +485,7 @@ int PS4_SYSV_ABI scePadResetLightBar(s32 handle) {
     if (!controller_id) {
         return ORBIS_PAD_ERROR_INVALID_HANDLE;
     }
-    auto controllers = *Common::Singleton<GameControllers>::Instance();
+    auto& controllers = *Common::Singleton<GameControllers>::Instance();
     s32 colour_index = UserManagement.GetUserByPlayerIndex(handle)->user_color - 1;
     Input::Colour colour{255, 0, 0};
     if (colour_index >= 0 && colour_index <= 3) {
@@ -522,7 +522,7 @@ int PS4_SYSV_ABI scePadResetOrientation(s32 handle) {
         return ORBIS_PAD_ERROR_INVALID_HANDLE;
     }
 
-    auto controllers = *Common::Singleton<GameControllers>::Instance();
+    auto& controllers = *Common::Singleton<GameControllers>::Instance();
     Libraries::Pad::OrbisFQuaternion defaultOrientation = {0.0f, 0.0f, 0.0f, 1.0f};
     controllers[*controller_id]->SetLastOrientation(defaultOrientation);
     controllers[*controller_id]->SetLastUpdate(std::chrono::steady_clock::now());
@@ -587,7 +587,7 @@ int PS4_SYSV_ABI scePadSetLightBar(s32 handle, const OrbisPadLightBarParam* pPar
             return ORBIS_PAD_ERROR_INVALID_LIGHTBAR_SETTING;
         }
 
-        auto controllers = *Common::Singleton<GameControllers>::Instance();
+        auto& controllers = *Common::Singleton<GameControllers>::Instance();
         controllers[*controller_id]->SetLightBarRGB(pParam->r, pParam->g, pParam->b);
         return ORBIS_OK;
     }
@@ -610,7 +610,7 @@ int PS4_SYSV_ABI scePadSetLightBarForTracker(s32 handle, const OrbisPadLightBarP
     if (!controller_id) {
         return ORBIS_PAD_ERROR_INVALID_HANDLE;
     }
-    auto controllers = *Common::Singleton<GameControllers>::Instance();
+    auto& controllers = *Common::Singleton<GameControllers>::Instance();
     controllers[*controller_id]->SetLightBarRGB(pParam->r, pParam->g, pParam->b);
     return ORBIS_OK;
 }
@@ -665,7 +665,7 @@ int PS4_SYSV_ABI scePadSetVibration(s32 handle, const OrbisPadVibrationParam* pP
     if (pParam != nullptr) {
         LOG_DEBUG(Lib_Pad, "scePadSetVibration called handle = {} data = {} , {}", handle,
                   pParam->smallMotor, pParam->largeMotor);
-        auto controllers = *Common::Singleton<GameControllers>::Instance();
+        auto& controllers = *Common::Singleton<GameControllers>::Instance();
         controllers[*controller_id]->SetVibration(pParam->smallMotor, pParam->largeMotor);
         return ORBIS_OK;
     }
--- a/src/core/libraries/pad/pad.cpp
+++ b/src/core/libraries/pad/pad.cpp
@@ -282,11 +282,14 @@ int PS4_SYSV_ABI scePadOpen(Libraries::UserService::OrbisUserServiceUserId userI
         }
         return ORBIS_DEVICE_SERVICE_ERROR_INVALID_USER;
     }
+    if (type == ORBIS_PAD_PORT_TYPE_REMOTE_CONTROL) {
+        return ORBIS_PAD_ERROR_INVALID_ARG;
+    }
     if (EmulatorSettings.IsUsingSpecialPad()) {
         if (type != ORBIS_PAD_PORT_TYPE_SPECIAL)
             return ORBIS_PAD_ERROR_DEVICE_NOT_CONNECTED;
     } else {
-        if (type != ORBIS_PAD_PORT_TYPE_STANDARD && type != ORBIS_PAD_PORT_TYPE_REMOTE_CONTROL)
+        if (type != ORBIS_PAD_PORT_TYPE_STANDARD)
             return ORBIS_PAD_ERROR_DEVICE_NOT_CONNECTED;
     }
     auto u = UserManagement.GetUserByID(userId);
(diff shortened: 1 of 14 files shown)


src/core/libraries/pad/pad.cpp
```diff
@@ -159,17 +161,33 @@ int PS4_SYSV_ABI scePadGetHandle(Libraries::UserService::OrbisUserServiceUserId
...
         return ORBIS_PAD_ERROR_DEVICE_NO_HANDLE;
     }
-    LOG_DEBUG(Lib_Pad, "(DUMMY) called");
-    return 1;
+    auto const u = UserManagement.GetUserByID(userId);
+    if (!u) {
+        return ORBIS_PAD_ERROR_DEVICE_NO_HANDLE;
+    }
+    s32 pad_handle = u->player_index;
+    LOG_DEBUG(Lib_Pad, "called, userid: {}, out pad handle: ", userId, pad_handle);
```
StevenMiller123: Pretty sure this format string is broken? I don't think the pad handle will log properly.

src/core/libraries/pad/pad.cpp
```diff
@@ -254,34 +272,47 @@ int PS4_SYSV_ABI scePadOpen(Libraries::UserService::OrbisUserServiceUserId userI
...
+            return 5;
+        }
+        return ORBIS_DEVICE_SERVICE_ERROR_INVALID_USER;
     }
-    if (Config::getUseSpecialPad()) {
+    if (EmulatorSettings.IsUsingSpecialPad()) {
         if (type != ORBIS_PAD_PORT_TYPE_SPECIAL)
             return ORBIS_PAD_ERROR_DEVICE_NOT_CONNECTED;
     } else {
         if (type != ORBIS_PAD_PORT_TYPE_STANDARD && type != ORBIS_PAD_PORT_TYPE_REMOTE_CONTROL)
```
StevenMiller123: Now that we're checking separately for it, `ORBIS_PAD_PORT_TYPE_REMOTE_CONTROL` should probably return `ORBIS_PAD_ERROR_INVALID_ARG` if it doesn't hit the earlier error return.

src/core/libraries/pad/pad.cpp
```diff
@@ -254,34 +272,47 @@ int PS4_SYSV_ABI scePadOpen(Libraries::UserService::OrbisUserServiceUserId userI
...
             return ORBIS_PAD_ERROR_DEVICE_NOT_CONNECTED;
     }
-    return 1; // dummy
+    return userId; // dummy
```
StevenMiller123: Is there a reason we're keeping the userId dummy return here? Now that we're doing proper handle checks, I imagine this could break some things?

src/core/libraries/pad/pad.cpp
StevenMiller123: Minor style nit, but I don't believe there's a point in keeping this extra handle check now that there's proper logic?

src/core/libraries/pad/pad.cpp
```diff
@@ -451,12 +484,26 @@ int PS4_SYSV_ABI scePadReadStateExt() {
 
 int PS4_SYSV_ABI scePadResetLightBar(s32 handle) {
     LOG_INFO(Lib_Pad, "(DUMMY) called");
```
StevenMiller123: I'll only leave this one comment about it, but I think at this point we should consider downgrading some of these pad logs to debug or trace, since these functions are properly implemented now.

src/core/libraries/pad/pad.cpp
```diff
@@ -555,8 +607,10 @@ int PS4_SYSV_ABI scePadSetLightBarBlinking() {
     return ORBIS_OK;
 }
 
-int PS4_SYSV_ABI scePadSetLightBarForTracker() {
-    LOG_ERROR(Lib_Pad, "(STUBBED) called");
+int PS4_SYSV_ABI scePadSetLightBarForTracker(s32 handle, const OrbisPadLightBarParam* pParam) {
```
StevenMiller123, line 626: Shouldn't there be some handle validation here?

src/input/input_handler.cpp
```diff
@@ -278,21 +396,39 @@ void ParseInputConfig(const std::string game_id = "") {
 
         std::string output_string = line.substr(0, equal_pos);
         std::string input_string = line.substr(equal_pos + 1);
-        // Remove trailing semicolon from input_string
-        if (!input_string.empty() && input_string[input_string.length() - 1] == ';' &&
-            input_string != ";") {
-            line = line.substr(0, line.length() - 1);
+        s8 input_gamepad_id = -1, output_gamepad_id = -1; // -1 means it's not specified
+
+        // todo: here the inputs and outputs are formatted and split, we need to extract the
```
StevenMiller123: I see a lot of todo's down here, are you sure this shouldn't be a draft?

StevenMiller123: I see a lot of todo's down here, are you sure this shouldn't be a draft?

StevenMiller123: I see a lot of todo's down here, are you sure this shouldn't be a draft?

StevenMiller123 (changes requested):
A couple nitpicks from looking over the code, I think most of this looks functional though.

StevenMiller123:
Ignore GitHub duplicating one of my comments 😅

src/core/libraries/pad/pad.cpp
kalaposfos13: I'll write a quick homebrew test for this tomorrow, unless you bother turning on your own console just to check which is the correct error code here.

src/core/libraries/pad/pad.cpp
kalaposfos13: > Is there a reason we're keeping the userId dummy return here?

This is a leftover from when handles were just userids, and I just forgot to update this function in particular after the refactor. Also, the whole EmulatorSettings.IsUsingSpecialPad() setup assumes only one controller in the first place, so this whole function and config setting will have to be refactored.

src/core/libraries/pad/pad.cpp
kalaposfos13: Yup, this was already removed in ReadState, I just missed this place.

src/input/input_handler.cpp
kalaposfos13: Looks like most of these are already resolved, and I just forgot to remove the comments, but yes, this is a valid observation, and it still does apply in one place.

src/core/libraries/pad/pad.cpp
StevenMiller123: The error check for this is present in the actual libScePad library, shouldn't warrant any testing in my opinion.
image

kalaposfos13: Well, can't really argue with that, feel free to push this if you like, or alternatively wait for me to do it tomorrow.

src/core/libraries/pad/pad.cpp
```diff
@@ -254,34 +272,47 @@ int PS4_SYSV_ABI scePadOpen(Libraries::UserService::OrbisUserServiceUserId userI
...
     }
-    LOG_INFO(Lib_Pad, "(DUMMY) called user_id = {} type = {} index = {}", userId, type, index);
+    auto u = UserManagement.GetUserByID(userId);
+    if (!u) {
+        return ORBIS_DEVICE_SERVICE_ERROR_USER_NOT_LOGIN;
+    }
+    s32 pad_handle = u->player_index;
+    LOG_INFO(Lib_Pad, "(DUMMY) called user_id = {} type = {} index = {}, pad_handle = {}", userId,
+             type, index, pad_handle);
     g_opened = true;
```
roamic: I think this should be per user.
Edit: or rather per handle.

src/core/libraries/pad/pad.cpp
```diff
@@ -254,34 +272,47 @@ int PS4_SYSV_ABI scePadOpen(Libraries::UserService::OrbisUserServiceUserId userI
...
     }
-    if (userId == -1) {
-        return ORBIS_PAD_ERROR_DEVICE_NO_HANDLE;
+    if (userId < 0) {
+        return ORBIS_DEVICE_SERVICE_ERROR_INVALID_USER;
+    }
+    if (userId == Libraries::UserService::ORBIS_USER_SERVICE_USER_ID_SYSTEM) {
+        if (type == ORBIS_PAD_PORT_TYPE_REMOTE_CONTROL) {
+            LOG_INFO(Lib_Pad, "Opened a TV remote device");
+            return 5;
```
roamic: `scePadGetHandle` should also return this handle if it's opened. The way it's done in the original library is they keep a map of opened handles where key is a tuple of `(userId, type, index)`. You can also make a global map like this to store the pad state and check if it's open.

src/core/libraries/pad/pad.cpp
```diff
@@ -538,8 +587,8 @@ int PS4_SYSV_ABI scePadSetLightBar(s32 handle, const OrbisPadLightBarParam* pPar
             return ORBIS_PAD_ERROR_INVALID_LIGHTBAR_SETTING;
         }
 
-        auto* controller = Common::Singleton<GameController>::Instance();
-        controller->SetLightBarRGB(pParam->r, pParam->g, pParam->b);
+        auto controllers = *Common::Singleton<GameControllers>::Instance();
```
roamic: Shouldn't this be a reference? If I understand correctly you're making a local copy every time. This issue is present in every call.

src/core/libraries/pad/pad.cpp
kalaposfos13: For now, I've made it so the setting sets either all or none of the controllers into special mode, it doesn't matter much anyway until said special pads are handled correctly.

StevenMiller123 (approved):
Looks good to me, wouldn't hurt to have others take a look though.

StevenMiller123:
Did a quick local test, local multiplayer in KINGDOM HEARTS Melody of Memory DEMO Version functions properly, with my controller as player one and my keyboard as player two.

image
image

yaya54840:
Hi kalafospos13
It works for the CUSA49379 Lego Voyager game which requires a second player.
I modified my user.json file, but I don't know if it's really necessary.

[shad_log.txt](https://github.com/user-attachments/files/26299197/shad_log.txt)

Capture d&#39;écran 2026-03-27 093538


[users.json](https://github.com/user-attachments/files/26299214/users.json)

kalaposfos13:
@yaya54840 your config only works because it is recognized as invalid, and the emulator falls back to the default configuration. I recommend moving/deleting it, letting the emulator regenerate it, and taking a look at the correct format.

yaya54840:
Okay kalaposfos13, I did as you said.
It works and it also allows me to launch the game, which was stuck on the menu.
Capture d&#39;écran 2026-03-27 105616

Thank you for the work you do.

LuismaSP89:
Everything working as expected, but not the R2 button, all the buttons are working but not this one in both controllers tested. Tested in resident evil 6 (Shoot button) and Guilty Gear XRD rev 2.

gandalfthewhite19890404:
Diablo III is working
image

gandalfthewhite19890404:
Can you prevent Player 1 removing? Some games (at least Diablo III) will hang

kalaposfos13:
> Can you prevent Player 1 removing? Some games (at least Diablo III) will hang

I have a strong suspicion that might also be the case on real hardware as well, and is Not My Bug™, plus from my own testing, games such as Minecraft will just throw up a popup saying "connect a controller to continue playing" instead. While the feature is indeed mostly useless, I don't really see the point in limiting this functionality anyway.

PR #4175, opened 2026-03-26
added transfer of sysmodules path and fonts path

georgemoralis:


StevenMiller123 (changes requested):
Gave it a quick test, font dir doesn't transfer.

StevenMiller123 (approved):
Scratch that, I got confused by my own config. Transferring options works fine for both paths.

PR #4187, opened 2026-03-29
translations: use cpp file for function

rainmakerv3:
Kinda not ideal to use a static header function (that I forgot to declare as static lol) instead of a cpp, so decided not to be lazy


rainmakerv3:
Screenshot of it working with the other trophy changes (temporarily used AI translation, won't be in this PR):

image

PR #4190, opened 2026-03-31
shader_recompiler: Generalize V_CMP_U64 to support arbitrary operands

martyniukyurii:
## Summary

- Extends `V_CMP_U64` to handle arbitrary operand types instead of asserting on non-constant operands
- Adds `ExecLo` support to both the thread-bit fast path in `V_CMP_U64` and `GetSrc64` in `translate.cpp`
- Adds a general 64-bit integer comparison fallback path using `GetSrc64<IR::U64>` with full `ConditionOp` support (`EQ`, `NE`, `GT`, `LT`, `LE`, `GE`, `F`, `TRU`)

## Problem

The previous `V_CMP_U64` implementation only supported `ConstZero` and `SignedConstIntNeg(-1)` as `src[1]`, and only `ScalarGPR`/`VccLo` as `src[0]`. Any other operand type triggered an assertion failure:

```
[Debug] <Critical> vector_alu.cpp:1141 operator(): Assertion Failed!
```

This was followed by a second crash in `GetSrc64` when `ExecLo` was used as an operand:

```
[Debug] <Critical> translate.cpp:446 GetSrc64: Unreachable code!
```

These crashes occurred during compute shader compilation in **Kona II: Brume** (CUSA19562).

### `vector_alu.cpp` — `V_CMP_U64`
- Thread-bit fast path now also handles `ExecLo` (via `ir.GetExec()`)
- Added general fallback: loads both operands as `IR::U64` via `GetSrc64` and performs standard integer comparison
- Removed `UNREACHABLE` guards for `is_signed` and `set_exec` — these are now properly handled

### `translate.cpp` — `GetSrc64`
- Added `OperandField::ExecLo` case: converts per-thread exec bit to a 64-bit value (`0xFFFFFFFFFFFFFFFF` if active, `0` otherwise) using `Select` + `PackUint2x32`

## Test plan

- [x] Tested with **Kona II: Brume** (CUSA19562) — game now boots past shader compilation and reaches the main menu (previously crashed immediately)
- [ ] Verify no regressions in other games that use `V_CMP_EQ_U64`, `V_CMP_NE_U64`, `V_CMP_GT_U64`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/shader_recompiler/frontend/translate/translate.cpp
+++ b/src/shader_recompiler/frontend/translate/translate.cpp
@@ -442,6 +442,20 @@ T Translator::GetSrc64(const InstOperand& operand) {
         }
         break;
     case OperandField::VccHi:
+        UNREACHABLE();
+    case OperandField::ExecLo: {
+        const auto exec_bit = ir.GetExec();
+        if constexpr (is_float) {
+            UNREACHABLE();
+        } else {
+            const auto ones32 = ir.Imm32(~u32{0});
+            const auto zero32 = ir.Imm32(u32{0});
+            const auto lo = IR::U32(ir.Select(exec_bit, ones32, zero32));
+            const auto hi = IR::U32(ir.Select(exec_bit, ones32, zero32));
+            value = ir.PackUint2x32(ir.CompositeConstruct(lo, hi));
+        }
+        break;
+    }
     default:
         UNREACHABLE();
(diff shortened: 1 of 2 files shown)


georgemoralis:
clearly AI generated , i don't bother try to read it

PR #4202, opened 2026-03-31
VOP3P instructions definitions

mikusp:
VOP3P instructions are introduced in Neo mode

### DIFF
--- a/src/shader_recompiler/frontend/decode.cpp
+++ b/src/shader_recompiler/frontend/decode.cpp
@@ -3,6 +3,7 @@
 
 #include <algorithm>
 #include "common/assert.h"
+#include "core/libraries/kernel/process.h"
 #include "shader_recompiler/frontend/decode.h"
 
 #include <magic_enum/magic_enum.hpp>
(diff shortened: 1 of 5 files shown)


squidbus (commented):
General comment: do we actually need to gate with `sceKernelIsNeoMode` calls? I don't think we need to worry about regular mode games trying to use these (and if they did the behavior here would be to fail anyway).

mikusp:
> General comment: do we actually need to gate with `sceKernelIsNeoMode` calls? I don't think we need to worry about regular mode games trying to use these (and if they did the behavior here would be to fail anyway).

Not really, but I guess just for correctness I've put them in place

StevenMiller123:
I mean, correctness-wise, it would be better to check HasNeoMode instead (so the config option directly). Though I agree that the checks aren't really necessary.

mikusp:
Removed the checks as requested

squidbus (approved):
Looks good to me. You're not wrong on correctness but for the recompiler code it seems better to me to not introduce more cross-dependency on kernel calls for something that would be undefined behavior and probably crash on base PS4 hardware anyway.

PR #4203, opened 2026-03-31
SDWA

mikusp:
SDWA (sub-dword addressing) is available for VOP1/2/C instructions in Neo mode. This PR parses the information but does not include them in source fetching for affected instructions


StevenMiller123 (commented):
Following the same logic as the VOP3P PR, probably don't need these IsNeoMode checks. If you want to keep them, it would be more accurate to check against HasNeoMode instead.

mikusp:
Updated with removed Neo checks and proper operand fields for non-vgpr case

PR #4204, opened 2026-04-01
VUID-vkCmdDraw-None-06887

Kravickas:
Implement split layout that keeps depth read-only while allowing stencil writes. Vulkan happy.

NHLs, MGSV and ton of other games.


georgemoralis:
does it fix a specific issue? or have better visual output?

PR #4209, opened 2026-04-02
[CI] Use VS 2026

Xphalnos:
Updating CI runner for using VS 2026 because it’s the latest stable release (newer STL + Windows SDK).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -143,7 +143,7 @@ jobs:
         shell: bash
 
   windows-sdl:
-    runs-on: windows-2025
+    runs-on: windows-2025-vs2026
     needs: get-info
     steps:
     - uses: actions/checkout@v6


Niram7777:
it will use VS 2026 automatically next month but we could put same labels for app and test (that you forgot) builds
```
As previously announced, the "windows-latest" and “windows-2025” labels in GitHub Actions will be migrated to use Visual Studio 2026 by default.

This change will be rolled out over a week beginning June 8, 2026 and will complete by June 15, 2026.

During this period your workflows will gradually switch over to the new image with Visual Studio 2026 as the default. If you want to do advance testing prior to the migration, you can follow the instructions below.

[IMPORTANT] The switch to Visual Studio 2026 may break your workflows if you depend on Visual Studio 2022.

What you need to do:

You do not need to do anything at this time if you want your workflows to migrate to the latest Visual Studio version.

If you want to test the new image:


update the runs-on: target in your YAML workflow file to windows-2025-vs2026.
```

PR #4210, opened 2026-04-02
For OpImageFetch, do not provide both Lod and Sample operands

mikusp:
An attempt to fix a Vulkan validation error:
```
[Render.Vulkan] <Error> (shadPS4:GpuComm) vk_platform.cpp:57 DebugUtilsCallback: VUID-VkShaderModuleCreateInfo-pCode-08737: vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
Image Operand Lod requires 'MS' parameter to be 0
  %131 = OpImageFetch %f32vec4_id %130 %129 Lod|Sample %u32_id_0 %u32_id_0
```


georgemoralis:
turtle agreed

PR #4213, opened 2026-04-03
GS fixes

mikusp:
Figured out by @raphaelthegreat:
- handles S_BFM_B32 in a copy shader
- computes the number of dwords in a vertex by the number of loads in a copy shader, rather than VERT_ITEMSIZE set in the registers


StevenMiller123:
Seems to work properly in my titles, tried everything I could think of that uses geometry shaders and saw no regressions.

PR #4216, opened 2026-04-04
FS: easy mods folder for games

rainmakerv3:
Enables CUSAXXXXX-mods folder for modified game files that games will prefer to load over patch/base game folders if it exists, allowing for easy modding without touching the original dumped files.

### DIFF
--- a/src/core/file_sys/fs.cpp
+++ b/src/core/file_sys/fs.cpp
@@ -42,7 +42,7 @@ void MntPoints::UnmountAll() {
 }
 
 std::filesystem::path MntPoints::GetHostPath(std::string_view path, bool* is_read_only,
-                                             bool force_base_path) {
+                                             HostPathType path_type) {
     // Evil games like Turok2 pass double slashes e.g /app0//game.kpf
     std::string corrected_path(path);
     size_t pos = corrected_path.find("//");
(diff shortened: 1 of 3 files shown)


RuskyDed:
Can confirm it works perfectly and gets loaded into the right order. Thank you.

StevenMiller123 (changes requested):
`MntPoints::IterateDirectory` will need updates for mod files to show up appropriately in sceKernelGetdents/sceKernelGetdirentries calls.

StevenMiller123:
Rest of the logic looks fine, address that and I'll approve.

rainmakerv3:
> Rest of the logic looks fine, address that and I'll approve.

Ok done. Not sure how to test it with the scekernel call and all that, but seemed fairly straightforward so I just went ahead with it

src/core/file_sys/fs.cpp
```diff
@@ -84,13 +84,20 @@ std::filesystem::path MntPoints::GetHostPath(std::string_view path, bool* is_rea
     mods_path += "-mods";
     mods_path /= rel_path;
 
+    if (path_type == HostPathType::Mod) {
```
StevenMiller123, line 87: This if block should probably be removed in favor of hitting those "std::filesystem::exists" checks right after. Otherwise you'll be forcibly returning mod/patch paths for things that don't actually exist in the mod or patch folder.

StevenMiller123 (changes requested):
Slight issue with the new commit.

src/core/file_sys/fs.cpp
rainmakerv3: Might be missing something here, but the intent of passing these args was indeed to force passing these paths regardless of whether or not they exist (such as for pass 3 of iteratedirectories, which checks if both the -mods path and the -patch path are non-existent).

This does mean that filesystem::exists checks need to be made separately when these args are used, but at least for iteratedirectories it seems there needs to be a way to get these paths, existing or not

StevenMiller123: I see now. This will probably work fine, though I'll give it a quick test with UE titles to make sure.

StevenMiller123 (approved):
Seems to work, gave a quick test with a UE title and it detects mod contents properly.

PR #4219, opened 2026-04-04
Fix edge case for touch states

kalaposfos13:
Apparently, pressing down touch 1, then touch 2, then releasing touch 1 makes touch 2 appear as the first touch, not the second, which was previously not handled like that, simply because SDL's backend gave this data in a way that didn't switch touch indexes while a touch wasn't unpressed.
The PR was validated with hardware testing.


StevenMiller123 (approved):
Thanks for the quick fix 😄

PR #4228, opened 2026-04-06
CMake: support Visual Studio 2022 ClangCL generator

kayforbe:
## Summary

- Fix GAS assembly (`fiber_context.s`) compilation on Windows by using a custom build command for ClangCL, since the VS build system cannot assemble `.s` files directly
- **Ninja and other generators are completely unaffected** — the fix only activates when `CMAKE_GENERATOR` matches `"Visual Studio"`

## Why

This allows developers to use the Visual Studio 2022 IDE with full debugging support (breakpoints, IntelliSense, F5 to run) while still compiling with **Clang** via the ClangCL toolset. shadPS4's requirement for Clang is fully met.

## How to build (step by step)

**Prerequisites:** Visual Studio 2022 with **C++ Clang tools for Windows** component installed.

### Step 0
Open **Developer Command Prompt for VS 2022** (search for it in Start Menu).

### Step 1: Clone with submodules

```bash
git clone --recurse-submodules https://github.com/shadps4-emu/shadPS4.git
cd shadPS4
```

If you already cloned without `--recurse-submodules`, run this to download all dependencies (make sure you are in `cd shadPS4`):

```bash
git submodule update --init --recursive
```

> **Why?** shadPS4 depends on many external libraries (boost, fmt, ffmpeg, zlib, imgui, etc.) stored as git submodules in the `externals/` folder. Without this step, those folders will be empty and CMake will fail.

### Step 2: Generate the Visual Studio solution

In Developer Command Prompt for VS 2022 (make sure you are in `cd shadPS4`):

```bash
cmake -S . -B build -G "Visual Studio 17 2022" -T ClangCL -A x64
```

### Step 3
Open `build/shadPS4.sln` in Visual Studio 2022.

### Step 4
In Solution Explorer, right-click on **shadps4** and select **"Set as Startup Project"**.

### Step 5
Select your build configuration (e.g. **Release | x64**) from the toolbar, then press **Ctrl+B** or go to **Build > Build Solution**.

### Step 6
Select **Local Windows Debugger** or press **F5** to start the application.

## What changed in CMakeLists.txt

When `WIN32 AND Clang AND Visual Studio generator`:
1. `fiber_context.s` is assembled via `add_custom_command` using Clang directly
2. The resulting `.obj` is linked into `shadps4` via `target_link_libraries`

Otherwise (Ninja, Make, etc.): behavior is identical to before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -575,13 +575,32 @@ set(USBD_LIB src/core/libraries/usbd/usbd.cpp
              src/core/libraries/usbd/emulated/skylander.h
 )
 
-set(FIBER_LIB src/core/libraries/fiber/fiber_context.s
-              src/core/libraries/fiber/fiber.cpp
-              src/core/libraries/fiber/fiber.h
-              src/core/libraries/fiber/fiber_error.h
-)
-
-set_source_files_properties(src/core/libraries/fiber/fiber_context.s PROPERTIES COMPILE_OPTIONS -Wno-unused-command-line-argument)
+if (WIN32 AND CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_GENERATOR MATCHES "Visual Studio")
+    # Visual Studio generator with ClangCL: assemble GAS .s file via custom command
+    # since the VS build system cannot handle .s files directly.
+    set(FIBER_ASM_SRC "${CMAKE_CURRENT_SOURCE_DIR}/src/core/libraries/fiber/fiber_context.s")
+    set(FIBER_ASM_OBJ "${CMAKE_CURRENT_BINARY_DIR}/fiber_context.obj")
+    add_custom_command(
+        OUTPUT ${FIBER_ASM_OBJ}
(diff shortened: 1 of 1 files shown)


kayforbe:
Closing this for now. I'll revisit the approach after more testing to make sure it doesn't introduce any issues. Thanks for the feedback.

Niram7777:
you should rather use the Ninja generator, I dont see any reason for adding complexity to support other generators
https://github.com/shadps4-emu/shadPS4/blob/main/documents/building-windows.md

also when a CMake project has presets, this is the recommended way of building (and you can create CMakeUserPresets.json for custom options)

kayforbe:
Thank you for the feedback! My main goal was just to improve the Developer Experience (DX) for Visual Studio users who are used to the classic .sln workflow and F5 debugging, as the current setup felt a bit complex for newcomers.

However, I completely understand your point about not adding unnecessary complexity to the CMake scripts and sticking to Ninja and CMakePresets as the standard. I will close this PR. Thanks for your time and for the great work on shadPS4

PR #4233, opened 2026-04-07
Define more GS regs

mikusp:


roamic:
Are you planning to use any of those? Otherwise it's just slightly increasing compile times for no reason.

squidbus (approved):
Not sure if these are really needed but I'm not opposed to having them defined

PR #4234, opened 2026-04-07
Update scePadGetInfo

kalaposfos13:
Attempted to progress on the topic of LLE computer visioning, which wasn't very successful, but here y'all go, here's a slightly better HLE scePad function that nothing except libScePadTracker and my testing homebrew uses.

### DIFF
--- a/src/core/libraries/pad/pad.cpp
+++ b/src/core/libraries/pad/pad.cpp
@@ -188,7 +188,7 @@ int PS4_SYSV_ABI scePadGetInfo(OrbisPadInfo* data) {
     data->unk1 = 0x1;
     data->pad_handle = 1;
     data->unk3 = 0x00000101;
-    data->colour = col.r + (col.r << 8) + (col.b << 16);
+    data->colour = col.r + (col.g << 8) + (col.b << 16);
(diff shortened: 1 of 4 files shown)


src/core/libraries/pad/pad.cpp
```diff
@@ -177,19 +177,18 @@ int PS4_SYSV_ABI scePadGetIdleCount() {
...
-    data[5] = 0x0;        // ?
-    data[6] = 0x00ff0000; // colour(?)
-    data[7] = 0x0;        // ?
+    auto& controllers = *Common::Singleton<GameControllers>::Instance();
+    auto col = controllers[0]->GetLightBarRGB();
+    std::memset(data, 0, sizeof(OrbisPadInfo));
+    data->unk1 = 0x1;
+    data->pad_handle = 1;
+    data->unk3 = 0x00000101;
+    data->colour = col.r + (col.r << 8) + (col.b << 16);
```
roamic: no `g`?

kalaposfos13: Thanks for being less blind than me. :)

roamic (approved):

PR #4236, opened 2026-04-08
GS: Only use number of comps if greater than VERT_ITEMSIZE

mikusp:
Should fix regressions from https://github.com/shadps4-emu/shadPS4/pull/4213


Kyoskii:
Tested in Bloodborne, fixes the sfx regressions from #4213 🎉

georgemoralis:
fixes bb issues

PR #4240, opened 2026-04-09
constant_propagation_pass: Fold unpack ops with constants and reverse ops

raphaelthegreat:
This improves codegen in shaders that use 64-bit integers by preserving the uint64_t types through operations, instead of constantly packing to and unpacking from a uvec2. Additionally it also allows us to use SetDst64 more freely in places that currently use pairs of SetScalarReg directly without breaking existing code


StevenMiller123:
No regressions as far as I can tell.

StevenMiller123 (commented):
Guess I should've checked under a debugger first. This causes a bunch of value.h debug asserts because you're running value.U32() on a U64 value (for that value_lo part)

PR #4241, opened 2026-04-09
Races and leak fixes part 2

georgemoralis:


StevenMiller123 (approved):
Doesn't seem to break anything, doesn't fix any of the race conditions I know of either though.

StevenMiller123:
This does fix regressions from #4239 as well.

georgemoralis:
still digging for races :D

PR #4242, opened 2026-04-09
shader_recompiler: Minor improvements to buffer atomics

raphaelthegreat:
Most relevant fix is correcting the buffer type when float atomics are used on a host that doesn't support them. Previously this would crash as the U32 alias required by the fallback was not always defined, now it shouldn't


StevenMiller123:
No regressions on my end. Someone with a GPU that doesn't support float atomics should probably test too though.

PR #4243, opened 2026-04-09
frontend: Add helper methods for thread bit getters and setters

raphaelthegreat:
This will be useful for reg type tracking and cleaner overall


StevenMiller123:
No regressions on my end (though I was testing before the newest rebase).

PR #4248, opened 2026-04-11
Implement screenshot functionality with overlays and game-only options

w1naenator:
What about the screenshot feature?

My solution enables screenshots as follows:
- If RenderDoc is disabled:
  - F12 captures game content only
  - Alt+F12 captures game content + emulator HUD
- If RenderDoc is enabled:
  - F12 still captures a RenderDoc frame


src/video_core/renderer_vulkan/vk_presenter.cpp
```diff
@@ -594,6 +981,46 @@ void Presenter::Present(Frame* frame, bool is_reusing_frame) {
...
+
+            cmdbuf.pipelineBarrier(vk::PipelineStageFlagBits::eColorAttachmentOutput,
+                                   vk::PipelineStageFlagBits::eTransfer,
+                                   vk::DependencyFlagBits::eByRegion, {}, {}, to_transfer);
+            CopyImageToReadback(cmdbuf, swapchain_image, vk::ImageLayout::eTransferSrcOptimal,
+                                readback);
+            swapchain_copied_for_screenshot = true;
+        }
+
+        if (swapchain_copied_for_screenshot) {
```
Niram7777: you can declare post_barrier here so it stays const ?

bigol83:
It works but that takes screenshots at 1280x720

CUSA07399_20260411_105008_811_game_000000

that's probably because it takes the emulator window size that is 1280x720
going to try at fullscreen
CUSA07399_20260411_105345_269_game_000000

yeah, when the emulator is in fullscreen takes 1920x1080

src/input/input_handler.cpp
```diff
@@ -761,7 +773,11 @@ void ControllerOutput::FinalizeUpdate(u8 gamepad_index) {
             PushSDLEvent(SDL_EVENT_MOUSE_TO_TOUCHPAD);
             break;
         case HOTKEY_RENDERDOC:
-            PushSDLEvent(SDL_EVENT_RDOC_CAPTURE);
+            if (IsAltPressedForHotkey()) {
+                PushSDLEvent(SDL_EVENT_SCREENSHOT_WITH_OVERLAYS);
+            } else {
+                PushSDLEvent(SDL_EVENT_RDOC_CAPTURE);
+            }
```
kalaposfos13: Why not simply add a new hotkey with the binding 'lalt, f12' instead? Also, the current hotkey might benefit from being renamed, since it's not just for Renderdoc captures anymore.

notanenergydrinkaddict:
> It works but that takes screenshots at 1280x720
> 
> CUSA07399_20260411_105008_811_game_000000
> that's probably because it takes the emulator window size that is 1280x720 going to try at fullscreen CUSA07399_20260411_105345_269_game_000000
> 
> yeah, when the emulator is in fullscreen takes 1920x1080

Yea I think it should really take it at whatever the game video out buffer size was registered and not output window too

w1naenator:
Game screenshots are now captured in native resolution, but HUD screenshots are still in output resolution because HUD elements are processed after scaling.

src/input/input_handler.cpp
w1naenator: Do you like actual names?

src/input/input_handler.cpp
```diff
@@ -191,6 +192,9 @@ std::filesystem::path GetInputConfigFile(const std::string& game_id) {
                 continue;
             }
             std::string output_string = line.substr(0, equal_pos);
+            if (output_string == "hotkey_renderdoc_capture") {
+                default_bindings_to_add.erase("hotkey_capture_frame");
+            }
```
kalaposfos13, line 200: This won't do what you think it does, this part of the code only adds to the file, so it won't remove old bindings, but I think due to backwards compatibility (something we are lacking quite a bit at the moment) we shouldn't try removing it anyway, plus it'll get written back the moment someone goes back to an older version. Having both won't cause any issues either, as while it does mean there will be two bindings with the same input, one of them will always have an unknown output and therefore be ignored.

src/input/input_handler.cpp
kalaposfos13: The ones you set are fine by me.

src/input/input_handler.cpp
w1naenator: I have recovered legacy names!

kalaposfos13 (approved):

PR #4250, opened 2026-04-11
Big Picture Mode

rainmakerv3:
Adds a CLI arg to launch emulator in a very simple Big Picture mode. It's done in ImGui, so everything can be fully navigated by controller, keyboard, or mouse. Apart from possible integration with Steam and other similar apps with Big Picture mode, this could be useful for handhelds as well.

Only one setting was added to emulator settings (to save UI scale). Window size / fullscreen follows settings for the main emulator. So as not to need to initialize a Vulkan instance (very painful), this uses sdl3 renderer instead. Since the rendering stuff was taken out of the existing imgui-sdl3 implementation and there's a bunch of changed stuff, I just used a separate imgui-sdl3 implementation for this. Also needed to add sdl3_image library to handle textures for the sdl renderer.  Everything used by this interface is completely shut down upon launching a game

This is just the launching interface. If this is approved, I can work on adding a settings adjustment component to this as well.

Ui:
image


georgemoralis:
that looks quite interesting!

src/main.cpp
kalaposfos13, line 124: Wouldn't this be more user-accessible if it opened when no CLI arguments are passed, instead of adding a new flag for it?

georgemoralis: so it is opened by default?? we don't want big picture mode as default

GHU7924:
I tested your build and I like the way it looks. Regarding activating this mode, I can say the following: we can replace the Full Screen button functionality with Big Picture, and it's also worth adding a checkbox if someone wants to launch this mode by default, as this should be more convenient for users, such as Steam Deck.

BP

The problem is, you don't have any functionality bound to controller buttons here, such as Cross for game launch, Triangle for patches, Square for controller settings, and so on. In that case, there should be a line that specifies this functionality.
This can be discussed if your work is approved.

rainmakerv3:
> I tested your build and I like the way it looks. Regarding activating this mode, I can say the following: we can replace the Full Screen button functionality with Big Picture, and it's also worth adding a checkbox if someone wants to launch this mode by default, as this should be more convenient for users, such as Steam Deck.
> 
> BP
> The problem is, you don't have any functionality bound to controller buttons here, such as Cross for game launch, Triangle for patches, Square for controller settings, and so on. In that case, there should be a line that specifies this functionality. This can be discussed if your work is approved.

Cross/A should work for lauching the selected game. Settings could be bound to square once implemented, I think that's a good idea.

For fullscreen though, I think we should keep that button because some users may just want to make the emulator window fullscreen and not use the big picture mode. Adding a button or option to launch big picture from the gui also sounds good, if this is merged I can do that

PR #4254, opened 2026-04-12
fixed trophy extraction

georgemoralis:
in case the trophies doens't start from 00 as they should


StevenMiller123 (approved):
Works good on my end. Fixes a trophy-related regression in Sonic Mania (which also would've affected Uncharted: The Nathan Drake Collection if that title still worked).

PR #4255, opened 2026-04-13
Big Picture: add settings configuration

rainmakerv3:
Title. Framework is kinda done, leaving it as draft for discussion while I put in all the settings (probably won't be all of them, just a selection)

image

image


rainmakerv3:
I think this should be done now.

Some other changes apart from adding the settings dialog:

1) Removed the SDL Image library, and used the stb_image library that we are happily already using for textures instead.
2) Various styling changes, mostly keeping items highlighted as necessary, tweaking colors and borders and stuff like that
3) Used the animated background the code calls  the PrettyBackground, the same one we use for save dialogs

Some settings (mostly those listing devices) are omitted for now, maybe I'll add them in later but I think this is mostly ok. Same with adding game folders because that probably would require another separate library since imGui doesn't natively have file a selection interface

This is how the completed interface looks:

image

georgemoralis:
let's see how it survives in main :D

PR #4261, opened 2026-04-15
Trophy: Replace sdl mixer library with minimp3 header for playing trophy sounds

rainmakerv3:
Minimp3 is a single header file as opposed to a sizeable library, works just as well since we're just playing mp3s


src/common/minimp3.h
```diff
@@ -0,0 +1,1870 @@
+// SPDX-FileCopyrightText: Copyright 2026 shadPS4 Emulator Project
```
roamic: I'd prefer that updating the 3rd party dependency didn't involve searching and downloading any files from the internet. You should add it as any other dependency, a submodule in the externals folder.

roamic: Also, this is not our work, not sure why you changed the copyright

rainmakerv3: ok done

src/core/libraries/np/trophy_ui.cpp
```diff
@@ -275,6 +317,64 @@ void TrophyUI::Draw() {
     }
 }
 
+void TrophyUI::PlayMp3(std::vector<unsigned char> mp3Data) {
+    mp3dec_t mp3d;
+    mp3dec_frame_info_t info;
+    short pcm[MINIMP3_MAX_SAMPLES_PER_FRAME];
```
roamic: This takes ~4KB of stack space which can be rather limited when executing inside guest threads. I suggest moving this to the heap. It's still a good idea even if it's not executed in game-created threads.

rainmakerv3: these are admittedly new concepts to me, is the new commit ok for this?

src/core/libraries/np/trophy_ui.cpp
```diff
@@ -275,6 +317,66 @@ void TrophyUI::Draw() {
     }
 }
 
+void TrophyUI::PlayMp3(std::vector<unsigned char> mp3Data) {
+    mp3dec_t mp3d;
+    mp3dec_frame_info_t info;
+    short* pcm = new short[MINIMP3_MAX_SAMPLES_PER_FRAME];
```
roamic: I meant with std::vector, please don't manage the memory manually

roamic (approved):
LGTM

PR #4265, opened 2026-04-15
Lib.Pad: Remove special pad errors

StevenMiller123:
I'm not sure where this sort of check should exist, but it's certainly not in scePadOpen. Removing these fixes the input regressions in several Unity titles after #4027.

This needs testing with titles that actually used special pads (if I remember right, this hack was for Rock Band titles?)


gandalfthewhite19890404:
FF IX works again
image

georgemoralis:
pad responses again in 

εικόνα

georgemoralis:
Aera works again

εικόνα

brad0demx:
Fixes the input issues in my affected games 
image
image
image

kalaposfos13:
This PR breaks input in Driveclub.

PR #4070, opened 2026-02-23
threads: initialize tls on thread creation (take 2)

roamic:
Needs testing with the games that were affected by the first attempt:
* Nier Automata (demo and full)
* Cyberdimension Neptunia: 4 Goddesses Online
* TOKYO GHOUL
* Nioh

And generally any other game

PR #1659, opened 2024-12-03
The way to Unity, pt.1

polybiusproxy:
This PR is the first of a series of changes with the main goal of getting Unity (and some FNA/MonoGame) games to boot on Windows:
- Windows' `CreateThread` does not set the stack as we want so we have to ditch it in favor of the kernel counterpart `NtCreateThread`—fixes crashes on Unity games.
- Use aligned size on file mapping

This should now bring Unity games back to a softlock due to missing exception support.

PR #4267, opened 2026-04-16
PoC: Wine, 39-bit VA and hex-emu

ElectronicHimate:
## Summary
The proposed changes allow the emulator to boot some games on Wine, 39-bit VA and hex-emu. The changes can't be merged as they'll break a lot of games on anything that isn't Wine or 39-bit VA.

## Motivation
To prove that it's possible and document my findings.

### 1. Replace NtCreateThread with CreateThread (partially reverts #1659)
Fixes Wine.
### 2. Revert initialize tls on thread creation (#4070)
Fixes Wine.
### 3. Remove image base address on Windows
Fixes hex-emu.
### 4. Add configurable supported user max
Changes supported user max to 0x7000000000 (448GB) and next addr to 0x80000000 (2GB). Fixes 39-bit VA.

### 4. Gravity Rush Remastered
Without null gpu it gets device lost.

## Any Other Things
Feel free to ask any questions or close this PR if it doesn't contribute to development in any way.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1238,13 +1238,6 @@ if (WIN32)
     else()
         target_link_options(shadps4 PRIVATE -Wl,--stack,2097152)
     endif()
-
-    # Change base image address
-    if (MSVC)
-        target_link_options(shadps4 PRIVATE /BASE:0x700000000000)
-    else()
-        target_link_options(shadps4 PRIVATE -Wl,--image-base=0x700000000000)
-    endif()
 endif()
 
 if (WIN32)
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -129,7 +129,7 @@ struct AddressSpace::Impl {
         s32 sdk_ver = Common::ElfInfo::Instance().CompiledSdkVer();
         if (os_version_info.dwBuildNumber <= AffectedBuildNumber ||
             sdk_ver >= Common::ElfInfo::FW_30) {
-            supported_user_max = 0x10000000000ULL;
+            supported_user_max = EmulatorSettings.GetSupportedUserMax();
             // Only log the message if we're restricting the user max due to operating system.
             // Since higher compiled SDK versions also get reduced max, we don't need to log there.
             if (sdk_ver < Common::ElfInfo::FW_30) {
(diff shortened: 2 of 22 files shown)


red-prig:
Why the hell are you running the emulator through Wine when there's a native Linux version? Naturally, the emulator isn't compatible with Wine because Windows requires special functionality.

kalaposfos13:
> Why the hell are you running the emulator through Wine

From the screenshots, I assume it's for Winlator, although you're correct, just using the native Linux version would still be preferable over there.

georgemoralis:
> Why the hell are you running the emulator through Wine when there's a native Linux version? Naturally, the emulator isn't compatible with Wine because Windows requires special functionality.

because he can

StevenMiller123:
I think if you could dynamically detect the need for that lower user max, then code for that could be made into a PR.

The rest of this isn't particularly helpful though, this just highlights issues or limitations in these other programs.

CMakeLists.txt
```diff
@@ -1238,13 +1238,6 @@ if (WIN32)
     else()
         target_link_options(shadps4 PRIVATE -Wl,--stack,2097152)
     endif()
-
-    # Change base image address
-    if (MSVC)
-        target_link_options(shadps4 PRIVATE /BASE:0x700000000000)
-    else()
-        target_link_options(shadps4 PRIVATE -Wl,--image-base=0x700000000000)
```
Niram7777, line 1246: 3./4. I believe some games require this location, so I guess 39-bit VA is not enough and 2^47 bytes = 140.737488 terabytes would be the minimum (0x700000000000 = 123.1453 TB) + ARMv8.2-LVA + Linux 5.14

(Otherwise I guess the relevant games have to be investigated if another solution can be made) 

-> already said previously but you should rather use the Linux version although I dont believe we support Android

https://opensource.com/article/20/12/52-bit-arm64-kernel
https://static.sched.com/hosted_files/devconfcz2021/15/devconf-cz-2021-arm64-large-address-space-support.pdf

squidbus: I think this would be one of the key issues, in theory you can dynamically detect at runtime if larger VA is available before using it but it also needs us to make sure the shadps4 executable is out of the way, which is done here at compile time.

PR #4269, opened 2026-04-16
Shadnet preparations

georgemoralis:
This pr is added as preparation for shadnet.

1. Replaced psnSignedIn with shadnet enabled variable .  In ps4 each user has it's own log in for psn so global variable is useless
2. Added shadnet server setting 
3. Improved loading of users.json , saving and added new variables neccesary for shadnet log in


georgemoralis:
@rainmakerv3 since you are mostly working on gui i expecting your comments :D

src/core/emulator_settings.cpp
```diff
@@ -499,7 +499,6 @@ bool EmulatorSettingsImpl::TransferSettings() {
         setFromToml(s.volume_slider, general, "volumeSlider");
         setFromToml(s.neo_mode, general, "isPS4Pro");
         setFromToml(s.dev_kit_mode, general, "isDevKit");
-        setFromToml(s.psn_signed_in, general, "isPSNSignedIn");
```
StevenMiller123, line 502: I don't see why the setting transfer had to be removed?

src/core/libraries/np/np_web_api2.cpp
```diff
@@ -207,8 +207,9 @@ s32 PS4_SYSV_ABI sceNpWebApi2SendMultipartRequest() {
 }
 
 s32 PS4_SYSV_ABI sceNpWebApi2SendRequest() {
-    if (!EmulatorSettings.IsPSNSignedIn()) {
-        LOG_INFO(Lib_NpWebApi2, "called, returning PSN signed out.");
+    if (!EmulatorSettings.IsShadNetEnabled()) {
+        LOG_INFO(Lib_NpWebApi2,
+                 "called, returning but shadnet is disabled simulate PSN signed out.");
```
StevenMiller123: If you want a message like this, then just change it to "called, returning shadNet signed out" or something along those lines. This error message is a grammatical nightmare.

StevenMiller123 (commented):
Some quick comments.

src/core/emulator_settings.cpp
georgemoralis: this is not the same thing the one was about psn the other is shadnet . Considering that psn signed in will be per user i don't think it is relative to transfer

src/core/libraries/np/np_web_api2.cpp
georgemoralis: done :D

StevenMiller123 (approved):
Code looks fine, not that it's really doing much yet.

georgemoralis:
it's preparation code :D (so @rainmakerv3 can do the gui stuff :p )

PR #4270, opened 2026-04-17
libraries/ult: Implement initial libSceUlt

RenO5:
Adds a basic implementation of the PS4 User-Level Threading library.

Implemented:
- sceUltInitialize
- WaitingQueueResourcePool create/work area size
- QueueDataResourcePool create/work area size
- Queue create/push/pop
- UlthreadRuntime create/destroy/work area size/opt param initialize
- Ulthread create/join

ULT threads are mapped to host threads for now. Queue synchronization uses std::mutex and std::condition_variable. This is sufficient to get libSceUlt-dependent titles past initialization.

Tested with Dead or Alive Xtreme 3 Fortune (CUSA04555), which previously crashed on launch due to missing ULT symbols and now boots in-game.

### DIFF
--- a/src/core/libraries/libs.cpp
+++ b/src/core/libraries/libs.cpp
@@ -73,6 +73,7 @@
 #include "core/libraries/web_browser_dialog/webbrowserdialog.h"
 #include "core/libraries/zlib/zlib_sce.h"
 #include "fiber/fiber.h"
+#include "core/libraries/ult/ult.h"
 
 namespace Libraries {
 
@@ -137,6 +138,7 @@ void InitHLELibs(Core::Loader::SymbolsResolver* sym) {
     Libraries::RazorCpu::RegisterLib(sym);
     Libraries::Move::RegisterLib(sym);
     Libraries::Fiber::RegisterLib(sym);
+    Libraries::Ult::RegisterlibSceUlt(sym);
     Libraries::Mouse::RegisterLib(sym);
     Libraries::WebBrowserDialog::RegisterLib(sym);
     Libraries::Zlib::RegisterLib(sym);
--- a/src/core/libraries/ult/ult.cpp
+++ b/src/core/libraries/ult/ult.cpp
@@ -1,0 +1,252 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <condition_variable>
+#include <cstring>
+#include <deque>
+#include <mutex>
+#include <thread>
+#include <vector>
+
+#include "common/logging/log.h"
+#include "core/libraries/libs.h"
+#include "core/libraries/ult/ult.h"
+#include "core/libraries/ult/ult_error.h"
+
+namespace Libraries::Ult {
+
+// Internal thread-safe queue implementation
+struct UltQueueInternal {
+    std::mutex mtx;
+    std::condition_variable cv_not_empty;
+    std::condition_variable cv_not_full;
+    std::deque<std::vector<u8>> items;
+    u32 data_size;
+    u32 max_items;
+    bool destroyed = false;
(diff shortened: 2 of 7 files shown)


src/core/libraries/ult/ult_error.h
```diff
@@ -0,0 +1,16 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+constexpr int ORBIS_OK = 0;
+constexpr int SCE_ULT_ERROR_NULL = 0x80810001;
```
StevenMiller123, line 7: Our convention is for errors to be formatted as `ORBIS_` instead of `SCE_`

src/core/libraries/ult/ult_error.h
```diff
@@ -0,0 +1,16 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+constexpr int ORBIS_OK = 0;
```
StevenMiller123, line 6: Please avoid redefining constants.
Include `core/libraries/error_codes.h` in code/header files that need our `ORBIS_OK` constant.

src/core/libraries/ult/ult.h
```diff
@@ -0,0 +1,88 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
```
StevenMiller123, line 1: New files should have an updated copyright year. These files were made in 2026, not 2024.

src/core/libraries/ult/ult.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+
+s32 PS4_SYSV_ABI sceUltQueuePop(UltQueue* queue, void* data) {
+    if (!queue || !queue->internal || !data) {
+        LOG_ERROR(Lib_Ult, "null parameter");
+        return SCE_ULT_ERROR_NULL;
+    }
+
+    auto* q = queue->internal;
+    std::unique_lock lock(q->mtx);
```
StevenMiller123, line 129: Do we know if there are any Unity games using this library?
If so, we can't use standard library functions here, or those games will have Windows-specific hangs when using this HLE.

src/core/libraries/libs.cpp
```diff
@@ -137,6 +138,7 @@ void InitHLELibs(Core::Loader::SymbolsResolver* sym) {
     Libraries::RazorCpu::RegisterLib(sym);
     Libraries::Move::RegisterLib(sym);
     Libraries::Fiber::RegisterLib(sym);
+    Libraries::Ult::RegisterlibSceUlt(sym);
```
StevenMiller123, line 141: libSceUlt is a library we run LLE. For these, you need to add the `RegisterLib` call in libSceSysmodule instead.
https://github.com/shadps4-emu/shadPS4/blob/2d559a4f77b2e78f17ea809fd062016f8a6c6d7a/src/core/libraries/sysmodule/sysmodule_internal.cpp#L215

Placing this in libs.cpp forces games to use this libSceUlt HLE instead of the LLE, and as far as I'm aware, LLE libSceUlt works fine.

StevenMiller123 (changes requested):
This is a promising start, though it needs testing. Plenty of small styling nitpicks, but overall the code looks good.

Biggest issue to fix is the RegisterlibSceUlt call, since that will override (and potentially break) LLE libSceUlt usage.

Frostwave-afk:
Hi, I've been trying to test this PR as DOAX3 (CUSA04555) is exactly the game you mentioned. The CI artifacts from April have expired. Could you trigger a fresh build run so I can download the Windows artifact and test it? Thanks

PR #4274, opened 2026-04-18
separate thread init and run stacks

roamic:
* Created a separate stack for thread initialization. The actual thread function provided by the game now runs on a separate stack.
* The stack is now zeroed on creation and/or reuse.

This should fix unity games on Linux that were broken by the previous stack manipulations.
Also worth checking:
* Gravity Rush 2
* Catherine
* TOKYO GHOUL
and any other game in general because this is widely used functionality


brad0demx:
Fixes my affected unity games on Linux 
image
image
image

StevenMiller123 (approved):
Only broken game is TOKYO GHOUL, but with how much this fixes, I say we should get this moving.

PR #4287, opened 2026-04-19
Core: Fix issues with #4273

StevenMiller123:
This PR adds asserts for failed module mappings, and for macOS, disables backing for mappings with execute permissions.

### DIFF
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -716,6 +716,11 @@ struct AddressSpace::Impl {
     void* Map(VAddr virtual_addr, PAddr phys_addr, u64 size, PosixPageProtection prot,
               int fd = -1) {
         m_free_regions.subtract({virtual_addr, virtual_addr + size});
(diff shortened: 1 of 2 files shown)


src/core/address_space.cpp
```diff
@@ -716,6 +716,11 @@ struct AddressSpace::Impl {
     void* Map(VAddr virtual_addr, PAddr phys_addr, u64 size, PosixPageProtection prot,
               int fd = -1) {
         m_free_regions.subtract({virtual_addr, virtual_addr + size});
+#ifdef __APPLE__
+        if (prot & PROT_EXEC != 0) {
+            phys_addr = -1;
```
squidbus, line 722: May want to assert here if `fd != -1`, which means file mapping was explicitly requested.

StevenMiller123: Added.

squidbus (approved):

PR #4291, opened 2026-04-20
renderer_vulkan: Add a few more feature support guards.

squidbus:
* Add check for `VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT` support before using with block compressed images, since it isn't strictly needed for many games.
* Add check for `VK_EXT_extended_dynamic_state3` support to determine whether `pDepthStencilState` can be `nullptr` in graphics pipelines.
* Only try `VK_EXT_portability_subset` for MoltenVK.

These were minor validation issues uncovered doing some early testing with the KosmicKrisp driver on macOS.


raphaelthegreat (approved):

UltraDaCat:
Games no longer crash on startup when using an x86 macOS device.
image
image

PR #4294, opened 2026-04-20
Dynamically detect 39-bit address space

yellowsparkles584:
Summary:
Based on #4267 feedback.
Removing image base address was unnecessary.
NtCreateThread was already removed in #4277


squidbus:
Why are we reverting these other PRs?

yellowsparkles584:
> Why are we reverting these other PRs?

It hangs when loading modules with them.

squidbus:
> It hangs when loading modules with them.

I think that issue would need to be addressed, then, not just a blanket revert.

yellowsparkles584:
Works now without reverting commits.

src/core/module.cpp
```diff
@@ -153,7 +154,8 @@ void Module::LoadModuleToMemory(u32& max_tls_index) {
             }
 
             // Map module segments
-            const auto memory_type = IsSystemLib() ? VMAType::Code : VMAType::Flexible;
+            const auto memory_type =
+                (IsSystemLib() || AddressSpace::Is39Bit()) ? VMAType::Code : VMAType::Flexible;
```
squidbus: I'm questioning the logic here as I'm not sure how 39-bit address space is related to this problem.

On macOS we faced a similar hang issue because it doesn't support mapping shared memory (which we use for the backing of direct/flexible blocks) as executable. What I would assume is maybe you're hitting some similar limitation running with wine (specifically on Android?).

In which case the check should really be whether we are running under wine, although I'm not sure if we intend to support running under wine. Like was said in the other PR related to this stuff, it would probably be better to use Linux builds on Android than Windows ones under wine.

src/core/address_space.cpp
```diff
@@ -141,7 +141,18 @@ struct AddressSpace::Impl {
         }
 
         // Determine the free address ranges we can access.
+        const VAddr va_size = reinterpret_cast<VAddr>(sys_info.lpMaximumApplicationAddress);
```
StevenMiller123: I think a better solution would be to respect this maximum address.

StevenMiller123 (changes requested):
Respecting what Windows reports as the maximum address is probably a good idea, though I think that limit should be applied directly (and if needed, overridden to avoid issues with Windows 10 or our image base).

The issue with module mappings needs to be looked into separately. If it is an actual Wine bug, it should be reported to their devs. Running under Wine is not considered a supported platform, so I don't think we should have workarounds for it.

src/core/module.cpp
TheKingFireS: Using Linux build on Android wouldn't solve issues because box64 will also set 39bit address if detected too. It's primarily Android Arm64 issue.

squidbus: Yes but like I said whatever is going on here with the code vs flexible mappings is not related to 39-bit address space, these are separate issues.

yellowsparkles584:
Changed supported user max to what Windows reports as the maximum address as requested, this needs testing on Windows 11.
The module mappings issue was a fex bug, I don't know if I should add workarounds for it.

PR #4298, opened 2026-04-20
Fix the FreeBSD submodule fix

kalaposfos13:
@xinitrcn1 already fixed hwinfo for FreeBSD, and even upstreamed the changes, but unfortunately, that is not how submodules work, so here's a PR to actually use that update.
https://github.com/shadps4-emu/ext-hwinfo/pull/2
Do not merge yet as I'm still going through things to ensure there aren't other things that also broke on FreeBSD.


kalaposfos13:
The two fixes were separately added by #4303 and #4318, I assume without either one's author seeing that I have already added those things here.

PR #4299, opened 2026-04-20
trying make eventflags alertable with BinarySemaphore

georgemoralis:
Seems to fix unity games , tests in other games needed too
(I am not responsible if you don't test and it breaks bloodborne , or driveclub or ...)


StevenMiller123:
Some tests I did prior to shadow's PR:
Fixes hangs when booting newer versions of PowerWash Simulator
image

Fixes hangs when finishing battles in NEO: The World Ends with You
image

Fixes hangs when booting Genshin Impact
image

I didn't do any non-Unity tests, I don't have a lot of spare time for testing right now.

raphaelthegreat:
The implementation appears to have been copied from semaphore file and doesn't look very clean, don't think this is mergeable at its current state. But its a good start.

LuismaSP89:
Subnautica still crashing on boot even with this PR

PR #4302, opened 2026-04-21
Dont use logger after Shutdown

Niram7777:
Emulator::Restart calls Log::Shutdown then quick_exit (which calls Flush)
thanks to @StevenMiller123 for reporting and help reproduce
- [x] I test when the Windows build finishes

### DIFF
--- a/src/common/logging/log.cpp
+++ b/src/common/logging/log.cpp
@@ -242,7 +242,12 @@ void Shutdown() {
 }
 
 void Flush() {
-    g_shad_file_sink->flush();
-    g_console_sink->flush();
+    if (g_shad_file_sink != nullptr) {
+        g_shad_file_sink->flush();
+    }
+
+    if (g_console_sink != nullptr) {
+        g_console_sink->flush();
+    }
 }
 } // namespace Common::Log
--- a/src/common/logging/log.h
+++ b/src/common/logging/log.h
@@ -37,7 +37,11 @@ void Flush();
 
 // Define the fmt lib macros
 #define LOG_GENERIC(log_class, log_level, ...)                                                     \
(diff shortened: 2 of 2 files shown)


StevenMiller123:
Doesn't seem to fix the issue I have.
image
[CUSA00803.log](https://github.com/user-attachments/files/26945199/CUSA00803.log)

Niram7777:
> Doesn't seem to fix the issue I have. image [CUSA00803.log](https://github.com/user-attachments/files/26945199/CUSA00803.log)

I suspec there is a HLE lib that is global (outlives the loggers) and tries to log.
Could you run in a dbg to check which function/thread cause this ? Or open C:/Windows/Minidump/*.dmp with Visual Studio ? 
(The issue has disappeared in the homebrew but I guess because too basic.)

StevenMiller123:
Seems you're correct, in a debugger I get an exception because of this debug log occurring after the logger is destroyed.
image

Niram7777:
> Seems you're correct, in a debugger I get an exception because of this debug log occurring after the logger is destroyed. image

this would need https://github.com/shadps4-emu/shadPS4/pull/4220 as a proper fix, but I will try a quick around until then....

StevenMiller123:
Couldn't you replace the log macros, or add a check to see if the logger is destroyed in them?

Niram7777:
> Couldn't you replace the log macros, or add a check to see if the logger is destroyed in them?

yes this is what I will do, (I was wandering if I throw, but probably safer to just do nothing in this case)

Niram7777:
@StevenMiller123 ready for test/merge

StevenMiller123:
Seems to work now.

image

PR #4306, opened 2026-04-23
Lib.NpTrophy: Fix workaround for missing trophy key.

StevenMiller123:
In #4177, @georgemoralis added an error return that triggers when trophy files fail to extract. Users who didn't specify a trophy key started seeing errors in some games.


georgemoralis:
> In #4177, @georgemoralis added an error return that triggers when trophy files fail to extract. Users who didn't specify a trophy key started seeing errors in some games.

yah blame me again , i started to get used it :D

shinra-electric:
With this PR Dark Souls, Darks Souls 2 and Yakuza 0 all go in-game again now when a trophy key is absent.

PR #4315, opened 2026-04-24
Save fixes VIIIIX

georgemoralis:
fix issue with concurrent writes , this fix racing with permission denied

e.g 

io_file.cpp:202 Open: Failed to open the file at path=.....\savedata\...\sce_sdmemory\memory.dat, error_message=permission denied

### DIFF
--- a/src/core/libraries/save_data/save_backup.cpp
+++ b/src/core/libraries/save_data/save_backup.cpp
@@ -34,6 +34,7 @@ static void CopyShareReadWrite(const std::filesystem::path& from, const std::fil
         }
         return;
     }
+
     Common::FS::IOFile src{from, Common::FS::FileAccessMode::Read, Common::FS::FileType::BinaryFile,
                            Common::FS::FileShareFlag::ShareReadWrite};
     if (!src.IsOpen()) {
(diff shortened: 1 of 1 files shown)


DanielSvoboda:
This fixes the problem where OKAMI wouldn't save.

PR 4315
image

0.15.1 Wip 615949e
image

georgemoralis:
i found that in beach buggy racing 2 btw

StevenMiller123:
Fixes the save-related regression in DanMachi BATTLE CHRONICLE, the title is now back to hanging on the intros.
image

[CUSA46310.log](https://github.com/user-attachments/files/27064129/CUSA46310.log)

src/core/libraries/save_data/save_backup.cpp
```diff
@@ -24,6 +25,41 @@ namespace fs = std::filesystem;
...
+    Common::FS::IOFile src{from, Common::FS::FileAccessMode::Read, Common::FS::FileType::BinaryFile,
+                           Common::FS::FileShareFlag::ShareReadWrite};
+    if (!src.IsOpen()) {
+        LOG_ERROR(Lib_SaveData, "Backup: failed to open source {}", fmt::UTF(from.u8string()));
+        return;
+    }
+    const u64 size = src.GetSize();
+    std::vector<u8> buf;
+    buf.resize(size);
+    if (size > 0) {
```
StevenMiller123, line 46: Might be better to just do `src.Close()` and return early if size is 0.

georgemoralis: done

src/core/libraries/save_data/save_backup.cpp
```diff
@@ -24,6 +25,45 @@ namespace fs = std::filesystem;
...
+    if (size == 0) {
+        src.Close();
+        return;
+    }
+    if (size > 0) {
+        src.ReadRaw<u8>(buf.data(), size);
+    }
+    src.Close();
+
+    Common::FS::IOFile dst{to, Common::FS::FileAccessMode::Write, Common::FS::FileType::BinaryFile,
```
StevenMiller123: Does the destination file always exist?
If not, it might be better to use FileAccessMode::Create here.

StevenMiller123 (approved):
Seems fine to me.

src/core/libraries/save_data/save_backup.cpp
georgemoralis: yes but size == 0 was remove now that i am thinking it because src might be an empty file (0 bytes ) but it will fail to create a empty file in dst

Linear524:
This finally fixed Last Guardian crashes with savefiles !!!
Thanks a million times !!! :) :) :)

Linear524:
Trico

georgemoralis:
> This finally fixed Last Guardian crashes with savefiles !!! Thanks a million times !!! :) :) :)

Glad you like it ;D

PR #4319, opened 2026-04-25
Fix log async thread name and custom setting not loaded

Niram7777:


Kravickas:
per-game log settings seem to be fixed

PR #4320, opened 2026-04-25
V_DIV_SCALE_F32

mikusp:
Hit by Resident Evil 2 in Neo mode.

The algorithm is based on the RDNA2 ISA pdf, with some changes where the spec clearly contains errors 😞 Tests included were confirmed with a PS4 Pro

Numerical issues found during testing on Nvidia:
- taking the exponent of Infinity returns 0 instead of 128 so that is checked explicitly
- denormals are flushed to zero (sometimes, but in the case of division they seem to be) so checking if division result would be denormal is done in a roundabout way
- input denormals are flushed, so even if the runtime shader info would request denormal preservation (unlikely for retail games), that would have no effect

Leaving as draft so others can test with different gpu/os configurations and point out any mistakes


StevenMiller123:
With how many of these edge cases Nvidia handles differently, would it be better to add a recompiler pass for injecting necessary workarounds?
I think it would be a whole lot nicer than adding these sorts of checks in basically every instruction.

mikusp:
I don't know if it's necessary Nvidia that does things differently, I don't have access to any different GPUs to test

StevenMiller123:
Well, if it's not just Nvidia, we can just have more GPUs run the pass.
My main point is, issues like what you're describing probably affect a ton of instructions (very few of which currently have these extra checks).

If it's feasible, injecting these checks through an extra recompiler pass would clean up instruction code and prevent us from needing to add these sorts of checks  to many of our currently implemented instructions.

Of course, if that's not feasible then you can ignore this comment 😅

squidbus:
Generally yeah, if something like preserve denorms, etc are needed but not available from the host features, we would put a flag for it in the profile and use an optional pass to rewrite the relevant IR to account for it.

mikusp:
Denorm preservation is not strictly needed (unless the game requests that, but I've haven't seen any doing that yet), and as I've explained in the discord, I don't see a value in creating a new pass just to handle that one instruction. It would effectively exchange a FPDiv and a comparison to a FPMul and a comparison. Moreover, for the check with FPDiv to work we'd need to enable denorm preservation globally for the shader, and that would usually not be something that the game requested.

Here the division denorm check is an implementation detail of the instruction and works the same regardless of whether the GPU is configured to preserve or flush denorms

georgemoralis:
this appears to be out of date @mikusp , care to rebase?

PR #4321, opened 2026-04-25
Fix big picture crash on some compiled builds

rainmakerv3:
Some compiled builds were crashing on std::filesystem::path ebootPath(*gamePath) when passing arg -b for big picture mode - without any game arg it would be constructing from an empty optional.

Instead of adding a !has_value check for this, I think it's just cleaner to move launching big picture up before any other args are checked.  The other args aren't passed to Big picture mode anyway

EDIT: also added passing executable name which is needed for cases such as restarting the emulator when loading new elfs from a collection (ex: Uncharted collection)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -140,6 +140,11 @@ int main(int argc, char* argv[]) {
     Common::Log::g_should_append |= EmulatorSettings.IsLogAppend();
     Common::Log::Setup("shad_log.txt");
 
+    if (bigPicture) {
+        BigPictureMode::Launch();
+        return 0;
+    }
+
     // ---- Utility commands ----
     if (addGameFolder) {
         EmulatorSettings.AddGameInstallDir(*addGameFolder);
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -140,6 +140,11 @@ int main(int argc, char* argv[]) {
     Common::Log::g_should_append |= EmulatorSettings.IsLogAppend();
     Common::Log::Setup("shad_log.txt");
 
+    if (bigPicture) {
```
Niram7777, line 143: I think you are ignoring some cli args so you may want to move it just before "// ---- Resolve game path or ID ----" but after the "EmulatorSettings.*(...)"

rainmakerv3: Honestly not sure which approach is better.  Respecting the args that override emulator settings may be confusing to users since the big picture settings dialog would show the overriden settings instead of what's in the actual settings files. And since big picture has to ignore some args anyway, I feel it's more intuitive if it just ignored all the other args as well for consistency.

The downside of course a the certain CLI functionality loss with big picture mode, but I think users would just bypass big picture mode anyway if they want to launch a game with the other args

So for me I think placing it here and ignoring all other args seems fine.

PR #4322, opened 2026-04-25
Fix Vulkan physical-device auto-selection comparator to satisfy strict weak ordering

qwe854896:
## Summary

This PR fixes the comparator used for Vulkan physical-device auto-selection.

That comparator is passed to `std::sort`, but it does not satisfy strict weak ordering:

1. the Vulkan API threshold preference is asymmetric
2. the device-type preference makes differing non-discrete types mutually equivalent, while still ordering same-type non-discrete devices by memory size

As a result, the sort can produce contradictory comparisons, and the auto-selection path has undefined behavior.

The intended priority appears to be:

1. Prefer devices meeting `TargetVulkanApiVersion`
2. Then prefer discrete GPUs
3. Then prefer larger device-local memory

but the current implementation does not encode those priorities as a valid ordering relation.

It makes both the Vulkan-API and discrete-GPU comparisons explicit and symmetric before the memory fallback, while preserving the original intended priority order.

### Witness 1: Vulkan API threshold asymmetry

Let the target API be Vulkan 1.3.

Consider two devices:

1. `A`: `apiVersion < TargetVulkanApiVersion`, same device type as `B`, larger device-local memory
2. `B`: `apiVersion >= TargetVulkanApiVersion`, same device type as `A`, smaller device-local memory

With the current comparator:

https://github.com/shadps4-emu/shadPS4/blob/5a15eb4d94332d6e11847a4ae067e2ad69b4036b/src/video_core/renderer_vulkan/vk_instance.cpp#L119-L122

1. `cmp(B, A) == true`
   because `B` satisfies the Vulkan API threshold and `A` does not
2. `cmp(A, B) == true`
   because the API-specific early return does not trigger in this direction, the device types are the same, and the comparator falls through to the memory comparison where `A` wins

So the comparator can report both `A < B` and `B < A`.

### Witness 2: non-discrete device-type equivalence is not transitive

Consider three devices with the same Vulkan API support status:

1. `A`: integrated GPU with 8 GB device-local memory
2. `B`: CPU or virtual GPU with 4 GB device-local memory
3. `C`: integrated GPU with 2 GB device-local memory

With the current comparator:

https://github.com/shadps4-emu/shadPS4/blob/5a15eb4d94332d6e11847a4ae067e2ad69b4036b/src/video_core/renderer_vulkan/vk_instance.cpp#L123-L125

1. `cmp(A, B) == false` and `cmp(B, A) == false`
   because the device types differ, and neither side is discrete
2. `cmp(B, C) == false` and `cmp(C, B) == false`
   for the same reason
3. `cmp(A, C) == true`
   because both are integrated GPUs, so the comparator falls through to the memory comparison and prefers `A`

So `A ~ B` and `B ~ C`, but not `A ~ C`, which violates transitivity of equivalence and therefore strict weak ordering.

## Potential Outcomes

Possible downstream effects include:

1. Undefined behavior in `std::sort`, which may cause out-of-bound read/write and potential crash.
2. Unstable or implementation-dependent device ordering
3. Wrong Vulkan physical-device selection during automatic startup
4. Selecting a less suitable non-discrete device because the comparator does not consistently express its intended priority order

This kind of bug can remain latent for a long time because many systems expose only one practical Vulkan device, or only device combinations that do not hit a contradictory comparison.

## Related Issues

I checked the public tracker for likely symptom-level matches.

- #232

## Fix

The current patch is:

1. Make the Vulkan API support comparison symmetric
2. Make the discrete-GPU preference symmetric
3. Preserve the existing memory-size fallback

That keeps the original selection intent and makes the comparator valid for use with `std::sort`.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_instance.cpp
+++ b/src/video_core/renderer_vulkan/vk_instance.cpp
@@ -116,13 +116,20 @@ Instance::Instance(Frontend::WindowSDL& window, s32 physical_device_index,
         std::sort(properties2.begin(), properties2.end(), [](const auto& left, const auto& right) {
             const vk::PhysicalDeviceProperties& left_prop = std::get<1>(left).properties;
             const vk::PhysicalDeviceProperties& right_prop = std::get<1>(right).properties;
-            if (left_prop.apiVersion >= TargetVulkanApiVersion &&
-                right_prop.apiVersion < TargetVulkanApiVersion) {
-                return true;
+            const bool left_supports_api = left_prop.apiVersion >= TargetVulkanApiVersion;
+            const bool right_supports_api = right_prop.apiVersion >= TargetVulkanApiVersion;
+            if (left_supports_api != right_supports_api) {
+                return left_supports_api;
             }
-            if (left_prop.deviceType != right_prop.deviceType) {
-                return left_prop.deviceType == vk::PhysicalDeviceType::eDiscreteGpu;
+
+            const bool left_is_discrete =
+                left_prop.deviceType == vk::PhysicalDeviceType::eDiscreteGpu;
+            const bool right_is_discrete =
+                right_prop.deviceType == vk::PhysicalDeviceType::eDiscreteGpu;
+            if (left_is_discrete != right_is_discrete) {
+                return left_is_discrete;
             }
+
             constexpr auto get_mem = [](const vk::PhysicalDeviceMemoryProperties& mem) -> size_t {
                 size_t max = 0;
                 for (u32 i = 0; i < mem.memoryHeapCount; i++) {


kalaposfos13:
This is just AI slop that doesn't even do anything productive, nor does it fix anything other than a vague possibility of the emulator selecting the wrong GPU (which to my knowledge has never been reported so far).

qwe854896:
I apologize for making this look like AI slop.

English is not my native language, so I used an LLM to help phrase the PR description. However, the observations, reasoning, and validation behind it are from my own investigation.

This is still a real problem. I currently have a proof-of-concept that can trigger a segmentation fault on x86_64 Linux systems. I also mentioned #232 because the full discussion strongly suggests a real wrong-GPU auto-selection problem: the reporter had a Vulkan-1.3-capable GPU, contributors agreed it should support 1.3, and later testing showed that auto-select still picked the wrong adapter while manually selecting the correct GPU made games work. At the same time, that discussion documents a manual-selection workaround rather than a definitive root cause and fix for the underlying auto-selection bug. That issue does not by itself prove this comparator bug is the root cause, but it is clearly consistent with incorrect physical-device selection.

If needed, I can provide the reproducer and more technical details later.

Also, for `std::sort`, the comparator is required to induce a strict weak ordering; otherwise the call is outside the algorithm's contract.

> `comp(a, b) && comp(b, c) implies comp(a, c)`
> `equiv(a, b) && equiv(b, c) implies equiv(a, c)`

Here, `equiv(a, b)` means `!comp(a, b) && !comp(b, a)`.

References:
- C++ draft, sorting requirement: https://eel.is/c++draft/alg.sorting.general
- C++ draft, strict weak order definition: https://eel.is/c++draft/concept.strictweakorder

squidbus (approved):

drekunia:
UPDATE: sorry, forgot to try the very obvious stuff to do, it's just the shader cache. it's all fine now, my bad.

---

i can't launch Bloodborne using BBLauncher after this update.

this is the console output:
```
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=1, number_format=9 is not fully supported (vk_format=R8Srgb, missing features={ ColorAttachment })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=3, number_format=9 is not fully supported (vk_format=R8G8Srgb, missing features={ SampledImage | ColorAttachment | TransferSrc | TransferDst })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=9, number_format=1 is not fully supported (vk_format=A2B10G10R10SnormPack32, missing features={ SampledImage | StorageImage | ColorAttachment | TransferSrc | TransferDst | StorageReadWithoutFormat | StorageWriteWithoutFormat })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=9, number_format=5 is not fully supported (vk_format=A2B10G10R10SintPack32, missing features={ SampledImage | StorageImage | ColorAttachment | TransferSrc | TransferDst | StorageReadWithoutFormat | StorageWriteWithoutFormat })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=13, number_format=4 is not fully supported (vk_format=R32G32B32Uint, missing features={ SampledImage | TransferSrc | TransferDst })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=13, number_format=5 is not fully supported (vk_format=R32G32B32Sint, missing features={ SampledImage | TransferSrc | TransferDst })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=13, number_format=7 is not fully supported (vk_format=R32G32B32Sfloat, missing features={ SampledImage | TransferSrc | TransferDst })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=16, number_format=0 is not fully supported (vk_format=R5G6B5UnormPack16, missing features={ StorageImage | StorageReadWithoutFormat | StorageWriteWithoutFormat })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=17, number_format=0 is not fully supported (vk_format=A1R5G5B5UnormPack16, missing features={ StorageImage | StorageReadWithoutFormat | StorageWriteWithoutFormat })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=18, number_format=0 is not fully supported (vk_format=R5G5B5A1UnormPack16, missing features={ StorageImage | ColorAttachment | StorageReadWithoutFormat | StorageWriteWithoutFormat })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:719 CollectImageFormatInfo: Surface format data_format=19, number_format=0 is not fully supported (vk_format=B4G4R4A4UnormPack16, missing features={ StorageImage | ColorAttachment | StorageReadWithoutFormat | StorageWriteWithoutFormat })
[Render.Vulkan] <Warning> (shadPS4:Main) vk_instance.cpp:729 CollectImageFormatInfo: Depth format z_format=1, stencil_format=1 is not fully supported (vk_format=D16UnormS8Uint, missing features={ DepthStencilAttachment })
[Debug] <Critical> (shadPS4:Main) vk_pipeline_serialization.cpp:283 lambda: Assertion Failed!
Permutation 0 is already inserted at 1! (fs_bf368417)
```

this is my BBLauncher settings and plugins:
image

sorry that i don't understand anything about the codebase, how emulation works, and what this PR meant to fix. just want to let you know (or to confirm if i post this on the right repo).

would be nice if anyone here can explain some context or how to fix this on my side, in the mean time, i'll try to compile the previous head manually for now. thanks.

squidbus:
I assume from the error message it's a shader cache thing in which case you might just need to clear it, either way I don't think that looks directly related to this change, most I can think is if this change just altered which GPU you end up defaulted to.

drekunia:
> I assume from the error message it's a shader cache thing in which case you might just need to clear it, either way I don't think that looks directly related to this change, most I can think is if this change just altered which GPU you end up defaulted to.

@squidbus yes, wrote an update above my previous message beforehand, sorry for the false alarm.

PR #4323, opened 2026-04-26
Recompiler: Use num_records to check if an inline buffer exists

StevenMiller123:
After #3816, we store inline buffers with their relative address, instead of the absolute address.
Some games create inline buffers with a relative address of 0, and after this change, those inline buffers would no longer detect properly (causing an attempt to read a sharp from sharp_idx = -1, which is hardcoded for inline buffers).

This PR changes our buffer bool() operator to check against num_records instead, which is usually non-zero (as zero would imply a buffer of 0 size). This fixes the regression in The Last of Us Remastered.

I'm not sure if this is a valid change, mainly since I don't know if there's a reason a game might intentionally register a buffer with 0 size (in which case, the check would need to be against all fields, or we'd need to check for inline buffers some other way).

### DIFF
--- a/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
+++ b/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
@@ -506,7 +506,7 @@ void PatchBufferSharp(IR::Block& block, IR::Inst& inst, Info& info, Descriptors&
         // is used to define an inline buffer resource
         std::array<u64, 2> raw;
         // Keep relative address, we'll do fixup of the address at buffer fetch later
-        raw[0] = (handle->Arg(0).U32() | u64(handle->Arg(1).U32()) << 32);
+        raw[0] = handle->Arg(0).U32() | u64(handle->Arg(1).U32()) << 32;
         raw[1] = handle->Arg(2).U32() | u64(handle->Arg(3).U32()) << 32;
         const auto buffer = std::bit_cast<AmdGpu::Buffer>(raw);
         buffer_binding = descriptors.Add(BufferResource{
--- a/src/shader_recompiler/resource.h
+++ b/src/shader_recompiler/resource.h
@@ -56,7 +56,7 @@ struct BufferResource {
         AmdGpu::Buffer buffer{};
         if (inline_cbuf) {
             buffer = inline_cbuf;
-            if (inline_cbuf.base_address > 1) {
+            if (inline_cbuf.base_address != 1) {
                 buffer.base_address += info.pgm_base; // address fixup
             }
         } else {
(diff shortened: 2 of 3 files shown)


StevenMiller123:
image

StevenMiller123:
Also fixes a similar crash in Uncharted: The Nathan Drake Collection
image

From what I know, this should fix the regression in The Order 1886 too, but I don't have that title.

Graubek:
Uncharted now crashing after displaying splashscreen (before, it was just crashing and now whole pc is lagging when game is turned on) on 
```
[Debug] <Critical> (shadPS4:GpuCommandProcessor) vk_presenter.cpp:1108 lambda: Assertion Failed!
```
The Order regression also seems to be fixed (game is in the menu status)

StevenMiller123:
Uncharted has always done that, make sure direct memory access is enabled in your config.

StevenMiller123:
For the record, the "whole pc is lagging when game is turned on" is an Nvidia driver thing, I've reproduced it before on my computer.
Though like I said, the core issue is that these titles need direct memory access enabled (and also readbacks for some graphical stuff).

Graubek:
Ok, Uncharted is in the menu with dma

chuwak86:
and this build allows to run game scene for uncharted 3

chronicer:
Perfect but again crashes at this

[Tty] <Info> (NdJobWorkerThread 0) logger.cpp:62 log_flush: [stdout] ******************************************************************************
[Tty] <Info> (NdJobWorkerThread 0) logger.cpp:62 log_flush: [stdout] *** ASSERTION: GetRenderFrameParams(frameToFreeInstancesFor)->m_gfxEopTick
[Tty] <Info> (NdJobWorkerThread 0) logger.cpp:62 log_flush: [stdout] ***
[Tty] <Info> (NdJobWorkerThread 0) logger.cpp:62 log_flush: [stdout] ***  Function: FrameBegin
[Tty] <Info> (NdJobWorkerThread 0) logger.cpp:62 log_flush: [stdout] ***      File: ndlib\render\fg-draw-mgr.cpp
[Tty] <Info> (NdJobWorkerThread 0) logger.cpp:62 log_flush: [stdout] ***      Line: 234
[Tty] <Info> (NdJobWorkerThread 0) logger.cpp:62 log_flush: [stdout] ***     Built:
[Tty] <Info> (NdJobWorkerThread 0) logger.cpp:62 log_flush: [stdout] ******************************************************************************
image

image


Without DMA Uncharted 3 in-game

StevenMiller123:
Nothing I can do about that crash, it's due to how long shader recompilation takes.

squidbus (approved):
The spec says:
> A resource set to all zeros acts as an unbound texture or buffer (return 0,0,0,0). Buffer Size (in bytes) =
(stride==0) ? num_elements : stride * num_elements.

So I assume this is alright to do as a comparable check.

PR #4324, opened 2026-04-27
Remove unused code from NTAPI

yellowsparkles584:


georgemoralis:
ok sure but that doesn't actually fix something? probably the removal code is some leftovers but we might need it again...

PR #4328, opened 2026-04-28
Add support for LsHsEsGs (tess + geometry) pipelines where tess in on…

baggins183:
…chip and geometry is offchip

Also refactor RuntimeInfo so some info specific to a logical stage can be shared across multiple HW stage RuntimeInfos. For example the TES can be put in ES or VS depending on whether GS is enabled.

Derive the geometry shader input primitive type from the topology set for the domain shader (not the abstract patch).

Mostly fixes trees in gravity rush 2. Tess factors are being read from buffers and don't seem accurate to the ps4 which results in more noticable pop-in for leaves. I think there's some other bug causing that.

Thanks to @junminlee2004 for finding that gr2 uses tess+geo

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv.cpp
@@ -373,13 +373,12 @@ void DefineEntryPoint(const Info& info, EmitContext& ctx, Id main) {
         break;
     case LogicalStage::TessellationEval: {
         execution_model = spv::ExecutionModel::TessellationEvaluation;
-        const auto& vs_info = ctx.runtime_info.vs_info;
-        ctx.AddExecutionMode(main, ExecutionMode(vs_info.tess_type));
-        ctx.AddExecutionMode(main, ExecutionMode(vs_info.tess_partitioning));
-        ctx.AddExecutionMode(main,
-                             vs_info.tess_topology == AmdGpu::TessellationTopology::TriangleCcw
-                                 ? spv::ExecutionMode::VertexOrderCcw
-                                 : spv::ExecutionMode::VertexOrderCw);
+        ctx.AddExecutionMode(main, ExecutionMode(ctx.runtime_info.es_vs_info.tess_type));
+        ctx.AddExecutionMode(main, ExecutionMode(ctx.runtime_info.es_vs_info.tess_partitioning));
+        ctx.AddExecutionMode(main, ctx.runtime_info.es_vs_info.tess_topology ==
+                                           AmdGpu::TessellationTopology::TriangleCcw
+                                       ? spv::ExecutionMode::VertexOrderCcw
+                                       : spv::ExecutionMode::VertexOrderCw);
         break;
     }
     case LogicalStage::Fragment:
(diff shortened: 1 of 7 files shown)


baggins183:
Needs decent testing because of the RuntimeInfo changes and I guarantee I broke shader cache (too bad lol)

StevenMiller123:
To be fair, shader cache was already broken for most things 😆

junminlee2004:
By the way, there's more low hanging fruit in terms of compatibility I could suggest. For fixing the green line artifacts in videos and cutscenes you need to edit the avplayer_source.cpp to do three things 
1.  Pitch alignment - PS4 GPU linear 8bpp textures need pitch aligned to 64 bytes. CopyNV12Data should write every row at stride = AlignUp(width, 64), and report this as .pitch. 
2. Height alignment - On real PS4, the video decoder always outputs height aligned to at least 16. This creates a guard band of black luma rows between the Y and UV planes. Without it, when the game's NV12 to RGB shader samples the bottom edge of the Y plane with bilinear filtering, it reads into the UV plane - UV bytes interpreted as luma produce the green/colored artifact. Height should be always AlignUp(height, 16) with crop_bottom_offset = aligned_height - real_height telling the game where the actual content ends. 
4. Pre-fill — The entire buffer should be pre-filled before frame copy: luma region with 0x00 (black), chroma region with 0x80 (neutral). This covers all padding areas: right-side pitch padding, bottom height padding, and any GPU overread.
The old use_vdec2 conditional alignment was the root issue because it skipped alignment entirely for the vdec2 path, leaving zero guard band between planes.

baggins183:
Thanks Im not familiar with avplayer at all so someone else will have to look

junminlee2004:
Also for the missing cel shader outline effect, I found that the reason it was missing was because of missing vertex buffers for the ES shader as revealed by vulkan validation: 
VUID-vkCmdDrawIndexed-Input-07939: Vertex shader uses input at location 0, 
but it was not provided with vkCmdSetVertexInputEXT().This requires tweaks in specialization.h and translate.cpp

roamic:
> By the way, there's more low hanging fruit in terms of compatibility I could suggest. For fixing the green line artifacts in videos and cutscenes you need to edit the avplayer_source.cpp to do three things
> 
>     1. Pitch alignment - PS4 GPU linear 8bpp textures need pitch aligned to 64 bytes. CopyNV12Data should write every row at stride = AlignUp(width, 64), and report this as .pitch.
> 
>     2. Height alignment - On real PS4, the video decoder always outputs height aligned to at least 16. This creates a guard band of black luma rows between the Y and UV planes. Without it, when the game's NV12 to RGB shader samples the bottom edge of the Y plane with bilinear filtering, it reads into the UV plane - UV bytes interpreted as luma produce the green/colored artifact. Height should be always AlignUp(height, 16) with crop_bottom_offset = aligned_height - real_height telling the game where the actual content ends.
> 
>     3. Pre-fill — The entire buffer should be pre-filled before frame copy: luma region with 0x00 (black), chroma region with 0x80 (neutral). This covers all padding areas: right-side pitch padding, bottom height padding, and any GPU overread.
>        The old use_vdec2 conditional alignment was the root issue because it skipped alignment entirely for the vdec2 path, leaving zero guard band between planes.

I think most of it is already fixed in a @Kravickas fork, we're still waiting for a PR though.

mikusp:
No regressions in my library

PR #4330, opened 2026-04-28
Uber web_api module fix

georgemoralis:
fixed truncate of request_id data . All findAndValidateContext context was creating nullptr context and was returning ORBIS_NP_WEBAPI_ERROR_LIB_CONTEXT_NOT_FOUND . Now at least it will go further :D


StevenMiller123 (approved):
Thanks for finding this.

PR #4336, opened 2026-04-30
Common: Restore old log filter behavior

StevenMiller123:
Not sure why this had to change in the first place, so this PR restores the old filter formatting, and brings back the old wildcard.

Would probably be good for @Niram7777 to review this, in case there's some good reason to keep the new filter format.

### DIFF
--- a/src/common/logging/log.cpp
+++ b/src/common/logging/log.cpp
@@ -204,8 +204,10 @@ void Setup(std::string_view log_filename) {
             const auto class_level_pair =
                 std::views::split(class_level, ':') | std::ranges::to<std::vector<std::string>>();
 
-            if (class_level_pair.size() == 1) {
-                default_log_level = spdlog::level_from_str(class_level_pair.front() |
+            ASSERT_MSG(class_level_pair.size() == 2, "bad log filter provided");
+
+            if (class_level_pair.front()[0] == '*') {
+                default_log_level = spdlog::level_from_str(class_level_pair.back() |
                                                            std::ranges::to<std::string>());
             } else {
                 log_level_per_class[class_level_pair.front() | std::ranges::to<std::string>()] =
@@ -222,16 +224,10 @@ void Setup(std::string_view log_filename) {
                                    : (EmulatorSettings.IsLogSync() ? sinks : async_sink));
 
         if (EmulatorSettings.IsLogEnable()) {
-            const auto wildcard_it = log_level_per_class.find(std::string("*"));
             const auto level_it = log_level_per_class.find(std::string(name));
 
-            if (level_it != log_level_per_class.end()) {
-                logger->set_level(level_it->second);
-            } else if (wildcard_it != log_level_per_class.end()) {
-                logger->set_level(wildcard_it->second);
-            } else {
-                logger->set_level(default_log_level);
-            }
+            logger->set_level(level_it != log_level_per_class.end() ? level_it->second
+                                                                    : default_log_level);
         } else {
             logger->set_level(spdlog::level::off);
         }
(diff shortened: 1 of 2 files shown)


src/common/logging/log.cpp
```diff
@@ -200,9 +200,9 @@ void Setup(std::string_view log_filename) {
...
 
     if (EmulatorSettings.IsLogEnable()) {
-        for (const auto class_level : std::views::split(EmulatorSettings.GetLogFilter(), ',')) {
+        for (const auto class_level : std::views::split(EmulatorSettings.GetLogFilter(), ' ')) {
             const auto class_level_pair =
-                std::views::split(class_level, '=') | std::ranges::to<std::vector<std::string>>();
+                std::views::split(class_level, ':') | std::ranges::to<std::vector<std::string>>();
 
             if (class_level_pair.size() == 1) {
                 default_log_level = spdlog::level_from_str(class_level_pair.front() |
```
Niram7777: instead of
```cpp
if (class_level_pair.size() == 1) {
    default_log_level = spdlog::level_from_str(class_level_pair.front() |
```
you can put
```cpp
ASSERT_MSG(class_level_pair.size() == 2, "bad log filter provided");

if (class_level_pair.front()[0] == '*') {
    default_log_level = spdlog::level_from_str(class_level_pair.back() |
```

so that you dont have to modify L.225-234

Niram7777 (changes requested):
the filter description should be reverted in
https://github.com/shadps4-emu/shadPS4/blob/main/documents/Debugging/Debugging.md

the qtlauncher should be aligned (but I can do is this weekend otherwise)

StevenMiller123:
I've updated the documentation and made the suggested code change, let me know if there's other changes I'm missing.

Before any sort of merge, it would probably be good to see how others feel. If others prefer the newer style, then I can just close this.

src/common/logging/log.cpp
```diff
@@ -200,12 +200,14 @@ void Setup(std::string_view log_filename) {
...
     if (EmulatorSettings.IsLogEnable()) {
-        for (const auto class_level : std::views::split(EmulatorSettings.GetLogFilter(), ',')) {
+        for (const auto class_level : std::views::split(EmulatorSettings.GetLogFilter(), ' ')) {
             const auto class_level_pair =
-                std::views::split(class_level, '=') | std::ranges::to<std::vector<std::string>>();
+                std::views::split(class_level, ':') | std::ranges::to<std::vector<std::string>>();
 
-            if (class_level_pair.size() == 1) {
-                default_log_level = spdlog::level_from_str(class_level_pair.front() |
+            ASSERT_MSG(class_level_pair.size() == 2, "bad log filter provided");
```
Niram7777: actually the log is not initialized yet, so we have to use intstead
```cpp
if (class_level_pair.size() != 2) {
 std::cerr <<  "bad log filter provided" << std::endl;
  continue; or quick_exit(1); // up to you (I think better to stop than continue)
}
```

StevenMiller123: Good catch. I'll just go for a continue, otherwise people who set up filters with the newer style would have issues.

Might as well save everyone time by just ignoring the bad filter :smile:

StevenMiller123:
Great, a lovely case of "works on my machine" 😩 
I'll fix compile when I have a chance, it's probably just a missing import.

StevenMiller123:
@Niram7777 I've made a PR for the QtLauncher side https://github.com/shadps4-emu/shadps4-qtlauncher/pull/337

Niram7777 (approved):

PR #4337, opened 2026-04-30
Change ASSERT macros to be do while loops instead of lambdas to preserve original function names in the log

kalaposfos13:
Recompilation speeds will be horrendous due to me touching a header that's used literally everywhere, but it is what it is.


georgemoralis:
challenge taken

PR #4340, opened 2026-05-01
Implement sceGnmInsertDingDongMarker

Kravickas:
Use in Marvel's Avengers.
Stub returned ORBIS_OK without writing. Game advanced its ACB write pointer by 4 dwords expecting a marker. Parser hit zeros, faulted on PM4 type-0. 
Now returns -1 on bad args, otherwise writes a 4-dword type-3 NOP and returns 0. No PM4 type-0 crash.


squidbus (approved):
Seems to match disassembly

PR #4341, opened 2026-05-01
Do not split targets when tests are enabled

mikusp:
ENABLE_TESTS now only enables tests and not removes the whole shadps4 executable from the configuration. This helps when switching back and forth between building tests and the emulator - previously I had to delete cache and reconfigure each time

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -193,9 +193,9 @@ endforeach()
 
 include(GoogleTest)
 
+add_custom_target(tests)
+
 foreach(t ${TEST_TARGETS})
-    gtest_discover_tests(${t}
-        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-        PROPERTIES TIMEOUT 60
-    )
+    add_dependencies(tests ${t})
+    add_test(NAME ${t} COMMAND ${t})
 endforeach()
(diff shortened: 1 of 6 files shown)


tests/CMakeLists.txt
```diff
@@ -193,9 +193,9 @@ endforeach()
...
 
+add_custom_target(tests)
+
 foreach(t ${TEST_TARGETS})
-    gtest_discover_tests(${t}
-        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-        PROPERTIES TIMEOUT 60
-    )
+    add_dependencies(tests ${t})
+    add_test(NAME ${t} COMMAND ${t})
```
Niram7777, line 200: I think it is better instead to put all the tests in the same executable

any reason for not running GcnTest on CI ?

mikusp: These are not run as there is no Vulkan device available, I've run some tests with lavapipe but the results were different to what is expected 😢

Niram7777: I still think all the tests should be in the same executable, this way we can use gtest without ctest to run all at once.
(otherwise you have to change `-E 'GcnTest'` by `-E 'shadps4_gcn_test'`)

-> any reason for replacing gtest_discover_tests by add_test ?

georgemoralis:
is that any use or should be close?

georgemoralis:
needs manual rebase @mikusp

PR #4069, opened 2026-02-22
Spdlog migration

Niram7777:
TODO:
- [x] filter
- [x] duplicate
- [x] update doc
- [x] group log config
- [x] refactor to avoid get
- [x] someone should test Windows
- [x] fix truncated log on SDL exit button (already was there before)
- [x] async

Various info:
- in async, the log thread is the 3rd one but it lost his name (thread 1 Main, thread 2 Tracy)
- bitfields have to be promoted to underlying type with + (I havent found variadic template with bitfield overloading possible)
- default:
```
[Log]
append = false
enable = true
filter = ""
maxSkipDuration = 5000
separate = false
sizeLimit = 104857600
skipDuplicate = true
sync = true
type = "wincolor" # Windows only
```
- log level first letter is lowercase
- maybe we should use utf-8 on Windows? https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page

Screenshot From 2026-02-22 18-12-06

PR #4342, opened 2026-05-02
Log read setting from old toml in Debug/General

Niram7777:
https://github.com/shadps4-emu/shadPS4/pull/4069#discussion_r3169597320

### DIFF
--- a/src/core/emulator_settings.cpp
+++ b/src/core/emulator_settings.cpp
@@ -545,6 +545,29 @@ bool EmulatorSettingsImpl::TransferSettings() {
 #endif
     }
 
+    if (og_data.contains("General")) {
+        const toml::value& general = og_data.at("General");
+        auto& s = m_log;
+
+        setFromToml(s.filter, general, "logFilter");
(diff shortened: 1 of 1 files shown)


src/core/emulator_settings.cpp
```diff
@@ -545,6 +545,29 @@ bool EmulatorSettingsImpl::TransferSettings() {
 #endif
     }
```
kalaposfos13, line 546: These should be removed, as they still don't do anything.

Niram7777: look inside your toml you will find them

kalaposfos13: Oh, you are correct, my bad. I guess that got added just a few commits before the config refactor? I don't remember interacting with that much.

Niram7777: spdlog was right after the config refactor
but [Log] section is done both in the toml and in the json
people will have in their toml both the old and new options:
- if they modified old options we will recover
- if they modified new options, they should remove the old options (or put the same values in old/new)

-> for simplicity at when we did the config refactor, we should just had removed the toml, and communicate

PR #4346, opened 2026-05-03
Do not crash on encountering an unknown option in PosixSocket::SetSocketOptions

kalaposfos13:
This makes Vue After Free not crash on opening its remote payload sender connection.

### DIFF
--- a/src/core/libraries/network/posix_sockets.cpp
+++ b/src/core/libraries/network/posix_sockets.cpp
@@ -590,7 +590,7 @@ int PosixSocket::SetSocketOptions(int level, int optname, const void* optval, u3
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/core/libraries/network/posix_sockets.cpp
```diff
@@ -590,7 +590,7 @@ int PosixSocket::SetSocketOptions(int level, int optname, const void* optval, u3
         }
     }
 
-    UNREACHABLE_MSG("Unknown level ={} optname ={}", level, optname);
```
StevenMiller123, line 593: Would probably be best to make this return an appropriate error instead of 0, for any cases that are genuinely using bad optname values.

mikusp (approved):

PR #4349, opened 2026-05-04
Vulkan add VK_EXT_DYNAMIC_RENDERING_UNUSED_ATTACHMENTS

Niram7777:
will fix when different formats are mixed

> [Render.Vulkan] <Error> (shadPS4:GpuComm) vk_platform.cpp:57 DebugUtilsCallback: VUID-vkCmdDrawIndexed-dynamicRenderingUnusedAttachments-08914: vkCmdDrawIndexed(): VkRenderingInfo::pDepthAttachment->imageView format (VK_FORMAT_D32_SFLOAT_S8_UINT) must match the corresponding format in VkPipelineRenderingCreateInfo::depthAttachmentFormat (VK_FORMAT_D32_SFLOAT).
The Vulkan spec states: If current render pass instance was begun with vkCmdBeginRendering, the dynamicRenderingUnusedAttachments feature is not enabled, and VkRenderingInfo::pDepthAttachment->imageView was not VK_NULL_HANDLE, the value of VkPipelineRenderingCreateInfo::depthAttachmentFormat used to create the bound graphics pipeline must be equal to the VkFormat used to create VkRenderingInfo::pDepthAttachment->imageView (https://docs.vulkan.org/spec/latest/chapters/drawing.html#VUID-vkCmdDrawIndexed-dynamicRenderingUnusedAttachments-08914)

### DIFF
--- a/src/video_core/renderer_vulkan/vk_instance.cpp
+++ b/src/video_core/renderer_vulkan/vk_instance.cpp
@@ -289,6 +289,8 @@ bool Instance::CreateDevice() {
     amd_shader_trinary_minmax = add_extension(VK_AMD_SHADER_TRINARY_MINMAX_EXTENSION_NAME);
     nv_framebuffer_mixed_samples = add_extension(VK_NV_FRAMEBUFFER_MIXED_SAMPLES_EXTENSION_NAME);
     amd_mixed_attachment_samples = add_extension(VK_AMD_MIXED_ATTACHMENT_SAMPLES_EXTENSION_NAME);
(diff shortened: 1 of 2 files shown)


squidbus:
This does not sound correct, that extension is for when the attachment slot is unused which it is not here.

I think you should focus more on why the formats are mismatching in the first place and getting them to match.

georgemoralis:
is that any used? if so needs to be rebase

Niram7777:
> is that any used? if so needs to be rebase

I am currently investigating what is happening with the attachments/slots and I am writting a shader debugger, this PR is probably useless

PR #4350, opened 2026-05-04
Http module implementation

georgemoralis:
- Added extensive logging
- Implementation of http calls
- Async state machine
- Openssl added
- Mockup up webapi calls (fake empty returns)


georgemoralis:
Here some proof of concept , bloodborne tries to connect to real server now (of course as expected it fails) 

εικόνα

georgemoralis:
[libssl-3-x64.zip](https://github.com/user-attachments/files/27368344/libssl-3-x64.zip)

added missing dlls if you miss them (need to fix ci)

StevenMiller123:
Should probably be marked as a draft until you properly fix CI and mac builds?

CMakeLists.txt
```diff
@@ -1257,6 +1255,15 @@ if (WIN32)
     target_sources(shadps4 PRIVATE src/shadps4.rc)
 endif()
 
+if (WIN32 AND OPENSSL_RUNTIME_DLLS)
+    foreach(dll ${OPENSSL_RUNTIME_DLLS})
+        add_custom_command(TARGET shadps4 POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E copy_if_different
+                    "${dll}" "$<TARGET_FILE_DIR:shadps4>"
+            COMMENT "Copying OpenSSL runtime DLL: ${dll}")
+    endforeach()
```
Niram7777: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_RUNTIME_DLLS

externals/CMakeLists.txt
```diff
@@ -331,3 +331,67 @@ endif()
...
+        file(GLOB OPENSSL_DLLS
+             "${OPENSSL_ROOT_DIR}/bin/libssl-*.dll"
+             "${OPENSSL_ROOT_DIR}/bin/libcrypto-*.dll")
+        if (OPENSSL_DLLS)
+            set(OPENSSL_RUNTIME_DLLS "${OPENSSL_DLLS}" CACHE INTERNAL
+                "OpenSSL DLLs to copy next to shadps4.exe (consumed by top-level CMakeLists)")
+            list(LENGTH OPENSSL_DLLS OPENSSL_DLL_COUNT)
+            message(STATUS "cpp-httplib: ${OPENSSL_DLL_COUNT} OpenSSL DLL(s) staged in OPENSSL_RUNTIME_DLLS for post-build copy")
+        endif()
+    endif()
```
Niram7777: not needed if you follow my link

.github/workflows/build.yml
```diff
@@ -176,11 +176,20 @@ jobs:
     - name: Build
       run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --parallel $env:NUMBER_OF_PROCESSORS
 
+    - name: Stage Windows SDL artifact
+      shell: pwsh
+      run: |
+        New-Item -ItemType Directory -Force -Path "${{github.workspace}}/upload" | Out-Null
+        Copy-Item "${{github.workspace}}/build/shadPS4.exe" "${{github.workspace}}/upload/"
+        Get-ChildItem "${{github.workspace}}/build/libssl-*-x64.dll" -ErrorAction SilentlyContinue | ForEach-Object { Copy-Item $_.FullName "${{github.workspace}}/upload/" }
+        Get-ChildItem "${{github.workspace}}/build/libcrypto-*-x64.dll" -ErrorAction SilentlyContinue | ForEach-Object { Copy-Item $_.FullName "${{github.workspace}}/upload/" }
```
Niram7777: those paths dont exist apparently (the zip only contains the .exe)
one `Get-ChildItem "${{github.workspace}}/build/*.dll"....` might be easier
(again I would recommend following my link)

metr1k-91:
[CUSA00192.txt](https://github.com/user-attachments/files/27380147/CUSA00192.txt)
[CUSA00900.txt](https://github.com/user-attachments/files/27380148/CUSA00900.txt)
[CUSA46747.txt](https://github.com/user-attachments/files/27380149/CUSA46747.txt)


I did not notice anything out of the ordinary.  The best testing we would have for this is bloodborne at this time, which stops short of ingame at the login / auth piece because of missing onlineid for the platformaccountid field.

I limited testing specific to this PR, so this is as much as I can provide.  Per wireshark data, the call/returns look good.

georgemoralis:
Macos openssl issues probably fixed if anyone can test . 

Also dll added in windows artifacts

externals/CMakeLists.txt
```diff
@@ -331,3 +331,73 @@ endif()
 add_library(Cpp_Httplib INTERFACE)
 target_include_directories(Cpp_Httplib INTERFACE cpp-httplib/)
 
+if (APPLE AND NOT OPENSSL_ROOT_DIR)
+    execute_process(
+        COMMAND brew --prefix openssl
```
squidbus: These brew calls will likely go to native ARM brew install and fetch the wrong library architecture. I don't think we should need to fetch from brew here at all, standard find mechanism should be fine.

externals/CMakeLists.txt
```diff
@@ -331,3 +331,73 @@ endif()
...
+    endif()
+    message(STATUS "macOS: using caller-supplied OPENSSL_ROOT_DIR=${OPENSSL_ROOT_DIR}")
+endif()
+
+find_package(OpenSSL QUIET)
+
+if (NOT OPENSSL_FOUND AND MSVC)
+    message(STATUS "OpenSSL not found in system, downloading FireDaemon prebuilt for Windows...")
+    if (NOT EXISTS "${CMAKE_BINARY_DIR}/external/openssl.zip")
+        file(DOWNLOAD https://download.firedaemon.com/FireDaemon-OpenSSL/openssl-3.6.1.zip
```
squidbus: If you're going to download from someone's OpenSSL pre-builts you should really at least check the SHA-256 matches a known value. But I don't know why we're downloading OpenSSL from whoever FireDaemon is to begin with.

squidbus (commented):
Using OpenSSL seems to be its own can of worms. I'm not sure why we're not using something that can be built as a submodule like LibreSSL or mbedtls, but if you're going to need OpenSSL then I think we should at least be doing a pre-built setup similar to what we have for FFmpeg, where we have our own repo that builds and publishes them for all targets.

externals/CMakeLists.txt
georgemoralis: i used vita3k's way of doing that , openssl is kinda hard to be manualy configured so there aren't actually official mirrors of it

georgemoralis:
> Using OpenSSL seems to be its own can of worms. I'm not sure why we're not using something that can be built as a submodule like LibreSSL or mbedtls, but if you're going to need OpenSSL then I think we should at least be doing a pre-built setup similar to what we have for FFmpeg, where we have our own repo that builds and publishes them for all targets.

lib-http has build-in support for openssl that's why i used that

externals/CMakeLists.txt
georgemoralis: yah that probably won't work , vita3k is using native openssl not x86-64 probably need something similar like i did with ci

Niram7777:
> > Using OpenSSL seems to be its own can of worms. I'm not sure why we're not using something that can be built as a submodule like LibreSSL or mbedtls, but if you're going to need OpenSSL then I think we should at least be doing a pre-built setup similar to what we have for FFmpeg, where we have our own repo that builds and publishes them for all targets.
> 
> lib-http has build-in support for openssl that's why i used that

since we are already using boost, I would have added beast with openssl (other SSL libs tend to have handshake issues in my experience)
(fyi its possible to add add beast without openssl)

squidbus:
> lib-http has build-in support for openssl that's why i used that

Last I checked it also has LibreSSL/BoringSSL compat

externals/CMakeLists.txt
georgemoralis: @squidbus pushed a fix now it should work ok i assume (don't have Mac to test) . Checked httplib header it also supports wolfssl and some others but i don't think that they are less complex than openssl...

squidbus: LibreSSL is accounted for here: https://github.com/shadexternals/cpp-httplib/blob/master/httplib.h#L402

Which you can run autogen and throw into a submodule repo then import as a CMake submodule: https://github.com/libressl/portable#building-from-a-git-checkout

squidbus: Can use this if you want and copy my submodule repo, works for me: https://github.com/squidbus/shadPS4/commit/964b0dedb33c076283edc97063c4ed48a8be7ce5

src/core/libraries/network/http.cpp
```diff
@@ -64,8 +443,50 @@ int PS4_SYSV_ABI sceHttpAddQuery() {
...
+    if (mode == SCE_HTTP_HEADER_OVERWRITE) {
+        (*target)[name] = value;
+    } else {
+        // SCE_HTTP_HEADER_ADD: refuse if the header already exists at this level.
+        if (target->contains(name)) {
+            LOG_ERROR(Lib_Http, "Header '{}' already exists at {} level in ADD mode", name, level);
+            return ORBIS_HTTP_ERROR_INVALID_VALUE;
+        }
+        target->emplace(name, value);
+    }
```
Niram7777: (regarding the discord comment)

instead ?
```cpp
} else if (mode == SCE_HTTP_HEADER_ADD) {
    // refuse if the header already exists at this level.
    if (target->contains(name)) {
        LOG_WARN(Lib_Http, "Header '{}' already exists at {} level in ADD mode", name, level);
        return ORBIS_HTTP_ERROR_INVALID_VALUE;
    }
    target->emplace(name, value);
} else {
    LOG_WARN(Lib_Http, "Header '{}' with unknown mode {}", name, mode);
    return ORBIS_OK;
}
```

maybe there is mode DELETE ?

~~maybe g_state.requests should be cleared when the request has been made?~~ (I guess `id` is unique so ok)

StevenMiller123: I guess the big question is how the actual library handles this case.

I'm decompiling stuff now to generally validate how this PR works (though that's probably going to take some time)

StevenMiller123: No, there's just the overwrite and add modes. The issue is probably somewhere else (or there's something this PR does wrong in this function).

src/core/libraries/network/http.cpp
```diff
@@ -282,75 +1186,120 @@ int PS4_SYSV_ABI sceHttpGetRegisteredCtxIds() {
...
-              libnetMemId, libsslCtxId, poolSize);
-    // return a value >1
-    static int id = 0;
-    return ++id;
+    LOG_INFO(Lib_Http, "called libnetMemId={}, libsslCtxId={}, poolSize={}", libnetMemId,
+             libsslCtxId, poolSize);
+    std::lock_guard<std::mutex> lock(g_state.m_mutex);
+    if (poolSize == 0) {
+        LOG_ERROR(Lib_Http, "poolSize is 0");
+        return ORBIS_HTTP_ERROR_INVALID_VALUE;
```
StevenMiller123: Decomp suggests this would end up returning ORBIS_KERNEL_ERROR_EINVAL through a sceKernelMapNamedFlexibleMemory error.

src/core/libraries/network/http.cpp
```diff
@@ -64,8 +443,50 @@ int PS4_SYSV_ABI sceHttpAddQuery() {
...
+        target = &it->second->headers;
+        level = "request";
+    }
+    if (!target) {
+        LOG_ERROR(Lib_Http, "Invalid id={} (not a template, connection, or request)", id);
+        return ORBIS_HTTP_ERROR_INVALID_ID;
+    }
+
+    if (mode == SCE_HTTP_HEADER_OVERWRITE) {
+        (*target)[name] = value;
```
StevenMiller123: Based on what I've decompiled, headers are only stored in the template. For request and connection ids, the library gets the template from that object.

src/core/libraries/network/http.cpp
```diff
@@ -64,8 +443,50 @@ int PS4_SYSV_ABI sceHttpAddQuery() {
...
+    } else {
+        // SCE_HTTP_HEADER_ADD: refuse if the header already exists at this level.
+        if (target->contains(name)) {
```
StevenMiller123: At least in sceHttpAddRequestHeader, there is not a check for duplicate headers.

src/core/libraries/network/http.cpp
```diff
@@ -89,67 +510,412 @@ int PS4_SYSV_ABI sceHttpAuthCacheImport() {
...
+        return ORBIS_HTTP_ERROR_BEFORE_INIT;
+    }
+
+    if (!g_state.templates.contains(tmplId)) {
+        LOG_ERROR(Lib_Http, "Invalid tmplId={}", tmplId);
+        return ORBIS_HTTP_ERROR_INVALID_ID;
+    }
+    if (!scheme) {
+        LOG_ERROR(Lib_Http, "scheme is null");
+        return ORBIS_HTTP_ERROR_UNKNOWN_SCHEME;
```
StevenMiller123: null scheme should return `ORBIS_HTTP_ERROR_INVALID_VALUE`

src/core/libraries/network/http.cpp
```diff
@@ -89,67 +510,412 @@ int PS4_SYSV_ABI sceHttpAuthCacheImport() {
...
+    }
+
+    std::string scheme_str = scheme;
+    if (scheme_str != "http" && scheme_str != "https") {
+        LOG_ERROR(Lib_Http, "Unknown scheme: {}", scheme_str);
+        return ORBIS_HTTP_ERROR_UNKNOWN_SCHEME;
+    }
+    if (!serverName) {
```
StevenMiller123, line 1017: null check for server name should come before the scheme check, and also returns `ORBIS_HTTP_ERROR_INVALID_VALUE`

src/core/libraries/network/http.cpp
```diff
@@ -89,67 +510,412 @@ int PS4_SYSV_ABI sceHttpAuthCacheImport() {
...
+        LOG_ERROR(Lib_Http, "serverName is null");
+        return ORBIS_HTTP_ERROR_INVALID_URL;
+    }
+    if (port == 0) {
+        if (scheme && std::strcmp(scheme, "https") == 0) {
```
StevenMiller123: Check should not be case sensitive and should use a function with proper length limits. Real library uses strncasecmp.
There's also no point in checking for non-null scheme here.

src/core/libraries/network/http.cpp
```diff
@@ -89,67 +510,412 @@ int PS4_SYSV_ABI sceHttpAuthCacheImport() {
...
+        return ORBIS_HTTP_ERROR_INVALID_ID;
+    }
+    if (!scheme) {
+        LOG_ERROR(Lib_Http, "scheme is null");
+        return ORBIS_HTTP_ERROR_UNKNOWN_SCHEME;
+    }
+
+    std::string scheme_str = scheme;
+    if (scheme_str != "http" && scheme_str != "https") {
+        LOG_ERROR(Lib_Http, "Unknown scheme: {}", scheme_str);
```
StevenMiller123: Same deal as my comment for line 566.

src/core/libraries/network/http.cpp
```diff
@@ -89,67 +510,412 @@ int PS4_SYSV_ABI sceHttpAuthCacheImport() {
...
+    }
+
+    int conn_id = ++g_state.next_obj_id;
+    HttpConnection conn;
+    conn.tmpl_id = tmplId;
+    conn.scheme = scheme_str;
+    conn.hostname = serverName;
+    conn.port = port;
+    conn.keep_alive = (isEnableKeepalive != 0);
+    conn.is_secure = (scheme_str == "https");
```
StevenMiller123: Again, shouldn't be case sensitive.

src/core/libraries/network/http.cpp
```diff
@@ -89,67 +510,412 @@ int PS4_SYSV_ABI sceHttpAuthCacheImport() {
...
+    if (parse_ret < 0) {
+        LOG_ERROR(Lib_Http, "URI parse failed: {:#x}", parse_ret);
+        return parse_ret;
+    }
 
-int PS4_SYSV_ABI sceHttpCreateRequest() {
-    LOG_ERROR(Lib_Http, "(STUBBED) called");
-    return ORBIS_OK;
+    std::string scheme_str = parsed.scheme ? parsed.scheme : "";
+    if (scheme_str != "http" && scheme_str != "https") {
```
StevenMiller123: Would be nice to avoid code duplication here. From this point onward, this should be the same behavior as sceHttpCreateConnection (and of course, also suffers the exact same complaints).

src/core/libraries/network/http.cpp
```diff
@@ -89,67 +510,412 @@ int PS4_SYSV_ABI sceHttpAuthCacheImport() {
...
+            return SCE_HTTP_METHOD_CONNECT;
+        return -1;
+    };
+    if (!path) {
+        LOG_ERROR(Lib_Http, "path is null");
+        return ORBIS_HTTP_ERROR_INVALID_VALUE;
+    }
+    int int_method = map_method(method);
+    if (int_method < 0) {
+        LOG_ERROR(Lib_Http, "Custom/unknown method '{}' dummy return GET", // TODO
```
StevenMiller123: No such thing as a custom/unknown method here, this case just returns ORBIS_HTTP_ERROR_UNKNOWN_METHOD.

src/core/libraries/network/http.cpp
```diff
@@ -89,67 +510,412 @@ int PS4_SYSV_ABI sceHttpAuthCacheImport() {
...
+    }
+    if (!url) {
+        LOG_ERROR(Lib_Http, "url is null");
+        return ORBIS_HTTP_ERROR_INVALID_URL;
+    }
 
-int PS4_SYSV_ABI sceHttpCreateRequestWithURL2() {
-    LOG_ERROR(Lib_Http, "(STUBBED) called");
-    return ORBIS_OK;
+    int req_id = ++g_state.next_obj_id;
```
StevenMiller123, line 1287: Real library has additional checks for newlines in any component of the URL, which trigger `ORBIS_HTTP_ERROR_INVALID_VALUE` returns.

src/core/libraries/network/http.cpp
```diff
@@ -89,67 +510,412 @@ int PS4_SYSV_ABI sceHttpAuthCacheImport() {
...
+    std::lock_guard<std::mutex> lock(g_state.m_mutex);
+    if (!g_state.inited) {
+        LOG_ERROR(Lib_Http, "Not initialized");
+        return ORBIS_HTTP_ERROR_BEFORE_INIT;
+    }
+    if (!g_state.active_contexts.contains(libhttpCtxId)) {
+        LOG_ERROR(Lib_Http, "Invalid libhttpCtxId={} (not in active contexts)", libhttpCtxId);
+        return ORBIS_HTTP_ERROR_INVALID_ID;
+    }
+    if (!userAgent) {
```
StevenMiller123: Null user agent is allowed in the actual library.

StevenMiller123 (changes requested):
I haven't gotten too far in decomp, but there's a lot of discrepancies I'm seeing so far.

src/core/libraries/network/http.cpp
StevenMiller123: Looking further at decomp, storing separately seems to be accurate. My bad.

georgemoralis:
@StevenMiller123 I think I fixed most of your reviews if you wanna check

src/core/libraries/network/http.cpp
```diff
@@ -1434,8 +1523,9 @@ int PS4_SYSV_ABI sceHttpInit(int libnetMemId, int libsslCtxId, u64 poolSize) {
              libsslCtxId, poolSize);
     std::lock_guard<std::mutex> lock(g_state.m_mutex);
     if (poolSize == 0) {
-        LOG_ERROR(Lib_Http, "poolSize is 0");
-        return ORBIS_HTTP_ERROR_INVALID_VALUE;
+        LOG_ERROR(Lib_Http, "poolSize=0 (returning ORBIS_KERNEL_ERROR_EINVAL "
```
StevenMiller123: I don't think we need to log this that verbosely 😅

src/core/libraries/network/http.cpp
```diff
@@ -31,326 +74,1449 @@ void NormalizeAndAppendPath(char* dest, char* src) {
...
-              std::string(name), std::string(value), mode);
-    return ORBIS_OK;
+inline const std::pair<std::string, std::string>* FindFirst(const HttpHeaders& headers,
+                                                            std::string_view name) {
+    for (const auto& kv : headers) {
+        if (IEquals(kv.first, name))
+            return &kv;
+    }
+    return nullptr;
 }
```
Niram7777, line 173: template

src/core/libraries/network/http.cpp
```diff
@@ -31,326 +74,1449 @@ void NormalizeAndAppendPath(char* dest, char* src) {
...
+                return true;
+            if (a == '\0' || b == '\0')
+                return false;
+            const char fa = (a >= 'A' && a <= 'Z') ? char(a - 'A' + 'a') : a;
+            const char fb = (b >= 'A' && b <= 'Z') ? char(b - 'A' + 'a') : b;
+            if (fa != fb)
+                return false;
+        }
+        return s[cap] == '\0' && lit[cap] == '\0';
+    };
```
Niram7777, line 195: you added IEquals before (or we could add boost:algorithm iequals)

src/core/libraries/network/http.cpp
```diff
@@ -31,326 +74,1451 @@ void NormalizeAndAppendPath(char* dest, char* src) {
...
+    if (auto it = g_state.connections.find(id); it != g_state.connections.end()) {
+        level = "connection";
+        return &it->second.settings;
+    }
+    if (auto it = g_state.requests.find(id); it != g_state.requests.end()) {
+        level = "request";
+        return &it->second->settings;
+    }
+    level = "";
+    return nullptr;
```
Niram7777, line 382: I imagine if the id is not found we dont need to bother clearing the level
(and we coud set it as std::string_view&)

src/core/libraries/network/http.cpp
```diff
@@ -426,294 +1593,508 @@ int PS4_SYSV_ABI sceHttpParseStatusLine(const char* statusLine, u64 lineLen, int
...
+        // either block on cv (blocking mode) or return EAGAIN (non-blocking).
+        req.state = HttpRequestState::Sending;
+        req_ptr = it->second;
+        plan.method = req.method;
+        plan.method_str = req.method_str;
+        plan.headers = req.headers;
+        plan.body = std::move(body_copy);
+        plan.settings = req.settings;
+        plan.nonblock = req.nonblock;
+        ParseRequestUrl(req.url, plan.scheme, plan.host, plan.port, plan.path);
```
Niram7777, line 1872: I think we should log if parsing failed
(I am wandering if the scheme is really mandatory but I guess we will find out)

src/core/libraries/network/http.cpp
```diff
@@ -31,326 +74,1451 @@ void NormalizeAndAppendPath(char* dest, char* src) {
...
+                if (a.size() != b.size())
+                    return false;
+                for (size_t i = 0; i < a.size(); ++i) {
+                    if (std::tolower(static_cast<unsigned char>(a[i])) !=
+                        std::tolower(static_cast<unsigned char>(b[i]))) {
+                        return false;
+                    }
+                }
+                return true;
+            };
```
Niram7777, line 630: same as above reuse iequals

src/core/libraries/network/http.cpp
```diff
@@ -31,326 +74,1451 @@ void NormalizeAndAppendPath(char* dest, char* src) {
...
+                    // Connection refused before TLS even started). Leave
+                    // ssl_error at 0; sceHttpGetLastErrno still reports it.
+                    break;
+                }
+            }
+            return false;
+        }
+        PopulateRealResponse(out_res, result);
+        // Debug stuff to be removed
+        if (result->status >= 400 && result->body.size() <= 8192) {
```
Niram7777, line 763: you aim to remove it in this PR or you need it in main for some time (the log should be debug in this case)?

src/core/libraries/network/http.cpp
```diff
@@ -31,326 +74,1451 @@ void NormalizeAndAppendPath(char* dest, char* src) {
...
+            return SCE_HTTP_METHOD_POST;
+        if (s == "HEAD")
+            return SCE_HTTP_METHOD_HEAD;
+        if (s == "PUT")
+            return SCE_HTTP_METHOD_PUT;
+        if (s == "DELETE")
+            return SCE_HTTP_METHOD_DELETE;
+        if (s == "TRACE")
+            return SCE_HTTP_METHOD_TRACE;
+        return -1;
```
Niram7777, line 1204: there is no SCE_HTTP_METHOD_CUSTOM here ?

src/core/libraries/network/http.cpp
```diff
@@ -31,326 +74,1451 @@ void NormalizeAndAppendPath(char* dest, char* src) {
...
+            int_method = SCE_HTTP_METHOD_POST;
+        else if (s == "HEAD")
+            int_method = SCE_HTTP_METHOD_HEAD;
+        else if (s == "PUT")
+            int_method = SCE_HTTP_METHOD_PUT;
+        else if (s == "DELETE")
+            int_method = SCE_HTTP_METHOD_DELETE;
+        else if (s == "TRACE")
+            int_method = SCE_HTTP_METHOD_TRACE;
+    }
```
Niram7777, line 1339: same?

src/core/libraries/np/np_web_api_internal.cpp
```diff
@@ -25,6 +27,39 @@ static s64 g_request_count = 0;
...
+static s32 NpMethodToHttpMethod(OrbisNpWebApiHttpMethod method) {
+    switch (method) {
+    case ORBIS_NP_WEBAPI_HTTP_METHOD_GET:
+        return 0; // ORBIS_HTTP_METHOD_GET
+    case ORBIS_NP_WEBAPI_HTTP_METHOD_POST:
+        return 1; // ORBIS_HTTP_METHOD_POST
+    case ORBIS_NP_WEBAPI_HTTP_METHOD_PUT:
+        return 4; // ORBIS_HTTP_METHOD_PUT
+    case ORBIS_NP_WEBAPI_HTTP_METHOD_DELETE:
+        return 5; // ORBIS_HTTP_METHOD_DELETE
```
mikusp: Why not use the symbols for ORBIS_HTTP_*?

src/core/libraries/network/http.h
```diff
@@ -13,6 +13,37 @@ class SymbolsResolver;
 
 namespace Libraries::Http {
 
+enum OrbisHttpsFlags : u32 {
```
mikusp, line 28: Why not enum class?

src/core/libraries/network/http.cpp
```diff
@@ -31,326 +75,1623 @@ void NormalizeAndAppendPath(char* dest, char* src) {
...
+
+constexpr int SCE_HTTP_METHOD_GET = 0;
+constexpr int SCE_HTTP_METHOD_POST = 1;
+constexpr int SCE_HTTP_METHOD_HEAD = 2;
+constexpr int SCE_HTTP_METHOD_OPTIONS = 3;
+constexpr int SCE_HTTP_METHOD_PUT = 4;
+constexpr int SCE_HTTP_METHOD_DELETE = 5;
+constexpr int SCE_HTTP_METHOD_TRACE = 6;
+constexpr int SCE_HTTP_METHOD_CONNECT = 7;
+constexpr int SCE_HTTP_METHOD_CUSTOM = 8;
```
mikusp: Why not make those SCE_HTTP_X_* HttpX enum classes?

brad0demx:
[PR-CUSA24267.log](https://github.com/user-attachments/files/27708945/PR-CUSA24267.log)
[CUSA24267.log](https://github.com/user-attachments/files/27709030/CUSA24267.log)

HTTP spam on Call of Duty Cold War is fixed on this PR, not sure what other logs may be useful

georgemoralis:
it might be worthy to enable isConnectedToNetwork switch (default is false)

src/core/libraries/np/np_web_api_internal.cpp
georgemoralis: done

src/core/libraries/network/http.h
georgemoralis: it won't work with bitmasks

src/core/libraries/network/http.cpp
georgemoralis: done

src/core/libraries/network/http.h
StevenMiller123: If you need support for that, use our DECLARE_ENUM_FLAG_OPERATORS macro. It's part of `src/common/enum.h`

georgemoralis: don't see a reason though

brad0demx:
Thats my bad forgot I needed other options enabled 
[CUSA24267.log](https://github.com/user-attachments/files/27728837/CUSA24267.log)

PR #4361, opened 2026-05-06
perf: Optimize RenderState equality operator

Hog185:
Add fast path checks for RenderState equality operator.
works in theory, havent been able to test for regressions or breakage

### DIFF
--- a/src/video_core/renderer_vulkan/vk_scheduler.h
+++ b/src/video_core/renderer_vulkan/vk_scheduler.h
@@ -47,6 +47,17 @@ struct RenderState {
     u16 num_color_attachments;
 
     bool operator==(const RenderState& other) const noexcept {
+        // Fast path: check most commonly different fields first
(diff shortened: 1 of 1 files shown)


src/video_core/renderer_vulkan/vk_scheduler.h
```diff
@@ -47,6 +47,17 @@ struct RenderState {
...
+            return false;
+        }
+        if (width != other.width || height != other.height) [[likely]] {
+            return false;
+        }
+        if (num_layers != other.num_layers) [[likely]] {
+            return false;
+        }
+        // Full comparison only if quick checks pass
         return std::memcmp(this, &other, sizeof(RenderState)) == 0;
```
squidbus, line 61: I think the `memcmp` is from before this struct was fixed up to be nicer recently. At this point you could probably just try out adding default equals operators (`bool operator==(const RenderState& other) const = default;`) for this and the above `RenderAttachment`, and let the compiler generate optimal comparisons.

PR #4365, opened 2026-05-06
Fix compilation with mingw-w64

oltolm:
These changes are necessary to build shadps4 with mingw-w64. Some changes are needed for GCC and Clang, some are needed for GCC only.

### DIFF
--- a/src/shader_recompiler/frontend/opcodes.h
+++ b/src/shader_recompiler/frontend/opcodes.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <magic_enum/magic_enum.hpp>
+
 #include "common/enum.h"
 #include "common/types.h"
 
@@ -2696,3 +2698,9 @@ struct SendMsgSimm {
 };
 
 } // namespace Shader::Gcn
+
+template <>
+struct magic_enum::customize::enum_range<Shader::Gcn::Opcode> {
+    static constexpr int min = static_cast<int>(Shader::Gcn::Opcode::S_ADD_U32);
+    static constexpr int max = static_cast<int>(Shader::Gcn::Opcode::V_MAD_MIXHI_F16);
+};
--- a/src/shader_recompiler/frontend/translate/translate.cpp
+++ b/src/shader_recompiler/frontend/translate/translate.cpp
@@ -17,8 +17,6 @@
 #include "video_core/amdgpu/resource.h"
 
 #include <numbers>
-#define MAGIC_ENUM_RANGE_MIN 0
-#define MAGIC_ENUM_RANGE_MAX 2143
 #include <magic_enum/magic_enum.hpp>
 
 namespace Shader::Gcn {
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -149,7 +149,7 @@ endif()
 # sirit
 add_subdirectory(sirit)
 if (WIN32)
-    target_compile_options(sirit PRIVATE "-Wno-error=unused-command-line-argument")
+    target_compile_options(sirit PRIVATE $<$<CXX_COMPILER_ID:Clang>:-Wno-error=unused-command-line-argument>)
 endif()
 
 # half
--- a/src/common/thread.cpp
+++ b/src/common/thread.cpp
@@ -170,7 +170,7 @@ bool AccurateSleep(const std::chrono::nanoseconds duration, std::chrono::nanosec
 
 #endif
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 
 // Sets the debugger-visible name of the current thread.
 void SetCurrentThreadName(const char* name) {
@@ -184,7 +184,7 @@ void SetThreadName(void* thread, const char* name) {
     SetThreadDescription(thread, UTF8ToUTF16W(name).data());
 }
 
-#else // !MSVC_VER, so must be POSIX threads
+#else // !_WIN32, so must be POSIX threads
 
 // MinGW with the POSIX threading model does not support pthread_setname_np
 #if !defined(_WIN32) || defined(_MSC_VER)


Niram7777:
I would recommend using clang even for MinGW
https://github.com/shadps4-emu/shadPS4/blob/main/documents/building-windows.md#option-3-msys2mingw
(otherwise the PR looks good to me)

oltolm:
Yes, Clang is better at building shadps4 than GCC right now, but I am working on improving GCC. BTW the build instruction recommend installing Clang in the `MINGW64` environment. That environment is [deprecated](https://www.msys2.org/news/#2026-03-15-deprecating-the-mingw64-environment) and `CLANG64` should be used instead.

src/shader_recompiler/frontend/translate/translate.cpp
```diff
@@ -17,8 +17,6 @@
 #include "video_core/amdgpu/resource.h"
 
 #include <numbers>
-#define MAGIC_ENUM_RANGE_MIN 0
-#define MAGIC_ENUM_RANGE_MAX 2143
```
squidbus, line 21: What is this being moved for? In the header it seems to me it can affect more source files than intended.

oltolm: It's to avoid the warning from GNU ld.

```
CMakeFiles/shadps4.dir/src/shader_recompiler/frontend/decode.cpp.obj: duplicate section `.rdata$_ZN10magic_enum6detail7names_vIN6Shader3Gcn6OpcodeELNS0_12enum_subtypeE0EEE[_ZN10magic_enum6detail7names_vIN6Shader3Gcn6OpcodeELNS0_12enum_subtypeE0EEE]' has different size
CMakeFiles/shadps4.dir/src/shader_recompiler/frontend/decode.cpp.obj: duplicate section `.rdata$_ZN10magic_enum6detail8values_vIN6Shader3Gcn6OpcodeELNS0_12enum_subtypeE0EEE[_ZN10magic_enum6detail8values_vIN6Shader3Gcn6OpcodeELNS0_12enum_subtypeE0EEE]' has different size
build finished with warning(s).
```
```c++
template <>
struct magic_enum::customize::enum_range<Shader::Gcn::Opcode> {
    static constexpr int min = 0;
    static constexpr int max = 2143;
};
```
should affect all source files that need this code, so that the definition is consistent across files. I don't see any harm in this code because it is scoped to `Shader::Gcn::Opcode` and does not use global defines like `MAGIC_ENUM_RANGE_MIN` and `MAGIC_ENUM_RANGE_MAX`.

oltolm: I could even improve it to

```c++
template <>
struct magic_enum::customize::enum_range<Shader::Gcn::Opcode> {
    static constexpr int min = static_cast<int>(Shader::Gcn::Opcode::S_ADD_U32);
    static constexpr int max = static_cast<int>(Shader::Gcn::Opcode::V_MAD_MIXHI_F16);
};
```

if you want.

squidbus: Sure

oltolm: done

src/shader_recompiler/frontend/opcodes.h
```diff
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <magic_enum/magic_enum.hpp>
```
squidbus, line 6: One more question here - does this need to be included in the header for the below to work? Same reason as before, we may not want everything that includes this header to also be forced to pull magic_enum into the compilation.

oltolm: If I don't include it, then I get errors like this
```
[519/655] Building CXX object CMakeFiles/shadps4.dir/src/shader_recompiler/recompiler.cpp.obj
FAILED: [code=1] CMakeFiles/shadps4.dir/src/shader_recompiler/recompiler.cpp.obj
C:\msys64\ucrt64\bin\c++.exe -DBOOST_ASIO_STANDALONE -DENABLE_DISCORD_RPC -DHWINFO_EXPORTS -DIMGUI_USER_CONFIG=\"imgui/imgui_config.h\" -DMINIZ_STATIC_DEFINE -DNOMINMAX -DNTDDI_VERSION=0x0A000006 -DSPDLOG_FUNCTION=__func__ -DTRACY_ENABLE -DTRACY_NO_CALLSTACK -DTRACY_NO_CODE_TRANSFER -DTRACY_NO_CONTEXT_SWITCH -DTRACY_NO_CRASH_HANDLER -DTRACY_NO_FRAME_IMAGE -DTRACY_NO_SAMPLING -DTRACY_NO_SYSTEM_TRACING -DTRACY_ONLY_LOCALHOST -DTRACY_ON_DEMAND -DWIN32_LEAN_AND_MEAN -DWINVER=0x0A00 -DZYCORE_STATIC_BUILD -DZYDIS_STATIC_BUILD -D_WIN32_WINNT=0x0A00 -IC:/src/shadPS4/src -IC:/src/shadPS4 -IE:/build-shadps4-gcc/src/video_core/host_shaders/include -IE:/build-shadps4-gcc/src/imgui/renderer/generated_fonts -IE:/build-shadps4-gcc/_cmrc/include -isystem C:/src/shadPS4/externals/spdlog/include -isystem C:/src/shadPS4/externals/magic_enum/include -isystem C:/src/shadPS4/externals/toml11/include -isystem C:/src/shadPS4/externals/robin-map/include -isystem C:/src/shadPS4/externals/xbyak -isystem C:/src/shadPS4/externals/tracy/public -isystem C:/src/shadPS4/externals/./renderdoc -isystem C:/src/shadPS4/externals/dear_imgui -isystem C:/src/shadPS4/externals/ImGuiFileDialog -isystem C:/src/shadPS4/externals/gcn/include -isystem C:/src/shadPS4/externals/half/include -isystem C:/src/shadPS4/externals/minimp3 -isystem C:/src/shadPS4/externals/vma/include -isystem C:/src/shadPS4/externals/LibAtrac9/C/src -isystem C:/src/shadPS4/externals/sirit/src/../include -isystem C:/src/shadPS4/externals/sirit/externals/SPIRV-Headers/include -isystem C:/src/shadPS4/externals/zydis/include -isystem E:/build-shadps4-gcc/externals/zydis -isystem C:/src/shadPS4/externals/zydis/dependencies/zycore/include -isystem E:/build-shadps4-gcc/externals/zydis/zycore -isystem C:/src/shadPS4/externals/glslang/glslang/.. -isystem E:/build-shadps4-gcc/include -isystem E:/build-shadps4-gcc/externals/glslang -isystem C:/src/shadPS4/externals/stb -isystem C:/src/shadPS4/externals/hwinfo/include -isystem C:/src/shadPS4/externals/miniz -isystem E:/build-shadps4-gcc/externals/miniz -isystem C:/src/shadPS4/externals/aacdec/fdk-aac/libAACdec/include -isystem C:/src/shadPS4/externals/aacdec/fdk-aac/libFDK/include -isystem C:/src/shadPS4/externals/aacdec/fdk-aac/libSYS/include -isystem C:/src/shadPS4/externals/aacdec/fdk-aac/libArithCoding/include -isystem C:/src/shadPS4/externals/aacdec/fdk-aac/libMpegTPDec/include -isystem C:/src/shadPS4/externals/aacdec/fdk-aac/libSBRdec/include -isystem C:/src/shadPS4/externals/aacdec/fdk-aac/libPCMutils/include -isystem C:/src/shadPS4/externals/aacdec/fdk-aac/libDRCdec/include -isystem C:/src/shadPS4/externals/aacdec/fdk-aac/libSACdec/include -isystem C:/src/shadPS4/externals/CLI11/include -isystem C:/msys64/ucrt64/include/AL -isystem C:/src/shadPS4/externals/cpp-httplib -isystem C:/src/shadPS4/externals/libusb/libusb -isystem C:/src/shadPS4/externals/ext-wepoll -isystem C:/src/shadPS4/externals/discord-rpc/include -O2 -g -DNDEBUG -std=gnu++23 -fdiagnostics-color=always -march=x86-64-v3 -MD -MT CMakeFiles/shadps4.dir/src/shader_recompiler/recompiler.cpp.obj -MF CMakeFiles\shadps4.dir\src\shader_recompiler\recompiler.cpp.obj.d -o CMakeFiles/shadps4.dir/src/shader_recompiler/recompiler.cpp.obj -c C:/src/shadPS4/src/shader_recompiler/recompiler.cpp
In file included from C:/src/shadPS4/src/shader_recompiler/frontend/instruction.h:7,
                 from C:/src/shadPS4/src/shader_recompiler/frontend/control_flow_graph.h:15,
                 from C:/src/shadPS4/src/shader_recompiler/recompiler.cpp:4:
C:/src/shadPS4/src/shader_recompiler/frontend/opcodes.h:2701:8: error: 'magic_enum' has not been declared
 2701 | struct magic_enum::customize::enum_range<Shader::Gcn::Opcode> {
      |        ^~~~~~~~~~
C:/src/shadPS4/src/shader_recompiler/frontend/opcodes.h:2701:41: error: expected unqualified-id before '<' token
 2701 | struct magic_enum::customize::enum_range<Shader::Gcn::Opcode> {
      |                                         ^
```

squidbus: Couldn't the struct specialization just be in the cpp file then, same as the `#define`s were?

oltolm: If I do that I get the warnings
```
[46/46] Linking CXX executable shadps4.exe
CMakeFiles/shadps4.dir/src/shader_recompiler/frontend/decode.cpp.obj: duplicate section `.rdata$_ZN10magic_enum6detail7names_vIN6Shader3Gcn6OpcodeELNS0_12enum_subtypeE0EEE[_ZN10magic_enum6detail7names_vIN6Shader3Gcn6OpcodeELNS0_12enum_subtypeE0EEE]' has different size
CMakeFiles/shadps4.dir/src/shader_recompiler/frontend/decode.cpp.obj: duplicate section `.rdata$_ZN10magic_enum6detail8values_vIN6Shader3Gcn6OpcodeELNS0_12enum_subtypeE0EEE[_ZN10magic_enum6detail8values_vIN6Shader3Gcn6OpcodeELNS0_12enum_subtypeE0EEE]' has different size
build finished successfully.
```
These warnings were the reason for this commit.

squidbus: Ugh... I guess it's fine then

squidbus (approved):

src/shader_recompiler/frontend/opcodes.h
oltolm: I mean I could create an additional header `opcodes_magic_enum.h` and include that.

squidbus: It's fine, I assume anywhere both magic_enum and the opcodes enum exist it will trigger that linker warning, unless the specialization is also present. Best to just pair them in the same file if we have to, to avoid accidentally re-introducing it.

Niram7777:
thats an example of why I think we should only support Clang.
~~here the issue seems to be on GCC, I dont see why it tries to make a magic_enum<Shader::Gcn::Opcode> in decoder.cpp~~ (the definitions should be tthe same)
maybe you could retry but with adding in the first line of decoder.cpp ?
```cpp
#define MAGIC_ENUM_RANGE_MIN 0
#define MAGIC_ENUM_RANGE_MAX 2143
#include <magic_enum/magic_enum.h>
```

oltolm:
It's because of code like this in `decode.cpp`
```c++
    ASSERT_MSG(instFormat.src_type != ScalarType::Undefined &&
                   instFormat.dst_type != ScalarType::Undefined,
               "Instruction format table incomplete for opcode {} ({}, encoding = 0x{:x})",
               magic_enum::enum_name(m_instruction.opcode), u32(m_instruction.opcode),
               u32(encoding));
```

squidbus:
> thats an example of why I think we should only support Clang. here the issue seems to be on GCC, I dont see why it tries to make a magic_enumShader::Gcn::Opcode in decoder.cpp

It's implicit to using `magic_enum` with the type, the issue is that you get conflicting definitions between using `magic_enum` with the raised limit and other files without the limit raise. Clang just doesn't complain about it.

Niram7777:
then the real fix is to add
> ```c++
> #define MAGIC_ENUM_RANGE_MIN 0
> #define MAGIC_ENUM_RANGE_MAX 2143
> ```

 before the ` #include <magic_enum/magic_enum.h>` in decode.cpp
(or define them with CMake if want to go further: `target_compile_definitions(magic_enum PUBLIC MAGIC_ENUM_RANGE_MIN=0 MAGIC_ENUM_RANGE_MAX=2143)`)

oltolm:
I already offered basically the same solution when I suggested a separate `opcodes_magic_enum.h`. There is nothing more to do for me here.

Niram7777:
> > thats an example of why I think we should only support Clang. here the issue seems to be on GCC, I dont see why it tries to make a magic_enumShader::Gcn::Opcode in decoder.cpp
> 
> It's implicit to using `magic_enum` with the type, the issue is that you get conflicting definitions between using `magic_enum` with the raised limit and other files without the limit raise. Clang just doesn't complain about it.

clang takes the first one it encounter and ignore the other definitions I believe,  I am not sure we want to rely on that
(gcc warning is legitimate)

squidbus:
I think I'm fine with the solution as it exists currently for now, I took a look and it seems we're doing these `magic_enum` struct specializers in other headers already anyway.

PR #4371, opened 2026-05-07
fix: use fallback np_comm_id when trophyIndexMap lookup fails

TdlifePeng:
## Problem

GOW III Remastered (CUSA01715) shows black screen with no audio after the logo. The render loop and draw calls run normally (~14K draws/60 frames), but the final output is all black.

## Bisect

Bisected to 5b60b73e (added new trophies and saves dirs #4177). That commit initially crashed (out-of-bounds array access), which masked the black screen. The crash was later fixed in 1c8ace66 (fixed trophy extraction #4254), at which point the black screen became visible.

## Root cause

In sceNpTrophyCreateContext, when service_label is not found in trophyIndexMap (e.g. no npbind.dat or trophy data not extracted), the function returns ORBIS_NP_TROPHY_ERROR_UNKNOWN. The game calls CreateContext -> RegisterContext -> GetGameInfo during boot, and the failed context creation causes it to enter an error state and stop rendering.

## Fix

When npCommId is not found in trophyIndexMap, fall back to a generated id (np_comm_<index>) instead of returning an error, so the trophy context can be created normally.

## Test

CUSA01715 proceeds past the logo into the game with correct video and audio.

Note: this PR is for reference only.

[Claude Code . deepseek-v4-pro]

### DIFF
--- a/src/core/libraries/np/np_trophy.cpp
+++ b/src/core/libraries/np/np_trophy.cpp
@@ -225,9 +225,12 @@ s32 PS4_SYSV_ABI sceNpTrophyCreateContext(OrbisNpTrophyContext* context,
     if (it != trophyMap.end()) {
         np_comm_id = it->second;
     } else {
-        LOG_ERROR(Lib_NpTrophy, "No npCommId found for trophy index/service_label: {}",
-                  service_label);
-        return ORBIS_NP_TROPHY_ERROR_UNKNOWN;
+        // Trophy data may not be available (no npbind.dat, extraction failed, etc.).
+        // Fall back to a generated id rather than returning an error, which can
+        // cause the game to stop rendering (e.g. black screen after logo).
+        np_comm_id = std::string("np_comm_") + std::to_string(service_label);
+        LOG_WARNING(Lib_NpTrophy, "No npCommId found for trophy index/service_label: {}, using {}",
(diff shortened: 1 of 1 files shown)


StevenMiller123:
This seems more like a bad game dump to me. Could you provide an emulator log?

StevenMiller123:
Others I've reached out to also seem to agree this is issue you've fixed is caused by a bad game dump, your copy of the game is likely missing some files.

TdlifePeng:
> This seems more like a bad game dump to me. Could you provide an emulator log?

[shad_log_black_screen.txt](https://github.com/user-attachments/files/27487433/shad_log_black_screen.txt)

Niram7777:
I dont have this issue with version 1.0 of this game, I doubt version 1.1 is removing the trophy file.
You can try to dump again, eventually without the update.
(I do have a black screen issue but its before the logo, and it looks unrelated to your change request, I'm investigating)

Niram7777:
@TdlifePeng it has been fixed, you can close this PR (or report the issue it happens again)

StevenMiller123:
@TdlifePeng's issue started before I even broke trophy code, and probably remains an issue until they redump the game.
Either way, this PR is effectively just a hack, so I'll close it.

PR #4373, opened 2026-05-07
Refactor memory size handling to better handle vdec titles

Hog185:


StevenMiller123:
No regressions in my tests, fixes The Last of Us™ Remastered crashing on cutscenes.
image
image
[shad_log.txt](https://github.com/user-attachments/files/27494893/shad_log.txt)

gandalfthewhite19890404:
image
now we can see FMVs without hacked build

src/core/libraries/videodec/videodec.cpp
```diff
@@ -9,7 +9,50 @@
 
 namespace Libraries::Videodec {
 
-static constexpr u64 kMinimumMemorySize = 16_MB; ///> Fake minimum memory size for querying
+static constexpr u64 kFallbackMemorySize = 16_MB;
+
+static constexpr u32 AlignUp(u32 value, u32 alignment) {
```
kalaposfos13: Just a nitpick, but https://github.com/shadps4-emu/shadPS4/blob/d85a5013ecf279f7b78c859b82d84fa486f8ee15/src/common/alignment.h#L12

StevenMiller123 (approved):
Pretty sure this isn't fully accurate, but it's significantly better than what we had before.

PR #4376, opened 2026-05-08
fix: Enhance image overlap resolution logic for ufc4

Hog185:
untested due to not having ufc4 locally, but swaps freeimage to expand image

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -575,16 +575,35 @@ ImageId TextureCache::FindImage(ImageDesc& desc, bool exact_fmt) {
         }
     }
 
-    if (image_id) {
+if (image_id) {
         Image& image_resolved = slot_images[image_id];
         if (exact_fmt && info.pixel_format != image_resolved.info.pixel_format) {
-            // Cannot reuse this image as we need the exact requested format.
+            // Cannot reuse this image
             image_id = {};
         } else if (image_resolved.info.resources < info.resources) {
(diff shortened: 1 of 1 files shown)


src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -575,16 +575,35 @@ ImageId TextureCache::FindImage(ImageDesc& desc, bool exact_fmt) {
...
+                         "Expanding image {} at {:#x}:\n"
+                         "  Mips:   {} -> {}\n"
+                         "  Layers: {} -> {}\n"
+                         "  Format: {} -> {}",
+                         image_id.index, info.guest_address,
+                         image_resolved.info.resources.levels, info.resources.levels,
+                         image_resolved.info.resources.layers, info.resources.layers,
+                         vk::to_string(image_resolved.info.pixel_format), vk::to_string(info.pixel_format));
+
+                image_id = ExpandImage(info, image_id);
```
squidbus, line 596: Not quite sure why we would be expanding here, expansion is supposed to be handled as part of `ResolveOverlap`. If there's some case where it should be but isn't, I would think that should be fixed there.

Hog185: resolveoverlap isnt catching it
ill see, but this is just to see if this fixes ufc4

squidbus: Any valid cases should be caught by `ResolveOverlap`, hence why this path was just a failure case before (not to say it's necessarily perfect, just that if there is a problem it should be addressed there).

PR #4378, opened 2026-05-08
Core: Fix argv[0] value

StevenMiller123:
Small fix for a small bug I noticed while debugging Uncharted hangs.


chronicer:
Yeah, more Uncharted)

PR #4379, opened 2026-05-08
ImGui:: Notifications

rainmakerv3:
Simple notifications window. Made this mostly for the upcoming shadNet (friend request / friend online notifications and such), but can be used for whatever else. For this PR, it's only use right now is for notifying about screenshot captures (previously there's no feedback to the user that screenshots were succesfully taken), as shown here.

image


kalaposfos13:
Oh nice, I can finally make a low-effort sceSysUtilSendSystemNotificationWithText implementation without having to learn ImGui for it.

PR #4380, opened 2026-05-09
shader_recompiler: Strip out manual bounds checking

squidbus:
Intent was to provide AMD bounds checking behavior on drivers without `robustBufferAccess2` (e.g. MoltenVK), but there were some unresolved issues with control flow in the result and it was disabled a long while back.

Just going to strip out the code for this to keep things simple. It doesn't seem to end up being that important overall, and long term, when KosmicKrisp improves enough to switch to that (mainly performance), it should have the support we need anyway.


mikusp (approved):

raphaelthegreat:
I think the vertex attribute implementation with buffers fallback path should be kept around as its useful for testing. Also can shared memory bounds checking be removed, I dont believe a definitive answer was brought regarding its usefulness

squidbus:
> I think the vertex attribute implementation with buffers fallback path should be kept around as its useful for testing. Also can shared memory bounds checking be removed, I dont believe a definitive answer was brought regarding its usefulness

Restored it behind a disabled `ifdef`. As for the shared memory bound checking, everything here was force-disabled anyway so none of it was in effect. I'd like to leave it to just removing that stuff for now.

squidbus:
https://github.com/shadps4-emu/shadPS4/pull/4385 for shared memory

PR #4382, opened 2026-05-09
Ufc2 fix

Hog185:
fix problems with ufc, now can get in game but rendering issues casue the game to be unplayable


Hog185:
this ended up fruitless

PR #4385, opened 2026-05-09
shader_recompiler: Remove shared memory bounds checking.

squidbus:
Remainder of removing manual bounds checking, this was actually in use so probably needs some regression testing.


Graubek:
No regressions on my end

raphaelthegreat (approved):

PR #4386, opened 2026-05-09
Refactor pad handles

kalaposfos13:
Following what the original library does, handles are now stored in a `[userid, type, index] -> pad_handle` mapping instead of a simple `userid -> pad_handle` map. This fixes a regression in A KINGS TALE: FINAL FANTASY XV.
Someone should probably go test what random edge case(s) I broke with this PR, I have only tested the bare minimum to ensure it at least works.


roamic (approved):

StevenMiller123:
Seems to work properly in what I've tested.

PR #4389, opened 2026-05-10
Fixes regressions and issues causing PSP emulated games to not run.

sammiq:
I hope you'll look these changes over and consider them, as I believe they fix core issues and are not hacks to make things work.

1. memory.cpp - Reserved Memory cannot be mapped, this seems to be incorrect and the PSP emulation relies on reserving then mapping memory at startup. This seems to be a core error in the emulation code, as reserved memory cannot be mapped.

2. appcontent.cpp - Temp directory output parameter may have garbage and the API is not null terminating the output, resulting in failures when the file path buffer is made into an invalid path name. This seems like a random issue, but managed to get this occurring on some games consistently.

3. videodec2_impl.cpp - Fix misaligned images when viewport is sized for PSP. This fixes garbage in movies on the PSP emulator, making the movies viewable, scaled correctly for the screen.  A lot of games work fine as linesize[0] == width but PSP moves are definitely not that case.

5.  videodec.cpp - Also implemented sceVideodec2GetAvcPictureInfo, which seems to be used over sceVideodec2GetPictureInfo by the PSP emulator.

6. texture_cache.cpp - Whie working on texutre transparency issues, i noticed that block texture size was not quite right while hashing, I think the revised code is more correct.

7. RGB4444 mapping was a strange swizzle, causing strange colors and lack of alpha transparency. I believe the swizzling order was incorrect and changing this makes textures no longer look strange.

 These changes combined allow Jean d'Arc, LocoRoco and Patapon (and possibly other PSP and enhanced PSP games) to run decently and with no visual corruption.

### DIFF
--- a/src/core/libraries/app_content/app_content.cpp
+++ b/src/core/libraries/app_content/app_content.cpp
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include <cmath>
-#include <cstring>
 
 #include "app_content.h"
 #include "common/assert.h"
@@ -401,9 +400,9 @@ int PS4_SYSV_ABI sceAppContentTemporaryDataMount2(OrbisAppContentTemporaryDataOp
     if (mountPoint == nullptr) {
         return ORBIS_APP_CONTENT_ERROR_PARAMETER;
     }
-    memset(mountPoint, 0, sizeof(OrbisAppContentMountPoint));
     static constexpr std::string_view TmpMount = "/temp0";
     TmpMount.copy(mountPoint->data, TmpMount.size());
+    mountPoint->data[TmpMount.size()] = '\0';
     LOG_INFO(Lib_AppContent, "sceAppContentTemporaryDataMount2: option = {}, mountPoint = {}",
              option, mountPoint->data);
     return ORBIS_OK;
--- a/src/core/libraries/app_content/app_content.cpp
+++ b/src/core/libraries/app_content/app_content.cpp
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 #include <cmath>
+#include <cstring>
 
 #include "app_content.h"
 #include "common/assert.h"
@@ -400,6 +401,7 @@ int PS4_SYSV_ABI sceAppContentTemporaryDataMount2(OrbisAppContentTemporaryDataOp
     if (mountPoint == nullptr) {
         return ORBIS_APP_CONTENT_ERROR_PARAMETER;
     }
+    memset(mountPoint, 0, sizeof(OrbisAppContentMountPoint));
     static constexpr std::string_view TmpMount = "/temp0";
     TmpMount.copy(mountPoint->data, TmpMount.size());
     LOG_INFO(Lib_AppContent, "sceAppContentTemporaryDataMount2: option = {}, mountPoint = {}",
--- a/src/core/libraries/videodec/videodec2_impl.cpp
+++ b/src/core/libraries/videodec/videodec2_impl.cpp
@@ -133,7 +133,7 @@ s32 VdecDecoder::Decode(const OrbisVideodec2InputData& inputData,
 
         // Only set framePitchInBytes if the game uses the newer struct version.
         if (outputInfo.thisSize == sizeof(OrbisVideodec2OutputInfo)) {
-            outputInfo.framePitchInBytes = frame->linesize[0];
+            outputInfo.framePitchInBytes = frame->width;
         }
 
         if (outputInfo.isValid) {
@@ -201,7 +201,7 @@ s32 VdecDecoder::Flush(OrbisVideodec2FrameBuffer& frameBuffer,
         outputInfo.codecType = 1; // FIXME: Hardcoded to AVC
         outputInfo.frameWidth = frame->width;
         outputInfo.frameHeight = frame->height;
-        outputInfo.framePitch = frame->linesize[0];
+        outputInfo.framePitch = frame->width;
         outputInfo.frameBufferSize = frameBuffer.frameBufferSize;
         outputInfo.frameBuffer = frameBuffer.frameBuffer;
 
@@ -211,7 +211,7 @@ s32 VdecDecoder::Flush(OrbisVideodec2FrameBuffer& frameBuffer,
 
         // Only set framePitchInBytes if the game uses the newer struct version.
         if (outputInfo.thisSize == sizeof(OrbisVideodec2OutputInfo)) {
-            outputInfo.framePitchInBytes = frame->linesize[0];
+            outputInfo.framePitchInBytes = frame->width;
         }
 
         // FIXME: Should we add picture info here too?
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -550,7 +550,7 @@ s32 MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, u64 size, Memo
                    virtual_addr);
         auto vma = FindVMA(virtual_addr)->second;
         auto remaining_size = vma.base + vma.size - virtual_addr;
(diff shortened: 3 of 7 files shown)


src/core/libraries/videodec/videodec2_impl.cpp
```diff
@@ -201,7 +201,7 @@ s32 VdecDecoder::Flush(OrbisVideodec2FrameBuffer& frameBuffer,
         outputInfo.codecType = 1; // FIXME: Hardcoded to AVC
         outputInfo.frameWidth = frame->width;
         outputInfo.frameHeight = frame->height;
-        outputInfo.framePitch = frame->linesize[0];
+        outputInfo.framePitch = frame->width;
```
sammiq, line 204: Note that this copy of the code differed from Decode, which I believe is an oversight from a previous change.

src/core/libraries/app_content/app_content.cpp
```diff
@@ -400,6 +401,7 @@ int PS4_SYSV_ABI sceAppContentTemporaryDataMount2(OrbisAppContentTemporaryDataOp
     if (mountPoint == nullptr) {
         return ORBIS_APP_CONTENT_ERROR_PARAMETER;
     }
+    memset(mountPoint, 0, sizeof(OrbisAppContentMountPoint));
     static constexpr std::string_view TmpMount = "/temp0";
     TmpMount.copy(mountPoint->data, TmpMount.size());
```
squidbus: Wouldn't it be more efficient to just add a null terminator after the end of the copy, rather than `memset` over the whole thing?

src/video_core/amdgpu/pixel_format.h
```diff
@@ -264,10 +264,10 @@ constexpr CompMapping RemapSwizzle(const DataFormat format, const CompMapping sw
...
         // Remap to a more supported component order.
         CompMapping result;
-        result.r = swizzle.g;
-        result.g = swizzle.b;
-        result.b = swizzle.a;
-        result.a = swizzle.r;
+        result.r = swizzle.a;
+        result.g = swizzle.r;
+        result.b = swizzle.g;
+        result.a = swizzle.b;
```
squidbus: Are you sure this is correct, I remember debugging this with someone a while back to fix issues in another game. The swizzle was a bit odd IIRC because we use `eB4G4R4A4UnormPack16` ("4-bit B component in bits 12..15, a 4-bit G component in bits 8..11, a 4-bit R component in bits 4..7, and a 4-bit A component in bits 0..3") since it's more widely supported by drivers.

squidbus:
Tagging @StevenMiller123 for the memory stuff.

Parotaku:
Patapon, japanese version still crash here...
[shad_log.txt](https://github.com/user-attachments/files/27564201/shad_log.txt)

src/video_core/amdgpu/pixel_format.h
sammiq: Thanks for looking!

CUSA41018_before shows the texture swizzling before, and after this change:
CUSA41018_after

Can you remember which game? I certainly don't want to cause regressions, but I tested almost every variation of swizzling and they were all incorrectly mapping colors or alpha.

sammiq:
> Patapon, japanese version still crash here... [shad_log.txt](https://github.com/user-attachments/files/27564201/shad_log.txt)

I'm not sure about the japanese version, but the US version (CUSA06171) works now.CUSA06171_20260510_174817_040_game_000000

src/core/libraries/app_content/app_content.cpp
sammiq: I't might be slightly more efficient, but I wasn't sure i wanted to stuff about getting string lengths and associated logic when a small memset on 32 bytes is pretty darn quick

src/video_core/amdgpu/pixel_format.h
squidbus: I think it was some Dragon Ball game... that does look right in your images though, I guess we'll see if there's a problem.

src/core/libraries/app_content/app_content.cpp
squidbus: You're right that it's pretty much a micro-optimization, but it should just be:
```
mountPoint->data[TmpMount.size()] = '\0';
```

Since the string length is just defined by the size of constant `TmpMount`.

sammiq: Ok I’ll push a change for it, thanks!

src/core/memory.cpp
```diff
@@ -550,7 +550,7 @@ s32 MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, u64 size, Memo
                    virtual_addr);
         auto vma = FindVMA(virtual_addr)->second;
         auto remaining_size = vma.base + vma.size - virtual_addr;
-        if (!vma.IsFree() || remaining_size < size) {
+        if (vma.IsMapped() || remaining_size < size) {
```
StevenMiller123: This doesn't look right. When NoOverwrite is specified, any area that isn't free should trigger this case.

src/core/memory.cpp
StevenMiller123: The actual bug is this. I must've missed this at some point, but swap the False to a True here and it should work.

src/core/memory.cpp
```diff
@@ -726,7 +726,7 @@ s32 MemoryManager::MapFile(void** out_addr, VAddr virtual_addr, u64 size, Memory
         auto vma = FindVMA(virtual_addr)->second;
 
         auto remaining_size = vma.base + vma.size - virtual_addr;
-        if (!vma.IsFree() || remaining_size < size) {
```
StevenMiller123, line 729: Same as my first comment, this is incorrect.

StevenMiller123 (changes requested):
Thanks for prompting me to actually look at this, seems the memory bug was simple all along 😅

StevenMiller123:
Breaks J-STARS Victory Vs+
image

[CUSA01593.log](https://github.com/user-attachments/files/27570985/CUSA01593.log)

squidbus:
Yeah that would be the 4444 format ordering.

src/core/memory.cpp
sammiq: Ah gotcha, I'll amend this, thanks for that I'm not sure I got the subtlety there, thanks for looking.

squidbus (commented):
Summary of where I am on this:
* Most of the changes here look good to me.
* Memory changes need to be updated based on feedback.
* 4444 format changes I'm still not sure on, especially given reported regression.
* I don't know much about the `videodec2_impl.cpp` changes, might need some testing with games using videodec2.

I'm okay if you fix the memory stuff and want to split off the 4444 part into a separate PR and land the rest of the fixes sooner, if there's some contention there.

sammiq:
> I'm okay if you fix the memory stuff and want to split off the 4444 part into a separate PR and land the rest of the fixes sooner, if there's some contention there.

I'll revert the 4444 fixes, and do some more testing, it definitely fixes some PSP games, but that's no good if it breaks others. I'll find some other games and see whether there's a divergent code path for texture handling somewhere.

Thanks to all for your patience and feedback.

squidbus:
@StevenMiller123 Can you re-test since the format change is removed?

StevenMiller123 (approved):
The parts I'm familiar with look good, I'll test J-STARS again as soon as I can.

StevenMiller123:
J-STARS is back to rendering properly.
image

sammiq:
Thanks again for test, I confirmed that several PSP games are back to having odd texture issues, so I'll do more digging. @squidbus are you happy for me to merge this?

squidbus (approved):

Linear524:
Initial idea is good - but I still getting "Device lost" crash during the boot of LocoRoco Remastered (CUSA06090)
Logs attached
[CUSA06090.txt](https://github.com/user-attachments/files/27594579/CUSA06090.txt)

StevenMiller123:
The GPU driver crashing is a separate issue.

PR #4392, opened 2026-05-11
Fixes to memory mapping on Windows

sammiq:
The previous PR (#4389) works on Linux fine, but I just got my build environment set up on Windows and it doesn't like mapping files with PAGE_NOACCESS. This adds a case to the mapping to map as read/write, and then protect the page after mapping, following the same logic as the allocation above. This allows the mapping to work, and the PSP emulation to continue.

@StevenMiller123 you should look at this in case I've done something silly,

Failed boot: [CUSA41018.log](https://github.com/user-attachments/files/27579520/CUSA41018.log)
Successful boot: [CUSA41018.log](https://github.com/user-attachments/files/27579490/CUSA41018.log)

### DIFF
--- a/src/core/address_space.cpp
+++ b/src/core/address_space.cpp
@@ -257,9 +257,20 @@ struct AddressSpace::Impl {
                 ret = VirtualProtect(ptr, size, prot, &resultvar);
                 ASSERT_MSG(ret, "VirtualProtect failed. {}", Common::GetLastErrorMsg());
             } else {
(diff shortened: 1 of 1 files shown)


StevenMiller123 (approved):
In the future I'll probably look closer at why MapViewOfFile3 fails here, as technically PAGE_NOACCESS shouldn't cause an error on it's own.

Since this is just a minor performance hit (and for an edge case we rarely hit), I'll approve as-is.

PR #4394, opened 2026-05-11
Add and handle hull shaders

Hog185:
Please Give me all the feedback you can. i am not entirely sure what i’m doing and any feedback will help


squidbus:
I'm not sure what this is supposed to be doing, you've added a file with some basic enum conversions and haven't used it anywhere.

Hog185:
turns out i’m stupid and didn’t see that this is already handled, :(

PR #4402, opened 2026-05-13
core: simplify sysv_abi wrappers for MinGW-w64 GCC

oltolm:
Use template <auto f> for HostCallWrapperImpl and OrbisWrapperImpl instead of passing the function type and value separately.

This keeps the wrapper behavior the same while matching the form MinGW-w64 GCC accepts.

---
This PR makes GCC on Windows able to build shadps4. You need [my patch](https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=b0f09b72d9e1783e8ccaf2d8c1c72b03e3973f0b), that is already upstream and that I plan to backport to MSYS2 GCC 16.1.

The problem with the templates is AFAIU the fact that GCC is more strict than Clang and takes the ABI attribute into account when matching templates.

### DIFF
--- a/src/core/libraries/kernel/threads.h
+++ b/src/core/libraries/kernel/threads.h
@@ -6,6 +6,7 @@
 #include <functional>
 #include "common/polyfill_thread.h"
 #include "core/libraries/kernel/threads/pthread.h"
+#include "core/tls.h"
 
 namespace Core::Loader {
 class SymbolsResolver;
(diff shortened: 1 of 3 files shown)


src/core/libraries/kernel/threads.h
```diff
@@ -52,7 +52,7 @@ class Thread {
         this->func = std::move(func);
         PthreadAttrT attr{};
         posix_pthread_attr_init(&attr);
-        posix_pthread_create(&thread, &attr, HOST_CALL(RunWrapper), this);
+        posix_pthread_create(&thread, &attr, RunWrapper, this);
```
squidbus: Seems like something that should not be removed, `RunWrapper` is a host function.

oltolm: fixed

squidbus (approved):

Issue #4390, opened 2026-05-10
[GAME BUG]: No input in resident evil 1 HD remake


Forgottenshadow89:
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.
- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [x] I have disabled all patches and cheats and the issue is still present.
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadPS4/wiki/I.-Quick-start-%5BUsers%5D#4-dumping-firmware-modules) installed.

### Describe the Bug

Strange issue in resident evil origins collection. The resident evil 0 works flawlessly but the resident evil 1 don't receive any kind of input. This is happening from the very beginning of the emulator (Not a recent issue) but I wonder why this is the only game that has no inputs at all and even more strange, why the resident evil 0 has input and the 1 doesn't?

### Reproduction Steps

Start resident evil origins collection. Select resident evil 1. You cannot do nothing, any button pressed does nothing. It works in resident evil 0.

### Specify OS Version

W11 pro x64

### CPU

13600KF

### GPU

RTX 4070

### Amount of RAM in GB

32GB DDR5

### Amount of VRAM in GB

12

### Log File

[shad_log.txt](https://github.com/user-attachments/files/27570518/shad_log.txt)

w1naenator:
There is another title with same issue - "Pump it BMX+". Both titles calls stubbed Move init API. As soon API returns 0, I think game's engine just switching to Move input.

Forgottenshadow89:
> There is another title with same issue - "Pump it BMX+". Both titles calls stubbed Move init API. As soon API returns 0, I think game's engine just switching to Move input.


Strange... what's the point of doing that? The resident evil 1 not compatible with psmove, and why only with re1 and not re0? And even it if were compatible I don't see a reason to switch to psmove again and again.

w1naenator:
> > There is another title with same issue - "Pump it BMX+". Both titles calls stubbed Move init API. As soon API returns 0, I think game's engine just switching to Move input.
> 
> Strange... what's the point of doing that? The resident evil 1 not compatible with psmove, and why only with re1 and not re0? And even it if were compatible I don't see a reason to switch to psmove again and again.

[...]

PR #4410, opened 2026-05-14
np-legacy-state-callback

w1naenator:
Core: Add user state callback management and improve user login/logout handling


w1naenator:
This fixes the `Pump it BMX+` softlock, and very possibly this issue as well: https://github.com/shadps4-emu/shadPS4/issues/4390

Huge regression testing is needed in other games.

w1naenator:
fix needed for RE1HD.

src/core/user_manager.cpp
```diff
@@ -195,11 +195,18 @@ LoggedInUsers UserManager::GetLoggedInUsers() const {
...
 
 void UserManager::LoginUser(User* u, s32 player_index) {
-    if (!u) {
+    if (!u || player_index < 1 || player_index > static_cast<s32>(logged_in_users.size())) {
         return;
     }
+
+    for (auto& logged_in_user : logged_in_users) {
+        if (logged_in_user == u) {
+            logged_in_user = nullptr;
```
kalaposfos13, line 204: For this very unlikely edge case, I believe a sign-out event should be queued for the old index, but I'm not sure if this can even happen, I'd put an unreachable here myself.

georgemoralis (approved):

georgemoralis:
i don't need other excuse that RE working to accept this

StevenMiller123:
Sounds like this (particularly the libScePad part) regressed a whole load of Unity games, to where they aren't accepting inputs.
@brad0demx was doing tests and narrowed the issue down to here.
I haven't done much testing of my own, but I've confirmed Slime Rancher v1.00 is impacted.

w1naenator:
> Sounds like this (particularly the libScePad part) regressed a whole load of Unity games, to where they aren't accepting inputs. @brad0demx was doing tests and narrowed the issue down to here. I haven't done much testing of my own, but I've confirmed Slime Rancher v1.00 is impacted.

I will check my Unity games tomorrow. And if I find any regression, then I will fix the issue.

w1naenator:
@StevenMiller123 could you explain in more detail what exactly this PR has broken?

StevenMiller123:
@kalaposfos13 already fixed it.

StevenMiller123:
The issue was with the new pad stuff you did for RE1HD. You were incorrectly setting device class to invalid for disconnected pads, and that broke a check Unity has (which in turn broke inputs in a bunch of Unity titles).

w1naenator:
> The issue was with the new pad stuff you did for RE1HD. You were incorrectly setting device class to invalid for disconnected pads, and that broke a check Unity has (which in turn broke inputs in a bunch of Unity titles).

Thank you for answer!

Issue #4387, opened 2026-05-09
[Aggregate issue] Issues with some Intel CPUs on Windows devices


StevenMiller123:
## This issue is for tracking bugs specific Intel CPUs running Windows. This is primarily known to affect 12th gen Intel CPUs and up, but some older CPUs are also affected.
Don't expect a fix anytime soon, this is just to keep track of the issue since we're finding a growing number of titles affected by this sort of problem.

- Ratchet & Clank™ has vertex explosions ingame (see https://github.com/shadps4-emu/shadPS4/issues/3658 for details)

Parotaku:
Don't know if this will help but here's a log with TLOU crashing on Win10 with Intel CPU:

[shad_log.txt](https://github.com/user-attachments/files/27559312/shad_log.txt)

StevenMiller123:
Looks like you are hitting the issue, thanks for the log!

bigol83:
Bloodborne crash on 12400f cpu windows 11 while attacking an enemy

[CUSA03173.log](https://github.com/user-attachments/files/27560722/CUSA03173.log)

Uncharted 2 crashing on main menu

[CUSA02343.log](https://github.com/user-attachments/files/27560743/CUSA02343.log)

Uncharted 3 crashing even before reaching main menu

[CUSA02343.log](https://github.com/user-attachments/files/27560748/CUSA02343.log)

Persona 5 broken intro movie

[CUSA06638.log](https://github.com/user-attachments/files/27560760/CUSA06638.log)

The Last of Us remastered crash on boot

[CUSA00556.log](https://github.com/user-attachments/files/27560773/CUSA00556.log)

gandalfthewhite19890404:
For me - it boots, i5-10400, Win10

<img width="1282" height="752" alt="Image" src="https://github.com/user-attachments/assets/96498a9c-a8a8-43dd-ace2-2330e455ad7a" />

[shad_log.txt](https://github.com/user-attachments/files/27562792/shad_log.txt)

Missake212:
Catherine: Full Body crashes when it's supposed to play the intro video.

[CUSA15036.log](https://github.com/user-attachments/files/27811149/CUSA15036.log)

Graubek:
InFamous First Light crashing silently after displaying splashscreeen (The only way to see why the game crashes is to run it in debug mode)

[...]

PR #4415, opened 2026-05-14
Flush log on terminate

Niram7777:
Should fix log truncated but needs Windows + Intel test
https://github.com/shadps4-emu/shadPS4/issues/4387#issuecomment-4451200988

I dont know what will be logged between "Unknown exception caught" and "Exception: {}" because I dont know if MS exception extends std::exception

(If we wanted to do something more fancy: needs to wrap all our threads)
https://learn.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=msvc-170

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -27,6 +27,23 @@
 #endif
 #include <core/user_settings.h>
 
+#ifdef _WIN32
+LONG TopLevelExceptionHandler(_EXCEPTION_POINTERS* ExceptionInfo) {
+    DWORD code = 0;
+    PVOID address = nullptr;
+
+    if (ExceptionInfo != nullptr && ExceptionInfo->ExceptionRecord != nullptr) {
+        code = ExceptionInfo->ExceptionRecord->ExceptionCode;
+        address = ExceptionInfo->ExceptionRecord->ExceptionAddress;
+    }
+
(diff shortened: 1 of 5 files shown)


Niram7777:
Microsoft actually dont throw SEH if there is no try catch around, amazing :-1: 
so we would have to wrap our threads with `__try/__catch`.... (hopefully I have enough time)

Niram7777:
Windows still refuses to throw...

src/main.cpp
```diff
@@ -27,6 +27,23 @@
...
+LONG TopLevelExceptionHandler(_EXCEPTION_POINTERS* ExceptionInfo) {
+    DWORD code = 0;
+    PVOID address = nullptr;
+
+    if (ExceptionInfo != nullptr && ExceptionInfo->ExceptionRecord != nullptr) {
+        code = ExceptionInfo->ExceptionRecord->ExceptionCode;
+        address = ExceptionInfo->ExceptionRecord->ExceptionAddress;
+    }
+
+    LOG_CRITICAL(Debug, "Unhandled Exception code {} at 0x{}", code, address);
```
kalaposfos13:
```suggestion
    LOG_CRITICAL(Debug, "Unhandled Exception code {} at {}", code, address);
```

Niram7777:
no matter what we ask Windows to do, it still doesnt want to...

PR #4429, opened 2026-05-15
Flush log on terminate

Niram7777:
https://github.com/shadps4-emu/shadPS4/pull/4415
(we cant reopen a forced push branch...)
@red-prig thanks for pointing me to the right direction

(normally EXCEPTION_CONTINUE_SEARCH of SignalHandler was not supposed to break my previous attemps but anyway)

### DIFF
--- a/src/core/signals.cpp
+++ b/src/core/signals.cpp
@@ -31,9 +31,16 @@ namespace Core {
 
 static LONG WINAPI SignalHandler(EXCEPTION_POINTERS* pExp) noexcept {
     const auto* signals = Signals::Instance();
+    DWORD code = 0;
+    PVOID address = nullptr;
+
+    if (pExp != nullptr && pExp->ExceptionRecord != nullptr) {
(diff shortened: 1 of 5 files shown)


src/core/signals.cpp
```diff
@@ -49,7 +56,14 @@ static LONG WINAPI SignalHandler(EXCEPTION_POINTERS* pExp) noexcept {
         break;
     }
 
-    return handled ? EXCEPTION_CONTINUE_EXECUTION : EXCEPTION_CONTINUE_SEARCH;
+    if (handled) {
+        return EXCEPTION_CONTINUE_EXECUTION;
+    }
+
+    LOG_CRITICAL(Debug, "Unhandled Exception code {} at {}", code, address);
```
kalaposfos13:
```suggestion
    LOG_CRITICAL(Debug, "Unhandled Exception code {:#x} at {}", code, address);
```

PR #4431, opened 2026-05-15
Http part1

georgemoralis:
Part1 of my http because no-one likes it because it is big . Contains more logging stuff , URI fixes and a possible way to report no internet connection available


StevenMiller123 (approved):
Traditionally, pretty sure sending requests is where errors should crop up. Accepting on the basis of getting the URI stuff fixed in main.

PR #4437, opened 2026-05-16
Reach control ingame fix

w1naenator:
Now Control completes the entire level loading process and reaches in-game, but with one nuance.


w1naenator:
Expected:
Screenshot_20260517-231933_YouTube_ReVanced_1

Actual:
CUSA11454_20260516_224630_744_game_000000

It looks like a post-processed effect that appears over the game content on a real PS4 and is almost invisible in the emulator absolutely has no transparency and is not clearing the previous state, but instead adds a new layer over the previous one each frame.
And still missing text.

PR #4441, opened 2026-05-17
Http part2

georgemoralis:
-Rewrote URI* functions more close to decompile
-Rewrote ParseStatusLine more close to decompile
-Implemented ParseResponseHeader
-Added minimal state machine
-SendRequest now returns proper error for no-internet connection


src/core/libraries/network/http.cpp
```diff
@@ -109,17 +160,52 @@ int PS4_SYSV_ABI sceHttpCookieImport(int libhttpCtxId, const void* buffer, u64 b
...
+        LOG_ERROR(Lib_Http, "Not initialized");
+        return ORBIS_HTTP_ERROR_BEFORE_INIT;
+    }
+    if (!g_state.templates.contains(tmplId)) {
+        LOG_ERROR(Lib_Http, "Invalid tmplId={}", tmplId);
+        return ORBIS_HTTP_ERROR_INVALID_ID;
+    }
+    if (!serverName) {
+        LOG_ERROR(Lib_Http, "serverName is null");
+        return ORBIS_HTTP_ERROR_INVALID_URL;
```
StevenMiller123: Should be `ORBIS_HTTP_ERROR_INVALID_VALUE`

src/core/libraries/network/http.cpp
```diff
@@ -128,38 +214,74 @@ int PS4_SYSV_ABI sceHttpCreateEpoll(int libhttpCtxId, OrbisHttpEpollHandle* eh)
...
 }
 
 int PS4_SYSV_ABI sceHttpCreateRequest2(int connId, const char* method, const char* path,
                                        u64 contentLength) {
-    LOG_ERROR(Lib_Http, "(STUBBED) called connId={}, method={}, path={}, contentLength={}", connId,
-              method ? method : "(null)", path ? path : "(null)", contentLength);
-    return ORBIS_OK;
+    LOG_INFO(Lib_Http, "called connId={}, method={}, path={}, contentLength={}", connId,
+             method ? method : "(null)", path ? path : "(null)", contentLength);
+    return sceHttpCreateRequestWithURL(connId, 0, path, contentLength);
```
StevenMiller123: null path should return `ORBIS_HTTP_ERROR_INVALID_VALUE`, this code results in returning `ORBIS_HTTP_ERROR_INVALID_URL`

src/core/libraries/network/http.cpp
```diff
@@ -128,38 +214,74 @@ int PS4_SYSV_ABI sceHttpCreateEpoll(int libhttpCtxId, OrbisHttpEpollHandle* eh)
 }
 
 int PS4_SYSV_ABI sceHttpCreateRequest(int connId, int method, const char* path, u64 contentLength) {
-    LOG_ERROR(Lib_Http, "(STUBBED) called connId={}, method={}, path={}, contentLength={}", connId,
-              method, path ? path : "(null)", contentLength);
-    return ORBIS_OK;
+    LOG_INFO(Lib_Http, "called connId={}, method={}, path={}, contentLength={}", connId, method,
+             path ? path : "(null)", contentLength);
+    return sceHttpCreateRequestWithURL(connId, method, path, contentLength);
```
StevenMiller123: null path should return `ORBIS_HTTP_ERROR_INVALID_VALUE`, this code results in returning `ORBIS_HTTP_ERROR_INVALID_URL`

src/core/libraries/network/http.cpp
StevenMiller123: Method should not be ignored, it needs to be converted to the proper int value.

StevenMiller123: Handling paths like urls seems to be incorrect.

src/core/libraries/network/http.cpp
StevenMiller123: Handling paths like urls seems to be incorrect.

src/core/libraries/network/http.cpp
```diff
@@ -128,38 +214,74 @@ int PS4_SYSV_ABI sceHttpCreateEpoll(int libhttpCtxId, OrbisHttpEpollHandle* eh)
...
 }
 
 int PS4_SYSV_ABI sceHttpCreateRequestWithURL2(int connId, const char* method, const char* url,
                                               u64 contentLength) {
-    LOG_ERROR(Lib_Http, "(STUBBED) called connId={}, method={}, url={}, contentLength={}", connId,
-              method ? method : "(null)", url ? url : "(null)", contentLength);
-    return ORBIS_OK;
+    LOG_INFO(Lib_Http, "called connId={}, method={}, url={}, contentLength={}", connId,
+             method ? method : "(null)", url ? url : "(null)", contentLength);
+    return sceHttpCreateRequestWithURL(connId, 0, url, contentLength);
```
StevenMiller123: Method should not be ignored, it needs to be converted to the proper int value.

StevenMiller123 (changes requested):
Focused on nitpicking things you're attempting to do which I've already decompiled.

src/core/libraries/network/http.cpp
```diff
@@ -216,14 +329,108 @@ int PS4_SYSV_ABI sceHttpCreateEpoll(int libhttpCtxId, OrbisHttpEpollHandle* eh)
...
 int PS4_SYSV_ABI sceHttpCreateRequest2(int connId, const char* method, const char* path,
                                        u64 contentLength) {
     LOG_INFO(Lib_Http, "called connId={}, method={}, path={}, contentLength={}", connId,
              method ? method : "(null)", path ? path : "(null)", contentLength);
-    return sceHttpCreateRequestWithURL(connId, 0, path, contentLength);
+    auto map_method = [](const char* m) -> int {
+        if (!m)
+            return -1;
+        std::string s = m;
+        if (s == "GET")
```
StevenMiller123, line 371: Real library uses a case insensitive check.

src/core/libraries/network/http.cpp
```diff
@@ -258,7 +465,39 @@ int PS4_SYSV_ABI sceHttpCreateRequestWithURL2(int connId, const char* method, co
                                               u64 contentLength) {
     LOG_INFO(Lib_Http, "called connId={}, method={}, url={}, contentLength={}", connId,
              method ? method : "(null)", url ? url : "(null)", contentLength);
-    return sceHttpCreateRequestWithURL(connId, 0, url, contentLength);
+    int int_method = -1;
+    if (method) {
+        std::string s = method;
+        if (s == "GET")
```
StevenMiller123, line 475: Check should be case insensitive.

StevenMiller123 (changes requested):

StevenMiller123:
Nevermind that review

StevenMiller123 (commented):
What I'm familiar with looks fine. I haven't decompiled much request logic though.

StevenMiller123 (approved):
Seems to function well enough.

PR #4443, opened 2026-05-17
http_part3

georgemoralis:


StevenMiller123 (approved):

StevenMiller123 (commented):
Missed this in my initial review, but this causes a deadlock. You're locking state in both sceHttpCreateRequest2 and sceHttpCreateRequestWithURL. Since your state mutex isn't recursive, that causes crashes.

PR #4449, opened 2026-05-18
Http part4

georgemoralis:
-Mostly Settings functions added , checked and verified with decompile


StevenMiller123 (approved):
Seems good to me

PR #4455, opened 2026-05-20
Fix assertion in UFC 2

Hog185:
also added hella logging


Hog185:
clang is not my friend

src/video_core/texture_cache/image.cpp
```diff
@@ -231,23 +231,41 @@ Image::Barriers Image::GetBarriers(vk::ImageLayout dst_layout, vk::AccessFlags2
...
+                                 info.resources.levels, info.resources.layers,
+                                 needs_partial_transition, partially_transited);
+                    if (subres_range) {
+                        LOG_CRITICAL(Render_Vulkan,
+                                     "GetBarriers subres_range: base.level={} base.layer={} "
+                                     "extent.levels={} extent.layers={}",
+                                     subres_range->base.level, subres_range->base.layer,
+                                     subres_range->extent.levels, subres_range->extent.layers);
+                    }
+                    ASSERT(false);
```
squidbus: Just use `ASSERT_MSG`, no need to log and then `ASSERT(false);`

src/video_core/texture_cache/image_view.cpp
```diff
@@ -60,7 +60,25 @@ ImageViewInfo::ImageViewInfo(const AmdGpu::Image& image, const Shader::ImageReso
...
+    } else {
+        range.base.layer = 0u;
+        if (image.base_array > 0) {
+            LOG_DEBUG(Render_Vulkan,
+                      "ImageViewInfo: non-array texture base_array={} >= NumLayers()={} "
+                      "(type={} depth={}) using local layer 0",
+                      image.base_array, image.NumLayers(), magic_enum::enum_name(image.GetType()),
+                      image.depth);
+        }
+    }
```
squidbus: Have you confirmed that the intended behavior would be to use slice 0? Also I don't see how base slice being more than layer count necessarily implies non-array texture.

src/video_core/texture_cache/image_view.cpp
```diff
@@ -71,8 +89,9 @@ ImageViewInfo::ImageViewInfo(const AmdGpu::Image& image, const Shader::ImageReso
 }
 
 ImageViewInfo::ImageViewInfo(const AmdGpu::ColorBuffer& col_buffer) noexcept {
-    range.base.layer = col_buffer.BaseSlice();
-    range.extent.layers = col_buffer.NumSlices() - range.base.layer;
+    const u32 base_slice = col_buffer.BaseSlice();
+    range.base.layer = base_slice;
+    range.extent.layers = col_buffer.NumSlices() - base_slice;
```
squidbus: What is the point of this, these have the same result since `range.base.layer` is set to `col_buffer.BaseSlice()`.

src/video_core/texture_cache/image_view.cpp
```diff
@@ -83,8 +102,9 @@ ImageViewInfo::ImageViewInfo(const AmdGpu::DepthBuffer& depth_buffer, AmdGpu::De
     format = Vulkan::LiverpoolToVK::DepthFormat(depth_buffer.z_info.format,
                                                 depth_buffer.stencil_info.format);
     is_storage = ctl.depth_write_enable;
-    range.base.layer = view.slice_start;
-    range.extent.layers = view.NumSlices() - range.base.layer;
+    const u32 base_slice = view.slice_start;
+    range.base.layer = base_slice;
+    range.extent.layers = view.NumSlices() - base_slice;
```
squidbus: Same here re: what is the point of this.

src/video_core/texture_cache/image.cpp
```diff
@@ -231,23 +231,41 @@ Image::Barriers Image::GetBarriers(vk::ImageLayout dst_layout, vk::AccessFlags2
...
+                                           info.resources.levels)
+                                : info.resources.levels;
+        const u32 layer_begin = needs_partial_transition ? subres_range->base.layer : 0u;
+        const u32 layer_end = needs_partial_transition
+                                  ? std::min(subres_range->base.layer + subres_range->extent.layers,
+                                             info.resources.layers)
+                                  : info.resources.layers;
+
+        const auto mips = std::views::iota(mip_begin, mip_end);
+        const auto layers = std::views::iota(layer_begin, layer_end);
```
squidbus, line 254: I guess it's fine to clamp it to the actual number of resources? Although I'd want to know where the bad ranges are coming from to confirm its not just a bug elsewhere in our code.

Hog185: The bad ranges in GetBarriers come specifically from non array Color2D textures with base_array > 0

PR #4456, opened 2026-05-20
Lib.VideoOut: Properly remove events on close

StevenMiller123:
This PR fixes Mass Effect™ Legendary Edition crashing when you try to load one of the games it contains.
image

### DIFF
--- a/src/core/libraries/videoout/driver.cpp
+++ b/src/core/libraries/videoout/driver.cpp
@@ -243,9 +243,10 @@ void VideoOutDriver::Flip(const Request& req) {
     }
 
     // Trigger flip events for the port.
-    for (auto& event : port->flip_events) {
-        if (event != nullptr) {
-            event->TriggerEvent(
+    for (auto event : port->flip_events) {
+        auto equeue = Kernel::GetEqueue(event);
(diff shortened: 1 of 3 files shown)


src/core/libraries/videoout/driver.cpp
```diff
@@ -71,17 +71,31 @@ void VideoOutDriver::Close(s32 handle) {
     std::memset(main_port.buffer_labels.data(), 0, sizeof(main_port.buffer_labels));
     std::memset(main_port.groups.data(), 0, sizeof(main_port.groups));
     std::memset(&main_port.flip_status, 0, sizeof(main_port.flip_status));
-    std::memset(&main_port.vblank_status, 0, sizeof(main_port.vblank_status));
+    main_port.flip_status = FlipStatus{};
```
squidbus, line 74: Seems redundant with the memset right above it? I assume should remove one or the other.

Also why a clear of flip status this replace clearing vblank status?

StevenMiller123: Oops, I messed that up. Will fix.

squidbus (approved):

PR #4457, opened 2026-05-20
Lib.Audio3d: implement sceAudio3dPortCreate and update sceAudio3dPortOpen

tatu16:


src/core/libraries/audio3d/audio3d.cpp
```diff
@@ -617,9 +617,27 @@ s32 PS4_SYSV_ABI sceAudio3dPortClose(const OrbisAudio3dPortId port_id) {
     return ORBIS_OK;
 }
 
-s32 PS4_SYSV_ABI sceAudio3dPortCreate() {
-    LOG_ERROR(Lib_Audio3d, "(STUBBED) called");
-    return ORBIS_OK;
+s32 PS4_SYSV_ABI sceAudio3dPortCreate(u32 max_objects, u32 queue_depth,
```
StevenMiller123: This doesn't seem to align with decompilation, where did you get this from?
From what I see, the header should be something like
```
s32 PS4_SYSV_ABI sceAudio3dPortCreate(u32 granularity, OrbisAudio3dRate rate, s64 param_3, OrbisAudio3dPortId* port_id);
```

Both the granularity and rate parameters are passed to sceAudio3dPortOpen as an OrbisAudio3dOpenParameters with `size_this = 0x10`.

StevenMiller123 (changes requested):

src/core/libraries/audio3d/audio3d.cpp
Xupie: okay that makes a lot more sense, i'm not sure how i came up with those

src/core/libraries/audio3d/audio3d.cpp
```diff
@@ -782,31 +796,150 @@ s32 PS4_SYSV_ABI sceAudio3dPortGetStatus() {
...
-    if (!parameters || !port_id) {
-        LOG_ERROR(Lib_Audio3d, "!parameters || !id");
+    OrbisAudio3dOpenParameters effective{
+        .size_this = 0x28,
+        .granularity = 0x100,
+        .rate = OrbisAudio3dRate::ORBIS_AUDIO3D_RATE_48000,
+        .max_objects = 512,
+        .queue_depth = 2,
+        .num_beds = 0,
+        .buffer_mode = OrbisAudio3dBufferMode::ORBIS_AUDIO3D_BUFFER_ADVANCE_AND_PUSH,
```
StevenMiller123: Default buffer_mode is `OrbisAudio3dBufferMode::NO_ADVANCE`, only the `size_this = 0x18` case hardcodes `OrbisAudio3dBufferMode::ORBIS_AUDIO3D_BUFFER_ADVANCE_AND_PUSH`

I suspect this is tied to having changed the open parameters struct.

src/core/libraries/audio3d/audio3d.h
```diff
@@ -36,9 +43,9 @@ struct OrbisAudio3dOpenParameters {
     OrbisAudio3dRate rate;
     u32 max_objects;
     u32 queue_depth;
-    OrbisAudio3dBufferMode buffer_mode;
-    int : 32;
     u32 num_beds;
+    u32 _pad;
+    OrbisAudio3dBufferMode buffer_mode;
```
StevenMiller123, line 45: Like mentioned above, why have you changed the open parameters struct?
Everything I've seen suggests main has the right struct. (For example, why would there be 3 valid buffer modes if sceAudio3dPortOpen requires `ORBIS_AUDIO3D_BUFFER_ADVANCE_AND_PUSH` or some mysterious unknown third value?)

StevenMiller123 (changes requested):
I think it's fine to add all the error checks, but make sure you're applying checks to the right values.

src/core/libraries/audio3d/audio3d.cpp
```diff
@@ -782,31 +796,143 @@ s32 PS4_SYSV_ABI sceAudio3dPortGetStatus() {
...
+        .rate = parameters->rate,
+        .max_objects = 512,
+        .queue_depth = 2,
+        .buffer_mode = OrbisAudio3dBufferMode::ORBIS_AUDIO3D_BUFFER_NO_ADVANCE,
+        ._pad = 0,
+        .num_beds = 2,
+    };
+
+    switch (parameters->size_this) {
+    case 0x10:
```
StevenMiller123, line 829: The bit-shifting logic the library does makes this more like a floor. Anything less than size 0x18, but greater than 0x10 would still fall in the 0x10 case.
Only sizes that would hit the error return are sizes less than 0x10 or size 0x36 and greater

StevenMiller123 (commented):
Looks pretty good, just one little comment. Probably is mergeable as-is though.

StevenMiller123 (approved):
Looks good

PR #4460, opened 2026-05-21
http part6

georgemoralis:
http epolls and connected functions

### DIFF
--- a/src/core/libraries/network/http.cpp
+++ b/src/core/libraries/network/http.cpp
@@ -179,6 +179,40 @@ static void SynthesizeTransportFailureResponse(HttpResponse& res) {
     res.all_headers_blob.clear();
 }
 
+// Map common HTTP status codes to strings for logs.
+static std::string HttpStatusLabel(int sc) {
+    switch (sc) {
+    case 0:
+        return "0 (no status)";
+    case 200:
+        return "200 OK";
+    case 204:
+        return "204 No Content";
+    case 301:
+        return "301 Moved Permanently";
+    case 302:
+        return "302 Found";
+    case 304:
+        return "304 Not Modified";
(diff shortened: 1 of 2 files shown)


src/core/libraries/network/http.cpp
```diff
@@ -689,6 +723,24 @@ int PS4_SYSV_ABI sceHttpSendRequest(int reqId, const void* postData, u64 size) {
...
+                constexpr u32 FailureEventBits =
+                    ORBIS_HTTP_NB_EVENT_RESOLVER_ERR | ORBIS_HTTP_NB_EVENT_HUP;
+                OrbisHttpNBEvent ev{};
+                ev.events = FailureEventBits;
+                ev.eventDetail = FailureEventBits;
+                ev.id = reqId;
+                ev.userArg = req_ptr->epoll_user_arg;
+                epoll_it->second->events.push_back(ev);
+                epoll_it->second->cv.notify_all();
+                LOG_INFO(Lib_Http, "pushed failure epoll event for reqId={} on epoll={}", reqId,
```
StevenMiller123: This function is doing a lot of logging at this point.
Might be nice to change some of these logs to debug to prevent too much spamming in user logs.

StevenMiller123 (commented):
Just a comment, I think the logic looks good in here.

src/core/libraries/network/http.cpp
georgemoralis: changed this one , as it is not probably neccesary atm , the others i still think i need them for further dev , so might be reduced in a few more prs but for now i need to trace stuff from games when someone post some logs :D

StevenMiller123 (approved):
Logic looks solid.

PR #4464, opened 2026-05-23
improve texture cache image expansion condition

Hog185:
this puts tony hawk skater pro 3+4 in menus (if you bypass the first pre existing issue) (that fix will come later)

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -335,7 +335,8 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
 
         // Size and resources are greater, expand the image.
         if (image_info.type == cache_image.info.type &&
-            image_info.resources > cache_image.info.resources) {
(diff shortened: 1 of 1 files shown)


src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -335,7 +335,8 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
 
         // Size and resources are greater, expand the image.
         if (image_info.type == cache_image.info.type &&
-            image_info.resources > cache_image.info.resources) {
+            image_info.tile_mode == cache_image.info.tile_mode &&
+            image_info.guest_size > cache_image.info.guest_size) {
```
squidbus, line 339: Tile mode equality is okay but I don't think this should be changed to guest size, if your guest size is greater but the resources are not there's something up.

raphaelthegreat: I think its more appropriate to check array mode and not tile mode. For example depth/color aliases have different tile modes (Thin2DThin vs Depth2DThin) but same array modes so they are compatible

PR #4467, opened 2026-05-23
Handle abs and clamp modifiers only for VOP3A instructions

mikusp:
VOP3B encoding repurposes abs and clamp bits for sdst, so we can only use them with VOP3A instructions. Another one suggested by @raphaelthegreat

### DIFF
--- a/src/shader_recompiler/frontend/decode.cpp
+++ b/src/shader_recompiler/frontend/decode.cpp
@@ -802,11 +802,19 @@ void GcnDecodeContext::decodeInstructionVOP3(uint64_t hexInstruction) {
 
     // update input modifier
     auto& control = m_instruction.control.vop3;
-    for (u32 i = 0; i != 3; ++i) {
-        if (control.abs & (1u << i)) {
-            m_instruction.src[i].input_modifier.abs = true;
-        }
 
+    // update output modifier
+    auto& outputMod = m_instruction.dst[0].output_modifier;
+
+    if (!IsVop3BEncoding(m_instruction.opcode)) {
+        for (u32 i = 0; i != 3; ++i) {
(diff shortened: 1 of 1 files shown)


src/shader_recompiler/frontend/decode.cpp
```diff
@@ -802,11 +802,19 @@ void GcnDecodeContext::decodeInstructionVOP3(uint64_t hexInstruction) {
...
+    // update output modifier
+    auto& outputMod = m_instruction.dst[0].output_modifier;
+
+    if (!IsVop3BEncoding(m_instruction.opcode)) {
+        for (u32 i = 0; i != 3; ++i) {
+            if (control.abs & (1u << i)) {
+                m_instruction.src[i].input_modifier.abs = true;
+            }
+        }
+        outputMod.clamp = static_cast<bool>(control.clmp);
```
squidbus, line 815: What's the source for the clamp bit, I was looking at Vega for example and clamp is still there for VOP3b.

Image

raphaelthegreat: I believe it's bit 11
| VOP3a | VOP3b |
| -------- | ------- |
| image  | Screenshot_20260524_003352  |

mikusp: Sea Islands ISA shows the overlap
Screenshot 2026-05-23 at 23 37 42
Screenshot 2026-05-23 at 23 37 31

squidbus: Hmm okay, I think I got it mixed up with the Neo VOP stuff and thought it wasn't there for Sea Islands, guess I should've checked closer.

squidbus (approved):

PR #4468, opened 2026-05-23
kernel: allow recursive read-locking on PthreadRwlock (fixes race condition in GR2)

junminlee2004:
MSVC's std::shared_timed_mutex won't grant recursive read-locks the way real PS4 pthread_rwlock does. Any PS4 title that relies on POSIX recursive-read semantics would experience a similar hang on windows. The fix adds a thread_local map tracking per-thread read-lock depth per rwlock. On rdlock/tryrdlock, if the thread already holds the lock for read, we bump the depth and return without touching the mutex; otherwise we acquire it for real.


src/core/libraries/kernel/threads/rwlock.cpp
```diff
@@ -10,6 +12,47 @@
...
+static thread_local std::unordered_map<const PthreadRwlock*, u32> g_rdlock_depth;
+
+// Returns true if the calling thread already holds a read lock on `lock`,
+// recording one more level of read recursion.
+static bool ReenterReadLock(const PthreadRwlock* lock) {
+    const auto it = g_rdlock_depth.find(lock);
+    if (it != g_rdlock_depth.end() && it->second > 0) {
+        ++it->second;
+        return true;
+    }
```
squidbus, line 31: Is this thread-safe? I see this check is called outside of a lock, and my understanding of C++ containers thread safety is that reading is not safe if another thread may be writing simultaneously.

junminlee2004: this one's safe: `g_rdlock_depth` is declared `static thread_local` (the declaration is a few lines above this snippet, so it's easy to miss in the inline view). Each thread gets its own independent instance of the map, so there's never more than one thread reading or writing a given container — the find/insert always happens on the owning thread, which is why no lock is needed.

squidbus: Ah, right. Sorry I missed that.

squidbus (approved):

raphaelthegreat:
Ideally the current STL mutex should be replaced by an implementation that matches freebsd on top of an address arbiter which would also solve this issue. But I'm not really sure how I feel about this change, would it be possible to use the existing shared_first_mutex for example to avoid this issue?

squidbus:
> would it be possible to use the existing shared_first_mutex for example to avoid this issue?

Good point that `SharedFirstMutex` may already have what we're looking for here, although I think it would need to be amended to support timed locking.

PR #4471, opened 2026-05-24
shader_recompiler: Pad number of TES passthrough outputs to fragment inputs if needed

squidbus:
Accounting for vertex exports in the quad/rect list tessellation shader I/O forwarding solved taking too many inputs from vertex, but in some games would give not enough outputs to fragment, which at least in MoltenVK can cause issues.

Instead of just taking the minimum of the two, use vertex exports for all forwarding except the tessellation evaluation outputs, where we will output all of the vertex exports and then also pad any remaining fragment inputs with zero values. This should ensure we only take as many values as vertex exports, and give as many values as fragment wants.


shinra-electric:
Fixes regression in Yakuza 0 with MVK

Screenshot 2026-05-24 at 18 15 54@0 5x

PR #4479, opened 2026-05-25
video_core: check if depth image is valid before using it

LNDF:


src/common/incremental_id.h
```diff
@@ -0,0 +1,22 @@
+// SPDX-FileCopyrightText: Copyright 2026 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+#include <atomic>
+
+namespace Common {
+template <typename CounterType>
```
raphaelthegreat, line 8: This is a superfluous abstraction imo

src/video_core/texture_cache/image.h
```diff
@@ -111,8 +112,14 @@ struct Image {
         return True(flags & ImageFlagBits::GpuModified) && False(flags & (ImageFlagBits::Dirty));
     }
 
-    void AssociateDepth(ImageId image_id) {
-        depth_id = image_id;
+    void AssociateDepth(const Image& depth_image) {
+        depth_id = depth_image.image_uid;
+        depth_uid = depth_image.image_uid;
```
squidbus: These change are confusing me a bit:
* What is the new UID counter for?
* Why are both of these fields set to it instead of just having one field?
* Later down you do `slot_images[image.depth_id]`, how are we guaranteeing here that the UID you generated is the right index in `slot_images`, and how is it different from just using the image ID from before?

LNDF: Actually you are right. I made a mistake when bringing over the patch from a testing branch. The first one should be image_id

squidbus: I assume we don't need the UID counter at all in this PR then, since `depth_uid` is not used anywhere.

LNDF: Pushed change with correct usage of uid

LNDF: The idea is to know whether an image that is associated with depth_id is still the same image or it has been freed/it is a different image.

squidbus: I see. I feel like we shouldn't need two IDs to accomplish this, but it is the texture cache... whole thing needs refactoring anyway.

squidbus (approved):

LNDF:
@StevenMiller123 please test with uncharted 3 to see whether it is better than main

src/common/incremental_id.h
LNDF: Maybe yes.

Is image creation thread safe?

StevenMiller123:
Seems to improve the stability of Uncharted 3 (from the Nathan Drake Collection). I was able to get through the bathroom fight without crashing or hanging like main does.
image
[CUSA02320.log](https://github.com/user-attachments/files/28272329/CUSA02320.log)

Here's a log from main trying the same fight. There I can't get through it because I consistently run into issues (in this case, it was an exception)
[CUSA02320 main.log](https://github.com/user-attachments/files/28272685/CUSA02320.main.log)

@Xcedf should be able to check too (if they're willing to test). I believe they're the one who initially informed me of this bug.

Xcedf:
@StevenMiller123 yes it's true, finished the toilet fight, unfortunately game doesn't start chapter 2, crashes after movie with the same error as I had in Uncharted 2.
[Debug] <Critical> (NdJobWorkerThread 0) signals.cpp:63 SignalHandler: Unhandled Exception code 0xc0000005 at 0x800e3b8fa

LNDF:
Both of those @StevenMiller123  @Xcedf  seem like internal game asserts. IIRC this game was very sensitive to GPU delays. 

Anyway, it doesn't seem related to the issue that this PR is fixing.

tochnonechelovek:
WATCH_DOGS™ (CUSA00016) now doesn't crash or hang (at least until I left the game myself after some time). Also, somewhat similar visual glitches to PAYDAY 2: CRIMEWAVE EDITION (CUSA01761). Not sure if it's from this PR, I haven't gotten that far before (and can't on main). Some vertex explosions and I think they're from this PR.

изображение
изображение

Log from main in the cutscene where it hangs usually:
[CUSA00016 main.log](https://github.com/user-attachments/files/28287533/CUSA00016.main.log)

Log from PR in the same cutscene in the same moment:
[CUSA00016.log](https://github.com/user-attachments/files/28287746/CUSA00016.log)

PR #4480, opened 2026-05-26
Attempt to fix yet another input regression in some Unity games

kalaposfos13:
Hopefully fixes the regression found in Slime Rancher, caused by #4410.

This PR is currently waiting for either @brad0demx or @StevenMiller123 to wake up and test it.

### DIFF
--- a/src/core/libraries/pad/pad.cpp
+++ b/src/core/libraries/pad/pad.cpp
@@ -51,11 +51,6 @@ static std::optional<HandleKey> FindHandleKeyByHandle(s32 handle) {
     return std::nullopt;
 }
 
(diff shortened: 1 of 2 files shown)


kalaposfos13:
The fix was to swap the default value of pInfo->deviceClass back to Standard instead of Invalid, but the rest of the changes should be valid as well, so I'm keeping them.

StevenMiller123 (approved):
Tested locally and it seems to work properly. Someone should test RE1HD to make sure that's still functional though.

w1naenator:
I don't have `RE1HD`, but `Pumped BMX +` still OK!

PR #4491, opened 2026-05-30
Compute->Texture Data Sync for God of War III

TdlifePeng:
# Compute→Texture Data Sync for God of War III

## Problem

PS4 GCN unified memory allows a compute shader to write a storage image in
one format (R32G32B32A32Uint) and a graphics shader to read the same memory
as a texture in a different format (BC3/BC1). Vulkan prohibits cross-format
aliasing — the emulator creates separate VkImages. Compute writes go to the
storage VkImage (GPU-local), but texture Refresh reads guest memory which is
never updated with the compute output. BC3 textures sample stale data.

In God of War III Remastered, compute shader `ccdebd80f02a77aa` performs a
cascaded dispatch filling 4 MB R32 buffers. The game then creates BC3 textures
at overlapping addresses via an A/B double-buffer pattern, expecting the
compute output as their data source.

## Approach

After a direct compute dispatch, record `vkCmdCopyImageToBuffer` from the
storage VkImage to a staging buffer, then memcpy into guest memory. Subsequent
texture Refresh finds the updated data and uploads to the BC3 VkImage.

Copies from multiple dispatches are batched into the same command buffer
without flushing. At the start of the next graphics `BindResources`, a single
`scheduler.Wait(tick)` submits the batch and waits for GPU completion, then
memcpys and marks overlapping textures dirty via `InvalidateMemoryRange`.

Gated by game serial: CUSA01623 / CUSA01715 / CUSA01740. Other games incur
zero overhead.

## Limitations

- Simplified implementation based on assumptions about the known compute
  shader behavior. Completeness and performance not thoroughly tested.
- Opening battle scene looks correct. Some particles appear black in later
  scenes — may be caused by other compute shaders not yet handled.
- Only mip 0 / layer 0 is downloaded — the compute shader produces a single
  flat R32 surface and does not generate higher mips. Any higher mips of the
  overlapping BC3 texture would need a separate mipmap generation pass, which
  is outside the scope of this fix.

An ideal solution would keep the entire pipeline on the GPU — a compute
shader that directly reads the R32 storage image and writes BC3-encoded
blocks to the texture VkImage, eliminating the round-trip through guest
memory and CPU.

**For reference only.**

---

[Claude Code, deepseek-v4-pro]

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1042,6 +1042,8 @@ set(VIDEO_CORE src/video_core/amdgpu/cb_db_extent.h
                src/video_core/renderer_vulkan/liverpool_to_vk.h
                src/video_core/renderer_vulkan/vk_common.cpp
                src/video_core/renderer_vulkan/vk_common.h
+               src/video_core/renderer_vulkan/vk_compute_download.cpp
+               src/video_core/renderer_vulkan/vk_compute_download.h
                src/video_core/renderer_vulkan/vk_compute_pipeline.cpp
                src/video_core/renderer_vulkan/vk_compute_pipeline.h
                src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
--- a/src/video_core/renderer_vulkan/vk_compute_download.cpp
+++ b/src/video_core/renderer_vulkan/vk_compute_download.cpp
@@ -5,9 +5,7 @@
 #include "common/logging/log.h"
 #include "core/memory.h"
 #include "video_core/renderer_vulkan/vk_compute_download.h"
-#include "video_core/renderer_vulkan/vk_instance.h"
 #include "video_core/renderer_vulkan/vk_scheduler.h"
-#include "video_core/renderer_vulkan/vk_platform.h"
 #include "video_core/texture_cache/image.h"
 #include "video_core/texture_cache/texture_cache.h"
 #include <vk_mem_alloc.h>
@@ -16,7 +14,7 @@ namespace Vulkan {
 
 namespace {
 struct PendingDownload {
-    vk::Fence fence{};
+    u64 download_tick{};     // scheduler tick when the copy was recorded
     VkBuffer staging_buffer{};
     VmaAllocation staging_alloc{};
     VmaAllocator vma_alloc{};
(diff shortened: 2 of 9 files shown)


squidbus:
Please do not rely on AI to solve these problems. For one, this seems to be under a false premise:

> PS4 GCN unified memory allows a compute shader to write a storage image in
one format (R32G32B32A32Uint) and a graphics shader to read the same memory
as a texture in a different format (BC3/BC1). Vulkan prohibits cross-format
aliasing

`VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT`, which we already make some use of, lets you create a compressed format image that can be aliased as uncompressed. In this situation the image should get replaced in the cache by a compressed one and interpreted as necessary for each operation.

raphaelthegreat:
Wasn't this implemented here https://github.com/shadps4-emu/shadPS4/pull/3056
This sounds like the exact same case. Can you describe what access pattern you encountered where this didnt work? GOW3 texture issues are mostly due to missing subrect renders last time I looked.
Also you mention different cases in PR description, writing to a storage image and buffer are different alias cases for BCn texture

TdlifePeng:
Screenshot 2026-05-30 221149
Screenshot 2026-05-30 221325

on amd 395

Hog185:
@TdlifePeng can you test it against nightly (0.15.1) not 0.15.0

TdlifePeng:
black screen

https://github.com/shadps4-emu/shadPS4/pull/4371

[Lib.NpTrophy] <Error> (Game:Main) np_trophy.cpp:228 sceNpTrophyCreateContext: No npCommId found for trophy index/service_label: 0

StevenMiller123:
It if works on this PR, then there's no reason it wouldn't boot on current WIP with proper settings. This PR is based on current WIP after all.

TdlifePeng:
after patch #4371

Xcedf:
@StevenMiller123 actualy vs the current main it's the same as vs 0.15.0, tried, the easiest way to reproduce issue is to change language in menu, no need to boot into game in this case.

EmulationChannell:
Captura de tela 2026-05-31 070546
Captura de tela 2026-05-31 070944
Captura de tela 2026-05-31 091851

[shad_log.txt](https://github.com/user-attachments/files/28437324/shad_log.txt)

Mohammedgfs:
reflection doesn't work 
Screenshot (70)

TdlifePeng:
could you share a save file at the screenshot location

TdlifePeng:
but if there's already a preferred direction

Mohammedgfs:
save data file iam using amd ryzen 5 7500f and rtx 5060 ti 16gb
[SD-260107120458.zip](https://github.com/user-attachments/files/28450266/SD-260107120458.zip)

TdlifePeng:
analysis on a 1060 suggests a cache-coherency-like cpu/gpu synchronization issue in the simplified implementation, leading to corrupted reflection textures (⁠٥⁠↼⁠_⁠↼⁠)

Mohammedgfs:
Is there a fix for it

TdlifePeng:
a sync-free solution, a gpu-only approach may be possible. need some time to investigate

Mohammedgfs:
Thanks

TdlifePeng:
xp with ai sometimes makes me want to fight it. time to visit Master Oogway ^⁠_⁠^

SoulToxic3119:
image
image
image
Here you can see that in version 8260d09 WIP 0.16.1 the excessive bloom was greatly improved, but there are still areas where the textures break, such as in the garden area where the statue gem is used, the glow of the hidden chests flickers, and the water is not textured correctly.
[shad_log.txt](https://github.com/user-attachments/files/28802962/shad_log.txt)

TdlifePeng:
Sorry, I only just saw this.

Could you try the latest simplified approach? Or, if possible, share a save file so I can take a look on my end.

https://github.com/shadps4-emu/shadPS4/pull/4591

PR #4492, opened 2026-05-30
Increase default inner deadzone

kalaposfos13:
https://github.com/shadps4-compatibility/shadps4-game-compatibility/issues/2545#issuecomment-4583124044


gandalfthewhite19890404:
Why not make it configurable? Or suggested values are viable for everyone?

kalaposfos13:
It is already configurable, this change is only so that the default aligns more closely with the PS4's default, so the requirement of the emulator being "vanilla" and on default settings for the compatibility list doesn't have a default so low, that stick drift can happen in games that don't have their own internal deadzones.

PR #4496, opened 2026-05-31
Implement save migration

kalaposfos13:
image

Symlinking requires admin privileges on Windows, but that fortunately can be solved by simply not using that OS. The wording is a bit awkward, so I'm open to suggestions on that part.

### DIFF
--- a/src/core/user_manager.cpp
+++ b/src/core/user_manager.cpp
@@ -25,11 +25,11 @@ bool UserManager::AddUser(const User& user) {
     const auto user_dir = EmulatorSettings.GetHomeDir() / std::to_string(user.user_id);
 
     std::error_code ec;
-    if (!fs::exists(user_dir)) {
-        fs::create_directory(user_dir, ec);
-        fs::create_directory(user_dir / "savedata", ec);
-        fs::create_directory(user_dir / "trophy", ec);
-        fs::create_directory(user_dir / "inputs", ec);
+    if (!std::filesystem::exists(user_dir)) {
+        std::filesystem::create_directory(user_dir, ec);
+        std::filesystem::create_directory(user_dir / "savedata", ec);
+        std::filesystem::create_directory(user_dir / "trophy", ec);
+        std::filesystem::create_directory(user_dir / "inputs", ec);
     }
 
     Save();
@@ -44,9 +44,9 @@ bool UserManager::RemoveUser(s32 user_id) {
 
     const auto user_dir = EmulatorSettings.GetHomeDir() / std::to_string(user_id);
 
-    if (fs::exists(user_dir)) {
+    if (std::filesystem::exists(user_dir)) {
         std::error_code ec;
-        fs::remove_all(user_dir, ec);
+        std::filesystem::remove_all(user_dir, ec);
     }
 
     m_users.user.erase(it, m_users.user.end());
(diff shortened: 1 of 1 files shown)


StevenMiller123:
I feel it would be better to do an ifdef and just remove the "move and link back" option on Windows.

kalaposfos13:
Well I did that, although I'd really appreciate if someone on Windows could actually test it.

src/core/user_manager.cpp
```diff
@@ -3,13 +3,16 @@
...
 #include <iostream>
+#include <SDL3/SDL_messagebox.h>
 #include <common/assert.h>
 #include <common/path_util.h>
 #include "emulator_settings.h"
 #include "libraries/system/userservice.h"
 #include "user_manager.h"
 #include "user_settings.h"
 
+namespace fs = std::filesystem;
```
Niram7777, line 14: I would rather not use this, or at minimum separate it in another commit - to keep review and blame clean

src/core/user_manager.cpp
```diff
@@ -138,11 +141,81 @@ Users UserManager::CreateDefaultUsers() {
...
+                        }
+                        fs::create_directory_symlink(new_save_dir, old_save_dir);
+                        break;
+                    case -1:
+                    case 3:
+                        break;
+                    default:
+                        UNREACHABLE();
+                    }
+                } catch (std::exception& e) {
```
Niram7777: const std::exception& e

src/core/user_manager.cpp
```diff
@@ -138,11 +141,81 @@ Users UserManager::CreateDefaultUsers() {
...
+#ifndef _WIN32
+                SDL_MessageBoxButtonData btns[4]
+#else
+                SDL_MessageBoxButtonData btns[3]
+#endif
+                    {
+                        {0, 0, "Copy"},
+                        {0, 1, "Move"},
+#ifndef _WIN32
+                        {0, 2, "Move and link back"},
```
Niram7777, line 163: for simplicity I would have done Copy/Move only

src/core/user_manager.cpp
```diff
@@ -138,11 +141,81 @@ Users UserManager::CreateDefaultUsers() {
...
-            std::filesystem::create_directory(user_dir / "savedata");
-            std::filesystem::create_directory(user_dir / "trophy");
-            std::filesystem::create_directory(user_dir / "inputs");
+        if (!fs::exists(user_dir)) {
+            fs::create_directory(user_dir);
+            fs::create_directory(user_dir / "savedata");
+            fs::create_directory(user_dir / "trophy");
+            fs::create_directory(user_dir / "inputs");
+            auto const old_save_dir =
+                Common::FS::GetUserPath(Common::FS::PathType::UserDir) / "savedata" / "1";
```
Niram7777, line 150: every folder in savedata should be migrated no ?

kalaposfos13: Apart from people who used my multiple controllers PR before it was integrated into shadow's new user manager should only have one folder in the old savedata location, for user id 1.

src/core/user_manager.cpp
kalaposfos13: Move and link back is very useful though, as you get to use the same saves on both before and after 0.15.0 builds.

PR #4498, opened 2026-05-31
Core: Clean up trophy extraction logic

StevenMiller123:
This code has been an absolute mess since the trophy rework. This eliminates several redundant loops through game trophy files, and separates the logic into it's own function.

I've done some minimal testing to make sure extraction still occurs when it should, not sure if I broke anything related to failed trophy extraction though.


Linear524:
Last Guardian is broken after this...

StevenMiller123:
Known regression, see #4504

PR #4499, opened 2026-05-31
Core: Remove old config code

StevenMiller123:
This PR removes the code for our old config backend. Trophy key migration now uses the toml library directly, like our config transfer code.


StevenMiller123:
This also fixes a very minor bug where new users would get the config migration popup.
That issue was caused by the old trophy migration code creating the old toml-based config before we try making the new one.

kalaposfos13 (approved):

Issue #4503, opened 2026-06-01
[GAME BUG]: God Of War III Remastered (CUSA01715): Black Screen


Gamegami49:
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.
- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [x] I have disabled all patches and cheats and the issue is still present.
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadPS4/wiki/I.-Quick-start-%5BUsers%5D#4-dumping-firmware-modules) installed.

### Describe the Bug

The game hangs on a black screen right after the Santa Monica Studio logo.

Log consistently displays "[Kernel.Fs] <Error> (snd_stream_parsing_thread) file_system.cpp:714 sceKernelStat: error = 2" twice right when it occurs.

### Reproduction Steps

1. Update shadPS4's version list
2. Use either latest Nightly (9a6d2d1) or 0.16.0 (Plutie-fueled)
3. Run the game and wait until the Santa Monica Studio logo
4. Black screen occurs

### Specify OS Version

Windows 10 Pro 22H2

### CPU

AMD Ryzen 7 5800X

### GPU

nVidia GeForce RTX 5070

### Amount of RAM in GB

64

### Amount of VRAM in GB

12

### Log File

[CUSA01715.log](https://github.com/user-attachments/files/28474701/CUSA01715.log)

StevenMiller123:
All the information I've gathered about this issue suggests that it comes from having a bad game dump. Specifically, the data in your game's `sce_sys/npbind.dat` is likely corrupt.

Please redump your game from your PS4, then retest this.

StevenMiller123:
Might be some merit in looking closer. Reopening for now while I investigate.

Akilan2006:
the game works perfectly without blackscreen issues in CLI version 0.15.0 but
not in version 0.16.0

azaelbr:
Here, God of War worked up to version 15v, although it had some texture bugs. I redid all the settings and tried every possible configuration and combination, but the game won’t get past the Santa Monica screen.

7800X3D
RTX 4070 Ti
32 GB RAM

[...]

PR #4504, opened 2026-06-01
Lib.NpTrophy: Move NP-related checks to sceNpTrophyRegisterContext

StevenMiller123:
Using OpenOrbis homebrew, I've verified that sceNpTrophyCreateContext will run successfully without any NP-related data files. The first checks against NP-related files appear to occur in sceNpTrophyRegisterContext.

Also fixed some regressions for users without a trophy key, which were caused by #4498


cuesta4:
Likely fixes #4503 and others as well.

georgemoralis:
seems ok

PR #4506, opened 2026-06-01
Add error logging for unsupported file descriptors

Hog185:
Log an error when an unsupported file descriptor is encountered.

### DIFF
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -1454,7 +1454,7 @@ s32 PS4_SYSV_ABI posix_select(s32 nfds, fd_set* readfds, fd_set* writefds, fd_se
                 }
             }();
             if (native_fd == -1) {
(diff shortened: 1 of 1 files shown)


Hog185:
uhhh can someone review this pls :3 its only small

src/core/libraries/kernel/file_system.cpp
```diff
@@ -1454,6 +1454,7 @@ s32 PS4_SYSV_ABI posix_select(s32 nfds, fd_set* readfds, fd_set* writefds, fd_se
                 }
             }();
             if (native_fd == -1) {
+                LOG_ERROR(Kernel_Fs, "Unsupported fd %d", i);
```
Niram7777: you  should use `{}`
https://fmt.dev/12.0/

Niram7777 (changes requested):

Niram7777 (commented):
probably better as LOG_WARNING
otherwise LGTM

Niram7777 (approved):

PR #4507, opened 2026-06-01
core/memory: handle fixed mappings with zero virtual address

cuesta4:
## Summary

Fixes a boot crash in Dead Nation: Apocalypse Edition (CUSA00176) caused by a fixed memory mapping request with `virtual_addr == 0`.

## Details

`MapMemory()` previously skipped the free-address search path when `MemoryMapFlags::Fixed` was set, even if the requested virtual address was zero. That allowed the mapping path to continue with address `0`, which later reached VMA lookup/creation logic that does not handle addresses below the VMA map start correctly.

This changes the non-fixed allocation branch to also handle `virtual_addr == 0`:

```cpp
} else if (False(flags & MemoryMapFlags::Fixed) || virtual_addr == 0) {
```

This keeps normal fixed mappings unchanged, but treats a zero requested address as needing allocator-selected placement.

## Rationale

A zero virtual address is not a valid final mapping target here. Falling back to the existing free-address search path matches the behavior already used for non-fixed mappings and avoids creating/looking up VMAs at an invalid low address.

## Testing

- Built shadPS4 successfully in Release mode.
- Verified Dead Nation: Apocalypse Edition (CUSA00176) boots past the previous crash.
- Verified Bloodborne (CUSA00900) still boots.

Made with the help of Codex

### DIFF
--- a/src/core/libraries/kernel/memory.cpp
+++ b/src/core/libraries/kernel/memory.cpp
@@ -169,7 +169,14 @@ s32 PS4_SYSV_ABI sceKernelReserveVirtualRange(void** addr, u64 len, s32 flags, u
 
     auto* memory = Core::Memory::Instance();
     const VAddr in_addr = reinterpret_cast<VAddr>(*addr);
-    const auto map_flags = static_cast<Core::MemoryMapFlags>(flags);
+    auto map_flags = static_cast<Core::MemoryMapFlags>(flags);
+
+    if (True(map_flags & Core::MemoryMapFlags::Fixed) && in_addr == 0) {
+        if (Common::ElfInfo::FW_170 <= g_sdk_version) {
+            return ORBIS_KERNEL_ERROR_EINVAL;
+        }
+        map_flags &= ~Core::MemoryMapFlags::Fixed;
+    }
 
     s32 result = memory->MapMemory(addr, in_addr, len, Core::MemoryProt::NoAccess, map_flags,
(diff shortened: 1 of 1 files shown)


src/core/memory.cpp
```diff
@@ -554,7 +554,7 @@ s32 MemoryManager::MapMemory(void** out_addr, VAddr virtual_addr, u64 size, Memo
             LOG_ERROR(Kernel_Vmm, "Unable to map {:#x} bytes at address {:#x}", size, virtual_addr);
             return ORBIS_KERNEL_ERROR_ENOMEM;
         }
-    } else if (False(flags & MemoryMapFlags::Fixed)) {
```
StevenMiller123, line 557: The appropriate fix (based on libkernel decomp) is to go through and add the following code snippet to sceKernelReserveVirtualRange, sceKernelMapNamedDirectMemory, and sceKernelMapNamedFlexibleMemory before their MapMemory calls:
```
if (True(map_flags & Core::MemoryMapFlags::Fixed) && in_addr == 0) {
    if (Common::ElfInfo::FW_170 <= g_sdk_version) {
        return ORBIS_KERNEL_ERROR_EINVAL;
    }
    map_flags &= ~Core::MemoryMapFlags::Fixed;
}
```

StevenMiller123 (changes requested):
Kinda shocked this case actually came up somewhere.

StevenMiller123 (approved):
Looks good to me.

PR #4508, opened 2026-06-01
input: reduce pad state latency for single-sample reads

cuesta4:
## Problem
`scePadRead` could return controller input roughly 125–250 ms out of date, most visibly in games that request one sample per frame. The controller timer produced separate gyro and acceleration snapshots every 4 ms, filling the 64-entry FIFO faster than single-sample callers consumed it and causing them to repeatedly receive the oldest queued state. Controller connection state was also maintained separately from each snapshot, which could produce inconsistent historical reports and left stale input available across reconnections.

## Fix
Coalesce periodic controller updates into one synchronized `PollState()` snapshot per tick. Single-sample reads now return the current state without consuming history, while multi-sample reads preserve FIFO ordering. Connection metadata, motion orientation, and touch timing are captured in each report; disconnected controllers continue producing neutral reports without logging out their users; and reconnecting clears old history and transient input while preserving the connection count. Initial controller discovery now occurs during pad library registration, after the kernel clock is initialized. The controller state and queue share one mutex, and obsolete touch, motion, queue, and connection bookkeeping was removed.

## Testing
No observed regressions in my games during manual testing.

### DIFF
--- a/src/core/libraries/pad/pad.cpp
+++ b/src/core/libraries/pad/pad.cpp
@@ -13,6 +13,7 @@
 #include "pad.h"
 
 #include <algorithm>
+#include <array>
 #include <optional>
 
 namespace Libraries::Pad {
@@ -137,10 +138,7 @@ int PS4_SYSV_ABI scePadGetControllerInformation(s32 handle, OrbisPadControllerIn
     if (it == handle_to_controller_map.end()) {
         return ORBIS_PAD_ERROR_INVALID_HANDLE;
     }
-    bool connected = false;
-    int connected_count = 0;
-    Input::State state{};
-    it->second->ReadState(&state, &connected, &connected_count);
+    const Input::State state = it->second->ReadState();
 
     std::memset(pInfo, 0, sizeof(OrbisPadControllerInformation));
     pInfo->touchPadInfo.pixelDensity = 1;
@@ -149,10 +147,10 @@ int PS4_SYSV_ABI scePadGetControllerInformation(s32 handle, OrbisPadControllerIn
     pInfo->stickInfo.deadZoneLeft = 1;
     pInfo->stickInfo.deadZoneRight = 1;
     pInfo->connectionType = ORBIS_PAD_CONNECTION_TYPE_LOCAL;
-    pInfo->connectedCount = static_cast<u8>(std::clamp(connected_count, 0, 0xff));
+    pInfo->connectedCount = state.connected_count;
     pInfo->deviceClass = OrbisPadDeviceClass::Standard;
-    pInfo->connected = connected;
-    if (connected) {
+    pInfo->connected = state.connected;
+    if (state.connected) {
         pInfo->deviceClass = EmulatorSettings.IsUsingSpecialPad()
                                  ? (OrbisPadDeviceClass)EmulatorSettings.GetSpecialPadClass()
                                  : OrbisPadDeviceClass::Standard;
@@ -367,26 +365,27 @@ int PS4_SYSV_ABI scePadOutputReport() {
     return ORBIS_OK;
 }
 
-int ProcessStates(s32 handle, OrbisPadData* pData, Input::GameController& controller,
-                  Input::State* states, s32 num, bool connected, u32 connected_count) {
-    if (!connected) {
-        pData[0] = {};
-        pData[0].orientation = {0.0f, 0.0f, 0.0f, 1.0f};
-        pData[0].connected = false;
-        return 1;
-    }
-
+int ProcessStates(OrbisPadData* pData, const Input::State* states, s32 num) {
     const bool gamepad_input_intercepted = ImGui::Core::IsGamepadInputCaptured();
     for (int i = 0; i < num; i++) {
+        pData[i] = {};
+        if (!states[i].connected) {
+            pData[i].leftStick = {128, 128};
+            pData[i].rightStick = {128, 128};
+            pData[i].orientation = {0.0f, 0.0f, 0.0f, 1.0f};
+            pData[i].connected = false;
+            pData[i].timestamp = states[i].time;
+            pData[i].connectedCount = states[i].connected_count;
+            continue;
+        }
         if (gamepad_input_intercepted) {
-            pData[i] = {};
             pData[i].buttons = OrbisPadButtonDataOffset::Intercepted;
             pData[i].leftStick = {128, 128};
             pData[i].rightStick = {128, 128};
             pData[i].orientation = {0.0f, 0.0f, 0.0f, 1.0f};
-            pData[i].connected = connected;
+            pData[i].connected = states[i].connected;
             pData[i].timestamp = states[i].time;
-            pData[i].connectedCount = connected_count;
+            pData[i].connectedCount = states[i].connected_count;
             pData[i].deviceUniqueDataLen = 0;
             continue;
         }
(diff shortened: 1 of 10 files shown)


kalaposfos13:
The code looks correct overall, someone will have to check if the games for which the motion input polling was added didn't break from the halved event count though.

roamic:
> The code looks correct overall, someone will have to check if the games for which the motion input polling was added didn't break from the halved event count though.

The `PopLatest` trick in the `ReadStates` seems like a hack to me.

kalaposfos13:
You're right, the implementation is a hack, and it should simply get the latest state, not pop it (that doesn't even pop it, it just straight up wipes the queue), although for what it's worth, I don't think any game will ever alternate using scePadRead and scePadReadState to actually encounter incorrect data.

roamic:
> You're right, the implementation is a hack, and it should simply get the latest state, not pop it (that doesn't even pop it, it just straight up wipes the queue), although for what it's worth, I don't think any game will ever alternate using scePadRead and scePadReadState to actually encounter incorrect data.

It shouldn't get the latest state. It should take the oldest states whether or not num is 1 or greater.

kalaposfos13:
My bad, I somehow managed to not see that this is scePadRead, not scePadReadState, you are right.

georgemoralis:
@kalaposfos13 can you re-review that?

cuesta4:
@StevenMiller123 

Could you please test (https://github.com/shadps4-compatibility/shadps4-game-compatibility/issues/46)?

src/input/controller.cpp
```diff
@@ -82,72 +83,83 @@ void State::UpdateAxisSmoothing() {
...
+        return 0;
     }
-    return ret_num;
+
+    // scePadRead commonly asks for one sample. Consume the newest queued snapshot directly
+    // instead of popping the stale backlog one element at a time.
+    if (states_num == 1) {
+        states[0] = *m_states_queue.PopLatest();
+        return 1;
+    }
```
roamic, line 84: This looks like a hack to me. I can't find any sources that suggest this should be the case. Can you provide any?

StevenMiller123:
Input latency issues in Kingdom Hearts seem to be addressed, can't test the titles @kalaposfos13 was concerned about though.
[CUSA05933.log](https://github.com/user-attachments/files/29973158/CUSA05933.log)

src/input/controller.cpp
roamic: Never mind. I found in the decompilation that `scePadReadState` calls `scePadRead` with count = 1. So seems like passing `1` is a special case where it should return current state. Still, I don't think you should pop the latest state. Instead, just return the current state.

src/input/controller.cpp
```diff
@@ -82,72 +83,85 @@ void State::UpdateAxisSmoothing() {
...
+    std::scoped_lock l(m_state_mutex);
     std::memcpy(accel_buf, acceleration, sizeof(accel_buf));
 }
 
-void GameController::UpdateAxisSmoothing() {
-    m_state.UpdateAxisSmoothing();
+void GameController::PollState() {
+    std::lock_guard lg(m_state_mutex);
+    // A disconnect is a single state transition. Do not continuously manufacture identical
+    // disconnected samples: scePadRead returns zero until the state changes again.
```
kalaposfos13: This is incorrect, that's exactly what happens on real hardware.

tests/test_ring_buffer_queue.cpp
kalaposfos13, line 1: Why do we need this test? This is testing a very likely never changing generic container class, not emulator code itself.

src/input/controller.cpp
```diff
@@ -291,6 +329,7 @@ void GameControllers::TryOpenSDLControllers() {
             }
             if (!still_connected) {
                 auto u = UserManagement.GetUserByID(controllers[i]->user_id);
+                UserManagement.LogoutUser(u);
```
kalaposfos13: This was removed intentionally.
https://github.com/shadps4-emu/shadPS4/pull/4532

kalaposfos13 (changes requested):
Aside from these three observations, I have no further issues, and the aforementioned games seem to behave correctly too.

src/input/controller.cpp
```diff
@@ -1,155 +1,16 @@
...
-            continue;
-        }
-        auto now = Libraries::Kernel::sceKernelGetProcessTime();
-        f32 t =
-            std::clamp((now - axis_smoothing_start_times[i]) / f32{axis_smoothing_time}, 0.f, 1.f);
-        axes[i] = s32(axis_smoothing_start_values[i] * (1 - t) + axis_smoothing_end_values[i] * t);
-    }
-}
-
-GameController::GameController() : m_states_queue(64) {}
```
kalaposfos13, line 92: I'm fine with Input::State getting its own controller_state.cpp file, but is there any particular reason as to why GameController code was also partially moved there? It doesn't make any sense in my opinion, and the only part that is probably fine is CalculateOrientation, but then you should change which class it belongs to as well, since it's not like that matters much, as it's a pure and static function.

src/sdl_window.cpp
```diff
@@ -173,7 +171,8 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameControllers* controller
     // input handler init-s
     Input::ControllerOutput::LinkJoystickAxes();
     Input::ParseInputConfig(std::string(Common::ElfInfo::Instance().GameSerial()));
-    controllers.TryOpenSDLControllers();
+    // Initial SDL discovery precedes the emulated kernel clock, so defer its first report.
```
kalaposfos13: This is a valid observation, however, I think simply deferring the first call of this to Pad::RegisterLib() is an easier solution than adding a function parameter just for this. Also, this attempted fix doesn't work anyway, because the motion sensor state generator loop starts right after and will supply garbage timestamps, but even that doesn't matter, as the queue will get filled up with correct states long before the library is actually initialized from the guest side, so all this does is failing to fix a bug that only exists in theory.

kalaposfos13 (changes requested):

src/input/controller.cpp
cuesta4: My bad, did that so I could run offline tests without messing with SDL discovery and UserService dependencies, should've reworked it after

src/core/libraries/pad/pad.cpp
```diff
@@ -367,26 +365,27 @@ int PS4_SYSV_ABI scePadOutputReport() {
...
     for (int i = 0; i < num; i++) {
+        pData[i] = {};
+        if (!states[i].connected) {
+            pData[i].leftStick = {128, 128};
+            pData[i].rightStick = {128, 128};
+            pData[i].orientation = {0.0f, 0.0f, 0.0f, 1.0f};
+            pData[i].connected = false;
+            pData[i].timestamp = states[i].time;
+            pData[i].connectedCount = states[i].connected_count;
+            continue;
```
kalaposfos13: This is still incorrect, just refer to the original code here, a disconnected handle will always return exactly one state, regardless of how many is actually requested. Honestly, I don't even know why you even changed this in the first place, as it's clearly unrelated to the main focus of the PR.

kalaposfos13 (changes requested):

src/core/libraries/pad/pad.cpp
cuesta4: Sorry for the repeated mistakes. Is this what you expected?

cuesta4:
@kalaposfos13 sorry for the tag, mind taking a look?

src/input/input_handler.cpp
```diff
@@ -846,18 +846,18 @@ void ControllerOutput::FinalizeUpdate(u8 gamepad_index) {
...
             ApplyDeadzone(new_param, righttrigger_deadzone[gamepad_index]);
-            controller->Axis(c_axis, GetAxis(0x0, 0x7f, *new_param));
+            controller->Axis(c_axis, GetAxis(0x0, 0x7f, *new_param), false);
             controller->Button(OrbisPadButtonDataOffset::R2, *new_param > 0x20);
             return;
         default:
             break;
         }
-        controller->Axis(c_axis, GetAxis(-0x80, 0x7f, *new_param * multiplier));
+        controller->Axis(c_axis, GetAxis(-0x80, 0x7f, *new_param * multiplier), false);
```
kalaposfos13: Unconditionally turning off axis smoothing will make KBM Bloodborne fans launch a crusade against you for regressing their game, so I suggest you either revert these to their original state, pure controller input won't trigger smoothing anyway (that only happens when a new axis value is more than 16 units away from the previous one, and virtually all controllers are capable of polling at resolutions much higher than that), if you really want to microoptimize this, just put the "skip to the end on small changes" logic into OnAxis too.

cuesta4: Thank you.

kalaposfos13 (approved):
I have no further observations about the code, although I have not yet checked if every game that might be affected by the PR still works or not in the latest version.

cuesta4:
No hurry, thanks for the patience

src/input/controller_state.cpp
```diff
@@ -0,0 +1,81 @@
+// SPDX-FileCopyrightText: Copyright 2024-2026 shadPS4 Emulator Project
```
StevenMiller123: New files should not have copyright attributed back to 2024.

StevenMiller123:
Latency in Kingdom Hearts titles is still fixed with the current version of this PR.

src/input/controller_state.cpp
cuesta4: Done.

PR #4509, opened 2026-06-01
Core: Only initialize logging once

StevenMiller123:
Somewhere during the mess of config swap and logger backend swap, we settled on initializing our logger multiple times to get around some annoyances with the new config code. While I think the expectation was that we'd retain all logs from prior inits, things would only behave as expected if users enabled the log_append config setting (which is impractical for most users).

This PR restores our old behavior, initializing loggers after we read the game id from the param.sfo (at which point we know what to name our log). Error and warning logs that could fire before logger initialization now use fmt::println instead, and I've removed the debug logs since there's not a good way to handle those properly (since these components need to initialize well before we even try to load a game).

While there's not a whole lot of benefit to this now that the trophy-related warnings all fire after logger's done initializing, this does prevent the creation of shad_log.txt for users with separated logging enabled. This also means that log append will now function properly for users with separated logging disabled, as before the first init would forcibly truncate the log.

### DIFF
--- a/src/common/key_manager.cpp
+++ b/src/common/key_manager.cpp
@@ -78,13 +78,12 @@ bool KeyManager::LoadFromFile() {
             SetDefaultKeys();
             TransferTrophyKey();
             SaveToFile();
-            LOG_DEBUG(KeyManager, "Created default key file: {}", keysPath.string());
             return true;
         }
 
         std::ifstream file(keysPath);
         if (!file.is_open()) {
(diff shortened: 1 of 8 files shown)


Niram7777:
I will review it, hopefully this evening...

Niram7777 (commented):
I will implement "shad_setup.log" this weekend and put back the LOG_ macros so it stays in some file

PR #4511, opened 2026-06-02
Implement a fragment shader emulating clip distance on NVIDIA

mikusp:


squidbus (commented):
General thoughts on the approach, we should probably organize this more like how we generate the TCS/TES shaders for rect/quad lists with an emitter in the shader recompiler backend, then set it at the Vulkan pipeline level same as those.

https://github.com/shadps4-emu/shadPS4/blob/main/src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp

https://github.com/shadps4-emu/shadPS4/blob/main/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp#L193-L204

georgemoralis:
it's getting hot here

squidbus (approved):

PR #4515, opened 2026-06-03
Refactor trophy extraction logic to improve error handling and maintin index mapping

w1naenator:
This fixes the Gravity Rush regression on my end. Other titles should also be tested.


StevenMiller123 (changes requested):
I already have an actual fix for similar issues in #4504.

w1naenator:
> I already have an actual fix for similar issues in #4504.

Ok, I jaust close mychack.

georgemoralis:
fixed with #4504

PR #4517, opened 2026-06-03
kernel: fix recursive read-lock deadlock on PthreadRwlock

junminlee2004:
Rewrote my fix in PR #4468 by amending SharedFirstMutex to allow timed-locking. The fix is now completely out of rwlock.cpp. I have tested and confirmed that the deadlock is fixed as in my previous PR

### DIFF
--- a/src/common/shared_first_mutex.h
+++ b/src/common/shared_first_mutex.h
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <chrono>
 #include <condition_variable>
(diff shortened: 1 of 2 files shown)


StevenMiller123 (approved):
Looks good to me, assuming no regressions are found.

junminlee2004:
oof, I pushed the version with the screwed up include directives, this one should build

georgemoralis:
looks good

PR #4521, opened 2026-06-05
fix: zlib-ng on GH runner using native instructions instead of the global config

xinitrcn1:
If the user oh so desires, `CMAKE_CXX_FLAGS="-mtune=native -march=naitve"` should satisfy this, otherwise this seems to cause issues with gh runners

### DIFF
--- a/externals/CMakeLists.txt
+++ b/externals/CMakeLists.txt
@@ -51,7 +51,8 @@ if (NOT TARGET ZLIB::ZLIB)
     set(ZLIB_ENABLE_TESTS OFF)
     set(WITH_GTEST OFF)
     set(WITH_NEW_STRATEGIES ON)
-    set(WITH_NATIVE_INSTRUCTIONS ON)
+    # Will mess with gh actions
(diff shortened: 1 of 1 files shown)


externals/CMakeLists.txt
```diff
@@ -51,7 +51,8 @@ if (NOT TARGET ZLIB::ZLIB)
     set(ZLIB_ENABLE_TESTS OFF)
     set(WITH_GTEST OFF)
     set(WITH_NEW_STRATEGIES ON)
-    set(WITH_NATIVE_INSTRUCTIONS ON)
+    # Will mess with gh actions
+    set(WITH_NATIVE_INSTRUCTIONS OFF)
```
squidbus, line 55: I don't think we need the comment, and technically we don't need to specify this at all because it's default `OFF` but I'm fine if you want to be explicit.

Also with this we should remove the extra compile flags from the actions, they're not needed.

squidbus:
I'll just follow up with a bit of clean up.

PR #4522, opened 2026-06-05
Fix avx512 for intel again

Niram7777:
-- 2 core AMD EPYC 9V74 80-Core Processor

shadps4-win64-sdl-2026-06-05-08df4e6  1  llvm-objdump -d shadPS4.exe | grep -e "vmovdqu64" -e "vpternlogq" | wc -l
0


squidbus:
I don't think this is needed with https://github.com/shadps4-emu/shadPS4/pull/4521

PR #4527, opened 2026-06-05
build: Clean up CPU instruction set handling a bit

squidbus:
* Remove setting `WITH_NATIVE_INSTRUCTIONS` for zlib entirely, the default is `OFF`. Similarly `WITH_RUNTIME_CPU_DETECTION` also defaults to `ON` as a result.
* Set `SDL_AVX512F` in CMake files for build consistency. Presumably in cases like Linux distros they'll be using SDL as a separate system lib anyway, this is just for our own submodule build.
* Remove AVX512 flags from CI, we don't need them there now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -171,7 +171,7 @@ jobs:
         key: ${{ env.cache-name }}-${{ hashFiles('**/CMakeLists.txt', 'cmake/**') }}
 
     - name: Configure CMake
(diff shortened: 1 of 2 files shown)


Niram7777 (changes requested):
you still need to deactivate AVX512, it is on by default for zlib-ng

> shadps4-win64-sdl-2026-06-05-6cf397f  1  llvm-objdump -d shadPS4.exe | grep -e "vmovdqu64" -e "vpternlogq" | wc -l
145

squidbus:
> you still need to deactivate AVX512, it is on by default for zlib-ng
> 
> > shadps4-win64-sdl-2026-06-05-6cf397f  1  llvm-objdump -d shadPS4.exe | grep -e "vmovdqu64" -e "vpternlogq" | wc -l
> > 145

Checking `objdump` is not relevant, because the default of zlib is to use runtime dispatch, i.e. it won't use AVX512 on systems without it.

Niram7777 (approved):

Niram7777:
- [x] waiting for test - before shadow merges it too fast :p

Hog185:
no regressions found on my end

PR #4528, opened 2026-06-06
Kernel.Pthreads: Restructure PthreadMutex struct

StevenMiller123:
As part of libc's mspace functions, when creating pthread mutexes for internal use, they modify the internal data of these mutexes to add the Private flag to m_flags. Previously, on Linux, this would modify internal data of our TimedMutex, and on Windows this would modify the m_yieldloops variable.

This PR reorganizes the struct to place m_flags at a accurate offset so these modifications work.

I would like to know if std::string is size 0x20 on Mac. It's size 0x20 on Windows and Linux, but I don't have a Mac, and I know their types tend to differ. If it is a different size, then I'll need to add filler to the PthreadMutex instead.

Credit to @liuk7071 for bringing this to my attention. As far as I've seen, this appears to be the only place and way libc does this.

### DIFF
--- a/src/core/libraries/kernel/threads/pthread.h
+++ b/src/core/libraries/kernel/threads/pthread.h
@@ -61,14 +61,15 @@ enum class PthreadMutexProt : u32 {
 };
 
 struct PthreadMutex {
-    std::string name;
-    PthreadMutexFlags m_flags;
-    int m_count;
-    TimedMutex m_lock;
     Pthread* m_owner;
+    int m_count;
     int m_spinloops;
     int m_yieldloops;
     PthreadMutexProt m_protocol;
+    u64 : 64;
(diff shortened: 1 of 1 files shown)


squidbus:
> I would like to know if std::string is size 0x20 on Mac. It's size 0x20 on Windows and Linux, but I don't have a Mac, and I know their types tend to differ. If it is a different size, then I'll need to add filler to the PthreadMutex instead.

As far as I can see it's 0x18 bytes. Honestly I would prefer not relying on C++ class types to be a specific size regardless, you can put the other fields first and if needed add some padding.

Also would recommend a `static_assert` for the offset of the field, with a comment explaining why.

StevenMiller123:
Made the suggested changes

squidbus (approved):

PR #4530, opened 2026-06-06
CMake log CPU info

Niram7777:
- [x] make @kalaposfos13 happy

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -170,6 +170,9 @@ jobs:
         append-timestamp: false
         key: ${{ env.cache-name }}-${{ hashFiles('**/CMakeLists.txt', 'cmake/**') }}
 
+    - name: Print CPU Info
+      run: bash -c "cat /proc/cpuinfo"
+
     - name: Configure CMake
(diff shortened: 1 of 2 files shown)


squidbus:
I think it would be more comprehensive/useful if we had a step as part of the CI actions instead that prints more detailed CPU info including supported extensions directly.

For Linux, it would be something like `cat /proc/cpuinfo`
For Windows, it seems from other projects you can do `bash -c "cat /proc/cpuinfo"` in the CI environment (or just set bash as the terminal for the actions step and do the same as Linux).
For macOS, `sysctl hw machdep.cpu` should be fine.

Windows example: https://github.com/bab2min/Kiwi/actions/runs/23979478850/job/69941527334

kalaposfos13 (commented):
VMs mess with the CPU names, so I'm not actually sure how useful this is for checking for AVX512.
```
-- 2 core 2794 MHz GenuineIntel Pentium III (0.18 micron) With 1 Or 2 MB On-Die L2 Cache 
```

squidbus (approved):
Output looks good

kalaposfos13 (approved):

PR #4531, opened 2026-06-06
Lib.Zlib: Use miniz instead of zlib-ng

StevenMiller123:
This might fix the random, Windows-specific crashes on boot in Rise of the Tomb Raider v1.00.
I'm not entirely confident, so I'm marking this as a draft until multiple testers verify.

Fix was suggested by @liuk7071, I did my own testing and it seems to address the issue, but since this is a random issue, I can't be certain.

### DIFF
--- a/src/core/libraries/zlib/zlib.cpp
+++ b/src/core/libraries/zlib/zlib.cpp
@@ -55,9 +55,9 @@ void ZlibTaskThread(const std::stop_token& stop) {
             task_queue.pop();
         }
 
-        uLongf decompressed_length = task.dst_length;
-        const auto ret = mz_uncompress(static_cast<Bytef*>(task.dst), &decompressed_length,
-                                       static_cast<const Bytef*>(task.src), task.src_length);
+        mz_ulong decompressed_length = task.dst_length;
+        const auto ret = mz_uncompress(static_cast<u8*>(task.dst), &decompressed_length,
+                                       static_cast<const u8*>(task.src), task.src_length);
 
         {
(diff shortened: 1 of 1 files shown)


StevenMiller123:
@bigol83 and @liuk7071 are both saying this fixes the Rise of the Tomb Raider crashes, so marking this as for review.

squidbus:
What is the context for this and how it's the solution?

Basically what is wrong with zlib-ng and can it be fixed? If there's an issue with it I'd rather just remove it entirely.

StevenMiller123:
I honestly don't know exactly what's wrong, the data zlib-ng produces for our inflate jobs doesn't match what miniz produces. May be random, as I've had situations where the game worked fine before too.

The output from zlib-ng causes crashes in Rise of the Tomb Raider, while the output from miniz works as expected.

squidbus:
If we're having unknown issues with zlib-ng I think we should just remove it altogether then, I don't see why we have both in the first place since they're both zlib-compatible.

StevenMiller123:
Not sure if this is the only use for zlib-ng, @georgemoralis was saying libpng uses it too.

squidbus:
Yes but if miniz is zlib-compatible it should work for that as well.

squidbus (approved):
I'll take a look at if we can remove zlib-ng later

georgemoralis:
png lib and the upcoming httplib are using zlib configurable in cmake . i am not sure it is so easy to detach zlib from them

PR #4537, opened 2026-06-07
Reset controller colours on emulator shutdown

### LINKED ISSUE #4536: [Feature Request]: Reset controller color on game exit
### Checklist

### Description

Currently, if you play a game that sets your controller to a special color, if you exit the game the controller will stay the same color, it would be nice if the controller could reset to the default color chosen in the GUI/user config instead.

### Reason

This is how it behaves on real hardware.

### Examples

_No response_

kalaposfos13:
Closes #4536


Niram7777:
why not just simply add ResetLightbarColors to quick_exit ?

src/common/assert.cpp
```diff
@@ -12,14 +13,19 @@
...
 #endif
 
+namespace Core {
+extern Emulator* g_emu;
+}
+
 void assert_fail_impl() {
-    Common::Log::Flush();
+    if (Core::g_emu) {
+        Core::g_emu->~Emulator();
```
Niram7777: if threads are still running and using Emulator objects, it will segfault

Niram7777 (changes requested):

kalaposfos13:
> why not just simply add ResetLightbarColors to quick_exit ?

Because this way, more shutdown things can be added in the future, and also, as you can see not all exits are from std::quick_exit.

src/common/assert.cpp
kalaposfos13: Good point, I guess I'll make a new function for this instead of using the destructor, although I believe this won't cause a segfault, as the emulator object won't be freed from memory from calling its destructor according to the C++ spec.

Niram7777: > I believe this won't cause a segfault

the WindowSDL will be destroyed, and it is used elsewhere

you should call your Shutdown now

fyi: I am planning to fix the shutdown at some point https://github.com/shadps4-emu/shadPS4/pull/4220

Niram7777 (changes requested):

Niram7777 (approved):

PR #4539, opened 2026-06-07
[Frontend] Polish fullscreen startup behavior

cuesta4:
### Problem

Starting a game with fullscreen enabled creates the window at the configured windowed size
and only then transitions it to fullscreen, letting the OS compositor animate the change
and briefly expose unpainted regions and the desktop. The cached window size is also never
refreshed after the transition, so the first swapchain and the ImGui display size start
from stale windowed values, stretching the splashscreen until the first resize event.

### Fix

- Set `SDL_PROP_WINDOW_CREATE_FULLSCREEN_BOOLEAN` at window creation when fullscreen is
  configured. The window is born fullscreen, so there is no transition to animate and
  nothing to hide. SDL sizes the window to the display internally and keeps the requested
  `window_width`/`window_height` as the windowed size, which it restores natively when
  leaving fullscreen — the config keeps defining the window size and no manual geometry
  bookkeeping is needed.
- Refresh the cached size with `SDL_GetWindowSizeInPixels()` after `SDL_SyncWindow()`, so
  the first swapchain and the splashscreen use the real drawable size. This also corrects
  the initial pixel size on HiDPI displays when starting windowed.
- Handle `SDL_EVENT_WINDOW_ENTER_FULLSCREEN` / `SDL_EVENT_WINDOW_LEAVE_FULLSCREEN` like
  `SDL_EVENT_WINDOW_RESIZED`, as suggested in review.
- Fix a dead error check: `SDL_GetDisplayForWindow()` returns `0` on failure
  (`SDL_DisplayID` is unsigned, so the `< 0` comparison was never true).

### DIFF
--- a/src/sdl_window.cpp
+++ b/src/sdl_window.cpp
@@ -119,6 +119,11 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameControllers* controller
     SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, height);
     SDL_SetNumberProperty(props, "flags", SDL_WINDOW_VULKAN);
     SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_RESIZABLE_BOOLEAN, true);
+    // Creating the window directly in fullscreen avoids a visible windowed -> fullscreen
+    // transition on startup. SDL sizes the window to the display and keeps the requested
+    // width/height as the windowed size to restore when leaving fullscreen.
+    SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_FULLSCREEN_BOOLEAN,
+                           EmulatorSettings.IsFullScreen());
     window = SDL_CreateWindowWithProperties(props);
     SDL_DestroyProperties(props);
     if (window == nullptr) {
@@ -129,7 +134,7 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameControllers* controller
 
     bool error = false;
     const SDL_DisplayID displayIndex = SDL_GetDisplayForWindow(window);
-    if (displayIndex < 0) {
+    if (displayIndex == 0) {
         LOG_ERROR(Frontend, "Error getting display index: {}", SDL_GetError());
         error = true;
     }
@@ -144,6 +149,9 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameControllers* controller
     }
     SDL_SetWindowFullscreen(window, EmulatorSettings.IsFullScreen());
     SDL_SyncWindow(window);
+    // The window geometry is only final once the fullscreen transition has settled; refresh
+    // the cached size so the first swapchain and the splashscreen use the real drawable size.
+    SDL_GetWindowSizeInPixels(window, &width, &height);
 
     SDL_InitSubSystem(SDL_INIT_GAMEPAD);
 
@@ -232,6 +240,8 @@ void WindowSDL::WaitEvent() {
     case SDL_EVENT_WINDOW_RESIZED:
     case SDL_EVENT_WINDOW_MAXIMIZED:
     case SDL_EVENT_WINDOW_RESTORED:
+    case SDL_EVENT_WINDOW_ENTER_FULLSCREEN:
+    case SDL_EVENT_WINDOW_LEAVE_FULLSCREEN:
         OnResize();
         break;
     case SDL_EVENT_WINDOW_MINIMIZED:


Niram7777 (changes requested):
this opacity hack look likes a workaround for an issue somewhere else.
but maybe you can try to open a issue to see if other people can provide details.
from what I found, the SDL setup we currently have is quite standard

georgemoralis:
@Niram7777 since you have already checked this care to review it again?

src/imgui/renderer/imgui_core.cpp
```diff
@@ -110,6 +110,7 @@ void Initialize(const ::Vulkan::Instance& instance, const Frontend::WindowSDL& w
 
     ::Core::Devtools::Layer::SetupSettings();
     Sdl::Init(window.GetSDLWindow());
+    Sdl::OnResize();
```
Niram7777: the window should already have the right size, so I dont see this as needed

src/video_core/renderer_vulkan/vk_presenter.cpp
```diff
@@ -1086,12 +1086,17 @@ void Presenter::Present(Frame* frame, bool is_reusing_frame) {
...
     {
         std::scoped_lock submit_lock{Scheduler::submit_mutex};
-        if (!swapchain.Present()) {
+        presented = swapchain.Present();
+        if (!presented) {
             swapchain.Recreate(window.GetWidth(), window.GetHeight());
         }
     }
+    if (presented) {
+        window.RequestStartupWindowReveal();
```
Niram7777: Hiding the window is a bad idea:
- some SDL functions expect the window to be visible
- as a user experience, if something is loading or failing, its better if we display something (even if now its just a black frame)

src/sdl_window.cpp
```diff
@@ -100,15 +102,38 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameControllers* controller
...
+        SDL_Rect display_bounds{};
+        const SDL_DisplayID primary_display = SDL_GetPrimaryDisplay();
+        if (primary_display != 0 && SDL_GetDisplayBounds(primary_display, &display_bounds)) {
+            // A hidden fullscreen window may retain its windowed client size until it is mapped.
+            // Seed the native window with the monitor geometry before creating the Vulkan surface
+            // so the first swapchain is already fullscreen-sized.
+            create_x = display_bounds.x;
+            create_y = display_bounds.y;
+            create_width = display_bounds.w;
+            create_height = display_bounds.h;
```
Niram7777: window_width/window_height config should define the window size.
eventually if it is set to 0, we could say "take the max display size" (but we shouldnt force it all the time)

src/sdl_window.cpp
```diff
@@ -100,15 +102,38 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameControllers* controller
...
     SDL_InitSubSystem(SDL_INIT_AUDIO);
 
+    const bool start_fullscreen = EmulatorSettings.IsFullScreen();
+    s32 create_x = SDL_WINDOWPOS_CENTERED;
+    s32 create_y = SDL_WINDOWPOS_CENTERED;
+    s32 create_width = width;
+    s32 create_height = height;
+    if (start_fullscreen) {
+        SDL_Rect display_bounds{};
+        const SDL_DisplayID primary_display = SDL_GetPrimaryDisplay();
```
Niram7777: the display to use could be choose regardless if it is fullscreen or not

src/sdl_window.h
```diff
@@ -76,20 +77,28 @@ class WindowSDL {
...
     void OnResize();
+    void RevealStartupWindow();
     void OnKeyboardMouseInput(const SDL_Event* event);
     void OnGamepadEvent(const SDL_Event* event);
 
 private:
     s32 width;
     s32 height;
+    s32 windowed_width;
+    s32 windowed_height;
```
Niram7777: those are the sames

src/sdl_window.cpp
```diff
@@ -100,15 +102,38 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameControllers* controller
...
-    SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_WIDTH_NUMBER, width);
-    SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, height);
+    SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_X_NUMBER, create_x);
+    SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_Y_NUMBER, create_y);
+    SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_WIDTH_NUMBER, create_width);
+    SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, create_height);
     SDL_SetNumberProperty(props, "flags", SDL_WINDOW_VULKAN);
     SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_RESIZABLE_BOOLEAN, true);
+    SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_FULLSCREEN_BOOLEAN, start_fullscreen);
+    SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_HIDDEN_BOOLEAN, start_fullscreen);
```
Niram7777: it goes with my "Hiding the window" comment

src/sdl_window.cpp
```diff
@@ -132,8 +157,18 @@ WindowSDL::WindowSDL(s32 width_, s32 height_, Input::GameControllers* controller
...
+    if (start_fullscreen) {
+        const bool defer_reveal = Libraries::SystemService::IsSplashVisible() &&
+                                  !Common::ElfInfo::Instance().GetSplashPath().empty();
+        startup_window_reveal_needed.store(defer_reveal, std::memory_order_release);
+        if (!defer_reveal) {
+            SDL_ShowWindow(window);
+            SDL_SyncWindow(window);
+            SDL_GetWindowSizeInPixels(window, &width, &height);
+        }
+    }
```
Niram7777: it goes with my "Hiding the window" comment

src/sdl_window.cpp
```diff
@@ -308,6 +363,23 @@ void WindowSDL::InitTimers() {
...
+            false)) {
+        startup_window_reveal_needed.store(true, std::memory_order_release);
+    }
+}
+
+void WindowSDL::RevealStartupWindow() {
+    SDL_ShowWindow(window);
+    SDL_SyncWindow(window);
+    OnResize();
+}
```
Niram7777: it goes with my "Hiding the window" comment

src/sdl_window.cpp
```diff
@@ -224,10 +260,29 @@ void WindowSDL::WaitEvent() {
...
+                SDL_SetWindowPosition(window, windowed_x, windowed_y);
+            } else {
+                SDL_SetWindowPosition(window, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
+            }
+            SDL_SyncWindow(window);
         } else {
-            SDL_SetWindowFullscreen(window, SDL_WINDOW_FULLSCREEN);
+            SDL_GetWindowPosition(window, &windowed_x, &windowed_y);
+            SDL_GetWindowSize(window, &windowed_width, &windowed_height);
+            has_windowed_position = true;
```
Niram7777: SDL_EVENT_WINDOW_ENTER_FULLSCREEN
SDL_EVENT_WINDOW_LEAVE_FULLSCREEN
should probably do like
SDL_EVENT_WINDOW_RESIZED
and call OnResize

Niram7777 (changes requested):
the rootcause is still unclear and the PR is adding too much complexity in critical code.
I have the impression the issue is outside our app, or that if we needed a fix, it would be simpler.
I would like more details on what is the issues, eventually reproduce it

Niram7777:
LGTM, I will test tomorrow morning on Linux/Windows

src/sdl_window.cpp
```diff
@@ -232,6 +240,8 @@ void WindowSDL::WaitEvent() {
     case SDL_EVENT_WINDOW_RESIZED:
     case SDL_EVENT_WINDOW_MAXIMIZED:
     case SDL_EVENT_WINDOW_RESTORED:
+    case SDL_EVENT_WINDOW_ENTER_FULLSCREEN:
+    case SDL_EVENT_WINDOW_LEAVE_FULLSCREEN:
         OnResize();
         break;
```
Niram7777, line 246: I thought for a second that when exiting fullscreen it would keep window size at the display size but actually from my tests it is able to go back down to the configuration size

Niram7777 (approved):
Tested and it is visually better to start directly in fullscreen, thank you

cuesta4:
@georgemoralis please merge this?

georgemoralis:
ok after ci finish

PR #4541, opened 2026-06-08
Refactor sRGB format handling for BC formats

Hog185:


src/video_core/renderer_vulkan/liverpool_to_vk.cpp
```diff
@@ -756,6 +769,7 @@ static auto surface_format_table = []() constexpr {
 }();
 
 vk::Format SurfaceFormat(AmdGpu::DataFormat data_format, AmdGpu::NumberFormat num_format) {
+    num_format = AmdGpu::RemapNumberFormat(num_format, data_format);
```
squidbus: This seems suspect, format should already be remapped before we get here. If there's somewhere that's not happening it may need to be addressed there.

Hog185: the call sites aren't doing, if you want i can do it in image-info, but it certainly aint redundant

squidbus: My point is that wherever the formats come from is intended to be doing it already. For example if you look at `ImageInfo::ImageInfo(image, desc)`, it does:

```
pixel_format = LiverpoolToVK::SurfaceFormat(image.GetDataFmt(), image.GetNumberFmt());
```

..and looking at those in turn, you will see:

```
    DataFormat GetDataFmt() const noexcept {
        return RemapDataFormat(DataFormat(data_format));
    }

    NumberFormat GetNumberFmt() const noexcept {
        return RemapNumberFormat(NumberFormat(num_format), DataFormat(data_format));
    }
```

Hog185: ah i see, i will remove and retest

src/video_core/renderer_vulkan/liverpool_to_vk.cpp
```diff
@@ -35,6 +35,19 @@ vk::StencilOp StencilOp(AmdGpu::StencilFunc op) {
         return vk::StencilOp::eDecrementAndWrap;
     case AmdGpu::StencilFunc::ReplaceOp:
         return vk::StencilOp::eReplace;
+    case AmdGpu::StencilFunc::Ones:
+        // No Vulkan equivalent; eReplace with reference value 0xFF is the closest approximation.
+        return vk::StencilOp::eReplace;
```
squidbus, line 41: You describe `eReplace with reference value 0xFF` here, but I don't see anything around ensuring that reference value.

Also would prefer a warning here similar to below.

Hog185: yeh dw about that comment ill remove it

squidbus (approved):
Not super happy with these stencil op replacements, but as long as there's a warning log for visibility, I guess it's better than crashing.

Hog185:
@squidbus as you have said,better than nothing

PR #4542, opened 2026-06-08
Fix stale guest memory when reusing GPU-written linear images through overlapping aliases

cuesta4:
### Problem

Some games (namely GOW3) reuse the same guest memory range through different image representations. A linear image may be written by the GPU and later reinterpreted as a tiled or compressed texture.

With **Enable Readback Linear Images** enabled, shadPS4 already scheduled GPU-to-CPU readbacks for linear images. However, those readbacks were asynchronous. This left a timing window where an overlapping image could be recreated from guest memory before the pending linear-image writeback had completed.

In that case, the new texture was uploaded from stale guest-memory data even though the most recent contents already existed on the GPU. This could result in progressive texture corruption when the aliased representation was used later.

### Fix

Keep the existing linear-image readback path asynchronous in the common case, but synchronize it on demand when an overlapping alias actually depends on the updated guest-memory contents.

The updated flow is:

GPU writes a linear image
- schedule the existing asynchronous readback

Later, an overlapping image needs to be reconstructed from guest memory
- detect pending linear-image readbacks for the overlapping range
- wait once for the relevant batch
- commit the completed data to guest memory
- reconstruct the alias from coherent data

This preserves the existing readback infrastructure and only adds the missing dependency point before alias reconstruction. The existing path performs preventive asynchronous readbacks. This change adds the required demand-driven synchronization when another image representation needs those bytes immediately. This avoids the broader synchronization cost of forcing every linear-image readback to complete immediately while still guaranteeing correctness for dependent alias reuse.

Before:
image

After:
image

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -86,17 +86,25 @@ ImageId TextureCache::GetNullImage(const vk::Format format) {
 }
 
 void TextureCache::ProcessDownloadImages() {
+    std::scoped_lock lock{mutex};
+    boost::container::small_vector<PendingImageDownload, 8> downloads;
+    downloads.reserve(download_images.size());
     for (const ImageId image_id : download_images) {
-        DownloadImageMemory(image_id, true);
+        if (auto download = ScheduleImageDownload(image_id)) {
+            downloads.push_back(*download);
+        }
     }
     download_images.clear();
+    CompleteImageDownloads(downloads);
 }
 
-void TextureCache::DownloadImageMemory(ImageId image_id, bool sync) {
+std::optional<TextureCache::PendingImageDownload> TextureCache::ScheduleImageDownload(
+    ImageId image_id) {
     Image& image = slot_images[image_id];
-    if (False(image.flags & ImageFlagBits::GpuModified)) {
-        return;
+    if (!image.SafeToDownload() || image.info.props.is_tiled || image.info.guest_address == 0) {
+        return {};
     }
+
     auto& download_buffer = buffer_cache.GetUtilityBuffer(MemoryUsage::Download);
     const u32 download_size = image.info.pitch * image.info.size.height * image.info.size.depth *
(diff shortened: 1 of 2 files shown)


EmulationChannell:
Captura de tela 2026-06-08 151314
Captura de tela 2026-06-08 151402
Captura de tela 2026-06-08 152011

[shad_log.txt](https://github.com/user-attachments/files/28721689/shad_log.txt)

StevenMiller123:
@EmulationChannell This PR only works with the readback linear images setting enabled. Your log indicates you left the setting disabled for your test.

EmulationChannell:
Captura de tela 2026-06-08 152547
Captura de tela 2026-06-08 152613
readback linear images setting enabled my test @StevenMiller123 
Perform a test on two option models here, same thing with the game GOW3 REMASTERD

cuesta4:
This only addresses the texture corruptions. The blooming problem is another thing entirely.

EmulationChannell:
> This only addresses the texture corruptions. The blooming problem is another thing entirely.

Yes, I noticed that with all the textures broken. I tested God of War 3 Remastered with both options available in the emulator settings.

Xcedf:
For the first time problmatic level loads with correct textures, but after it textures will get corruped anyway, and Readback Linear Images enabled.

TdlifePeng:
> This only addresses the texture corruptions. The blooming problem is another thing entirely.
Screenshot 2026-06-09 134434

[fix auto-exposure chain by copying RT data to aliased texture views](https://github.com/TdlifePeng/shadPS4/commit/a0469fa07a20ce71205a19a11f269f4de6590d3a)

EmulationChannell:
> > This only addresses the texture corruptions. The blooming problem is another thing entirely.
> 
> Screenshot 2026-06-09 134434
> [fix auto-exposure chain by copying RT data to aliased texture views](https://github.com/TdlifePeng/shadPS4/commit/a0469fa07a20ce71205a19a11f269f4de6590d3a)

yes fixed thanks!

PR #4549, opened 2026-06-10
NP Part 1: NpSignaling

metr1k-91:
PR Split 1 of X, adds the NpSignaling Library and some of the dependencies to other libraries (Namely NpCommon.  Any stubs are highlighted by the stubs files and will be resolved upon other PRs.  (Dependencies are to shadnet and further PR's to Matching2)


src/core/libraries/np/np_common.cpp
```diff
@@ -1,33 +1,115 @@
...
+    if (sce_rc == ORBIS_KERNEL_ERROR_EBUSY) {
+        return ORBIS_NP_LW_MUTEX_ERROR_BUSY;
+    }
+    return static_cast<u32>(sce_rc);
+}
+
+void* PS4_SYSV_ABI NpCalloutThreadMain(void* arg) {
+    auto* ctx = static_cast<OrbisNpCalloutContext*>(arg);
+    const auto mutex = &ctx->mutex;
+    (void)sceNpMutexLock(mutex);
```
StevenMiller123: We don't typically do unnecessary type casts like this.

src/core/libraries/np/np_common.cpp
```diff
@@ -90,13 +169,454 @@ s32 PS4_SYSV_ABI sceNpCmpOnlineId(OrbisNpOnlineId* online_id1, OrbisNpOnlineId*
...
 }
 
+using NpMutexStorage = std::recursive_mutex;
+using NpCondStorage = std::condition_variable_any;
+struct NpThreadStorage {
+    std::thread t;
+};
+
+u32 PS4_SYSV_ABI sceNpMutexLock(void* mutex) {
+    // scepthread should be used here
```
StevenMiller123: There's nothing wrong with exporting our HLE'd pthread functions and typedefs to use here. I did something similar when working on LibcInternal file HLE (and you seem to be doing this for the NpSignalling code anyway).

src/core/libraries/np/np_common.cpp
```diff
@@ -90,13 +169,454 @@ s32 PS4_SYSV_ABI sceNpCmpOnlineId(OrbisNpOnlineId* online_id1, OrbisNpOnlineId*
...
+    } else {
+        c->wait_for(lock, std::chrono::microseconds(usec));
+    }
+    lock.release();
+    return 0;
+}
+
+s32 PS4_SYSV_ABI sceNpCreateThread(void* thread, NpThreadEntry start_routine, void* arg,
+                                   s32 priority, u64 stack_size, u64 affinity_mask,
+                                   void* thread_name) {
```
StevenMiller123: Shouldn't name be a const char*?

src/core/libraries/np/np_common.cpp
```diff
@@ -90,13 +169,454 @@ s32 PS4_SYSV_ABI sceNpCmpOnlineId(OrbisNpOnlineId* online_id1, OrbisNpOnlineId*
...
+    Libraries::Kernel::OrbisKernelTimespec ts{};
+    const s32 rc =
+        Libraries::Kernel::sceKernelClockGettime(Libraries::Kernel::ORBIS_CLOCK_MONOTONIC, &ts);
+    if (rc == ORBIS_OK) {
+        *usec = ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
+    }
+    return rc;
+}
+
+s32 PS4_SYSV_ABI sceNpGetPlatformType(const void* np_id) {
```
StevenMiller123: Why are you using void* as a parameter instead of OrbisNpId*

src/core/libraries/np/np_error.h
```diff
@@ -5,11 +5,12 @@
 
 #include "core/libraries/error_codes.h"
 
-// Base NP errors (0x80550001 - 0x8055001F)
+// Base NP errors (0x80550001 � 0x8055001F)
```
StevenMiller123: Is there any reason you reverted the comment change here?

src/core/libraries/np/np_signaling.cpp
```diff
@@ -0,0 +1,890 @@
...
+        SignalingMutexGuard lock;
+        g_contexts.clear();
+        g_connections.clear();
+        g_npid_to_conn.clear();
+        g_peer_netinfo_results.clear();
+    }
+
+    DestroySignalingMutex();
+
+    LOG_INFO(Lib_NpSignaling, "sceNpSignalingTerminate: cleared all state");
```
StevenMiller123: No point in putting the function name here, our log macros already place it in the message.

src/core/libraries/np/np_signaling_helpers.cpp
```diff
@@ -0,0 +1,535 @@
...
+s32 InitSignalingHeap(s64 pool_size) {
+    LOG_DEBUG(Lib_NpSignaling, "Helpers::InitSignalingHeap pool_size={}", pool_size);
+
+    if (pool_size == 0 || g_signaling_heap_initialized) {
+        return ORBIS_NP_SIGNALING_INTERNAL_ERROR_ALLOCATOR;
+    }
+
+    const u64 alignment_mask = 0x3fff;
+    const u64 remainder = static_cast<u64>(pool_size) & alignment_mask;
+    const u64 aligned_size = remainder == 0 ? static_cast<u64>(pool_size)
```
StevenMiller123: Should use one of our alignment helpers, this is just an align up to the nearest page size.

StevenMiller123 (changes requested):
Comment list would be endless if I commented on every instance of these nitpicks. Aside from those, this looks relatively decent.

metr1k-91:
I'll get this stuff sorted.  The line comment removal readd was probably due to copy paste on my end.  Trying to split this off from the rest of my work has not been fun.  The rest I'll get fixed.  On the impl itself do you guys have specific guidelines on the way files are broken down? (ie state, helpers, etc.) That good how it's split or do you want it refactored in any way?

Thanks

StevenMiller123:
I think you've split things up fine, we don't have strict requirements in that regard.

src/core/libraries/np/np_signaling.cpp
metr1k-91: I'm fairly certain these are all handled now.

src/core/libraries/np/np_signaling_helpers.cpp
metr1k-91: I went ahead and replaced with alignup

src/core/libraries/np/np_error.h
metr1k-91: I was working off an older version of the error file specifically.  I've corrected these.

src/core/libraries/np/np_common.cpp
metr1k-91: This is good now.

src/core/libraries/np/np_common.cpp
metr1k-91: Corrected this.

src/core/libraries/np/np_common.cpp
metr1k-91: Thanks.  I had already done this for the most part but cond* related work I didn't find actually publically declared anywhere.  So I pulled back to what I actually have in my fork and readded everything.  thread.h is now included with those declarations and we should be good.

src/core/libraries/np/np_common.cpp
metr1k-91: I've typed this appropriately.

src/core/libraries/network/netctl.cpp
```diff
@@ -258,8 +258,30 @@ int PS4_SYSV_ABI sceNetCtlGetInfoV6IpcInt() {
...
+        nat_info->mapped_addr = ext_ip;
+    } else {
+        nat_info->stun_status = 0;
+        nat_info->mapped_addr = inet_addr("127.0.0.1");
+        if (netinfo->RetrieveIp()) {
+            nat_info->mapped_addr = inet_addr(netinfo->GetIp().c_str());
+        }
+    }
+
+    LOG_DEBUG(Lib_NetCtl, "GetNatInfo: stun_status={} nat_type={} mapped_addr={:#x}",
```
StevenMiller123: Missed this log

src/core/libraries/np/np_common.cpp
StevenMiller123: Nothing was changed here, or in any other of these explicit casts to void either. Did you forget to push something?

To be clear, what I'm meaning is that you don't need to cast the result of sceNpMutexLock (or anything similar) to void just because the result is ignored. Just don't assign the result to anything.

src/core/libraries/np/np_common.cpp
```diff
@@ -90,13 +165,446 @@ s32 PS4_SYSV_ABI sceNpCmpOnlineId(OrbisNpOnlineId* online_id1, OrbisNpOnlineId*
...
+    Libraries::Kernel::OrbisKernelTimespec ts{};
+    const s32 rc =
+        Libraries::Kernel::sceKernelClockGettime(Libraries::Kernel::ORBIS_CLOCK_MONOTONIC, &ts);
+    if (rc == ORBIS_OK) {
+        *usec = ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
+    }
+    return rc;
+}
+
+s32 PS4_SYSV_ABI sceNpGetPlatformType(const OrbisNpId* npid) {
```
StevenMiller123, line 337: While you fixed the parameter type here, there's several other places you're marking parameters as a void* then casting to the appropriate type. Please fix all of these.

src/core/libraries/np/np_common.cpp
```diff
@@ -90,13 +165,446 @@ s32 PS4_SYSV_ABI sceNpCmpOnlineId(OrbisNpOnlineId* online_id1, OrbisNpOnlineId*
...
+    return sceNpMutexUnlock(mutex);
+}
+
+u32 PS4_SYSV_ABI sceNpLwMutexInit(void* mutex, void* mutex_name, u64 flags) {
+    LOG_DEBUG(Lib_NpCommon, "mutex={:p} name={:p} flags={:#x}", fmt::ptr(mutex),
+              fmt::ptr(mutex_name), flags);
+    return sceNpMutexInit(mutex, mutex_name, flags);
+}
+
+void PS4_SYSV_ABI sceNpLwMutexDestroy(void* mutex) {
```
StevenMiller123: Pretty sure this shouldn't be a void return.

src/core/libraries/np/np_common.cpp
```diff
@@ -90,13 +165,446 @@ s32 PS4_SYSV_ABI sceNpCmpOnlineId(OrbisNpOnlineId* online_id1, OrbisNpOnlineId*
...
+        }
+        if (rc == 0) {
+            rc = Libraries::Kernel::scePthreadMutexInit(pthread_mutex, &attr, pthread_name);
+        }
+        Libraries::Kernel::posix_pthread_mutexattr_destroy(&attr);
+    }
+    return static_cast<u32>(NormalizeNpCommonResult(rc));
+}
+
+void PS4_SYSV_ABI sceNpMutexDestroy(void* mutex) {
```
StevenMiller123: Pretty sure this shouldn't be a void return.

src/core/libraries/np/np_common.cpp
```diff
@@ -90,13 +165,446 @@ s32 PS4_SYSV_ABI sceNpCmpOnlineId(OrbisNpOnlineId* online_id1, OrbisNpOnlineId*
...
+    Libraries::Kernel::PthreadCondAttrT attr = nullptr;
+    int rc = Libraries::Kernel::posix_pthread_condattr_init(&attr);
+    if (rc == 0) {
+        rc = Libraries::Kernel::scePthreadCondInit(pthread_cond, &attr, cond_name_cstr);
+        Libraries::Kernel::posix_pthread_condattr_destroy(&attr);
+    }
+    return NormalizeNpCommonResult(rc);
+}
+
+void PS4_SYSV_ABI sceNpCondDestroy(void* cond) {
```
StevenMiller123: Shouldn't be void.

src/core/libraries/np/np_common.cpp
```diff
@@ -90,13 +165,446 @@ s32 PS4_SYSV_ABI sceNpCmpOnlineId(OrbisNpOnlineId* online_id1, OrbisNpOnlineId*
...
+    auto* pthread_cond = static_cast<Libraries::Kernel::PthreadCondT*>(cond);
+    LOG_DEBUG(Lib_NpCommon, "cond={:p}", fmt::ptr(cond));
+    (void)Libraries::Kernel::posix_pthread_cond_destroy(pthread_cond);
+}
+
+void PS4_SYSV_ABI sceNpCondSignal(void* cond) {
```
StevenMiller123: Shouldn't be void.

src/core/libraries/np/np_signaling_state.cpp
```diff
@@ -0,0 +1,1364 @@
...
+        }
+
+        u32 peer_addr = 0;
+        u16 peer_port = 0;
+        const bool resolved = Stubs::ResolvePeer(act.peer_online_id, &peer_addr, &peer_port) &&
+                              peer_addr != 0 && peer_port != 0;
+
+        if (!resolved) {
+            LOG_WARNING(Lib_NpSignaling,
+                        "ProcessPendingActivations: peer '{}' endpoint unresolved; connection {} "
```
StevenMiller123: Don't need to log the function name here.

src/core/libraries/np/np_signaling_state.cpp
```diff
@@ -0,0 +1,1364 @@
...
+    if (!Stubs::TransportIsReady()) {
+        return;
+    }
+
+    const u32 server_addr = Stubs::MmServerAddr();
+    const u16 server_udp = Stubs::MmServerUdpPort();
+
+    if (server_addr == 0 || server_udp == 0) {
+        LOG_WARNING(Lib_NpSignaling,
+                    "SendStunPing: ctxId={} skipped (server_addr={:#x} udp_port={})", ctx_id,
```
StevenMiller123: Don't need to log the function name here.

StevenMiller123 (changes requested):
Most of my comments weren't fully addressed, so I've made my comments more clear. If you need more clarification, let me know.

metr1k-91:
I believe I've nailed a majority if not all of these now.  Let me know if I'm missing anything else.

StevenMiller123 (approved):
Looks good to me

StevenMiller123:
@georgemoralis Would be nice to have you review, since you've looked at a handful of NP libs yourself.

PR #4553, opened 2026-06-11
texture_cache: Improve handling of mips of compressed views

squidbus:
* Update `ImageInfo::IsCompatible` to check two-way format compatibility instead of exact pixel format equality.
* Update mip/slice checks and mip copy to check block dimensions instead of plain image dimensions.
* Restore storage image usage to compressed formats. This is needed if uncompressed views are used for storage, which is supported, and should not cause errors with extended usage create flag set.

Together these allow the texture cache to identify uncompressed mips that need to be copied to and replaced by a larger compressed texture. Previously, only the first mip would be copied with equal address, and subsequent mips would be skipped for copy in the left overlap case. Typically when compressed images are also used uncompressed, the uncompressed view happens one subresource at a time, which then all need to be coalesced into the larger compressed image.

### DIFF
--- a/src/video_core/texture_cache/image.cpp
+++ b/src/video_core/texture_cache/image.cpp
@@ -35,6 +35,11 @@ static vk::ImageUsageFlags ImageUsageFlags(const Vulkan::Instance* instance,
             // flag is also used.
             usage |= vk::ImageUsageFlagBits::eStorage;
         }
+    } else {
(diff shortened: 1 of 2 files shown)


squidbus:
Note that at least in the newer Sonic games I was testing this in, it improves the situation with textures but does not fully make things right, I assume there's still cases where the data can get lost.

Hog185:
no regressions nor impovements, LGTM

PR #4556, opened 2026-06-11
UPNP Support

metr1k-91:
Adds UPNP (currently not wired to anything) for NAT support utilizing the miniupnpc library.

Adds Settings:
enable_upnp
signaling_addr
signaling_port

If upnp is disabled, signaling will fall back to defined signaling_addr and signaling_port for signaling / matching2 activities, for those that have port forwarding and not upnp.

### DIFF
--- a/src/core/emulator_settings.h
+++ b/src/core/emulator_settings.h
@@ -112,7 +112,7 @@ struct OverrideItem {
 };
 
 template <typename Struct, typename T>
-inline OverrideItem make_override(const char* key, Setting<T> Struct::* member) {
+inline OverrideItem make_override(const char* key, Setting<T> Struct::*member) {
     return OverrideItem{
         key,
         [member, key](void* base, const nlohmann::json& entry, std::vector<std::string>& changed) {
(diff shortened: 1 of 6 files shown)


src/core/emulator_settings.h
```diff
@@ -202,7 +205,9 @@ struct GeneralSettings {
             make_override<GeneralSettings>("trophy_notification_side",
                                            &GeneralSettings::trophy_notification_side),
             make_override<GeneralSettings>("connected_to_network",
```
georgemoralis, line 207: enable_upnp should't be overrideable?

metr1k-91: I left it as a global rather than a per game override as if you're using it for one game I don't see the reason why you wouldn't use it for all of them.

Happy to make that change.

georgemoralis: it will go to experimental settings (like the other ones so, it should be ovverideable to be configured per game only

metr1k-91: Should be all set now.

PR #4557, opened 2026-06-11
texture_cache: expand image on larger guest_size

Hog185:
title is self explanitory


squidbus:
This has the same problem I've noted before, there has to be a reason *why* the guest size is bigger, not just that it is bigger.

PR #4558, opened 2026-06-11
Add native sampling for color cubemaps

Kravickas:
Fixes hard cubemap edge on the shoulder piece.
Main ----------------------------------------PR
ctr

Fixes visible seams in shadows.
Main ---------------------------------------------------------------------PR
bb

### DIFF
--- a/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
+++ b/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
@@ -924,7 +924,7 @@ Id ImageType(EmitContext& ctx, const ImageResource& desc, Id sampled_type) {
     const auto format = desc.is_atomic ? GetFormat(image) : spv::ImageFormat::Unknown;
     const auto type = image.GetViewType(desc.is_array);
     const u32 sampled = desc.is_written ? 2 : 1;
-    if (image.IsCube() && !desc.is_written && !desc.is_storage) {
+    if (image.IsCube() && !desc.is_written && !desc.is_storage && !desc.is_depth) {
         ctx.AddCapability(spv::Capability::SampledCubeArray);
         return ctx.TypeImage(sampled_type, spv::Dim::Cube, false, true, false, sampled, format);
     }
--- a/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
+++ b/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
@@ -924,6 +924,10 @@ Id ImageType(EmitContext& ctx, const ImageResource& desc, Id sampled_type) {
     const auto format = desc.is_atomic ? GetFormat(image) : spv::ImageFormat::Unknown;
     const auto type = image.GetViewType(desc.is_array);
     const u32 sampled = desc.is_written ? 2 : 1;
+    if (image.IsCube() && !desc.is_written && !desc.is_storage) {
+        ctx.AddCapability(spv::Capability::SampledCubeArray);
+        return ctx.TypeImage(sampled_type, spv::Dim::Cube, false, true, false, sampled, format);
+    }
     switch (type) {
     case AmdGpu::ImageType::Color1D:
         return ctx.TypeImage(sampled_type, spv::Dim::Dim1D, false, false, false, sampled, format);
--- a/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
+++ b/src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
@@ -265,6 +265,7 @@ public:
         auto& image = image_resources[index];
         image.is_atomic |= desc.is_atomic;
         image.is_written |= desc.is_written;
+        image.is_storage |= desc.is_storage;
         return index;
     }
 
@@ -556,6 +557,7 @@ void PatchImageSharp(IR::Block& block, IR::Inst& inst, Info& info, Descriptors& 
         .is_atomic = is_atomic,
         .is_array = bool(inst_info.is_array),
         .is_written = is_written,
+        .is_storage = is_storage,
         .is_r128 = bool(inst_info.is_r128),
     };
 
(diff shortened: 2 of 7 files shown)


Hog185:
how do you even notice this in gameplay 😭😭,

StevenMiller123:
To be fair, the BB case was rather noticeable 😅

squidbus:
The reason it was changed to 2D array is that some games will give you cube maps or use them in ways that are incompatible with Vulkan's restrictions on cube maps, this is likely to cause problems.

brad0demx:
image
image

Seems to fix lines in the skybox in Slime Rancher

squidbus:
Looking more closely, I guess if it's limited to sampled cubes that might be fine?

However, Bloodborne is giving validation errors:
```
[Render.Vulkan] <Error> (shadPS4:GpuCommandProcessor) vk_platform.cpp:57 DebugUtilsCallback: VUID-VkImageViewCreateInfo-image-01003: vkCreateImageView(): pCreateInfo->viewType (VK_IMAGE_VIEW_TYPE_CUBE_ARRAY) is not compatible with image type VK_IMAGE_TYPE_2D.
The Vulkan spec states: If image was not created with VK_IMAGE_CREATE_CUBE_COMPATIBLE_BIT then viewType must not be VK_IMAGE_VIEW_TYPE_CUBE or VK_IMAGE_VIEW_TYPE_CUBE_ARRAY (https://vulkan.lunarg.com/doc/view/1.4.350.0/mac/antora/spec/latest/chapters/resources.html#VUID-VkImageViewCreateInfo-image-01003)
```

I'm not sure you can rely on cube check for the base image's compatible bit as the game may view it as a non-cube first, you may just need to do some heuristic like if it's 2D, has equal 2D dimensions, has >= 6 layers, and sample count equals 1 (see valid usage requirements in https://docs.vulkan.org/refpages/latest/refpages/source/VkImageCreateInfo.html).

junminlee2004:
@squidbus And how would those issues usually manifest?

squidbus:
> And how would those issues usually manifest?

By playing the game? These log messages show up with Vulkan validation enabled.

junminlee2004:
My bad, you commented 2 minutes before I did so I didn't see it pop up. I was referring to the comment you made 4 hours ago

Xcedf:
Fixes long standing cubemap reflection issue in Infamous
Screens from local builds
before:
image
after:
image

src/video_core/texture_cache/image_view.cpp
```diff
@@ -64,6 +68,18 @@ ImageViewInfo::ImageViewInfo(const AmdGpu::Image& image, const Shader::ImageReso
     range.extent.levels = image.NumViewLevels(desc.is_array);
     range.extent.layers = image.NumViewLayers(desc.is_array);
     type = image.GetViewType(desc.is_array);
+    if (image.IsCube() && !desc.is_written && !desc.is_storage) {
```
squidbus, line 71: Recommend pulling these checks into `GetViewType` and adjusting its usages accordingly. Should simplify the change in `spirv_emit_context` to just be adding a switch case, too.

raphaelthegreat:
Do the issues with cubemaps also occur on AMD gpus that use native cubemap instructions? Also note using cubemaps again will cause the bloodborne LSD issue on polaris iirc

UltraDaCat:
> Do the issues with cubemaps also occur on AMD gpus that use native cubemap instructions? Also note using cubemaps again will cause the bloodborne LSD issue on polaris iirc

Well, our beloved kornjača was right, the LSD is back in Bloodborne! Tested on my RX 560;

MAIN:
image 
PR:
image

Hog185:
we wouldn’t dare upset the bloodborne fans 
image

squidbus:
> Do the issues with cubemaps also occur on AMD gpus that use native cubemap instructions?

The issues are not from the coordinate calculations, they're from 2D arrays not having seamless sampling.

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
```diff
@@ -889,11 +891,29 @@ IR::Value FixCubeCoords(IR::IREmitter& ir, const AmdGpu::Image& image, const IR:
...
     }
-    // AMD cube math results in coordinates in the range [1.0, 2.0]. We need
-    // to convert this to the range [0.0, 1.0] to get correct results.
-    const auto fixed_x = ir.FPSub(IR::F32{x}, ir.Imm32(1.f));
-    const auto fixed_y = ir.FPSub(IR::F32{y}, ir.Imm32(1.f));
-    return ir.CompositeConstruct(fixed_x, fixed_y, face);
+    // The cube ALU already projected the sample direction onto the selected face, so x and y are
+    // the face coordinates and face is slice * 8 + face_id. Invert that here to rebuild the
+    // direction Vulkan samples a cube with.
+    const IR::F32 s{ir.FPSub(IR::F32{x}, ir.Imm32(1.f))};
```
raphaelthegreat, line 897: You should stub the CUBE* instructions instead to avoid undoing their work. Look at how they were implemented before https://github.com/shadps4-emu/shadPS4/pull/2083 for this

src/video_core/texture_cache/image.cpp
```diff
@@ -136,6 +136,9 @@ Image::Image(const Vulkan::Instance& instance_, Vulkan::Scheduler& scheduler_,
     if (info.props.is_block && instance->IsBlockTexelViewSupported()) {
         flags |= vk::ImageCreateFlagBits::eBlockTexelViewCompatible;
     }
+    if (info.props.is_cube) {
```
raphaelthegreat, line 139: I think this is what breaks polaris? Might worth testing if not including that flag works there

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
```diff
@@ -928,6 +924,10 @@ Id ImageType(EmitContext& ctx, const ImageResource& desc, Id sampled_type) {
     const auto format = desc.is_atomic ? GetFormat(image) : spv::ImageFormat::Unknown;
     const auto type = image.GetViewType(desc.is_array);
     const u32 sampled = desc.is_written ? 2 : 1;
+    if (image.IsCube() && !desc.is_written && !desc.is_storage && !desc.is_depth) {
```
raphaelthegreat, line 927: Storage cubemaps do exist, so why block storage? Also AddCapability should not be here, it should be in SetupCapabilities

raphaelthegreat (commented):
Needs some work, also in general you should handle cubemap and cubemap arrays depending on if the image sharp has the DA flag set

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
squidbus: The instructions are needed for storage case where 2D array is used.

src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
squidbus: Because storage cases are primarily where Vulkan cube map capabilities are violated. In this case they remain 2D arrays.

src/shader_recompiler/ir/passes/resource_tracking_pass.cpp
raphaelthegreat: I think in that case we should fully move the cube instruction handling in the spirv backend. So resource tracking can decide if it wants to use cubemap sampling or 2d array sampling it can deterministically recover the original coordinates

src/video_core/texture_cache/image.cpp
squidbus: Mentioned above, but yeah this causes validation errors currently as we don't always know when the base image is created if it will be used as a cube. It does need heuristics mentioned to ensure we only put cube compatible on images it is allowed on, though.

raphaelthegreat: In general this ties to a texture cache rework that needs to be done at some point where image usage is part of the key, to allow not being super conservative with creation flags.

For now a simple heuristic can be added for example image is 2D, width == height and layers >= 6. And disable this entirely for polaris GPUs on windows

georgemoralis:
@Kravickas this needs manual rebase

PR #4565, opened 2026-06-12
Update ImageInfo::UpdateSize for mip calculations

Hog185:
Refactor mip size calculations to use power of two dimensions.

fixes resolve overlap issue in tony hawk skater pro skater 2+3

### DIFF
--- a/src/video_core/texture_cache/image_info.cpp
+++ b/src/video_core/texture_cache/image_info.cpp
@@ -152,23 +152,20 @@ bool ImageInfo::IsCompatible(const ImageInfo& info) const {
 
 void ImageInfo::UpdateSize() {
     guest_size = 0;
+    const u32 pow2_pitch = props.is_pow2 ? std::bit_ceil(pitch) : pitch;
+    const u32 pow2_height = props.is_pow2 ? std::bit_ceil(size.height) : size.height;
+    const u32 pow2_depth = props.is_pow2 ? std::bit_ceil(size.depth) : size.depth;
     for (s32 mip = 0; mip < resources.levels; ++mip) {
-        u32 mip_w = pitch >> mip;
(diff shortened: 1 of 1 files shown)


Hog185:
@squidbus can you review this

src/video_core/texture_cache/image_info.cpp
```diff
@@ -152,24 +152,21 @@ bool ImageInfo::IsCompatible(const ImageInfo& info) const {
 
 void ImageInfo::UpdateSize() {
     guest_size = 0;
+    const u32 pow2_pitch = props.is_pow2 ? std::bit_ceil(pitch) : pitch;
+    const u32 pow2_height = props.is_pow2 ? std::bit_ceil(size.height) : size.height;
+    const u32 pow2_depth = props.is_pow2 ? std::bit_ceil(size.depth) : size.depth;
```
squidbus, line 157: Can you explain why this should be done up-front and not on the mip size? What is this based on? Also why doesn't it apply to mip 0?

Hog185: pow2pad affects only mip-chain layout/offsets for levels>1, never the base level's size.

squidbus: Where is this information from?

Hog185: my massive brain

georgemoralis: ok that is probably not a great answer and if you can't verify your finding that pr can't be merged and will eventually close . Random fixes can cause issues in other games and we need to be careful of what we are merging

Hog185:
i can’t verify this, outside of the fact that the game gets further

PR #4568, opened 2026-06-12
sceMouse HLE

kalaposfos13:
This is a quick and dirty (well not really dirty per se, it works just fine) SDL implementation of the library made specifically for my own homebrew, but as the post below shows, it does work with commercial games as well. There's a new input setting for the config named use_mice_as_mice to enable this (defaults to `false` to avoid complaints like "why did the mouse stop working after I updated" from people who already had it bound to something), and it supports handling mice both unified and separately.


brad0demx:
[Screencast_20260615_160917.webm](https://github.com/user-attachments/assets/c991fa85-a738-4c95-9b2f-fb21dc20a8a4)

Mouse control works in Terraria 🎉

src/core/libraries/mouse/mouse.h
```diff
@@ -1,29 +1,95 @@
...
-int PS4_SYSV_ABI sceMouseDisconnectPort();
-int PS4_SYSV_ABI sceMouseGetDeviceInfo();
-int PS4_SYSV_ABI sceMouseInit();
-int PS4_SYSV_ABI sceMouseMbusInit();
-int PS4_SYSV_ABI sceMouseOpen();
-int PS4_SYSV_ABI sceMouseRead();
-int PS4_SYSV_ABI sceMouseSetHandType();
-int PS4_SYSV_ABI sceMouseSetPointerSpeed();
-int PS4_SYSV_ABI sceMouseSetProcessPrivilege();
+template <class T>
```
StevenMiller123: Since we're using this RingBufferQueue in multiple places, it might be better to place it in a header in `src/common`, then import that header here and for the pad stuff using it.

src/core/libraries/mouse/mouse.cpp
```diff
@@ -54,14 +74,44 @@ int PS4_SYSV_ABI sceMouseMbusInit() {
...
+    mouse_handles[index] = index;
+    LOG_INFO(Lib_Mouse, "out handle: {}", mouse_handles[index]);
+    return mouse_handles[index];
 }
 
-int PS4_SYSV_ABI sceMouseRead() {
-    LOG_DEBUG(Lib_Mouse, "(STUBBED) called");
-    return ORBIS_OK;
+int PS4_SYSV_ABI sceMouseRead(s32 handle, OrbisMouseData* pData, s32 num) {
+    // LOG_DEBUG(Lib_Mouse, "(DUMMY) called, h: {}, n: {}", handle, num);
```
StevenMiller123: Might be nice to have this debug log uncommented before merging?

src/core/libraries/mouse/mouse.cpp
```diff
@@ -54,14 +74,44 @@ int PS4_SYSV_ABI sceMouseMbusInit() {
...
+    if (pParam->flag == MouseOpenBehaviour::Merged && index != 0) {
+        LOG_ERROR(Lib_Mouse, "Only one mouse can be opened in merged mode!");
+        return ORBIS_MOUSE_ERROR_ALREADY_OPENED;
+    }
+    g_is_merged_mode = pParam->flag == MouseOpenBehaviour::Merged;
+    if (mouse_handles[index] != -1) {
+        LOG_ERROR(Lib_Mouse, "already opened");
+        return ORBIS_MOUSE_ERROR_ALREADY_OPENED;
+    }
+    mouse_handles[index] = index;
```
StevenMiller123, line 94: Decompilation suggests the library has 8 global handles, each handle must have either a different user_id or index. I doubt this will cause trouble any time soon, but it's worth noting.

src/core/libraries/mouse/mouse.cpp
```diff
@@ -54,14 +74,44 @@ int PS4_SYSV_ABI sceMouseMbusInit() {
...
+    LOG_INFO(Lib_Mouse, "out handle: {}", mouse_handles[index]);
+    return mouse_handles[index];
 }
 
-int PS4_SYSV_ABI sceMouseRead() {
-    LOG_DEBUG(Lib_Mouse, "(STUBBED) called");
-    return ORBIS_OK;
+int PS4_SYSV_ABI sceMouseRead(s32 handle, OrbisMouseData* pData, s32 num) {
+    // LOG_DEBUG(Lib_Mouse, "(DUMMY) called, h: {}, n: {}", handle, num);
+    if (mouse_handles[0] != handle && mouse_handles[1] != handle) {
```
StevenMiller123, line 104: Before checking handles, you should check `pData` and `num` parameters. Real library has checks against null `pData` and num greater than 64.

StevenMiller123 (commented):
Looks mostly good, but I do have a few comments based on some quick decompiling I've done.

src/core/libraries/mouse/mouse.cpp
kalaposfos13: It does appear that real hardware can indeed handle two mice per user, and if someone shows me a game that is actually set up to handle all 8 slots (which I can't even test neither on real hardware not on my laptop, due to the lack of USB ports, even with my one USB hub added), I will gladly add support for that as well, but until then, the first user gets to have up to two mice, and that's it.

PR #4576, opened 2026-06-14
build: Put system library use behind flag

squidbus:
Build using bundled libraries by default, to avoid unexpectedly developing and testing against different versions of libraries than we ship.

cc: @abouvier so you're aware of this change


ConnorG512:
Just wondering if there are any updates to merging this flag so I can work with in the Nix derivation? Does this also mean that if a library is not found it will fallback to compiling the submodule from source? Many thanks.

squidbus:
> Does this also mean that if a library is not found it will fallback to compiling the submodule from source? Many thanks.

* Now: System libraries will be used if found, otherwise submodules.
* After this PR:
  * If the build flag is not specified, only submodules will be used.
  * If the build flag is specified, system libraries will be used if found, otherwise submodules, same as before.

ConnorG512:
Ah okay thank you for clarifying for me. Appreciate it.

PR #4579, opened 2026-06-14
np manager improvements and settings

georgemoralis:
Added np settings and some preparations for shadnet

### DIFF
--- a/src/common/elf_info.h
+++ b/src/common/elf_info.h
@@ -94,6 +94,7 @@ public:
     static constexpr u32 FW_700 = 0x7000000;
     static constexpr u32 FW_750 = 0x7500000;
     static constexpr u32 FW_800 = 0x8000000;
+    static constexpr u32 FW_900 = 0x9000000;
     static constexpr u32 FW_1000 = 0x10000000;
     static constexpr u32 FW_1150 = 0x11500000;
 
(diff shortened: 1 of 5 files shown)


StevenMiller123 (changes requested):
Most of this looks fine, however these functions are mostly all lacking proper user_id validity checks.
Now that these functions are actually depending on other emulator components, I think it's important we make sure inputted user_ids actually correspond to an active user, and we return proper errors if they don't.

Other than that, the rest of this stuff seems fine to me.

georgemoralis:
> Most of this looks fine, however these functions are mostly all lacking proper user_id validity checks. Now that these functions are actually depending on other emulator components, I think it's important we make sure inputted user_ids actually correspond to an active user, and we return proper errors if they don't.
> 
> Other than that, the rest of this stuff seems fine to me.

User_ids coming from np_handler which is the connector for shadnet . Tried to keep the pr simple and didn't add any shadnet stuff yet . All calls return the default user id , which 99% of the time will be ok . Also without enabled shadnet switch that doesn't make any use still :D

StevenMiller123:
We've seen games throw garbage at user_id fields before, which is why I choose to mention it.
While it doesn't matter yet, it could cause strange issues later down the line when per-user stuff is actually respected.

georgemoralis:
@StevenMiller123 added some user checks, rest will go with a new pr , (add more shadnet stuff as well)

StevenMiller123 (approved):
I'd rather my review gets properly addressed, but for now this will do. Aside from the missing user_id validity checks, this seems fine.

PR #4581, opened 2026-06-15
Fix P.T. demo crash

itsmattbytes:
Fixes the following error which caused the P.T. demo to crash at launch. (CUSA01127)

> [Debug] <Critical> (shadPS4:GpuComm) image.cpp:258 GetBarriers: Assertion Failed!


squidbus:
Your PR seems to be improperly rebased and is removing parts of a recent commit before it.

src/video_core/texture_cache/image.cpp
```diff
@@ -234,28 +229,43 @@ Image::Barriers Image::GetBarriers(vk::ImageLayout dst_layout, vk::AccessFlags2
         if (!partially_transited) {
             subresource_states.resize(info.resources.levels * info.resources.layers);
             std::fill(subresource_states.begin(), subresource_states.end(), last_state);
+        } else if (subresource_states.size() != info.resources.levels * info.resources.layers) {
+            // The image's resource extents have changed since subresource_states was last
+            // populated (e.g. the image was expanded). Resize to match, inheriting the last
+            // known state for any newly added subresources.
+            subresource_states.resize(info.resources.levels * info.resources.layers, last_state);
         }
```
squidbus: Where is this supposedly happening? We don't change the resource count of an image, `ExpandImage` creates and copies to a new one.

src/video_core/texture_cache/image.cpp
```diff
@@ -234,28 +229,43 @@ Image::Barriers Image::GetBarriers(vk::ImageLayout dst_layout, vk::AccessFlags2
...
+            needs_partial_transition ? std::min(subres_range->base.level, mip_end) : 0u;
+        const auto layer_end = needs_partial_transition
+                                    ? std::min(subres_range->base.layer + subres_range->extent.layers,
+                                               info.resources.layers)
+                                    : info.resources.layers;
+        const auto layer_begin =
+            needs_partial_transition ? std::min(subres_range->base.layer, layer_end) : 0u;
+
+        const auto mips = std::views::iota(mip_begin, mip_end);
+        const auto layers = std::views::iota(layer_begin, layer_end);
```
squidbus: This seems to me like there's incorrect `subres_range` being passed in rather than something we should be sanitizing here.

squidbus: Or alternatively, seems like it could be a case where the cache was supposed to expand the image, making the resource range valid.

itsmattbytes: Thanks for the review, still looking into the source cause.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -694,12 +694,19 @@ void Rasterizer::BindTextures(const Shader::Info& stage, Shader::Backend::Bindin
             if (mip_fallback_mode == Shader::MipStorageFallbackMode::ConstantIndex) {
                 ASSERT(num_bindings == 1);
                 desc.view_info.range.base.level += image_desc.constant_mip_index;
+                // Cap to the last available mip so the view range stays within the image's
+                // actual subresources. Out-of-range constant LODs can occur when the shader's
+                // LOD immediate exceeds the TSHARP's last_level.
+                desc.view_info.range.base.level =
+                    std::min(desc.view_info.range.base.level, desc.info.resources.levels - 1);
```
squidbus, line 714: Technically I think in this case we would bind a null descriptor, for out of bounds access behavior.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -694,12 +694,19 @@ void Rasterizer::BindTextures(const Shader::Info& stage, Shader::Backend::Bindin
...
                 desc.view_info.range.extent.levels = 1;
             } else if (mip_fallback_mode == Shader::MipStorageFallbackMode::DynamicIndex) {
                 desc.view_info.range.base.level += i;
                 desc.view_info.range.extent.levels = 1;
             }
 
+            // Save range before FindImage can overwrite base.level/base.layer from a parent image.
+            desc.initial_range = desc.view_info.range;
```
squidbus, line 722: This seems like something we should just save in the `ImageViewInfo` constructors for later reference.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -731,10 +738,22 @@ void Rasterizer::BindTextures(const Shader::Info& stage, Shader::Backend::Bindin
...
-                old_image.binding = {};
-                image_id = texture_cache.FindImage(desc);
+            {
+                auto& old_image = texture_cache.GetImage(image_id);
+                // Re-find if (a) ExpandImage marked this image for rebind, or (b) the slot was
+                // silently reused between passes by another TSHARP's FindImage — detected by the
+                // stored range now being OOB for whatever image lives in that slot.
+                const bool range_oob =
+                    desc.view_info.range.base.level >= old_image.info.resources.levels ||
+                    desc.view_info.range.base.layer >= old_image.info.resources.layers;
```
squidbus, line 755: Suggest using the `image_uid` that was added to `Image` in some way, if your goal is to detect that the image has changed. This method here seems like it won't always work.

georgemoralis:
rebased

PR #4584, opened 2026-06-16
Add Page protected writes

Hog185:
should fix black screen issue in ufc 1, however it is untested due to lack of a decent GPU right now


Kravickas:
NHL19
```
[Debug] <Critical> (RenderThread) page_manager.cpp:206 Protect: Assertion Failed!
Attempted to protect region as write-only which is not a valid permission
```

PR #4586, opened 2026-06-17
emulator: Set window icon to game icon.

squidbus:
When a game is loaded, attempt to read its `sce_sys/icon0.png` and set the SDL window icon.


gandalfthewhite19890404:
What fo you think about using icon for taskbar too?

PR #4588, opened 2026-06-17
camera: fix incorrect parameter and SDK check and add some more error logging

kalaposfos13:
This fixes Camera initialization in THE PLAYROOM.


StevenMiller123 (approved):
Looks good

PR #4591, opened 2026-06-18
Synchronization Between Storage Images and Guest Memory

TdlifePeng:
A collection of game-specific rendering fixes for God of War III Remastered, all gated behind a game serial check in `Gow3Features::Init()` and active only for CUSA01623 / CUSA01715 / CUSA01740. No impact on other titles.

For reference only.

### `storage_image_sync`
Downloads compute shader storage image output to guest memory after each CS dispatch. PS4 unified memory lets textures at the same address read CS output directly; in shadPS4 each format gets its own VkImage, so data must be explicitly copied through guest memory. Includes re-tiling via `TileLinearBuffer` (CopyImageToBuffer always produces linear data; guest memory must stay tiled to match the T# descriptor) and buffer cache notification so downstream texture refreshes pick up the new data.

### `rt_alias_copy` (pull model)
When a texture at an address that previously had a render target recorded is bound for sampling, copies from the RT VkImage to the texture VkImage. Handles cross-format and cross-sample-count copies (vkCmdResolveImage for MSAA->1x). Deduplicated per (addr, tex_id) per submit.

### `1x1_readback`
Force-downloads 1x1 render targets to guest memory so the CPU can read GPU-computed results (auto-exposure).

### `depth_clear_skip`
Skips depth buffer clear when `depth_write_enable` is false, matching PS4 hardware behaviour that preserves previous depth data through passes that disable depth writes.

## Known Issues

Giant (Kronos) textures randomly render incorrectly on all GPU vendors. Root cause traced to game-side non-deterministic T# descriptor aliasing: during the loading window, the game sometimes writes the same base_address to multiple BC1 texture slots instead of their own distinct data. Not introduced by these patches.

Tracing one of the striped textures (2e3e3cb00.png) shows that it originates from a 1024x1024 BC3 texture, which later gets copied to address 2e3e3cb00 and then interpreted as a 2048x2048 BC1 texture. The source appears to come from R_POSHADES00.WAD, processed through CPU-side swizzling, and then written through a storage image to an address overlapping the giant texture. Its intended use seems to be a mountain texture used later in the level.

I'm not sure whether this behavior is caused by this patch or whether the patch is merely making an existing issue easier to reproduce.

## Future Direction

The current approach relies on per-feature sync points (storage sync, RT alias copy, 1x1 readback) that each handle one specific data path. A more comprehensive solution would be a unified GPU-tracked memory manager that
monitors all guest memory and GPU objects through the buffer cache, copies between them when aliasing is detected, and writes back to guest memory on demand -- effectively emulating PS4 unified memory at the right granularity rather than patching individual producer -> consumer pairs.

---

[Claude Code, deepseek-v4-pro]

Screenshot5
Screenshot2
Screenshot4
Screenshot3
renderdoc1
renderdoc2


squidbus:
> A collection of game-specific rendering fixes for God of War III Remastered, all gated behind a game serial check in Gow3Features::Init() and active only for CUSA01623 / CUSA01715 / CUSA01740. No impact on other titles.

This is an immediate red flag, we don't accept game-specific hacks upstream.

TdlifePeng:
That's fair. I only know it works for the game I'm testing and have no idea whether it would generalize.

A proper solution would likely require comprehensive guest memory tracking, implying major buffer cache rework and broader refactoring, which is well beyond the scope of this experiment.

TdlifePeng:
I've removed the game-specific condition.

Unfortunately, it's also the only game I currently have that can reliably reach this code path. The giant texture corruption still shows up three times on the way to Poseidon, and I still have no idea what's actually causing it.

I also tried testing with The Order 1886, but the scene becomes so dark it's barely playable, as if dynamic lighting textures are never being updated. Those textures don't appear to involve overlapping addresses, so I'm equally stuck there.

If anyone is able to test compatibility with other games, I'd really appreciate it.

PR #4594, opened 2026-06-18
Net: Various setsockopt and getsockopt fixes

StevenMiller123:
Currently, it would appear as if our socket code is trying to mimic... Winsock.
This PR makes various changes to our behavior so we more closely match what we're actually trying to emulate.

Opening as draft as I would still like to work on some other things before a merge. This also needs testing because I disabled a few ops that Winsock openly claims don't match POSIX, and I've removed some hacks that were presumably in place because of Winsock.


StevenMiller123:
Actual changelist:
- Return ORBIS_NET_EINVAL for incorrect optlen in setsockopt
    - Winsock defaults to EFAULT, and we were returning EFAULT. PS4 returns EINVAL.
- Remove ORBIS_NET_EFAULT returns for low optlen in getsockopt
    - PS4 doesn't error, it just copies how much data is requested to the output.
- Support the non-net variants of SO_SNDTIMEO and SO_RCVTIMEO
    - setsockopt receives a timeval, and getsockopt outputs a timeval. PS4 internally stores timeout in milliseconds.
- Fix setsockopt for net-variants of SO_SNDTIMEO and SO_RCVTIMEO.
    - optval is in milliseconds, not microseconds.
- Fix getsockopt for net-variants of SO_SNDTIMEO and SO_RCVTIMEO
    - We were just running the host's getsockopt with these opts. This would work fine on Windows, but not on POSIX-supporting platforms.
- Properly support SO_NAME for setsockopt and getsockopt
    - This uses the same name set by sys_socketex, and can be set through setsockopt too.
- Fix handling of SO_ERROR_EX
    - PS4 internally or's the error code with ORBIS_NET_ERROR_BASE for this opt.
- Fix getsockopt on SO_LINGER
    - We were running host getsockopt directly, which works on POSIX platforms, but would return invalid data on Windows due to linger struct differences.

I've also removed some hacky opt implementations. I still need to decompile more of this stuff to judge how to properly handle these operations. 
- Disable IP_MULTICAST_LOOP
    - Winsock documents explicitly state their implementation of this opt is not compliant with the POSIX version. As such, we'll need to emulate this.
- Disable SO_ONESBCAST
    - I can't find proper word on how this is supposed to work, but the kernel does treat this differently from SO_BROADCAST.
- Remove ORBIS_NET_EPROCUNAVAIL return from SO_LINGER handling.
    - PS4 does not return an error for this case. If Winsock does, then we need a better workaround. Needs testing.
- Remove ORBIS_NET_EPROCUNAVAIL return from IP_HDRINCL handling.
    - PS4 does not return an error for this case. If Winsock does, then we need a better workaround. Needs testing.

StevenMiller123:
Marking as ready for review since work on this has stalled, but it does fix a couple games (and fixes a homebrew I'm still working on).

georgemoralis:
what games it fixes? (so i can test on windows)

StevenMiller123:
Should fix Among Us, I originally worked on this to fix an issue that shows up in Bleach: Brave Souls with network on (though that's hard to dump since it's a free-to-play digital-only game)

georgemoralis:
AmongUs crashes earlier with 

[Lib.Net] <Error> (Background Job.Worker 10) sys_net.cpp:414 sys_recvfrom: s = 25 (anon) returned error code: 40
[Tty] <Error> (Background Job.Worker 10) logger.cpp:60 log_flush: [stderr] Ping: Error receiving ICMP packet response. Possibly a timeout.: Message too long

StevenMiller123:
@georgemoralis please send a log with the Lib.Net:Debug filter?

PR #4599, opened 2026-06-19
Implement sceMouseClose and improve disconnected state handling

kalaposfos13:
Let's say that hypothetically, someone wants to write a homebrew that wants to know if a mouse is connected or not. For this, Sony has graciously provided us with the following interface:
- Initialize the library. (completely fine, almost everything needs this as the zeroth step)
- Open a mouse handle.
- Read mouse state.
  - If a mouse is connected for this index, then provided you didn't insert a sleep between the last two steps, you get 0 reported states here.
  - If a mouse is not connected, you always get one default state with `connected` set to false.
- Going by the above description, if you get 0 states, then assume a mouse is connected to that handle, otherwise check the connected field in the latest state, as it is technically possible that you were shaking the mouse while initialization occurs, and it sent some data (if you don't add a wait, the chance for this is practically zero, but still).
- You then close the handle, as you only needed to check connection, not start consuming the mouse states.

I personally thank Sony for giving us this very pleasant and nice to work with interface for this.

### DIFF
--- a/src/core/libraries/mouse/mouse.cpp
+++ b/src/core/libraries/mouse/mouse.cpp
@@ -19,10 +19,15 @@ namespace Libraries::Mouse {
 RingBufferQueue<OrbisMouseData> mouse_states[2]{{64}, {64}};
 s32 mouse_handles[2]{-1, -1};
 s32 mouse_sdl_handles[2]{-1, -1};
-bool g_lib_init = false, g_is_merged_mode = false;
+bool g_lib_init = false, g_is_merged_mode = false, g_are_mice_enabled;
 
 int PS4_SYSV_ABI sceMouseClose(s32 handle) {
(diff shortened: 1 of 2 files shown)


src/core/libraries/mouse/mouse.cpp
```diff
@@ -94,28 +100,31 @@ int PS4_SYSV_ABI sceMouseOpen(Libraries::UserService::OrbisUserServiceUserId use
         LOG_ERROR(Lib_Mouse, "already opened");
         return ORBIS_MOUSE_ERROR_ALREADY_OPENED;
     }
-    mouse_handles[index] = index;
+    mouse_handles[index] = index + 0;
```
StevenMiller123: `index + 0`

georgemoralis: @kalaposfos13 ?

PR #4605, opened 2026-06-22
Lib.Mouse: Fix sceMouseOpen checks

StevenMiller123:
Looking closer at decompilation, null parameters is allowed. If pParam is null, or if the game has SDK ver < FW 2.00, it will just follow the code path for non-merged mode. If param is non-null, SDK is greater than FW 2.00, and the first bit of the flag is true, then it follows the code for merged mode.

For simplicity, this PR ignores the FW 2.00 check. If it comes up in the future, we can handle it.


kalaposfos13 (commented):
This is only a style nitpick, but while `pParam && True(pParam->flag & MouseOpenBehaviour::Merged)` works, it's a) duplicated and b) a bit hard to read at a glance, so I'd factor it out into a new variable on the top of the function, and make it more explicit that Normal is the default when no parameter is given, for example with `pParam ? pParam->flag : MouseOpenBehaviour::Normal)`.

StevenMiller123:
Removed duplicated check and made it explicitly clear that Normal behavior is default.

kalaposfos13 (approved):

PR #4607, opened 2026-06-23
First ShadNet integration

georgemoralis:
Check our new wiki page!

https://wiki.shadps4.net


src/core/libraries/np/np_handler.cpp
```diff
@@ -19,16 +22,1796 @@ NpHandler& NpHandler::GetInstance() {
     return s_instance;
 }
 
-bool NpHandler::IsPsnSignedIn(s32 user_id) const {
-    if (EmulatorSettings.IsShadNetEnabled()) {
-        return true; // fake just return true if shadnet is enabled //TODO
+std::pair<std::string, u16> NpHandler::ParseServerAddress() const {
+    const std::string server_str = EmulatorSettings.GetShadNetServer();
+    std::string host = server_str;
```
StevenMiller123: Would be good to have error logs for when the shadnet server field is empty.

src/core/libraries/np/np_handler.cpp
```diff
@@ -19,16 +22,1796 @@ NpHandler& NpHandler::GetInstance() {
...
+        proto.set_data(std::string(reinterpret_cast<const char*>(gameInfoData), gameInfoSize));
+    }
+
+    const std::string proto_bytes = proto.SerializeAsString();
+    const std::string com_id = GetNpCommId(service_label);
+    if (!IsValidNpCommId(com_id)) {
+        LOG_ERROR(NpHandler,
+                  "{}: no valid NP Communication ID (npbind.dat missing or blank); "
+                  "rejecting request",
+                  __func__);
```
StevenMiller123: As always, there is no point in including the function name, the log macros already add it.

I won't comment on other cases of this, but do try to fix it all.

georgemoralis:
Updated with instructions

src/shadnet/client.cpp
```diff
@@ -0,0 +1,730 @@
...
+    if (payload.empty()) {
+        LOG_ERROR(ShadNet, "Empty GetToken reply");
+        m_sem_authenticated.release();
+        return;
+    }
+    const ErrorType err = static_cast<ErrorType>(payload[0]);
+    if (err != ErrorType::NoError) {
+        LOG_WARNING(ShadNet,
+                    "GetToken returned error={} — WebAPI calls will be unauthenticated. "
+                    "(Server build may predate cmd 39.)",
```
StevenMiller123: What does this even mean?

src/shadnet/client.cpp
```diff
@@ -0,0 +1,730 @@
...
+            ::getsockopt(m_sock, SOL_SOCKET, SO_ERROR, reinterpret_cast<char*>(&err), &len);
+            connected = (err == 0);
+        }
+    }
+    if (!connected) {
+        LOG_ERROR(ShadNet, "connect() timed out or failed for {}:{}", m_host, m_port);
+        SHAD_CLOSE(m_sock);
+        m_sock = SHAD_INVALID_SOCK;
+        m_state = ShadNetState::FailureConnect;
+        return false;
```
StevenMiller123, line 324: While testing, I've had internet flickers completely disable ShadNet because of these timed out errors. I think we should probably just retry connection, not error out.

src/shadnet/client.h
```diff
@@ -0,0 +1,284 @@
...
+    Delete = 3,
+    SendToken = 4,
+    SendResetToken = 5,
+    ResetPassword = 6,
+    ResetState = 7,
+    AddFriend = 8,
+    RemoveFriend = 9,
+    AddBlock = 10,
+    RemoveBlock = 11,
+    // 12–29: room/lobby/ticket NOT implemented in shadNet
```
StevenMiller123: Probably best to have those in the enum, even if they'd be unused currently.

src/sdl_window.cpp
```diff
@@ -386,6 +387,12 @@ Uint32 wheelOffCallback(void* og_event, Uint32 timer_id, Uint32 interval) {
 void WindowSDL::OnKeyboardMouseInput(const SDL_Event* event) {
     using Libraries::Pad::OrbisPadButtonDataOffset;
 
+    // Global UI hotkey: F2 toggles the shadNet Friends window.
+    if (event->type == SDL_EVENT_KEY_DOWN && !event->key.repeat && event->key.key == SDLK_F2) {
```
StevenMiller123: We need a good way to do this on controller-only setups, or setups without proper keyboards.

StevenMiller123 (changes requested):

src/shadnet/client.h
metr1k-91: Agree with this.  I have matching2 coming shortly, can add these back in with no changes to how they were, I'll modify if needed.

src/core/emulator_settings.h
```diff
@@ -605,8 +602,8 @@ class EmulatorSettingsImpl {
     SETTING_FORWARD(m_general, ConsoleLanguage, console_language)
     SETTING_FORWARD(m_general, BigPictureScale, big_picture_scale)
     SETTING_FORWARD(m_general, ShadNetServer, shadnet_server)
-    SETTING_FORWARD(m_general, SignalingAddr, signaling_addr)
-    SETTING_FORWARD(m_general, SignalingPort, signaling_port)
+    SETTING_FORWARD(m_general, ShadNetWebApiServer, shadnet_webapi_server)
+    SETTING_FORWARD(m_general, ShadNetSignalingServer, shadnet_signaling_server)
```
metr1k-91: You can change this from ShadNetSignalingServer to SignalingInfo and keep the addr:port setup you showed me.  This will be for upnp fallback when it isn't available on the users router / modem setup, and / or for local testing scenarios until we have proper external vs local handling on the shadnet side.

src/shadnet/client.h
georgemoralis: fixed

src/core/libraries/np/np_handler.cpp
georgemoralis: fixed

src/core/emulator_settings.h
```diff
@@ -185,9 +185,11 @@ struct GeneralSettings {
     Setting<int> big_picture_scale{1000};
     Setting<std::string> shadnet_server{"srv.shadps4.net:31313"};
     Setting<std::string> shadnet_webapi_server{"http://srv.shadps4.net:31315"};
-    Setting<std::string> shadnet_signaling_server{"srv.shadps4.net:31314"};
+    Setting<std::string> signaling_info{};
     Setting<bool> enable_upnp{true};
 
+    SETTING_FORWARD(m_general, SignalingInfo, signaling_info)
```
metr1k-91: Good with this.

src/shadnet/client.cpp
georgemoralis: i have improved reconnections if you can retry

src/shadnet/client.cpp
georgemoralis: simplified it

src/core/libraries/np/np_handler.cpp
georgemoralis: done

src/sdl_window.cpp
georgemoralis: done

src/shadnet/client.cpp
StevenMiller123: It takes a longer disconnection, but the same issue still ends up present. Once it detects disconnection, PSN emulation is completely disabled.
It's been a while since I last used a PS4 that's on latest FW, but I'm pretty sure we shouldn't completely disable shadNet if internet dies long enough to break connection.

In my opinion, if the shadNet connection dies from a network error, then we should report not signed in or some sort of network error, but it should try to reconnect on the next NP call made by the game or something along those lines.

georgemoralis: Improved it a bit. now Network drop reports SignedOut to the game, then the worker transparently reconnects with backoff once connectivity returns.  logout/shutdown still cancel any pending reconnect so they're never resurrected.  I think anyway that further improvements are QOL so that should not stop the merging of the pr , it can be improved in another one :D

src/common/elf_info.h
UltraDaCat, line 1: You forgot the Copyright shadow :kek:

UltraDaCat (commented):
How could you forget? 💔

RuskyDed:
ImGui stuff from this PR broke external fps counters on both Windows and Linux for me (tied to vblank value now rather than the actual fps). The embedded one by f10 is still acting correctly.

PR #4610, opened 2026-06-23
video_core: Implement IT_SET_PREDICATION using VK_EXT_conditional_rendering

cuesta4:
This PR implements PM4 draw predication (`IT_SET_PREDICATION`), used by titles such as *God of War III Remastered* for occlusion culling.

On real hardware the GPU's command processor evaluates the predicate when the predicated packets **execute** — applications never wait for occlusion results before enabling predication (`kPredicationZPassHintWait` is resolved by the GPU). Since the emulator parses PM4 far ahead of actual GPU execution, evaluating predicates at parse time would either stall (full CPU/GPU sync per predicated object) or always fail open. This implementation instead builds the predicate on the GPU timeline:

- **Zpass**: draws between pixel-pipe stat dumps are measured with per-draw Vulkan occlusion queries. On `SET_PREDICATION`, the query results are copied into a predicate buffer on the GPU (single-query windows use a direct 32-bit `vkCmdCopyQueryPoolResults`; multi-query and `continue` chains go through a small reduction compute shader). Predicated draws/dispatches are wrapped in `vkCmdBegin/EndConditionalRenderingEXT`. The CPU never waits for query results.
- **Bool32/Bool64**: the guest value is sampled on the GPU timeline via a buffer copy, since it may be produced by GPU work queued earlier in the same stream.
- **Guest counter writeback**: pixel-pipe ZPASS counters are written back asynchronously once the producing submission completes (no stalls).
- **Fail-open policy**: missing extension, unknown ops (`PrimCount`), invalid addresses or unproducible predicates all execute draws, so predication bugs can never incorrectly hide geometry.
- The index base staged by a GPU-predicated `DrawIndex2` is restored for packets outside the predicate, since hardware would also skip the base update when the draw is skipped (avoids geometry ghosting).

Requires `VK_EXT_conditional_rendering` (widely supported); falls open with a warning when unavailable.

**Validation** (GoW III Remastered, instrumented runs): ~180k predication windows over a play session, 0 fail-open, ~56% of predicated draws discarded by the GPU in occluded scenes; guest ZPASS counter writeback verified against `OcclusionQueryResults::isReady()/getZPassCount()` semantics.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1104,6 +1104,8 @@ set(VIDEO_CORE src/video_core/amdgpu/cb_db_extent.h
                src/video_core/renderer_vulkan/vk_pipeline_serialization.h
                src/video_core/renderer_vulkan/vk_platform.cpp
                src/video_core/renderer_vulkan/vk_platform.h
+               src/video_core/renderer_vulkan/vk_predication.cpp
+               src/video_core/renderer_vulkan/vk_predication.h
                src/video_core/renderer_vulkan/vk_presenter.cpp
                src/video_core/renderer_vulkan/vk_presenter.h
                src/video_core/renderer_vulkan/vk_rasterizer.cpp
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -65,6 +65,28 @@ static std::span<const u32> NextPacket(std::span<const u32> span, size_t offset)
     return span.subspan(offset);
 }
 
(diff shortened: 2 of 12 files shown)


squidbus:
General comment, I can't imagine waiting on the GPU for readbacks to predicate on host is very efficient, does something like `VK_EXT_conditional_rendering` not work here to predicate operations on the GPU side?

cuesta4:
Yep, probably. Converted to a draft to think about how to do it.

PR #4611, opened 2026-06-24
Barrier vertex/index buffers written by compute

Kravickas:
BindVertexBuffers and BindIndexBuffer had no barrier, so a buffer written by a compute dispatch and then read as vertex or index input had no synchronization between the write and the read, causing vertex explosions on AMD GPUs.

I want to gate it behind a new experimental setting.

Test for performance regressions against nightly.


Hog185:
no noticable performance difference 
7600xt/5 5600

raphaelthegreat:
This looks good. I wonder if in the future we could better utilize the synchronization PM4 packets like AcquireMem to avoid manual barrier tracking like this but for now this is okay

tochnonechelovek:
No notable changes compared to nightly.
i5 12400F, RTX 3060.

Kyoskii:
tested with bloodborne, performance seems to be on par with nightly.

Kravickas:
Hmm, live test without the setting? 😏

StevenMiller123:
Fixes Persona 5 Royal vertex explosions.
image
[CUSA17419.log](https://github.com/user-attachments/files/29388106/CUSA17419.log)



Does not impact Call of Duty Advanced Warfare vertex explosions (seems to be a different vulkan race, it's another thing that Vulkan Crash Diagnostics fixes)
image
[CUSA00803.log](https://github.com/user-attachments/files/29388128/CUSA00803.log)

georgemoralis:
since it doesn't seem we have any complains about this , let's add it and if it breaks something we add a setting for it

PR #4614, opened 2026-06-24
fix select function file descriptors on posix based systems

Hog185:
Refactor select function to handle immediately-ready disk and device file descriptors.


Hog185:
gets rid of "unsupported file descriptors" errors on P.T on linux

src/core/libraries/kernel/file_system.cpp
```diff
@@ -1439,12 +1445,22 @@ s32 PS4_SYSV_ABI posix_select(s32 nfds, fd_set* readfds, fd_set* writefds, fd_se
                 return -1;
             }
 
+            // Regular files and devices are always ready; stdin (fd 0) never read-ready
+            if (file->type == Core::FileSys::FileType::Regular ||
+                file->type == Core::FileSys::FileType::Device) {
```
StevenMiller123, line 1450: I think it would be best to keep running normal files through native select, though either route probably isn't 100% accurate.

Hog185: what’s your reasoning behind that?

StevenMiller123: While I'd need to test real hardware to be certain, I would expect file permissions to impact read/write-readiness in select.

Hog185: i feel like the performance difference is incredibly negligible, especially as this is select()

StevenMiller123: Performance difference doesn't matter (unless it's really severe), what matters is accuracy.

Hog185: as you said yourself neither is fully accurate, and in my opinion it’s negligible, however if this is a requirement, i’ll rewrite tmmw

Hog185:
also solves the same error message in sbk 22

Hog185:
and motogp 25

thethunderturner:
@Hog185 can you rebase please?

Hog185:
@thethunderturner done

thethunderturner:
Thanks, will try when i get home.

georgemoralis:
needs manual rebase

Hog185:
shit PR ngl

PR #4293, opened 2026-04-20
shader_recompiler: Only forward declared number of vertex inputs.

squidbus:
Use the declared number of vertex exports in determining how many parameters to forward from vertex to fragment in generated rect/quad list tessellation control shaders. In my testing this value was accurate to the actual number of vertex exports.

Fixes instances of the following validation error (e.g. in Bloodborne):
```
VUID-RuntimeSpirv-OpEntryPoint-08743: vkCreateGraphicsPipelines(): pCreateInfos[0] (SPIR-V Interface) VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT has a declared Input at Location 1 Component 0  but the previous stage (VK_SHADER_STAGE_VERTEX_BIT) has no Output declared there.
The input variable is:
  pointer to Input -> array[32] of vec4 of float32
The Vulkan spec states: Any user-defined variables shared between the OpEntryPoint of two shader stages, and declared with Input as its Storage Class for the subsequent shader stage, must have all Location slots and Component words declared in the preceding shader stage's OpEntryPoint with Output as the Storage Class (https://vulkan.lunarg.com/doc/view/1.4.341.0/mac/antora/spec/latest/appendices/spirvenv.html#VUID-RuntimeSpirv-OpEntryPoint-08743)

```

PR #4616, opened 2026-06-25
Fix Vulkan validation errors in image descriptors and auxiliary tessellation interfaces

cuesta4:
Fixes invalid VS-to-TCS interfaces in auxiliary tessellation shaders used for rect and quad lists.

The generated TCS now declares inputs only for locations actually exported by the vertex shader, while preserving the outputs required by later stages. Missing inputs are synthesized as zero without adding invalid interface declarations. Clip-distance emulation is handled as a separate shifted location, and the pipeline cache version is bumped to discard stale auxiliary shaders.

The image descriptor portion of the original implementation has been removed. The new implementation is now based on the recent robustness2 changes.

Validated with sparse vertex exports, clip-distance emulation, SPIR-V validation, and the GCN test suite.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
+++ b/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
@@ -8,6 +8,8 @@
 #include "common/assert.h"
 #include "shader_recompiler/backend/spirv/emit_spirv_discard_frag.h"
 #include "shader_recompiler/backend/spirv/emit_spirv_quad_rect.h"
+#include "shader_recompiler/info.h"
+#include "shader_recompiler/ir/attribute.h"
 #include "video_core/renderer_vulkan/liverpool_to_vk.h"
 #include "video_core/renderer_vulkan/vk_graphics_pipeline.h"
 #include "video_core/renderer_vulkan/vk_instance.h"
@@ -17,6 +19,34 @@
 namespace Vulkan {
 
 using Shader::Backend::SPIRV::AuxShaderType;
+
+namespace {
+
+u64 BuildAuxTessPreviousStageOutputMask(const Shader::Info* vs_info,
+                                        const Shader::FragmentRuntimeInfo& fs_info) {
+    if (!vs_info) {
+        return 0;
+    }
+
+    u64 mask = 0;
+    if (fs_info.clip_distance_emulation &&
+        vs_info->stores.GetAny(Shader::IR::Attribute::ClipDistance)) {
+        mask |= 1ull;
+    }
+
+    for (u32 i = 0; i < Shader::IR::NumParams; ++i) {
+        const auto param = Shader::IR::Attribute::Param0 + static_cast<int>(i);
+        if (!vs_info->stores.GetAny(param)) {
+            continue;
+        }
+        const u32 location =
+            Shader::Backend::SPIRV::AuxTessAttributeLocation(i, fs_info.clip_distance_emulation);
+        mask |= 1ull << location;
+    }
+    return mask;
+}
+
+} // Anonymous namespace
 
 static constexpr std::array LogicalStageToStageBit = {
     vk::ShaderStageFlagBits::eFragment,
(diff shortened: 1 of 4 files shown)


src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -20,6 +20,104 @@
...
+
+    LOG_ERROR(Render_Vulkan,
+              "Image numeric type mismatch before descriptor write: stage={} image_id={} "
+              "expected={} requested_format={} actual={} actual_format={} storage={}",
+              stage.stage, image_id.index, NameOf(expected), vk::to_string(desc.view_info.format),
+              NameOf(actual), vk::to_string(image_view.format),
+              desc.type == VideoCore::TextureCache::BindingType::Storage);
+}
+
+} // Anonymous namespace
```
squidbus: This seems like a lot to add just to log an error message, not involved at all in the actual handling of the issue? If we really need to log this case and not just handle it, would prefer it be done where the issue is actually detected/handled.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -742,7 +840,12 @@ void Rasterizer::BindTextures(const Shader::Info& stage, Shader::Backend::Bindin
             if (instance.IsNullDescriptorSupported()) {
                 image_infos.emplace_back(VK_NULL_HANDLE, VK_NULL_HANDLE, vk::ImageLayout::eGeneral);
             } else {
-                auto& null_image_view = texture_cache.FindTexture(VideoCore::NULL_IMAGE_ID, desc);
+                const auto null_format = desc.view_info.format != vk::Format::eUndefined
+                                             ? desc.view_info.format
+                                             : desc.info.pixel_format;
+                const auto null_image_id = texture_cache.GetNullImage(null_format);
+                auto& null_image_view = texture_cache.FindTexture(null_image_id, desc);
+                LogImageViewNumberClassMismatch(stage, desc, null_image_view, null_image_id);
```
squidbus: I would think in this case we should just get a null image that matches the desired format, and this should never mismatch.

src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp
```diff
@@ -17,6 +19,53 @@ namespace Vulkan {
...
+        if (input.IsDefault()) {
+            continue;
+        }
+        const u32 location = Shader::Backend::SPIRV::AuxTessAttributeLocation(
+            input.param_index, fs_info.clip_distance_emulation);
+        if (location < 64u && ((output_mask >> location) & 1ull) == 0) {
+            LOG_DEBUG(Render_Vulkan,
+                      "Aux tess input location {} for PS input {} has no VS output; "
+                      "synthesizing zero",
+                      location, i);
```
squidbus: Think you should just log in the actual shader creation when you have a missing input, instead of this extra checking function.

src/shader_recompiler/specialization.h
```diff
@@ -134,20 +136,30 @@ struct StageSpecialization {
...
-                         spec.is_cube = sharp.IsCube();
-                         if (spec.is_storage) {
-                             spec.dst_select = sharp.DstSelect();
-                         } else {
-                             spec.is_srgb = sharp.GetNumberFmt() == AmdGpu::NumberFormat::Srgb;
-                         }
-                         spec.num_conversion = sharp.GetNumberConversion();
-                         spec.num_bindings = desc.NumBindings(*info);
-                     });
+        for (const auto& desc : info->images) {
```
squidbus: Why is this changed from `ForEachSharp`?

src/shader_recompiler/specialization.h
```diff
@@ -134,20 +136,30 @@ struct StageSpecialization {
...
+            if (!sharp.Valid() || sharp.GetDataFmt() == AmdGpu::DataFormat::FormatInvalid) {
+                binding++;
+                continue;
+            }
+            bitset.set(binding++);
+            const auto num_format = sharp.GetNumberFmt();
+            spec.type = sharp.GetViewType(desc.is_array);
+            spec.num_class = AmdGpu::GetNumberClass(num_format);
+            spec.data_format = sharp.GetDataFmt();
+            spec.num_format = num_format;
```
squidbus: Specializing on the specific data/number format sounds like a bad idea when not strictly necessary, will increase combinations. I don't see why this is even necessary, specializing on integer vs float (which we do) should already be sufficient to deal with the issue you specified.

squidbus (commented):
General note: The approach looking [at vertex outputs](https://github.com/shadps4-emu/shadPS4/pull/4293) and [both vertex outputs and fragment inputs](https://github.com/shadps4-emu/shadPS4/pull/4471) has been tried before and [there were still issues](https://github.com/shadps4-emu/shadPS4/pull/4476) (IIRC for the revert, it was causing weird colors in the Persona 5 intro video). So I'm a bit wary of doing this again.

src/shader_recompiler/specialization.h
cuesta4: It was needed for the discarded image-format experiment. I removed it.

squidbus:
Seems alright, would like if it can get some testing though.

cuesta4:
This approach differs from #4293 and #4471 by not using the aggregate vertex-export count to truncate or pad the first N fragment inputs. That approach loses location identity when vertex exports are sparse or fragment inputs are remapped, forwarding the wrong values even when the total interface counts match. Instead, it builds a mask from the actual parameter locations written by the vertex shader, as recorded in Info::stores. Each non-default fragment input is matched through its param_index, including the location shift required by clip-distance emulation. The generated TCS declares an input only when that exact VS output location exists, while still producing every output required by the TCS > TES > FS interface chain and zero-filling only genuinely missing locations. This preserves the identity of sparse and remapped exports and limits the change to the generated auxiliary TCS/TES path.

PR #4617, opened 2026-06-25
fix: FBSD 15 libc++ quirks with includes

xinitrcn1:
generally using `std::ranges` without either `<ranges>` and `<algorithm>` ought to trip up atleast one libc++...
such libc++ is called FBSD15 default clang libc++.

### DIFF
--- a/src/common/cstring.h
+++ b/src/common/cstring.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <algorithm>
+#include <ranges>
 #include <string_view>
 
 #include "assert.h"
--- a/src/core/file_format/psf.cpp
+++ b/src/core/file_format/psf.cpp
@@ -1,7 +1,9 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include <algorithm>
 #include <cstring>
+#include <ranges>
 
 #include "common/assert.h"
 #include "common/io_file.h"
--- a/src/core/libraries/network/net_epoll.cpp
+++ b/src/core/libraries/network/net_epoll.cpp
@@ -1,6 +1,9 @@
 // SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include <algorithm>
+#include <ranges>
+
 #include "common/assert.h"
 #include "common/types.h"
 #include "net_epoll.h"
--- a/src/core/libraries/network/net_upnp.cpp
+++ b/src/core/libraries/network/net_upnp.cpp
@@ -12,6 +12,7 @@
 #include <ws2tcpip.h>
 #else
 #include <arpa/inet.h>
+#include <net/if.h>
 #endif
 
 namespace Libraries::Net {
--- a/src/core/libraries/np/object_manager.h
+++ b/src/core/libraries/np/object_manager.h
@@ -1,8 +1,10 @@
 // SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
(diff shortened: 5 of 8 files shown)


CMakeLists.txt
```diff
@@ -266,6 +266,15 @@ if (ENABLE_SYSTEM_LIBRARIES)
     list(POP_BACK CMAKE_MODULE_PATH)
 endif()
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+    # libc++ requires -fexperimental-library to enable std::jthread and std::stop_token support.
+    include(CheckCXXSymbolExists)
+    check_cxx_symbol_exists(_LIBCPP_VERSION version LIBCPP)
+    if(LIBCPP)
+        add_compile_options(-fexperimental-library)
```
Niram7777: the cleanest way would be to use llvm-devel (because version 19 is too old, and  this lib was finished in 20)
https://libcxx.llvm.org/Status/Cxx20.html#cxx20-status
https://github.com/RPCS3/rpcs3/pull/17899/changes
(we should update requirements and doc which I am planning when I update LLVM on CI)

Niram7777 (approved):
I approve because CMake can be cleaned in the future

CMakeLists.txt
xinitrcn1: most FreeBSD users would expect things to work with default clang
it's not really affecting any other platform (bar a potential modern macOS with appleclang)

CMakeLists.txt
```diff
@@ -266,6 +266,15 @@ if (ENABLE_SYSTEM_LIBRARIES)
     list(POP_BACK CMAKE_MODULE_PATH)
 endif()
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
```
squidbus: If you're going to re-add this please at least limit it to FreeBSD, or based on libc++ version.

georgemoralis:
clang should be fixed first

xinitrcn1:
clang fixed, will deal with cmake shenanigans in another pr

CMakeLists.txt
Niram7777: we shouldn't lock to llvm 19 just because people dont want to use a recent compiler.

Niram7777:
> clang fixed, will deal with cmake shenanigans in another pr

actually even recent libcxx doesnt have chrono fully implemented (std::chrono::current_zone).
I suggest you test with
```cmake
option(ENABLE_LIBCXX_EXPERIMENTAL_LIBRARY "Enable more complete libc++" OFF)

if (ENABLE_LIBCXX_EXPERIMENTAL_LIBRARY)
    add_compile_options(-fexperimental-library)
    add_link_options(-fexperimental-library)
endif ()
```
(+ ENABLE_LIBCXX_EXPERIMENTAL_LIBRARY=ON)
it works on my side, if it does on your side you can create the PR (or I can)

or
```cmake
option(ENABLE_LIBCXX_USE "Enable libc++ usage" OFF)

if (ENABLE_LIBCXX_USE)
    add_compile_options(-stdlib=libc++ -fexperimental-library)
    add_link_options(-stdlib=libc++ -lc++abi -fexperimental-library)
endif ()
```
(+ ENABLE_LIBCXX_USE=ON)

squidbus:
> actually even recent libcxx doesnt have chrono fully implemented (std::chrono::current_zone).

We already use the date library as a substitute for this.

Niram7777:
> > actually even recent libcxx doesnt have chrono fully implemented (std::chrono::current_zone).
> 
> We already use the date library as a substitute for this.

for Apple and Freebsd yes, but not for Linux

squidbus:
> for Apple and Freebsd yes, but not for Linux

Then just change it to enable date library if libc++ is detected, if that’s actually needed. I’d rather that than fiddling with a partially done version.

PR #4619, opened 2026-06-25
Memory Patch Fixes

notanenergydrinkaddict:
There's a lot of local string copies but I didn't bother fixing them

### DIFF
--- a/src/common/memory_patcher.cpp
+++ b/src/common/memory_patcher.cpp
@@ -184,9 +184,18 @@ void ApplyPatchesFromXML(std::filesystem::path path) {
                             maskOffsetValue = std::stoi(maskOffsetStr, 0, 10);
                         }
 
-                        MemoryPatcher::PatchMemory(currentPatchName, address, patchValue, targetStr,
-                                                   sizeStr, false, littleEndian, patchMask,
-                                                   maskOffsetValue);
+                        const patchInfo patch = {
+                            .gameSerial = "*",
+                            .modNameStr = currentPatchName,
+                            .offsetStr = address,
+                            .valueStr = patchValue,
+                            .targetStr = targetStr,
+                            .sizeStr = sizeStr,
+                            .littleEndian = littleEndian,
+                            .patchMask = patchMask,
+                            .maskOffset = maskOffsetValue,
+                        };
+                        MemoryPatcher::PatchMemory(patch);
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


src/common/memory_patcher.cpp
```diff
@@ -30,20 +30,24 @@ std::string toHex(u64 value, size_t byteSize) {
     return ss.str();
 }
 
+static int isHex(const std::string& s) {
+    return (s.size() >= 3 && s[0] == '0' && (s[1] == 'x' || s[1] == 'X')) ? 16 : 10;
+}
```
kalaposfos13, line 45: This does not follow the spec from https://github.com/illusionyy/PS-Game-Patch, and will break a lot of patches from both there and from the shadPS4 patches repo as well. If you have a patch that uses decimal numbers, then it is the patch that is wrong, not everyone else.

kalaposfos13 (dismissed):

src/common/memory_patcher.cpp
notanenergydrinkaddict: Are you sure? This patch uses decimal https://github.com/illusionyy/PS-Game-Patch/blob/46a0b10b60247d7405bf8902cf6ea8f3502cf1ae/patches/xml/ArmoredCoreVI-Orbis.xml#L41-L42

kalaposfos13: Huh, it appears you're correct, and it's only "bytes" that is unconditionally hex, my bad.

georgemoralis:
@kalaposfos13 you approve it then?

georgemoralis:
needs clang fix anyway

notanenergydrinkaddict:
> needs clang fix anyway

fixed

src/common/memory_patcher.cpp
```diff
@@ -30,20 +30,33 @@ std::string toHex(u64 value, size_t byteSize) {
...
+static bool isHexSym(const std::string& s) {
+    return (s.size() >= 2 && (s[0] == '$' || s[0] == '#'));
+}
+
+static bool isHex0x(const std::string& s) {
+    return (s.size() >= 3 && s[0] == '0' && (s[1] == 'x' || s[1] == 'X'));
+}
+
+// possible prefixes == (s == `0x` || s == `#` || s == `$`)
+static int isHex(const std::string& s) {
```
kalaposfos13: I guess one other nitpick could be that the name of this function is misleading, as isHex implies it'll return a bool, while it returns the number's number system itself, so a better name would be NumberBaseOf or something.

kalaposfos13 (approved):

src/common/memory_patcher.cpp
notanenergydrinkaddict: changed to `convertNumBase`

PR #4621, opened 2026-06-25
cmake: enable system protobuf library

abouvier:


georgemoralis:
i am a bit worried about this since we can' guarantee that shadnet and shadps4 will run the same version

abouvier:
Is this really important if the `shadnet.proto` schema stay the same?

thethunderturner:
> Is this really important if the `shadnet.proto` schema stay the same?

Yeah I agree with shadow. shadnet.proto must be identical in both shadps4 (emulator) and shadnet

metr1k-91:
It is extremely important that proto stay lined up on both ends, yes.

PR #4624, opened 2026-06-26
Core: code patches for cmp and xor instructions involving the TCB

StevenMiller123:
This PR adds code patches for `cmp reg, fs:[0]` and `xor reg, fs:[0]` instructions, where reg is a typical 64bit register. The former is encountered in PeacockTV (now that the memory assert is fixed), while the latter is seen in updated versions of Roblox.

Needs review before merge, I would like a second eye on these changes since I don't typically mess with code patches.


src/core/cpu_patches.cpp
```diff
@@ -127,6 +133,70 @@ static void GenerateTcbAccess(void* /* address */, const ZydisDecodedOperand* op
...
+
+    // Set rsp to before red zone and save scratch register
+    c.lea(rsp, ptr[rsp - 128]);
+    c.pushfq();
+    c.push(scratch);
+
+    // Retrieve value from TCB and store it in the scratch register
+    RetrieveTcbPointer(scratch, c);
+
+    // Preform compare op
```
squidbus:
```suggestion
    // Perform compare op
```

src/core/cpu_patches.cpp
```diff
@@ -127,6 +133,70 @@ static void GenerateTcbAccess(void* /* address */, const ZydisDecodedOperand* op
...
+
+    // Set rsp to before red zone and save scratch register
+    c.lea(rsp, ptr[rsp - 128]);
+    c.pushfq();
+    c.push(scratch);
+
+    // Retrieve value from TCB and store it in the scratch register
+    RetrieveTcbPointer(scratch, c);
+
+    // Preform xor
```
squidbus:
```suggestion
    // Perform xor
```

squidbus (commented):
Looks okay other than two small typos.

PR #4627, opened 2026-06-26
Reapply #4619 and handle patch value hex exceptions

notanenergydrinkaddict:
Needs to be checked by bb fanatics, also better to fix the actual patches. report any that are broken from the log please!

### DIFF
--- a/src/common/memory_patcher.cpp
+++ b/src/common/memory_patcher.cpp
@@ -30,20 +30,34 @@ std::string toHex(u64 value, size_t byteSize) {
     return ss.str();
 }
 
+static bool isHexSym(const std::string& s) {
+    return (s.size() >= 2 && (s[0] == '$' || s[0] == '#'));
+}
+
+static bool isHex0x(const std::string& s) {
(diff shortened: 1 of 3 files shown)


kalaposfos13:
In my opinion, silently skipping patches here is worse than asserting on such occurrences, I'd have put an unreachable in the catch block myself, with a log clearly saying that the issue lies with the patch, not the user or the emulator, as most people don't check logs for silent failures, only for crashes, although for Bloodborne specifically, it's likely enough to have only one person knowledgeable about patches go through the lists of patches and fix them manually.

notanenergydrinkaddict:
Made it unreachable

kalaposfos13:
Well then looks good, one thing that might still be worth adding is explicitly mentioning the change in backward compatibility, so users don't just see a cryptic error message, but also know why it appeared (something like "If the patch was working on earlier versions, then it was using a format that shadPS4 handled incorrectly, and shouldn't have worked in the first place"), but aside from that, if you undraft it, I think it's fine to merge.

notanenergydrinkaddict:
Updated to suggesting the user to fix the patch. since this code is gonna be reached if parsing fails, so I think it should always be an assertion instead.

PR #4635, opened 2026-06-27
ShadNET hotfix

UltraDaCat:
Makes shadNET stuff work with custom configs.

### DIFF
--- a/src/core/emulator_settings.cpp
+++ b/src/core/emulator_settings.cpp
@@ -513,10 +513,6 @@ bool EmulatorSettingsImpl::TransferSettings() {
         setFromToml(s.show_splash, general, "showSplash");
         setFromToml(s.trophy_notification_side, general, "sideTrophy");
         setFromToml(s.connected_to_network, general, "isConnectedToNetwork");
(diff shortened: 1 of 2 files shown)


thethunderturner:
I think the team is planning on deprecating TOML support, and just sticking with JSON

src/core/emulator_settings.cpp
```diff
@@ -513,6 +513,10 @@ bool EmulatorSettingsImpl::TransferSettings() {
         setFromToml(s.show_splash, general, "showSplash");
         setFromToml(s.trophy_notification_side, general, "sideTrophy");
         setFromToml(s.connected_to_network, general, "isConnectedToNetwork");
+        setFromToml(s.shad_net_enabled, general, "IsShadNetEnabled");
```
StevenMiller123: What's the point in this? Most of these options were never added to the old toml configs

src/core/emulator_settings.cpp
georgemoralis, line 1: these didn't exist pre 0.15.0 so we don't need to

StevenMiller123:
Clang failed 😅

PR #4636, opened 2026-06-27
Push events for webapi and some other functions fixes

georgemoralis:
Tested with shadnet webapi and seems functional


src/core/libraries/np/np_web_api/np_web_api.cpp
```diff
@@ -292,26 +325,27 @@ s32 PS4_SYSV_ABI sceNpWebApiGetConnectionStats(s32 userCtxId, const char* pApiGr
 }
 
 s32 PS4_SYSV_ABI sceNpWebApiGetErrorCode() {
-    LOG_ERROR(Lib_NpWebApi, "(STUBBED) called");
-    return ORBIS_OK;
+    const s32 code = getLastWebApiError();
```
StevenMiller123, line 328: Based on what I've decompiled, I don't think this is accurate. I'll look into what the call looks like since I've got a game using it.

StevenMiller123 (commented):
Most of this looks fine, I had a feeling that notification callback was important given it seemed to be the only thing interacting with those filters 😅 
The sceNpWebApiGetErrorCode implementation doesn't look right though.

src/core/libraries/np/np_web_api/np_web_api.cpp
georgemoralis: The real lib sources sceNpWebApiGetErrorCode() from sceNpServerErrorJson* global state as we don't have that subsystem, i capture it from the error body the gamereads via sceNpWebApiReadData. Check captureWebApiError as well it is decompiled code

StevenMiller123: From what I can tell it isn't some global state, there are parameters that get passed from sceNpWebApiGetErrorCode to those sceNpServerErrorJson functions. Ghidra just doesn't detect them because the parameters are passed straight along.

PR #4639, opened 2026-06-27
fix(vk): reduce staging buffer from 512_MiB to 128_MiB so specific NVIDIA drivers don't immediately OOM

xinitrcn1:
same-ish issue observed on Eden and Citron NEO with increased staging buffers
basically *nix NVIDIA drivers dont like big chunks of allocations >128MiB
so they oom (VK_ERROR_OUT_OF_MEMORY), usually

NVIDIA windows drivers seem to be fine of this through. But I haven't checked on shadPS4 proper.
an #ifdef would've been ugly and I don't know if it's entirely *nix specific

as always with everything NVIDIA the reason why this is OOM'ing when there is clearly memory available, is a mystery

drivers that oom:
```
nvidia-driver-580-580.142
nvidia-drm-66-kmod-580-580.142.1500068_2
nvidia-drm-kmod-580-580.142
nvidia-kmod-580-580.142.1500068
nvidia-xconfig-595.58.03
```

uname:
`FreeBSD xerix 15.0-RELEASE FreeBSD 15.0-RELEASE releng/15.0-n280995-7aedc8de6446 GENERIC amd64`

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -17,7 +17,7 @@
 namespace VideoCore {
 
 static constexpr size_t DataShareBufferSize = 64_KB;
-static constexpr size_t StagingBufferSize = 512_MB;
+static constexpr size_t StagingBufferSize = 128_MB;
 static constexpr size_t DownloadBufferSize = 32_MB;
 static constexpr size_t UboStreamBufferSize = 64_MB;
 static constexpr size_t DeviceBufferSize = 128_MB;


src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -17,7 +17,7 @@
 namespace VideoCore {
 
 static constexpr size_t DataShareBufferSize = 64_KB;
-static constexpr size_t StagingBufferSize = 512_MB;
+static constexpr size_t StagingBufferSize = 128_MB;
```
squidbus, line 20: If this is an NVIDIA *nix problem it should probably be reduced for only NVIDIA *nix.

xinitrcn1: Thing is, I don't know if it happens for Windows as well (especially since shadPS4 is vastly different, allocations-wise)

StevenMiller123: Most of our devs are Nvidia GPU owners running Windows. If there were issues caused by the staging buffer size for that setup, I think it would've been addressed by now.

squidbus: Do we even have reports of this on Linux or is it FreeBSD only?

StevenMiller123: @xinitrcn1 Do you have specific steps for reproducing the out of memory issues?

xinitrcn1: Yes.
1. open shadPS4 on freebsd
2. have a beautiful gpu like 1030
3. instant oom, always, no matter what you open, you could open "hello world"homebrew and it would still oom

...step 2 is a must

StevenMiller123: As a 750 Ti owner, I can pretty safely say this doesn't affect Linux.

raphaelthegreat:
This has been solved

PR #4646, opened 2026-06-29
Lib.NpWebApi2: Basic implementation of request logic

StevenMiller123:
This PR implements enough libSceNpWebApi2 logic to get requests sent to shadNet. As is, most games won't behave because of missing shadNet end points, so I figure getting this part out is a good start.

libSceNpWebApi2 relies on libSceHttp2. In order to work around that without injecting hacks into my code, I have basic libSceHttp2 calls routing to our libSceHttp implementation. Since this library only hits basic synchronous operations (and our libSceHttp implementation has very poor error checking), this is safe enough to do.

In the future, I will probably look into implementing libSceHttp2 properly. We can't emulate the library's async calls using our libSceHttp implementation, so a proper, separate implementation is warranted. We can probably keep using cpp-httplib for it though.


StevenMiller123:
I'll reopen this once I've finished implementing everything.

PR #4648, opened 2026-06-30
Lib.AvPlayer: Fixes to decoder loops

StevenMiller123:
This PR fixes some deadlock issues I've noticed while looking through our AvPlayer code.

Firstly, I've adjusted the sync logic to actually check if the audio decoder thread is active. In games where video would sync to audio, if the audio decoder stopped producing frames, threads calling sceAvPlayerGetVideoData would no longer pop frames and push buffers, and the video decoder would deadlock. On real hardware, such videos continue to completion.

Secondly, I've stopped the decoders from waiting when there are still valid packets or buffers to decode. I don't think this should happen, but I don't know our AvPlayer code well enough to know for certain.

Finally, if the decoder or demuxer threads were to exit on their own, their Kernel::Thread object would still be marked as Joinable. This means that if the AvPlayer decoding and demuxer threads all exit, sceAvPlayerIsActive would still return true. To fix this, I've made the threads attempt to join with themselves, which sets the actual pthread object to null and resolves the issue.

There's still more deadlock potential in these decoders, but this does partially fix the regression in Batman - Return to Arkham games, where they would hang part-way through videos since the audio decoder exited naturally. Unfortunately, they're still prone to hanging when trying to exit AvPlayer, due to some internal race involving their code for exiting AvPlayer playback.

This PR needs a proper review before merging. I haven't spotted any regressions on my end, but I would like some testers to check other games (perhaps some of the recently regressed ones too?). It would also be a good idea to have a reviewer take a look since this is mostly based on game behavior, not actual decompilation.

### DIFF
--- a/src/core/libraries/avplayer/avplayer_source.cpp
+++ b/src/core/libraries/avplayer/avplayer_source.cpp
@@ -361,7 +361,8 @@ bool AvPlayerSource::GetVideoData(AvPlayerFrameInfoEx& video_info) {
 
     const auto& new_frame = m_video_frames.Front();
     if (m_state.GetSyncMode() == AvPlayerAvSyncMode::Default) {
-        if (m_audio_stream_index) {
+        if (m_audio_stream_index && m_audio_decoder_thread.Joinable()) {
+            // Audio is available, sync video with it.
             if (new_frame.info.timestamp > m_last_audio_ts.value_or(0)) {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


StevenMiller123:
Might be the longest description I've written for a PR with as low as 28 changed lines 😅

roamic (approved):
LGTM

Parotaku:
Doesn't seem to change anything here in games with problematic videos (Snoopy, Battle Crashers, No straight roads...)...
Don't know if that's a good thing or a bad thing...! ^^;

Missake212:
Gravity Rush™ Remastered doesn't hang anymore and gets back in game:

[CUSA01113.log](https://github.com/user-attachments/files/29513548/CUSA01113.log)

Matterfall™ behaves exactly the same, the video flickers to black and green a lot and it still hangs before reaching the main menu:

[CUSA02475.log](https://github.com/user-attachments/files/29513569/CUSA02475.log)

(If you need my testing on one of your PR in the future feel free to @ me on it)

StevenMiller123:
Thanks for all the tests. I'll merge this for now as it seems there's no new regressions.
I've got other AvPlayer-related issues I'm still debugging. If I come up with more bug fixes, I'll make a second PR.

PR #4649, opened 2026-06-30
ShadNet:Presence support

georgemoralis:
This is shadps4 of presense support for shadnet . Not yet working since it needs shadnet side (coming soon)


georgemoralis:
for use with 0.0.8 shadnet (updated)

PR #4650, opened 2026-06-30
Matching2 fixup 2, I'll take all the buffer I want.

metr1k-91:
Fixed what appeared to be a overflow caused by context movement from terminate calls.


georgemoralis:
seems to work fine in xenoverse2!

PR #4653, opened 2026-07-01
Invitation Dialog

georgemoralis:
WIP


georgemoralis:
Some progress on the UI part of it , still not sending events

εικόνα

StevenMiller123 (approved):
Seems to work pretty well in games I'm set up to test, aside from some potential server-side bugs.
Only nitpick is that there's seemingly no way to accept invites on a purely controller-based setup.

PR #4655, opened 2026-07-01
shader_recompiler: fix SRT-walker fault handler (boot crash + transient non-residency)

Andrew401428:
Two fixes to `SrtWalkerSignalHandler` (`src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp`) for titles that lean heavily on the JIT'd SRT walker — Bluepoint / FromSoft engines such as Shadow of the Colossus, Bloodborne and Demon's Souls.

### 1. Code-range bug → boot crash on ~half of boots

The handler decides whether a fault came from the JIT'd walker using `g_srt_codegen_start`, which is set to the address of the **first flatten-pass walker**. Walkers emitted earlier via `RegisterWalkerCode` live **below** that address in the same codegen buffer, so their faulting RIPs fall under `code_start`, fail the range check, and the fault goes unhandled → segfault on roughly half of boots for SRT-heavy titles.

Fix: bound the range by the codegen buffer itself — `g_srt_codegen.getCode()` … `getCode() + getSize()`, which contains only walker code.

With the wider range the handler can also now see walker faults whose instruction isn't the simple `MOV rdi/r10d, [mem]` load it handles. Those previously hit `ASSERT` / `UNREACHABLE_MSG` and aborted the process; they now `return false` (decline) and let the fault propagate to normal handling.

### 2. Page-in-and-retry before zeroing the load

On a walker load fault the old code **immediately** rewrote the load to `xor reg, reg`, baking a zero pointer into the walker. These walkers chase guest pointers into per-draw scratch / descriptor tables that are frequently only **transiently** non-resident at scene start, so the zero is fed to the geometry as a null descriptor pointer (e.g. a null material `T#`), contributing to a black/untextured world even after the memory becomes valid.

Fix: if the faulting address is a valid mapping, make its page resident (`AddressSpace::Protect(…, ReadWrite)`) and **re-execute** the load (return `true` without advancing RIP) — the same self-heal `GuestFaultSignalHandler` already performs for orphaned `NoAccess` protections. A bounded per-address `thread_local` retry guards against an infinite fault loop on a genuinely-unmapped pointer, in which case the handler falls back to the original zero-the-load behavior.

### Scope / notes

- One file, `+56 / −7`, no new APIs — uses existing `Core::Memory` / `AddressSpace::Protect` / `Common::AlignDown` / `Common::GetRip`.
- Compile-verified against current `main` (single-TU clang-cl build, exit 0).
- These are stability fixes only; they do **not** by themselves make SOTC fully render (its virtual-texture material system is a separate, larger issue). Fix 1 makes SRT-heavy titles boot reliably instead of crashing ~50% of the time; Fix 2 stops a class of transient-null-descriptor black-material artifacts.

### DIFF
--- a/src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
+++ b/src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
@@ -5,13 +5,16 @@
 #include <boost/container/flat_map.hpp>
 #include <xbyak/xbyak.h>
 #include <xbyak/xbyak_util.h>
+#include "common/alignment.h"
 #include "common/arch.h"
 #include "common/decoder.h"
 #include "common/io_file.h"
 #include "common/logging/log.h"
 #include "common/path_util.h"
 #include "common/signal_context.h"
+#include "core/address_space.h"
 #include "core/emulator_settings.h"
+#include "core/memory.h"
 #include "core/signals.h"
 #include "shader_recompiler/info.h"
 #include "shader_recompiler/ir/breadth_first_search.h"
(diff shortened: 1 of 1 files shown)


georgemoralis:
First of all we try not to depend 100% on AI tools . Stop posting what claude told you and at least provide a human explanation of what you are trying to fix and if the applied fix really fixed something particular.  Give us the real issue and what that pr fixed if it fixed something

Andrew401428:
Hey so I made this very much last second right before I was leaving for a while so I rly didn’t even look over it. I was more or less just curious to see if someone was even looking at SOTC. I also get not using AI 100% but I  was being incredibly straightforward with it and said if you don’t like it I’ll keep working on my own. It seems I will keep working on my own. I don’t rly use GitHub so idk.

PR #4662, opened 2026-07-02
Render.Recompiler: V_LOG_CLAMP_F32

StevenMiller123:
Like the other clamp ops we've already got, I'm just ignoring the clamp here and using our V_LOG_F32 for this. Let me know if that's not fine.

Should fix a bug in F1® 2021 that @UltraDaCat reported


Niram7777 (approved):
so I guess we have to clamp everywhere like that at some point:
https://llvm.org/docs/AMDGPUModifierSyntax.html#amdgpu-synid-clamp

UltraDaCat:
F1 2021 gets ingame now, tends to be crashy
image
image
image
image

Log:
[shad_log.txt](https://github.com/user-attachments/files/29627957/shad_log.txt)

PR #4663, opened 2026-07-03
Implemented MultipartPart content for npwebapi

georgemoralis:
Done using decompile , not tested though


StevenMiller123 (approved):
Been a while since I looked at this lib, but this does seem correct to me.

PR #4664, opened 2026-07-03
macOS: Enable stack-switching thread helper compilation on ARM64

burhanmoin1:
### Summary
This PR enables compiling the emulator's native stack-switching assembly thread helper (`stack.S`) on ARM64 architectures (such as Apple Silicon macOS).

### Details
1. **CMake Configuration (`CMakeLists.txt`):**
   * **Problem:** Compiling the emulator on ARM64 macOS would fail because the stack-switching assembly file (`stack.S`) was omitted from the build compilation list (which was hardcoded to `x86_64` architecture only).
   * **Fix:** Updated the architecture check in `CMakeLists.txt` to include `arm64` and append `stack.S` for compilation.

2. **Assembly and Thread Helpers (`pthread.cpp` and `stack.S`):**
   * **Problem:** `pthread.cpp` declares `_runOnAnotherStack` only for `ARCH_X86_64`, which fails to link on ARM64.
   * **Fix:** Added compiler checks for `__arm64__` / `__aarch64__` in `pthread.cpp` and implemented the corresponding ARM64 assembly routine for `_runOnAnotherStack` in `stack.S`.

### DIFF
--- a/src/core/libraries/kernel/threads/stack.S
+++ b/src/core/libraries/kernel/threads/stack.S
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright 2026 shadPS4 Emulator Project
 # SPDX-License-Identifier: GPL-2.0-or-later
 
+#if defined(__x86_64__)
+
 .global _runOnAnotherStack
 _runOnAnotherStack:
   pushq %r12
@@ -35,3 +37,41 @@ _runOnAnotherStack:
   popq %r13
   popq %r12
   ret
+
+#elif defined(__arm64__) || defined(__aarch64__)
+
+.global _runOnAnotherStack
+_runOnAnotherStack:
+  // Save frame pointer (x29), link register (x30), and callee-saved registers x19, x20
+  stp x29, x30, [sp, #-32]!
+  mov x29, sp
+  stp x19, x20, [sp, #16]
+
+  // Save the current stack pointer into x19 (callee-saved)
+  mov x19, sp
+  // Save the current frame pointer into x20 (callee-saved)
+  mov x20, x29
+
+  // Align the target stack pointer (x2) to 16 bytes (ARM64 constraint)
+  and x2, x2, #-16
+
+  // Switch stack pointer to the new stack (x2)
+  mov sp, x2
+  // Switch frame pointer to the new stack (x2)
+  mov x29, x2
+
+  // Call the function (target address is in x1)
+  // Note: the argument is already in x0!
+  blr x1
+
+  // Restore the old stack pointer and frame pointer
+  mov sp, x19
+  mov x29, x20
+
+  // Restore saved registers and return
+  ldp x19, x20, [sp, #16]
+  ldp x29, x30, [sp], #32
+  ret
+
+#endif
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -363,9 +363,9 @@ set(KERNEL_LIB src/core/libraries/kernel/coredump/coredump.cpp
                src/core/libraries/kernel/aio.h
 )
 
-if (ARCHITECTURE STREQUAL "x86_64")
+if (ARCHITECTURE STREQUAL "x86_64" OR ARCHITECTURE STREQUAL "arm64")
     list(APPEND KERNEL_LIB src/core/libraries/kernel/threads/stack.S)
-    set_source_files_properties(src/core/libraries/kernel/threads/stack.s PROPERTIES COMPILE_OPTIONS -Wno-unused-command-line-argument)
+    set_source_files_properties(src/core/libraries/kernel/threads/stack.S PROPERTIES COMPILE_OPTIONS -Wno-unused-command-line-argument)
 endif()
 
 set(NETWORK_LIBS src/core/libraries/network/http.cpp
--- a/src/core/libraries/kernel/threads/pthread.cpp
+++ b/src/core/libraries/kernel/threads/pthread.cpp
@@ -12,7 +12,7 @@
 #include "core/libraries/libs.h"
 #include "core/memory.h"
 
-#ifdef ARCH_X86_64
+#if defined(ARCH_X86_64) || defined(__arm64__) || defined(__aarch64__)
 extern "C" void* PS4_SYSV_ABI _runOnAnotherStack(void* arg, void* func,
                                                  void* stackb) asm("_runOnAnotherStack");
 #else


src/core/libraries/kernel/threads/stack.S
```diff
@@ -1,5 +1,4 @@
-# SPDX-FileCopyrightText: Copyright 2026 shadPS4 Emulator Project
-# SPDX-License-Identifier: GPL-2.0-or-later
```
notanenergydrinkaddict, line 2: Why is this removed

burhanmoin1: Sorry about that! The headers were accidentally removed in a previous commit. I've pushed a commit restoring them, and the PR has been updated.

notanenergydrinkaddict: Thank you

notanenergydrinkaddict (approved):

src/core/cpu_patches.cpp
```diff
@@ -881,9 +881,9 @@ void RegisterPatchModule(void* module_ptr, u64 module_size, void* trampoline_are
 }
 
 void PrePatchInstructions(u64 segment_addr, u64 segment_size) {
-#if !defined(_WIN32) && !defined(__APPLE__)
-    // Linux and others have an FS segment pointing to valid memory, so continue to do full
-    // ahead-of-time patching for now until a better solution is worked out.
```
Niram7777, line 886: the comment is still true for Linux at least I believe

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -260,17 +264,39 @@ vk::UniqueInstance CreateInstance(Frontend::WindowSystemType window_type, bool e
...
     setenv("VK_DRIVER_FILES", icd_path.c_str(), true);
 #endif
 
+#if defined(__APPLE__)
+    static const auto mvk_path = [] {
+        char path[PATH_MAX];
+        u32 length = PATH_MAX;
+        _NSGetExecutablePath(path, &length);
+        auto exe_dir = std::filesystem::path(path).parent_path();
+        return (exe_dir / "libMoltenVK.dylib").string();
```
Niram7777: I dont believe we support Molten anymore, only KosmicKrisp
@squidbus ?

Niram7777 (changes requested):

src/core/cpu_patches.cpp
squidbus: Yes, the Linux info should stay. Also while there may be a point with caching efficiency, in my experience it hasn’t made a huge difference since it’s a one time cost for each patch location. I don’t really want to rely on the AOT patching if we don’t have to since it scans the binary without any awareness of code vs data, and thus is not totally reliable.

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -196,6 +196,10 @@ std::vector<const char*> GetInstanceExtensions(Frontend::WindowSystemType window
         extensions.push_back(VK_KHR_SURFACE_EXTENSION_NAME);
     }
 
+#if defined(__APPLE__)
+    extensions.push_back(VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME);
+#endif
```
squidbus: Portability subset is not supported, we use KosmicKrisp now.

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -260,17 +264,39 @@ vk::UniqueInstance CreateInstance(Frontend::WindowSystemType window_type, bool e
 
 #if defined(__APPLE__)
     // Initialize the environment with the path to the included ICD, so that the loader will
-    // find it.
+    // find it. VK_DRIVER_FILES requires the full path to the JSON file, not a directory.
```
squidbus: No it does not.

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -260,17 +264,39 @@ vk::UniqueInstance CreateInstance(Frontend::WindowSystemType window_type, bool e
...
     static const auto icd_path = [] {
         char path[PATH_MAX];
         u32 length = PATH_MAX;
         _NSGetExecutablePath(path, &length);
-        return std::filesystem::path(path).parent_path();
+        // First try: bundled MoltenVK_icd.json next to the executable
```
squidbus: We do not use MoltenVK anymore, nor do we want to use the Homebrew installed lib anyway.

src/video_core/renderer_vulkan/vk_platform.cpp
squidbus: Correct, we only support KosmicKrisp, which is loaded via libvulkan.

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -399,8 +425,14 @@ vk::UniqueInstance CreateInstance(Frontend::WindowSystemType window_type, bool e
         },
     };
 
+    vk::InstanceCreateFlags flags;
+#if defined(__APPLE__)
+    flags |= vk::InstanceCreateFlagBits::eEnumeratePortabilityKHR;
```
squidbus: Again, not supported anymore.

src/sdl_window_apple.mm
```diff
@@ -10,39 +10,7 @@
...
-        [maskPath addClip];
-        [baseIcon drawInRect:bounds
-                    fromRect:NSZeroRect
-                   operation:NSCompositingOperationSourceOver
-                    fraction:1.0f];
-        [nativeIcon unlockFocus];
-
-        [NSApp setApplicationIconImage:nativeIcon];
-    }
+    // No-op on macOS to prevent symbol conflicts between Homebrew libpng and system ImageIO/AppKit
```
squidbus: This is entirely for macOS… what symbol conflicts are you even talking about, this is normal Objective-C using macOS APIs.

squidbus (commented):
None of the Vulkan changes in here look valid to me, we don’t support MoltenVK anymore and need to use libvulkan. For macOS we include both libvulkan and KosmicKrisp as submodules and bundle them in with the output executable, so there’s no situation of the user not having a compatible version.

src/sdl_window_apple.mm
burhanmoin1: Hi @squidbus and @Niram7777,

Thank you so much for the feedback. I completely understand and agree with your points.

To be clear, I originally created this fork just to see if I could experiment and run Bloodborne locally on my M1 Pro Mac. I added the AOT JIT patching workaround and custom Vulkan driver loading purely as local hacks to solve my own framerate/startup issues, without fully grasping the broader architecture and safety implications of the emulator (like the unreliability of blind AOT memory scanning or the shift to KosmicKrisp).

In my haste to submit a clean open-source contribution for compilation support, I incorrectly grouped those experimental tweaks into the same branch.

I have completely cleaned up the branch and force-pushed a clean update. I reverted the Vulkan, JIT, and icon changes back to upstream master. The PR now only contains the structural ARM64 assembly compilation fixes (CMakeLists.txt, pthread.cpp, and stack.S) which enable native building on Apple Silicon.

Moving forward, I will be much more mindful of the overall architecture and safety constraints of the codebase. Thank you both for your patience and for teaching me!

squidbus (approved):
I’m unsure how much utility implementing the assembly runOnStack would have with an actual ARM64 JIT in place for native ARM builds, but I suppose the implementation looks okay.

Niram7777 (approved):

PR #4669, opened 2026-07-04
vdec: fix flush and decode

roamic:
* Flush now sends a flush packet to the ffmpeg codec
* Removed the setting of `thisSize` as this should be set by caller
* Removed the setting of `isErrorPic` because this should come from vdec internals that are not investigated yet. Should probably never be set to true when using ffmpeg.
* Removed multi-frame decoding as this can break without error. Now it will cause an API error so we can analyze games that do this.

This PR should fix the videos hanging in Burnout Paradise Remastered.


brad0demx:
image
image
image

Fixes the intro cinematics in burnout

bigol83:
works flawless for me 👌

georgemoralis:
did someone test if it breaks other videos?:P

huntyboy102:
I love you

brad0demx:
> did someone test if it breaks other videos?:P

I tested on a couple other games mainly Call of Duty titles and it seems to be same behaviour as main. Not sure what other games I could test for it though

tochnonechelovek:
No notable changes or regressions in my titles.

PR #4670, opened 2026-07-04
Updated is_pow2 property calculation to include NumLevels check.

Hog185:
Brings tony hawk skater pro 3+4 to menus, and goat simulator 3 to gameplay (barely)


Hog185:
brings sbk 22 in game

georgemoralis:
pow2pad affects the alignment of arrayLayers, Width, Height, Depth, so given the fact that this fix launches games, it means there are problems somewhere in the tiler (I suspect that this will break the textures in Slime Rancher, where almost all the textures are unaligned)  , @red-prig notes credits to him

Hog185:
@georgemoralis so what changes do you want

PR #4675, opened 2026-07-05
shader_recompiler: GCN thread mask accuracy fix (GR2 Grass Fix)

junminlee2004:
Fixes 2 bugs in recompiler code that restore some Tess+Geo pipelines (grass) that are currently absent (produce 0 vertices in shader dumps) and not rendered correctly in GR2. 
The first bug is a gcn thread mask parked in VGPR lanes via `v_writelane` and restored via `v_readlane` losing its per-lane meaning and coming back as `Undef` . The fix is to add a shadow variable that saves the mask's lane-bit view into the shadow so that the mask's meaning travels through the spill instead of being dropped. This restores the rendering of the grass.
The second bug is due to the plain overwrite in V_CMPX. A plain overwrite in EXEC forgets that some lanes were already excluded, unintentionally reviving them, causing garbage data to be served, causing ghost grass to spawn, floating around everywhere in the world. The fix is to make V_CMPX `AND` its result with EXEC before writing EXEC and VCC/SDST, preserving the inactive lanes.

no grass
Before GCN thread mask accuracy fixes
grass
After GCN thread mask accuracy fixes

Example of what "ghost grass" looks like when inactive lanes are not preserved


georgemoralis:
looking good probably , can someone test for regressions? (need for speed rivals had a similar grass issue)

raphaelthegreat:
I don't understand the CMPX bug, can you show a few GCN instructions that present the problem? EXEC state should always be implicitly enforced by control flow. If EXEC = 0 then a compare is not gonna be executed

junminlee2004:
> I don't understand the CMPX bug, can you show a few GCN instructions that present the problem? EXEC state should always be implicitly enforced by control flow. If EXEC = 0 then a compare is not gonna be executed
```

asm
; inside the allocation loop (PC 1288) - s[100:101] is a RUNNING active-mask,
; re-narrowed every iteration; a lane drops out the iteration it exhausts:
s_and_b64       s[100:101], s[100:101], vcc

; after the loop (PCs 4020-4056), the overflow-saturation region:
s_mov_b64       exec, s[100:101]      ; restore loop-participant mask
v_cmpx_ge_u32   vcc, 3, v5            ; v5 = exit status: 3 = broke on pool-full,
                                      ;                   0 = exited by exhaustion
s_cbranch_vccz  +7                    ; skip if NO lane qualifies
...
ds_write_b32    LDS[0] = s93          ; cursor := pool capacity (0x20000)
```

You are right about EXEC but the real bug is the VCC destination: its consumer is a scalar branch, which EXEC doesn't gate, so control flow can't enforce anything there.

junminlee2004:
```
shader disassembly of the grass shader- (cs_0x3f0a3c48) for reference:
asm
/*04d8*/ s_mov_b32   s27, 0                ; scalar trip counter
/*04e0*/ s_mov_b64   s[100:101], exec      ; running mask := loop participants
.L1280:
/*0500*/ v_cmp_lt_u32 vcc, s27, v38        ; counter < this lane's candidate count?
/*0504*/ s_orn2_b64   vcc, vcc, exec       ; don't drop lanes not in exec for this compare
/*0508*/ s_and_b64    s[100:101], s[100:101], vcc
/*050c*/ s_and_b64    exec, exec, vcc      ; exhausted lanes leave BOTH masks
/*0510*/ s_cbranch_execz .L4020            ; exit 1: all lanes exhausted
/*0524*/ ds_add_rtn_u32 v40, v5, v6        ; shared cursor allocation (LDS)
/*0530*/ v_cmp_le_u32 vcc, s93, v40        ; s93 = pool capacity
/*0534*/ v_cndmask_b32 v5, 2, 3, vcc       ; status: 3 = pool full, 2 = allocated
/*053c*/ s_mov_b64    exec, s[100:101]     ; re-widen
/*0540*/ v_cmpx_ge_u32 vcc, 2, v5          ; full lanes leave exec (but stay in s[100:101])
/*0544*/ s_cbranch_vccz .L4020             ; exit 2: nobody allocated -> pool full
         ...body: build blade, stores, saveexec pairs...
/*0f94*/ s_add_i32    s27, s27, 1
/*0fb0*/ s_branch     .L1280
.L4020:
/*0fb4*/ s_mov_b64    exec, s[100:101]     ; the restore in question
/*0fb8*/ v_cmpx_ge_u32 vcc, 3, v5
/*0fbc*/ s_cbranch_vccz .L4060             ; skip the stomp
/*0fd0*/ ds_write_b32 v1, v2               ; LDS[0] := capacity

```

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -112,6 +135,26 @@ void Translator::V_WRITELANE_B32(const GcnInst& inst) {
...
     const IR::U32 old_value{GetSrc(inst.dst[0])};
     ir.SetVectorReg(dst, ir.WriteLane(old_value, value, lane));
+
+    // Shadow the per-lane (U1) view of spilled thread masks so V_READLANE_B32 can restore it
+    // (see the comment there). For non-mask (plain data) sources the shadowed U1 resolves to
+    // Undef, so a later mask-typed read of the restored SGPR behaves as if no shadow existed.
+    if (lane.IsImmediate()) {
+        const u32 key = (inst.dst[0].code << 6) | (lane.U32() & 63u);
+        switch (inst.src[0].field) {
+        case OperandField::ScalarGPR:
```
baggins183: This switch seems redundant since since its repeating logic from GetSrc1. I think you can just use GetSrc1 instead

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -112,6 +135,26 @@ void Translator::V_WRITELANE_B32(const GcnInst& inst) {
     const IR::U32 lane{GetSrc(inst.src[1])};
     const IR::U32 old_value{GetSrc(inst.dst[0])};
     ir.SetVectorReg(dst, ir.WriteLane(old_value, value, lane));
+
+    // Shadow the per-lane (U1) view of spilled thread masks so V_READLANE_B32 can restore it
+    // (see the comment there). For non-mask (plain data) sources the shadowed U1 resolves to
+    // Undef, so a later mask-typed read of the restored SGPR behaves as if no shadow existed.
+    if (lane.IsImmediate()) {
+        const u32 key = (inst.dst[0].code << 6) | (lane.U32() & 63u);
```
baggins183: At an IR level instead of mixing the vector reg dst + the lane into a single arg Id rather have 2 separate operands for that

src/shader_recompiler/frontend/translate/data_share.cpp
junminlee2004: I stripped out some redundant lines by using GetSrc1 but from my  understanding we can't completely remove it because  GetSrc1 asserts on any literal that isn't 0 or -1 right?

baggins183: yeah that makes sense I think how it is now works

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -104,6 +104,26 @@ void Translator::V_READLANE_B32(const GcnInst& inst) {
...
+    // exec restores or store guards fed by it constant-fold to false. Restore the U1 view from
+    // the mask-lane shadow saved by V_WRITELANE_B32. A write to an SGPR architecturally
+    // clobbers both views, so unconditionally redefining the U1 view is strictly more faithful
+    // than leaving it stale.
+    if (lane.IsImmediate()) {
+        switch (inst.dst[0].field) {
+        case OperandField::ScalarGPR:
+        case OperandField::VccLo:
+            SetDst1(inst.dst[0],
+                    ir.GetMaskLaneVariable(IR::VectorReg(inst.src[0].code), lane.U32() & 63u));
```
baggins183: why the & 63?
There are more vector regs than that

Also relevant for the << 6 during SSA

baggins183: nvm im dumb its number of lanes

baggins183: I dont think the & is needed either way, can just take lane as is right?

junminlee2004: I think it's needed to protect the packed key from lane bits bleeding into the vgpr bits but we can just move it into the SSA table's MaskLaneKey and the call sites can take the lane as is.

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -104,6 +104,26 @@ void Translator::V_READLANE_B32(const GcnInst& inst) {
     const IR::U32 value{GetSrc(inst.src[0])};
     const IR::U32 lane{GetSrc(inst.src[1])};
     SetDst(inst.dst[0], ir.ReadLane(value, lane));
+
+    // GCN compilers spill EXEC/VCC/SGPR-pair thread masks into VGPR lanes (v_writelane) and
```
baggins183: Would prefer a more concise comment/in your own words.

My idea would be "This could be a thread mask spilled from scalar to vector register. Additionally store a bool value in case a user of the dst register sources it as a thread mask. Get/SetMaskLaneVariables will be connected by SSA in that case."

Similar opinion on other comments

src/shader_recompiler/frontend/translate/data_share.cpp
baggins183: You mean like the src reg for lane can be an immediate that's >63? I don't actually know, I guess the hardware could clamp or wrap it (don't see the docs describe it), but I don't think it's something that we've considered

baggins183: I would say just add asserts in read/writelane inside the if(immediate) that its < 64

baggins183:
Mask spilling stuff looks fine if acceptable to others in principle

I agree with the other point that the CMPX thing is more a symptom of inaccuracies in control-flow/exec stuff and needs future rework (which will also handle the mask spill part). I'm not brushed up on that to approve that part

src/shader_recompiler/frontend/translate/data_share.cpp
```diff
@@ -104,6 +104,22 @@ void Translator::V_READLANE_B32(const GcnInst& inst) {
     const IR::U32 value{GetSrc(inst.src[0])};
     const IR::U32 lane{GetSrc(inst.src[1])};
     SetDst(inst.dst[0], ir.ReadLane(value, lane));
+
+    // This could be a thread mask spilled from a scalar to a vector register. Additionally
+    // store a bool value in case a user of the dst register sources it as a thread mask;
+    // Get/SetMaskLaneVariable are connected by SSA in that case.
+    if (lane.IsImmediate()) {
+        ASSERT(lane.U32() < 64);
+        switch (inst.dst[0].field) {
```
baggins183: Just thought that we may want to do this (and the writelane part) only for LO registers (VCCLO/EXECLO/even numbered SGPRs). Doing it for odd numbered might screw stuff up.

Will think about it tomorrow

Also for readlane unlike writelane it shouldn't need a switch and can just do a SetDst1. I'm not sure if dst can be EXECLO/HI but even if it could we wouldn't want to ignore EXECLO.

DarkBoy08:
Will this PR affect other games? Not only d Need For Speed Rivals. And also, for example, in The Last Guardian? In theory, it can affect Trico's feathers if you run the game without the readbacks feature. I could be wrong.
P.S for 10.07.26- I also decided to take the feather issue into my own hands and delved into the vector_alu.cpp file under shader_recompiler. I made a couple of code changes, so in theory, the tricot feathers should work without readbacks.

DarkBoy08:
> looking good probably , can someone test for regressions? (need for speed rivals had a similar grass issue)

I tested Need For Speed Rivals and I found out that nothing had changed at all. In theory, the grass should have disappeared from the road, but it didn't, and in fact, it became even more abundant. This is clearly not normal.
Native PR
Need for Speed Rivals Native PR
PR4675
Need For Speed Rivals PR4675

Pipi86:
@DarkBoy08  I was sure i did saw that exact same bug some years ago, so i looked into RPCS3 github issues, and found it. In that case it seems it happens when setting shader compiler to async multithreaded instead of single threaded. https://github.com/RPCS3/rpcs3/issues/13947#issuecomment-1761982428
I'm not sure this PR could fix the misplacement of grass and bushes but from your screenshots it looks to me that with this PR they are properly rendered.

baggins183:
LGTM

Xcedf:
Improves uncharted 3 cloth a lot, still not perfect but better
Before:

After:

PR #4688, opened 2026-07-09
LLE libSceAt9Enc

kalaposfos13:
this trivially LLEable library is used by exactly nothing except Dreams, a homebrew by me in a few hours and maybe some games that have voicechat, but this last one is not relevant for the emulator at the moment

edit: i got skill issued, so no homebrew for y'all


georgemoralis:
forgot to update readme with it

PR #4690, opened 2026-07-09
Fix heap_api checks

StevenMiller123:
Accidentally broke them with #4622


kalaposfos13:
This fixes crashes upon thread creation in certain homebrew apps made with OpenOrbis.

PR #4691, opened 2026-07-10
Improvements on shadnet reachabilty

georgemoralis:
two checks and notifications added

1) Check if configured shadnet server is reachable
2) Check if protocol version matches

If any of the above aren't true , it will disable a notification message and will disable shadnet features for the current run. This will avoid retries if shadnet server isn't available and will force users to go to latest version if there is a protocol missmatch


StevenMiller123 (approved):
Code looks fine, and this worked for others in testing.

PR #4693, opened 2026-07-11
Fix flaky GCN tests

al-jibran:
The GCN tests would sometimes pass on Windows with Visual Studio 2022 but consistently failed on Linux.

On Linux, a segfault during teardown caused CTest to report every GCN test as failed. After fixing the segfault, some tests were still flaky because the generated instructions depended on uninitialized values in memory.

This PR explicitly destroys the GCN test runner during test suite teardown and initializes the instruction fields so the generated encodings are deterministic everywhere.

Tested on:
- Ubuntu 25.10, NVIDIA GTX 1650 / driver 595.71.05
- Ubuntu 25.10 with llvmpipe
- Windows 11 with Visual Studio 2022

### DIFF
--- a/tests/gcn/gcn_test_runner.cpp
+++ b/tests/gcn/gcn_test_runner.cpp
@@ -136,6 +136,11 @@ std::expected<Runner*, ErrorInfo> Runner::instance() {
     return g_runner.get();
 }
 
+void Runner::DestroyInstance() {
+    std::lock_guard lock{g_runner_mutex};
+    g_runner.reset();
+}
+
 std::expected<void, ErrorInfo> Runner::initialize() {
     VULKAN_HPP_DEFAULT_DISPATCHER.init();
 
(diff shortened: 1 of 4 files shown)


DarkBoy08:
Is this the reason I can't build my .exe file? Or is there another problem?

al-jibran:
> Is this the reason I can't build my .exe file? Or is there another problem?

This PR only touches test files, so it wouldn't affect the normal executable build.

DarkBoy08:
> This PR only touches test files, so it wouldn't affect the normal executable build

Understood. How do I create an shadps4.exe file? The reason I ask is that I’ve tried almost everything, and I just can't get it to work at all.

red-prig:
> > This PR only touches test files, so it wouldn't affect the normal executable build
> 
> Understood. How do I create an shadps4.exe file? The reason I ask is that I’ve tried almost everything, and I just can't get it to work at all.

If you are new to programming, please ask this question on [Discord](https://discord.com/invite/bFJxfftGW6), but not in this thread, which is not related to compilation.

DarkBoy08:
Sorry I can't, because I've been banned for pirate(

PR #4699, opened 2026-07-12
[Core] Add configurable PS4 HDD timing for /app0 reads

cuesta4:
## The problem

After two fucking months of debugging, finally: God of War 3 Remastered has multiple texture corruption problems during normal gameplay on shadPS4. This PR addresses one of them. After investigating a giant multi-frame RenderDoc capture, I isolated the problem to the CPU production side. After a painfully long analysis using CDB, I tracked the bug to a fucking use-after-free.

The problem? It's in the game's logic, not shadPS4's. The game uses a compactor running on the main thread that was tuned, terribly one might say, around the PS4's original asset-streaming and HDD speeds. When under memory pressure due to the increased I/O speed possible on shadPS4, or even a warm cache on the PS4 as documented in some user reports, the allocator fails to allocate memory for high-resolution textures and correctly returns `RAX = 0`, BUT THE GAME IGNORES IT.

The game then proceeds to use the memory slot in the texture descriptor despite not actually having ownership over it. In most cases, this causes garbage data to be uploaded to the GPU and therefore corrupt textures. Other times, it causes crashes. It's technically possible to create a per-game patch that circumvents this, but it would take a lot of work that I'm now too burned out to do after this incredible experience.

## The fix

Implement a completion-rate limiter that preserves the resulting FIOS2 FIFO order - it is disabled by default and gated by a setting. The limiter uses a token-bucket virtual-time cursor. Each read advances the cursor by `bytes / configured bandwidth`, while accumulated credit is clamped to a 5 ms burst window. The host `PRead` starts immediately into a reusable staging buffer. The storage worker waits only for any modeled time remaining after the host read completes. Data is then copied to the guest spans, and the operation is completed. This avoids large completion bursts on SSDs and faster HDDs.

Add a time-dilation factor derived from the guest's presentation cadence. Real guest flips feed an exponential moving average of the observed flip period. It scales future token charges from 1× to 4× in 25% increments. This prevents the game from receiving too much throughput during slower moments. Time dilation can be disabled independently. Both synchronous reads and AIO read commands respect this limiter.

Configured bandwidths between 50 and 200 MiB/s enable the limiter. Anything below 50 MiB/s is clamped to 50 MiB/s to avoid issues. Setting it to 0 or anything above 200 MiB/s disables the limiter.

The launcher options are already implemented. I will open a PR for them when this one is approved.

## Testing

Manually tested in the Gaia scene. The fread PR helps with general texture corruption, but this limiter remained necessary to fix the scene in the screenshots.

## Before and After

BEFORE
AFTER

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -867,6 +867,8 @@ set(CORE src/core/aerolib/stubs.cpp
          src/core/file_sys/directories/normal_directory.h
          src/core/file_sys/directories/pfs_directory.cpp
          src/core/file_sys/directories/pfs_directory.h
+         src/core/file_sys/storage_scheduler.cpp
+         src/core/file_sys/storage_scheduler.h
          src/core/file_format/pfs.h
          src/core/file_format/psf.cpp
          src/core/file_format/psf.h
--- a/src/core/emulator_settings.h
+++ b/src/core/emulator_settings.h
@@ -188,6 +188,8 @@ struct GeneralSettings {
     Setting<bool> neo_mode{false};
     Setting<bool> dev_kit_mode{false};
     Setting<int> extra_dmem_in_mbytes{0};
+    Setting<u32> app0_read_bandwidth_mibps{0};
+    Setting<bool> app0_read_disable_time_stretching{false};
     Setting<bool> shad_net_enabled{false};
     Setting<bool> trophy_popup_disabled{false};
     Setting<double> trophy_notification_duration{6.0};
@@ -211,6 +213,10 @@ struct GeneralSettings {
             make_override<GeneralSettings>("dev_kit_mode", &GeneralSettings::dev_kit_mode),
             make_override<GeneralSettings>("extra_dmem_in_mbytes",
                                            &GeneralSettings::extra_dmem_in_mbytes),
+            make_override<GeneralSettings>("app0_read_bandwidth_mibps",
+                                           &GeneralSettings::app0_read_bandwidth_mibps),
+            make_override<GeneralSettings>("app0_read_disable_time_stretching",
+                                           &GeneralSettings::app0_read_disable_time_stretching),
             make_override<GeneralSettings>("shad_net_enabled", &GeneralSettings::shad_net_enabled),
             make_override<GeneralSettings>("trophy_popup_disabled",
                                            &GeneralSettings::trophy_popup_disabled),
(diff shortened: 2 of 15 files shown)


DarkBoy08:
Really, there is a difference, but without the readbacks function, some textures turn black. Here are my tests:
Without Readbacks
Native PR
GOW 3 Native PR Without Readbacks
PR4699
GOW 3 PR4699 Without Readbacks
With Readbacks Precise
Native PR 
GOW 3 Native PR With Readbacks Precise
PR4699
GOW 3 PR4699 With Readbacks Precise

Gih-pt:
I can't see any clear difference between the two images without the readbacks function enabled.

DarkBoy08:
> I can't see any clear difference between the two images without the readbacks function enabled.

Sorry, maybe I just it seemed, it when I looked. It happens to everyone. Or maybe I have bad hardware.

roamic:
I've tried running the game from my 20yo WD Blue HDD and immediately got the glitched textures which makes me doubt the proposed analysis. Are you sure that this is purely an issue with the drive speed?

cuesta4:
No. I can only affirm that:

1. The game's compactor can fall behind when asset streaming is too fast.
2. This can exhaust the pool and cause the allocator to return RAX = 0.
3. The game ignores that allocation failure and continues using the pointer already stored in the recycled texture descriptor.
4. Reducing I/O throughput gives the compactor enough time to reclaim memory before the pool is exhausted.

Increasing the pool size or changing WAD padding could sometimes make Gaia render correctly, but only by changing the memory layout or timing. The game still used a stale pointer, it simply happened to reference harmless mapped memory. When it pointed to invalid or unmapped memory, the game crashed.

Limiting I/O throughput was the only approach I found that consistently prevented the issue, but like I said, I am open to better solutions.

kalaposfos13:
While most devs don't really agree with the correctness of this PR, it should be noted that this appears to help SotC as well, as confirmed by tatu on the Discord server.

georgemoralis:
it is not that we don't agree maybe let's say it's questionable :D

Heliumrich:
I know Dolphin has a feature to emulate the slow CD/DVD read speeds, and it's enabled by default ! (for accuracy I imagine)

So even if read speed being way faster is never/rarely an issue for the PS4 moderne architecture, I think it still makes sense to have this feature 🤔 even if disabled by default and buried down in the settings

raphaelthegreat:
I personally find the premise of the PR acceptable tbh. Before reviewing the code tho, why cant this be implemented by sleeping the thread before a file read for appropriate time depending on bytes read and its done witb a complex async scheduler?

cuesta4:
I tried a fixed sleep on the submitter, but it led to poor performance overall and inconsistent results, especially in multiple small file reads, the sleep quickly adds up. It still does hit performance, but it's much better. There's probably a much better way of doing it.

raphaelthegreat:
For small file reads you could try to use a spin lock to burn some cpu cycles in a more precise manner or userspace available x64 instructions for small sleeps.

StevenMiller123:
Has GoW3 been retested after my fread PR?
It's fixed a wide variety of file streaming issues, I wouldn't be surprised if it at least reduced the frequency of these corruption issues.

cuesta4:
@raphaelthegreat I'll give it a shot.

@StevenMiller123 yes, just tested it. Same behavior in my control scene (the one in the PR's screenshots).

bigol83:
> Has GoW3 been retested after my fread PR? It's fixed a wide variety of file streaming issues, I wouldn't be surprised if it at least reduced the frequency of these corruption issues.

i tested god of war 3 yesterday with main build and it effectively reduced those corruptions. I played 20 minutes before reaching the first corrupted texture.

PR #4702, opened 2026-07-13
fix: null dereference in internal__Foprep on file exhaustion

### LINKED ISSUE #4698: [APP BUG]: internal__Foprep missing return after file == nullptr check — sets ENOMEM then dereferences the null file
### Checklist

### Describe the Bug

```
In src/core/libraries/libc_internal/libc_internal_io.cpp,
internal__Foprep checks its `file` argument for null, sets the error
code, but DOES NOT RETURN — and then unconditionally dereferences file:

https://github.com/shadps4-emu/shadPS4/blob/cf3bb346bf9d73471e7bb64a8bcafd52f36c3685/src/core/libraries/libc_internal/libc_internal_io.cpp#L74

    if (file == nullptr) {
        *Kernel::__Error() = POSIX_ENOMEM;
    }

    // Preserve mode and index
    Libraries::Kernel::PthreadMutexT mtx = file->_Mutex;   // <-- null deref
    ...
    memset(file, 0, sizeof(OrbisFILE));                    // <-- memset into null

The null case is reachable: the caller internal_fopen obtains file from
internal__Fofind() and even checks internal__Foprep's result for null
afterwards — so the function is clearly expected to return null on this
path. When the FILE table is exhausted, the emulator crashes inside
internal__Foprep instead of failing the fopen cleanly.

Suggested fix:

    if (file == nullptr) {
        *Kernel::__Error() = POSIX_ENOMEM;
        return nullptr;
    }

Found with ZeroDefect, a C/C++ dataflow static analyzer I am developing
(reported as `null-deref: 'file' may be null` on every dereference after
the check), then verified by hand.
```

### Reproduction Steps

```
Source-level defect, no specific game required:

1. Read internal__Foprep at libc_internal_io.cpp:72-88 (commit cf3bb34,
   current main).
2. At runtime: exhaust the FILE table so internal__Fofind() returns
   null, then any guest fopen through libSceLibcInternal crashes in
   internal__Foprep instead of returning null with ENOMEM.
```

### Expected Behavior

```
internal__Foprep should return nullptr with *__Error() = POSIX_ENOMEM
(exactly what the existing check intends, and what the caller already
handles), letting the guest see a failed fopen instead of crashing the
emulator.
```

### Specify OS Version

Windows 11 Home (build 26200)

### CPU

Intel Core i5-1235U

### GPU

Intel Iris Xe Graphics (integrated)

### Amount of RAM in GB

16 GB

### Log File

_No response_

Harshalj2108:
Closes #4698
This PR fixes a bug in `libc_internal_io.cpp` where the emulator would crash upon exhausting the `FILE` table. 
Previously, `internal__Foprep` checked if the `file` argument was `nullptr` and set the `POSIX_ENOMEM` error, but it was missing an explicit `return` statement. This caused the function to fall through and immediately dereference the null `file` pointer on the subsequent lines. 
By adding the missing `return nullptr;`, this function now cleanly propagates the error back to the guest application instead of crashing the emulator. 
# Changes
* Added `return nullptr;` after setting `*Kernel::__Error() = POSIX_ENOMEM;` inside `internal__Foprep`.

### DIFF
--- a/src/core/libraries/libc_internal/libc_internal_io.cpp
+++ b/src/core/libraries/libc_internal/libc_internal_io.cpp
@@ -73,6 +73,7 @@ OrbisFILE* PS4_SYSV_ABI internal__Foprep(const char* path, const char* mode, Orb
                                          s32 fd, s32 s_mode, s32 flag) {
     if (file == nullptr) {
         *Kernel::__Error() = POSIX_ENOMEM;
+        return nullptr;
     }
 
     // Preserve mode and index


georgemoralis:
it appears ok

PR #4708, opened 2026-07-14
video_core: synchronize independently backed aliased images

cuesta4:
## Problem
TextureCache allocates distinct image slots when descriptors differ, even if sharing guest_address. Storage writes (`ImageFlagBits::GpuModified`) update only the writer's Vulkan image, causing aliased sampler reads in FindTexture (e.g., GOW3 luminance reduction chains) to consume stale allocations.

## Solution
Introduce generation-based alias tracking via alias_states in TextureCache. SynchronizeAlias checks writer `alias_generation` during UpdateImageForTexture to copy overlapping subresources via ScheduleImageDownload, while an asynchronous ReadbackTracker handles CPU backing writebacks (`TryWriteBacking`) and invalidates stale aliases on CPU/GPU writes and unmaps.

## Testing
image
image

### DIFF
--- a/src/video_core/buffer_cache/buffer.cpp
+++ b/src/video_core/buffer_cache/buffer.cpp
@@ -160,6 +160,13 @@ void Buffer::Fill(u64 offset, u32 num_bytes, u32 value) {
     });
 }
 
+void Buffer::Invalidate(u64 offset, u64 size) {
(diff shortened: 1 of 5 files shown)


cuesta4:
@raphaelthegreat mind reviewing? You seem to be more familiar with this stuff

PR #4709, opened 2026-07-14
Lib.Kernel: Check to ensure heap_api is valid before running heap mallocs in TlsGetAddr

StevenMiller123:
Technically libkernel uses libSceLibcInternal's malloc, but we're already doing this host malloc workaround in several other places, so this should be fine.


kalaposfos13:
This fixes a highly specific crash with Http Live Streaming with both libSceAvPlayer and libSceAvPlayerStreaming LLEd, and makes the only currently known PS4 app using that feature, which is a homebrew I just made today, functional once the aforementioned libraries are LLEd.

PR #4710, opened 2026-07-14
Don't load HLE libraries if they're force-LLEd

kalaposfos13:
This makes messing around with LLE libraries easier, as it now doesn't require manually editing the source code for libraries already having a HLE version.


src/core/libraries/libs.cpp
```diff
@@ -78,91 +80,116 @@
...
+            {"libSceAudio3d.sprx", RegisterAudio3d},
+            {"libSceHttp.sprx", Libraries::Http::RegisterLib},
+            {"libSceHttp2.sprx", Libraries::Http2::RegisterLib},
+            {"libSceNet.sprx", Libraries::Net::RegisterLib},
+            {"libSceNetCtl.sprx", Libraries::NetCtl::RegisterLib},
+            {"libSceSaveData.sprx", Libraries::SaveData::RegisterLib},
+            {"libSceSaveDataDialog.sprx", Libraries::SaveData::Dialog::RegisterLib},
+            {"libSceSsl2.sprx", Libraries::Ssl2::RegisterLib},
+            {"libSceSysmodule.sprx", Libraries::SysModule::RegisterLib},
+            {"libScePosix.sprx", Libraries::Posix::RegisterLib},
```
StevenMiller123: Might as well get rid of this one, libScePosix isn't a real firmware lib (since it's a sub-library of libkernel), and we don't register anything through it anyway.

StevenMiller123 (commented):
I think this looks mostly fine to me, just one nit

src/core/libraries/libs.cpp
```diff
@@ -78,91 +79,115 @@
...
+            {"libSceNpSnsFacebookDialog.sprx", Libraries::Np::NpSnsFacebookDialog::RegisterLib},
+            {"libSceNpAuth.sprx", Libraries::Np::NpAuth::RegisterLib},
+            {"libSceNpParty.sprx", Libraries::Np::NpParty::RegisterLib},
+            {"libSceNpPartner.sprx", Libraries::Np::NpPartner::RegisterLib},
+            {"libSceNpTus.sprx", Libraries::Np::NpTus::RegisterLib},
+            {"libSceScreenShot.sprx", Libraries::ScreenShot::RegisterLib},
+            {"libSceAppContent.sprx", Libraries::AppContent::RegisterLib},
+            {"libScePngDec.sprx", Libraries::PngDec::RegisterLib},
+            {"libScePlayGo.sprx", Libraries::PlayGo::RegisterLib},
+            {"libScePlayGo.sprx", Libraries::PlayGo::Dialog::RegisterLib},
```
StevenMiller123: Should be libScePlayGoDialog.sprx

src/core/libraries/libs.cpp
```diff
@@ -78,91 +79,115 @@
...
+            {"libSceRemoteplay.sprx", Libraries::Remoteplay::RegisterLib},
+            {"libSceRazorCpu.sprx", Libraries::RazorCpu::RegisterLib},
+            {"libSceMove.sprx", Libraries::Move::RegisterLib},
+            {"libSceMouse.sprx", Libraries::Mouse::RegisterLib},
+            {"libSceWebBrowserDialog.sprx", Libraries::WebBrowserDialog::RegisterLib},
+            {"libSceZlib.sprx", Libraries::Zlib::RegisterLib},
+            {"libSceHmd.sprx", Libraries::Hmd::RegisterLib},
+            {"libSceHmdSetupDialog.sprx", Libraries::HmdSetupDialog::RegisterLib},
+            {"libSceDiscMap.sprx", Libraries::DiscMap::RegisterLib},
+            {"libSceUlobjmgr.sprx", Libraries::Ulobjmgr::RegisterLib},
```
StevenMiller123: Should be ulobjmgr.sprx

src/core/libraries/libs.cpp
```diff
@@ -78,91 +79,115 @@
...
+
+    auto* game_info = Common::Singleton<Common::ElfInfo>::Instance();
+    const auto& sys_module_path = EmulatorSettings.GetSysModulesDir();
+    const auto& game_specific_modules_path =
+        sys_module_path /
+        (game_info->GameSerial().empty() ? std::string_view("no_serial") : game_info->GameSerial());
+    {
+        constexpr auto ModulesToLoad = std::to_array<Core::SysModules>({
+            {"libkernel.sprx", Libraries::Kernel::RegisterLib},
+            {"libSceGnmDriver.sprx", Libraries::GnmDriver::RegisterLib},
```
StevenMiller123, line 107: If we're running in neo mode (sceKernelIsNeoMode() returns true) this should technically be libSceGnmDriverForNeoMode.sprx instead.

src/core/libraries/libs.cpp
```diff
@@ -181,10 +181,18 @@ void InitHLELibs(Core::Loader::SymbolsResolver* sym) {
         });
 
         for (auto mod : ModulesToLoad) {
-            if (!std::filesystem::exists(game_specific_modules_path / mod.module_name)) {
-                mod.callback(sym);
-            } else {
+            if (mod.module_name == "libSceGnmDriver.sprx") {
+                auto lle_file =
+                    EmulatorSettings.IsNeo() ? "libSceGnmDriverForNeoMode.sprx" : mod.module_name;
```
StevenMiller123: Not entirely correct, the normal libSceGnmDriver still loads if a game doesn't support Neo mode. `Common::ElfInfo::Instance().GetPSFAttributes().support_neo_mode`
That's why I specifically referenced the libkernel function.

StevenMiller123 (approved):
Looks good to me.

squidbus (commented):
~~I don't think sysmodules that cannot run LLE (e.g. need IPMI) should load, otherwise you will have people dumping all the sysmodules and putting them in and wondering their games are broken.~~

~~If you really want to be able to load *anything* LLE I think it should be a config option off by default.~~

Okay, I see this is just for the game-specific library path, carry on then...

squidbus (approved):

src/core/libraries/libs.cpp
```diff
@@ -78,91 +79,123 @@
...
+            {"libSceNpSignaling.sprx", Libraries::Np::NpSignaling::RegisterLib},
+            {"libSceNpScore.sprx", Libraries::Np::NpScore::RegisterLib},
+            {"libSceNpTrophy.sprx", Libraries::Np::NpTrophy::RegisterLib},
+            {"libSceNpWebApi.sprx", Libraries::Np::NpWebApi::RegisterLib},
+            {"libSceNpWebApi2.sprx", Libraries::Np::NpWebApi2::RegisterLib},
+            {"libSceNpProfileDialog.sprx", Libraries::Np::NpProfileDialog::RegisterLib},
+            {"libSceNpSnsFacebookDialog.sprx", Libraries::Np::NpSnsFacebookDialog::RegisterLib},
+            {"libSceNpAuth.sprx", Libraries::Np::NpAuth::RegisterLib},
+            {"libSceNpParty.sprx", Libraries::Np::NpParty::RegisterLib},
+            {"libSceNpPartner.sprx", Libraries::Np::NpPartner::RegisterLib},
```
StevenMiller123, line 137: Missed this earlier, but this should be libSceNpPartner001.sprx.
Shouldn't really matter anytime in the next 5 years, but I'll fix it when I have a chance.

PR #4711, opened 2026-07-14
Handle V_CMP_U64 const zero

S1ii:
## Summary

- handle `OperandField::ConstZero` in `V_CMP_U64`
- add regression coverage for `EQ`, `NE`, and `GT`

## Root cause

VOP3 `V_CMP_*_U64` shaders using zero as both operands reached the `src0` switch without a `ConstZero` case, causing `UNREACHABLE`.

## Validation

- `shadps4_gcn_test.exe --gtest_filter=GcnTest.cmp_u64_const_zero`
- built Release with Clang 19.1.7
- NHL 21 / CUSA18182 reaches the menu after language selection


StevenMiller123:
Replaced by #4972

PR #4713, opened 2026-07-14
CI: Check Windows visual studio generator

notanenergydrinkaddict:
CCache don't work with clang-cl and visual studio generator, it never gets sent any cache, so post workflow and it never uploads anything. So ninja is still preferred for CI builds. 
```
ccache stats
  C:\Strawberry\c\bin\ccache.exe -s
  Local storage:
    Cache size (GB): 0.0 / 0.5 ( 0.00%)
  C:\Strawberry\c\bin\ccache.exe --version
  ccache version 4.9.1

Not saving cache because no objects are cached.
```


georgemoralis:
but there are 2 targets now? windows-sdl and windows-vs-sdl?

PR #4714, opened 2026-07-15
renderer_vulkan: omit semantically inactive depth attachments

cuesta4:
## Problem

Vulkan dynamic rendering limits the render area to the smallest active attachment, while PS4 color and depth buffers are independent. A configured but semantically inactive depth target could therefore clip a full-resolution color pass and preserve stale pixels outside its extent.

## Fix

Depth/stencil attachments are now retained only when they can produce an observable effect: reject fragments, write depth or stencil, apply bounds, perform copies or decompression, or modify HTILE metadata. States such as `DepthClearEnable`, `ForceZValid`, compression policy, and sample selection no longer make depth active by themselves.

## Testing

Added focused coverage for inactive and side-effecting DB states; all five DepthStencilStateTest.* cases pass. The modified Vulkan renderer translation units compile successfully, and the fix was runtime-tested with God of War III Remastered.

Before and After:
image
image
image
image

### DIFF
--- a/src/video_core/renderer_vulkan/vk_depth_stencil_state.h
+++ b/src/video_core/renderer_vulkan/vk_depth_stencil_state.h
@@ -1,0 +1,81 @@
+// SPDX-FileCopyrightText: Copyright 2026 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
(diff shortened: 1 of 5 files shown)


raphaelthegreat:
Can you describe what configuration was seen? Did the size of color and depth attachments not match?

PR #4716, opened 2026-07-15
Core: Fix libSceNpPartner module name

StevenMiller123:
A quick and completely useless fix for a bug introduced by #4710.


kalaposfos13 (approved):
Here you go, have a quick and completely useless PR review for your quick and completely useless fix.

PR #4717, opened 2026-07-16
shader_recompiler: Initialize GS vertex offsets from the input primitive type

junminlee2004:
Restores rendering of the contour cel shader outline in Gravity Rush 2 by fixing a bug in which GS input vertex-offset registers were initialized from output primitive type, leaving adjacency vertices undefined. 
Unlike most shaders in Gravity Rush 2, cel shader outlines utilize triangles with adjacency (6 vertices). Currently, the GS prologue initializes ES-vertex-offset registers by switching on output primitive type, which limits the number of vertices that can be set up for input to 3, causing vertices 3-5 to be folded into garbage. The fix is to switch on input primitive type, intializing every provided offset (6) V0-V1, V3-V6. 

Before (switching on output primitive type)
image

After (switching on input primitive type)
image

### DIFF
--- a/src/shader_recompiler/frontend/translate/translate.cpp
+++ b/src/shader_recompiler/frontend/translate/translate.cpp
@@ -226,11 +226,28 @@ void Translator::EmitPrologue(IR::Block* first_block) {
         }
         break;
     case LogicalStage::Geometry:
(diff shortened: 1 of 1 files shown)


squidbus (approved):

georgemoralis:
looking good

PR #4593, opened 2026-06-18
 Refactor texture cache image download handling

Hog185:
Makes UFC 1 visible, obnoxious vertex explosions tho

PR #4719, opened 2026-07-16
Remove image ids from download_images on TextureCache::DeleteImage calls

StevenMiller123:
If we delete images while they're queued for download, then ProcessDownloadImages will attempt to retrieve a deleted image and trigger debug asserts (and probably more issues outside of debug builds).

This became more apparent after #4593, which both enabled image deletions during ProcessDownloadImages (through `scheduler.Finish()` running deferred operations), and added more images to `download_images`.

This PR adds a mutex to operations involving the `download_images` set, and removes images from `download_images` on DeleteImage calls, preventing these issues.

This fixes a regression in Assassin's Creed Unity with ReadbackLinearImages after #4593. This doesn't appear to break Marvel's Spider-Man, but will need testing in UFC titles (and needs a proper review).

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -61,6 +61,7 @@ TextureCache::TextureCache(const Vulkan::Instance& instance_, Vulkan::Scheduler&
 TextureCache::~TextureCache() = default;
 
 void TextureCache::ProcessDownloadImages() {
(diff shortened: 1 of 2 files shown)


squidbus (approved):
Looks fine if testing goes alright.

StevenMiller123:
@Hog185 tested on Discord, said they had no regressions. https://discord.com/channels/1080089157554155590/1258386683691274361/1527347281974788298

PR #4723, opened 2026-07-17
Modules updates

georgemoralis:
1. SDL3
2. OPENAL
3. CLI11
4. HTTPLIB
5. SPDLOG
6. JSON
7. VULKAN-HEADERS
8. VMA
9. GSLANG
10. ROBIN-MAP
11. MAGIC_ENUM


squidbus:
Not sure about the rest, but if you're updating Vulkan-Headers you should probably also update the version we look for with `find_package`.

Niram7777 (changes requested):
SPDLOG gives a glimpse of what the AI future will look likes...
We shouldn't update it until smart people maintain this repo instead of an AI agent.
https://github.com/gabime/spdlog/issues/3636

PR #4724, opened 2026-07-17
shader_recompiler: Always specialize vertex attributes on number class

Kravickas:
With VK_EXT_legacy_vertex_attributes present, the same vertex shader did not get separate variants for different attribute number classes (e.g. sint vs float) whichever variant was compiled first served all of them, reading the attribute through the wrong type.
In practice this corrupted vertex fetch for the mismatched draws, causing vertex explosions and missing UI elements.

Should I do the cleanup? What's the plan for now unused code?

### DIFF
--- a/src/shader_recompiler/profile.h
+++ b/src/shader_recompiler/profile.h
@@ -33,7 +33,6 @@ struct Profile {
     bool support_fp16_signed_zero_inf_nan_preserve{};
     bool support_fp32_signed_zero_inf_nan_preserve{};
     bool support_fp64_signed_zero_inf_nan_preserve{};
-    bool support_legacy_vertex_attributes{};
     bool supports_image_load_store_lod{};
     bool supports_native_cube_calc{};
     bool supports_trinary_minmax{};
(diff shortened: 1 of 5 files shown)


squidbus:
Can you elaborate on the need for this? `VK_EXT_legacy_vertex_attributes` does not require the numeric type to match, and the translated shader should be bit-casting to the necessary type for whatever instruction uses the inputs.

Kravickas:
With legacyVertexAttributes the numeric types don't have to match, but that only makes the binding legal, nothing defines what values arrive when the types differ. The proposal resolves that explicitly:

> Should implementations convert float/integer values? No. When fetching an integer data type from float values or float data types from integer values, the resulting shader values are implementation-dependent.
https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_legacy_vertex_attributes.html#_should_implementations_convert_floatinteger_values

The bitcasts operate on whatever ends up in the input variable, on implementations that convert during fetch according to the binding format, the original bits are gone before the shader runs. Log showed thousands of sint<->float mismatched reuses on main, zero with specialization active.

squidbus:
I see, I think you should also just remove the use of `VK_EXT_legacy_vertex_attributes` entirely while you're at it then, that was its purpose here. As it is here it's just enabled for nothing.

squidbus:
Should also remove enabling the extension in vk_instance

Kravickas:
I think removing the rest would regress unaligned vertex attributes?

squidbus:
> I think removing the rest would regress unaligned vertex attributes?

I don't think we deal with those, at the very least this extension was only added for number types. If those are a problem we need handling for them for drivers that don't have this extension anyway.

squidbus (approved):

PR #4729, opened 2026-07-17
Add ZArchive support for compressed game files

Aspenini:
This adds support for .zar files, offering a method of having dumped games stored as single files with a bit of compression.


RuskyDed:
Compiled locally, works great! You should probably handle applying -UPDATE, -patch and -mods folders as well, as it only launches the main file atm and packing them all into one isn't really convenient.

shinra-electric:
I tested on macOS with the base version of BB, and it worked fine.

I haven't tested updates. DLC should probably also be handled. 

If anyone else on macOS wants to try out this PR, I've published a macOS build of the ZArchive tool [here](https://github.com/shinra-electric/ZArchive/releases/tag/v0.1.2). The official repo doesn't have a macOS build yet (though I've already PR'd one).

Aspenini:
For updates/mods/dlc, would something like this be sufficient?
image
And for DLC have each one be able to be loaded as a .zar (I have no DLC to test on)

shinra-electric:
> For updates/mods/dlc, would something like this be sufficient? image And for DLC have each one be able to be loaded as a .zar (I have no DLC to test on)

The base game, updates and dlc can usually all be contained in one `.zar` file. You can look at Cemu's `.wua` file format as an example (it is essentially the same as `.zar`).

RuskyDed:
Is it possible make it boot the base game from zar and mount the loose files from -patch folder after?

Aspenini:
> Is it possible make it boot the base game from zar and mount the loose files from -patch folder after?

Yes, I am working on another branch that does this if you would like to test
https://github.com/Aspenini/shadPS4/tree/extra-content-mounting

Aspenini:
> The base game, updates and dlc can usually all be contained in one `.zar` file. You can look at Cemu's `.wua` file format as an example (it is essentially the same as `.zar`).

It could be made to support both loose .zars and that. For said format, what do you suggest for the layout?

shinra-electric:
> It could be made to support both loose .zars and that. For said format, what do you suggest for the layout?

If you use ZArchive to unzip a `.wua` file, it's just a directory with the three things in it: Base game, update and DLC (if the update and DLC exist)

shinra-electric:
Here is an example from BotW: 

Screenshot 2026-07-19 at 10 57 40

RuskyDed:
> > Is it possible make it boot the base game from zar and mount the loose files from -patch folder after?
> 
> Yes, I am working on another branch that does this if you would like to test https://github.com/Aspenini/shadPS4/tree/extra-content-mounting

Tested that branch on Linux and couldn't make it mount additional loose folders unfortunately. However, it mounts additional zar files just fine (tested -UPDATE, -patch and -mods). Another thing that i noticed is if you have both the zar and the loose folder with the same name (CUSAxxxxx-patch.zar  and CUSAxxxxx-patch for example) in the same folder, it'd not mount any additional files whatsoever. Loading the base game as loose files with additional zar results in segmentation fault.

src/core/libraries/avplayer/avplayer_source.cpp
```diff
@@ -49,7 +51,16 @@ bool AvPlayerSource::Init(const AvPlayerInitData& init_data, std::string_view pa
         }
     } else {
         const auto mnt = Common::Singleton<Core::FileSys::MntPoints>::Instance();
-        const auto filepath = mnt->GetHostPath(path);
+        auto filepath = mnt->GetHostPath(path);
+        if (Common::FS::Zar::IsZarInnerPath(filepath)) {
+            // FFmpeg cannot read from inside a ZArchive; extract the media to a host file.
```
roamic: This is incorrect, ffmpeg CAN read from an arbitrary source through AVIO. See `avplayer_file_streamer.h` for example.

Aspenini: I replaced the extraction path with an IOFile-backed AVIO context, so archived media is now streamed directly.

roamic: Please separate ZArchiver streaming code from the PS4 callbacks code. Make a separate file/class for your `AVIOContext` wrapper. Mixing them will lead to confusion when reading the code as they are very different things.

Heliumrich:
+1 on supporting loading a loose folder full of (merged) mods (the ones replacing/adding game files) ontop of a single game+DLC+update .zar (not like games are getting updates or new dlc)

Can we choose the compression ratio ? Or it's adding too much complexity for end-users for not much benefit
Would be interesting to see stats about size reduction (and read speed/loading times on HDDs)

Aspenini:
> +1 on supporting loading a loose folder full of (merged) mods (the ones replacing/adding game files) ontop of a single game+DLC+update .zar (not like games are getting updates or new dlc)

I think support for a single combined `.zar` containing the base game, updates, and DLC would be better handled in a separate PR

> Can we choose the compression ratio ? Or it's adding too much complexity for end-users for not much benefit Would be interesting to see stats about size reduction (and read speed/loading times on HDDs)

ZArchive already uses a fixed Zstd level of 6, and its tool does not expose compression settings. Since PS4 assets are often already compressed, adding that option would likely provide little benefit

georgemoralis:
adding base , updates and dlc inside one archive will make some games more than 120GB . I don't think it is a good idea , or it should be used as option

StevenMiller123:
If the whole point of this is to allow game compression, I think it's fitting to add support for one archive that contains both the base game and at least one update. Adding support for DLCs would be a pain with how we currently handle those, so I wouldn't really request that much.

georgemoralis:
compression is minimal as told (1-2%) . The point is to have 1 file instead of 15.000 game files per game and save some HDD life :D

src/core/libraries/avplayer/avplayer_source.cpp
Aspenini: > Please separate ZArchiver streaming code from the PS4 callbacks code. Make a separate file/class for your `AVIOContext` wrapper. Mixing them will lead to confusion when reading the code as they are very different things.

AvPlayerFileStreamer now handles only the PS4 file callbacks, ZArchive streaming is isolated in AvPlayerZarStreamer, and the AVIOContext lifecycle/callback bridge is encapsulated in a separate AvioContext class.

roamic: > the AVIOContext lifecycle/callback bridge is encapsulated in a separate AvioContext class.

... why? Please don't make things more convoluted, just leave file streamer as it was, with no changes and create your own that is based on `IDataStreamer`.

larryscruff616:
I just did a 20 game run compressing to ZAR. Here are the result...

TOTAL (20 game(s) with a matching .zar)
Folder - 357.41GB
ZAR - 268.62GB
Reduction - 24.8%

The results are a real mixed bag. Around half the games show a 1-10% reduction in file size, while the other half are around 30-50%. The average reduction in size is 25% for this sample.

See attached file for individual game results.

[zar_comparison.txt](https://github.com/user-attachments/files/30223811/zar_comparison.txt)

georgemoralis:
Some games have already compressed stuff  , so others don't . So it is mostly if the game already has it's assets compressed or not.

Nicktamer:
I convert some games I have to see the gain.

For some games the gain is very limited 1 or 2go, but for some the gain is
almost 50% (Wipeout omega folder is 40 go, zar is 21go)
Hope this can be merged, the gain can be interesting and have 1 file is more interesting than a folder with many files (when copie, it's more faster 1 files, than 3000!)

georgemoralis:
@Aspenini there is a conflict with latest main . @roamic  is already agree on his part . I am ok in others waiting for latest approval from @StevenMiller123 as it touches some filesystem stuff he is more familiar with

StevenMiller123 (changes requested):
I'm not a fan of how tightly the Zar handling is mixed with our normal filesystem code.
I feel the Zar code should be something modular that gets called into when a Zar file is in use, rather than being called into everywhere and having all these checks for whether a Zar is relevant.

Heliumrich:
> compression is minimal as told (1-2%) . The point is to have 1 file instead of 15.000 game files per game and save some HDD life :D

Does it really change much about HDD life, if the "15000" files are not fragmented, versus a "big zip" ?
I mean in both scenarios the HDD will still have to read ton of small stuff in random places.
Unless there's a way to put data that's often read together, close together. With some "runtime map" generated for each game

georgemoralis:
this was reworked on #4749 Thank you for the suggestion and your work!

PR #4734, opened 2026-07-19
texture_cache: Fix MipOf and SliceOf checks

StevenMiller123:
After #4553 modified these checks, UE games started exhibiting a green glow. The issue was that the BlockDim width was coming from image pitch. The mip's pitch value is padded, while the pitch calculated from the parent image was not. As a result, the width validity check was failing, and actual image mips were not getting recognized correctly.

This PR changes the check back to using the actual width and height, and manually shifting values to appropriately account for block images. I've also added a separate check for the actual pitch, accounting for the padding by checking through mips_layout instead of calculating based on mip index.

These changes were suggested by @raphaelthegreat, and fixes the UE green glow.

### DIFF
--- a/src/video_core/texture_cache/image.cpp
+++ b/src/video_core/texture_cache/image.cpp
@@ -672,12 +672,15 @@ void Image::CopyImageWithBuffer(Image& src_image, vk::Buffer buffer, u64 offset)
 void Image::CopyMip(Image& src_image, u32 mip, u32 slice) {
     const auto& src_info = src_image.info;
 
-    const auto src_block_dim = src_info.BlockDim();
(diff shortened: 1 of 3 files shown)


squidbus (approved):

StevenMiller123:
Seems this reveals an issue in some of my other titles, gonna debug that before this gets merged.

StevenMiller123:
Had to fix the assert checks in CopyMip, since those were checking through BlockDim and could fail now that MipOf and SliceOf checks are fixed.

Fixes an assert Phantasy Star Online 2: New Genesis was hitting in prior builds of this PR.

PR #4735, opened 2026-07-20
core: initialize NGS2 rack query output

MathieuARS:
## Summary

  Initialize the NGS2 rack query output before populating its implemented fields.

  This prevents callers from observing uninitialized host-stack data in fields that shadPS4 does not currently fill, while preserving the existing implemented values and ABI layout.

  ## Validation

  - Added focused NGS2 rack-query coverage
  - RelWithDebInfo build completed successfully with clang-cl
  - CTest: 411 passed, 1 HTTP-dependent test skipped, 0 failed
  
Game: Shadow of the Colossus


georgemoralis:
that doesn't look bad , but you should also reference the game that was causing the problem with . Or it happends to any game that uses HLE ngs2? (which is incomplete in any case)

MathieuARS:
> that doesn't look bad , but you should also reference the game that was causing the problem with . Or it happends to any game that uses HLE ngs2? (which is incomplete in any case)

Oh you are right, sorry

I'm currently trying to make Shadow of the colossus work

georgemoralis:
you can use lle ngs2 btw most of the games works fine with it

PR #4737, opened 2026-07-20
shader_recompiler: implement V_MBCNT masked bit counts

MathieuARS:
## Summary

  This replaces the previous compute lane-ID post-pass with a proper implementation of `V_MBCNT_LO_U32_B32` and `V_MBCNT_HI_U32_B32`.

  shadPS4 previously lowered `v_mbcnt_lo_u32_b32 v0, -1, v0` directly to `LaneId`, which discarded the `v0` accumulator. This exact instruction (`0x460000c1`) appears in Shadow of the Colossus compute shader `0x91c6c95e`.

  Compute shaders now use a separate guest lane ID derived from `LocalInvocationIndex & 63`. The existing `LaneId` still represents the host subgroup lane, so real subgroup operations are not broadly rewritten.

  `EXEC`-backed masks retain their existing behavior for now, since reconstructing guest wave64 `EXEC` across multiple host subgroups needs a separate implementation.

  ## Validation

  - Rebased onto current `main` (`4c72de07`)
  - Full RelWithDebInfo executable and PDB build succeeded with clang-cl 22.1.3
  - Full GCN shader suite: 49 passed, 0 failed
  - Added a Vulkan regression test using the exact `0x460000c1` instruction
  - Added translation tests for LO/HI masked counts, compute guest lane IDs, non-compute subgroup lane IDs, and shader metadata tracking
  
Game: Shadow of the Colossus

### DIFF
--- a/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
@@ -463,6 +463,7 @@ Id EmitImageAtomicExchange32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id co
 Id EmitImageAtomicCmpSwap32(EmitContext& ctx, IR::Inst* inst, u32 handle, Id address, Id value,
                             Id cmp_value);
 Id EmitCubeFaceIndex(EmitContext& ctx, IR::Inst* inst, Id cube_coords);
+Id EmitGuestLaneId(EmitContext& ctx);
 Id EmitLaneId(EmitContext& ctx);
 Id EmitWarpId(EmitContext& ctx);
 Id EmitQuadShuffle(EmitContext& ctx, Id value, Id index);
--- a/src/shader_recompiler/backend/spirv/emit_spirv_warp.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_warp.cpp
@@ -6,6 +6,13 @@
 
 namespace Shader::Backend::SPIRV {
 
+Id EmitGuestLaneId(EmitContext& ctx) {
+    ASSERT(ctx.l_stage == LogicalStage::Compute);
+    constexpr u32 guest_wave_size = 64;
+    const Id invocation_index = ctx.OpLoad(ctx.U32[1], ctx.local_invocation_index);
+    return ctx.OpBitwiseAnd(ctx.U32[1], invocation_index, ctx.ConstU32(guest_wave_size - 1));
+}
+
 Id SubgroupScope(EmitContext& ctx) {
     return ctx.ConstU32(static_cast<u32>(spv::Scope::Subgroup));
 }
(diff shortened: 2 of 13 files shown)


raphaelthegreat:
What bug is this supposed to fix, can you show an example? I think this being a post-pass is also a bit redudant, you could add a new IR opcode for a guest lane id or make lane id be this actual value. But I feel this could be a hack too that hides a more real issue

MathieuARS:
> What bug is this supposed to fix, can you show an example? I think this being a post-pass is also a bit redudant, you could add a new IR opcode for a guest lane id or make lane id be this actual value. But I feel this could be a hack too that hides a more real issue

I found this while debugging Shadow of the Colossus. One of its compute shaders uses the lane ID to index into shared memory, but on my RTX card shadPS4 hands it 0-31 twice instead of 0-63, so the second half ends up overwriting the first.

You're right that the post-pass is too broad though. I'll rework it with a separate guest lane ID in the IR so it doesn't interfere with actual host subgroup operations

raphaelthegreat:
What is the GCN isa used to for this indexing?

MathieuARS:
> What is the GCN isa used to for this indexing?

It's v_mbcnt_lo_u32_b32 v0, -1, v0 (0x460000c1), which shows up twice in compute shader 0x91c6c95e. shadPS4 currently lowers that to LaneId whenever the first source is -1, ignoring the v0 accumulator entirely. I'm checking whether that lowering is actually the underlying problem you were pointing at

raphaelthegreat:
Is the local workgroup size 64 too or not?

MathieuARS:
> Is the local workgroup size 64 too or not?

Yeah, it's 8x8x1, so 64 local invocations, and I confirmed that in the dumped SPIR-V. For this shader LocalInvocationIndex is already in 0-63, so the & 63 isn't needed here. It's only there for the general case, where a workgroup can hold more than one guest wave64.

Sorry for the couple edits, english isn't my main language so i'm trying to explain as best as i can with a translator

thethunderturner:
CoD Infinite warfare seems to crash further now.
Before:
```
[Debug] <Critical> (shadPS4:GpuComm) vector_alu.cpp:724 V_MBCNT_U32_B32: Unreachable code!
```

After:
```
[Debug] <Critical> (shadPS4:GpuComm) vector_alu.cpp:1262 V_CMP_U64: Assertion Failed!
```

Might need to be combined with #4711

raphaelthegreat:
The lowering to lane id is correct in this case, though I can't remember any exact examples. Implementing mbcnt properly requires also adjusting DS_APPEND in a non-trivial manner so I wouldn't recommend, I think your last commit might have broken that. A simpler less intrusive approach might be to still have a pass that replaces lane_id with local_invocation_index if local workgroup size <= 64 and shader has no subgroup ops (could hijack the existing readlane elimination pass for this to avoid adding a new one)

StevenMiller123:
I believe this is effectively replaced by #4972? Can be reopened if I'm wrong.

PR #4741, opened 2026-07-21
avplayer: fixes for frame dimensions

roamic:
During the extensive testing on PS4 using the homebrew I have yet to find a case where these are not aligned for AVC videos. Let me know if it breaks videos in any games.

Also:
* Extended video data to the cropped part of the video as observed on the PS4 (only vertical for now)
* Synced vdec implementation to avplayer and vdec2 implementation to vdec improvements in flush function.
* Removed excessive memset from frame data fill
* Aligned pitch to 64 after testing videos with width 720 on PS4
* Fixed stream tracks selection by language
* Changed the number of audio buffers to `max_frame_buffers * 2` as observed on the PS4. Each audio buffer is now 32KB in size and aligned to 16 as it is on PS4.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -612,7 +612,9 @@ set(FIBER_LIB src/core/libraries/fiber/fiber_context.cpp
               src/core/libraries/fiber/fiber_error.h
 )
 
-set(VDEC_LIB src/core/libraries/videodec/videodec2_impl.cpp
+set(VDEC_LIB src/core/libraries/videodec/video_utils.cpp
+             src/core/libraries/videodec/video_utils.h
+             src/core/libraries/videodec/videodec2_impl.cpp
              src/core/libraries/videodec/videodec2_impl.h
              src/core/libraries/videodec/videodec2.cpp
              src/core/libraries/videodec/videodec2.h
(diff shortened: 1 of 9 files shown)


roamic:
Updated with the pitch aligned to 64

thethunderturner:
I've tested Black Ops 3 with this PR, which has some issues with avplayer in the intro sequence. It still has green bars, and the audio still lags behind the video. I wasnt expecting anything to change, but it is worth noting.

[CUSA02624.log](https://github.com/user-attachments/files/30274157/CUSA02624.log)

roamic:
Rebranded green line into black line, shouldn't be so noticeable now while I figure out if it's correct.

roamic:
Extended video data to the cropped part of the video (only vertical for now)

Synced vdec implementation to avplayer and vdec2 implementation to vdec improvements in flush function.

thethunderturner:
Looks ok. I will test in a minute

thethunderturner:
well, doesnt build on linux :p

bigol83:
i could test on windows and i think the latest push fixed the issue

PR
immagine

Main
immagine

don't just look at the green bar but also outlines in the big blue O letter that are now fixed with this PR

Same for the Naughty Dod logo

PR

immagine


Main
immagine

roamic:
> well, doesnt build on linux :p

Fixed the build

thethunderturner:
image

This fixes the green lines for me as well.

PR #4742, opened 2026-07-22
Kernel.Fs: Use intermediate buffer for storing file read data before copying to host memory

StevenMiller123:
In a variety of titles, our io_file ReadRaw would fail silently in the std::fread call. These failing calls would return 0 bytes read, set the stream's error flag for std::ferror to detect, and set errno to einval.

For some reason, std::fread does not error if we read into a host buffer, then copy that data to the guest. While I feel this might be too hacky to actually merge, I think it's useful as a test to get more information about affected titles.

The issues this caused depended on the title, but it's responsible for old UE4 titles attempting to map more memory than the PS4 contains, the InvalidFileFourCC errors in Shadow of the Colossus, and probably is responsible for issues in other titles too.

Credits to @Kravickas for (indirectly) giving me the idea to try this.

### DIFF
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -358,6 +358,7 @@ s64 ReadFile(Common::FS::IOFile& file, void* buf, u64 nbytes) {
     const auto remaining = file.GetSize() - file.Tell();
     memory->InvalidateMemory(reinterpret_cast<VAddr>(buf), std::min<u64>(nbytes, remaining));
     void* file_buf = std::malloc(nbytes);
+    ASSERT_MSG(file_buf != nullptr, "malloc failed on size {:#x}", nbytes);
     u64 bytes = file.ReadRaw<u8>(file_buf, nbytes);
     std::memcpy(buf, file_buf, bytes);
     std::free(file_buf);
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -352,16 +352,18 @@ s64 PS4_SYSV_ABI sceKernelWrite(s32 fd, const void* buf, u64 nbytes) {
     return result;
 }
 
+static std::vector<u8> file_buf{};
+
 s64 ReadFile(Common::FS::IOFile& file, void* buf, u64 nbytes) {
     const auto* memory = Core::Memory::Instance();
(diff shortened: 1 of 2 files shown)


Parotaku:
image

This helps Devil May Cry 5 loading a bit further on Windows here...

[shad_log.zip](https://github.com/user-attachments/files/30257425/shad_log.zip)

Parotaku:
'Little Nightmares'(1) doesn't crash anymore on death on Windows with this...

Here's the log with me dying twice but no crash:
[shad_log.zip](https://github.com/user-attachments/files/30257683/shad_log.zip)

raphaelthegreat:
Could it be related to protection again? We do invalidate the memory region but maybe gpu thread represents it before the read finishes? Also would switching to raw syscalls for nails ops help? I feel this should be a last resort due to the copy overhead

src/core/libraries/kernel/file_system.cpp
```diff
@@ -357,8 +357,11 @@ s64 ReadFile(Common::FS::IOFile& file, void* buf, u64 nbytes) {
     // Invalidate up to the actual number of bytes that could be read.
     const auto remaining = file.GetSize() - file.Tell();
     memory->InvalidateMemory(reinterpret_cast<VAddr>(buf), std::min<u64>(nbytes, remaining));
-
-    return file.ReadRaw<u8>(buf, nbytes);
+    void* file_buf = std::malloc(nbytes);
```
notanenergydrinkaddict: You should check if malloc actually succeed, if failed, memcpy below is ub

StevenMiller123:
@raphaelthegreat Not sure how important it is, but these issues still occur when nullgpu is enabled.

Randomuser8219:
image
This fixes the race condition and LODs of maps not loading in inFAMOUS: Second Son.

StevenMiller123:
Did some further investigation, the error is coming from native calls to OS-specific read functions. When converting IOFile code to use POSIX functions, the errors still occurred.

For now I've changed the allocation per read into a static std::vector which reserves more space when larger reads occur. This should help performance.

If anyone has better ideas for a fix, feel free to suggest them. For now I'm opening this up for review.

StevenMiller123:
Between other tests I've done, and the remaining things I've pushed here, I think this branch is good to go.

georgemoralis:
Yah let's goooo

PR #4745, opened 2026-07-22
Audio3d: Refactor mix queue logic for advance and flush

tatu16:
Previously sceAudio3dPortFlush was calling sceAudio3dPortAdvance to process mix queue, but it fails validation if buffer mode is set to ORBIS_AUDIO3D_BUFFER_NO_ADVANCE. This is causing flush to fail without emptying queue leading to ConvertAndEnqueue crashing the game.


georgemoralis:
does that fix applies to a specific game?

tatu16:
fixes Putty Squad CUSA00271 crashing at splash screen, I don't have any other game with this specific issue

georgemoralis:
doesn't look bad

PR #4749, opened 2026-07-24
Abstract FS

georgemoralis:
Inspired from #4729 so we can support multiple fs (now zarchieve) .

The abstraction is done in IFile.h . There are 2 backends the normal operation and zarchieve one .

### DIFF
--- a/src/core/file_sys/backends/host_fs.h
+++ b/src/core/file_sys/backends/host_fs.h
@@ -1,0 +1,91 @@
+// SPDX-FileCopyrightText: Copyright 2026shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include <filesystem>
+#include <memory>
+#include <string_view>
+#include "common/io_file.h"
+#include "core/file_sys/ifile.h"
+
+namespace Core::FileSys {
+
+/// IFile implementation backed by a real host file.
+class HostFile final : public IFile {
+public:
+    HostFile(std::filesystem::path host_path, Common::FS::FileAccessMode mode, bool read_only);
+    ~HostFile() override = default;
+
+    s64 Read(void* dst, u64 size) override;
+    s64 Write(const void* src, u64 size) override;
+    bool Seek(s64 offset, Common::FS::SeekOrigin origin) override;
+    u64 Tell() const override;
+    u64 Size() const override;
+    bool Flush() override;
+    bool IsOpen() const override;
+
+    std::optional<std::filesystem::path> GetHostPath() const override {
+        return m_path;
+    }
+
+    Common::FS::IOFile* GetHostFile() override {
+        return &m_file;
+    }
+
+    MmapPolicy GetMmapPolicy() const override {
+        return MmapPolicy::Native;
+    }
+
+    bool IsReadOnly() const override {
+        return m_read_only;
+    }
+
+private:
+    std::filesystem::path m_path;
+    Common::FS::IOFile m_file;
+    bool m_read_only;
+};
+
+/// Streaming iterator over a host directory.
+class HostDirectory final : public IDirectory {
+public:
+    explicit HostDirectory(const std::filesystem::path& root);
+    ~HostDirectory() override = default;
+
+    bool Next(DirEntry& out) override;
+    void Rewind() override;
+
+private:
+    std::filesystem::path m_root;
+    std::filesystem::directory_iterator m_it;
+    std::filesystem::directory_iterator m_end;
+};
+
+class HostFsBackend final : public IBackend {
+public:
+    HostFsBackend(std::filesystem::path root, bool read_only);
+    ~HostFsBackend() override = default;
+
+    bool Exists(std::string_view rel_path) override;
+    bool IsDirectory(std::string_view rel_path) override;
+
+    std::unique_ptr<IFile> Open(std::string_view rel_path, bool writable) override;
+    std::unique_ptr<IDirectory> OpenDir(std::string_view rel_path) override;
+
+    bool IsReadOnly() const override {
+        return m_read_only;
+    }
+
+    std::optional<std::filesystem::path> RootHostPath() const override {
+        return m_root;
+    }
+
+private:
+    std::filesystem::path Resolve(std::string_view rel_path) const;
+
+    std::filesystem::path m_root;
+    bool m_read_only;
+};
+
+} // namespace Core::FileSys
--- a/src/core/file_sys/backends/host_fs.h
+++ b/src/core/file_sys/backends/host_fs.h
@@ -41,6 +41,10 @@ public:
         return m_read_only;
     }
 
+    bool IsWriteOnly() const override {
+        return m_file.IsOpen() && m_file.IsWriteOnly();
+    }
+
 private:
     std::filesystem::path m_path;
     Common::FS::IOFile m_file;
@@ -70,7 +74,8 @@ public:
     bool Exists(std::string_view rel_path) override;
     bool IsDirectory(std::string_view rel_path) override;
 
-    std::unique_ptr<IFile> Open(std::string_view rel_path, bool writable) override;
+    std::unique_ptr<IFile> Open(std::string_view rel_path,
+                                Common::FS::FileAccessMode mode) override;
     std::unique_ptr<IDirectory> OpenDir(std::string_view rel_path) override;
 
     bool IsReadOnly() const override {
--- a/src/core/file_sys/backends/host_fs.h
+++ b/src/core/file_sys/backends/host_fs.h
@@ -76,7 +76,8 @@ public:
     bool Exists(std::string_view rel_path) override;
     bool IsDirectory(std::string_view rel_path) override;
 
-    std::unique_ptr<IFile> Open(std::string_view rel_path, Common::FS::FileAccessMode mode) override;
+    std::unique_ptr<IFile> Open(std::string_view rel_path,
+                                Common::FS::FileAccessMode mode) override;
     std::unique_ptr<IDirectory> OpenDir(std::string_view rel_path) override;
 
     bool IsReadOnly() const override {
--- a/src/core/file_sys/backends/host_fs.h
+++ b/src/core/file_sys/backends/host_fs.h
@@ -45,6 +45,8 @@ public:
         return m_file.IsOpen() && m_file.IsWriteOnly();
     }
 
+    void Stat(FileStat& out) override;
+
 private:
     std::filesystem::path m_path;
     Common::FS::IOFile m_file;
@@ -74,8 +76,7 @@ public:
     bool Exists(std::string_view rel_path) override;
     bool IsDirectory(std::string_view rel_path) override;
 
-    std::unique_ptr<IFile> Open(std::string_view rel_path,
-                                Common::FS::FileAccessMode mode) override;
+    std::unique_ptr<IFile> Open(std::string_view rel_path, Common::FS::FileAccessMode mode) override;
     std::unique_ptr<IDirectory> OpenDir(std::string_view rel_path) override;
 
     bool IsReadOnly() const override {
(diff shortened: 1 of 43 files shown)


georgemoralis:
this should be ready for testing

src/core/file_sys/backends/host_fs.h
```diff
@@ -0,0 +1,91 @@
+// SPDX-FileCopyrightText: Copyright 2026shadPS4 Emulator Project
```
StevenMiller123: Missing a space in the copyright text

src/core/file_sys/fs.cpp
```diff
@@ -24,7 +27,69 @@ void MntPoints::Mount(const std::filesystem::path& host_folder, const std::strin
...
+                        std::string_view suffix) -> std::shared_ptr<IBackend> {
+        std::filesystem::path dir_path = base;
+        dir_path += std::string{suffix};
+        if (auto b = make_backend(dir_path, /*ro=*/true)) {
+            return b;
+        }
+        std::filesystem::path zar_path = base;
+        zar_path += std::string{suffix};
+        zar_path += ".zar";
+        if (auto b = make_backend(zar_path, /*ro=*/true)) {
```
StevenMiller123: Looking at this code, I think it would be simpler to make `make_backend` test appending the .zar file extension instead.

src/core/file_sys/fs.cpp
```diff
@@ -24,7 +27,69 @@ void MntPoints::Mount(const std::filesystem::path& host_folder, const std::strin
...
+        }
+    }
+    std::shared_ptr<IBackend> base = make_backend(host_folder, read_only);
+    if (!base) {
+        std::filesystem::path zar_candidate = host_folder;
+        zar_candidate += ".zar";
+        base = make_backend(zar_candidate, read_only);
+    }
+    if (!base) {
+        base = std::make_shared<HostFsBackend>(host_folder, read_only);
```
StevenMiller123: If `make_backend` fails, then the folder doesn't exist. Since host path is expected to exist, we should probably assert rather than falling back on creating a HostFsBackend object.

src/core/file_sys/fs.cpp
```diff
@@ -191,61 +256,185 @@ std::filesystem::path MntPoints::GetHostPath(std::string_view path, bool* is_rea
...
+    if (writable && mount->read_only) {
+        return nullptr;
+    }
+    const auto rel = RelativeToMount(*corrected, *mount);
+    const std::string rel_copy{rel};
+
+    for (const auto& backend : mount->backends) {
+        // Skip read-only overlays when we need write access; writes
+        // always land on the base backend.
+        if (writable && backend->IsReadOnly()) {
```
StevenMiller123: I don't think this is necessary? Since alternate backends are for storing games, and games are read-only, attempting to open a file with write access would just trigger the `writable && mount->read_only` error above.

src/core/file_sys/ifile.h
```diff
@@ -0,0 +1,165 @@
+// SPDX-FileCopyrightText: Copyright 2025 shadPS4 Emulator Project
```
StevenMiller123: New files shouldn't have a 2025 copyright

src/core/file_sys/fs.h
```diff
@@ -103,12 +122,82 @@ struct File {
     std::filesystem::path m_host_name;
     std::string m_guest_name;
     Common::FS::IOFile f;
```
StevenMiller123: Why are we still storing an IOFile here? Isn't it unnecessary since the IFile handle stores it's own IOFile object?

src/core/libraries/kernel/file_system.cpp
```diff
@@ -223,31 +233,33 @@ s32 PS4_SYSV_ABI open(const char* raw_path, s32 flags, u16 mode) {
...
-            if (append) {
-                // Open for reading and appending
-                e = file->f.Open(file->m_host_name, Common::FS::FileAccessMode::ReadAppend);
-            } else {
-                // Open for reading and writing
-                e = file->f.Open(file->m_host_name, Common::FS::FileAccessMode::ReadWrite);
+        if (!non_host) {
+            if (read) {
+                // Open exclusively for reading
+                e = file->f.Open(file->m_host_name, Common::FS::FileAccessMode::Read);
```
StevenMiller123: If you make mnt->Open run IOFile::Open on it's own file, that should remove the need for these `file->f.Open` calls.
Just make sure the mnt->Open properly handles the logic used here.

src/core/libraries/kernel/file_system.cpp
```diff
@@ -767,33 +790,50 @@ s32 PS4_SYSV_ABI fstat(s32 fd, OrbisKernelStat* sb) {
...
+                    std::chrono::system_clock::now());
+                const auto secs = std::chrono::time_point_cast<std::chrono::seconds>(sctp);
+                const auto nsecs =
+                    std::chrono::duration_cast<std::chrono::nanoseconds>(sctp - secs);
+
+                sb->st_mtim.tv_sec = static_cast<int64_t>(secs.time_since_epoch().count());
+                sb->st_mtim.tv_nsec = static_cast<int64_t>(nsecs.count());
+                sb->st_atim = sb->st_mtim;
+                sb->st_ctim = sb->st_mtim;
+            }
```
StevenMiller123: If there's no way to retrieve these timestamps from the contents of zar archives, then you could probably get good enough dates from stat-ing the archive itself.
Also might be a good idea to have the burden of populating a stat struct moved to the file backends instead, so host_fs has the code for getting the actual times, while zarchive_fs has code for getting the archive's times.

src/core/memory.cpp
```diff
@@ -703,14 +703,22 @@ s32 MemoryManager::MapFile(void** out_addr, VAddr virtual_addr, u64 size, Memory
...
 
-    if (False(file->f.GetAccessMode() & Common::FS::FileAccessMode::Write) &&
-        False(file->f.GetAccessMode() & Common::FS::FileAccessMode::Append)) {
-        // If the file does not have write access, ensure prot does not contain write
-        // permissions. On real hardware, these mappings succeed, but the memory cannot be
-        // written to.
+    if (non_host_backed) {
+        // Non-host backends are read-only
         prot &= ~MemoryProt::CpuWrite;
+    } else {
```
StevenMiller123, line 713: Could probably simplify this by just adding `non_host_backend` to the access mode check?

src/core/memory.cpp
```diff
@@ -761,7 +769,46 @@ s32 MemoryManager::MapFile(void** out_addr, VAddr virtual_addr, u64 size, Memory
...
+        if (True(prot & MemoryProt::CpuRead)) {
+            perms |= Core::MemoryPermission::Read;
+        }
+        if (True(prot & MemoryProt::CpuReadWrite)) {
+            perms |= Core::MemoryPermission::ReadWrite;
+        }
+        if (True(prot & MemoryProt::CpuExec)) {
+            perms |= Core::MemoryPermission::Execute;
+        }
+        impl.Protect(mapped_addr, size, perms);
```
StevenMiller123: I don't like having all this extra logic here 🤔
Since the logic needs to be different based on backend capabilities, maybe this logic should be part of the backends themselves?

Maybe something like having a `file->Map` function that takes requested permissions, and calls into address space and other places as needed to do the rest of the work?

src/emulator.cpp
```diff
@@ -200,22 +211,31 @@ void Emulator::Run(std::filesystem::path file, std::vector<std::string> args,
...
+                const std::string base_name =
+                    game_folder_name.substr(0, game_folder_name.rfind('-'));
+                const auto base_path = game_folder.parent_path() / base_name;
+                if (std::filesystem::is_directory(base_path)) {
+                    game_folder = base_path;
+                }
             }
         }
-    }
+    } // if (!from_archive)
```
StevenMiller123: Is there anything this comment accomplishes?

StevenMiller123 (changes requested):
This honestly looks really good, just needs a little more refining in my opinion.

src/emulator.cpp
georgemoralis: fixed

src/core/file_sys/fs.h
georgemoralis: Devices still needs them

georgemoralis:
fixes a bunch of issues . now it is ready for some testing

src/core/file_sys/fs.h
StevenMiller123: I don't see what you're referring to here. From doing some local tests, it seems rather easy to get rid of this.

StevenMiller123:
Clang failed by the way :smile:

src/core/memory.cpp
```diff
@@ -759,9 +768,36 @@ s32 MemoryManager::MapFile(void** out_addr, VAddr virtual_addr, u64 size, Memory
...
+        if (True(mp & MemoryProt::CpuReadWrite)) {
+            perms |= Core::MemoryPermission::ReadWrite;
+        }
+        if (True(mp & MemoryProt::CpuExec)) {
+            perms |= Core::MemoryPermission::Execute;
+        }
+        impl.Protect(reinterpret_cast<VAddr>(addr), sz, perms);
+    };
+
+    file->handle->Map(reinterpret_cast<u8*>(mapped_addr), size, phys_addr, std::bit_cast<u32>(prot),
```
StevenMiller123, line 799: Not exactly the way I expected this to be handled 😅 
I guess it's fine though.

StevenMiller123 (commented):
Code looks fine to me, assuming the AvPlayer-related logic looks fine to @roamic 
I'm saving my approval for when I've had a chance to test this though.

RuskyDed:
Can't get zars working on Linux with this. Running with the terminal, it sees the CUSA, applies game-specific settings, launches the window for a couple of seconds and then crashes with:
> [Debug] <Critical> (Game:Main) log.cpp:257 Terminate: Exception: filesystem error: directory iterator cannot open directory: Not a directory [/mnt/WD2/Games/PS4/CUSA01322.zar]

On Windows it at least launches fine.

georgemoralis:
Added a possible fix for linux if you can re-test

RuskyDed:
That fixed it. Loose -UPDATE, -patch or -mods folders didn't get loaded though (same in form of additional zars)

StevenMiller123:
Breaks in games with multiple executables, it's only ever loading the eboot when it should load a game elf.
[CUSA05795.log](https://github.com/user-attachments/files/30394505/CUSA05795.log)

StevenMiller123:
Also would be nice to handle what @RuskyDed is mentioning, given that PR #4729 did allow it.

Additionally, worth noting that the regression I mentioned only applies to games tested as zar files. I didn't run into the multiple executable issue when testing my usual dumps.

georgemoralis:
> Breaks in games with multiple executables, it's only ever loading the eboot when it should load a game elf. [CUSA05795.log](https://github.com/user-attachments/files/30394505/CUSA05795.log)

fixed

georgemoralis:
> That fixed it. Loose -UPDATE, -patch or -mods folders didn't get loaded though (same in form of additional zars)

should be fixed

RuskyDed:
> should be fixed

Yes, it indeed now applies the additional loose files and zars from updates and mods, even zars on top of unpacked game. Tried even chaining loose base game + loose update + mods in zar, seems fine. I'd prefer someone else with wider library to test it still. 
While i was testing it, spotted another issue, it doesn't load splash screens from zars

Nicktamer:
Tried some games and works fine for me.

Just tested with base game without update  and/or dlc

georgemoralis:
> While i was testing it, spotted another issue, it doesn't load splash screens from zars
should be fixed too

StevenMiller123 (approved):
Works in my tests, and in other people's tests. I think this is good to go?

RuskyDed:
> > should be fixed too

Splashes are now displaying correctly. One more thing popped up while i was testing compressed Bloodborne copy, a UNIX-specific one. It applies the loose patch folder ok, applies the mods folder partially, but crashes if trying to enable the in-game debug menu as if no specific fonts are present. This PR could help clarifying it: #4312

StevenMiller123 (changes requested):
Should probably fix the newly located Bloodborne issue before the Bloodborne people get too upset 😅

georgemoralis:
ok let me see

georgemoralis:
ok if you can test if latest fixed the issue

RuskyDed:
> ok if you can test if latest fixed the issue

Fixed. I have no more complaints, for now. Thank you and the original author. Now we need a launcher integration :smiley:

StevenMiller123 (approved):

PR #4756, opened 2026-07-25
video_core: Implement T# min_lod (VK_EXT_image_view_min_lod)

Kravickas:
Traced in Driveclub: it uploads only mips 2-11 of its streaming textures and sets min_lod=2.0 in the T# while mips 0-1 are still empty memory. After writing them it moves the clamp in 1/256 steps down to 0, and back up when streaming out. The garbage on close up textures is the ignored clamp as we sample the empty mips.

driveclub main
driveclub pr

### DIFF
--- a/src/video_core/renderer_vulkan/vk_instance.cpp
+++ b/src/video_core/renderer_vulkan/vk_instance.cpp
@@ -203,8 +203,7 @@ bool Instance::CreateDevice() {
                           vk::PhysicalDevicePrimitiveTopologyListRestartFeaturesEXT,
                           vk::PhysicalDeviceShaderAtomicFloat2FeaturesEXT,
                           vk::PhysicalDeviceWorkgroupMemoryExplicitLayoutFeaturesKHR,
-                          vk::PhysicalDeviceImage2DViewOf3DFeaturesEXT,
-                          vk::PhysicalDeviceImageViewMinLodFeaturesEXT>();
+                          vk::PhysicalDeviceImage2DViewOf3DFeaturesEXT>();
     features = feature_chain.get().features;
 
     const vk::StructureChain properties_chain = physical_device.getProperties2<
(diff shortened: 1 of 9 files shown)


src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -334,6 +335,12 @@ bool Instance::CreateDevice() {
         LOG_INFO(Render_Vulkan, "- sampler2DViewOf3D: {}",
                  image_2d_view_of_3d_features.sampler2DViewOf3D);
     }
+    image_view_min_lod = add_extension(VK_EXT_IMAGE_VIEW_MIN_LOD_EXTENSION_NAME);
+    if (image_view_min_lod) {
+        image_view_min_lod_features =
+            feature_chain.get<vk::PhysicalDeviceImageViewMinLodFeaturesEXT>();
+        LOG_INFO(Render_Vulkan, "- minLod: {}", image_view_min_lod_features.minLod);
```
squidbus: No need to query the features or print for this, if the extension is exposed it must support this feature. Just checking the extension support is fine.

src/video_core/renderer_vulkan/vk_instance.cpp
```diff
@@ -499,6 +506,9 @@ bool Instance::CreateDevice() {
             .image2DViewOf3D = image_2d_view_of_3d_features.image2DViewOf3D,
             .sampler2DViewOf3D = image_2d_view_of_3d_features.sampler2DViewOf3D,
         },
+        vk::PhysicalDeviceImageViewMinLodFeaturesEXT{
+            .minLod = image_view_min_lod_features.minLod,
```
squidbus: Similarly here, can always be `true`.

src/video_core/texture_cache/image.cpp
```diff
@@ -206,14 +206,18 @@ ImageView& Image::FindView(const ImageViewInfo& view_info, bool ensure_guest_sam
     if (ensure_guest_samples && backing->num_samples > 1 != info.num_samples > 1) {
         SetBackingSamples(info.num_samples);
     }
+    ImageViewInfo lookup_info = view_info;
+    if (lookup_info.min_lod != 0 && !instance->IsImageViewMinLodSupported()) {
+        lookup_info.min_lod = 0;
+    }
```
squidbus: This seems unnecessary given we also check `IsImageViewMinLodSupported` before applying a min lod. I think we can leave this unmodified, as I don't think we need to worry about micro-optimizing the view lookup if its unsupported.

squidbus (approved):

PR #4758, opened 2026-07-25
Use pthread* instead of std::mutex in PthreadOnce

StevenMiller123:
mutex goes unused, so this is all that needs fixing for this to be accurate. pthread_once struct is something guest applications can use, so it's relatively valuable to have it done correctly.

### DIFF
--- a/src/core/libraries/kernel/threads/pthread.h
+++ b/src/core/libraries/kernel/threads/pthread.h
@@ -242,7 +242,7 @@ enum class PthreadOnceState : u32 {
 
 struct PthreadOnce {
     std::atomic<PthreadOnceState> state;
-    std::mutex mutex;
+    Pthread* mutex;
 };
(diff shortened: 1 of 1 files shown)


src/core/libraries/kernel/threads/pthread.h
```diff
@@ -242,7 +242,7 @@ enum class PthreadOnceState : u32 {
 
 struct PthreadOnce {
     std::atomic<PthreadOnceState> state;
-    std::mutex mutex;
+    Pthread* mutex;
```
squidbus: Wouldn't it be more accurate for this to be a `PthreadMutexT`?

StevenMiller123: You are correct, my bad

squidbus (approved):

PR #4761, opened 2026-07-26
avplayer: improve sync between audio and video

roamic:
The original library has a complex mechanism of tracking the sync timings. It uses multiple timepoints, like last audio, video timepoint, video hint timepoint, and last data access time. While I do not fully understand most of the algorythms it uses I can make this simple change that improves, but not fully fixes, jittering in some videos. Tested with Kingdom Hearts Birth by Sleep that was fully jittering before this change, now it only jitters occasionally.

### DIFF
--- a/src/core/libraries/avplayer/avplayer_source.cpp
+++ b/src/core/libraries/avplayer/avplayer_source.cpp
@@ -310,6 +310,7 @@ bool AvPlayerSource::Stop() {
     m_video_frames.Clear();
 
     m_last_audio_ts.reset();
+    m_last_data_time.reset();
     m_start_time.reset();
     m_pause_time = {};
     m_pause_duration = {};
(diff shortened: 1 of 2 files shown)


bigol83:
This improves Birth by Sleep but makes 358/2 days video skip like stuttering. Not just the video itself, but the emulation stutters too and you can see framedrops like here, where it is 37 fps

immagine

immagine


You can also see this by just openinig the game collection. Logos slowdowns.

thethunderturner:
No difference for Black Ops 3. Audio still lags behind the video
[CUSA02624.log](https://github.com/user-attachments/files/30389769/CUSA02624.log)

roamic:
> No difference for Black Ops 3. Audio still lags behind the video [CUSA02624.log](https://github.com/user-attachments/files/30389769/CUSA02624.log)

This game doesn't seem to use the PS4 avplayer library, so it won't be affected.

Treshco:
Improves Devil May Cry 4.
It plays now the opening video:
dmc4pr

[CUSA01599.PR.log](https://github.com/user-attachments/files/30390918/CUSA01599.PR.log)

PR #4766, opened 2026-07-28
Fix emulator read issues with WinFSP/rclone storage

Masamune3210:
First time working with the codebase, please let me know if I did anything wrong

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -305,6 +305,7 @@ void Emulator::Run(std::filesystem::path file, std::vector<std::string> args,
 
     const auto resolve_relative_path = [](const std::filesystem::path& path,
                                           const std::filesystem::path& base) {
+        // WinFSP-backed mounts can reject canonical path queries while normal reads still work.
         std::error_code relative_error;
         auto relative_path = std::filesystem::relative(path, base, relative_error);
         if (relative_error) {
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -303,15 +303,38 @@ void Emulator::Run(std::filesystem::path file, std::vector<std::string> args,
         rebase_to_base_game(game_folder);
     }
 
+    const auto resolve_relative_path = [](const std::filesystem::path& path,
+                                          const std::filesystem::path& base) {
+        std::error_code relative_error;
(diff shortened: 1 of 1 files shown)


Niram7777:
a description of the issue would be welcome (I am guessing a filesystem exception ?)
and why lexically_relative fixes the issue

Masamune3210:
> a description of the issue would be welcome (I am guessing a filesystem exception ?) and why lexically_relative fixes the issue

The attached issue explains it so I didn't even think about adding it as a comment in the code. I'll add one when I get home later today from work, sorry

Masamune3210:
Short and sweet of it is that std::filesystem::relative() calls weakly_canonical() on Windows, which blows up whenever a mount is created by using DefineDosDeviceW instead of through Mount Manager. Windows promptly (and in my and a lot of other people's opinion, incorrectly) assumes the device doesn't exist when it tries and fails to get info about the mount from Mount Manager, which gets translated out as ERROR_UNRECOGNIZED_VOLUME

Niram7777 (approved):
thanks, I was going to ask what is the error_code you hit (in case we need different behavior for different error_code in the future)

Masamune3210:
[Debug] <Info> () main.cpp:125 main: Run: ["C:\\Users\\User\\AppData\\Roaming\\shadPS4QtLauncher\\versions\\Pre-release\\shadPS4.exe", "**rclonepath**"]
[Debug] <Critical> (shadPS4:Main) log.cpp:257 Terminate: Exception: weakly_canonical: The volume does not contain a recognized file system.
Please make sure that all required file system drivers are loaded and that the volume is not corrupted.

Masamune3210:
Added a comment in the code to explain the need for the change as well as caught back up with upstream, let me know if anything else is needed

Niram7777 (changes requested):
there is an issue with your rebase

Masamune3210:
Sorry, I'm still not super comfortable with git, I'll have to try to fix it tomorrow after work, it's 1 AM here

Masamune3210:
Had some time, hopefully that fixed it

Niram7777 (approved):

PR #4767, opened 2026-07-28
video_core: Respect depth write state for register clears

cuesta4:
## Problem

`DB_RENDER_CONTROL.DEPTH_CLEAR_ENABLE` controls how depth writes are processed, but the Vulkan backend treated the bit as an unconditional depth attachment clear. Games can leave it enabled during color-only draws with depth writes disabled, which erased existing scene depth and caused particles such as fire to render through occluding geometry.

## Fix

Gate register-driven depth attachment clears on an enabled depth test, enabled guest Z-write, and a valid depth buffer. HTILE metadata clears remain unchanged, preserving legitimate depth clears while preventing color-only draws from destroying previously rendered depth.

## Testing

God of War III Remastered no longer renders particle effects over occluding geometry. Testing is welcomed.

Before:
before

After:
after

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -895,8 +895,10 @@ RenderState Rasterizer::BeginRendering(const GraphicsPipeline* pipeline) {
         auto& image = texture_cache.GetImage(image_id);
 
         const auto slice = image_view.info.range.base.layer;
-        const bool is_depth_clear = regs.depth_render_control.depth_clear_enable ||
-                                    texture_cache.IsMetaCleared(htile_address, slice);
+        const bool is_depth_clear =
+            (regs.depth_render_control.depth_clear_enable && regs.depth_control.depth_enable &&
+             regs.depth_control.depth_write_enable) ||
+            texture_cache.IsMetaCleared(htile_address, slice);
         const bool is_stencil_clear = regs.depth_render_control.stencil_clear_enable;
         texture_cache.TouchMeta(htile_address, slice, false);
         ASSERT(desc.view_info.range.extent.levels == 1 && !image.binding.needs_rebind);


src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -31,6 +31,11 @@ static Shader::PushData MakeUserData(const AmdGpu::Regs& regs) {
     return push_data;
 }
 
+static bool HasRegisterDepthClearWrite(const AmdGpu::Regs& regs) {
+    return regs.depth_render_control.depth_clear_enable && regs.depth_control.depth_enable &&
+           regs.depth_control.depth_write_enable && regs.depth_buffer.DepthValid();
```
squidbus: `depth_enable` and `DepthValid()` seems redundant, they are prerequisites to get to where this is called.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
```diff
@@ -895,8 +900,8 @@ RenderState Rasterizer::BeginRendering(const GraphicsPipeline* pipeline) {
         auto& image = texture_cache.GetImage(image_id);
 
         const auto slice = image_view.info.range.base.layer;
-        const bool is_depth_clear = regs.depth_render_control.depth_clear_enable ||
-                                    texture_cache.IsMetaCleared(htile_address, slice);
+        const bool is_depth_clear =
+            HasRegisterDepthClearWrite(regs) || texture_cache.IsMetaCleared(htile_address, slice);
```
squidbus: Going off of above, seems it would be simpler to just add `depth_write_enable` here, no need to make a new function for it.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
cuesta4: Yeah, DepthValid() is redundant here. depth_enable is not guaranteed, though, because db_desc can also be populated through the (stencil_enable && StencilValid()) path. I kept depth_enable to avoid clearing depth during stencil-only draws.

src/video_core/renderer_vulkan/vk_rasterizer.cpp
cuesta4: Done

squidbus (approved):

PR #4768, opened 2026-07-29
video_core: Emulate scaled min/max blending

meh7an:
## The problem

On GCN the color blender applies blend factors to min/max operations; Vulkan ignores them by spec. A pipeline blending `MIN/MAX` with `src_factor=SRC_COLOR, dst_factor=DST_COLOR` expects `min/max(src*src, dst*dst)` — squared-domain accumulation — but gets plain `min/max(src, dst)`. This is the case the existing "Unimplemented use of min/max blend op with blend factor not equal to one" warning points at. God of War 3 Remastered builds its screen-space light/shadow accumulation with exactly this state, washing out environment materials and leaving its stochastic shadow dither unresolved on screen.

## The fix

Flag the pattern in the graphics pipeline key; the fragment shader squares its RGB output for that attachment at export and the Vulkan blend factors are forced to `ONE`, so fixed-function min/max produces exactly `min/max(src^2, dst^2)` like the hardware does. Clear values 0 and 1 are fixed points of the transform, so the attachment stays coherent end to end. Pipelines without this blend state are unaffected.


raphaelthegreat:
The scaled min/max blending thing I would consider as an acceptable workaround for now since full min/max blending emulation is cumbersome. However the readback thing is a hack and not that clean. The real issue that needs solving is having texture cache properly validate rendering a subrect of an image with same pitch

raphaelthegreat (approved):

src/video_core/renderer_vulkan/vk_pipeline_cache.cpp
```diff
@@ -413,6 +413,14 @@ bool PipelineCache::RefreshGraphicsKey() {
         color_buffer.num_conversion = col_buf.GetNumberConversion();
         color_buffer.export_format = regs.color_export_format.GetFormat(cb);
         color_buffer.swizzle = col_buf.Swizzle();
+
+        const auto& bc = regs.blend_control[cb];
+        color_buffer.blend_self_scale =
```
raphaelthegreat, line 422: When this is true add a warning log so it can be known when this emulation triggers

PR #4770, opened 2026-07-29
video_core: Validate sampled sub-rect images from containing image

meh7an:
## The problem

Games reuse one scratch allocation for GPU reduction chains (auto-exposure and similar), sampling each stage through a smaller descriptor over the same base address and pitch. The texture cache resolves those descriptors to never-rendered sibling images instead of the GPU-written container, so the luminance shader reads stale zeros — in God of War 3 Remastered the measured exposure collapses and bloom floods the entire frame white.

## The fix

A first conservative step toward proper sub-rect validation, following review feedback on [#4768](https://github.com/shadps4-emu/shadPS4/pull/4768#issuecomment-5111691734): when a sampled texture resolves to a never-rendered image whose base address, pitch and format match a GPU-written image of larger extent, validate its contents from that image with a GPU-side copy before sampling. Deliberately narrow — offset-zero, sampling side only — with the generalization (tracked validation, arbitrary offsets, render-side handling) planned as a follow-up.

## Testing

God of War 3 Remastered, before/after below. Tested with "Enable readback linear images" on and readbacks mode at its default (Disabled).

## Before and After
Before 1:
BEFORE1
After 1:
AFTER1

Before 2:
BEFORE2
After 2:
AFTER2

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -573,6 +573,60 @@ ImageId TextureCache::FindImage(ImageDesc& desc, bool exact_fmt) {
         RegisterImage(image_id);
     }
 
+    // Reduction chains sample never-rendered sub-views of a GPU-written container through
+    // the same base address and pitch. Their extent is semantically meaningful, so validate
+    // their texels from the container's top-left region before each use. This has to happen
+    // on the lookup path: once the sub-view exists, its binds resolve by exact match and no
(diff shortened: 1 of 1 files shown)


raphaelthegreat:
This does not belong in FindImage it should be done in ResolveOverlap in the equal address path. Also this code is overly bloated and checks way too many things that don't need to be checked

Heliumrich:
Why before/after image2 show the same fix as #4708 ?

PR #4771, opened 2026-07-29
video_core: Resolve equal-address views with matching row layout

meh7an:
## The problem

Games can sample a multisampled render target as a single-sample texture over the same allocation: the T# describes the same base address and extent, with a guest size of exactly `container size / samples`. The texture cache's exact match rejects this on `guest_size`, so the bind resolves to a never-rendered single-sample sibling image and the shader samples stale zeros instead of the scene.

God of War 3 Remastered reads its 2-sample interlaced scene color this way (a single-sample UNORM view over the sRGB target, gamma decoded in-shader) to feed refraction and glass materials. Everything seen through the frozen floor of the "remains of Ares" room — the body under the ice, the stained-glass windows — renders black.

## The fix

The multisample counterpart of the sub-rect validation proposed in #XXXX, using the same approach: when a sampled, never-rendered single-sample image aliases a GPU-written multisampled container with the same extent and tile mode, a compatible format class, and a guest size of exactly `container / samples`, validate its contents by resolving the container into it. Since the container is redrawn every frame, the view is re-validated once per submission tick and stays marked as derived rather than GPU-authored, so it never goes stale.

Deliberately narrow, consistent with the sub-rect validation: offset-zero aliases, single-level, sampling side only. Independent of #XXXX — the two validations cover different aliasing patterns and either can land alone.

## Testing

God of War 3 Remastered, before/after below — both the remains under the ice and the stained-glass windows recover (same mechanism). Previously working areas of the game were checked for regressions.

Before:
CUSA01715_20260729_102334_896_game_000000
After:
CUSA01715_20260729_103720_016_game_000000

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -281,8 +281,13 @@ std::tuple<ImageId, int, int> TextureCache::ResolveOverlap(const ImageInfo& imag
     if (image_info.guest_address == cache_image.info.guest_address) {
         const u32 lhs_block_size = image_info.num_bits * image_info.num_samples;
         const u32 rhs_block_size = cache_image.info.num_bits * cache_image.info.num_samples;
-        if (image_info.BlockDim() != cache_image.info.BlockDim() ||
-            lhs_block_size != rhs_block_size) {
+        // Uncompressed views with the same pitch and texel size are layout-compatible partial
+        // views of the allocation even when their heights differ, and are resolved below.
+        const bool same_row_layout =
+            !image_info.props.is_block && !cache_image.info.props.is_block &&
+            image_info.pitch == cache_image.info.pitch && lhs_block_size == rhs_block_size;
+        if (!same_row_layout && (image_info.BlockDim() != cache_image.info.BlockDim() ||
+                                 lhs_block_size != rhs_block_size)) {
             // Very likely this kind of overlap is caused by allocation from a pool.
             if (safe_to_delete) {
                 FreeImage(cache_image_id);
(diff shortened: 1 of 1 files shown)


raphaelthegreat:
I dont see how this would work on real hardware. Msaa images pack N samples together so just interpreting the memory with the same format without any samples would result in a garbage image. The only case that is handled is if its interpreted with a new format whose block size == old block size * N. This should be investigated further

raphaelthegreat:
Can you try the following, add a std::max here with 2
https://github.com/shadps4-emu/shadPS4/blob/cc728e1c7c97bb9dca71dea78e9e26a87a220280/src/video_core/amdgpu/resource.h#L249

This should also fix your issue

meh7an:
> Can you try the following, add a std::max here with 2
> 
> https://github.com/shadps4-emu/shadPS4/blob/cc728e1c7c97bb9dca71dea78e9e26a87a220280/src/video_core/amdgpu/resource.h#L249
> 
> This should also fix your issue

Tried the `NumSamples` clamp on a clean checkout, no effect on this title, and the captured descriptors explain why: the reads that break are not MSAA-typed T#s decoding as one sample. I instrumented the title to dump the raw T#s binding this allocation, and it is bound two ways:

- `Color2DMsaa, samples=2, dfmt 10, nfmt 9 (srgb), 1920x1080, pitch 1920, tiling 14` — the ordinary multisampled reads, which already work (and where the clamp is a no-op, since `last_level=1`).
- `Color2D, samples=1, dfmt 10, nfmt 0 (unorm)`, same extent/pitch/tiling, over the same base — a plain single-sample descriptor with the same block size, ~1000 binds per captured window. These are the consumers that read stale zeros, and the clamp can't reach them.

On the layout question you raised: you were right, and I tested it directly. Three reconstructions of what that single-sample view should read: a fragment-interleaved reconstruction (samples packed inside micro tiles) renders garbage, confirming your point about same-format no-samples reinterpretation; a resolve average looks plausible but soft; an exact fragment 0 copy reproduces the intended content precisely. So the fragments of these color surfaces are stored plane-separated, and the single-sample window over the allocation is fragment 0's plane read with ordinary single-sample addressing.

The PR now does exactly that: a small compute kernel copies fragment 0 (no filtering, no averaging, no reinterpretation) into the otherwise never-rendered single-sample sibling before it is sampled, re-validated once per submission tick. Kept deliberately narrow like the sub-rect validation: offset-zero, single-level, 32bpp, sampling side only.

CUSA01715_20260729_122740_169_game_000000

raphaelthegreat:
First of all would appreciate if you spoke directly instead of copy pasting what I presume are LLM answers. Secondly, this is still an issue that needs investigation, as the texture cache job is not to "reproduce the intended content" rather it should always preserve the layout of said data. So this alias does not make sense, maybe there is a resolve in-between

meh7an:
> First of all would appreciate if you spoke directly instead of copy pasting what I presume are LLM answers. Secondly, this is still an issue that needs investigation, as the texture cache job is not to "reproduce the intended content" rather it should always preserve the layout of said data. So this alias does not make sense, maybe there is a resolve in-between

Sorry for LLM usage, it was just to keep the comment writings look "Perfect".
I think I am getting the concept now, it's not about "making it work". Thanks for the notice.

PR #4772, opened 2026-07-29
Rewrite kernel event flags

kalaposfos13:
This fixes random hangs in SotC, and likely in other titles as well. Credit to MathieuARS and GhostRider584 for the initial problem description.


StevenMiller123:
No regressions in what I've tested so far. Shadow of the Colossus hangs appear to be addressed on my end.

No change to Unity event flag hangs (though that's a Windows-only problem, I think the world would end before kalaposfos ever directly fixes a Windows-only issue 😆)

PR #4773, opened 2026-07-30
video_core: preserve coherence across aliased images

cuesta4:
## Problem

PS4 applications can bind the same guest surface memory through image descriptions with different extents or Vulkan-compatible formats. The texture cache represented these bindings as independent host images, so writes performed through one alias were not consistently visible through the others. Asynchronous readbacks could also overwrite newer guest writes with stale GPU data. This caused missing post-processing effects, stale render-target contents, and incorrect composition in titles such as God of War III Remastered.

## Fix

Track compatible images sharing the same guest address as a coherent alias group with an authoritative backing image, pending writer, and generation counters. Synchronize contents through explicit Vulkan image copies and barriers, supporting both complete coverage and compatible partial intersections. Track asynchronous readback pages by generation so invalidated guest pages are not overwritten, and invalidate non-coherent download allocations before host reads. Alias copy compatibility is isolated as a pure operation shared by the runtime and offline tests.

## Testing

Fixes God of War III Remastered's hidden chest effects and hyper-exposure bug.

image
image
before3
after3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1122,6 +1122,8 @@ set(VIDEO_CORE src/video_core/amdgpu/cb_db_extent.h
                src/video_core/renderer_vulkan/host_passes/fsr_pass.h
                src/video_core/renderer_vulkan/host_passes/pp_pass.cpp
                src/video_core/renderer_vulkan/host_passes/pp_pass.h
+               src/video_core/texture_cache/aliasing.cpp
+               src/video_core/texture_cache/aliasing.h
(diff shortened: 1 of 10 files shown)


Randomuser8219:
This does the exact same thing that #4770 does,

cuesta4:
> This does the exact same thing that #4770 does,

No, it doesn't. 

And I had a previous PR (#4708) which fixed the exposure first, but I closed it because this is a better implementation.

PR #4781, opened 2026-07-31
Harden GS copy shader parsing and fix GPU buffer synchronization

Nyces:
These changes primarily target two *THE IDOLM@STER* series titles on the PS4 platform:

*THE IDOLM@STER: Platinum Stars* (CUSA03628, CUSA05706)
*THE IDOLM@STER: Stella Stage* (CUSA07753, CUSA09153)

This changes resolves the problems outlined in compatibility issue :
https://github.com/shadps4-compatibility/shadps4-game-compatibility/issues/836

The issue report noted:
1. Vertex explosion glitches during character facial rendering.
2. Immediate crashes upon entering story sequences within "Produce" mode.

Analysis of `shad_log`, `debugdump`, and `shader dump` revealed:
1. Character facial vertex explosion: The root cause appears to be incomplete buffer synchronization and cache page tracking following GPU writes.
2. Crash after the first dialogue segment: The root cause is an assertion failure in the GS copy shader parser when encountering the game's non-standard copy shader format.

Additionally, I discovered another bug prior to these changes:
1. Crash when starting a "Live" performance in S4U mode.

Analysis showed:
The S4U "Live" crash stems from the same category of GS copy shader issues as the Produce mode crash.

changes made:
`copy_shader.cpp`: Fixed the overly rigid GS copy shader parsing logic; the first instruction is no longer required to be the fixed `s_mov_b32 vcc_hi, #imm`. This is the core fix for both the Produce mode crash and the S4U "Live" crash.

`copy_shader.cpp`: Added scalar constant propagation to enable parsing of more complex offset calculations within copy shaders.

`copy_shader.cpp`: Added support for `BUFFER_LOAD_DWORDX2/X3/X4` (copy shader formats that read multiple dwords in a single operation).

`ring_access_elimination.cpp`: Added a fallback mapping for GS copy shaders to prevent immediate crashes when parsing is incomplete.

`ring_access_elimination.cpp`: Changed strict assertions into warnings and skip operations to prevent unknown mappings from causing the emulator to exit abruptly.

`buffer_cache.cpp`: Preserved tracking for GPU-written buffer pages to ensure that data written by the GPU is not re-read as standard CPU memory. vk_rasterizer.cpp: Implemented read/write access barriers when shaders write to buffers to mitigate GPU buffer synchronization issues.

Results of changes:
1. The facial vertex explosion issue has been resolved.
2. "Produce" mode now functions correctly (though rapidly skipping through story sequences causes the screen to freeze).
3. "Live" mode (S4U) is playable, but the screen freezes midway through; the visuals lock up completely while the audio continues to play.
4. The "First instruction is not s_mov_b32 vcc_hi, #imm" error mentioned in previous issues still appears as a warning in the logs but no longer causes a fatal crash.
5. Gameplay stability remains suboptimal, with occasional crashes to desktop.
6. A friend tested the impact on other titles; currently, the only game observed to be unplayable is *Demon Slayer -Kimetsu no Yaiba- The Hinokami Chronicles 2* (CUSA44357, CUSA50716).

The game is not yet fully playable; significant issues regarding audio stuttering, latency, and screen freezing persist.

To be honest, these changes were generated with the assistance of AI tools and subsequently verified through my own testing.

Feedback and discussion are welcome.

6AFE63F1666941DE4FF09E46B35E780E
ECEB3F1ECA5CCD4DA68C9CCF905FF015
233ED0365296CEF30FA49C3F8392E072
F1E259C7EA16A1343D0A3DEABA13C618
AA1860AE80C7100AF515D60A82B224E4
10DD76C0C8622EB2A786B56B71DBC52F
0F2BCD6C0B6089360FCC48614672F8D8
B6BF5C75CB0D2B44004F86D013A0021E

### DIFF
--- a/src/shader_recompiler/frontend/copy_shader.cpp
+++ b/src/shader_recompiler/frontend/copy_shader.cpp
@@ -1,36 +1,111 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include <array>
+#include <optional>
+
 #include "shader_recompiler/frontend/copy_shader.h"
 #include "shader_recompiler/frontend/decode.h"
 #include "shader_recompiler/ir/attribute.h"
 
 namespace Shader {
 
+static bool IsDwordBufferLoad(const Gcn::GcnInst& inst) {
+    return inst.opcode == Gcn::Opcode::BUFFER_LOAD_DWORD ||
+           inst.opcode == Gcn::Opcode::BUFFER_LOAD_DWORDX2 ||
+           inst.opcode == Gcn::Opcode::BUFFER_LOAD_DWORDX3 ||
+           inst.opcode == Gcn::Opcode::BUFFER_LOAD_DWORDX4;
+}
+
+static u32 GetBufferLoadDwords(const Gcn::GcnInst& inst) {
+    switch (inst.opcode) {
+    case Gcn::Opcode::BUFFER_LOAD_DWORD:
+        return 1;
+    case Gcn::Opcode::BUFFER_LOAD_DWORDX2:
+        return 2;
+    case Gcn::Opcode::BUFFER_LOAD_DWORDX3:
+        return 3;
+    case Gcn::Opcode::BUFFER_LOAD_DWORDX4:
+        return 4;
+    default:
+        return 0;
+    }
+}
+
+static std::optional<s32> ReadConstant(const Gcn::InstOperand& operand,
+                                       const std::array<std::optional<s32>, 256>& sources) {
+    switch (operand.field) {
+    case Gcn::OperandField::ConstZero:
+        return 0;
+    case Gcn::OperandField::SignedConstIntPos:
+        return operand.code - Gcn::OperandFieldRange::SignedConstIntPosMin + 1;
+    case Gcn::OperandField::SignedConstIntNeg:
+        return static_cast<s32>(Gcn::OperandFieldRange::SignedConstIntNegMin) -
+               static_cast<s32>(operand.code) - 1;
+    case Gcn::OperandField::LiteralConst:
+        return static_cast<s32>(operand.code);
+    case Gcn::OperandField::ScalarGPR:
+        if (operand.code < sources.size()) {
+            return sources[operand.code];
+        }
+        return std::nullopt;
+    default:
+        return std::nullopt;
+    }
+}
+
 CopyShaderData ParseCopyShader(std::span<const u32> code) {
     Gcn::GcnCodeSlice code_slice{code.data(), code.data() + code.size()};
     Gcn::GcnDecodeContext decoder;
 
-    constexpr u32 token_mov_vcchi = 0xBEEB03FF;
-    ASSERT_MSG(code[0] == token_mov_vcchi, "First instruction is not s_mov_b32 vcc_hi, #imm");
-
     std::array<s32, 64> offsets{};
     offsets.fill(-1);
 
-    std::array<s32, 256> sources{};
-    sources.fill(-1);
+    std::array<std::optional<s32>, 256> sources{};
 
     CopyShaderData data{};
     auto last_attr{IR::Attribute::Position0};
     while (!code_slice.atEnd()) {
         auto inst = decoder.decodeInstruction(code_slice);
         switch (inst.opcode) {
+        case Gcn::Opcode::S_ADD_U32: {
+            const auto lhs = ReadConstant(inst.src[0], sources);
+            const auto rhs = ReadConstant(inst.src[1], sources);
+            if (lhs && rhs) {
+                sources[inst.dst[0].code] = *lhs + *rhs;
+            }
+            break;
+        }
+        case Gcn::Opcode::S_AND_B32: {
+            const auto lhs = ReadConstant(inst.src[0], sources);
+            const auto rhs = ReadConstant(inst.src[1], sources);
+            if (lhs && rhs) {
+                sources[inst.dst[0].code] = *lhs & *rhs;
+            }
+            break;
+        }
+        case Gcn::Opcode::S_OR_B32: {
+            const auto lhs = ReadConstant(inst.src[0], sources);
+            const auto rhs = ReadConstant(inst.src[1], sources);
+            if (lhs && rhs) {
+                sources[inst.dst[0].code] = *lhs | *rhs;
+            }
+            break;
+        }
         case Gcn::Opcode::S_LSHL_B32: {
-            ASSERT(inst.src[0].field == Gcn::OperandField::SignedConstIntPos &&
-                   inst.src[1].field == Gcn::OperandField::SignedConstIntPos);
-            sources[inst.dst[0].code] =
-                (inst.src[0].code - Gcn::OperandFieldRange::SignedConstIntPosMin + 1)
-                << (inst.src[1].code - Gcn::OperandFieldRange::SignedConstIntPosMin + 1);
+            const auto lhs = ReadConstant(inst.src[0], sources);
+            const auto rhs = ReadConstant(inst.src[1], sources);
+            if (lhs && rhs) {
+                sources[inst.dst[0].code] = *lhs << *rhs;
+            }
+            break;
+        }
+        case Gcn::Opcode::S_LSHR_B32: {
+            const auto lhs = ReadConstant(inst.src[0], sources);
+            const auto rhs = ReadConstant(inst.src[1], sources);
+            if (lhs && rhs) {
+                sources[inst.dst[0].code] = static_cast<s32>(static_cast<u32>(*lhs) >> *rhs);
+            }
             break;
         }
         case Gcn::Opcode::S_MOVK_I32: {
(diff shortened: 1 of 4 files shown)


raphaelthegreat:
The copy shader changes look good to me. The ring access elimination ones Im not sure about, parsing of copy shader should not fail and it emulator should crash when it does. The buffer cache changes dont quite get. Aside from the buffer stuff you are making gpu modified ranges page aligned which dont think will help much and results in flushing more data. Also adding the registration check to fast path can unnecessary reduce how many times it's used

Nyces:
> The copy shader changes look good to me. The ring access elimination ones Im not sure about, parsing of copy shader should not fail and it emulator should crash when it does. The buffer cache changes dont quite get. Aside from the buffer stuff you are making gpu modified ranges page aligned which dont think will help much and results in flushing more data. Also adding the registration check to fast path can unnecessary reduce how many times it's used

Thanks for the reply.

If I understand correctly, your suggestions imply the following:

1. Accept the changes to `copy_shader.cpp`.
2. Remove the fallback mechanism (`BuildFallbackGsCopyData` + `FillMissingGsCopyData`) and let parsing failures trigger an immediate `ASSERT` crash.
3. Remove `CachePageRange` page alignment and mark regions directly using the raw `device_addr` + `size`.
4. Remove the `!IsRegionRegistered` condition to allow more buffers to use the fast path.

However, I believe these are critical changes for these two games and should not be removed. I conducted a simple experiment regarding points 2 through 4 to test the impact of removing them:

Each experiment was tested on one of the target games (*The Idolmaster: Stella Stage*, CUSA09153).

Baseline: `b1820a65` (Fallback retained, `CachePageRange` page alignment retained, `!IsRegionRegistered` retained).

Experiment A (`48892143`): Fallback retained, `CachePageRange` page alignment removed, `!IsRegionRegistered` removed.

Experiment B (`05578a21`): Fallback removed, `CachePageRange` page alignment retained, `!IsRegionRegistered` retained.

Experiment C (`714c4dc6`*): Fallback removed, `CachePageRange` page alignment removed, `!IsRegionRegistered` removed.

*I accidentally deleted the log data for Experiment C and cannot recover it.

Test results:

Baseline `b1820a65`: Everything normal.

Experiment A (`48892143`): Vertex explosion in character facial rendering; S4U Live/Produce cutscenes normal.

2065ABB30A246EAEFE642C86AA40488D


Experiment B (`05578a21`): Character facial rendering normal; S4U Live/Produce cutscenes crash.

580C42FEE9410409E383A3640A5E58D4


Experiment C (`714c4dc6`): Vertex explosion in character facial rendering; S4U Live/Produce cutscenes crash.

Based on the results:
1. The fallback mechanism is linked to the fix for S4U Live/Produce cutscenes.
2. `CachePageRange` page alignment and `!IsRegionRegistered` are linked to the fix for character facial vertex explosions. The `shad_log` and shader dumps for Experiment A and Experiment B are provided below for analysis:

A:
[log_48892143.zip](https://github.com/user-attachments/files/30622354/log_48892143.zip)



B:
[log_05578a21.zip](https://github.com/user-attachments/files/30622368/log_05578a21.zip)



A brief review of the logs reveals the following:
In Experiment A, after removing both the `CachePageRange` page alignment and the `!IsRegionRegistered` check, managed buffers (such as the buffer containing facial vertices) were no longer blocked from the "fast path" by the registration check; the registration condition effectively allowed them to pass through. Simultaneously, the `IsRegionGpuModified` check shifted from a whole-page basis to a precise byte-range basis. This created a scenario where the fast path conditions were all met: the buffer was read-only and small (satisfying initial criteria), the registration check no longer blocked managed regions, and the precise-byte GPU modification check failed to detect modifications in "blind spots" (areas within the same page but not overlapping with recorded modified bytes). Consequently, the system proceeded directly to `stream_buffer.Copy`, copying stale CPU-side memory into the stream buffer and sending it to the rendering pipeline. This resulted in the misalignment of all facial vertex positions and normals and the scattering of triangular faces—essentially an "exploding face" artifact.

The `CachePageRange` page alignment and the `!IsRegionRegistered` check act as two conservative gates in series on the `ObtainBuffer` read-only fast path, working together to block incorrect copies of stale CPU memory: The first gate, `!IsRegionRegistered`, uses `buffer_ranges.Intersects` to determine if the buffer has been registered as a managed region via `CreateBuffer` (facial vertex buffers are always managed and registered during `CreateBuffer`); if registered, this condition evaluates to `false`, short-circuiting the fast path `if` statement and preventing execution of the `stream_buffer.Copy` branch at line 389. The second gate, `CachePageRange`, aligns the requested range to page boundaries before passing it to `IsRegionGpuModified`. Even if the GS pipeline writes to the ring buffer while bypassing the `ObtainBuffer` `is_written` branch (leaving `gpu_modified_ranges` without a record or with records only for other bytes on the same page—creating a byte-level blind spot), the page-level intersection check still detects the modification for the entire page, preventing a modified cached page from being erroneously allowed onto the fast path. With the two-stage gating mechanism in place, the relevant facial regions are first intercepted and kept out of the "fast path" during registration checks; a secondary full-page alignment acts as a fallback, forcing the process onto the "slow path" where `SynchronizeBuffer` retrieves the latest vertex positions and normal vectors from the GPU—thereby preventing the "exploding face" artifact where triangular facets scatter wildly.
(I am not entirely sure if my understanding here is correct.)

In Experiment B, after removing the fallback, the last eight lines of the shad_log (captured as the "Produce" cutscene began) read as follows:

[Render.Recompiler] <Warning> (shadPS4:GpuCommandProcessor) ring_access_elimination.cpp:204 lambda: Missing GS copy mapping for offset 0x1000 in copy shader 0xf3bd17fb
[Render.Recompiler] <Warning> (shadPS4:GpuCommandProcessor) ring_access_elimination.cpp:204 lambda: Missing GS copy mapping for offset 0x1200 in copy shader 0xf3bd17fb
(Repeated four times)
The application crashed immediately afterward.

The assumption that parsing results in only two states—success (complete `attr_map`) or failure (empty `attr_map`)—does not hold true in practice.

If the `attr_map` is fully successful, no crash occurs. In the original version, however, a complete failure or even partial success would trigger an assertion failure and crash the program—at least, that is my understanding. In the legacy code, `ParseCopyShader` derives the ring buffer offset corresponding to each `EXP` instruction using a combination of a single linear pass, static constant folding, and a limited instruction whitelist. It does not actually execute the shader; instead, it relies on a single sequential scan to propagate known constants within the `sources[]` and `offsets[]` arrays. If the offset calculation involves instructions outside the whitelist (such as `S_MUL_I32`, `S_SUB_U32`, `V_*` vector instructions, or conditional branches), or if the `soffset` value cannot be determined (triggering the `if (!soffset) break` condition at the `BUFFER_LOAD` stage), or if the offset for a given `VGPR` remains -1 when accessed by the `EXP` instruction, then that `EXP` is not written to the `attr_map`, resulting in a "hole" for the corresponding key. Commit `b1820a65` reduced the likelihood of such holes by removing the initial instruction assertion, introducing a generic `ReadConstant()` function, expanding the whitelist to include `S_ADD_U32`, `S_AND_B32`, `S_OR_B32`, and `S_LSHR_B32`, and adding support for multi-DWORD loads (`DWORDX2/3/4`). However, it did not eliminate the blind spots inherent in this pattern-matching approach; for instance, the `GS 0xf3bd17fb` shader in *The Idolmaster* still lacks the keys `0x1000` and `0x1200`.

The changes in `b1820a65` involved three steps:

1. First, it checks whether the `attr_map` produced by `ParseCopyShader` is empty, incomplete, or complete.
2. Scenarios:
① If empty: Directly replace it entirely with a complete `attr_map` (where every dword has a mapping) generated by `BuildFallbackGsCopyData`.

② If incomplete: Proceed to `FillMissingGsCopyData` to fill in the gaps. `FillMissingGsCopyData` takes the complete fallback map generated earlier and merges it with the `attr_map` obtained from actual parsing, filling only the holes missed during actual parsing (using `try_emplace` so existing keys aren't overwritten).
(There is a possibility of incorrect filling, though the two *The Idolmaster* games seem unaffected.)

③ If complete: Still go through the process.

3. Proceed to `ForEachInstruction` for `StoreBufferU32` replacement, then finally check for any remaining missing mappings; if any exist, log "Missing GS copy mapping for offset..." and continue.

In Experiment B, we removed the fallback mechanism (`BuildFallbackGsCopyData` + `FillMissingGsCopyData`)—eliminating both the detection and the filling steps—but the warning log code remained, resulting in "Missing offset 0x1000/0x1200" logs. The IR compiled with the incomplete GS (`0xf3bd17fb`) retained two unreplaced ring buffer `StoreBufferU32` instructions, leading to a crash upon entering the "Produce" (live performance) cutscene.

Therefore, I believe the fallback mechanism, `CachePageRange` (page alignment), and `!IsRegionRegistered` checks should not be removed. Together, they ensure correct facial rendering and the proper functioning of S4U Live and cutscenes.

Finally, if there are better approaches, please feel free to suggest them.

As for the startup issue affecting *Demon Slayer -Kimetsu no Yaiba- The Hinokami Chronicles 2* (CUSA44357, CUSA50716), I currently have no leads.

Also, please excuse the fact that I used some AI assistance for such a long reply.

(Please correct me if I've said anything wrong.)

PR #4782, opened 2026-07-31
shader_recompiler: split resource tracking and flatten load from buffer for sharp source

LNDF:
This splits resource tracking pass into resource discovery and patching passes. Then we put flattening pass in the middle.

That way:
1. We discover all resources first, accepting ReadConstBuffer as source and marking it for flattening.
2. Flattening pass flattens all ReadConsts + marked ReadConstBuffers
3. Patch resource instructions, also accepting ReadConstBuffer as source (since it was flattened)


src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -185,7 +202,7 @@ static void VisitPointer(u32 off_dw, IR::Inst* subtree, PassInfo& pass_info,
         c.mov(r10d, ptr[rdi + (src_off_dw << 2)]);
```
LNDF: Do i need to handle buffer stride etc here if ReadConstBuffer?

brad0demx:
On this PR, Call of Duty Cold War crashes in roughly the same point with this new critical 
`[Debug] <Critical> (shadPS4:GpuComm) resource_discover_pass.cpp:141 FindSharpSource: Unreachable code!
Unable to find sharp sources pc=0x70a8`
[CUSA24267.log](https://github.com/user-attachments/files/30599258/CUSA24267.log)

brad0demx:
Killzone Shadow Fall still crashes while loading into offline match, now crashing with
```
[Debug] <Critical> (shadPS4:GpuComm) resource_discover_pass.cpp:141 FindSharpSource: Unreachable code!
Unable to find sharp sources pc=0x1484
```
[CUSA00008.log](https://github.com/user-attachments/files/30599348/CUSA00008.log)

bigol83:
Skyrim immediately crashes with this PR
[CUSA05486.log](https://github.com/user-attachments/files/30606308/CUSA05486.log)

```
[Debug] <Critical> (shadPS4:GpuCommandProcessor) resource_discover_pass.cpp:141 FindSharpSource: Unreachable code!
Unable to find sharp sources pc=0x80
```

StevenMiller123:
Current push fixes the Skyrim regression. Now the game behaves like main again.

LNDF:
> Skyrim immediately crashes with this PR [CUSA05486.log](https://github.com/user-attachments/files/30606308/CUSA05486.log)
> 
> ```
> [Debug] <Critical> (shadPS4:GpuCommandProcessor) resource_discover_pass.cpp:141 FindSharpSource: Unreachable code!
> Unable to find sharp sources pc=0x80
> ```

Should be fixed 


> Killzone Shadow Fall still crashes while loading into offline match, now crashing with
> 
> ```
> [Debug] <Critical> (shadPS4:GpuComm) resource_discover_pass.cpp:141 FindSharpSource: Unreachable code!
> Unable to find sharp sources pc=0x1484
> ```
> 
> [CUSA00008.log](https://github.com/user-attachments/files/30599348/CUSA00008.log)

Try latest commit


> On this PR, Call of Duty Cold War crashes in roughly the same point with this new critical `[Debug] <Critical> (shadPS4:GpuComm) resource_discover_pass.cpp:141 FindSharpSource: Unreachable code! Unable to find sharp sources pc=0x70a8` [CUSA24267.log](https://github.com/user-attachments/files/30599258/CUSA24267.log)

Try latest commit

bigol83:
> > Skyrim immediately crashes with this PR [CUSA05486.log](https://github.com/user-attachments/files/30606308/CUSA05486.log)
> > ```
> > [Debug] <Critical> (shadPS4:GpuCommandProcessor) resource_discover_pass.cpp:141 FindSharpSource: Unreachable code!
> > Unable to find sharp sources pc=0x80
> > ```
> 
> Should be fixed

It is fixed for me

StevenMiller123:
This PR causes regressions in Uncharted: The Nathan Drake Collection and Bloodborne. I've reported these directly to @LNDF, just mentioning them here so people know these issues are already known.

bigol83:
Far Cry 5 shows this on this PR

```
[Render.Vulkan] <Info> (shadPS4:GpuCommandProcessor) vk_pipeline_cache.cpp:602 CompileModule: Compiling fs shader 0xc42b4153
[Debug] <Critical> (shadPS4:GpuCommandProcessor) breadth_first_search.h:139 DominatingBreadthFirstSearch: Assertion Failed!
Unable to deduce correct finding
```

Main build

```
[Render.Vulkan] <Info> (shadPS4:GpuCommandProcessor) vk_pipeline_cache.cpp:602 CompileModule: Compiling fs shader 0xc42b4153
[Debug] <Critical> (shadPS4:GpuCommandProcessor) resource_tracking_pass.cpp:489 TrackSharp: Assertion Failed!
Unable to deduce sharp source
```

crashes on both cases without even showing a logo

Final Fantasy XV crashes with 
```
[Render.Vulkan] <Info> (shadPS4:GpuCommandProcessor) vk_pipeline_cache.cpp:602 CompileModule: Compiling cs shader 0x66f5202
[Debug] <Critical> (shadPS4:GpuCommandProcessor) ir_emitter.cpp:770 CompositeConstruct: Unreachable code!
Composite construct with 0 elements, only 2-4 are supported
```


Main 
```
[Kernel.Vmm] <Info> (LuminousTask1) memory.cpp:246 sceKernelMapNamedDirectMemory: out_addr = 0x33e1e0000
[Debug] <Critical> (shadPS4:GpuCommandProcessor) resource_tracking_pass.cpp:411 FindSharpSources: Unreachable code!
Bindless sharp access detected pc=0x584
```

brad0demx:
COD Cold War now hangs rather than crashing. 
```
[Render.Recompiler] <Error> (shadPS4:GpuComm) flatten_extended_userdata_pass.cpp:546 ComputeOffset: Unexpected instruction for offset computation, BitCastU32F32
[Render.Recompiler] <Error> (shadPS4:GpuComm) flatten_extended_userdata_pass.cpp:589 VisitPointer: Failed to compute offset for SRT walker
[Lib.Http] <Info> (bdHTTPWorker(PSN)) http.cpp:1915 sceHttpWaitRequest: epoll id=2 returned 0 events after wait
[Lib.Http] <Info> (bdHTTPWorker(PSN)) http.cpp:1915 sceHttpWaitRequest: epoll id=2 returned 0 events after wait
[Lib.Http] <Info> (bdHTTPWorker(PSN)) http.cpp:1915 sceHttpWaitRequest: epoll id=2 returned 0 events after wait
```
[CUSA24267.log](https://github.com/user-attachments/files/30908995/CUSA24267.log)

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -159,20 +191,388 @@ namespace Shader::Optimization {
...
+
+#define POP_ABORT_ON_FAILURE(expr)                                                                 \
+    if (!(expr)) {                                                                                 \
+        c.add(rsp, 8);                                                                             \
+        return false;                                                                              \
+    }
+
+static bool EmitComputeOffsetIAdd32(Xbyak::CodeGenerator& c, Xbyak::Reg32 reg, PassInfo& pass_info,
+                                    IR::Inst* inst) {
+    if (inst->AreAllArgsImmediates()) {
```
raphaelthegreat: I feel this check might not be needed as constant propagation has already run before this and should have constant folded the operation, have you noticed this in some code?

src/shader_recompiler/ir/srt_gvn_table.h
raphaelthegreat: I feel this also suffers from the logic bug in main. If a shader has something like that
```
addr = ...
read_const(addr)
...
addr += 4
read_const(addr)
```
then the GVN might consider them identical since it skips over the intermediate add sees same addr. I have not confirmed this so worth checking. Another question is if we really need GVN pass, how much de-duplication actually happens in practice

src/shader_recompiler/ir/breadth_first_search.h
```diff
@@ -71,4 +73,91 @@ auto BreadthFirstSearch(Value value, Pred&& pred) -> std::invoke_result_t<Pred,
...
+            }
+        }
+    }
+    if (findings.empty()) {
+        return std::nullopt;
+    }
+
+    // Perform dominance analysis on findings and eliminate ones that don't pass
+    // If a finding is dominated by another, the former can be eliminated.
+    size_t num_findings = findings.size();
```
raphaelthegreat: I think this should be a separate function in dominance.h that takes the findings array by reference and prunes it

src/shader_recompiler/ir/breadth_first_search.h
```diff
@@ -71,4 +73,91 @@ auto BreadthFirstSearch(Value value, Pred&& pred) -> std::invoke_result_t<Pred,
     return BreadthFirstSearch(value.InstRecursive(), pred);
 }
 
+template <typename Instruction, typename Pred>
+auto DominatingBreadthFirstSearch(Instruction* inst, const IR::Block& current_parent, bool deep,
```
raphaelthegreat: The BFS header is getting crowded so worth doing a more holistic review.
The normal BFS
* Only finds 1 instruction its looking for and stops
* Keeps searching until it finds it without any way to stop it

This BFS
* Can find multiple sources if presents
* Is more selective where it searches and deep option stops it from going thru phis

A review of all BFS usages in recompiler could be useful to deduce if its possible to make a unified function that handles everything (also could use std::expected as predicate return to allow user to specify if they want to stop search on the current path or entirely etc)

src/shader_recompiler/ir/breadth_first_search.h
```diff
@@ -71,4 +73,91 @@ auto BreadthFirstSearch(Value value, Pred&& pred) -> std::invoke_result_t<Pred,
...
+                                  Pred&& pred) -> std::invoke_result_t<Pred, Instruction*> {
+    if (std::optional result = pred(inst)) {
+        return result;
+    }
+
+    boost::container::small_vector<IR::Inst*, 8> visited, findings;
+    std::queue<IR::Inst*> queue;
```
raphaelthegreat: Small optimization; since we spam  this function a lot in resource discover might be useful to either pass this as argument in function so resource discover can maintain a queue or make this static and clear it on function entry, to avoid reconstructing the object on every call and associated heap allocations

src/shader_recompiler/ir/passes/resource_pass.h
```diff
@@ -0,0 +1,166 @@
...
+
+#include "shader_recompiler/ir/opcodes.h"
+#include "shader_recompiler/ir/value.h"
+
+namespace Shader::Optimization {
+
+struct ResourceDiscovery {
+    IR::Inst* user{};
+    IR::Block* user_block{};
+    const IR::Inst* sharp_source{};
```
raphaelthegreat: Should this be an array of sharp sources (for each dword of the sharp, could be more in the future if batched loads are added like ReadConstU32X4)?

src/shader_recompiler/ir/passes/resource_discover_pass.cpp
```diff
@@ -0,0 +1,184 @@
...
+    ASSERT(marked_count == count);
+}
+
+void DiscoverBufferSharp(IR::Block& block, IR::Inst& inst, ResourceDiscoveryList& sharp_usages) {
+    IR::Inst* handle = inst.Arg(0).InstRecursive();
+    if (handle->AreAllArgsImmediates()) {
+        // For inmediates, add a sharp usage with null sharp source.
+        sharp_usages.emplace_back(ResourceDiscovery{&inst, &block, nullptr});
+    } else {
+        IR::Inst* buffer_handle = handle->Arg(0).InstRecursive();
```
raphaelthegreat: For buffers handle is a composite that contains all 4 dwords of V#, can just call FindSharpSource for each, store them in an array (as mentioned in other review comment) and in resource patching verify offsets of flattened instructions are contiguous. If they are not, additional logic could be introduced (for example using gather intrinsic like _mm_i32gather_epi32 with a slim abstraction could be suitable)

src/shader_recompiler/ir/passes/resource_discover_pass.cpp
```diff
@@ -0,0 +1,184 @@
...
+            auto [sampler_handle, found] =
+                CheckDisableAnisoLod0Pattern(sampler->Arg(0).InstRecursive());
+            sampler_sharp_source = FindSharpSource(sampler_handle, block);
+            disable_aniso = found;
+        }
+    }
+
+    if (sharp_source && sharp_source->GetOpcode() == IR::Opcode::ReadConstBuffer) {
+        const auto texture_flags = inst.Flags<IR::TextureInstInfo>();
+        const auto is_r128 = texture_flags.is_r128.Value();
```
raphaelthegreat: Mentioning here too, image sample ir op should be updated to hold 2 composites for the low and high part of the T# and do the same as buffers. The r128 check should be entirely a frontend thing and leave the high composite null

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -159,20 +191,388 @@ namespace Shader::Optimization {
 
 namespace {
 
-static inline void PushPtr(Xbyak::CodeGenerator& c, u32 off_dw) {
+static inline u16 GetFlatbufOffset(const IR::Inst* inst) {
+    if (inst->GetOpcode() == IR::Opcode::ReadConstBuffer) {
+        auto inst_info = inst->Flags<IR::BufferInstInfo>();
+        return inst_info.flatbuf_off_dw;
+    }
+    return inst->Flags<u16>();
```
raphaelthegreat: for good measure check if its IR::Opcode::ReadConst and add an UNREACHABLE at the end

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -245,51 +655,104 @@ void FlattenExtendedUserdataPass(IR::Program& program) {
...
 
     for (auto r_it = program.post_order_blocks.rbegin(); r_it != program.post_order_blocks.rend();
          r_it++) {
         IR::Block* block = *r_it;
         for (IR::Inst& inst : *block) {
-            if (inst.GetOpcode() == IR::Opcode::ReadConst) {
-                if (!inst.Arg(1).IsImmediate()) {
-                    LOG_WARNING(Render_Recompiler, "ReadConst has non-immediate offset");
-                    continue;
+            if (inst.GetOpcode() == IR::Opcode::ReadConst ||
```
raphaelthegreat: condition can be inverted to reduce nesting

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -245,51 +655,104 @@ void FlattenExtendedUserdataPass(IR::Program& program) {
...
 
-                IR::Inst* ptr_composite = inst.Arg(0).InstRecursive();
-
-                const auto pred = [](IR::Inst* inst) -> std::optional<IR::Inst*> {
+                queue.push(offset.InstRecursive());
+                while (!queue.empty()) {
+                    IR::Inst* inst{queue.front()};
+                    queue.pop();
+
+                    if (inst->GetOpcode() == IR::Opcode::ReadConstBuffer) {
```
raphaelthegreat: Should  this stop when it encounters a Phi? iirc the dynamic offset emit cannot handle them atm though not sure what the plan is to add them in the future. Also this is another BFS variant that searches the entire tree without stopping, would be good to unify it with the rest

DarkBoy08:
Regression on Uncharted The Nathan Drake Collection. It doesn't matter whether DMA is enabled or not, it crashes at the shader compilation stage. But I decided to launch it without DMA.
Main
[Uploading Main-log! Without DMA.txt…]()
Freezes while compiling shaders
PR
[PR log Without DMA.txt](https://github.com/user-attachments/files/31077782/PR.log.Without.DMA.txt)
The game doesn't start AT ALL.

LNDF:
> Regression on Uncharted The Nathan Drake Collection. It doesn't matter whether DMA is enabled or not, it crashes at the shader compilation stage. But I decided to launch it without DMA. Main Uploading Main-log! Without DMA.txt… Freezes while compiling shaders PR [PR log Without DMA.txt](https://github.com/user-attachments/files/31077782/PR.log.Without.DMA.txt) The game doesn't start AT ALL.

Thanks for the report. This regression is already know and will be delt with before merging. There are more games thatare affected.

First will apply the suggested changes on the review though.

src/shader_recompiler/ir/srt_gvn_table.h
LNDF: For other insts that are not GetUserData, CompositeConstructU32x2, ReadConst and ReadConstBuffer GVN will associate an incrementing number to the instruction itself. For those 4 instructions it will insteed assign an incrementing number to a hash of the numbers given to the argument instructions. 

Because of this, if you have
```
%10 = ReadConst %9 ...
%11 = IAdd32 %10 ...
%12 = ReadConst %11 ...
```

Both readconsts should have different number since the argument is different.

I figured out this by reading the gvn table .h file so i might have understood something wrong. Is there a game that is affected by this in main?

src/shader_recompiler/ir/passes/resource_pass.h
LNDF: Maybe in the future it is a good idea. but right now flattening pass will order readconsts in order.

Resource tracking pass will use the first source instruction and, since they are are contiguous and ordered by offset in flatbuf.

src/shader_recompiler/ir/passes/resource_discover_pass.cpp
LNDF: Do you think this is in scope for this PR or out of scope?

IMO its a better idea to do this in another PR where we also include the refactor of image instructions to include both the hish and low vectors.

LNDF: Do you think this is in scope for this PR or out of scope?

IMO its a better idea to do this in another PR where we also include the refactor of image instructions to include both the hish and low vectors.

src/shader_recompiler/ir/passes/resource_discover_pass.cpp
LNDF: Answered in the previous comment

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
LNDF: Currently will go through all possible previous blocks. Then this will log an error when generating the x64 code.

src/shader_recompiler/ir/breadth_first_search.h
LNDF: Done

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
LNDF: Done.

Tho min and max afaik are not constant folded so i left them

src/shader_recompiler/ir/breadth_first_search.h
LNDF: Yes. But is dominance.h the correct place? That is for frontend and what we are trying to separate is ir stuff.

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
LNDF: Done

src/shader_recompiler/ir/breadth_first_search.h
raphaelthegreat: Well this code prunes sources based on dominance analysis I feel it matches better in dominance than a what is supposed to be a generic BFS algorithm header impl

src/shader_recompiler/ir/passes/resource_pass.h
raphaelthegreat: Well anything is possible, this was mostly proposed for validation. A shader could load lower 128bits and upper ones from separate memory, it could use constants for upper half maybe. There is probably nothing out there that does this tho, so ig you could skip it like main does

src/shader_recompiler/ir/passes/resource_discover_pass.cpp
raphaelthegreat: Sure it could be possible. You should add an assertion at least that the ReadConstBuffer insts you find have a use in the composite handle of the buffer opcode

src/shader_recompiler/ir/srt_gvn_table.h
raphaelthegreat: I think the example given was incorrect, what about this
```
addr = user_data()
if (cond0)
  addr0 = addr + 4;
else
  addr0 = addr;
read_const(addr0);

if (cond1)
   addr1 = addr + 8
else
   addr1 = addr;
read_const(addr1);
```
Each ReadConst is gonna have a phi as address where 1 argument is the same SSA value directly from user data and another is an IAdd. The bfs is gonna return the addr SSA in both cases resulting in same value number

src/shader_recompiler/ir/breadth_first_search.h
LNDF: Done

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
LNDF: Done

src/shader_recompiler/ir/srt_gvn_table.h
LNDF: Done. We now handle phis as any other inst (so assugn a new gvn value for each phi)

LNDF: Actualy, just updated it to use the same logic as in flattening pass

src/shader_recompiler/ir/breadth_first_search.h
LNDF: So BFS can be unified like this.

A single BFS function that returns an std::optional<T> where T is the value type of the invoke result of the predicate, which should return std::expected<T, bool>. bool here is wheter to do BFS recursively on the args of the instruction in question.
Predicate can also have an optional arg that is an push_inst() functor which will allow the predicate to push arbitarry insts to the bfs queue (required for this pr).

For BFS function itself i have 2 options. Have a BFS and BFSDominant function (where the first will return the first match and the other will do the dominance analysis) or have a single BFS function which will take a bool argument on wheter to return the first match or do dominance analysis.

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -159,20 +191,381 @@ namespace Shader::Optimization {
...
+        c.cmova(reg, dword[rsp]);
+        c.add(rsp, 8);
+    }
+    return true;
+}
+
+static bool EmitComputeOffsetUMax32(Xbyak::CodeGenerator& c, Xbyak::Reg32 reg, PassInfo& pass_info,
+                                    IR::Inst* inst) {
+    if (inst->AreAllArgsImmediates()) {
+        c.mov(reg, std::max(inst->Arg(0).U32(), inst->Arg(1).U32()));
```
baggins183: Can this be done in const folding instead?

LNDF: Yes. It should

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
```diff
@@ -159,20 +191,381 @@ namespace Shader::Optimization {
...
+
+#define POP_ABORT_ON_FAILURE(expr)                                                                 \
+    if (!(expr)) {                                                                                 \
+        c.add(rsp, 8);                                                                             \
+        return false;                                                                              \
+    }
+
+static bool EmitComputeOffsetIAdd32(Xbyak::CodeGenerator& c, Xbyak::Reg32 reg, PassInfo& pass_info,
+                                    IR::Inst* inst) {
+    ASSERT(!inst->AreAllArgsImmediates());
```
baggins183, line 236: Just glancing at it it looks like these binop functions can be refactored into one since they share a lot of logic

src/shader_recompiler/ir/srt_gvn_table.h
baggins183: I think there are still issues.
I think it'd be more accurate if you searched for the source that contains one of these 4 instructions, but  set "source" to the operand to "inst" that the search succeeded on, not the return val of the predicate which could be deeper into the search.

Or just take the next Value Number for phi's. I forget what exactly I was going for (I dont think value numbering is necessarry for correctness, just deduplicating the srt program and number of bindings).

baggins183: Also I think you can add all the insts you've implemented in ComputeOffset to the switch below so they can be dedup'd instead of just taking the next VN.

baggins183: > I think it'd be more accurate if you searched for the source that contains one of these 4 instructions, but set "source" to the operand to "inst" that the search succeeded on

This might also be an issue for how the flatten pass builds up the user lists, how it searches for ptr_lo. But it might be fine, not sure.

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
LNDF: Now min and max are also handled in constant folding

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
LNDF: It is now done

src/shader_recompiler/ir/passes/flatten_extended_userdata_pass.cpp
LNDF: While all ops share a general structure (if arg 1 is imm, then if 2 is imm, and then if none is imm), the logic for each case is slightly different for each case.

For example, ISub32 has a negate when arg 0 is immediate or none of them are immediate, but not when arg 1 is immediate.
Another example: left/right shift have all 3 paths follow a different logic, as well as min/max.
Additionaly, while logic for bitwise ops seems similar, different overloads fo xbyak functions are used for each path (like an immediate or mem)

src/shader_recompiler/ir/srt_gvn_table.h
```diff
@@ -50,13 +74,9 @@ class SrtGvnTable {
...
-        case IR::Opcode::Phi: {
-            const auto pred = [](IR::Inst* inst) -> std::optional<IR::Inst*> {
-                if (inst->GetOpcode() == IR::Opcode::GetUserData ||
-                    inst->GetOpcode() == IR::Opcode::CompositeConstructU32x2 ||
-                    inst->GetOpcode() == IR::Opcode::ReadConst ||
-                    inst->GetOpcode() == IR::Opcode::ReadConstBuffer) {
+        if (inst->GetOpcode() == IR::Opcode::Phi) {
+            const auto pred = [this](IR::Inst* inst) -> std::optional<IR::Inst*> {
+                if (IsArgHashInst(inst)) {
                     return inst;
```
baggins183: Actually I meant the predicate should stay the same (4 opcodes) and only change the switch cases below. Adding all the others would increase the chance of choosing the wrong edge for a phi.

For phis specifically I'd loop over the (edge, edge_value) pairs and test the predicate for each edge_value, and then  return the value number of edge_value if it passes the predicate (this will account for any adds/other evaluatable insts in between). If multiple edge_value's pass the predicate then it would be good to have a warning since ambiguous. Hopefully that makes sense

src/shader_recompiler/ir/passes/resource_discover_pass.cpp
LNDF: An assert added, but only for buffer insts.
For image insts the reafctor mentioned earlier to have all sharp parts referenced in image insts (in this pr or another)

src/shader_recompiler/ir/srt_gvn_table.h
LNDF: > I think it'd be more accurate if you searched for the source that contains one of these 4 instructions, but set "source" to the operand to "inst" that the search succeeded on, not the return val of the predicate which could be deeper into the search.

I don't understand this clearly. Is this the same as what you have written in your latest review comment?

> Or just take the next Value Number for phi's. I forget what exactly I was going for (I dont think value numbering is necessarry for correctness, just deduplicating the srt program and number of bindings).

This can't be done. Flattening pass and gvn should have same search logic. If the search logic differs, it could happen that an instrucion with same offset and base is considered a duplicate of another instruction that is identical, but flattening pass will replace it with the previous instruction in the PointerUserList.. Then since the VN is different an offset will not be set when duplicate instructions are handled at the end.

src/shader_recompiler/ir/srt_gvn_table.h
LNDF: If I understand correctly, what you mean is to test the predicate with the original 4 insts, then find out which edge that instruction comes from and use that edge inst as a gvn right?

src/shader_recompiler/ir/srt_gvn_table.h
baggins183: > Is this the same as what you have written in your latest review comment?

Yeah, basically I want it to match strategy of the search in the flatten pass. SRT branches need to originate from ReadConsts/GetUserData's (and I guess now ReadConstBuffer's), so to reduce the risk of choosing bad phi edges we should only consider these. Also we may be able to get rid of composite construct in the predicate now that I think of it.

src/shader_recompiler/ir/srt_gvn_table.h
baggins183: Yes, before "source" would end up being the ReadConst/GetUserData/etc directly, but we should be using the value number of the phi's operand instead since there could be any number of instructions (like the new insts you evaluate) in between the phi and the ReadConst/GetUserData/etc.

src/shader_recompiler/ir/srt_gvn_table.h
baggins183: I hadnt looked at this part though, I dont think my previous suggestions addressed it.

```
addr = user_data()
if (cond0)
  addr0 = addr + 4;
else
  addr0 = addr;
v = read_const(addr0);
```

I think this is more involved. Youd have to do something like:

```
addr = user_data()
if (cond0)
  v = read_const(addr + 4);
else
  v = read_const(addr);
```

and this might get bad with loops

src/shader_recompiler/ir/srt_gvn_table.h
LNDF: This is now done

src/shader_recompiler/ir/passes/readconst_address_add_transform.cpp
```diff
@@ -0,0 +1,68 @@
...
+                addr->SetArg(0, lo->Arg(0));
+                addr->SetArg(1, hi->Arg(0));
+                for (auto [user, operand] : addr->Uses()) {
+                    ASSERT(user->GetOpcode() == IR::Opcode::ReadConst && operand == 0);
+                    IR::IREmitter ir{*block, IR::Block::InstructionList::s_iterator_to(inst)};
+                    IR::Value offset = user->Arg(1);
+                    if (offset.IsImmediate()) {
+                        user->SetArg(1, ir.Imm32(offset.U32() + (add_offset.U32() >> 2u)));
+                    } else {
+                        auto shift = ir.ShiftRightLogical(IR::U32(add_offset), ir.Imm32(2u));
```
raphaelthegreat: it has been checked above that add_offset is a constant, so ig the shift in ir is not needed

src/shader_recompiler/ir/passes/readconst_address_add_transform.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+void ReadConstAddressAddTransform(IR::Program& program) {
+    // This handles the following pattern by combining the addition with the offset
+    // %82 = IAdd32 %65, #28816
+    // %83 = ULessThan32 %82, %65
+    // %84 = SelectU32 %83, #1, #0
+    // %85 = IAdd32 %66, %84
+    // %86 = CompositeConstructU32x2 %82, %85
+    // %87 = ReadConst (flags=0x0)  %86, #0
+    for (auto& block : program.blocks) {
```
raphaelthegreat: This can be done inline in flattening pass too, there isnt much reason to pay the cost of iterating all IR again

src/shader_recompiler/frontend/dominance.cpp
```diff
@@ -0,0 +1,41 @@
+// SPDX-FileCopyrightText: Copyright 2026 shadPS4 Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include <queue>
+#include "shader_recompiler/frontend/dominance.h"
+
+namespace Shader::Gcn {
+
+bool IsCfgBlockDominatedBy(const Shader::Gcn::Block* maybe_dominator,
```
raphaelthegreat: This can be deleted after a rebase

src/shader_recompiler/frontend/dominance.h
```diff
@@ -0,0 +1,43 @@
...
+#include "shader_recompiler/ir/value.h"
+
+namespace Shader::Gcn {
+
+bool IsCfgBlockDominatedBy(const Shader::Gcn::Block* maybe_dominator,
+                           const Shader::Gcn::Block* block, const Shader::Gcn::Block* dest_block);
+
+template <typename Container>
+    requires std::same_as<typename Container::value_type, IR::Inst*>
+void EliminateNonDominantInstructions(Container& insts, const IR::Block& current_parent) {
```
raphaelthegreat: Can be deleted too

src/shader_recompiler/ir/passes/resource_discover_pass.cpp
```diff
@@ -0,0 +1,188 @@
...
+        }
+        break;
+    }
+    ASSERT(marked_count == count);
+}
+
+void DiscoverBufferSharp(IR::Block& block, IR::Inst& inst, ResourceDiscoveryList& sharp_usages) {
+    IR::Inst* handle = inst.Arg(0).InstRecursive();
+    if (handle->AreAllArgsImmediates()) {
+        // For inmediates, add a sharp usage with null sharp source.
```
raphaelthegreat: minor spelling mistake :kek:

src/shader_recompiler/ir/srt_gvn_table.h
```diff
@@ -50,24 +74,41 @@ class SrtGvnTable {
...
+                case IR::Opcode::ReadConstBuffer:
                     return inst;
+                default:
+                    return std::nullopt;
                 }
-                return std::nullopt;
             };
-            IR::Inst* source = IR::BreadthFirstSearch(inst, pred).value();
-            vn = GetValueNumber(source);
+            boost::container::small_vector<std::pair<IR::Inst*, IR::Inst*>, 2> src_inst_map;
```
raphaelthegreat: should GVN pass no longer handle phi nodes? ReadConst pointers should no longer have phis now unless shader actually does conditional loads. Dynamic offset stuff might need phi node if its in a loop. Maybe just giving each phi a unique value number is sufficient for this?

StevenMiller123 (approved):
Has both mine and @raphaelthegreat's approval, should be good to merge based on what I've tested.

PR #4784, opened 2026-08-01
Fix cube array layer in FixCubeCoords

Kravickas:
Port the Cube Array fix from #4558

Cube arrays sample the wrong layer for slice >= 1. GCN packs the array layer
into the face coordinate as slice * 8 + face_id, but the 2D array image the
cube is sampled from has 6 faces per cube, so the layer needs remapping to
slice * 6 + face_id.


squidbus:
Does this fix a particular case? Just want to make sure as I would have assumed there would just be padding layers in the array image or something.

Kravickas:
It does, but I would like some additional tests.

inFAMOUS FL - https://github.com/shadps4-emu/shadPS4/pull/4558#issuecomment-4687661699
Ratchet & Clank
Dragon Quest - https://discord.com/channels/1080089157554155590/1258386683691274361/1529958445036343336

Randomuser8219:
> It does, but I would like some tests.
> 
> inFAMOUS FL - [#4558 (comment)](https://github.com/shadps4-emu/shadPS4/pull/4558#issuecomment-4687661699) Ratchet & Clank Dragon Quest - https://discord.com/channels/1080089157554155590/1258386683691274361/1529958445036343336

Sure can. I also noticed more UE games, such as Sackboy: A Big Adventure having cubemap issues.

squidbus:
FWIW the code changes themselves look okay to me.

StevenMiller123:
Before anyone gets their hopes up, this doesn't change anything for BB and it's shadows. (I figure that was probably a given, but still worth noting for the BB fans)
image

squidbus:
> Before anyone gets their hopes up, this doesn't change anything for BB and it's shadows. (I figure that was probably a given, but still worth noting for the BB fans)

BB will need full fix with proper cube sampler to eliminate the seams (I'm not sure there's a good way to emulate the seamless mapping otherwise).

The approach from the previous PR is generally in the right direction for that, just had some problems with when to use cube compatible images and where the cube face calculations belong.

Randomuser8219:
image

I _think_ this improved Ratchet & Clank? The crates used to have jagged colors, but I'm in an area where cubemap issues are less obvious, and I don't feel like starting the game anew.

Xcedf:
Cubemaps are still fixed with thia as 4558 in Infamous First light

squidbus (approved):

PR #4792, opened 2026-08-02
voice: Return ORBIS_FAIL in sceVoiceReadFromOPort to fix BO3 hang

Modified17:
Fixes a startup hang in BO3 where the game gets stuck polling for microphone input. Returning ORBIS_FAIL bypasses the loop. 

With this fix, BO3 successfully reaches in-game (tested via null gpu).

### DIFF
--- a/src/core/libraries/voice/voice.cpp
+++ b/src/core/libraries/voice/voice.cpp
@@ -93,7 +93,7 @@ s32 PS4_SYSV_ABI sceVoicePausePortAll() {
 
 s32 PS4_SYSV_ABI sceVoiceReadFromOPort() {
     LOG_ERROR(Lib_Voice, "(STUBBED) called");
-    return ORBIS_OK;
+    return ORBIS_FAIL;
 }
 
 s32 PS4_SYSV_ABI sceVoiceResetPort() {


src/core/libraries/voice/voice.cpp
```diff
@@ -93,7 +93,7 @@ s32 PS4_SYSV_ABI sceVoicePausePortAll() {
 
 s32 PS4_SYSV_ABI sceVoiceReadFromOPort() {
     LOG_ERROR(Lib_Voice, "(STUBBED) called");
-    return ORBIS_OK;
+    return ORBIS_FAIL;
```
StevenMiller123, line 96: There's probably a better error code to use here, libraries don't usually return the value ORBIS_FAIL.

DanielSvoboda: maybe test this 0x804e0801 = The voice library has not been initialized.

StevenMiller123: I'd argue a not-initialized error is probably just as inaccurate. For a stub error code, probably best to do 0x804E0803 (ORBIS_VOICE_ERROR_GENERAL) or 0x804E0808 (ORBIS_VOICE_ERROR_RESOURCE_INSUFFICIENT).
Best solution would be testing the function to see how it behaves when there's no data to read, and stubbing whatever that behavior is. I can do that myself if needed.

georgemoralis: int sceVoiceReadFromOPort(
    uint32_t ops,
    void *data,
    uint32_t *size
);  maybe you can just *size=0;

PR #4793, opened 2026-08-02
Allow override of `protobuf_FORCE_FETCH_DEPENDENCIES`

ngoquang2708:
`flatpak-builder` disallow internet access in build-time.

This allow us to build `protobuf`'s dependencies (abseil in this case) with `clang` to avoid linking issue.

Ref: https://github.com/flathub/net.shadps4.shadPS4/pull/77.


squidbus:
@georgemoralis Is this something we needed to download at build time to begin with, feel like we could've just submodule'd this ourselves.

georgemoralis:
not sure how that switch works ;/

ngoquang2708:
`protobuf_FORCE_FETCH_DEPENDENCIES` force `protobuf` to fetch and build `abseil` by it self.
Currently we hardcoded `protobuf_FORCE_FETCH_DEPENDENCIES`  to ON.
This PR allow us to override that.
@squidbus suggestion is good to me.

ConnorG512:
Nix also dissalows internet access through building. I worked around it also by pulling the abseil source before the build and overwriting the source path when CMake calls its fetch content. If we can use a submodule for it then that step could probably be avoided completely.

misterj05:
Gentoo by default also disallows internet access in the build process.

I didn't find this as much of an issue as I am building with system libraries and already have the needed deps installed, what I found more of a problem was that my system abseil by default is compiled with GCC whereas shadps4 prefers clang which leads to the linking error mentioned in the PR description when compiling the vendored protobuf.

If I were to disable system libraries now there's the issue of the build process having no internet to fetch the dependencies which would fix the issue above.

It's likely @ConnorG512's submodule suggestion would fix this stalemate. And of course a way to make sure the system abseil isn't used.

squidbus:
https://github.com/shadps4-emu/shadPS4/pull/4872

PR #4794, opened 2026-08-02
Fix DCB copy buffer overflow implement missing libkernel libSceLibcInternal exports

wikinto:
## What this fixes

Chasing a crash in inFAMOUS Second Son (CUSA00004) turned up one hard crash with a
memory-safety bug sitting behind it. On the way there I ran into a handful of missing HLE
functions and a few log lines that point at problems which don't exist. Sections below, roughly
by subsystem.

### 1. `video_core`: DCB copy buffer overflow and use-after-free

`Liverpool::CopyCmdBuffers` hands out `std::span`s into `GpuQueue::dcb_buffer` and
`ccb_buffer`. Both are `std::vector`s, `reserve()`d once to `GfxReservedSize` (2 MiB) and rewound
only in `SubmitDone()`. That makes 2 MiB a budget for the whole frame, shared by every submit in
it, not a per-submit budget.

The budget doesn't even fit a single legal command buffer.
`sceGnmSubmitCommandBuffersForWorkload` accepts DCBs up to `0x3ffffc` bytes, which is twice
the entire frame reserve. inFAMOUS Second Son hits it every time:

```
liverpool.cpp:1181 CopyCmdBuffers: Assertion Failed!
dcb copy buffer out of reserved space
```

The assert hides a worse bug than the one it reports. The lines right after it `resize()` the
vector, so once asserts are compiled out the vector reallocates and every span already handed to
the GPU thread's suspended `ProcessGraphics` coroutines dangles.

Both vectors are now a chunked arena (`std::deque<std::vector<u32>>`). Each chunk gets
`reserve()`d once when it is `emplace_back()`ed and never grows past that capacity, so no chunk
reallocates and spans handed out earlier stay valid. `Reset()` won't free or shrink a chunk, and
that is deliberate: `sceGnmSubmitDone` calls `WaitGpuIdle()`, which waits on `submission_lock`
rather than `num_submits`, and then goes ahead anyway when `!IsGpuIdle()`. A suspended coroutine
can still be holding a span into any chunk at that point. The arena also has its own leaf mutex,
because `Allocate()` and `Reset()` were previously reachable under two different locks
(`GpuQueue::m_access` and `submit_mutex`). That is a data race on the containers, not just stale
data.

Only reachable with `copyGPUBuffers` turned on, which is not the default.

### 2. `libkernel`: `sceKernelTruncate` reported success while doing nothing

`sceKernelTruncate` (`WlyEA-sLDf0`) had no implementation, so the aerolib stub returned
`ORBIS_OK` and claimed the call had worked. Games that write savedata through libSceFios2 can
corrupt it silently that way. Only the fd-based `sceKernelFtruncate` existed.

This adds path-based `posix_truncate`/`sceKernelTruncate`, written the same way as
`posix_rename` and `posix_unlink`: it resolves through `MntPoints::GetHostPath` and reuses an
already-open handle via `HandleTable::GetFile`, so truncating a file the guest currently has open
works. `EROFS`, `ENOENT`, `EISDIR`, `EINVAL` and `FileType::Device` are handled. The `SetSize()`
return value in `posix_ftruncate`, which was thrown away, is checked now too.

`getrusage` (`hHlZQUnlxSM`) is exported as well. Its stub returned 0 without touching the caller's
`rusage`, so guest `clock()` and `times()` were reading uninitialized memory.

### 3. `libSceLibcInternal`: exports imported by the game's own `libSceFios2.prx`

None of these were exported, so the aerolib stub returned 0. Inside Fios2 that means NULL
pointers and empty strings.

- `malloc` and `free` now go through the application heap installed by
  `_sceKernelRtldSetApplicationHeapAPI` if there is one, the same way
  `Linker::AllocateTlsForThread` already does it, and fall back to the host allocator otherwise.
  This needs a public accessor for `Linker::heap_api`.
- `printf`, `vprintf` and `vsnprintf` are built on the vendored formatter that was already in
  tree. That formatter had `maxlen` hardcoded to `SIZE_MAX`, so every write path was unbounded.
  Any format expansion over 255 bytes could overflow `printf_ctx`'s `char buffer[256]`, and
  `vsnprintf(buf, 0, …)`, which is the standard way to measure a length, did
  `strcpy(s, nullptr)`. `maxlen` is a real parameter now, threaded through properly; the
  `strcpy`s are gone; and formatting writes into the caller's buffer rather than a temporary
  sized by a guest-controlled `n`. The `snprintf` export that already existed had the same
  defect and is fixed along with the rest. Return values are the count that would have been
  written, per C99, not the truncated one.
- `time` is implemented. I left `gmtime`, `gmtime_s`, `localtime` and `mktime` unresolved on
  purpose. They take a guest `struct tm` and there is no confirmed layout for it in tree, so a
  wrong guess at the field order corrupts guest memory without saying anything.

### 4. `sceSystemServiceParamGetString` had the wrong signature

It was declared with no parameters at all, yet registered against the real three-argument ABI,
and it returned `ORBIS_OK` without writing the caller's buffer. The game then read uninitialized
memory as the value. It takes `(param_id, char* buf, size_t buf_size)` now, checks its arguments,
and always writes `buf`.

### 5. Log severities that send people chasing problems that aren't there

- `scePadInit` logged `(STUBBED)` at ERROR, and it isn't a stub; it sets state that the rest of
  the library reads. In the log that started all this, a DualSense was working fine, gyro, accel
  and speaker included, while this line read as a failure.
- The RenderDoc/Wayland notice is advisory. Surface creation carries on right after it.
- The `default:` arm for program headers in `module.cpp` fired at ERROR for `PT_INTERP`, SCE
  Comment and SCE Library Version, all of which are informational.
- Trophy extraction logged the cause you can act on (missing key, `trp.cpp`) at INFO and the
  symptom you can't at ERROR, once per `.trp` file. That's now a single warning, gated on the
  existing `KeyManager::HasKeys()`.
- Around 19 `LOG_ERROR(Kernel_Fs, "error = {}", …)` sites logged an errno and nothing else, which
  made the failures impossible to diagnose (`posix_stat` only logs its path at DEBUG). Each one
  carries the path or fd now. A plain `ENOENT` from `stat` drops to DEBUG, since probing for a
  file that isn't there is something applications do all the time.

## Testing

Built with clang 19 (Release) and run against inFAMOUS Second Son (CUSA00004). It used to die on
the `CopyCmdBuffers` assert; it gets past it now.

That test doesn't exercise `sceKernelTruncate` (it only fires on Fios2 writes), `getrusage` or
`sceSystemServiceParamGetString`. `vprintf` the title did call.

## What this doesn't solve

- The arena never gives memory back. It keeps chunks at the frame high-water mark for the life of
  the process, because releasing them is the use-after-free I'm fixing. The floor rises from
  4 MiB (2 x 2 MiB) to 8 MiB (2 x `ChunkDwords`).
- `Reset()` rewinds to chunk 0, so the next frame writes over bytes a suspended coroutine might
  still be reading. The pointer stays valid, the contents don't. Upstream does exactly this today
  (`dcb_buffer_offset = 0` on the same reserved vector), so it isn't a regression, but it isn't
  fixed either.
- `printf_ctx` still cuts off at 255 characters while reporting the full length.
- `internal_malloc` chooses its allocator per call, so a pointer allocated before the guest
  installs its heap and freed afterwards goes to the wrong one. The window is small, and
  `linker.cpp` already has the same asymmetry.
- `posix_truncate` falls back to `EIO` on archive-backed mounts, because `GetHostPath` knows
  nothing about the dump-to-temp-file path those use.
- `getrusage` sits behind `#ifndef _WIN64` and passes the guest pointer straight through to the
  host. The FreeBSD and Linux x86-64 `struct rusage` layouts match, but on Windows the NID stays
  unregistered.
- `sceSystemServiceParamGetString` answers `SystemName` with the user name. On hardware that's
  the console name, and shadPS4 has no separate notion of one.


src/core/libraries/kernel/time.cpp
```diff
@@ -525,6 +525,21 @@ s32 PS4_SYSV_ABI sceKernelConvertUtcToLocaltime(time_t time, time_t* local_time,
     return ORBIS_OK;
 }
 
+#ifndef _WIN64
```
StevenMiller123: I would rather we keep the function defined for Windows, but had the ifdef inside the function to log a stub error.

src/core/libraries/libc_internal/libc_internal_io.cpp
```diff
@@ -464,8 +480,21 @@ s32 PS4_SYSV_ABI internal_fclose(OrbisFILE* file) {
     return 0;
 }
 
+// NID "2sWzhYqFH4E" (_Stdout) was previously registered here as a LIB_OBJ pointing at a
```
StevenMiller123: Please remove this comment that was very clearly left by your LLM that has absolutely nothing to do with the changes you've made here.

src/core/libraries/libc_internal/libc_internal_time.h
```diff
@@ -0,0 +1,17 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
```
StevenMiller123: New files should not have 2024 copyright.

src/core/libraries/libc_internal/libc_internal_io.cpp
```diff
@@ -25,6 +25,22 @@ s32 PS4_SYSV_ABI internal_snprintf(char* s, u64 n, VA_ARGS) {
...
+
+s32 PS4_SYSV_ABI internal_printf(VA_ARGS) {
+    VA_CTX(ctx);
+    return printf_ctx(&ctx);
+}
+
+s32 PS4_SYSV_ABI internal_vprintf(const char* format, Common::VaList* arg) {
+    char buffer[256];
+    s32 result = vsnprintf_ctx(buffer, sizeof(buffer), format, arg);
+    std::printf("%s", buffer);
```
StevenMiller123: I don't see a good reason for us to unconditionally log buffer contents on each vprintf call. This should be removed, or at least swapped to LOG_TRACE so it doesn't cause performance issues for people who are using the emulator normally.

src/core/libraries/libc_internal/libc_internal_memory.cpp
```diff
@@ -30,12 +34,36 @@ s32 PS4_SYSV_ABI internal_memcmp(const void* s1, const void* s2, size_t n) {
...
 }
 
+void* PS4_SYSV_ABI internal_malloc(u64 size) {
+    // The guest may install its own heap via _sceKernelRtldSetApplicationHeapAPI when it has
+    // allocations must go through it so that the matching free() sees memory it owns
+    const auto heap_api = Common::Singleton<Core::Linker>::Instance()->HeapAPI();
+    if (heap_api && heap_api->heap_malloc) {
+        return heap_api->heap_malloc(size);
+    }
+    return std::malloc(size);
```
StevenMiller123, line 45: If you are going to implement LibcInternal's malloc, you need to do it properly, or add a stub log to indicate that it's not properly implemented.

src/core/libraries/libc_internal/libc_internal_memory.cpp
```diff
@@ -30,12 +34,36 @@ s32 PS4_SYSV_ABI internal_memcmp(const void* s1, const void* s2, size_t n) {
...
+void PS4_SYSV_ABI internal_free(void* ptr) {
+    if (ptr == nullptr) {
+        return;
+    }
+    const auto heap_api = Common::Singleton<Core::Linker>::Instance()->HeapAPI();
+    if (heap_api && heap_api->heap_free) {
+        heap_api->heap_free(ptr);
+        return;
+    }
+    std::free(ptr);
```
StevenMiller123, line 58: My comment about your malloc applies here too.

src/core/libraries/libc_internal/libc_internal_time.cpp
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
```
StevenMiller123: New files should not have 2024 copyright.

src/core/libraries/libc_internal/libc_internal_time.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+s64 PS4_SYSV_ABI internal_time(s64* tloc) {
+    const std::time_t result = std::time(nullptr);
+    if (tloc != nullptr) {
+        *tloc = static_cast<s64>(result);
+    }
+    return static_cast<s64>(result);
+}
+
+// NOTE: gmtime/gmtime_s/localtime/mktime are intentionally not implemented here. They all
```
StevenMiller123: Clean up after your LLM please. This comment is unnecessary, all that's needed is someone intending to actually do what they're supposed to and decompile the library.

src/core/libraries/libc_internal/libc_internal_time.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <ctime>
+
+#include "core/libraries/libs.h"
+#include "libc_internal_time.h"
+
+namespace Libraries::LibcInternal {
+
+s64 PS4_SYSV_ABI internal_time(s64* tloc) {
+    const std::time_t result = std::time(nullptr);
```
StevenMiller123: If the API for time is supposed to be different, we need to check for other differences too. Please properly decompile this function instead of just passing it to the host.

src/core/libraries/libc_internal/printf.h
```diff
@@ -409,11 +409,15 @@ static inline size_t _ftoa(out_fct_type out, char* buffer, size_t idx, size_t ma
 #endif // PRINTF_SUPPORT_FLOAT
 
 // internal vsnprintf
-static inline int _vsnprintf(out_fct_type out, char* buffer, const char* format,
+// `maxlen` is a real bound on `buffer` supplied by the caller (it used to be hardcoded to
+// SIZE_MAX here, which made every write path below unbounded). `out()` (_out_buffer) already
```
StevenMiller123: This issue seems to be something the Vita3K team introduced in their variant of this file (which judging from the copyright, is the version we adopted as a base). I would recommend taking inspiration from how the original tiny printf works, instead of whatever your LLM is trying to do here.

Either that or look into why the Vita3K messed with this whole maxlen logic in the first place, as there might be a valid reason.

src/core/libraries/libc_internal/printf.h
```diff
@@ -732,32 +736,50 @@ static inline int _vsnprintf(out_fct_type out, char* buffer, const char* format,
...
+    // Bound the write to sizeof(buffer): _vsnprintf still returns the true, untruncated length.
+    int result = _vsnprintf(_out_buffer, buffer, sizeof(buffer), format, &ctx->va_list);
+    // Safe: _vsnprintf's termination step always places a NUL at an index less than
+    // sizeof(buffer), so buffer is a valid, in-bounds C string here regardless of the true
+    // (possibly larger) formatted length.
     printf("%s", buffer);
     return result;
 }
 
+// NOTE: unreachable dead code (no LIB_FUNCTION/LIB_OBJ in this codebase calls fprintf_ctx).
```
StevenMiller123: Please remove these LLM-generated comments bloating the file.

src/core/libraries/system/systemservice.cpp
```diff
@@ -1976,8 +1978,38 @@ s32 PS4_SYSV_ABI sceSystemServiceParamGetInt(OrbisSystemServiceParamId param_id,
...
+        LOG_ERROR(Lib_SystemService, "buf is null");
+        return ORBIS_SYSTEM_SERVICE_ERROR_PARAMETER;
+    }
+    if (buf_size == 0) {
+        LOG_ERROR(Lib_SystemService, "buf_size is 0");
+        return ORBIS_SYSTEM_SERVICE_ERROR_PARAMETER;
+    }
+    switch (param_id) {
+    case OrbisSystemServiceParamId::SystemName: {
+        std::string name = "shadPS4";
```
StevenMiller123: Judging from the name of this parameter ID, it sounds as if this gets the system name. On PS4, this is configurable, but always defaults to PS4- followed by some numbers.

Instead of writing a user name here, this should default to shadPS4, and maybe have a separate config entry for configuring it.

src/core/libraries/system/systemservice.cpp
```diff
@@ -1976,8 +1978,38 @@ s32 PS4_SYSV_ABI sceSystemServiceParamGetInt(OrbisSystemServiceParamId param_id,
...
+            buf[0] = '\0';
+            return ORBIS_SYSTEM_SERVICE_ERROR_PARAMETER;
+        }
+        snprintf(buf, buf_size, "%s", name.c_str());
+        break;
+    }
+    default:
+        LOG_ERROR(Lib_SystemService, "param_id {} unsupported!", u32(param_id));
+        buf[0] = '\0';
+        return ORBIS_SYSTEM_SERVICE_ERROR_PARAMETER;
```
StevenMiller123: Unsupported param_ids should be stubbed to success, until we've handled all the ones that show up.

src/video_core/amdgpu/liverpool.h
```diff
@@ -186,12 +186,65 @@ struct Liverpool {
...
+        }
+
+        // Rewinds the arena for the next frame. This must never free or shrink a chunk:
+        // a coroutine suspended from a previous submission may still hold a span pointing
+        // into any chunk, even one other than the first, since SubmitDone() can run while
+        // num_submits != 0. So every chunk is retained for the lifetime of the process at
+        // its frame high-water mark, and only the logical fill state is reset here.
+        // Because each chunk's capacity() is fixed once by the single reserve() at
+        // emplace_back() time, and resize() never grows a chunk past that capacity,
+        // no chunk ever reallocates, so previously-returned spans remain valid forever.
```
StevenMiller123: Comments should tell us what a function does, not what a function does not do.
As I've mentioned in other review comments, please clean these up.

src/video_core/renderer_vulkan/vk_platform.cpp
```diff
@@ -88,8 +88,8 @@ vk::SurfaceKHR CreateSurface(vk::Instance instance, const Frontend::WindowSDL& e
         }
     } else if (window_info.type == Frontend::WindowSystemType::Wayland) {
         if (EmulatorSettings.IsRenderdocEnabled()) {
-            LOG_ERROR(Render_Vulkan,
-                      "RenderDoc is not compatible with Wayland, use an X11 window instead.");
+            LOG_WARNING(Render_Vulkan,
```
StevenMiller123: This should stay an error, or perhaps be turned into an assert.
Unless something with Wayland has changed recently, hitting this case consistently causes a crash during `createWaylandSurfaceKHR` calls.

src/core/libraries/kernel/file_system.cpp
```diff
@@ -859,14 +864,92 @@ s32 PS4_SYSV_ABI posix_ftruncate(s32 fd, s64 length) {
...
+            return -1;
+        }
+        if (!host->SetSize(length)) {
+            *__Error() = POSIX_EIO;
+            return -1;
+        }
+        return ORBIS_OK;
+    }
+
+    Common::FS::IOFile new_file(host_path, Common::FS::FileAccessMode::ReadWrite);
```
StevenMiller123, line 937: truncate shouldn't create new files. If the file to modify doesn't exist, then we should set *__Error() to POSIX_ENOENT and return -1.

StevenMiller123 (changes requested):
There are a few good changes here, but almost none of this can be accepted as-is in my opinion.

src/core/libraries/libc_internal/libc_internal_memory.cpp
wikinto: I added stub log

src/core/libraries/libc_internal/libc_internal_memory.cpp
wikinto: I added stub log

src/core/libraries/system/systemservice.cpp
```diff
@@ -1976,8 +1977,34 @@ s32 PS4_SYSV_ABI sceSystemServiceParamGetInt(OrbisSystemServiceParamId param_id,
     return ORBIS_OK;
 }
 
-int PS4_SYSV_ABI sceSystemServiceParamGetString() {
-    LOG_ERROR(Lib_SystemService, "(STUBBED) called");
+s32 PS4_SYSV_ABI sceSystemServiceParamGetString(OrbisSystemServiceParamId param_id, char* buf,
```
georgemoralis, line 1980: does any game uses this? haven't seen it

StevenMiller123: I've seen it used by some stuff

wikinto: This wasn't found via a game log, it came from reading the registration table, the function was declared with no parameters while LIB_FUNCTION registers it against the real 3-argument ABI, and it returned ORBIS_OK without writing to the callers buffer, so any game that does call it reads uninitialized stack believing it got a string

src/core/libraries/kernel/file_system.cpp
```diff
@@ -292,7 +292,7 @@ s32 PS4_SYSV_ABI posix_close(s32 fd) {
 s32 PS4_SYSV_ABI sceKernelClose(s32 fd) {
     s32 result = close(fd);
     if (result < 0) {
-        LOG_ERROR(Kernel_Fs, "error = {}", *__Error());
+        LOG_ERROR(Kernel_Fs, "fd = {} error = {}", fd, *__Error());
```
StevenMiller123, line 295: For file descriptor logging to be useful, I think you should add an info log in `open` that logs the fd it returns.

PR #4797, opened 2026-08-03
Add a bunch of signal stuff qemu needs

kalaposfos13:
the amount of incorrect defines the OpenOrbis toolchain has is astounding btw


georgemoralis:
at the very least it will break linux (ultimate goal) so merging

kalaposfos13:
It's not breaking Linux, Stephen's comment was only about the fact that pthread_kill works better on Windows on a technicality, but apart from his one single test case, it doesn't affect anything else. Also, that only takes one line to fix it the boring way, and the not boring way is much more work, but makes signal emulation perfect on every platform except for Windows.

PR #4798, opened 2026-08-04
Issues observed in The Order: 1886 intro sequence

TdlifePeng:
## Overview

These are some issues observed while getting through the intro sequence.

Some issues were bypassed temporarily just to continue execution. This is not intended as a patch proposal; the main purpose is to document the observations and hopefully get feedback from people more familiar with the hardware behavior.

## 1. Invalid texture descriptors

Affected shaders:

* `cs_0x00000000a7b66f58_0`
* `cs_0x00000000efaaab2b`

These shaders contain a large number of invalid texture descriptors. Entering the game triggers an assertion failure.

I did not find a good way to handle these invalid bindings, so I temporarily skipped all texture bindings for these shaders to continue testing.

## 2. VSM shadow generation

Affected shader:

* `cs_0x000000008ea0ef5b_0`
* `cs_0x000000009846aaff_0`

This shader converts the depth buffer into VSM data.

It appears to rely on hardware HTILE data, but the bound HTILE buffer does not seem to contain the expected contents. The result is that shadows disappear, as if the generated data is based on far surfaces.

For testing purposes, I temporarily enabled another internal path that directly reads the depth buffer. This makes the result look correct, but I am not sure what the intended implementation should be.

A proper solution may require a better understanding of HTILE generation and usage.

## 3. TAA ghosting

Affected shader:

* `cs_0x0000000092311aca_0`

This TAA implementation uses 4x sampling, while the bound depth buffer appears to be 2x.

The mismatch produces incorrect results and severe ghosting.

It is not clear whether this is the actual cause or just a related symptom. The current change only makes this path compatible with 2x depth buffers.

## 4. Missing large bubbles in the pool

Several possible issues were observed:

### a. Dynamic GDS offset

Affected shader:

* `cs_0x00000000873b1ade_0`

The GDS offset is constant-folded during translation, but the runtime value can change.

### b. Indirect SSBO binding

Affected shader:

* `vs_0x00000000c75dd317_0`

This shader renders the large bubbles in the new game pool.

The particle data is stored in `ssbo10` and accessed through indirect `v#` addressing. The correct `v#` value is not known when creating the binding.

### c. addr64 support

The shader requires `addr64` support to read the correct offsets from push constants.

### d. Runtime constructed static v#

The shader constructs and accesses static `v#` values at runtime. They are currently not bound, so the correct data cannot be read.

This likely requires BDA support.

## Other observations

After passing the intro sequence, another assertion occurs due to buffers aligned to 2-byte boundaries.

## Notes

These are exploratory findings rather than confirmed root causes. Any feedback on the expected hardware behavior or correct implementation approach would be appreciated.

---

Claude Code / deepseek-v4-pro

Screenshot 2026-08-04 165209

Screenshot 2026-08-04 165615

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1019,6 +1019,7 @@ set(SHADER_RECOMPILER src/shader_recompiler/profile.h
                       src/shader_recompiler/ir/passes/hull_shader_transform.cpp
                       src/shader_recompiler/ir/passes/identity_removal_pass.cpp
                       src/shader_recompiler/ir/passes/inject_clip_distance_attributes.cpp
+                      src/shader_recompiler/ir/passes/insert_ssbo_store_load_barrier.cpp
                       src/shader_recompiler/ir/passes/ir_passes.h
                       src/shader_recompiler/ir/passes/lower_buffer_format_to_raw.cpp
                       src/shader_recompiler/ir/passes/lower_fp64_to_fp32.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -978,6 +978,8 @@ set(SHADER_RECOMPILER src/shader_recompiler/profile.h
                       src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
                       src/shader_recompiler/backend/spirv/emit_spirv_integer.cpp
                       src/shader_recompiler/backend/spirv/emit_spirv_logical.cpp
+                      src/shader_recompiler/backend/spirv/emit_spirv_per_vertex.cpp
+                      src/shader_recompiler/backend/spirv/emit_spirv_per_vertex.h
                       src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.cpp
                       src/shader_recompiler/backend/spirv/emit_spirv_quad_rect.h
                       src/shader_recompiler/backend/spirv/emit_spirv_select.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -828,6 +828,7 @@ set(COMMON src/common/logging/classes.h
            src/common/number_utils.cpp
            src/common/memory_patcher.h
            src/common/memory_patcher.cpp
+           src/common/hack_features.cpp
            ${CMAKE_CURRENT_BINARY_DIR}/src/common/scm_rev.cpp
            src/common/scm_rev.h
            src/common/key_manager.cpp
--- a/src/common/bit_array.h
+++ b/src/common/bit_array.h
@@ -5,6 +5,7 @@
 
 #include <array>
 #include <cstddef>
+#include <type_traits>
 #include "common/types.h"
 
 #ifdef __AVX2__
@@ -22,6 +23,12 @@ class BitArray {
     static constexpr size_t WORD_COUNT = N / BITS_PER_WORD;
     static constexpr size_t WORDS_PER_AVX = 4;
     static constexpr size_t AVX_WORD_COUNT = WORD_COUNT / WORDS_PER_AVX;
+
+    // summary: one bit per data word. Type picks the smallest that fits.
+    static_assert(WORD_COUNT <= 64, "BitArray summary requires WORD_COUNT <= 64 (N <= 4096)");
+    using SummaryType = std::conditional_t<(WORD_COUNT <= 8), u8,
+                        std::conditional_t<(WORD_COUNT <= 16), u16,
+                        std::conditional_t<(WORD_COUNT <= 32), u32, u64>>>;
 
 public:
     using Range = std::pair<size_t, size_t>;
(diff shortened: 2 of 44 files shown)


dante3732:
Hello, i have crash all time in this place. 
shadPS4_2026_08_17_18_04_35_712

End of chapter 8.

TdlifePeng:
Could you share a save file from around that point? I'm still investigating the smoke particle issue in Chapter 1, so it'll take me a while to reach that point myself.

dante3732:
> Could you share a save file from around that point? I'm still investigating the smoke particle issue in Chapter 1, so it'll take me a while to reach that point myself.

Maybe it only AMD GPUs issue. Someone already completed game.

[CUSA00035.zip](https://github.com/user-attachments/files/31149204/CUSA00035.zip)

TdlifePeng:
After loading the save, I hit a deadlock between the game and render threads shortly after the dialogue starts, caused by memory access protection and a GPU readback. Not sure if this is the same issue you're seeing.

dante3732:
> After loading the save, I hit a deadlock between the game and render threads shortly after the dialogue starts, caused by memory access protection and a GPU readback. Not sure if this is the same issue you're seeing.

 I just launched the game with render set to null, so I managed to finish watching the cutscene without the video and start Chapter  9. But after I turn off the null setting, I keep getting crashes after the game loads.

TdlifePeng:
On an AMD 395, it deadlocked a few times, but eventually made it through the tunnel and reached "Chapter IX – Mistrust" without any other major issues.
Could you check the crash information at the end of shad_log.txt and see if it gives any clues?

dante3732:
> On an AMD 395, it deadlocked a few times, but eventually made it through the tunnel and reached "Chapter IX – Mistrust" without any other major issues.
> Could you check the crash information at the end of shad_log.txt and see if it gives any clues?

[shad_log.txt](https://github.com/user-attachments/files/31171346/shad_log.txt)

RainKikyou:
I used this PR build to finishied the game. The biggest impact was the caching issue (it was extremely significant, causing stuttering every few steps). Starting with build 6d27ee8, the performance penalty for the two readback options wasn't as severe as before.However, the frame rate in these two scenarios was quite poor, generally only around 20 FPS.（My PC ：AMD 9700X + RTX 3080 20G + 16Gx2 DDR5 6000MHz CL36）
lots of fire effects and gunfights
image
Assassinating the gatekeeper in the courtyard
image
In other scenarios, it can maintain a stable 30 frames per second throughout.If you disable the two readback options, you can achieve a consistent FPS of over 60 (but the graphics quality will be very poor).

Main problems with graphics
1.black spots on the face（The black spots on the werewolf are likely caused by the same issue.）
image
image
2.Shadow problem (I don't know if the shadow issues in The Order: 1886 were fixed in PR4839, but the shadow issues in GOW3 have definitely been fixed.）
image
3.Smoke texture incorrectly （This problem should be the same as the one with Knack.）
image

If these three graphics issues are fixed, and the cache generation problem is improved, then the game will soon be playable.

TdlifePeng:
`vk::Result = ErrorOutOfHostMemory` is a pretty vague error and can have many possible causes. Does the save at the end of Chapter 8 produce the same error?

Also, where did the previous screenshot come from? I haven't seen that particular camera angle.

The game needs DMA enabled; for example, without it, the bubbles in the water at the beginning of the game are invisible.

The shadow issue requires precise readbacks, as the CPU needs to read the scene depth.

I'm still investigating the smoke particle issue. The GTX 1060 I'm using for frame captures doesn't support barycentric, so I need to fix that first.

Shader translation and compilation stuttering has always been an issue. Precompilation is the usual solution, but that's probably difficult for an emulator. Maybe someday someone could build a shader cache hub to collect, merge, and distribute shader caches.

raphaelthegreat:
I will work on upstreaming these slowly there is a lot of useful stuff here that can work

dante3732:
> Does the save at the end of Chapter 8 produce the same error?

Yes, I had a crash between the end of chapter 8 and the start of chapter 9.

Linear524:
On my HW second chapter crashes with strange error in the log:

[shad_log.txt](https://github.com/user-attachments/files/31654305/shad_log.txt)

georgemoralis:
> On my HW second chapter crashes with strange error in the log:  [shad_log.txt](https://github.com/user-attachments/files/31654305/shad_log.txt)

ArrayPtr is not implemented indeed but have you tried in main? that branch is kinda outdated

Linear524:
> > On my HW second chapter crashes with strange error in the log:  [shad_log.txt](https://github.com/user-attachments/files/31654305/shad_log.txt)
> 
> ArrayPtr is not implemented indeed but have you tried in main? that branch is kinda outdated

I will try Main a bit later, thanks :)

StevenMiller123:
This PR is not mergable. The findings are useful, so I'd suggest making an issue discussing the findings you've made that are still relevant.
In my opinion, it's preferrable to have an issue discussing these sorts of problems instead of a primarily AI-generated PR.

PR #4799, opened 2026-08-04
shader_recompiler: Lower user clip planes to clip distances

junminlee2004:
Fixes corrupt water reflections in inFAMOUS Second Son by implementing GCN user clip planes (PA_CL_UCP_* / UCP_ENA in PA_CL_CLIP_CNTL, previously decoded but never applied). The issue was that none of the geometry below the water surface was being clipped by the user clip plane (UCP0 in rdoc) as it does on hardware. This is problematic because the rendering trick the game uses requires nothing to be rendered below the water's surface (UCP0) to produce accurate reflections. The key is that Vulkan only accepts the user clip plane's evaluation from shader-written clip distance so the evaluation has to be lowered into the vertex shader as clip distances for clipping to occur. The remaining changes keep the descriptor layout and shader cache consistent to avoid crashes.

Before (no clipping - you can see the submerged dock pillars portrayed as reflections):

After (clipping active - no submerged pillar reflections):

The game utilizes a rendering trick where a second camera is placed under the third person camera beneath the user clip plane (submerged below sea level). Any light traveling straight to this submerged camera simulates light bouncing off of the surface of the water and hitting your eye (the third person camera) so the video output from this submerged camera is displayed as a simulated water reflection. Problem is, if the game fails to clip objects below the water, the reflections will show the giant submerged dock pillars (like a submarine cam) as reflections in addition to the reflections for objects above the water.


roamic:
Someone needs to check Bloodborne water with this since it uses clip distance (that we also had to emulate for nvidia).

georgemoralis:
no bloodborne volunteers?

Kyoskii:
Breaks Bloodborne reflections/layering, this is incorrect to how it appears on real hardware, I feel like someone needs to test infamous to ensure that's how it looks like on actual ps4 hardware.
image
image

sorry for horrible reference image, but on real hardware that issue actually exists too, the previous incorrect appearance was correct, at least for Bloodborne.
image

junminlee2004:
> Breaks Bloodborne reflections/layering, this is incorrect to how it appears on real hardware, I feel like someone needs to test infamous to ensure that's how it looks like on actual ps4 hardware.

I figured this would happen based on Roamic's comment, I updated the PR to skip UCP lowering when a shader exports its own clip distances. See if it's fixed now. Also the before screenshot for infamous doesn't do it justice, when you spin the camera around, those reflections on the water also spin around due to the way the game's rendering trick is set up. This is definitely not what happens on hardware.

junminlee2004:
image

Here is how it looks on real hardware

Kyoskii:
> > Breaks Bloodborne reflections/layering, this is incorrect to how it appears on real hardware, I feel like someone needs to test infamous to ensure that's how it looks like on actual ps4 hardware.
> 
> I figured this would happen based on Roamic's comment, I updated the PR to skip UCP lowering when a shader exports its own clip distances. See if it's fixed now. Also the before screenshot for infamous doesn't do it justice, when you spin the camera around, those reflections on the water also spin around due to the way the game's rendering trick is set up. This is definitely not what happens on hardware.

https://github.com/user-attachments/assets/6ed50860-eb63-488a-9cff-9426396a378e


I think the distance is fine but the layering seems incorrect for Bloodborne still,

Kyoskii:
> I think the distance is fine but the layering seems incorrect for Bloodborne still,

more visible example
https://github.com/user-attachments/assets/c08d40e0-f5ae-434d-a9c6-8e55e3125aeb

junminlee2004:
> > I think the distance is fine but the layering seems incorrect for Bloodborne still,
> 
> more visible example https://github.com/user-attachments/assets/c08d40e0-f5ae-434d-a9c6-8e55e3125aeb

Are you using an Nvidia gpu? I just made a commit that emulates UCP lowering to Nvidia gpus since clip distance is emulated there. Should also fix the layering issue.

Kyoskii:
amd apu
 ryzen ai max 395+ with 8060s
 
 edit: I did test your new commit, the behavior seems appropriate with your new commit after checking elsewhere im bloodborne.

junminlee2004:
Infamous seems to be fixed on both AMD and NVIDIA gpus as well with my latest commit but anyone can feel free to test for regressions.

Graubek:
Reflections seems to be fixed for me but water is flickering when readbacks are turned on (I'm using red zone patch)

junminlee2004:
> Reflections seems to be fixed for me but water is flickering when readbacks are turned on (I'm using red zone patch)

The flickering water in inFAMOUS with readbacks set to precise is a pre-existing issue that occurs without the red zone patch or this PR. This PR implements clipping to fix water reflections, not readback issues.

src/shader_recompiler/ir/passes/lower_user_clip_planes.cpp
```diff
@@ -0,0 +1,99 @@
...
+    auto it = std::ranges::find(program.info.buffers, BufferType::ClipPlanes,
+                                &BufferResource::buffer_type);
+    if (it == program.info.buffers.end()) {
+        it = program.info.buffers.insert(it, {
+                                                 .used_types = IR::Type::F32,
+                                                 .inline_cbuf = AmdGpu::Buffer::Placeholder(
+                                                     NumUserClipPlanes * 4 * sizeof(float)),
+                                                 .buffer_type = BufferType::ClipPlanes,
+                                             });
+    }
```
squidbus, line 39: Does it make sense to always insert this? I would assume this should be right before `IR::IREmitter ir{*target_block, target_it};`, where we've committed to inserting the user clip plane logic.

junminlee2004: I think the insertion above the mask check is necessary because the mask check can end the pass early when mask=0 on the first compile, which is an issue because the buffer registration wouldn't be written to shared info.

squidbus: If that's the case we should probably at least skip and bind it to the null buffer later then if it's not enabled, otherwise we're always copying the planes and binding this extra buffer every time.

squidbus (approved):

PR #4801, opened 2026-08-04
Appends last time played for each game in play_time.txt

Kody-Schram:
A really small change which just appends the Unix epoch time for games when playtime is updated in the UpdatePlayTime function.

I have tested, this does not affect the QT launcher in any way, it currently ignores this new meta data.


georgemoralis:
looking good but clang is broken

PR #4804, opened 2026-08-05
Various fixes around Media Player

StevenMiller123:
This PR is a collection of various things I did while debugging various issues that occur in newer versions of Media Player (CUSA02012).

- Improve Windows-specific stub of sigaction to set oact based on act parameters.
- Basic implementation of posix_access
- Stub success on attempts to call stat and posix_access on the filesystem root
- Fix our export for environ, we were giving the address to the vector object instead of the actual string data
- Ensure PthreadSpecificElem data is stored at the correct offset, since Sony's mono-runtime accesses these directly through pthreads instead of the pthread_set/getspecific functions
- Improve TCB patches to handle any src offset that would be within the guest TCB.
  - Sony's mono-runtime was accessing tcb_thread through fs:[0x10], which failed on Windows and Linux because it wasn't getting patched.
  - Removes need for the GenerateStackCheck and GenerateStackCanary patches, as fs:[0x28] is also part of the guest tcb.
- Add various libkernel_psmkit exports for implemented functions.

With these changes on Windows, Media Player now initializes the mono runtime, and starts actually executing C# code. It crashes before it can display anything, from failing to find some functions. I'll need to debug it.

On Linux, Media Player is crashing extremely early. The current issue is that early mono initialization tries walking through signal values to find ones it can use. Out of the signals they try using, all the available ones are above Linux's SIGRTMAX value, so host calls to sigaction fail with EINVAL. There's some other issues that crop up on Linux too, but @kalaposfos13 fixes most of them with their open PR for signal stuff.


georgemoralis:
seems to have conflict for merging

StevenMiller123:
Rebase conflicts should be fixed now.

PR #4807, opened 2026-08-05
Minor mistype fix

UltraDaCat:
Let's not forget about the copyright year while we're at it right?


georgemoralis:
copyright to @DanielSvoboda for this

PR #4808, opened 2026-08-05
shader_recompiler: add unorm/snorm/uint/sint 10_11_11 number formats

Wint3rNight:
`10_11_11` only implemented `NumberFormat::Float`, so anything else crashed:

```
Unpack10_11_11: Unreachable code!
Unsupported 10_11_11 number format: Snorm
```

(reported in #496)

`2_10_10_10` already handles Unorm/Snorm/Uint/Sint, so I added the same four
here following it. Fields are 11/11/10, so the divisors are 2047/2047/1023
for unorm and 1023/1023/511 for snorm. The Float path is unchanged.

Tested locally on NVIDIA and AMD — endpoints, one-LSB values to pin the
divisors, sign extension, pack/unpack round trips, clamping. Existing suite
still passes.

Two things worth flagging: only Snorm has an actual report behind it, the
other three are for symmetry. And snorm doesn't clamp the most-negative
value to -1.0 — `2_10_10_10` behaves the same way, so I matched it rather
than making the two inconsistent.


squidbus:
What game does this effect (I couldn't find the report glancing in the aggregate thread), and has it been checked that its working correctly there?

Wint3rNight:
It's God of War (CUSA07410), reported by @bigol83 here: https://github.com/shadps4-emu/shadPS4/issues/496#issuecomment-3133877242

Line 9288 of the log attached to that comment:

[Debug] <Critical> ir_emitter.cpp:1073 Unpack10_11_11: Unreachable code!
Unsupported 10_11_11 number format: Snorm

It came up as a side thread about PS4 Pro mode rather than as its own report, and since it isn't an Unknown opcode error it doesn't match what that issue is nominally collecting.

On your second question: no, I haven't confirmed the game renders correctly past that point. I don't own it, so I can't. What I did check is that the formats produce the right values, the tests/gcn suite only builds ALU encodings so it can't reach buffer-format code, so I injected the pack/unpack ops into shaders through a throwaway harness and verified the round-trips on both an NVIDIA GPU and an AMD one. The divisors are 2047/2047/1023 for unorm and 1023/1023/511 for snorm, matching the 11/11/10 field widths the same way the existing 2_10_10_10 path does.

Also worth being upfront: only Snorm appears in that report. I added all four because 2_10_10_10 already has all four and there was no reason for 10_11_11 to be the odd one out, but Uint/Sint/Unorm aren't exercised by any log I could find.

If someone with God of War can run it against this branch that'd settle it properly. I am happy to wait on that rather than have it go in unverified.

PR #4821, opened 2026-08-09
FMask

DanielSvoboda:


StevenMiller123 (changes requested):
Fmask formats should be patched out in ResourceTrackingPass. They shouldn't make their way to surface format creation if I remember right.

squidbus:
It seems fine, I'm not sure adding to num components / sizes is necessary though since we usually eliminate these before we get there.

squidbus (approved):

PR #4823, opened 2026-08-10
net: implement sceNetGetSockInfo, AP callbacks and POSIX socket options

ntamas94:
Unity/Mono titles using PlayLink cannot get past network start-up. Three separate
problems, found while bringing up Knowledge is Power (CUSA12435).

### `sceNetGetSockInfo` was a parameterless stub returning `ORBIS_OK`

Callers treat a return value <= 0 as failure, so the game destroyed and recreated its
UDP socket forever — **16,015 `sys_socketex` calls with zero `sys_socketclose`** in a
single run, and no socket ever stayed open long enough to receive anything.

The struct layout is recovered from how the game and its `PS4NativeSockets` plugin use
the buffer: they reserve `0xA0` bytes per entry, read a readable-byte count at `+0x2C`
and the socket type at `+0x44` (compared against `SOCK_DGRAM` and `SOCK_DGRAM_P2P`).
Fields whose meaning is not established are left zeroed rather than guessed at — happy
to adjust if someone has the real layout.

### `NameOf(OrbisNetSocketSoOption)` aborted on unknown options

It is called to build a log string and the lambda is evaluated eagerly, so an unknown
option killed the emulator through `UNREACHABLE_MSG` even with debug logging off. It
now returns `"unknown"`, and `Set`/`GetSocketOptions` report `ENOPROTOOPT` instead of
aborting.

### POSIX socket options were missing

The POSIX entry points pass FreeBSD-numbered options (`0x1003`-`0x1006`), which were
absent from the enum, and carry the timeout as a `timeval` rather than the microsecond
`int` the sceNet-numbered options use. Both shapes are now accepted.

### `sceNetCtlApCheckCallback` never dispatched

It now delivers to callbacks registered through `sceNetCtlApRegisterCallback`, and only
on an actual state change — games poll it every frame, so unconditional delivery meant
2,099 duplicate events in one 55-minute run.

### Result

The title reaches a stable listening state on UDP 9066-9071, `sys_socketex` drops from
16,015 to 7, and two phones complete the PlayLink handshake.

### Open question for maintainers

The non-AP `sceNetCtlCheckCallback` is also a stub even though
`NetCtlInternal::CheckCallback()` is fully implemented. I left it alone — it would
affect other titles and I could not tell whether the current behaviour is deliberate.

Tested on Windows, clang-cl, NVIDIA RTX 2080.

### DIFF
--- a/src/core/libraries/network/net.cpp
+++ b/src/core/libraries/network/net.cpp
@@ -1033,9 +1033,32 @@ int PS4_SYSV_ABI sceNetGetRouteInfo() {
     return ORBIS_OK;
 }
 
-int PS4_SYSV_ABI sceNetGetSockInfo() {
-    LOG_ERROR(Lib_Net, "(STUBBED) called");
-    return ORBIS_OK;
+int PS4_SYSV_ABI sceNetGetSockInfo(OrbisNetId s, OrbisNetSockInfo* info, int n, int flags) {
+    if (info == nullptr || n <= 0) {
+        *Libraries::Kernel::__Error() = ORBIS_NET_EINVAL;
+        return -1;
+    }
+    auto file = FDTable::Instance()->GetSocket(s);
+    if (!file) {
+        *Libraries::Kernel::__Error() = ORBIS_NET_EBADF;
+        LOG_ERROR(Lib_Net, "socket id is invalid = {}", s);
+        return -1;
+    }
+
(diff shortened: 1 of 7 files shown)


StevenMiller123:
Please respect our AI-use rules listed in https://github.com/shadps4-emu/shadPS4/blob/main/CONTRIBUTING.md

PR #4824, opened 2026-08-10
app_content: don't abort when the dump has no TITLE_ID

ntamas94:
`sceAppContentInitialize` read `TITLE_ID` from `param.sfo` and hit `UNREACHABLE_MSG`
when it was absent, taking the emulator down with "Failed to get TITLE_ID".

A dump with a missing or unreadable `param.sfo` is not an emulator invariant violation —
it just means there is no title id to key addon lookups on, and so no addons to mount.
The very next branch already treats "no addon directory" as success, so the missing id
now logs an error and returns `ORBIS_OK` the same way.

The title boots and plays with this change; only addon content stays unavailable, which
is the correct consequence of not knowing the title id.


StevenMiller123 (changes requested):
The reason we have this unreachable is because all games have a param.sfo. If your game dump does not have a param.sfo, then you probably need to re-dump it.

PR #4830, opened 2026-08-11
Implement sceSysmoduleLoadModuleByNameInternal

kalaposfos13:
Force-LLEing stuff from sys_modules/title_id/ is already a thing, and this PR makes that possible for apps that use the internal function sceSysmoduleLoadModuleByNameInternal to load modules instead of the standard API for games, such as VSH. This by itself is not enough to satisfy module loading logic for that though, at least not in a clean way, as some modules ideally should be skipped without reporting an error, so it can both proceed, while at the same time we don't have to deal with having to emulate even more stuff for misc libraries.

I only tagged Stephen in one commit, but his contributions were more encompassing than just that one addition.

### DIFF
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -1583,6 +1583,8 @@ void RegisterFileSystem(Core::Loader::SymbolsResolver* sym) {
     LIB_FUNCTION("c7ZnT7V1B98", "libScePosix", 1, "libkernel", posix_rmdir);
     LIB_FUNCTION("c7ZnT7V1B98", "libkernel", 1, "libkernel", posix_rmdir);
     LIB_FUNCTION("naInUjYt3so", "libkernel", 1, "libkernel", sceKernelRmdir);
+    LIB_FUNCTION("8vE6Z6VEYyk", "libkernel_psmkit", 1, "libkernel", posix_access);
+    LIB_FUNCTION("8vE6Z6VEYyk", "libkernel", 1, "libkernel", posix_access);
     LIB_FUNCTION("E6ao34wPw+U", "libScePosix", 1, "libkernel", posix_stat);
     LIB_FUNCTION("E6ao34wPw+U", "libkernel", 1, "libkernel", posix_stat);
     LIB_FUNCTION("eV9wAD2riIA", "libkernel", 1, "libkernel", sceKernelStat);
--- a/src/core/libraries/kernel/file_system.cpp
+++ b/src/core/libraries/kernel/file_system.cpp
@@ -659,6 +659,27 @@ s32 PS4_SYSV_ABI sceKernelRmdir(const char* path) {
     return result;
 }
 
+s32 PS4_SYSV_ABI posix_access(const char* path, s32 mode) {
+    LOG_INFO(Kernel_Fs, "(PARTIAL) path = {}, mode = {}", path, mode);
+    if (strlen(path) > 255) {
+        *__Error() = POSIX_ENAMETOOLONG;
+        return -1;
+    }
+
+    auto* mnt = Common::Singleton<Core::FileSys::MntPoints>::Instance();
+    const bool is_dir = mnt->IsDirectory(path);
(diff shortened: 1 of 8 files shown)


src/core/libraries/kernel/file_system.cpp
```diff
@@ -661,6 +659,27 @@ s32 PS4_SYSV_ABI sceKernelRmdir(const char* path) {
...
+    const bool is_root = strncmp(path, "/", 2) == 0;
+    if (!is_dir && !is_file && !is_root) {
+        *__Error() = POSIX_ENOENT;
+        return -1;
+    }
+    if (is_root) {
+        LOG_WARNING(Kernel_Fs, "Checking accessibility of filesystem root");
+    }
+    return ORBIS_OK;
+}
```
StevenMiller123, line 687: I don't think you need this in here?
PR #4804 already has the exact same implementation for it.

src/core/libraries/sysmodule/sysmodule.cpp
```diff
@@ -98,9 +99,20 @@ s32 PS4_SYSV_ABI sceSysmoduleLoadModule(OrbisSysModule id) {
...
 
-s32 PS4_SYSV_ABI sceSysmoduleLoadModuleByNameInternal() {
-    LOG_ERROR(Lib_SysModule, "(STUBBED) called");
-    return ORBIS_OK;
+s32 PS4_SYSV_ABI sceSysmoduleLoadModuleByNameInternal(char const* name, u64 args, void const* argp,
+                                                      void const* popt, s32* res) {
+    LOG_ERROR(Lib_SysModule, "(DUMMY) called, name: {}", name);
+    std::string filename = std::string(name) + ".sprx";
+    s32 exists;
+    exists = Libraries::Kernel::posix_access(("/sys/common/lib/" + filename).c_str(), 0);
```
StevenMiller123: Should use sceKernelGetFsSandboxRandomWord instead of hardcoding sys here.

StevenMiller123 (changes requested):
Mostly looks good to me, just some minor nits.

src/core/libraries/kernel/file_system.cpp
kalaposfos13: I guess I'll be waiting until that PR gets merged, then.

src/core/libraries/sysmodule/sysmodule.cpp
```diff
@@ -98,9 +100,22 @@ s32 PS4_SYSV_ABI sceSysmoduleLoadModule(OrbisSysModule id) {
...
-s32 PS4_SYSV_ABI sceSysmoduleLoadModuleByNameInternal() {
-    LOG_ERROR(Lib_SysModule, "(STUBBED) called");
-    return ORBIS_OK;
+s32 PS4_SYSV_ABI sceSysmoduleLoadModuleByNameInternal(char const* name, u64 args, void const* argp,
+                                                      void const* popt, s32* res) {
+    LOG_ERROR(Lib_SysModule, "(DUMMY) called, name: {}", name);
+    std::string filename = std::string(name) + ".sprx";
+    s32 exists;
+    using namespace Kernel;
+    std::string system_base = std::string("/") + sceKernelGetFsSandboxRandomWord();
```
StevenMiller123, line 150: I personally think we should tie this into some allow/block list instead of relying on users making game-specific module folders for things calling this (even if it's only called by firmware stuff).
Maybe try to have this use the same array of allowed modules that's currently placed in the loadModule code, with a bypass for modules present in game-specific module folders?

StevenMiller123 (commented):
Just some more thoughts on this now that the blocking PR (my fixes for Media Player) are now merged.

src/core/libraries/sysmodule/sysmodule.cpp
kalaposfos13: That's fine by me, do you have a list of safely LLEable modules loaded through this function perchance?

StevenMiller123: Been too long since I glanced at VSH, I forget what all it tries loading through this.

PR #4832, opened 2026-08-11
Add config option extra_fmem_in_mbytes

kalaposfos13:
This is the same as extra_dmem_in_mbytes, but targeting the other memory type, and is more roundabout and less rigid.


StevenMiller123:
Works on my machine

PR #4836, opened 2026-08-11
Add --env flag to pass environment variables to the guest

kalaposfos13:
Validated by merging hacks on this branch locally until VSH got far enough to test this.


Gih-pt:
It's interesting to see work related to VSH now.

PR #4837, opened 2026-08-11
Fix argv parsing when the -- argument is not specified.

StevenMiller123:
Thanks @kalaposfos13 for breaking this 😄


kalaposfos13:
You're welcome. :)
Fortunately, this still managed to fit in the limit of at most one dumb regression per day per person, so it should be fine.
So, are we merging this now or when the CI gets itself back together?

PR #4838, opened 2026-08-12
FileSys: Open host-backed files with POSIX-style full sharing

sgdc3:
Games expect POSIX behaviour when they open the same file twice: if one fd already has it open for writing, opening it again on another one should still work. That is already the case on Linux and macOS, since the share mode in IOFile only exists on the Windows path, but on Windows the open falls back to IOFile's default (ShareReadOnly, `_SH_DENYWR`) and the second open comes back as EACCES.

I ran into this with LittleBigPlanet 3, which keeps reopening its own temp_orbis cache while it is still open for writing: those opens are rejected and the game gives up on loading the current level.

Disclosure: analysis, debugging and development of this fix were done with the help of Claude Code; I reviewed the reasoning and the final change myself before opening this PR.


georgemoralis:
@StevenMiller123 i believe this is proper , can you give me your thoughts?

sgdc3:
Any news about this? @georgemoralis @StevenMiller123

StevenMiller123 (commented):
Remove or simplify the comment, and I think this will be good to go?

StevenMiller123 (approved):
Assuming I understand the code correctly, I think this looks fine.

PR #4839, opened 2026-08-12
video_core: service readback faults at widened window

junminlee2004:
This PR fixes the open world breaking in inFAMOUS Second Son and First Light when precise readbacks is enabled due to a high frequency of faults scattered across gpu modified ranges. Servicing pages at a widened 512 kb window around the fault fixes the issue. There is also a performance improvment when precise readbacks are enabled. 

Before (servicing one 4kb page per fault)
before

After (servicing 128 pages/512kb window per fault)
after

Before (servicing one 4kb page per fault)
bb before

After (servicing 128 pages/512kb window per fault)
bb after

### DIFF
--- a/src/video_core/buffer_cache/buffer_cache.cpp
+++ b/src/video_core/buffer_cache/buffer_cache.cpp
@@ -86,7 +86,7 @@ void BufferCache::ReadMemory(VAddr device_addr, u64 size, bool is_write) {
             std::max<VAddr>(Common::AlignDown(device_addr, WindowSize), buf_start);
         const VAddr window_end = std::min<VAddr>(
             std::max<VAddr>(window_start + WindowSize, device_addr + size), buf_end);
-        DownloadBufferMemory<false>(buffer, window_start, window_end - window_start, false);
+        DownloadBufferMemory<false>(buffer, window_start, window_end - window_start);
         if (is_write) {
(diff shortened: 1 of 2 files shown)


src/video_core/buffer_cache/buffer_cache.cpp
```diff
@@ -77,7 +77,19 @@ void BufferCache::InvalidateMemory(VAddr device_addr, u64 size) {
...
+        // is widened to a window around the request
+        constexpr u64 WindowSize = 512_KB;
+        const VAddr buf_start = buffer.CpuAddr();
+        const VAddr buf_end = buf_start + buffer.SizeBytes();
+        const VAddr window_start =
+            std::max<VAddr>(Common::AlignDown(device_addr, WindowSize), buf_start);
+        const VAddr window_end = std::min<VAddr>(
+            std::max<VAddr>(window_start + WindowSize, device_addr + size), buf_end);
+        DownloadBufferMemory<false>(buffer, window_start, window_end - window_start, false);
+        if (is_write) {
```
raphaelthegreat, line 90: this is already called inside DownloadBufferMemory, so you might want to remove it from there if your intention is to avoid marking the whole window as cpu modified as well

raphaelthegreat (approved):

PR #4842, opened 2026-08-13
vdec2: revert CPU and GPU memory calculations

roamic:
should fix uncharted, avplayer lle still works


DarkBoy08:
All clear. Can be merge)

PR #4844, opened 2026-08-13
Fix more random UE crashes tied to readbackLinearImages

StevenMiller123:
More proper solution to the issue I was dealing with in #4748. Having looked closer at the code, there isn't a race between PopPendingOperations and the priority op runners since they use different queues. The actual issue is that PopPendingOperations and DeferOperation can be run simultaneously, and since both functions modify the pending ops queue, this can trigger crashes.

#4748 probably helped as a coincidence because it meant the priority runners would now block the GPU command processor thread, but this should be more consistently working solution.

This helps with more random crashes in some UE titles, including Kingdom Hearts 3.

### DIFF
--- a/src/video_core/renderer_vulkan/vk_scheduler.cpp
+++ b/src/video_core/renderer_vulkan/vk_scheduler.cpp
@@ -119,7 +119,7 @@ void Scheduler::Wait(u64 tick) {
 }
 
 void Scheduler::PopPendingOperations() {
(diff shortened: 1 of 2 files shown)


georgemoralis:
of course KH3 what else

StevenMiller123:
I dunno, I only tested KH3 😆 
Technically should affect stability of DMA and readbackLinearImages, and potentially other random crashes that could come from deferred non-priority ops.

PR #4847, opened 2026-08-13
Core: Support for memory patches targeting a specific module

sgdc3:
At present, patches can only target the game's main executable: patch addresses are resolved via `g_eboot_address`, pattern scans are performed on the eboot image, and the changes are applied during `OnGameLoaded`. Therefore, it is not possible to write a patch for a bug in one of the PRX modules supplied with a game: when patches are applied, those modules have not yet been loaded and their base address is not that of the eboot. Furthermore, their address is not even the same from one run to the next, as it depends on what has been mapped before it.

This PR allows a patch line to specify the name of its target module:

```xml
<Line Type="bytes" Address="0x9908" Value="0F1F00" Module="spu.prx" />
```

Such patches read their own address as an offset within the module, resolve it (along with any mask patterns) based on the module's base address and size, and are applied when the module is actually mapped, just before its code is executed. If the module has not yet been loaded, the patch waits in a pending list that the module loader drains as each module loads; if the patch arrives later via IPC and the module is already known, it is applied immediately. Patches lacking this attribute behave exactly as before.

The second commit applies the changes necessary to integrate these patches with the IPC. PATCH_MEMORY has no field for the module, and its fields are positional: extending it would leave the parser one field out of sync with a sender that does not send that additional field. PATCH_MEMORY_MODULE is therefore defined as its own command, negotiated via the ENABLE_MEMORY_PATCH_MODULE capability, so as to keep the IPC backwards compatible.
A follow-up PR for the launcher will implement sending the new command; until then it is only advertised, and senders that do not know about it keep using PATCH_MEMORY.

~~The reason behind this PR stems from my need to fix a bug that freezes LittleBigPlanet 3 a few seconds after its main menu, the Pod, has loaded. The game includes a module called spu.prx, inherited from the PS3 version, which masks a memory address down to 32 bits before an atomic increment. On PS3 that was harmless, as guest memory sat below 4 GiB; but in shadPS4 the eboot maps beyond 4 GiB, so the increment lands on a low alias instead of the counter it is meant to update, and the queue that counter guards overflows. The fix involves 15 bytes at two points within that module, and without this change there is no way to express it.~~

Disclosure: analysis, debugging and development of this change were done with the help of Claude Code; I reviewed the reasoning and the final change myself before opening this PR.

### DIFF
--- a/src/common/memory_patcher.cpp
+++ b/src/common/memory_patcher.cpp
@@ -4,8 +4,10 @@
 #include <algorithm>
 #include <codecvt>
 #include <fstream>
+#include <mutex>
 #include <sstream>
 #include <string>
+#include <unordered_map>
 #include <nlohmann/json.hpp>
 #include <pugixml.hpp>
 #include "common/elf_info.h"
@@ -23,6 +25,17 @@ std::string g_game_serial;
 std::string patch_file;
 bool patches_applied = false;
 std::vector<patchInfo> pending_patches;
+
+// A patch that names a module has to wait for it to be mapped. Modules load on guest threads
+// while patches can come in from the IPC thread, hence the mutex.
+struct LoadedModule {
+    uintptr_t base{};
+    uint64_t size{};
+};
+
+static std::mutex g_module_mutex;
+static std::unordered_map<std::string, LoadedModule> g_loaded_modules;
+static std::vector<patchInfo> g_pending_module_patches;
 
 std::string toHex(u64 value, size_t byteSize) {
     std::stringstream ss;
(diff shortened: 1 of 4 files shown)


raphaelthegreat:
This is an emulator bug, the eboot should be loaded at the lowest possible address and guest modules from 0x80000000, currently everything gets mapped at 0x800000000ULL

sgdc3:
@raphaelthegreat Well, that makes a lot more sense now! I tried changing the value of ModuleLoadBase in module.cpp and the game works as expected! 😄 I’m not sure whether this PR would be relevant in other scenarios; I’ll leave it to the repository maintainers to decide whether or not to close it.

PR #4848, opened 2026-08-13
Core: Load guest modules below 4 GiB

sgdc3:
At the moment, modules are mapped starting from 0x800000000. This address lies beyond the SYSTEM_MANAGED_MAX limit (0x7FFFFBFFF), so the modules end up outside the system-managed area to which they belong, and no pointer to them fits within 32 bits.

A title ported from PS3 may contain code that truncates an address to 32 bits, which is harmless on a real console because modules are mapped low, but causes addressing issues here.

Let's map them starting from 0x80000000 as recommended by @raphaelthegreat in #4847

### DIFF
--- a/src/core/module.cpp
+++ b/src/core/module.cpp
@@ -20,7 +20,9 @@ namespace Core {
 
 using EntryFunc = PS4_SYSV_ABI int (*)(size_t args, const void* argp, void* param);
 
-static constexpr u64 ModuleLoadBase = 0x800000000;
+// In the system managed area, and low enough that guest code truncating an address to 32 bits
+// still lands on the right module.
+static constexpr u64 ModuleLoadBase = 0x80000000;
 
 static u64 GetAlignedSize(const elf_program_header& phdr) {
     return (phdr.p_align != 0 ? (phdr.p_memsz + (phdr.p_align - 1)) & ~(phdr.p_align - 1)
--- a/src/core/module.cpp
+++ b/src/core/module.cpp
@@ -20,6 +20,7 @@ namespace Core {
 
 using EntryFunc = PS4_SYSV_ABI int (*)(size_t args, const void* argp, void* param);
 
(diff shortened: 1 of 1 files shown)


StevenMiller123:
This should also fix Final Fantasy XII, which was also expecting addresses to be 32-bit.

StevenMiller123 (commented):
I think it wouldn't hurt to follow the suggestion closer and separate the eboot load address from this.
Specifically, the eboot should load at the lowest possible address, game-supplied modules should use a base address of 0x80000000, and system modules should use a base of 0x800000000 (the current base we use for all of them).

These base addresses come from an old message from @red-prig, I haven't confirmed these for myself or really looked at this recently.

red-prig:
More precisely, eboot.bin always loads at 0x0 00400000, game PRX starts at 0x0 80000000, and system PRX starts at 0x8 000000000.
eboot.bin can be moved to 0x0 80000000 and it will work. Using 0x0 00400000 requires a complex hack, at least on Windows.

raphaelthegreat:
Isnt 0x00400000 only with goldhen running? Otherwise ASLR will pick a random address before 0x80000000? So its fine to pick lowest available address on windows that might be a little higher.

red-prig:
oh yeah exactly, with ASLR enabled this is the interval:[0x400000..0x80000000-size] but only for the ET_SCE_DYNEXEC type

StevenMiller123 (commented):
Seems to work as expected, but it's causing some weird crashes in some of my titles. I think it might be best to load eboots at the game modules base for now, and I'll need to investigate the crashing I'm seeing.

StevenMiller123 (commented):
Apparently the weird crashes were a bug from another PR. Could you revert your most recent change? (sorry for the flip-flopping around on this 😓)
I'll approve once you do that, right now the eboot is mapping too high so Final Fantasy XII is broken again.

PR #4852, opened 2026-08-14
Net: Implement P2P sockets on a real UDP socket

sgdc3:
The current P2PSocket implementation is a stub. Native() returns nothing, so a P2P descriptor cannot be added to an epoll and select() has nothing to wait on; sceNetEpollControl handles every operation by returning an error and printing the `P2P socket cannot be added to epoll (unimplemented)` message in the console and no datagram ever leaves or reaches the machine. A title that runs its netcode over P2P sockets waits forever on readiness that cannot arrive.

This PR gives every host port a real UDP socket. P2PPort owns one socket per (address, port) pair, shared by every vport bound on it, and each datagram carries a 6-byte {dst_vport, src_vport, flags} header. The layout follows RPCS3's, which solves the same problem (there was no reason to re-invent the encapsulation).

The port's receive thread demultiplexes on dst_vport and forwards the payload into a loopback "inbox" socket belonging to the matching endpoint. That inbox is what Native() now returns, so every guest P2P socket owns a real descriptor and the existing epoll & select paths work on it unchanged.

Connect on a datagram socket records the destination instead of discarding it, which sceNetSend needs since it carries no address of its own. And with networking disabled the port binds the loopback only: a title probing its own signaling port still gets its reply without needing an exposed UDP port.

I ran into this with LittleBigPlanet 3, which opens a P2P socket while it is still booting and probes its own signaling port through the loopback. With the current stub implementation the reply never comes back, so the title sits on its connection screen and never reaches the main menu. With a real socket underneath, it does (then crashes due rendering issues).

Scope of testing: one title, on Windows. I have not looked into the stream (SOCK_STREAM_P2P) path, which stays unimplemented.

Disclosure: analysis, debugging and development of this change were done with the help of Claude Code; I reviewed the reasoning and the final change myself before opening this PR.

Given how heavily this one was scaffolded with the help of an LLM, I'm keeping this PR as a draft, I would like to have it read first by anyone with more experience of the shadPS4 codebase, and of the netcode implementation in particular.


Randomuser8219:
No regressions with my titles. LBP3 remains improved.

PR #4856, opened 2026-08-15
Silence the extremely spammy compiler warning after #4804

kalaposfos13:


StevenMiller123 (commented):
I think the static assert for pthread size should be kept in pthreads.h like the mutex one.
Other than that, if this silences the compiler warnings, I'm all for it.

StevenMiller123 (approved):

CMakeLists.txt
```diff
@@ -1377,6 +1377,10 @@ if (WIN32)
     target_sources(shadps4 PRIVATE src/dist/shadps4.rc)
 endif()
 
+if (NOT WIN32)
```
StevenMiller123: Wait a moment, why not WIN32? Is there not a way to deal with these warnings there?

kalaposfos13: Different compiler, different flags, I only added this for Clang, not MSVC.

PR #4857, opened 2026-08-15
video_core: honor STENCILOPVAL for GCN ReplaceOp stencil writes

junminlee2004:
Fixes stale shadow maps, broken alpha blending, missing video output on in-game TVs in inFAMOUS Second Son and First Light and potentially in other games (I think God of War 3 had something similar to this? I don't have it to test though). 

So in Infamous, DB_STENCILREFMASK writes fields test_val and op_val but updatedepthstencilstate() converted it to vulkan dynamic state, which can only use one reference value per face, which always forced the test_val to be used for both compare and write. As such, both front.stencil_test_val and back.stencil_tes_val became the test_val of 0, causing replaceop to always write 0. This gates any pass that requires certain stencil values, which removes the rendering of Reggie, video output on TVs, breaks alpha blending on hair, and since the shadow mask repainted every frame, replaceop writing 0 forever made it so that this is never done per frame, causing the mask to be rasterized and then remain stale, causing this dirty camera lens effect. The fix is to gate which of the two fields get passed to dyanmic_state.SetStencilReferences() with the following rule: if any of the face's stencil ops is replaceop, pass op_val, otherwise, test_val.

before: test_val only ( no gating logic)
stale shadow mask 1

after: gating logic active
fixed shadow mask 1

before: test_val only ( no gating logic)
stale shadow mask 2

after: gating logic active
stale shadow mask 3

### DIFF
--- a/src/video_core/renderer_vulkan/vk_rasterizer.cpp
+++ b/src/video_core/renderer_vulkan/vk_rasterizer.cpp
@@ -1316,7 +1316,23 @@ void Rasterizer::UpdateDepthStencilState() const {
         const auto front = regs.stencil_ref_front;
         const auto back =
             regs.depth_control.backface_enable ? regs.stencil_ref_back : regs.stencil_ref_front;
-        dynamic_state.SetStencilReferences(front.stencil_test_val, back.stencil_test_val);
+        // GCN REPLACE_OP writes DB_STENCILREFMASK.STENCILOPVAL, so a face whose stencil ops
+        // include ReplaceOp takes its Vulkan reference from op_val.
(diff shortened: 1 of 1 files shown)


raphaelthegreat:
What is the stencil compare op? Should probably also check that using the opval as ref doesnt break the stencil check itself and emit a warning if it does

junminlee2004:
In the affected draws, the compare op is Always. The stamp draws test nothing and only write. As for the warning, I actually had it in my  first draft. I'll add it back in.

raphaelthegreat (approved):

PR #4858, opened 2026-08-15
Rewrite signal emulation again

kalaposfos13:
VSH has a check that failed under the previous implementation, because it tried to install a handler for every possible signal, of which there are 128 on the PS4, but only 64 and 32 on Linux and macOS, respectively (on Windows, this wasn't an issue because signal code is still mostly stubbed, and the things that aren't, are software emulated anyway). Previously, every guest signal was mapped to a host signal, and then everything was transparently forwarded to the host OS, which worked pretty well up until something actually came along and started using those higher signal numbers not available on host platforms. Since VSH only checked their availibility, but didn't actually make use of them, it was rather easy to just ignore those signals and return dummy successes, but it nonetheless prompted me to rewrite signal emulation on POSIX systems once again.

The new logic works roughly like so:
Signals can be grouped into two categories, hardware ones that are raised by the host hardware/OS, like SIGSEGV, SIGILL, SIGFPE etc, and there are software ones, that can only come from the guest invoking raise/pthread_kill themselves as an internal signaling mechanism. The first group has a host signal handler installed for them as well, and (after first checking if our HLE was responsible for the signal in cases of SIGSEGV or SIGBUS) the signal and its accompanying info gets forwarded to g_curthread for consumption (if no handler is installed, do the default action which is "crash" for most and "ignore" for some signals, if the thread is in sigwait, wake it, if it's in sigsuspend, call the handler then wake it, otherwise call the handler). For the other set of signals, pthread_kill will first queue the signal on the thread, check if the signal is blocked, and if not, sends a notification to the thread via a host SIGUSR1 signal, whose handler on the other end will inspect its thread's state, and figure out what to do from there (see above, but prepend a signal queue lookup to it with respect to the signal mask). The other sig\* functions follow a similar setup as well.

~~The Pthread backend can be reused for Windows I'm pretty sure, and it might even allow for a better hardware signal emulation as well (just make use of the same backend from Windows's SignalHandler at signals.cpp:29, the only code I know is that 0xc0000005 is the SIGSEGV equivalent), but for now I've kept everything POSIX only because I use Arch Linux (BTW Edition) btw don't use Windows nor do I care about it.~~ Done. The Windows code has been updated to use the same backend, and can now emulate SIGSEGV and SIGILL signals to some degree (testing needed).

sigaction also had a quite dumb bug that I can't even blame on anyone else because it was me who introduced it months ago, which is that it used the sce wrapper functions' callback function signature instead of the POSIX signature(s), meaning guest apps were getting garbage parameters aside from the signal number itself, which only wasn't found earlier because most games only use the sce wrappers, and those were fine.

Testing would be appreciated, because while I'm pretty sure I didn't break anything (on Linux), one must never be too sure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -328,6 +328,7 @@ set(KERNEL_LIB src/core/libraries/kernel/coredump/coredump.cpp
                src/core/libraries/kernel/threads/event_flag.cpp
                src/core/libraries/kernel/threads/exception.cpp
                src/core/libraries/kernel/threads/exception.h
+               src/core/libraries/kernel/threads/exception_codes.h
                src/core/libraries/kernel/threads/mutex.cpp
                src/core/libraries/kernel/threads/pthread_attr.cpp
                src/core/libraries/kernel/threads/pthread_clean.cpp
--- a/src/core/libraries/kernel/kernel.cpp
+++ b/src/core/libraries/kernel/kernel.cpp
@@ -105,49 +105,55 @@ s32 PS4_SYSV_ABI sceKernelError(s32 posix_error) {
     return posix_error + ORBIS_KERNEL_ERROR_UNKNOWN;
 }
 
-void SetPosixErrno(s32 e) {
+s32 NativeToPosixErrno(s32 const e) {
     // Some error numbers are different between supported OSes
     switch (e) {
+    case 0:
+        return 0;
     case EPERM:
-        g_posix_errno = POSIX_EPERM;
+        return POSIX_EPERM;
         break;
     case ENOENT:
-        g_posix_errno = POSIX_ENOENT;
+        return POSIX_ENOENT;
         break;
     case EINTR:
-        g_posix_errno = POSIX_EINTR;
+        return POSIX_EINTR;
         break;
     case EDEADLK:
-        g_posix_errno = POSIX_EDEADLK;
+        return POSIX_EDEADLK;
         break;
     case ENOMEM:
-        g_posix_errno = POSIX_ENOMEM;
+        return POSIX_ENOMEM;
         break;
     case EACCES:
-        g_posix_errno = POSIX_EACCES;
+        return POSIX_EACCES;
         break;
     case EFAULT:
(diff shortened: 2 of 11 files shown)


StevenMiller123 (approved):
I think this is pretty good, and your tests all pass on Windows. Not seeing any regressions on my end either.

tochnonechelovek:
No regressions for me on Windows, except change in behaviour for CUSA07587 - ARMORED WARFARE. Now it hangs instead of the usual crash.
This PR hangs on the last line: [CUSA07587 hang.log](https://github.com/user-attachments/files/31124940/CUSA07587.hang.log)
Latest main crash (Critical is not the last line at all): [`4cc54cd`](https://github.com/shadps4-emu/shadPS4/commit/4cc54cdadd3947bd88b379ac8e96a619ecec25bc) [CUSA07587.log](https://github.com/user-attachments/files/31124954/CUSA07587.log)

StevenMiller123:
Looks like the critical still logged, but the crash didn't happen?

tochnonechelovek:
> Looks like the critical still logged, but the crash didn't happen?

Oh, didn't see that. Yeah, probably. Will test tomorrow to see if anything will change (I doubt it), but the hang was consistent in a few launches between changing the versions. This game really likes to log a lot more even after Critical and this PR just doesn't close the game for some reason, not a big problem though.

kalaposfos13:
> Looks like the critical still logged, but the crash didn't happen?

This was an issue with non-signal-related asserts and unreachables in general, the latest push should address this problem.

kalaposfos13:
It also introduces a side effect of asserts not being trivially skippable by debuggers anymore, because no signal handling means any upcoming signal, of which there are many in the memory tracker for example, would crash the program anyway, but since that's not an officially supported thing and can be fixed by removing one line, it's technically not a problem. Also, in my opinion, if one really wants to skip asserts, removing them one by one at their source is always preferable to skipping every assert unconditionally.

tochnonechelovek:
Tested the last commit and CUSA07587 crashes like normal. No regressions in other games.

PR #4860, opened 2026-08-15
Kernel.Pthreads: Do not double destroy threads when free threads exceeds MaxCachedThreads

StevenMiller123:
pthreads should only be deleted in the should_free path, where thread_heap.Free() calls std::destroy_at for us. Right now, we are always deleting threads, and when thread_heap.Free() is called, this re-runs the pthread destructor on the already-destroyed pthread and triggers crashes.

I don't know why this wasn't causing issues on Windows before, but it's been blocking PAYDAY 2 on Linux for ages.


georgemoralis:
What is THE GAME?

StevenMiller123:
Updated my description to clarify :laughing:

PR #4862, opened 2026-08-16
video_core: respect SPI_BARYC_CNTL front-face encoding

w1naenator:
- Read FRONT_FACE_ALL_BITS from SPI_BARYC_CNTL
- Emit +1.0/-1.0 when sign encoding is selected
- Emit integer 1/0 when all-bits encoding is selected
- Carry the register state through FragmentRuntimeInfo
- Add a boolean IR attribute path for FrontFacing
- Include the mode in shader specialization and cache metadata
- Add regression tests covering both AMD-defined encodings

Fixes according Roamic, Raphael, and IndecisiveTurtle recommendations.


georgemoralis:
@raphaelthegreat does this looks OK now?

raphaelthegreat:
The review comments from previous PR remain unaddressed

raphaelthegreat:
Need to fix test compilation and should be fine

PR #4865, opened 2026-08-17
Fix issues related to platinum trophies

kalaposfos13:
This PR fixes a wrong file read bug that prevented platinum trophies from being unlocked at all, makes the check for its unlock take into account that DLC trophies don't count towards it, and delay its popup to only after the last trophy that triggered it has finished presenting its own popup, because otherwise, the platinum popup was just overwriting the previous one.

### DIFF
--- a/src/core/libraries/np/np_trophy.cpp
+++ b/src/core/libraries/np/np_trophy.cpp
@@ -916,7 +916,7 @@ int PS4_SYSV_ABI sceNpTrophyUnlockTrophy(OrbisNpTrophyContext context, OrbisNpTr
     }
 
     pugi::xml_document doc;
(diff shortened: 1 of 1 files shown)


rainmakerv3:
the trophy queue should prevent the trophy popups from overwriting each other, and I recall that it used to work before. If the platinum pop-up is overwriting the penultimate trophy pop-up, then the queue itself may have been broken at some point. 

If this is the case, then the fix here only covers platinum trophies, but will not fix other instances of trophy pop-ups being not being queued properly when two or more are earned at the same time.

kalaposfos13:
Yes, that is correct here, I was using SotC's debug menu to test this, and each new trophy overwrote the previous one's popup.

PR #4872, opened 2026-08-18
externals: Use abseil as submodule instead of fetching at build time

squidbus:
Helps build environments with locked down internet access. This also allows using the system installed version, as usual.


squidbus:
Ignoring the test CI build failures as they exist on main.

misterj05:
We might want to disallow using system abseil period, as using system abseil compiled with GCC will lead to a linking error when compiling the vendored protobuf with clang. Since we're already not allowing protobuf to be a system library this should not be much of an issue.

This issue doesn't show on Arch because it's using GCC/G++ to compile shadps4.

PR #4882, opened 2026-08-20
video_core/gnmdriver: fix Naughty Dog titles stuck at black screen / 3 fps (Uncharted TNDC boots to main menu)

sergek88:
Uncharted: The Nathan Drake Collection (CUSA02320) is listed as "Nothing" in the compatibility tracker: it boots to a black screen and crawls at 3-4 fps, never reaching the title screen in reasonable time. After instrumenting the emulator (Tracy + custom fence/event tracing) we found three independent problems. With all three fixed, the game reaches the full title screen and its main menu is rendered and navigable:

| Title screen | Main menu |
|---|---|
| title | menu |

### 1. Replace the submission lock with a bounded gfx-ring drain

`sceGnmSubmitDone` currently sets `submission_lock` whenever the GPU is not idle; every subsequent submission then blocks until the WHOLE queue drains (reset only by the GpuIdle IRQ). ND engines keep their async compute rings parked in `WaitRegMem` at all times, so "whole queue idle" effectively means "once per interrupt cadence": Tracy shows `WaitGpuIdle` costing a mean of 58 ms (max 30 s!) per call and `sceGnmSubmitDone` 86 ms — with actual GPU work being ~6 s out of a 90 s capture. The game is fully serialized by the emulator, not by its own pacing.

The replacement: `sceGnmSubmitDone` performs a *bounded* (2 s) wait for the *gfx ring only* to drain. This preserves the guarantee ND relies on — all EOP fences of the submitted frame are written before the guest's next `FrameBegin` (their `fg-draw-mgr.cpp:234` assert checks exactly that) — without waiting on compute rings, and the timeout prevents deadlocks when ring progress depends on labels the guest writes after `submitDone` returns (observed during loading phases).

### 2. Defer the flip signal until the flip submit retires

`PatchFlipRequest` appends the frame-tick `EventWriteEop` AFTER the `PatchedFlip` NOP in the DCB. Signaling `GfxFlip` while parsing the NOP lets the present thread deliver the flip event to the guest *before* the tick value is written — the guest wakes up, reads a zero EOP tick and asserts. The upstream submission lock was masking this race by keeping the guest too slow to observe it. On real hardware an eop-flip cannot precede the EOP of its own submit either, so the signal is now accumulated and fired when the containing gfx submit retires.

### 3. Tame "unbounded" V# descriptors (num_records = 0xffffffff)

ND shaders bind V# descriptors with `num_records = 0xffffffff`, `stride = 4` — a legal "pointer to all of memory" pattern. The u32 multiply overflows to 0xfffffffc and `ClampRangeSize` clamps it to the end of the contiguous mapping (~2.7 GiB). Treating that as a regular SSBO is catastrophic:

* `gpu_modified_ranges.Add()` marks the whole guest heap GPU-modified, killing the stream-buffer fast path for every other buffer;
* every draw re-synchronizes whatever the guest wrote anywhere in the range — we measured a single `ObtainBuffer` call taking **12.25 seconds** during the menu loading phase, and 24 ms average `Draw` cost (vs 34 µs in light scenes);
* re-protecting ~700k pages per draw plus region locks held across uploads burns ~5 CPU cores in page-fault storms.

The fix: for descriptors matching this pattern (num_records == 0xffffffff and clamped size >= 1 GiB) don't mark the range written, and synchronize only a 16 MiB window near the base address, at most once per guest flip. Reads stay coherent (ND uses the descriptor for data near its base); a proper long-term solution would route such descriptors through the existing BDA page-table path.

Also demotes the per-draw `ClampRangeSize`/`ClampMode` warnings to debug — with sync logging they alone cost milliseconds per frame.

### Results (RTX 5080, Windows 11, base 1.00 + update 1.02)

| | upstream main | this PR |
|---|---|---|
| intro logos | ~35 min | ~2 min |
| full title screen | not reached (bare text after ~50 min) | ~23 min, full art |
| main menu | never | rendered, navigable |
| guest EOP-tick assert | n/a (masked by lock) | not reproducible |

The remaining slowness (menu load phase) is the guest's own streaming loop paying HLE event-delivery latency per job wave; it is independent of these fixes and we plan to look into it next.

Happy to split this into three PRs if preferred, gate any part behind a config option, or adapt the descriptor heuristic.

### DIFF
--- a/src/core/libraries/gnmdriver/gnmdriver.cpp
+++ b/src/core/libraries/gnmdriver/gnmdriver.cpp
@@ -2313,11 +2313,17 @@ s32 PS4_SYSV_ABI sceGnmSubmitCommandBuffers(u32 count, const u32* dcb_gpu_addrs[
 int PS4_SYSV_ABI sceGnmSubmitDone() {
     HLE_TRACE;
     LOG_DEBUG(Lib_GnmDriver, "called");
-    WaitGpuIdle();
-    if (!liverpool->IsGpuIdle()) {
-        submission_lock = true;
-    }
     liverpool->SubmitDone();
+    // Pace the guest by (bounded) draining of the gfx ring instead of the
+    // submission lock: the lock forces every submission after submitDone to
+    // wait until the WHOLE queue is idle, including async compute rings that
+    // engines like Naughty Dog's keep busy at all times - which serializes
+    // rendering to the interrupt cadence (~50-90 ms per frame observed in
+    // Uncharted TNDC). Draining just the gfx ring before returning keeps the
+    // engine's own frame pacing intact: all EOP fences of this frame are
+    // written before the guest's next FrameBegin. The wait is bounded because
+    // ring progress may depend on labels the guest writes after submitDone.
+    liverpool->WaitForGfxRingDrain();
     send_init_packet = true;
     ++frames_submitted;
     DebugState.IncGnmFrameNum();
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -123,7 +123,7 @@ u64 MemoryManager::ClampRangeSize(VAddr virtual_addr, u64 size) {
     clamped_size = std::min(clamped_size, size);
 
     if (size != clamped_size) {
-        LOG_WARNING(Kernel_Vmm, "Clamped requested buffer range addr={:#x}, size={:#x} to {:#x}",
+        LOG_DEBUG(Kernel_Vmm, "Clamped requested buffer range addr={:#x}, size={:#x} to {:#x}",
                     virtual_addr, size, clamped_size);
     }
     return clamped_size;
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -139,14 +139,14 @@ void Liverpool::Process(std::stop_token stoken) {
                     }
                     submit_cv.notify_all();
                 }
-                if (curr_qid == GfxQueueId) {
-                    while (pending_flip_signals) {
-                        --pending_flip_signals;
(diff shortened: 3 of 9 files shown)


sergek88:
Pushed an improvement to fix #2 (`c1d7c4c`): the flip signal is now delivered at the **end of DCB execution** instead of at submit retire.

Signaling at retire turned out to create a head-of-line dependency — during streaming phases the gfx ring can legally stall for seconds on guest-written labels, postponing presentation far beyond the point where the flip fence is already written. The DCB-end point keeps the original guarantee (every packet after the `PatchedFlip` NOP has executed, including the frame fence in all four `PrepareFlip` variants) while delivering the event microseconds after the NOP.

Measured impact on Uncharted TNDC (same setup as above): intro logos **2 min → 41 s**, menu.pak streaming phase **~20 min → ~5 min**, still no guest asserts.

StevenMiller123:
Please review our AI-use rules, located in https://github.com/shadps4-emu/shadPS4/blob/main/CONTRIBUTING.md

sergek88:
Fair point, and my bad — I should have read the contributing rules before opening this. So, to be transparent: yes, I work with an AI assistant. It helps me instrument the emulator, profile and draft code. The diagnosis itself is backed by real measurements on my machine — Tracy captures, timing series, logs, before/after runs — and I'm happy to share any of it and to explain every decision in this patch myself, in my own words. Ask me anything.

I'm still actively working on this title (loading is already several times faster on my end, and I'm aiming for playable next), so I'd like to ask for this to be reopened — or, if you prefer, I'll resubmit a properly disclosed PR when the next batch of fixes is ready. Whichever works better for the project. Thanks for the quick review, and sorry for the noise.

PR #4884, opened 2026-08-20
Improve tsharp robustness

raphaelthegreat:
* Check if the address if a valid gpu address and also reject sharps that have invalid data or number formats
* Clamp base_array to actual allocated slices in image view

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -123,8 +123,8 @@ u64 MemoryManager::ClampRangeSize(VAddr virtual_addr, u64 size) {
     clamped_size = std::min(clamped_size, size);
 
     if (size != clamped_size) {
-        LOG_WARNING(Kernel_Vmm, "Clamped requested buffer range addr={:#x}, size={:#x} to {:#x}",
-                    virtual_addr, size, clamped_size);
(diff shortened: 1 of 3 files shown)


StevenMiller123:
I think we should at least log something when we reject invalid sharps, given we know our current sharp tracking can cause them.

diogoaduartereis:
Hi @raphaelthegreat not sure if this is helpful in any way but although this PR and the other one seem to help with The Order 1886. For my rtx 4060 and i5-12400F, it still crashes. Do let me know if you find anything useful here.
[CUSA00785.log](https://github.com/user-attachments/files/31270664/CUSA00785.log)
Edit: it seemed to not write everything into teh file, it crashed here: [Debug] <Critical> (shadPS4:GpuCommandProcessor) vk_rasterizer.cpp:675 BindBuffers: Assertion Failed!

raphaelthegreat:
Are you maybe further in the game or trying to load a different area?

diogoaduartereis:
The first chapter is fine until the end. But is the second one that crashes. I had a previous save file on that part already, but any attempts on chapter 2 crash.

PR #4885, opened 2026-08-20
AbstractFs cleanup

georgemoralis:
Solving some slowness issues , caching added


src/core/file_sys/backends/host_fs.cpp
```diff
@@ -133,6 +134,21 @@ void HostDirectory::Rewind() {
 HostFsBackend::HostFsBackend(std::filesystem::path root, bool read_only)
     : m_root(std::move(root)), m_read_only(read_only) {}
 
+std::string HostFsBackend::CacheKey(std::string_view rel_path) {
+    std::string key = Common::ToLower(std::string{rel_path});
+    std::replace(key.begin(), key.end(), '\\', '/');
+    return key;
+}
```
squidbus: How necessary is this cache key? If we're looking to optimize here by skipping resolves, I think also not having to do these scans over the string to convert to lower and change slashes would help. I'm not sure we need to care much about optimizing for paths being used with different casing/slash style, since the main thing we care about (in my view, to get back toward previous performance) is that repeated uses of a path within the same file operation are not re-resolving. Plus I'm not sure that many games would realistically be using the exact same path with a bunch of different slash/capitalization styles enough to make it worth it.

georgemoralis: you are right this was unneccasery optimization , might case more issues than solved , removed

squidbus: Well, I was fine with the cache, I just meant that I'm not sure we need to sanitize the key like this vs just key-ing it on the path as-is. But if it didn't make much of a difference anyway then that's fine.

src/core/libraries/kernel/file_system.cpp
```diff
@@ -343,22 +354,12 @@ s64 PS4_SYSV_ABI sceKernelWrite(s32 fd, const void* buf, u64 nbytes) {
...
 }
 
-static thread_local std::vector<u8> file_buf{};
-
 s64 ReadFile(Core::FileSys::File* file, void* buf, u64 nbytes) {
     const auto* memory = Core::Memory::Instance();
     // Invalidate up to the actual number of bytes that could be read.
     const auto remaining = file->GetSize() - file->Tell();
     memory->InvalidateMemory(reinterpret_cast<VAddr>(buf), std::min<u64>(nbytes, remaining));
-    if (file_buf.capacity() < nbytes) {
```
StevenMiller123, line 353: Why were these changes removed?
As far as I can tell, you haven't fixed the issue with page tracking breaking file reads in this PR, so this workaround should stay.

StevenMiller123 (changes requested):

src/common/io_file.h
```diff
@@ -169,8 +169,14 @@ class IOFile final {
 
     template <typename T>
     size_t ReadRaw(void* data, size_t size) const {
-        u64 read = std::fread(data, sizeof(T), size, file);
-        ASSERT_MSG(std::ferror(file) == 0, "Failed to read file, error = {}", std::strerror(errno));
+        if (!IsOpen()) {
+            return 0;
+        }
+        const u64 read = std::fread(data, sizeof(T), size, file);
+        if (std::ferror(file) != 0) {
```
StevenMiller123: Most of the IOFile code is based on the assumption that these reads never fail, which is why I made this an assert. Is there a good reason to downgrade this?

src/common/io_file.h
```diff
@@ -185,7 +191,6 @@ class IOFile final {
         u64 written = std::fwrite(data.data(), sizeof(T), data.size(), file);
         ASSERT_MSG(std::ferror(file) == 0, "Failed to write to file, error = {}",
                    std::strerror(errno));
-        std::fflush(file);
```
StevenMiller123, line 188: We need to flush after writes, otherwise games that go straight into reading will have issues reading outdated data.
This will break saves in Dragon Ball Xenoverse titles.

src/common/io_file.h
```diff
@@ -222,8 +227,8 @@ class IOFile final {
         }
 
         u64 bytes = std::fwrite(&object, sizeof(T), 1, file) == 1;
-        ASSERT_MSG(std::ferror(file) == 0, "Failed to read file, error = {}", std::strerror(errno));
-        std::fflush(file);
```
StevenMiller123, line 226: This fflush can't be removed, same reason as above.

StevenMiller123 (changes requested):
More things I noticed.

StevenMiller123 (commented):
Not sure how much benefit there is to complicating the file reading workaround this much, but the regression is fixed.

tochnonechelovek:
Latest commit of this PR finally fixes slow loading in CUSA05078 - Paladins.

src/core/file_sys/backends/host_fs.cpp
```diff
@@ -119,7 +151,7 @@ bool HostDirectory::Next(DirEntry& out) {
     std::error_code ec;
     out.name = entry.path().filename().string();
     out.is_directory = entry.is_directory(ec);
-    out.size = out.is_directory ? 0 : entry.file_size(ec);
```
StevenMiller123, line 122: Why was this changed?

src/core/libraries/kernel/file_system.cpp
```diff
@@ -234,13 +240,18 @@ s32 PS4_SYSV_ABI open(const char* raw_path, s32 flags, u16 mode) {
...
+        errno = 0;
+        file->handle = mnt->OpenResolved(resolved, access_mode);
         if (!file->handle || !file->handle->IsOpen()) {
             h->DeleteHandle(handle);
-            *__Error() = (write || rdwr || truncate) && mnt->GetMount(raw_path)->read_only
-                             ? POSIX_EROFS
-                             : POSIX_EIO;
-            LOG_ERROR(Kernel_Fs, "Opening {} failed, backend did not serve the file", raw_path);
+            if (const int host_errno = errno; host_errno != 0) {
+                SetPosixErrno(host_errno);
```
StevenMiller123, line 248: Relying on host error code is never a good idea, unless you're only planning to make shadPS4 run on FreeBSD and MacOS 😆 

Why was this changed?

src/core/libraries/kernel/file_system.cpp
```diff
@@ -343,22 +354,53 @@ s64 PS4_SYSV_ABI sceKernelWrite(s32 fd, const void* buf, u64 nbytes) {
...
+        file_buf.resize(chunk);
     }
-    s64 bytes = file->Read(file_buf.data(), nbytes);
-    if (bytes < 0) {
-        return bytes;
+
+    auto* dst = static_cast<u8*>(buf);
+    s64 total = 0;
+    while (static_cast<u64>(total) < nbytes) {
+        const u64 want = std::min<u64>(ReadChunkSize, nbytes - total);
```
StevenMiller123, line 376: I'm guessing the chunk size thing is to avoid having too much memory consumed by the file buffer, but for games that read much larger chunks of data, this could harm performance.
I'd at least suggest bumping chunk size. Most games aren't dedicating too many threads to file reads, so even a size of 1MB or more shouldn't really cause harm.

StevenMiller123 (changes requested):
This needs proper testing before I'll accept a merge. Too many things that could easily slip through the cracks here.

src/core/libraries/kernel/file_system.cpp
kalaposfos13: This also breaks the fix I just made here, but in the opposite way. Now, in guest readonly but host read-write mounts, creating a file in a nonexistent folder, the host will return EIO, even though we need to return EROFS anyway (unless we don't, I don't know, I haven't tested it myself). Also, Linux having different enums is not an issue because SetPosixErrno handles conversion, and as long as the error codes it returns are the same as what FreeBSD would return too, which they should be, given that both follow POSIX standards, it should not cause problems.

StevenMiller123: Linux isn't always returning the same code as FreeBSD or PS4, that's why I mention what I have.

src/core/libraries/kernel/file_system.cpp
red-prig: Unfortunately, the POSIX standard only declares names, so the error codes for Linux and FreeBSD don't match at all.

PR #4889, opened 2026-08-21
More gpu improvements 

raphaelthegreat:
Upstreams 2 commits from https://github.com/shadps4-emu/shadPS4/pull/4798 that I found LGTM
Stubs HTILE to have uncompressed ZMASK bits. Still not ideal but better than before


diogoaduartereis:
image
With this PR I can finally boot into the second chapter! I couldn't do that previously or with the claude PR

PR #4896, opened 2026-08-21
shader_recompiler: decorate written storage buffers as Coherent

junminlee2004:
This PR fixes uncontrollably stretching hair on movement for NVIDIA gpus in TLOU Remastered by decorating written storage buffers as `Coherent`. 
According to OpenGL Shading Language 4.60 specification, section 4.10 "Memory Qualifiers":
https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.html

"_When accessing memory using variables not declared as coherent, the memory accessed by a shader may be cached by the implementation to service future accesses to the same address. Memory stores may be cached in such a way that the values written may not be visible to other shader invocations accessing the same memory. **The implementation may cache the values fetched by memory reads and return the same values to any shader invocation accessing the same memory, even if the underlying memory has been modified since the first memory read**. While variables not declared as coherent may not be useful for communicating between shader invocations, using non-coherent accesses may result in higher performance._"

Hair threads exchange positional data through buffers and without `Coherent`, NVIDIA can legally reuse stale, cached copies. The fix is to decorate written storage buffers as `Coherent`.

Before (stale):
TLOU-before

After: (Coherent)
TLOU-after

### DIFF
--- a/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
+++ b/src/shader_recompiler/backend/spirv/spirv_emit_context.cpp
@@ -782,6 +782,11 @@ EmitContext::BufferSpv EmitContext::DefineBuffer(bool is_written, u32 elem_shift
     Decorate(id, spv::Decoration::DescriptorSet, 0U);
     if (!is_written) {
         Decorate(id, spv::Decoration::NonWritable);
+    } else {
+        // Guest shaders may use buffer stores with s_barrier for handoff
+        // between invocations of a workgroup. Mark written buffers coherent
+        // so accesses are not cached in registers across the barrier.
+        Decorate(id, spv::Decoration::Coherent);
     }
     switch (buffer_type) {
     case BufferType::GdsBuffer:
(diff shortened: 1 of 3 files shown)


raphaelthegreat (approved):

baggins183:
Im not sure about perf impact but this could be significant. Can we track if the sharp is also read and only emit  coherent then?

red-prig:
The MUBUF and MIMG instructions literally have a GLC (globaly coherent) bit field. Shouldn't we just move this attribute to spirv?

raphaelthegreat:
Hmm yes I forgot about the GLC bit, that should probably be checked

junminlee2004:
Every one of the 219 MUBUF instructions in the hair shader carries the GLC bit field and the game's compiler wrote 0s into every single one.

red-prig:
There is also such an option as the bit parameters of the L1, L2 cache written in V#

junminlee2004:
I think the better idea is to implement the suggestion by @baggins183. The only way to forbid a compiler optimization (NVIDIA driver specific in this case) is a compile-time contract, and under the GLSL450 memory model that contract is the Coherent decoration

red-prig:
I think this might affect game performance if you always set it to storage buffers

baggins183:
I think you should check mtype of the buffer as red prig said. Looks like it's not actually used at all by us. About the "compile-time contract thing" you mentioned, this could be factored into buffer specialization so it becomes a compile time contract if that's what you meant

junminlee2004:
I've now moved the coherency attribute from always being set to written storage buffers to guest V# mtype derived from the game.

PR #4899, opened 2026-08-22
shader_recompiler: Improve ssa rewrite pass correctness and readability

raphaelthegreat:
The ssa rewrite pass is based on the spirv-opt pass of the same name:
https://github.com/KhronosGroup/SPIRV-Tools/blob/main/source/opt/ssa_rewrite_pass.cpp

This commit brings the pass closer to that reference implementation. Notably this fixes applying the pass to unstructured graphs which seems to be broken otherwise

Additionally the manual stack iteration of ReadVariable has been replaced with recusion which is much easier to reason about

Undef handling has also been changed, before each block would define its own Undef instruction, now the pass will keep track and reuse compatible instructions add insert them to the first block

### DIFF
--- a/src/shader_recompiler/ir/passes/ir_passes.h
+++ b/src/shader_recompiler/ir/passes/ir_passes.h
@@ -13,7 +13,7 @@ void InjectClipDistanceAttributes(IR::Program& program, RuntimeInfo& runtime_inf
 
 namespace Shader::Optimization {
 
-void SsaRewritePass(IR::BlockList& program);
+void SsaRewritePass(IR::Program& program);
 void IdentityRemovalPass(IR::BlockList& program);
 void DeadCodeEliminationPass(IR::Program& program);
 void ConstantPropagationPass(IR::BlockList& program);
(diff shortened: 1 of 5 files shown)


diogoaduartereis:
[CUSA00785.log](https://github.com/user-attachments/files/31336328/CUSA00785.log)
Seems to regress the order 1886 on my build compared to your latest PR. Chapter 2 opening results in black screen

raphaelthegreat:
I can't reproduce said black screen

StevenMiller123:
Why is your vblank frequency set to 9999?

raphaelthegreat:
Also maybe you should disable pipeline cache?

diogoaduartereis:
@raphaelthegreat  it also happened before your latest PR: https://github.com/shadps4-emu/shadPS4/pull/4889 where some people could actually enter the game. Only after that PR was I able to enter chapter 2. I will try different settings.
@StevenMiller123 No reason whatsover tbh. But it shouldn't be the issue since I had it before as well. I will report back.

diogoaduartereis:
Well my bad, it seems that putting readbacks to precise makes me get into the game once again. Before I had it at relaxed.

diogoaduartereis:
image

diogoaduartereis:
Ignore what I had said  before guys. It seems to also be working with readbacks at relaxed and shader cache enabled. So not sure what happened. The game seems to be looking a bit better. I've seem some effects from gunshot wounds appear where they didn't before.

PR #4900, opened 2026-08-22
shader_recompiler: Add support for virtual registers

raphaelthegreat:
Useful for expressing load/stores to non guest architectural state.
Each virtual register is identified by a unique (index, type) pair


diogoaduartereis:
Makes the Order 1886 work without DMA!

PR #4908, opened 2026-08-23
Http2 fixes

georgemoralis:
- sceHttp2DeleteRequest now uses find()/erase() instead of operator[]
- sceHttp2CreateRequestWithURL deletes the connection when request creation fails afterward.
- sceHttp2Term now forwards to sceHttpTerm
- sceHttp2GetResponseContentLength now forwards instead of returning OK

### DIFF
--- a/src/core/libraries/network/http2.cpp
+++ b/src/core/libraries/network/http2.cpp
@@ -8,10 +8,14 @@
 #include "core/libraries/network/http2.h"
 
 #include <map>
+#include <mutex>
+#include <set>
 
 namespace Libraries::Http2 {
 
+std::mutex g_http2_mutex;
 std::map<s32, s32> requests_to_connections{};
+std::set<s32> g_active_contexts{};
 
 s32 PS4_SYSV_ABI sceHttp2AbortRequest(s32 req_id) {
     LOG_ERROR(Lib_Http2, "(STUBBED) called");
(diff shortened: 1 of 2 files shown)


src/core/libraries/network/http2.cpp
```diff
@@ -99,17 +112,34 @@ s32 PS4_SYSV_ABI sceHttp2DeleteCookieBox() {
...
+    const s32 result = Libraries::Http::sceHttpDeleteRequest(req_id);
     if (result < 0) {
         LOG_ERROR(Lib_Http2, "Failed to delete HTTP request, error = {:#x}", result);
         return result;
     }
-    s32 conn_id = requests_to_connections[req_id];
-    result = Libraries::Http::sceHttpDeleteConnection(conn_id);
-    if (result < 0) {
-        LOG_ERROR(Lib_Http2, "Failed to delete HTTP connection, error = {:#x}", result);
+    s32 conn_id = -1;
```
StevenMiller123: Connections are created with requests, so maybe we should start running these functions under a mutex, so validating the connections isn't necessary?

src/core/libraries/network/http2.cpp
```diff
@@ -179,6 +214,12 @@ s32 PS4_SYSV_ABI sceHttp2Init(s32 net_id, s32 ssl_id, u64 pool_size, s32 max_req
     s32 ctx_id = Libraries::Http::sceHttpInit(net_id, ssl_id, pool_size);
     if (ctx_id < 0) {
         LOG_ERROR(Lib_Http2, "Failed to init HTTP context, error = {:#x}", ctx_id);
+        return ctx_id;
+    }
+
+    {
+        std::lock_guard<std::mutex> lock(g_http2_mutex);
+        g_active_contexts.insert(ctx_id);
```
StevenMiller123: I don't see a point in tracking active contexts if we're still returning the http1 context as-is. Not a bad idea if we want to start actually implementing this library, but until then, it's unnecessary in my opinion.

StevenMiller123 (commented):
Mostly fine, though I think this can be simplified a bit and still accomplish what it's looking to do.

georgemoralis:
@StevenMiller123 done

StevenMiller123 (approved):
Looks good

PR #4910, opened 2026-08-23
Net Fixes

georgemoralis:
Some error code corrections and sceNetGetSockInfo


StevenMiller123 (approved):
The code itself looks fine, I'm assuming you've properly checked to make sure you're returning the right stuff?

georgemoralis:
yes although it appears that this function mostly used in P2P scokets which are unimplemented so it fails as expected

PR #4916, opened 2026-08-25
usbd: add a loopback listener for the emulated Dimensions Toypad

harrysof:
Placing a figure is only possible through the Dimensions Manager dialog, which means reaching for a mouse every time a game asks for a different character. This adds an opt-in loopback TCP listener so an external companion app can do it from a controller instead.

The protocol is the one the Cemu and RPCS3 Toypad forks already speak (LOAD / REMOVE / MOVE, plus GET_LED), so an app written for either of those works here unchanged. It binds 127.0.0.1 only and does nothing at all unless dimensionsListenerPort is set - the default of 0 leaves it disabled, so nothing binds for users who do not ask for it.

GET_LED returns a snapshot of the LED state the game is driving. The toypad now parses the 0xC0..0xC8 LED commands as they pass through the interrupt path and keeps a per-region mirror of mode, colour and timings, with a serial that only advances when a command actually changed something, so a poller can skip what it has already applied and never restarts a flash mid-cycle.

The Dimensions Manager dialog and the existing USB_*_FIGURE IPC commands are untouched; this is an additional interface, not a replacement.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -605,6 +605,8 @@ set(USBD_LIB src/core/libraries/usbd/usbd.cpp
              src/core/libraries/usbd/usb_backend.h
              src/core/libraries/usbd/emulated/dimensions.cpp
              src/core/libraries/usbd/emulated/dimensions.h
+             src/core/libraries/usbd/emulated/dimensions_listener.cpp
+             src/core/libraries/usbd/emulated/dimensions_listener.h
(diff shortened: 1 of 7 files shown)


StevenMiller123:
Your PR does not meet our contribution rules, please review the AI rules we have in https://github.com/shadps4-emu/shadPS4/blob/main/CONTRIBUTING.md

Issue #4645, opened 2026-06-29
[APP BUG]: Shader cache stops compiling on consequent launches of games


isshininu:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.

### Describe the Bug

Shader cache stops compiling on consequent launches of all games

### Reproduction Steps

at first launch of game with option enabled, shader cache folder is created, with shaders and profile.bin
on consequent launches new shaders are not compiling.
This problem can be avoided by deleting profile.bin, then new shaders continue to compile

### Specify OS Version

Windows 10

### CPU

Intel core i5-10600kf

### GPU

Nvidia 3070 RTX

### Amount of RAM in GB

48

PR #4928, opened 2026-08-27
video_core: rebuild the pipeline cache instead of disabling it forever

sergek88:
**About AI, up front:** I work with an AI assistant. It helps me instrument the emulator, profile it and draft code. Everything below was measured on my own machine, and I can walk through every line of this change myself — ask me anything.

## The problem

Once the pipeline cache is found incompatible, it never recovers.

The profile blob is written in exactly one place — when it is missing. On a mismatch `WarmUp` closes the database and returns, and the stale blob stays on disk. The next launch loads that same stale profile, mismatches again, closes again. Shaders are recompiled from scratch on every single launch until the user deletes the cache folder by hand, which most people don't know exists.

This is what #4645 reported: *"shader cache stops compiling on consequent launches... this problem can be avoided by deleting profile.bin"*. #4704 fixed one cause of the mismatch (the padding one), but not the inability to recover from any other cause.

## It is happening right now

#4874 removed `max_ubo_size` from `Shader::Profile`. That changed `sizeof(Profile)` from 72 to 60, so every cache written before that commit fails the size check on every launch.

Same machine, same game, same cache folder (2712 files, 72-byte profile), only the build differs.

Current main (pre-release `54ec7d0`):

```
run 1  Pipeline cache profile has unexpected size (72 != 60). Ignoring the cache
       -> profile.bin still 72 bytes, 2712 files
run 2  Pipeline cache profile has unexpected size (72 != 60). Ignoring the cache
       -> profile.bin still 72 bytes, 2712 files
```

Nothing is ever preloaded, and nothing ever will be.

With this patch:

```
run 1  Pipeline cache profile has unexpected size (72 != 60). Discarding it
       -> cache reset, profile.bin 60 bytes
run 2  Preloaded 51 pipelines
```

## What the change does

Discard the incompatible cache and start a fresh one instead of leaving it in place. `DataBase::Reset()` closes the database, removes the cache directory (or archive) and reopens it empty; `WarmUp` then writes the current profile and continues exactly as it does on a first launch.

Dropping the stored pipelines is deliberate: they were built against the previous profile, so reusing them would not be safe. Only the profile blob is rewritten in the missing-blob path, so this is the only way the cache can come back.

I also folded the two compatibility checks into one path with a single exit. To me it reads better than three separate returns, but if you would rather keep the original shape, say so and I will restore it.

## Testing

- CI: Windows, Linux and macOS.
- By hand on Windows with Uncharted: The Nathan Drake Collection (CUSA02320), the runs shown above, against a real cache left over from an older build.

### DIFF
--- a/src/video_core/cache_storage.cpp
+++ b/src/video_core/cache_storage.cpp
@@ -120,6 +120,20 @@ void DataBase::Open() {
     opened = true;
 }
 
+void DataBase::Reset() {
+    Close();
+
+    std::error_code ec{};
+    std::filesystem::remove_all(cache_path, ec);
+    if (ec) {
+        LOG_ERROR(Render, "Failed to discard the pipeline cache at {}: {}",
+                  cache_path.string().c_str(), ec.message());
+    }
+
+    ar_is_read_only = true;
+    Open();
+}
+
 void DataBase::Close() {
     if (!IsOpened()) {
         return;
--- a/src/video_core/cache_storage.h
+++ b/src/video_core/cache_storage.h
@@ -28,6 +28,8 @@ public:
 
     void Open();
     void Close();
+    /// Discards the stored cache and reopens it empty.
+    void Reset();
     [[nodiscard]] bool IsOpened() const {
         return opened;
     }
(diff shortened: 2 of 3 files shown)


StevenMiller123:
Your PR does not meet our contribution rules, please review the AI rules we have in https://github.com/shadps4-emu/shadPS4/blob/main/CONTRIBUTING.md

PR #4932, opened 2026-08-28
Add an option to orient motion controls for a handheld

### LINKED ISSUE #3871: [Feature Request]: Toggle for Vertical Oriented Motion Controls
### Checklist

### Description

Most implementations of motion controls in PS4 games account for the controller being held horizontally, relatively flat. But for certain controllers, or namely, handheld PCs, you will be holding the device vertically. This can change the functionality of the gyroscope pretty drastically.

What I am requesting, is a toggle that adjusts the parameters of the emulated Dualshock motion controls to behave as expected for handhelds, or any controller that you may hold vertically. This could behave similarly to how the gyroscope functions through Remote Play on PS Vita and (I believe) PS Portal, but more helpfully, how it behaves through the open source client, Chiaki-NG.

I’ve been testing Gravity Rush Remastered, on Steam Deck, using a solution I will detail below, and it feels as if the yaw and roll functions are reversed (I’m not educated enough on the matter to tell you if that is literally the case, if it is that simple, but it may be). 

For anyone looking to test any of this on Steam Deck, getting its gyroscope to work on ShadPS4 requires the following:

1. Add the launch option “SDL_GAMECONTROLLER_IGNORE_DEVICES=0 %command%” not including quotation marks, to the emulator in Game Mode

2. Step one alone will work, but to avoid the visual of an on-screen mouse, in controller settings, make your Gyro Behavior “Gyro To Mouse [Beta]” and in its settings, change the button behavior to a random back button.

And you’re done. Saw someone on Reddit doing this for Vita3K, on a whim tested it on ShadPS4, and it worked. If you do implement this vertically-orientated gyroscope feature, you may even want to put some of this information in the details that appear when hovering over the toggle, as I don’t believe it’s common knowledge.

### Reason

I understand that a feature which would primarily help handheld PC users may not be of the highest priority, but unless I’m underestimating it greatly, this seems like a relatively simple feature that fixes one of the few, fundamental hardware incompatibilities that comes with playing PS4 games on a handheld. A feature that will future-proof ShadPS4 as handheld PCs only become more prevalent.

### Examples

The open source PlayStation Remote Play client, Chiaki-NG, has a toggle in its general settings named “Steam Deck Vertical” that corrects the function of the gyroscope when being held vertically, and I recommend looking into it. I’ve tested Gravity Rush through it as well, and the difference is night and day with this feature.

elpendor:
Closes #3871.

SDL reports gamepad motion in the frame of a controller lying face up, which is how a DualShock is held — so passing `event->gsensor.data` straight through is correct for one. A handheld is held upright instead, and its frame is 90 degrees away about X. X is left-right in both, so pitch arrives correctly while yaw and roll land on each other: tilting the device moves the camera, turning it does nothing. On a Steam Deck a real 90 degree yaw arrives as -89.1 on `z` against +23.5 on `y`.

`Input.motion_controls_vertical` rotates motion -90 degrees about X — `(x, y, z)` to `(x, -z, y)` — before it reaches the emulated pad. Gyro and accelerometer both, since they share the frame. Per-game overrideable like the other input settings.

Off by default, and not auto-detected: nothing SDL reports says how a device is being held, and the rotation is wrong for a controller resting face up.

Built and run on a Steam Deck (OLED) in a retail game with motion controls. Off, behaviour is unchanged; on, a 90 degree yaw reads 92.0 about true vertical and the camera turns with the device.

No GUI checkbox here — config and per-game overrides only. Happy to add one to the Qt settings dialog if that is wanted.

### DIFF
--- a/src/sdl_window.cpp
+++ b/src/sdl_window.cpp
@@ -423,13 +423,13 @@ void WindowSDL::OnGamepadEvent(const SDL_Event* event) {
         case SDL_SENSOR_GYRO:
             gamepad = controllers.GetGamepadIndexFromJoystickId(event->gsensor.which);
             if (gamepad < 5) {
-                controllers[gamepad]->UpdateGyro(event->gsensor.data);
+                controllers[gamepad]->UpdateGyro(event->gsensor.data, true);
             }
             break;
         case SDL_SENSOR_ACCEL:
             gamepad = controllers.GetGamepadIndexFromJoystickId(event->gsensor.which);
             if (gamepad < 5) {
-                controllers[gamepad]->UpdateAcceleration(event->gsensor.data);
+                controllers[gamepad]->UpdateAcceleration(event->gsensor.data, true);
             }
             break;
         default:
--- a/src/core/emulator_settings.h
+++ b/src/core/emulator_settings.h
@@ -324,7 +324,8 @@ struct InputSettings {
     Setting<int> usb_device_backend{UsbBackendType::Real}; // specific
     Setting<bool> use_special_pad{false};
     Setting<int> special_pad_class{1};
-    Setting<bool> motion_controls_enabled{true}; // specific
+    Setting<bool> motion_controls_enabled{true};   // specific
+    Setting<bool> motion_controls_vertical{false}; // specific
     Setting<bool> use_unified_input_config{true};
     Setting<std::string> default_controller_id{""};
     Setting<bool> background_controller_input{false}; // specific
@@ -342,6 +343,8 @@ struct InputSettings {
             make_override<InputSettings>("usb_device_backend", &InputSettings::usb_device_backend),
             make_override<InputSettings>("motion_controls_enabled",
                                          &InputSettings::motion_controls_enabled),
+            make_override<InputSettings>("motion_controls_vertical",
+                                         &InputSettings::motion_controls_vertical),
             make_override<InputSettings>("background_controller_input",
                                          &InputSettings::background_controller_input),
             make_override<InputSettings>("ime_accessibility_enabled",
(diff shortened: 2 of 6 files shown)


src/sdl_window.cpp
```diff
@@ -42,6 +42,24 @@ namespace Frontend {
...
 using namespace Libraries::Pad;
 
+/// SDL reports gamepad motion in the frame of a controller lying face up. A
+/// handheld is held upright instead, so its Y and Z are that frame's Z and -Y
+/// -- which is why yaw and roll feel swapped. Rotating -90 degrees about X puts
+/// them back where a game written for a DualShock expects them.
+///
+/// Off by default: nothing SDL reports says how the device is being held.
+static void OrientMotionForDevice(const float in[3], float out[3]) {
+    if (!EmulatorSettings.IsMotionControlsVertical()) {
```
kalaposfos13: You can save an unnecessary copy of this buffer by inlining this logic into GameController::UpdateGyro, replacing the memcpy there.

elpendor: Done — the transform now writes straight into `gyro_buf`/`accel_buf`, so the stack buffer is gone.

One thing it ran into: `input_mouse.cpp`'s `EmulateGyro` calls both functions too, with values already in the frame a game expects (a constant `(0, 9.81, 0)` accel and mouse deltas mapped to pitch/yaw). Rotating those would break mouse gyro whenever the option is on, so the functions take a `device_frame` flag defaulting to false, and only `sdl_window.cpp` passes true.

Also added the setting to the Big Picture Input tab, under "Enable Motion Controls" — it had no UI before, so the only way to reach it was hand-editing the config.

Rebased on current main.

StevenMiller123:
This PR is valid, but it does not meet our contributing guidelines. Please review our guidelines at https://github.com/shadps4-emu/shadPS4/blob/main/CONTRIBUTING.md (particularly the ones surrounding AI use), adjust your branch accordingly, then open a new PR.

PR #4933, opened 2026-08-28
Http module fixups and implementations

georgemoralis:
- Send User-Agent configured in HTTP template
- Send default User-Agent if not any


georgemoralis:
@sgdc3 this fixes mostly applies to LBP3 , if you can check them , NpLookup requires latest shadnet (already online)

sgdc3:
@georgemoralis I tried to rebase my branch on your PR but then I am no longer able to connect to the LBP Bonsai private server.

Turns out it's the new User-Agent handling in sceHttpCreateTemplate: it always appends the libhttp tag to whatever the game passes. So LBP3's PatchworkLBP3 1.0 becomes PatchworkLBP3 1.0 libhttp/4.50 (PlayStation 4) but the server wants the exact string, and it just answers 403 on login.

This fixed it for me: only build the libhttp UA when the game doesn't pass one, otherwise send it untouched:
```
tmpl.user_agent = (userAgent && *userAgent) ? userAgent : BuildLibhttpUserAgent("");
```

With that, login returns 200 and everything works again.

No idea what retail libhttp actually does when the app sets its own UA, but judging from the server it probably doesn't append anything, real consoles connect to Bonsai just fine.

sgdc3:
@georgemoralis While at it: LBP3 online also needs this one https://github.com/sgdc3/shadPS4/commit/9ecf0722b82c66c0b2e015a63f9e59ff51ea7871

sceHttpSetInflateGZIPEnabled is mapped onto httplib's transparent decompression, which also decodes deflate, but LBP3 asks for Accept-Encoding: deflate and decodes the responses itself, so they came back corrupted and the game just kept retrying.

kalaposfos13:
Out of curiosity, does LLE sceHttp+sceSsl2 also work for LBP3? You'll have to mount system certs for the latter LLE library to work, which you can do with the --mount flag.

georgemoralis:
> @georgemoralis I tried to rebase my branch on your PR but then I am no longer able to connect to the LBP Bonsai private server.
> 
> Turns out it's the new User-Agent handling in sceHttpCreateTemplate: it always appends the libhttp tag to whatever the game passes. So LBP3's PatchworkLBP3 1.0 becomes PatchworkLBP3 1.0 libhttp/4.50 (PlayStation 4) but the server wants the exact string, and it just answers 403 on login.
> 
> This fixed it for me: only build the libhttp UA when the game doesn't pass one, otherwise send it untouched:
> 
> ```
> tmpl.user_agent = (userAgent && *userAgent) ? userAgent : BuildLibhttpUserAgent("");
> ```
> 
> With that, login returns 200 and everything works again.
> 
> No idea what retail libhttp actually does when the app sets its own UA, but judging from the server it probably doesn't append anything, real consoles connect to Bonsai just fine.

check http.sprx decompile and the lib actually append the libhttp/4.50 (PlayStation 4) after any game user agent . Is the server ps3 or ps4 compatible?

georgemoralis:
> @georgemoralis While at it: LBP3 online also needs this one [sgdc3@9ecf072](https://github.com/sgdc3/shadPS4/commit/9ecf0722b82c66c0b2e015a63f9e59ff51ea7871)
> 
> sceHttpSetInflateGZIPEnabled is mapped onto httplib's transparent decompression, which also decodes deflate, but LBP3 asks for Accept-Encoding: deflate and decodes the responses itself, so they came back corrupted and the game just kept retrying.

yah you are right about this . Added a more generic option for this on my latest commit if you want to check if it works

src/core/libraries/network/http.cpp
```diff
@@ -955,6 +953,26 @@ static bool HeaderNameMatches(std::string_view a, std::string_view b) {
     return true;
 }
 
+static std::string GetLibhttpSystemVersionString() {
+    u32 sw_hex = CURRENT_FIRMWARE_VERSION;
+    Libraries::Kernel::SwVersionStruct sw{};
+    sw.struct_size = sizeof(sw);
+    if (Libraries::Kernel::sceKernelGetSystemSwVersion(&sw) == ORBIS_OK) {
```
StevenMiller123: I don't remember us having any way to adjust system software version, is there any reason you're calling this despite already retrieving the sw version through our CURRENT_FIRMWARE_VERSION define?

sgdc3:
@georgemoralis Tried both. You're right about the UA, it was a mistake on my side. The gzip one needs a bit more.

**gzip:** `ade79b05` fixes the request side, but the body is still getting mangled. `cli.set_decompress(plan.settings.inflate_gzip)` is unconditional and `inflate_gzip` defaults to `true` (`http.cpp:57`), and httplib's `set_decompress()` also inflates deflate/br. So when LBP3 asks for `Accept-Encoding: deflate`, we inflate it behind its back.

Since the flag is gzip-only, we can reuse your Accept-Encoding scan here too:
```cpp
cli.set_decompress(plan.settings.inflate_gzip &&
                   (!game_set_accept_encoding || game_asked_for_gzip));
```
where `game_asked_for_gzip` is a case-insensitive search for "gzip" in that header's value.

**User-Agent:** you were right. I looked at the [server software code](https://github.com/LittleBigRefresh/Refresh/blob/main/Refresh.Common/Extensions/RequestContextExtensions.cs): the 403 comes from Refresh's `IsPatchworkVersionValid`, which does `StartsWith("PatchworkLBP")` and then throws the rest of the string at a `Version.TryParse` method. So nothing was wrong on the shadps4 side; the UA I was sending is a PS3 Patchwork spoof from my own game patch, and it was my spoof that needed to override the user agent entirely.

georgemoralis:
@sgdc3 it is more complex that than . I forced a manual inflate that seems more correct , if you can test

src/core/libraries/network/http.cpp
georgemoralis: yah it was pointless to do it like this , removed

sgdc3:
@georgemoralis works as expected, thanks

PR #4942, opened 2026-08-31
BugFix: LOG_GENERIC evaluated arguments before checking level

AnishGRao:
Was doing some debugging with:
`strace -f -c -p $shadps4_pid`

And saw nearly 450000 calls to `prctl`. Went a step further, and found that when tracing specifically for `prctl`, it's all to `PR_GET_NAME` which felt insane, because why are we checking for who we are so often? Going through the codebase I found that pthread_getname was in GetCurrentThreadName, which led me to the LOG_GENERIC call. 

Adding this check here significantly reduced the prctl spam.


StevenMiller123 (approved):
Looks good to me.

PR #4952, opened 2026-09-01
Fix incorrect error return on attempting to create a file in a nonexistent folder in a writable mount

kalaposfos13:
Tearaway Unfolded creates files in its savedata by first trying to create the file, and if that fails to get created due to the parent folder(s) not existing, it creates the folder(s) and tries again. After the Abstract Filesystem PR, this case was failed to be considered and `open` incorrectly returned EROFS in errno instead of EIO, causing the game to think the savedata isn't writable and abort. This PR fixes the errno return value in this case, which fixes savedata handling in the aforementioned game.


StevenMiller123:
Clang failed :smile:

Linear524:
Huge thanks !!!
This bug was pretty annoying for a long time...

Linear524:
Sorry - issue still persist :(

PR #4955, opened 2026-09-02
video decoding: prevent write to an uninitialized heap

roamic:


georgemoralis:
we do actually

hgh32:
sekiro hit assertion
[shad_log.txt](https://github.com/user-attachments/files/31814478/shad_log.txt)

Issue #4925, opened 2026-08-26
[Feature Request]: Issues found PVS studio static analyzer: np_common.cpp


Seraphimt:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

Analyzer message:
V547 Expression `compare < 0` is always false. Unsigned type value is never < 0. np_common.cpp 49

The typo is obvious.
```c++
s32 PS4_SYSV_ABI sceNpCmpNpIdInOrder(OrbisNpId* np_id1, OrbisNpId* np_id2, 
u32* out_result) {
  ....
  // Compare data
  u32 compare =
    std::strncmp(np_id1->handle.data, np_id2->handle.data, 
    ORBIS_NP_ONLINEID_MAX_LENGTH);
  if (compare < 0) {
    *out_result = -1;
      return ORBIS_OK;
  } else if (compare > 0) {
    *out_result = 1;
    return ORBIS_OK;
  }
  ....
}
```
Link to an article with additional information: https://pvs-studio.com/en/blog/posts/cpp/1405/

### Reason

Fixing potential bug and\or code quality.

Issue #4924, opened 2026-08-26
[Feature Request]: Issues found PVS studio static analyzer: hull_shader_transform.cpp


Seraphimt:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

Analyzer message:
V796 It is possible that `break` statement is missing in switch statement. hull_shader_transform.cpp 247

A highly probable error or very strange code that requires a comment\[[fallthrough]].
```c++
void WalkUsersOfTessConstantHelper(IR::Use use, u32 inc, bool propagateError) {
  IR::Inst* inst = use.user;

  switch (use.user->GetOpcode()) {
  case IR::Opcode::LoadSharedU32:
  case IR::Opcode::LoadSharedU64:
  case IR::Opcode::WriteSharedU32:
  case IR::Opcode::WriteSharedU64: {
    bool is_addr_operand = use.operand == 0;
    if (is_addr_operand) {
      u32 counter = inst->Flags<u32>();
      inst->SetFlags<u32>(counter + inc);
      ASSERT_MSG(!propagateError, 
          "LDS instruction {} accesses ambiguous attribute type",
           fmt::ptr(use.user));
      // Stop here
      return;
    }
  }
  case IR::Opcode::Phi: {
    auto it = phi_infos.find(use.user);
    ....
  }
  ....
}
```
Link to an article with additional information: https://pvs-studio.com/en/blog/posts/cpp/1405/ 

### Reason

Fixing potential bug and\or code quality.

Issue #4923, opened 2026-08-26
[Feature Request]: Issues found PVS studio static analyzer: decoder.cpp


Seraphimt:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

Analyzer message:
V579 The ZydisDecoderDecodeFull function receives the pointer and its size as arguments. It is possibly a error. Inspect the third argument. decoder.cpp 29

Instruction longer than 8 byte will be cut off. A highly probable error or very strange code that requires a comment\simplification.
```c++
void DecoderImpl::printInstruction(void* code, u64 address) {
  ZydisDecodedInstruction instruction;
  ZydisDecodedOperand operands[ZYDIS_MAX_OPERAND_COUNT_VISIBLE];
  ZyanStatus status =
    ZydisDecoderDecodeFull(&m_decoder, code, sizeof(code), 
                           &instruction, operands);
  if (!ZYAN_SUCCESS(status)) {
    fmt::print("decode instruction failed at {}\n", fmt::ptr(code));
  } else {
    printInst(instruction, operands, address);
  }
}
```
Link to an article with additional information: https://pvs-studio.com/en/blog/posts/cpp/1405/ 

### Reason

Fixing potential bug and\or code quality.

Issue #4922, opened 2026-08-26
[Feature Request]: Issues found PVS studio static analyzer: liverpool_to_vk.cpp


Seraphimt:
### Checklist

- [x] I have searched for a similar issue in this repository and did not find one.

### Description

Analyzer message:
V634 The priority of the `*` operation is higher than that of the `<<` operation. It`s possible that parentheses should be used in the expression. liverpool_to_vk.cpp 759

Suspicious expression and it's easy to break something due to operator precedence. Requires a comment\simplification.
```c++
// Table 8.13 Data and Image Formats [Sea Islands Series Instruction Set Architecture]
static const size_t amd_gpu_data_format_bit_size = 6;   // All values are under 64
static const size_t amd_gpu_number_format_bit_size = 4; // All values are under 16

static auto surface_format_table = []() constexpr {
  std::array<vk::Format, 1 << amd_gpu_data_format_bit_size * 1 << amd_gpu_number_format_bit_size>
    result;
  for (auto& entry : result) {
    entry = vk::Format::eUndefined;
  }
  for (const auto& supported_format : SurfaceFormats()) {
    result[GetSurfaceFormatTableIndex(supported_format.data_format,
                                      supported_format.number_format)] =
        supported_format.vk_format;
  }
  return result;
}();
```
Link to an article with additional information: https://pvs-studio.com/en/blog/posts/cpp/1405/

### Reason

Fixing potential bug and\or code quality.

PR #4959, opened 2026-09-03
Random, mostly pointless fixes

StevenMiller123:
- Remove dead code in DecoderImpl #4923
- Fix weird harmless double-assign in TextEditor #4921
- Fix FindStreams return check #4917
- Fix broken memset and clean up a bit of Aio code #4918
- Fix type for compare check in sceNpCmpNpIdInOrder #4925
- Fix scePlayGoConvertLanguage types #4920
- Fix scePlayGoGetProgress nullptr dereference #4919
- Fix (I think?) missing break in HullShaderTransform #4924
- Fix (I think?) vk::Format result table definition #4922

Also changes unrelated to the main goal of this PR:
- Tidy up comments in sceNpCheckPlus
- Fix OrbisNpState defines based on extracted VSH defines

Needs review, I think most of this is fine, but I'd like a second opinion (especially on the ones I'm not too confident about).

For future reference, @Seraphimt (or really anyone thinking to do something like this in the future), just group these into one issue. There is no value in making so many issues each to document one-line fixes.


tochnonechelovek:
No regressions in my titles, if it matters.

src/shader_recompiler/ir/passes/hull_shader_transform.cpp
```diff
@@ -244,6 +244,7 @@ class TessConstantUseWalker {
                 // Stop here
                 return;
             }
+            break;
```
georgemoralis, line 247: i have doubts about that break , @raphaelthegreat ?

src/video_core/renderer_vulkan/liverpool_to_vk.cpp
```diff
@@ -759,7 +759,8 @@ static size_t GetSurfaceFormatTableIndex(AmdGpu::DataFormat data_format,
 }
 
 static auto surface_format_table = []() constexpr {
-    std::array<vk::Format, 1 << amd_gpu_data_format_bit_size * 1 << amd_gpu_number_format_bit_size>
+    std::array<vk::Format,
```
georgemoralis, line 765: and for this

src/shader_recompiler/ir/passes/hull_shader_transform.cpp
mikusp: looks correct

src/video_core/renderer_vulkan/liverpool_to_vk.cpp
mikusp: also correct

PR #4960, opened 2026-09-03
Upstreaming imgui 

georgemoralis:
We use a quite old imgui from 2024 , upstreamed it and added the neccesary fixes to latest version.

@rainmakerv3 if you can take a look since most of changes are on big picture


rainmakerv3:
It looks like for big picture, the change was basically just adding ImGuiChildFlags_Borders, which could affect spacing, though probably not in any meaningful way. If big picture / settings window still look okay, then it should be fine

PR #4967, opened 2026-09-04
fix imgui crash

georgemoralis:


ConnorG512:
Yep, with this PR I can hit F3 normally again 
image

PR #4968, opened 2026-09-04
Fix regressions from #4952

kalaposfos13:
Files outside of mount points for some reason get far into opening that they hit the error path I modified, instead of just erroring out much sooner, so here's a fix for all those games that try opening debug files in the current directory, which is not allowed on the PS4, but we need to handle the errors coming from that regardless.


bigol83:
Fixes No Man's Sky crash
immagine

PR #4970, opened 2026-09-05
Improve log message for fp rounding modes

mikusp:


georgemoralis:
clang is not your friend

PR #4975, opened 2026-09-06
Fix cut off logging in log files after the signals rewrite

kalaposfos13:
This also fixes a small and inconsequential bug about guest breakpoint handling and cleans up some dead code.


StevenMiller123 (approved):
Tested and confirmed working on Windows and Linux.

PR #4977, opened 2026-09-07
renderer_vulkan: Handle QuadStrip as a triangle strip

junminlee2004:
This PR fixes the following crash when attempting to aim a gun in The last of us Remastered and Uncharted 3 by adding the missing `QuadStrip` case to `PrimitiveType()`:
```
[Debug] <Critical> (shadPS4:GpuComm) liverpool_to_vk.cpp:138 PrimitiveType: Unreachable code!
Unimplemented primitive type: 20
```
 Since a triangle strip over the same vertices covers the same area at even vertex count, differing only in which diagonal each quad is split along, the existing topology can simply be reused directly.

Uncharted
Guns can now be aimed and fired in Uncharted 3

Before
[Screencast_20260906_203304.webm](https://github.com/user-attachments/assets/601e0bcc-5040-49d3-8c68-f3f4ce4e72a8)

After
https://github.com/user-attachments/assets/7a28d4d7-76e2-4e43-ad8e-71b16a755950

### DIFF
--- a/src/video_core/renderer_vulkan/liverpool_to_vk.cpp
+++ b/src/video_core/renderer_vulkan/liverpool_to_vk.cpp
@@ -121,6 +121,8 @@ vk::PrimitiveTopology PrimitiveType(AmdGpu::PrimitiveType type) {
         return vk::PrimitiveTopology::eTriangleFan;
     case AmdGpu::PrimitiveType::TriangleStrip:
         return vk::PrimitiveTopology::eTriangleStrip;
+    case AmdGpu::PrimitiveType::QuadStrip:
+        return vk::PrimitiveTopology::eTriangleStrip;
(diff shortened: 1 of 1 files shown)


diogoaduartereis:
Just tried on my end and is working as intended, no more crashes.

raphaelthegreat:
Are you sure QuadStrip and TriangleStrip are that compatible? I believe there must be some difference between them

red-prig:
If you don't do anything unusual with the barycentric coordinates, there won't be any differences.

red-prig:
but normally this should be emulated in the same way as the rect list in the tassellation shader

raphaelthegreat:
tessellation runs before geometry so this is not so easy. But I guess you should add warning log here and it should be fine

PR #4993, opened 2026-09-09
Stub sceNpNotifyPlusFeature

mikusp:
Spammed by NBA 2K22


metr1k-91:
I'll see if I can get this proper at some point soon.

mikusp:
@metr1k-91 I've decided to leave it as a stub as it seems inconsequential to the emulation itself, probably it's more about compliance and the certification process

PR #4994, opened 2026-09-09
imgui: stop Triangle from opening context menus

junminlee2004:
This PR fixes a regression caused by PR #4960 that created a binding for `ImGuiKey_NavGamepadContextMenu` to the triangle button, giving the button the ability to open context menus. The fix is to mark triangle as owned by the emulator to remove this ability since imgui nav's context menu request only accepts unowned keys


junminlee2004:
@gandalfthewhite19890404 this should fix triangle triggering the font size dialog when fps counter is enabled.

gandalfthewhite19890404:
Yes, works awesome now

PR #4995, opened 2026-09-10
shader_recompiler: Lower wave64 constructs for wave32

raphaelthegreat:
Credits to @GhostRider584 for initially spotting this issue and deriving a prototype solution. 

After inverse ballot elimination runs, there might be surviving Ballot opcodes. There are cases where the full mask is required for properly performing an operation. An example could be an append buffer

```
u32 base = 0;
uint64_t mask = subgroupBallot(cond);
if (thread_id == 0) {
    base = atomicAdd(..., bitCount(mask));
}
base = subgroupBroadcast(base, 0);
```

to properly execute this, both mask and subgroupBroadcast have to use shared memory to share results between pairs of wave32 subgroups. Sharing uses aliasing for efficiency; only 1 store and 1 load are required.

```
uint64_t mask_wave32 = ...
uint offset = scratch_base + ((gl_LocalInvocationIndex / 32) << 2u);
shared_mem_u32[offset >> 2u] = unpackUint2x32(mask_wave32).x;
barrier();
uint64_t mask_wave64 = shared_mem_u64[offset >> 3u];
```

Using subgroupElect would be even more ideal to limit shared mem access to 1 thread but its not simple to introduce branching ops in IR level and the goal is for the lowering to be transparent to the backend.

Another case handled here is LDS spills, namely when variables get stored to shared memory and loaded back again from same location. This has been handled for fragment stage for a long while as its simple; no graphics API allows shared memory in non compute stages so any shared access in fragment must be a compiler generated pattern. But for compute a detection pass is required to promote it back to SSA.

The code is incomplete, requires cleanup and a few more changes and handling more stuff. Also still not sure if LDS spilling should be a pattern matched replacement or if the wave64 lowering should also handle that aspect.


Parotaku:
This PR gives FIST's 3D some blue artifact on title menu

image

[shad_log.zip](https://github.com/user-attachments/files/32046536/shad_log.zip)

Parotaku:
It also makes Resident Evil 4 crash before showing the menu 3D (the glitchy forest)...

[shad_log.zip](https://github.com/user-attachments/files/32047391/shad_log.zip)

Graubek:
There is a warning in logs about `[Render.Recompiler] <Warning> (shadPS4:GpuCommandProcessor) lower_wave64_pass.cpp:102 LowerWave64BallotPass: Ballot instruction in non uniform control flow` when playing infamous second son
[shad_log.txt](https://github.com/user-attachments/files/32050062/shad_log.txt)

carrrrrrrrot:
Rendering in sotc is much improved.
castle_main
castle_turtle
interior_main
interior_turtle
[shad_log.txt](https://github.com/user-attachments/files/32068420/shad_log.txt)

Parotaku:
Blue artifact is fixed...
image

Parotaku:
image

Broken forest 3d is back...

bigol83:
Lines are still visiable for me in FF VII Remake

immagine

raphaelthegreat:
this is done for now as an initial implementation

PR #4996, opened 2026-09-10
shader recompiler multiple fixes, common, video core and buffer cashe

nikmparis217-ux:
common: Destroy pooled objects when a chunk is released. ObjectPool destroyed the storage union, never the objects, so every non-trivial member leaked 3,800 small blocks per shader, 15 million live allocations.

video_core: Free the coroutine frame of a finished indirect buffer.Three IndirectBuffer handlers drove the task to completion and dropped. final_suspend suspends and Task has no destructor, so one frame leaked per packet. Measured: host heap 3991 → 1283 MB.

buffer_cache: Barrier a buffer whose last access was a write. An identical access mask skipped the barrier even when the previous access had been a write.

shader_recompiler: Translate the remaining 64-bit compares and F64 ops. Only 4 of the 32 opcodes reached the function; the other 28 fell through to the unknown-opcode path.

shader_recompiler: Emit DS_SWIZZLE_B32 as a subgroup shuffle.Both swizzle modes compute the source lane from the lane id, but the code used Broadcast/QuadBroadcast, which require a dynamically uniform index undefined.

shader_recompiler: Guard loop exits that can wrap. while (i != end) with a zero or inverted count walks the counter through the whole 32-bit range. Also caps an exit bound loaded from memory.

### DIFF
--- a/src/common/object_pool.h
+++ b/src/common/object_pool.h
@@ -77,7 +77,9 @@ private:
         }
 
         void Release() {
-            std::destroy_n(storage.get(), used_objects);
+            for (size_t i = used_objects; i > 0; --i) {
+                std::destroy_at(std::addressof(storage[i - 1].object));
+            }
             used_objects = 0;
         }
 
--- a/src/video_core/amdgpu/liverpool.cpp
+++ b/src/video_core/amdgpu/liverpool.cpp
@@ -202,6 +202,7 @@ Liverpool::Task Liverpool::ProcessCeUpdate(std::span<const u32> ccb) {
                 YIELD_CE();
                 RESUME_CE(task);
             }
+            task.handle.destroy();
             break;
         }
         default:
@@ -837,6 +838,7 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
                     YIELD_GFX();
                     RESUME_GFX(task);
                 }
+                task.handle.destroy();
                 break;
             }
             case PM4ItOpcode::IncrementDeCounter: {
@@ -975,6 +977,7 @@ Liverpool::Task Liverpool::ProcessCompute(std::span<const u32> acb, u32 vqid) {
                 YIELD_ASC(vqid);
                 RESUME_ASC(task, vqid);
             }
+            task.handle.destroy();
             break;
         }
         case PM4ItOpcode::DmaData: {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1030,6 +1030,7 @@ set(SHADER_RECOMPILER src/shader_recompiler/profile.h
                       src/shader_recompiler/ir/passes/lower_fp64_to_fp32.cpp
                       src/shader_recompiler/ir/passes/lower_user_clip_planes.cpp
                       src/shader_recompiler/ir/passes/phi_simplification_pass.cpp
+                      src/shader_recompiler/ir/passes/loop_wrap_guard_pass.cpp
                       src/shader_recompiler/ir/passes/readlane_elimination_pass.cpp
                       src/shader_recompiler/ir/passes/resource_discover_pass.cpp
                       src/shader_recompiler/ir/passes/resource_pass.h
--- a/src/shader_recompiler/backend/spirv/emit_spirv.cpp
+++ b/src/shader_recompiler/backend/spirv/emit_spirv.cpp
@@ -300,6 +300,9 @@ void SetupCapabilities(const Info& info, const Profile& profile, const RuntimeIn
     if (info.uses_group_quad) {
         ctx.AddCapability(spv::Capability::GroupNonUniformQuad);
     }
+    if (info.uses_group_shuffle) {
+        ctx.AddCapability(spv::Capability::GroupNonUniformShuffle);
+    }
     if (info.uses_group_ballot || info.loads.Get(IR::Attribute::SubgroupLtMask)) {
         ctx.AddCapability(spv::Capability::GroupNonUniformBallot);
     }
--- a/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
+++ b/src/shader_recompiler/backend/spirv/emit_spirv_instructions.h
@@ -476,6 +476,7 @@ Id EmitCubeFaceIndex(EmitContext& ctx, IR::Inst* inst, Id cube_coords);
 Id EmitLaneId(EmitContext& ctx);
 Id EmitWarpId(EmitContext& ctx);
 Id EmitQuadShuffle(EmitContext& ctx, Id value, Id index);
+Id EmitShuffle(EmitContext& ctx, Id value, Id index);
 Id EmitReadFirstLane(EmitContext& ctx, Id value);
 Id EmitReadLane(EmitContext& ctx, Id value, Id lane);
 Id EmitWriteLane(EmitContext& ctx, Id value, Id write_value, u32 lane);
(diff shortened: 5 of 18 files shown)


nikmparis217-ux:
i tested these changes on a personal fixed gt7. no personal fixes are added to this request only general emulator fixes

raphaelthegreat:
Can you give more details on the loop issue, wouldn't the guest shader also not work on real hw if that is the case

nikmparis217-ux:
> Can you give more details on the loop issue, wouldn't the guest shader also not work on real hw if that is the case

I will provide more details once i get home from work

src/common/object_pool.h
```diff
@@ -77,7 +77,9 @@ class ObjectPool {
         }
 
         void Release() {
-            std::destroy_n(storage.get(), used_objects);
+            for (size_t i = used_objects; i > 0; --i) {
```
StevenMiller123, line 80: `std::destroy_n`should be the same as looping through `std::destroy_at`

StevenMiller123 (changes requested):
Not reviewing the rest since most of this is code I'm not too familiar with.

src/common/object_pool.h
StevenMiller123: Nevermind, I've been told there is actually a difference.

Graubek:
All regressions:
**Concrete Genie** crashes before reaching main menu `[Debug] <Critical> (shadPS4:GpuCommandProcessor) signals.cpp:143 SignalHandler: Unhandled Exception code 0xc0000005 at 0x7000010af140` [concrete.txt](https://github.com/user-attachments/files/32061800/concrete.txt)
**Gravity Rush 2** crashes before displaying splashscreen `[Debug] <Critical> (shadPS4:GpuCommandProcessor) signals.cpp:143 SignalHandler: Unhandled Exception code 0xc0000005 at 0x7000010af140` [Gravity2.txt](https://github.com/user-attachments/files/32061828/Gravity2.txt)
**Gravity Rush Remastered** crashes before going ingame `[Debug] <Critical> (shadPS4:GpuCommandProcessor) signals.cpp:143 SignalHandler: Unhandled Exception code 0xc0000005 at 0x7000010af140`  [Gravity.txt](https://github.com/user-attachments/files/32061909/Gravity.txt)
**Infamous Second Son** crashes before loading the game `[Debug] <Critical> (shadPS4:GpuCommandProcessor) signals.cpp:143 SignalHandler: Unhandled Exception code 0xc0000005 at 0x7000010af140`  [second.txt](https://github.com/user-attachments/files/32061986/second.txt)

Ok, I just noticed that nearly all my games are crashing on the same error

src/video_core/amdgpu/liverpool.cpp
```diff
@@ -202,6 +202,7 @@ Liverpool::Task Liverpool::ProcessCeUpdate(std::span<const u32> ccb) {
                 YIELD_CE();
                 RESUME_CE(task);
             }
+            task.handle.destroy();
```
Niram7777, line 205: I have a feeling all of them should be in the task destructor ?
(I had a draft I never pushed to rewrite the task as a c++ object with ctor/dtor)

PS: you would be better submitting several PRs

baggins183:
I am skeptical of this loop thing, it seems like you are bounding the max number of iterations? If the loop bound is a garbage value from memory or something then this is hiding other issues

nikmparis217-ux:
> Can you give more details on the loop issue, wouldn't the guest shader also not work on real hw if that is the case

You're right, the guest shader isn't buggy.The loop this was written for is a tessellation control-point do-while whose counter decrements by one and whose exit test is counter != 1. On real hardware the count is always >= 1, so != selects exactly the same iteration count an ordered compare would. It only wraps under emulation, because the count arrives as 0: the flattened user-data snapshot gets read before the page backing it is registered, so the shader is handed a value the console would never produce. The second case, which the bound cap was added for, is the same story — an ordered loop whose bound comes out of srt_flatbuf[16], where stale bytes turn (garbage + 63)  6 into tens of millions of iterations.
So both cases are emulator-side data bugs in the SRT/flatbuf path. The pass doesn't fix a guest bug; it stops a bad value we produced from becoming a device loss.

src/video_core/amdgpu/liverpool.cpp
nikmparis217-ux: > I have a feeling all of them should be in the task destructor ? (I had a draft I never pushed to rewrite the task as a c++ object with ctor/dtor)
> 
> PS: you would be better submitting several PRs

I checked how Task is actually used before answering: all four are function-local (auto task = ...), none is copied, and the only assignment is ce_task = ProcessCeUpdate(ccb). So an owning, move only Task destructor calling handle.destroy() when the handle is non-null, move ctor and move-assign via std::exchange, copy deleted drops in without touching a single call site, and all four explicit destroy() calls go away. It also covers a case the manual calls can't. Today a frame is only freed on the path that runs the while (!done) loop to completion. If the enclosing coroutine is itself destroyed while suspended inside that loop shutdown, or any early exit added later the inner frame leaks, because nothing owns it. With the handle owned by the object, destroying the parent frame destroys its locals and the inner coroutine goes with it. One implementation detail worth being explicit about is promise_type::final_suspend() returns suspend_always, so a completed coroutine is suspended rather than destroyed and the destructor still has to free it. destroy() on a suspended-but-not-done handle is equally well-defined, which is what makes the abandoned case above work. 

note taken about the seperate PRs.

nikmparis217-ux:
> All regressions: **Concrete Genie** crashes before reaching main menu `[Debug] <Critical> (shadPS4:GpuCommandProcessor) signals.cpp:143 SignalHandler: Unhandled Exception code 0xc0000005 at 0x7000010af140` [concrete.txt](https://github.com/user-attachments/files/32061800/concrete.txt) **Gravity Rush 2** crashes before displaying splashscreen `[Debug] <Critical> (shadPS4:GpuCommandProcessor) signals.cpp:143 SignalHandler: Unhandled Exception code 0xc0000005 at 0x7000010af140` [Gravity2.txt](https://github.com/user-attachments/files/32061828/Gravity2.txt) **Gravity Rush Remastered** crashes before going ingame `[Debug] <Critical> (shadPS4:GpuCommandProcessor) signals.cpp:143 SignalHandler: Unhandled Exception code 0xc0000005 at 0x7000010af140` [Gravity.txt](https://github.com/user-attachments/files/32061909/Gravity.txt) **Infamous Second Son** crashes before loading the game `[Debug] <Critical> (shadPS4:GpuCommandProcessor) signals.cpp:143 SignalHandler: Unhandled Exception code 0xc0000005 at 0x7000010af140` [second.txt](https://github.com/user-attachments/files/32061986/second.txt)
> 
> Ok, I just noticed that nearly all my games are crashing on the same error


Thanks for the logs. They're one failure, not four. The last line before the crash in every one of them is vk_pipeline_cache.cpp:615 CompileModule: Compiling <stage> shader 0x... so the fault is inside the shader recompiler, and it lands on three different stages across the four titles: vs in Concrete Genie, cs in Gravity Rush 2 and Infamous Second Son, fs in Gravity Rush Remastered. That rules out the two commits here that only run on shaders using a particular instruction the 64-bit compares and DS_SWIZZLE_B32  since neither can fire on all of those. Of what's left, LoopWrapGuardPass is the one that is new, runs unconditionally on every shader, and has only ever been exercised against a single title in my own testing. The others (object pool, buffer barrier, coroutine frame) run unconditionally in my tree and have been through every session of two games without producing this .I haven't identified the faulting instruction inside the pass yet, so that's elimination rather than a diagnosis, and I'd rather not change the PR on a guess. There's also the correctness point from the other thread: that commit's bound cap rewrites any memory-loaded loop bound to min(bound, 16384), which silently truncates correct shaders.How would you like it handled? drop that commit from this PR?
One thing I should ask before going further. has anyone confirmed these are regressions  that the same builds of these four games get past this point on current main? All four logs are from this PR's build, and I don't want to chase a commit here if the same crash reproduces without it. CompileModule also runs the JIT'd SRT walker, which reads guest memory and can fault, and this PR doesn't touch that path at all so a main log from any one of the four would tell us a lot.

PR #4999, opened 2026-09-10
shader_recompiler: Bind the masked value of a descriptor dword

nikmparis217-ux:
Found this problem on god of war 2018. Logging the opcode the walk gave up on named BitwiseAnd32, 291 times across 273 shaders in a single frame, which was also the number of programs that ended up with no usable resource list. The mask, 0xc3ffffff, clears bits 26-29 the top of the stride field of V# word 1. PeelDwordMasks records the cleared bits per dword and continues the walk to the value underneath, which is a sharp source again; SharpFetch reapplies them when it assembles the descriptor.  this fix was developed with AI (Claude). I reviewed the change and wrote this description and checked the code. 

before Στιγμιότυπο οθόνης 2026-09-10 223226
after Στιγμιότυπο οθόνης 2026-09-10 224535

### DIFF
--- a/src/shader_recompiler/ir/passes/resource_discover_pass.cpp
+++ b/src/shader_recompiler/ir/passes/resource_discover_pass.cpp
@@ -246,6 +246,27 @@ IR::Inst* FindSharpSource(IR::Inst* handle) {
     return handle;
 }
 
+static void PeelDwordMasks(SharpReference& sharp) {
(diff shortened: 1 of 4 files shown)


baggins183:
I was looking into the same thing a few days ago, I believe this is is only being done to T#'s so this would be the num_format field and not V# stride.

I think this approach would work although it doesn't really match other cases that use SharpFetchPostOp. But I would rather handle this in srt program itself

PR #5010, opened 2026-09-12
shader_recompiler: Cache fetch shader parsing

Kyoskii:
Potential optimization: caches the fetch shader parse so it runs once per shader instead of once per draw, cutting pipeline lookup cost + more cpu headroom in scenes with heavy draw calls.

The parse only depends on the fetch shader's instruction bytes, so the result is safe to reuse. Entries are keyed by code address with a hash of the parsed bytes as a guard, so code placed at a reused address can't hit a stale entry. The map is thread_local, which keeps the lookup lock-free without assuming it only ever has one caller. 

Scenes with a larger amount of draw calls should see more of a benefit, from few user tests this also gives more CPU headroom.

### DIFF
--- a/src/shader_recompiler/frontend/fetch_shader.cpp
+++ b/src/shader_recompiler/frontend/fetch_shader.cpp
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include <unordered_map>
 #include "common/assert.h"
 #include "shader_recompiler/frontend/decode.h"
 #include "shader_recompiler/frontend/fetch_shader.h"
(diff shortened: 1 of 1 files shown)


src/shader_recompiler/frontend/fetch_shader.cpp
```diff
@@ -45,12 +46,42 @@ const u32* GetFetchShaderCode(const Info& info, u32 sgpr_base) {
     return code;
 }
 
+// FNV-1a over the shader's dwords. Kept local rather than pulling in an external hash, because
```
Niram7777, line 49: maybe a SIMD algo like xxHash would be faster

StevenMiller123 (changes requested):
Right now this doesn't meet our AI rules @Kyoskii 
Please properly disclose AI use, and do some level of proofreading instead of leaving AI generated comments intact. I'll leave this open since it's been tested with decent results.

src/shader_recompiler/frontend/fetch_shader.cpp
```diff
@@ -45,12 +46,42 @@ const u32* GetFetchShaderCode(const Info& info, u32 sgpr_base) {
...
+    // per shader stage per draw, and every build disassembles the whole shader again. The parse
+    // only depends on the instruction bytes, so cache it and skip the repeats.
+    //
+    // Keyed by code address, with a hash of the parsed bytes so a reused address can't return a
+    // stale entry. thread_local keeps the lookup lock-free without assuming a single caller.
+    struct CacheEntry {
+        u64 code_hash{};
+        FetchShaderData data{};
+    };
+    thread_local std::unordered_map<const u32*, CacheEntry> parse_cache;
```
Niram7777, line 76: if playing for a long time (and there are lot of shaders) this could grow very big ?
maybe use fixed length container 🤷

src/shader_recompiler/frontend/fetch_shader.cpp
```diff
@@ -108,6 +139,7 @@ std::optional<FetchShaderData> ParseFetchShader(const Shader::Info& info) {
         }
     }
 
+    parse_cache.insert_or_assign(code, CacheEntry{HashShaderCode(code, data.size), data});
```
Niram7777, line 142: add config option in case it consumes too much memory ?

Niram7777 (commented):
nice, I was looking at the perf issue in Wipeout, I will test your PR tomorrow

Image

Niram7777:
it looks like a couple of fps faster, but going from 10 to 12 is not like a game changer.
shadps4_pr
I still have the +3000 draw indexed and pipeline graphics, my bottleneck is somewhere else.
(I have to investigate BindResources, BeginRendering, ForEachSharp, RuntimeInfo comparator)

src/shader_recompiler/frontend/fetch_shader.cpp
Niram7777: but your algo might be good enough for starter

src/shader_recompiler/frontend/fetch_shader.cpp
Niram7777: in Wipeout, I reach a 1000 entries after a finishing a couple of races.
For simplicity, as a starter you could just do a
```cpp
if parse_cache.size > 10'000
    parse_cache.clear()
```

Kyoskii:
> it looks like a couple of fps faster, but going from 10 to 12 is not like a game changer.
 I still have the +3000 draw indexed and pipeline graphics, my bottleneck is somewhere else. (I have to investigate BindResources, BeginRendering, ForEachSharp, RuntimeInfo comparator)

yeah i appreciate the feedback, I'll look into improving it and maybe adding a LRU cache+limit, mostly just trying to find ways to improve the cpu bottleneck in high draw call scenes but not quite sure how we could further improve the bindbuffers/bindtextures besides maybe like, an indexed memory so we can avoid constantly walking pages.

realistically the better/appropriate solution would be to properly implement set_predication and occlusion so we can cut down on those unnecessary draws overall

PR #5009, opened 2026-09-12
Implement some V_CMPX_U64 ops

StevenMiller123:
Just the two I know are needed right now, can add more if others have been seen in games.

@Kravickas I believe Marvel's Avengers is affected by this, could you test?

PR #5011, opened 2026-09-12
shader_recompiler: Implement V_ASHR_I64

Gih-pt:
Implements the missing vector ALU opcode V_ASHR_I64, reported by @Kravickas in comments on PR #5009 and aggregate issue #496 (crash in Marvel's Avengers, CUSA14030).

Note: AI was used for code generation and investigation; changes were
validated against the existing merged 64-bit shift implementations
and the instruction metadata table in format.cpp. 

I also compiled the build locally (to save @raphaelthegreat from having to do extra work).

### DIFF
--- a/src/shader_recompiler/frontend/translate/translate.h
+++ b/src/shader_recompiler/frontend/translate/translate.h
@@ -275,6 +275,7 @@ public:
     void V_CVT_PK_U8_F32(const GcnInst& inst);
     void V_LSHL_B64(const GcnInst& inst);
     void V_LSHR_B64(const GcnInst& inst);
+    void V_ASHR_I64(const GcnInst& inst);
     void V_ALIGNBIT_B32(const GcnInst& inst);
     void V_ALIGNBYTE_B32(const GcnInst& inst);
     void V_MUL_F64(const GcnInst& inst);
--- a/src/shader_recompiler/frontend/translate/vector_alu.cpp
+++ b/src/shader_recompiler/frontend/translate/vector_alu.cpp
@@ -447,6 +447,8 @@ void Translator::EmitVectorAlu(const GcnInst& inst) {
         return V_LSHL_B64(inst);
     case Opcode::V_LSHR_B64:
         return V_LSHR_B64(inst);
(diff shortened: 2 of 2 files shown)


Kravickas:
Works, new crash on
```

[Debug] <Critical> (shadPS4:GpuCommandProcessor) inverse_ballot_elimination_pass.cpp:90 InverseBallotEliminationPass: Unreachable code!
Unexpected immediate argument for InverseBallot 0x1
```
[shad_log.txt](https://github.com/user-attachments/files/32149981/shad_log.txt)

Gih-pt:
A question for the devs: is it better to update a PR by merging the main branch or by rebasing?

raphaelthegreat (approved):

raphaelthegreat:
> Works, new crash on
> 
> ```
> 
> [Debug] <Critical> (shadPS4:GpuCommandProcessor) inverse_ballot_elimination_pass.cpp:90 InverseBallotEliminationPass: Unreachable code!
> Unexpected immediate argument for InverseBallot 0x1
> ```
> 
> [shad_log.txt](https://github.com/user-attachments/files/32149981/shad_log.txt)

This could be implemented by folding the InverseBallot to either
```
gl_SubgroupInvocationID == 0
```
or
```
(gl_LocalInvocationIndex & 63) == 0
```
depending if the operation is wave64 sensitive or not. the operation needs to be examined to determine that

Gih-pt:
@raphaelthegreat 
Opened #5021 for one of the two InverseBallot call sites that can hit this — SetScc, which should just be 'result != 0' like the 32-bit siblings already do. Not confirmed yet whether that's the one actually hit by this specific crash.

The other call site is SetDst64's ExecLo case in translate.cpp:1021. Didn't want to guess at that one blind since it sounds like an actual design decision — profile.subgroup_size is populated but never read anywhere in shader_recompiler, and IsSubgroupSize64Supported() will always be false on NVIDIA's fixed 32-wide warp architecture, so gl_SubgroupInvocationID alone won't be reliable there (matches what you flagged between that and gl_LocalInvocationIndex & 63).

PR #5014, opened 2026-09-12
Harden GS copy shader parsing and fix GPU buffer synchronization

Nyces:
Ok I'm submitting this PR again; I think it's in good enough shape now (?

In short, these changes fix the character facial vertex explosion issue in *The Idolmaster* and resolve crashes that occurred during story sequences and at the start of live performances.

For specific details, please refer to https://github.com/shadps4-emu/shadPS4/pull/4781

Additionally, a friend of mine, @RainKikyou, used these modifications to try and fix the vertex explosion issue in *GUNDAM VERSUS*, and the results look good.

However, during testing, we discovered that these changes prevent *Demon Slayer: Kimetsu no Yaiba* from launching; I haven't investigated this specific issue in depth.

These changes were generated with AI assistance and have been tested to ensure they work.

### DIFF
--- a/src/shader_recompiler/frontend/copy_shader.cpp
+++ b/src/shader_recompiler/frontend/copy_shader.cpp
@@ -1,36 +1,111 @@
 // SPDX-FileCopyrightText: Copyright 2024 shadPS4 Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+#include <array>
+#include <optional>
+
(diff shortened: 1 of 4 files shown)


raphaelthegreat:
A lot of code here is probably not needed after changes to main branch. Did you hit any assertion that required implementing the expanded ReadConstant? I think that cleanup is good but other changes probably not mergeable

PR #5016, opened 2026-09-13
ngs2: propagate SystemSetup validation errors

Dev-next-gen:
In `SystemSetup`, in `src/core/libraries/ngs2/ngs2_impl.cpp`, the `result` variable is declared as `u32`, which directly contradicts the `ORBIS_NGS2_ERROR_*` constants returned by `SystemSetupCore`, all of which hold negative values. Because of that implicit conversion to an unsigned value, the two `if (result < 0)` guards at lines 133 and 163 can never mathematically be true, which turns those blocks into dead branches.

This logic flaw leads to a serious behaviour: `sceNgs2SystemQueryBufferSize` returns `ORBIS_OK` even for an invalid option, and fills `hostBufferInfo` anyway, while `sceNgs2SystemCreate` runs to completion and writes `*outHandle = 1` for a configuration it has just rejected internally.

The problem has stayed hidden until now because `sceNgs2SystemCreateWithAllocator` is the only execution path spared by the bug, simply because it re-evaluates the validation test on a local `s32` variable.

The fix is a one-line change, switching the variable from `u32` to `s32`. It has no public impact: `SystemSetup` is already declared as `s32` in the `ngs2_impl.h` header, and every caller already stores its return value in an `s32`.

Four new test cases were added to `tests/test_ngs2.cpp`, in the `shadps4_ngs2_test` target the CI already runs. Three of them fail on `main` and the fourth covers the healthy path, which confirms the fix holds.

Disclosure: Claude Opus 5 found the defect and wrote both the fix and the tests. I reviewed the change myself, checked the sign of the `ORBIS_NGS2_ERROR_*` constants and the callers in `ngs2.cpp`, and wrote this description.

### DIFF
--- a/src/core/libraries/ngs2/ngs2_impl.cpp
+++ b/src/core/libraries/ngs2/ngs2_impl.cpp
@@ -116,7 +116,7 @@ s32 SystemSetup(const OrbisNgs2SystemOption* option, OrbisNgs2ContextBufferInfo*
     SystemInternal setupResult;
     void* systemList = NULL;
     size_t requiredBufferSize = 0;
-    u32 result = ORBIS_NGS2_ERROR_INVALID_BUFFER_SIZE;
+    s32 result = ORBIS_NGS2_ERROR_INVALID_BUFFER_SIZE;
 
     if (option) {
         if (option->size != 64) {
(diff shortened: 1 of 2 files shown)


StevenMiller123:
> Disclosure: Claude Opus 5 found the defect and wrote both the fix and the tests. I reviewed the change myself, checked the sign of the ORBIS_NGS2_ERROR_* constants and the callers in ngs2.cpp, and wrote this description.

This whole PR description reads like it's been AI generated. Not sure if this comes from using a translator or from genuine intent to hide AI use, but I'm closing this PR.

Dev-next-gen:
Helle
I am french.
I wrote this in french and translated it.

StevenMiller123:
I'll reopen this. For future PRs please try to write descriptions yourself, or try to find a proper translator instead of using an LLM for translations.

Dev-next-gen:
Alright

PR #5020, opened 2026-09-13
video_core: Only enable sampler comparison for comparison fetches

junminlee2004:
This PR fixes the corrupt bloom in GR2 for Pascal and older Nvidia gpus by disabling vulkan depth comparison operations for plain (non _C) GCN instructions. GR2's bloom blur fragment shader `fs_0x0000000058bfbd5f_0` has no `IMAGE_SAMPLE_C` GCN instructions, only `IMAGE_SAMPLE`. Therefore, there should be no sampler comparison at all. The S#'s `depth_compare_func`, which in this case is Always, only decides which compareop is used. Whether the comparison should happen is decided by  the opcode. 
Currently,` .compareEnable = sampler.depth_compare_func != AmdGpu::DepthCompare::Never`, so vulkan depth comparison is enabled solely based on depth_compare_func, and textures get read by non-_C instructions through a comparison sampler.
The shader creates 48 samples per pixel for two textures, a 480×270` R11G11B10_FLOAT` (32 samples) and a 16×16 `R32_FLOAT` LUT (16 samples), through the same sampler. The 16x16 LUT holds the blur weights.
The behavioral difference between newer Nvidia gpu architectures like Turing and older ones like pascal is that for newer gpus, the 16 LUT fetches return the correct value  0 to ~0.15 (the sampled weight) with no vulkan comparison being made. On Pascal, these fetches return 1.0 (the Always comparison result), which results in saturated white squares instead of proper bloom.
The fix is to derive the _C bit for each instruction from the existing 64bit MIMG instruction decode as `IR::TextureInstInfo::is_depth` and save the `is_depth` result to a bool variable,` SamplerResource::is_depth`. The vulkan depth comparison is refactored to `.compareEnable=is_depth` so that only sampler bindings used by _C instructions will have vulkan depth comparison enabled. A dedup predicate was added so that the shader has slots for two samplers when the same S# is used for both plain and compare reads, which allows a separate sampler object to be used for each case. A sampler hash change in the texture cache was added so that when two sampler objects are requested, a separate sampler object is created for each case. 

CUSA04943_20260910_200050_728_game_000000

Before (Pascal GPU)

CUSA04943_20260911_083227_707_game_000000

After (Pascal GPU)

### DIFF
--- a/src/video_core/texture_cache/texture_cache.cpp
+++ b/src/video_core/texture_cache/texture_cache.cpp
@@ -804,11 +804,15 @@ void TextureCache::RefreshImage(Image& image) {
 }
 
 vk::Sampler TextureCache::GetSampler(const AmdGpu::Sampler& sampler,
-                                     AmdGpu::BorderColorBuffer border_color_base) {
-    const u64 hash = XXH3_64bits(&sampler, sizeof(sampler));
+                                     AmdGpu::BorderColorBuffer border_color_base,
+                                     const bool is_depth) {
+    // Compare and plain uses of one S# need separate samplers.
+    const u64 hash =
+        XXH3_64bits(&sampler, sizeof(sampler)) ^ (is_depth ? 0x9E3779B97F4A7C15ULL : 0);
 
     std::scoped_lock lock{samplers_mutex};
-    const auto [it, new_sampler] = samplers.try_emplace(hash, instance, sampler, border_color_base);
+    const auto [it, new_sampler] =
+        samplers.try_emplace(hash, instance, sampler, border_color_base, is_depth);
     if (new_sampler) {
         samplers.at(hash).lru_id = sampler_lru_cache.Insert(hash, gc_tick);
     } else {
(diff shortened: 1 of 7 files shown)


src/video_core/texture_cache/texture_cache.cpp
```diff
@@ -804,11 +804,15 @@ void TextureCache::RefreshImage(Image& image) {
 }
 
 vk::Sampler TextureCache::GetSampler(const AmdGpu::Sampler& sampler,
-                                     AmdGpu::BorderColorBuffer border_color_base) {
-    const u64 hash = XXH3_64bits(&sampler, sizeof(sampler));
+                                     AmdGpu::BorderColorBuffer border_color_base,
+                                     const bool is_depth) {
+    // Compare and plain uses of one S# need separate samplers.
+    const u64 hash =
+        XXH3_64bits(&sampler, sizeof(sampler)) ^ (is_depth ? 0x9E3779B97F4A7C15ULL : 0);
```
raphaelthegreat: Can you use HashCombine from common/hash.h instead?

raphaelthegreat (approved):
Looks good, just a minor style nit

PR #5021, opened 2026-09-13
shader_recompiler: fix SCC computation for S_AND/OR/XOR/NOT_B64

Gih-pt:
SetScc was using InverseBallot(result), valid only for genuine per-lane ballot masks (VCC/EXEC-shaped). These SOP2 opcodes are also used for generic 64-bit scalar bitwise logic, where hardware defines SCC simply as 'result != 0'. Mirrors the existing 32-bit SetScc(INotEqual(...)) pattern.

It may or may not fix the crash reported by @Kravickas in Marvel's Avengers (CUSA14030). This refers to comment #5011.

Note: AI was used for code generation and investigation; the changes were reviewed locally.

I think this PR needs a bit more review because I'm not sure if this is the best approach; let me know if it's a good change.


raphaelthegreat:
This wont fix the reported issue, as the constant is an argument of a S_AND_SAVEEXEC_B64 instruction. While changing SCC to not use InverseBallot is semantically correct its gonna break many things due to how the recompiler currently translates branch conditions, so this isnt mergeable without a control flow graph rewrite

Gih-pt:
Do you think I should close the PR now or wait for the control flow graph rewrite?

Gih-pt:
I'm going to close this PR and reopen it when this can be done properly.

PR #5023, opened 2026-09-13
shader_recompiler: Implement S_SETVSKIP

Gih-pt:
Implements the missing S_SETVSKIP scalar opcode in the shader recompiler, part of the aggregate missing-opcode tracker in #496. 

Note: AI was used for code generation and investigation; all changes were validated against the AMD GCN ISA Reference (SOPC instructions). 

This opcode is hit by SAMURAI WARRIORS 4 (CUSA00783).


red-prig:
So what exactly is the implementation? you just skip it, and in the meantime S_SETVSKIP significantly changes the behavior of subsequent instructions, so this PR is unacceptable

Gih-pt:
@red-prig S_SETVSKIP does modify a hardware mode state, but its effect is only on VALU instruction issue. This implementation does not model that state, so vector instructions continue executing normally. This may result in extra VALU work, but does not change the program results.

red-prig:
Actually, it does have a significant impact, because VALUs can both write to buffers and output data to the pixel shader; so, a full implementation within the flow graph is required.

Gih-pt:
Do I have to wait for the control flow graph rewrite to do this, or can I adapt my implementation for it now?

raphaelthegreat:
you probably dont have to wait for a rewrite, can you examine what the VSKIP usage is, if it is of this pattern
```
S_SETVSKIP S0 S1
...
S_SETVSKIP -1, 0 // or any other combination of S0, S1 that always enable it
```
then you can implement this in SplitDivergenceScopes in control flow graph

Gih-pt:
@raphaelthegreat I'll check that.

Gih-pt:
> you probably dont have to wait for a rewrite, can you examine what the VSKIP usage is, if it is of this pattern
> 
> ```
> S_SETVSKIP S0 S1
> ...
> S_SETVSKIP -1, 0 // or any other combination of S0, S1 that always enable it
> ```
> 
> then you can implement this in SplitDivergenceScopes in control flow graph



@raphaelthegreat Done. Sorry for the late reply; things have been busy lately.

raphaelthegreat:
Can you show the ISA pattern being hit here, so we can confirm if this fix is appropriate handling for it

Gih-pt:
@raphaelthegreat I don't have the game myself, but I dug into the original crash log from SAMURAI WARRIORS 4, Posted by @StevenMiller123. The decode-time assert hits right after `Compiling cs shader 0xd8d791029a1ba418` — so it's a compute shader, hash 0xd8d791029a1ba418. 

Unfortunately, this involves having the .bin file—which is an issue due to repository rules and copyright material. It would be great if @StevenMiller123 could help in some way, but I cannot confirm the full pattern without the shader dump file.

StevenMiller123:
As is described in the aggregate issue, I just cross-posted the missing opcode from a compatibility listing.
@brad0demx is the one who has the affected title.

Gih-pt:
Sorry @StevenMiller123, I don't know how I missed that detail. I’d like to ask if it’s possible to make an exception so that @brad0demx can publish this specific file, or if there is another way to validate this implementation.

PR #5025, opened 2026-09-13
Fix a crash in the case of the guest trying to open a file with a valid mount point but an invalid malformed unicode string in the relative to mount part of the path string

kalaposfos13:
I encountered this issue while messing around with VSH.


src/core/file_sys/fs.cpp
```diff
@@ -261,7 +262,8 @@ std::filesystem::path MntPoints::GetHostPath(std::string_view path, bool* is_rea
                     path_parts.pop_back();
                 };
                 // Can happen when the mismatch is in upper folder.
-                if (std::filesystem::exists(current_path / part)) {
+                std::error_code ec;
```
Niram7777, line 270: in case of error it should be log ?
(same for other one above)

Niram7777 (approved):

PR #5035, opened 2026-09-14
Config and input fixes

### LINKED ISSUE #4853: [APP BUG]: Config file can become corrupted when shadPS4 is terminated
### Checklist

### Describe the Bug

Sometimes the shadPS4 global and/or game-specific configs becomes corrupted after closing the emulator.

I'm using shadPS4 through SteamOS, and this seems to happen when I use Steam's "Close Game" option, especially after shadPS4 has crashed or frozen.

I don't have to change any settings before this happens.

From looking at the code, `EmulatorSettings` can save the config during shutdown, and the config is written directly to the file. If the process is killed while the file is being written, it could leave a partially written/corrupted config.

### Reproduction Steps

I don't have a reliable reproduction yet, but the issue seems to happen like this:

1. Start shadPS4 and a game.
2. Play normally without changing any settings.
3. If shadPS4 crashes/freezes, use Steam's "Close Game" option.
4. Start shadPS4 again.
5. Sometimes `config.json` is corrupted and has to be restored/deleted.

### Expected Behavior

The config should never become corrupted when the emulator is terminated.

It would probably be safer to write the config to a temporary file and rename it after the write succeeds, so an interrupted write cannot leave an invalid config.

### Specify OS Version

SteamOS 3.8.16

### CPU

AMD Custom CPU 1772

### GPU

AMD Radeon Graphics (RADV NAVI33)

### Amount of RAM in GB

16 GB

### Log File

_No response_

kalaposfos13:
This fixes config corruption on close, replaces errors with unreachables for failing to load json config files, cleans up config init code, and fixes a bug where some hotkeys accidentally triggered both on keydown and on keyup events.

Closes #4853


src/core/memory.cpp
```diff
@@ -1108,13 +1109,18 @@ s32 MemoryManager::Protect(VAddr addr, u64 size, MemoryProt prot) {
...
     // Ensure the range to modify is valid
     std::scoped_lock lk{mutex, unmap_mutex};
-    ASSERT_MSG(IsValidMapping(addr, size), "Attempted to access invalid address {:#x}", addr);
 
     // Appropriately restrict flags.
     constexpr static MemoryProt flag_mask =
         MemoryProt::CpuReadWrite | MemoryProt::CpuExec | MemoryProt::GpuReadWrite;
     MemoryProt valid_flags = prot & flag_mask;
 
+    if (!IsValidMapping(addr, size)) {
```
StevenMiller123: I don't think this can be accepted 😆

StevenMiller123 (changes requested):
Also needs a clang fix

PR #5036, opened 2026-09-14
Make userfaultfd configurable at runtime

mikusp:
As part of userfaultfd restoration, it changes how it is enabled from a compile time flag to a runtime one, facilitating easier debugging and testing.

Creates a new flag --userfaultfd that enables userfaultfd as the method of tracking memory. Obviously it's Linux only. If we ever end up using features not present in the user's kernel, it will fallback to the original (and current) implementation based on SIGSEGV


src/main.cpp
```diff
@@ -88,6 +89,8 @@ int main(int argc, char* argv[]) {
 
     app.add_flag("--wait-for-debugger", waitForDebugger);
     app.add_option("--wait-for-pid", waitPid);
+    app.add_flag("--userfaultfd", userfaultfd,
```
StevenMiller123, line 93: Might not be a bad idea to if-def away the add_flag call for non-Linux platforms.
That way Windows users don't see it in the help text and stuff

src/core/emulator_settings.h
```diff
@@ -215,6 +215,7 @@ struct GeneralSettings {
     Setting<std::string> shadnet_webapi_server{"http://srv.shadps4.net:31315"};
     Setting<std::string> signaling_info{};
     Setting<bool> enable_upnp{true};
+    Setting<bool> userfaultfd{false};
```
StevenMiller123: I think this would fit better in our GPU settings, as opposed to having it in general.

src/core/emulator_settings.h
```diff
@@ -759,6 +760,13 @@ class EmulatorSettingsImpl {
         }
     }
 
+    bool GetUserfaultfdTracking() {
+        return m_general.userfaultfd.get();
+    }
+    void SetUserfaultfdTracking(bool userfaultfd = false) {
+        m_general.userfaultfd.value = userfaultfd;
+    }
```
StevenMiller123: Also can just use SETTING_FORWARD_BOOL, that macro defines a set and get function for each setting if I remember right.

kalaposfos13 (commented):
It's not really a code thing per se, more like a bookkeeping one, but I kinda don't like how this PR hides the fact that it does more than just make the page manager backend configurable, but also fixes some long-standing bugs with our userfaultfd backend and makes use some new feature flags as well, you should at least mention it in the PR name and maybe also edit the description to document what you did.

kalaposfos13 (changes requested):
Currently there is no way to distinguish which page manager backend is in use from the log, it should be logged ideally.

kalaposfos13 (approved):

georgemoralis (approved):

PR #4183, opened 2026-03-27
[WIP] Out of order flip IRQ fix

Kravickas:
**Crash:** 
`ASSERT_MSG(port->buffer_labels[buf_id] == 1, "Out of order flip IRQ")`

**Main:** 
sceVideoOutSubmitEopFlip pushes a callback into a FIFO (RegisterOnce). The callback has buf_id baked in. When GPU finishes, Signal(GfxFlip) grabs the first one in line. With multiple flips queued, that's not necessarily the one that belongs to the submission that just finished — buffer_labels[buf_id] isn't set for it, assert crashes.

**PR:**
PatchFlipRequest still writes buf_id and flip_arg into the NOP data block in the command buffer. But instead of pushing a callback into the FIFO via RegisterOnce, the flip data is passed to SubmitGfx as optional<FlipRequest> and stored alongside the task in Liverpool's submission queue. When a submission completes (task.done()), Liverpool checks if that specific submission has a flip attached. If it does, it calls EnqueueFlip with that submission's own buf_id and flip_arg — not whatever happens to be at the front of a shared queue.

Main: NHL 19 - playing for ~10min - ooof IRQ assert.
PR: NHL 19 - playing for 1h 50min - no crash.

Thoughts ?
Players can try to break it.

PR #5039, opened 2026-09-16
[testers needed] gnmdriver: fix out-of-order flip IRQ race in EOP flip submission

Gih-pt:
Updated fix for the "Out of order flip IRQ" assertion failure in video_out.cpp (sceVideoOutSubmitEopFlip), initially reported in #4183, this PR needs testing because several games are affected by this crash.

The assert fires because the GfxFlip callback (registered via IrqC) and the command buffer that triggers it (enqueued via Liverpool) live in two independent FIFOs with no shared lock, so their relative order can drift apart when flips are submitted from more than one thread.

Instead of removing the IRQ mechanism, this serializes "register the callback" and "enqueue its command buffer" into one atomic step with a single mutex around sceGnmSubmitAndFlipCommandBuffersForWorkload. The two FIFOs can no longer drift apart, and the rest of the architecture is untouched.

Note: AI was used for code generation and investigation of the original PR from @Kravickas. Also verified against the crash logs, and confirmed GfxFlip has no other emitters/handlers,


StevenMiller123 (changes requested):
Remove the giant AI generated comments please

Gih-pt:
I'm going to remove them all for now; let me know if any comments are needed.

StevenMiller123:
With how small this is, I don't think a comment is necessary. If you want to put one, write a short line about the change in your own words.

Gih-pt:
> Remove the giant AI generated comments please

@StevenMiller123 done

StevenMiller123:
I'm looking in decomp, there seems to be several other places where the real library does a mutex lock and we don't. Could you look into that and see if there are other places worth locking in?

StevenMiller123:
For the record, this PR does not fix the flip IRQ issues. Still a valid fix, but it's not the fix for this issue.
[CUSA37521.log](https://github.com/user-attachments/files/32313451/CUSA37521.log)

bigol83:
Dark Souls 3 test

crash with out of order flip irq in three minutes

[CUSA03365.log](https://github.com/user-attachments/files/32333197/CUSA03365.log)

tochnonechelovek:
No regressions and changes in my titles.

Gih-pt:
The reported crashes might be fixed now; please also let me know if the new comment is acceptable. Regarding @StevenMiller123's comment, I would appreciate it if you could point out those other locations where the real library performs a mutex lock but the emulator doesn't yet.

src/video_core/amdgpu/liverpool.cpp
```diff
@@ -729,7 +729,18 @@ Liverpool::Task Liverpool::ProcessGraphics(std::span<const u32> dcb, std::span<c
                 const u32 data_size = (header->type3.count.Value() - 2) * 4;
                 u64* address = write_data->Address<u64*>();
                 if (!write_data->wr_one_addr.Value()) {
-                    std::memcpy(address, write_data->data, data_size);
+                    if (vo_port->IsVoLabel(address)) {
+                        // Flip label writes race with VideoOutDriver::Flip(), which resets
```
StevenMiller123, line 733: What little I've found so far suggests this is not the cause of the issue.

StevenMiller123:
I have a local branch with a proper fix for the locking issue in libSceGnmDriver, I'll make a PR for what I've got.
Flip IRQ seems to be tied to something weird going on in submitted command buffers. I still need to figure out what causes it, but it's definitely not a race with flip logic at this point (judging from both UE and Dragon Ball Xenoverse 2, can't judge the Dark Souls issues since I don't own that title).

Gih-pt:
Do you think the fixes in this PR are good enough, or should I close it in favor of yours? (I still need to check if Dark Souls has improved.)

StevenMiller123:
I need someone more familiar with the VideoOut logic to take a look, I'm not sure if there's value in the extra mutex locking you've done there.

Gih-pt:
@StevenMiller123 I'm going to close this in favor of your PR #5043; thanks for the time and review work.

Gih-pt:
Reopening because I noticed that #5043 doesn't completely solve the problem.

StevenMiller123:
Retest my PR with my newest push. From what I've seen locally I seem to have fixed the flip IRQ issue, though it needs more testing.

PR #4414, opened 2026-05-14
Neo: MIMG instructions

mikusp:
Neo adds a couple of new features for MIMG encoding:
- a16/d16 bits
- new modifier _G16 used for gradients(?)

PR #5041, opened 2026-09-17
Implement Neo extensions for MIMG instructions

mikusp:
Reintroduces https://github.com/shadps4-emu/shadPS4/pull/4414 after it has been reverted due to regressions in https://github.com/shadps4-emu/shadPS4/pull/4902.

The fix is to ignore the _G16 bit when Neo mode is not enabled


src/shader_recompiler/frontend/decode.cpp
```diff
@@ -1200,7 +1200,7 @@ u32 GcnDecodeContext::getMimgModifier(OpcodeMIMG opcode) {
 void GcnDecodeContext::decodeInstructionMIMG(uint64_t hexInstruction) {
     u32 op_msb = bit::extract(hexInstruction, 0, 0);
     u32 op = bit::extract(hexInstruction, 24, 18);
-    op = op + (op_msb * (1 << 7));
+    op = op + (Libraries::Kernel::sceKernelIsNeoMode() ? op_msb * (1 << 7) : 0);
```
squidbus, line 1203: Is there not some hardware register or something for this if it needs to be ignored? I.e. how does this work for a non-Neo game running on a Neo GPU?

PR #5043, opened 2026-09-17
Lib.GnmDriver: Submit fixes

StevenMiller123:
Fixes some discrepancies with libSceGnmDriver HLE and kernel behaviors. Inspired by #5039.

Was trying to dig into flip IRQ issues, but those seem to be tied to some weirdness in guest command buffers (and as such, are not fixed by this PR). When I was debugging, I found that the submitted command buffer has an unexpected PrepareFlip Nop packet before the WriteData packet responsible for writing VideoOut buffer labels. That fires the flip signal before the buffer label can be written, triggering the assert.


StevenMiller123:
Newest push should fix flip IRQ asserts. Tested in Dragon Ball Xenoverse 2 (while based on an older build to get past a recent regression), could use testing in other games.

StevenMiller123:
For reference, when I was debugging the weird extra nop thing, I found that the nop that causes the "out-of-order flip IRQ" assert was one written by sceGnmDrawInitToDefaultContextState400.

Specifically, these games (whether by coincidence or intentional behavior) overwrite old command buffers with this init state. In rare (but sometimes consistent) cases, this would overlap such that the final nop in the init state packet would overwrite a patched flip nop. In these circumstances, because our init state arrays were missing a zero, the data label from the original nop would be left intact, and end up triggering a patched flip early in the command buffer.

Of course, I checked both the base and neo libSceGnmDriver libraries to confirm, and they do write this extra zero. You won't see it in Ghidra unless you manually retype the cmdbuf parameter to a uint* though :smile:

Gih-pt:
This is what should be called a rare fix.

bigol83:
I played almost 3 consecutive hours on dark souls 3, no crash at all. 

Just had that out of device memory crash, just once. The game has some vertex explosions, but right now it's 10 times more enjoyable to play. 👍

PR #5044, opened 2026-09-17
Do not check library version and symbol type when searching for exports in the linker

kalaposfos13:
Apparently, the PS4 does not take these fields into account for linking, so we shouldn't either. The library version has so far not been an issue as far as I know, but dropping the symbol type check stops an OpenOrbis bug where every export is a function export, even object ones, from failing to link non-function symbols.

This issue was encountered while trying to get OpenOrbis's Piglet sample working.


src/core/linker.cpp
```diff
@@ -32,6 +32,8 @@
 #include <signal.h>
 #endif
 
+#pragma optimize off
```
StevenMiller123: Naturally can't leave this in.

src/core/loader/symbols_resolver.cpp
```diff
@@ -35,13 +46,18 @@ void SymbolsResolver::DebugDump(const std::filesystem::path& file_name) {
...
-                                  symbol.virtual_address, ids.at(0), nid_name, ids.at(1), ids.at(2),
-                                  ids.at(3), ids.at(4)));
+                                  symbol.virtual_address, id.name, nid_name, id.library,
+                                  id.library_version, id.module, SymbolTypeToS(id.type)));
     }
 }
 
+bool SymbolResolver::operator==(SymbolResolver const& o) const {
+    // not checking the version and type as they're ignored (second pending confirmation)
+    return this->name == o.name && this->library == o.library && this->module == o.module;
```
StevenMiller123, line 60: Needs to be confirmed before merge in my opinion

kalaposfos13: It's already confirmed, I just forgot to remove that part from the comment. :jim:

StevenMiller123 (approved):
I think this is fine, assuming nothing breaks.

red-prig:
I’ll add that there are two particular features:
```
 STT_NOTYPE        = 0; //module_start/module_stop
 STT_SCE           =11; //module_start/module_stop 
```
STT_NOTYPE -> this is not a hashed string but is stored as the original name (observed only for module_start/module_stop)
STT_SCE -> this is a hashed string but does not contain module library indices (observed only for module_start/module_stop; e.g., 'BaOKcng8g88')

kalaposfos13:
@red-prig I didn't add STT_SCE handling, simply because that in theory should hit an unreachable in our current code, and given that I have never seen that unreachable ever being hit, it's safe to say it doesn't matter.

red-prig:
It's hard to say whether this will cause problems on its own: module_start/module_stop are rare, but they need to be initialized correctly.

src/core/linker.cpp
```diff
@@ -391,15 +391,22 @@ void Linker::Relocate(Module* module) {
 bool Linker::Resolve(const std::string& name, Loader::SymbolType sym_type, Module* m,
                      Loader::SymbolRecord* return_info) {
     const auto ids = Common::SplitString(name, '#');
-    if (ids.size() != 3) {
+    const LibraryInfo* library = m->FindLibrary(ids[1]);
```
red-prig, line 394: One more thing: very, *very* rarely, I come across indices longer than a single character (if I recall correctly, I saw this in *Cat Quest*).

kalaposfos13:
> It's hard to say whether this will cause problems on its own: module_start/module_stop are rare, but they need to be initialized correctly.

We'ven't been handling module_start at all as far as I'm aware (and haven't been doing any amount of shutdown procedures for that matter), and yet it worked out fine for now, for module_start I assume I need to manually dlsym it and call it somewhere in the neighbourhood of Module::Start() (provided it exists in the module to be loaded)?

red-prig:
> > It's hard to say whether this will cause problems on its own: module_start/module_stop are rare, but they need to be initialized correctly.
> 
> We'ven't been handling module_start at all as far as I'm aware (and haven't been doing any amount of shutdown procedures for that matter), and yet it worked out fine for now, for module_start I assume I need to manually dlsym it and call it somewhere in the neighbourhood of Module::Start() (provided it exists in the module to be loaded)?

Yes and Yes

kalaposfos13:
Is it called before or after the entry point?

red-prig:
> Is it called before or after the entry point?

In the original firmware, libkernel is called first, which does everything before the game's entry point, So, it turns out that before this...

Like:

```
DT_INIT
if dl_sym(module_start) module_start();
```

PR #5047, opened 2026-09-18
buffer_cache: Refactor buffer manager with sparse arenas

raphaelthegreat:
The buffer cache is responsble for mirroring the GPU accessed address space in device local memory. On main it uses a simple scheme where the address space is divided into 16KB pages. Buffers are allocated with that granularity and all pages of that buffer point to the same buffer object. When a range overlaps multiple buffers, a join overlap operation is performed where a new buffer is created and the old buffer contents are copied over to the new one.

This scheme is quite simple and works well but has some issues

* It requires a 2-pass approach to buffer binding, because different bindings in the same draw/dispatch can cause a join overlap between them. If one binding was a storage buffer being written to, then data would be lost otherwise. This also causes various other issues when trying to batch memory uploads

* Join overlap can cause large VRAM usage spikes and requires heuristics to avoid repeated recreations. If the cache holds a large 1GB+ buffer, and then a range that partially overlaps it is accessed, the full 1GB range + the non overlapped size is going to get allocated. The old buffer will eventually be freed but this is still a large VRAM spike that can OOM or force system memory usage on lower VRAM gpus

To resolve these issues a new buffer management scheme is proposed in this pull request, that uses sparse residency buffers. The address space is divided into 2^32 VA arenas, and each arena holds a certain amount of blocks/tiles which are determined by the sparse bind granularity of the host GPU.

The arena size has been chosen because some lower end apple sillicon GPUs only support creating buffers up to 4GB, even though the GPU hardware support much large address spaces. Currently the code assumes that arena merges are always successfull, but that case will fail on those GPUs, for that bridge VA buffers need to be created and cached. However accesses that straddle arenas are expected to be rare enough for this to not be an actual issue in practice.

Any buffer access is checked and overlapping blocks are made resident on demand by allocating memory. Residency requests are batched and submitted before scheduler submits the main command buffer. If an access overlaps two adjacent arenas, then the arenas are merged into a single larger one and all their backings are brought up.

This new design avoids the old issues. There is no join operation aside from arena joining, which doesnt allocate any memory and is mostly for bookeeping purposes. The buffer objects are also stable now so the 2-pass approach has been dropped.

For tracking residency a custom sorted vector based interval set has been added which is optimized for fast queries, as checking if a range is resident is in quite a hot path.

Before the PR is mergeable some way to improve barrier tracking is required


squidbus:
Note: This isn't expected to work on Mac yet, there is some work I need to do to put in the features of sparse we need.

Xcedf:
Killzone Shadow Fall goes ingame no problem, no more OOM errors, just require to disable some geometry and tess shaders
image

metr1k-91:
> Killzone Shadow Fall goes ingame no problem, no more OOM errors, just require to disable some geometry and tess shaders image

Unrelated, but if possible give me a shout on discord, I'm working on something related to KZSF and was curious how you accomplished this.

Thanks.

EmulationChannell:
Captura de tela 2026-09-20 125836

EmulationPsycho:
It's not perfect yet, but it's a significant improvement over P.T.

Master:
Captura de Tela (856)

PR:
Captura de Tela (853)

Master:
Captura de Tela (857)

PR:
Captura de Tela (854)

Master:
Captura de Tela (858)

PR:
Captura de Tela (855)

EmulationPsycho:
It also fixes the intro videos for Alienation and Matterfall they don't jitter anymore!

Captura de Tela (865)

EmulationChannell:
Captura de tela 2026-09-21 045918

EmulationChannell:
Captura de tela 2026-09-21 161819

PR #5053, opened 2026-09-18
libc_internal: implement sceLibcMspace family, __cxa_guard, _Getpctype

Marco-marcheseragona:
## Summary
MediEvil (CUSA12982) crashes on every boot right after NpToolkit2 maps its private memory pool ("PSN: Failed to initialize the np toolkit" -> LowLevelFatalError -> unhandled access violation in the signal handler). Root cause traced to several completely unimplemented libSceLibcInternal symbols that this game's boot path depends on:

- `sceLibcMspaceCreate`/`Malloc`/`Free`/`Destroy` (plus `Calloc`/`Realloc`) were registered in the NID database but never implemented, always returning 0/NULL. NpToolkit2 calls `sceLibcMspaceCreate` on its freshly-mapped flexible memory region to build a private heap; getting NULL back makes it conclude initialization failed.
- `__cxa_guard_acquire`/`Release`/`Abort` (the Itanium C++ ABI guard protocol for thread-safe function-local `static` initialization) were stubbed to always report "already initialized", so any function-local static with non-trivial construction was used **uninitialized** on first access — a likely source of similar crashes in other titles, not just this one.
- `_Getpctype` (ctype classification table used by isalpha/isdigit/etc.) was unimplemented, hit once boot reaches the sign-in dialog module.
- `stubs.cpp`'s `MAX_STUBS` (8192) is exhausted by MediEvil's large UE4 import surface; once exceeded, further unresolved imports silently fall back to an anonymous catch-all stub with no name/NID logged, which made this bug much harder to diagnose. Raised to 16384.

## Changes
- `src/core/libraries/libc_internal/libc_internal_memory.cpp`: minimal first-fit allocator implementing the sceLibcMspace* family over a caller-provided memory region (mirrors the dlmalloc-style mspace API the Sony wrapper is based on); `__cxa_guard_*` using a recursive mutex; `_Getpctype` returning a standard ctype table.
- `src/core/aerolib/stubs.cpp`: `MAX_STUBS` 8192 -> 16384.

## Result
MediEvil now boots past the intro cinematic into gameplay instead of crashing on every launch (previously reproduced identically across shadps4 v0.17.0/v0.18.0/current main and the ShadPs4Plus fork).

## Test plan
- [x] Built with `x64-Clang-Release` preset, zero warnings/errors introduced
- [x] MediEvil (CUSA12982, both base v1.00 and patched v1.02) launches, plays intro cinematic, reaches gameplay on Linux/RADV (RX 5700)
- [ ] Would appreciate testing on other titles that use function-local statics / mspace allocators to check for regressions


StevenMiller123:
We do not accept PRs that are entirely AI-generated. Some of the changes here are decent, but we expect all contributors to properly write PR descriptions, properly understand the code they're submitting, and properly revise the code they've generated to better follow our contributing guidelines.

Also, there is no point in bothering to test unofficial forks.

Issue #5050, opened 2026-09-18
[GAME BUG]: Shadwen CUSA04540 [Debug] <Critical> (shadPS4:GpuCommandProcessor) resource_discover_pass.cpp:257 MarkReadConstBufferSharpSources: Assertion Failed!


ScottNash042:
### Checklist (we expect you to perform these steps before opening the issue)

- [x] I have searched for a similar issue in this repository and did not find one.
- [x] I am using an official build obtained from [releases](https://github.com/shadps4-emu/shadPS4/releases) or updated one of those builds using its in-app updater.
- [x] I have re-dumped the game and performed a clean install without mods and the issue is still present.
- [x] I have disabled all patches and cheats and the issue is still present.
- [x] I have all the required [system modules](https://github.com/shadps4-emu/shadPS4/wiki/I.-Quick-start-%5BUsers%5D#4-dumping-firmware-modules) installed.

### Describe the Bug

crashes out after the very first screen
[Debug] <Critical> (shadPS4:GpuCommandProcessor) resource_discover_pass.cpp:257 MarkReadConstBufferSharpSources: Assertion Failed!

### Reproduction Steps

just load the game

### Specify OS Version

Windows 11

### CPU

AMD Ryzen 7 5700X 8-Core Processor | 16 Threads

### GPU

NVIDIA GeForce RTX 3060

### Amount of RAM in GB

32GB

### Amount of VRAM in GB

16GB

### Log File

[shad_log.txt](https://github.com/user-attachments/files/32375288/shad_log.txt)

Gih-pt:
@ScottNash042 Please test my new PR and post your results.

ScottNash042:
@Gih-pt 
anywhere to download a built version already

Gih-pt:
> [@Gih-pt](https://github.com/Gih-pt) anywhere to download a built version already

You can always test the latest builds from the published PRs in Actions; here is the most up-to-date one for my PR. If it improves or another error occurs, please post a log: https://github.com/shadps4-emu/shadPS4/actions/runs/35454327162

ScottNash042:
Thank you 
Will try it out later

ScottNash042:
just tried it 
still the same issue
here is the log

[shad_log.txt](https://github.com/user-attachments/files/32432543/shad_log.txt)

ScottNash042:
just tried your update
still crashes out with the same error but it starts to play music now but crashes out about a second after the music starts

[...]

PR #5059, opened 2026-09-19
shader_recompiler: recognize sharp sources broadcast via V_READFIRSTLANE_B32

Gih-pt:
Fixes a crash in MarkReadConstBufferSharpSources() where a sharp (V#/T#/S#) dword produced by V_READFIRSTLANE_B32 wasn't recognized as a valid sharp source, since v_readfirstlane_b32 is common for descriptors computed with per-thread arithmetic the compiler has proven is wave-uniform; per raphaelthegreat's suggestion, the fix unwraps ReadFirstLane right where V# dwords are gathered in DiscoverBufferSharp, so the rest of the pipeline sees the real producer with no other code changes needed. Confirmed against a real IR dump showing each dword is exactly ReadFirstLane(ReadConstBuffer(...)). (Closes issue [#5050](https://github.com/shadps4-emu/shadPS4/issues/5050))

AI was used for code generation and investigation; the fix was validated against the existing shadPS4 codebase itself, and against @ScottNash042's IR dump confirming the pattern before implementing the fix.


raphaelthegreat:
Well the assertion exists to catch weird usages, so I think the first attempt should be if you can handle the pattern instead rather than ignoring it

Gih-pt:
Well, that makes sense.

Gih-pt:
@raphaelthegreat Since I don't own this game, I'm not sure if I can properly handle the pattern and verify it the way I'd like; I think it's better to just issue a log warning for now and fully fix it in another PR when I'm able to.

Gih-pt:
> Well the assertion exists to catch weird usages, so I think the first attempt should be if you can handle the pattern instead rather than ignoring it


@raphaelthegreat The pattern is now properly handled and includes a fallback to prevent a crash. (I would like to thank @ScottNash042 for making it possible for me to complete the correction)

raphaelthegreat:
Can you describe a bit what pattern is being seen? Sharp loads are typically only allowed from ReadConst which is an SGPR so just seeing through a READFIRSTLANE doesn't sound enough to me.

Gih-pt:
@raphaelthegreat In Shadwen, the four dwords of a buffer V# are each produced by a ReadFirstLane, so MarkReadConstBufferSharpSources sees ReadFirstLane as the source instead of GetUserData/ReadConst/ReadConstBuffer. The attached log (before the pattern was handled) shows it: `Unrecognized sharp source instruction ReadFirstLane for dword 0..3`.

example

The patch doesn't accept ReadFirstLane as a source. It only follows the RFL chain to its producer and applies the existing IsSharpSource check to that.

[shad_log.txt](https://github.com/user-attachments/files/32434953/shad_log.txt)

raphaelthegreat:
Can you dump IR right before resource discover pass and inspect in IR dump what the full pattern looks like. It's not enough to handle just the ReadFirstLane you should confirm the handling is correct

Gih-pt:
Is it allowed to share shader dumps in this repository?

StevenMiller123:
No, shader binary dumps are considered copyrighted material. This is why we're asking for an IR dump instead. IR and SPV dumps are fine, since they're the product of us recompiling the game shader.

Gih-pt:
How do I extract an IR dump?

raphaelthegreat:
In src/shader_recompiler/recompiler.cpp add a call
```
Shader::IR::DumpProgram(program, info, "pre-res-discover."); // <- add this before ResourceDiscoverPass
auto resources = Shader::Optimization::ResourceDiscoverPass(program, profile);
```
and enable shader dumping in config.json. Then inspect the IR dump for those ReadFirstLane opcodes. Don't send the whole thing just the relevant small snippet. I recommend you try do this without AI to get a proper feel for the workflow of examining recompiler output

src/shader_recompiler/ir/passes/resource_discover_pass.cpp
raphaelthegreat, line 274: Can you try this change instead? If this works you can replace all the other code written with this
```suggestion
        vsharp.dwords[i] = handle->Arg(i);
        if (auto* inst = vsharp.dwords[i].TryInst(); inst && inst->GetOpcode() == IR::Opcode::ReadFirstLane) {
            vsharp.dwords[i] = inst->Arg(0);
        }
```

Gih-pt: @raphaelthegreat Done. Once again, I’d like to thank @ScottNash042 for allowing me to confirm the complete pattern as demonstrated here. 

%69 = ReadConstBuffer (sharp_source=false, flatbuf_off_dw=0x0) %68, #4
%70 = ReadConstBuffer (sharp_source=false, flatbuf_off_dw=0x0) %68, #5
%71 = ReadConstBuffer (sharp_source=false, flatbuf_off_dw=0x0) %68, #6
%72 = ReadConstBuffer (sharp_source=false, flatbuf_off_dw=0x0) %68, #7
...
%85 = ReadFirstLane %70
%87 = ReadFirstLane %71
%89 = ReadFirstLane %72
%91 = ReadFirstLane %69

...
%95 = CompositeConstructU32x4 %91, %85, %87, %89

[cs_0x000000008673ea57.pre-res-discover.irprogram.txt](https://github.com/user-attachments/files/32489523/cs_0x000000008673ea57.pre-res-discover.irprogram.txt)

raphaelthegreat: I see, with this context, your fix is correct. The 3 line change I recommended should be enough to resolve the issue, is a bit cleaner I think

raphaelthegreat (approved):
thx

PR #5060, opened 2026-09-19
Lib.GnmDriver: Remove extra wait in sceGnmSubmitDone

StevenMiller123:
Seems I accidentally misunderstood something I saw in my decomp of the submit done ioctl. This one change effectively made submits synchronous, which was good for local testing, but also effectively a hack.

As far as I can tell, fixing this oversight will only bring regressions. Should we make a config setting for this instead?


squidbus:
> As far as I can tell, fixing this oversight will only bring regressions. Should we make a config setting for this instead?

I feel like it should be fine as you have it in this PR if this is the intended behavior? Do you know of any unintended regressions here?

StevenMiller123:
Right now, the extra wait is a mostly harmless hack. My goal was to model a wait I saw in the submit done ioctl, but I accidentally made sceGnmSubmitDone block until all prior submits complete.
As far as I know, sceGnmSubmitDone is not supposed to block until prior submits finish (rather, games are supposed to wait for sceGnmAreSubmitsAllowed to return true, indicating that the GPU is idle after a sceGnmSubmitDone call).

That extra wait broke low framerate/hang detection in UE4 and Naughty Dog games, and prevents most of the PM4 type 0 races we had. I haven't seen any meaningful regressions, but it is still a hack, which is why I've opened this PR to revert that specific change from #5043.

squidbus (approved):

StevenMiller123:
Did not intend to do that 😅

PR #5062, opened 2026-09-20
host_shaders: embed shaders as spv

LNDF:
Embed host shaders as SPV insteed of embeddin the GLSL and compiling them at runtime.
Some values are now specialization constants insteed of preprocessor macros.

Additionaly, a bug with the tile manager is fixed:

> the last days I have been looking and i think there are some mistakes:
> The index for the tilers pipelines array is computed as:
> const u32 pl_id = u32(info.tile_mode) * NUM_BPPS + std::bit_width(info.num_bits) - 4;
> 
> however, we also use this define:
> fmt::format("NUM_SAMPLES={}", info.num_samples),
> 
> Which we don't count for the index.
> Also the bit_with for 64bpp and 96bpp would be the same, so they would colide.
> If this are bugs, i can resolve them in a pr i have about host shaders. This is the last thingg i have pending.

Also, maibe I need to add back the #define logs I removed from tile_manager.cpp...

AI disclosure: Used AI to help with thigs related to cmake scripts.


raphaelthegreat:
Doesn't glslang support compiling a GLSL file into SPIRV and generating a header with the bytecode, is that insufficient for this usecase?

LNDF:
> Doesn't glslang support compiling a GLSL file into SPIRV and generating a header with the bytecode, is that insufficient for this usecase?

Didn't know about that feature. Will check it out.

Keep in. Mind however that we  need to track cmake dependencies. E.g. if s shader includes a header file and that header file is modified, clske should rebuild trigger the shader rebuild.

Currently that is done with a custom included in the shader compiler program

LNDF:
Maybe stil pending to add #include <stdint> to all generated headers

raphaelthegreat (approved):

PR #5069, opened 2026-09-21
Uffd fix iii

mikusp:
More userfaultfd fixes:
- now errors returned from UFFDIO_{,UN}REGISTER are logged
- the handler thread sets it's own name
- the whole memory range is registered at all times

The most substantial change in this PR is how exactly registrations work. Previously, a memory range that was determined to be a mapping usable by the GPU (IsValidGpuMapping()). That caused issues with titles that try to watch memory that may be partially unmapped (as evidenced by a "Tracking memory region {:#x} - {:#x} which is not fully GPU mapped." line in the log), as then at least one of the VMAs spanning the range would not be registered with the userfaultfd, and as such could not be writeprotected. mprotect/SIGSEGV handling did not have that problem.

Now, the emulator maintains the whole reserved memory range registered at all times, registering it when starting the userfault handler thread, and then reregistering every new VMA that is created by mmap calls. Due to how memory unmapping tries to minimise the number of VMAs maintained for the process, the mmap call might span more memory than requested, so Unmap functions have been modified to return the real range being modified. Additionally, no memory range gets unregistered as that does not seem to be beneficial in any way, and might limit the number of ioctls issued.

Lastly, one title caused a hang during fault handling, when a write protection fault originating from GpuCommandProcessor ended up trying to send a GPU command and taking a semaphore in the process. Now gpu thread has been waiting for the fault to be handled, which in turn waited for the gpu thread to release the semaphore. Thanks to @raphaelthegreat's suggestion, a thread id of the faulting thread is now used to decide if sending a GPU commands may skip taking the semaphore, and progress without deadlocking the program.


src/video_core/amdgpu/liverpool.h
```diff
@@ -96,8 +96,8 @@ struct Liverpool {
     }
 
     template <bool wait_done = false>
-    void SendCommand(auto&& func) {
-        if (std::this_thread::get_id() == gpu_id) {
+    void SendCommand(auto&& func, bool handling_userfault_from_gpu_thread = false) {
+        if (std::this_thread::get_id() == gpu_id || handling_userfault_from_gpu_thread) {
```
raphaelthegreat: I think you should remove the whole "std::this_thread::get_id() == gpu_id" check from here, and BufferCache::ReadMemory should be the one to decide if it will use SendCommand or not

PR #5073, opened 2026-09-22
CI: Do not zip AppImages

mikusp:


.github/workflows/build.yml
```diff
@@ -300,15 +300,17 @@ jobs:
...
 
     - name: Run AppImage packaging script
-      run:  ./.github/workflows/scripts/linux-appimage-sdl.sh
+      run: |
+        ./.github/workflows/scripts/linux-appimage-sdl.sh
+        cp Shadps4-sdl.AppImage shadps4-linux-sdl-${{ needs.get-info.outputs.date }}-${{ needs.get-info.outputs.shorthash }}.AppImage
 
     - name: Package and Upload Linux SDL artifact
       run: |
         tar cf shadps4-linux-sdl.tar.gz -C ${{github.workspace}}/build shadps4
```
squidbus: Maybe slightly unrelated, but why are we still doing this tar-ing if we don't upload it?

mikusp: I wondered that myself 😅 @georgemoralis?

squidbus: We probably used to upload a tar here, I think you can just remove this.

brad0demx:
This PR seems to no longer synergise with the previous PR #3730, this means the resulting .AppImage artifacts are not marked as executable and don't function with the launcher out the box.

There is also the more launcher specific side effect/issue that I mentioned on discord with the new naming scheme of the .AppImage, some work might potentially need to be done with the version manager for that to remain functional with this change

squidbus:
> This PR seems to no longer synergise with the previous PR https://github.com/shadps4-emu/shadPS4/pull/3730, this means the resulting .AppImage artifacts are not marked as executable and don't function with the launcher out the box.

The launcher would have to mark them itself, there's no way to make it come executable without a container around it.

brad0demx:
Funnily enough I actually used to have a PR on the launcher to mark it as executable there, but it was decided by others that doing that here would be better. Is this just a nuance of not zipping it anymore? 

Also with it not being possible to mark as executable, should line 366 be removed from the file as well as the tar stage? 
`chmod -R a+x ./artifacts/shadps4-linux-sdl-*`

.github/workflows/build.yml
georgemoralis: i probably didn't do that i hate linux ;p

squidbus (approved):
Seems okay, but we need to have a corresponding launcher change to set it as executable for this to go in.

PR #5078, opened 2026-09-23
shader_recompiler: Fix geometry invocation count comparison

IfkumRfnl:
Fix a likely typo in `HwGeometryRuntimeInfo::operator==`: it checked whether both invocation counts were nonzero instead of comparing them for equality. Add a regression test.


raphaelthegreat:
I think the test is a little bit too specific (would be nice to have a test suite for these stuff for sure, but I think its better to leave it to another PR). Can you split the relevant fix into this PR to merge it

IfkumRfnl:
@raphaelthegreat thanks for looking into it, removed the tests, kept only the 1 line fix

PR #5081, opened 2026-09-23
Consider overlapping images differing only in pow2 compatible

mikusp:
Resolve image overlaps if the only difference is pow2pad.

Allows Stray to progress ingame
image

If you have a game that crashes on ResolveOverlap and logs sth like:
```
=== COMPARISON ===
Same format:           true
Same type:             true
Same tile mode:        true
Same block size:       true
Same BlockDim:         true
Same pitch:            true
Same pow2:             false
Same alt tile:         true
Old resources <= new:  true (old: 1, new: 1)
Old size <= new size:  true
```
please test it with this PR


EmulationChannell:
Captura de tela 2026-09-23 183059

raphaelthegreat (approved):
I think this is fine, for single layer and level images with same power-of-two pitch, the difference in computed size comes from the height which doesnt really matter anyway. In the upcoming texture cache rewrite there needs to be some consideration on how to integrate this more cleanly

PR #5083, opened 2026-09-23
Define V_CMP_*_I64

mikusp:
V_CMP_GT_I64 has been used by YouTube so I've defined all ops


src/shader_recompiler/frontend/translate/vector_alu.cpp
```diff
@@ -371,6 +371,24 @@ void Translator::EmitVectorAlu(const GcnInst& inst) {
     case Opcode::V_CMPX_TRU_U32:
         return V_CMP_U32(ConditionOp::TRU, false, true, inst);
 
+        //     V_CMP_{OP8}_I64
+    case Opcode::V_CMP_F_I64:
+        return V_CMP_U64(ConditionOp::F, false, false, inst);
```
StevenMiller123: is_signed bool should be true

StevenMiller123 (approved):

Gih-pt:
For obvious reasons, the emulator cannot connect to the official PSN server, but is it possible to connect to other servers, such as Google? I remember seeing YouTube and Netflix working on RPCS3 (or at least they used to), so I suppose the same question applies to ShadPS4.

PR #5087, opened 2026-09-23
shader_recompiler: Support P10/P20 V_INTERP_MOV_F32 selectors

w1naenator:


raphaelthegreat:
When using fragment shader barycentric, radv seems to force LDS passthrough mode for attributes:
https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/amd/vulkan/radv_cmd_buffer.c#L3888

So this is only correct when both flat_shade and offset5 are set, otherwise you need to construct P10, P20 with subtraction. I think it should be fine with that

PR #5090, opened 2026-09-23
shader_recompiler: implement fragment sample coverage input

w1naenator:


src/shader_recompiler/backend/spirv/emit_spirv_context_get_set.cpp
```diff
@@ -197,6 +199,18 @@ Id EmitGetAttributeU32(EmitContext& ctx, IR::Attribute attr, u32 comp) {
...
     case IR::Attribute::SampleIndex:
         return ctx.OpLoad(ctx.U32[1], ctx.sample_index);
+    case IR::Attribute::SampleCoverage: {
+        const u32 num_samples = std::clamp<u32>(ctx.runtime_info.hw.fs.num_samples, 1U, 16U);
+        Id coverage = ctx.u32_one_value;
+        if (num_samples > 1) {
+            coverage = ctx.OpLoad(ctx.U32[1], ctx.OpAccessChain(ctx.input_u32, ctx.sample_mask_in,
+                                                                ctx.u32_zero_value));
+            const u32 valid_samples = (1U << num_samples) - 1U;
+            coverage = ctx.OpBitwiseAnd(ctx.U32[1], coverage, ctx.ConstU32(valid_samples));
```
raphaelthegreat: This should be done in IR level I feel. Why does the helper invocation check exist here?

w1naenator: Ok, I'll move the sample coverage construction to IR and keep the SPIR-V backend limited to loading the raw builtins. The HelperInvocation check was intended to make coverage deterministically zero for helper lanes, since Vulkan does not guarantee that their SampleMask input is zero. I'll expose SampleMask and HelperInvocation separately and perform the mask/select in the fragment prologue

raphaelthegreat (approved):

